PR #800, opened 2020-07-08
Introduce conditional to activestate.yaml

Naatan:
https://www.pivotaltracker.com/story/show/173427846


Naatan:
Suggest filtering out the first commit when you review

internal/constraints/constraints.go
```diff
@@ -315,19 +385,35 @@ func IsConstrained(constraint projectfile.Constraint) (bool, int) {
...
+		logging.Error("FilterUnconstrained called with nil conditional")
+	}
+
 	for i, item := range items {
-		c := item.ConstraintsFilter()
-		constrained, specificity := IsConstrained(c)
-		if !constrained {
-			if s, exists := selected[item.ID()]; !exists || s.specificity < specificity {
-				selected[item.ID()] = itemIndex{specificity, i}
+		if conditional != nil && item.ConditionalFilter() != "" {
```
mdrohmann, line 400: Maybe we could print an error message, if both `ConditionalFilter` and `ConstraintsFilter` are provided.  I think it is the correct way to not mix them together, as we are deprecating the "constraints" anyways.  But if a user has defined both, they might wonder why their `constraints` are being ignored...

internal/constraints/constraints.go
```diff
@@ -1,21 +1,91 @@
...
+func (c *Conditional) RegisterFunc(name string, value interface{}) {
+	c.funcs[name] = value
+}
+
+func (c *Conditional) RegisterParam(name string, value interface{}) {
+	c.params[name] = value
+}
+
+func (c *Conditional) Eval(conditional string) (bool, error) {
+	tpl, err := template.New("letter").Funcs(c.funcs).Parse(fmt.Sprintf(`{{if %s}}1{{end}}`, conditional))
```
mdrohmann, line 78: :+1: for piggy-backing on the go template parser to implement all the complex parsing in one line.

internal/constraints/constraints.go
Naatan: This is easier said than done since `print` is deprecated and passing `outputer` in to this level would be pretty awkward. We could pre-validate the activestate.yaml for this type of stuff but that feels like overkill for the low severity of this use-case.

PR #801, opened 2020-07-08
Indicate cached projects

MDrakos:
https://www.pivotaltracker.com/story/show/173526829


internal/runners/projects/projects_test.go
```diff
@@ -23,14 +31,14 @@ func TestProjects(t *testing.T) {
...
+	pjs := NewProjects(catcher.Outputer, authentication.Get(), &configMock{})
 
 	projects, fail := pjs.fetchProjects()
 	assert.NoError(t, fail.ToError(), "Fetched projects")
 	assert.Equal(t, 1, len(projects), "One project fetched")
-	assert.Equal(t, "test project", projects[0].Name)
+	assert.Equal(t, "test project (test description)", projects[0].Name)
 	assert.Equal(t, "organizationName", projects[0].Organization)
-	assert.Equal(t, "test description", projects[0].Description)
+	assert.Equal(t, []string{"/some/local/path/"}, projects[0].LocalCheckouts)
```
MDrakos, line 41: Attempted to write integration tests for this however due to the string wrapping of the table output the test suite is not able to match the output easily.

cmd/state/internal/cmdtree/projects.go
```diff
@@ -6,10 +6,11 @@ import (
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/runners/projects"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/spf13/viper"
 )
 
 func newProjectsCommand(outputer output.Outputer, auth *authentication.Auth) *captain.Command {
-	runner := projects.NewProjects(outputer, auth)
+	runner := projects.NewProjects(outputer, auth, viper.GetViper())
```
Naatan, line 13: Fyi filed this story to address - https://www.pivotaltracker.com/story/show/173734115

internal/output/plain.go
```diff
@@ -263,6 +264,10 @@ func sprintTable(slice []interface{}) (string, error) {
 				stringValue = trimValue(stringValue, termWidth)
 			}
 
+			if funk.Contains(field.opts, string(EmptyNil)) {
+				stringValue = useEmptyNil(stringValue)
```
Naatan, line 268:
```suggestion
			if funk.Contains(field.opts, string(EmptyNil)) && stringValue == nilText {
				stringValue = ""
```

And delete `useEmptyNil`

internal/runners/projects/projects.go
```diff
@@ -55,19 +64,48 @@ func (r *Projects) fetchProjects() ([]projectWithOrg, *failures.Failure) {
...
+			if localPaths, ok := localConfigProjects[fmt.Sprintf("%s/%s", strings.ToLower(org.URLname), strings.ToLower(project.Name))]; ok {
+				p.LocalCheckouts = localPaths
 			}
-			projectsList = append(projectsList, projectWithOrg{project.Name, desc, org.Name})
+			orgProjects[i] = p
 		}
+
+		sort.Slice(orgProjects, func(i, j int) bool {
+			return orgProjects[i].LocalCheckouts != nil && orgProjects[j].LocalCheckouts == nil
+		})
```
Naatan, line 97: Run this outside of the loop on `projects`. No point sorting this every time an entry is added, only need to sort it once.

pkg/projectfile/projectfile.go
```diff
@@ -67,7 +67,7 @@ var strReg = fmt.Sprintf(`https:\/\/%s\/([\w_.-]*)\/([\w_.-]*)(?:\?commitID=)*(.
 // ProjectURLRe Regex used to validate project fields /orgname/projectname[?commitID=someUUID]
 var ProjectURLRe = regexp.MustCompile(strReg)
 
-const localProjectsConfigKey = "projects"
+const LocalProjectsConfigKey = "projects"
```
Naatan, line 70: We gotta find a better solution than just exposing the key, but I also don't want to start making the projectfile or project package a catchall for ANYTHING project.. Let's just keep it in mind for now, don't want to add scope.

internal/runners/projects/projects.go
MDrakos: Running this particular sort outside of the organizations loop will move all of the projects to the top of the table. One of the ACs was to keep the overall ordering of the table based on organizations.

Naatan: Let's make it:

```
return (orgProjects[i].LocalCheckouts != nil && orgProjects[j].LocalCheckouts == nil) ||
	orgProjects[i].Organization < orgProjects[j].Organization ||
	orgProjects[i].Name < orgProjects[i].Name
```

And move it outside the loop. That way it stays performant, the AC's are handled AND we get name sorting as a bonus.

Naatan:
@MDrakos tested it locally;

 - field_localcheckouts is not showing up as localized
 - local checkouts are not sorted on top

Naatan:
Also worth noting, Organizations are not sorted at all, the AC on that is wrong.

MDrakos:
@Naatan thanks for testing! I've added a `locale` tag to address the `Local Checkouts` field. 

As for sorting, the local checkouts should be at the top of their organization, not the top of the whole listing... at least that's how I read the AC. So the overall table is sorted by Org and within that the local projects are at the top. Confirmed that output in my local testing.

internal/runners/projects/projects.go
MDrakos: I modified your suggestion slightly to maintain the organization ordering. It is not sorted by name within the organizations but should maintain how we received it from the platform.

MDrakos: After testing this further it also can also break the organization ordering. I think we need to decide which ordering we want. 

If it should be ordered by organization then the initial sorting can be kept. The sort will only be called once for every organization that a user is a part of, which for most users I imagine won't be very high.

If we want the local projects at the top of the list and all others below that we can do one sort which should handle it.

Naatan: Let's not get hung up on it for this PR seeing as it's proving to be a challenge. Checkouts should be sorted on top as per the AC, the rest will have to follow in a separate story.

PR #803, opened 2020-07-09
improve relocation replacement performance

mdrohmann:
https://www.pivotaltracker.com/story/show/173566469

Here are results about how much the performance has improved:
 
https://docs.google.com/spreadsheets/d/1qNf1eptVhsTjx9WwB7H5WhARLMktfOKZkHjpaMuqbxA/edit?usp=sharing


internal/fileutils/fileutils.go
```diff
@@ -76,53 +74,22 @@ func ReplaceAll(filename, find string, replace string, include includeFunc) erro
...
-			// Pad replacement with NUL bytes.
-			//logging.Debug("Padding replacement text by %d byte(s)", len(findBytes)-len(replaceBytes))
-			paddedReplaceBytes := make([]byte, len(findBytes)+len(regexExpandBytes))
-			copy(paddedReplaceBytes, replaceBytes)
-			replaceBytes = paddedReplaceBytes
-		}
-	} else {
-		replaceRegex = regexp.MustCompile(fmt.Sprintf(`%s`, quoteEscapeFind))
-		//logging.Debug("Assuming file '%s' is a text file", filename)
+	count, byts, err := replacePathInFile(fileBytes, find, replace)
```
Naatan, line 76:
```suggestion
	count, byts, err := replaceInFile(fileBytes, find, replace)
```

Don't think this needs to have an opinion on what it is replacing.

internal/fileutils/fileutils.go
```diff
@@ -134,6 +101,9 @@ func ReplaceAllInDirectory(path, find string, replace string, include includeFun
 		return err
 	}
 
+	end := time.Now()
+	logging.Debug("ReplaceAllInDirectory took %v", end.Sub(start))
```
Naatan: Remove test code

internal/fileutils/fileutils.go
```diff
@@ -76,53 +74,22 @@ func ReplaceAll(filename, find string, replace string, include includeFunc) erro
...
+		return nil
+	}
 
-	return WriteFile(filename, buffer.Bytes()).ToError()
+	return WriteFile(filename, byts).ToError()
 }
 
 // ReplaceAllInDirectory walks the given directory and invokes ReplaceAll on each file
 func ReplaceAllInDirectory(path, find string, replace string, include includeFunc) error {
+	start := time.Now()
```
Naatan: Remove test code

internal/fileutils/replace.go
```diff
@@ -0,0 +1,178 @@
...
+		// fast forward to the end of the string
+		i = endOffset - 1
+	}
+	return count, buf, nil
+}
+
+// replacePathInFileRegex is the old implementation of replacePathInFile based on regular expressions.
+// It is now only used in the benchmark functions
+// Note that it always returns 1 for the count of replacements
+func replacePathInFileRegex(buf []byte, oldpath, newpath string) (int, []byte, error) {
```
Naatan, line 134: Drop this please, we have git history for this.

internal/fileutils/fileutils.go
```diff
@@ -114,11 +131,9 @@ func ReplaceAll(filename, find string, replace string, include includeFunc) erro
 		//logging.Debug("Assuming file '%s' is a text file", filename)
 	}
 
-	replaced := replaceRegex.ReplaceAll(fileBytes, replaceBytes)
-	buffer := bytes.NewBuffer([]byte{})
-	buffer.Write(replaced)
+	replaced := replaceRegex.ReplaceAll(buf, replaceBytes)
 
-	return WriteFile(filename, buffer.Bytes()).ToError()
+	return !bytes.Equal(replaced, buf), replaced, nil
```
mdrohmann, line 136: I did not use `ReplaceAllFunc` here, because the above version of this function already installs an ActivePython project as fast as the most optimized version that I just got rid off.  So, that tells us that optimizing this function has no impact at all.  Therefore, I think that I can avoid the refactoring that `ReplaceAllFunc` would have required.

PR #810, opened 2020-07-13
Updates need to be generated on right branch

Naatan:
https://www.pivotaltracker.com/story/show/173785481


.github/workflows-src/steps.lib.yml
```diff
@@ -103,6 +102,7 @@ with:
 name: "Validate"
 shell: bash
 run: |
+  echo "${{ toJson(github.event) }}"
```
Naatan, line 105: This step was randomly failing, and I can't get it to fail anymore, so leaving it in here for now.

PR #815, opened 2020-07-15
use absolute path to powershell.exe if available

mdrohmann:
https://www.pivotaltracker.com/story/show/173815872


mdrohmann:
I did not iterate through the `PATH` variable but instead just print the windows error message in a message window.  Something that should have been done anyways.

PR #816, opened 2020-07-15
Add MSI artifact testing

daved:
https://www.pivotaltracker.com/story/show/173756455


daved:
@Naatan For the moment, I just need general guidance on whether this PowerShell code looks rational. I'll be working on getting a ps1 file called from the Go tests in the meantime. I'd like to hide the lifting code behind a "windows" build tag, and an "msi" build tag. I think it would be useful for local development to not include this logic, and the CI can call for it specifically.

daved:
@Naatan This is now working locally from an elevated cmd.exe process. Please let me know if I've continued in the right direction. If so, I'd appreciate having your guidance on how to marry this to CI and other infra. Specifically, I'm not sure what you'd like to see regarding the gather of artifacts to be tested.

daved:
Also, I'm not sure what to make of this, but I'll check it out in the morning: https://github.com/ActiveState/cli/runs/879818154#step:6:126

test/installers/msi_windows_test.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, msiFileName := range perlMsiFileNames {
+		t.Run(msiFileName, func(t *testing.T) {
+			m := newMsiFile(msiFileName)
+			s := newPwshSession(t)
+
+			cp := s.Spawn(installAction.cmd(m.path))
+			cp.Expect("exitcode:0:", time.Minute*3)
+			cp.ExpectExitCode(0)
+
+			cp = s.Spawn(checkPerlCmd)
```
Naatan, line 101: Would be good if we can also test if modules are working, this is what we used for that in the activation tests:

http://github.com/ActiveState/cli/blob/d73cb9956fb7b4f3e2d1e90c8fe08be4fd96c51c/test/integration/activate_int_test.go#L188-L188

Naatan:
> Specifically, I'm not sure what you'd like to see regarding the gather of artifacts to be tested.

As said in the AC's:

> Should target the languages under the build dir

The `build-msi-languages` places them there, which you can spoof for your local tests of course.

test/installers/msi_windows_test.go
```diff
@@ -0,0 +1,115 @@
+package installers_test
```
Naatan, line 1: This should be under `test/integration`

daved: The suffix must be `_int_windows_test.go` for the implicit windows build tag to be valid. Should I use an explicit build tag with `_int_test.go` suffix, or should I use the implicit `_int_windows_test.go` suffix?

mdrohmann:
Here is the proof that it now works: https://github.com/ActiveState/cli/runs/924499224

test/integration/msi_int_windows_test.go
```diff
@@ -0,0 +1,127 @@
...
+	if !e2e.RunningOnCI() && false {
+		t.Skipf("Skipping; Not running on CI")
+	}
+
+	perlMsiFilePaths, err := msiFilePaths(msiDir, perlMsiPrefix)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(perlMsiFilePaths) == 0 {
```
Naatan, line 89: This is already doing a bit too much by "detecting" perl MSI files, that wasn't an AC. We know it's two specific files. Now this test could succeed with only 1 file that might not even be the file we want to test. Please make this a little "dumber".

test/integration/msi_pwsh_int_windows_test.go
```diff
@@ -0,0 +1,56 @@
...
+	pwshCmd := fmt.Sprintf(pwshCmdForm, msiAct, msiPath, a.logFileName(msiPath), installPath)
+	return pwshCmd
+}
+
+func (a msiExecAction) logFileName(msiPath string) string {
+	msiName := filepath.Base(strings.TrimSuffix(msiPath, filepath.Ext(msiPath)))
+	return filepath.Join(environment.GetRootPathUnsafe(), fmt.Sprintf(`%s_%s.log`, msiName, string(a)))
+}
+
+type pwshSession struct {
```
Naatan, line 40: @mdrohmann is this really necessary? Seems we're just using powershell to capture the exit code, I have a hard time believing we really need to do that.

test/integration/msi_int_windows_test.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, msiFilePath := range perlMsiFilePaths {
+		t.Run(filepath.Base(msiFilePath), func(t *testing.T) {
+			m := newMsiFile(msiFilePath)
+			s := newPwshSession(t)
+
+			cp := s.Spawn(installAction.cmd(m.path, installPath))
+			cp.Expect("exitcode:0:", time.Minute*3)
+			cp.ExpectExitCode(0)
+
+			assertRegistryPathInclues(t, installPath)
```
Naatan, line 104:
```suggestion
			assertRegistryPathIncludes(t, installPath)
```

test/integration/msi_int_windows_test.go
```diff
@@ -0,0 +1,127 @@
...
+	msiDir = mustFilepathByProjectRoot(`/build/msi`)
+	logDir = mustFilepathByProjectRoot(`/build`)
+
+	msiExt        = ".msi"
+	perlMsiPrefix = "ActivePerl"
+
+	asToken = "ActiveState"
+
+	checkPerlVersionCmd = "perl -v"
+	checkPerlModulesCmd = "perldoc -l DBD::Pg"
```
Naatan, line 28: Please incorporate this into the tests itself so far as that is feasible. Right now I dislike the amount of scanning I have to do to grok the code. Constantly scrolling up to reference some thing that really could've just been right there in the code.

test/integration/msi_pwsh_int_windows_test.go
mdrohmann: It is really for the `Start-Process` cmdlet.  The `msiexec.exe` returns right away otherwise, and we have no idea when the installation actually finishes... Well, we could watch the log file, but I think that would be even worse...

The perl cmds do not necessarily have to be called by a powershell.

PR #818, opened 2020-07-15
install.ps1 does not use global errors

mdrohmann:
This stops relying on the global $Errors variable.  It sometimes failed due to an error that occurred in a different command-let than we were looking at.

### DIFF
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -80,7 +80,7 @@ function promptYN([string]$msg)
     return $True
 }
 
-function errorOccured($suppress, $err) {
+function errorOccured($suppress, $errMsg) {
     if($errMsg) {
         if (-Not $suppress){
             Write-Warning $errMsg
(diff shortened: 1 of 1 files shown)


installers/install.ps1
```diff
@@ -98,54 +96,19 @@ function hasWritePermission([string] $path)
     # $acl = Get-Acl $path -ErrorAction 'silentlycontinue'
     # return (($acl.Access | Select-Object -ExpandProperty IdentityReference) -contains $user)
     $thefile = "activestate-perms"
-    New-Item -Path (Join-Path $path $thefile) -ItemType File -ErrorAction 'silentlycontinue'
-    $occurance = errorOccured $True
+    New-Item -Path (Join-Path $path $thefile) -ItemType File -ErrorAction 'SilentlyContinue' -ErrorVariable err
```
MDrakos, line 99: Does the casing matter here? This occurrence is not the same as others in the diff where it is `silentlycontinue`.

mdrohmann: It is all case insensitive. Windows!

PR #821, opened 2020-07-16
State show presents remote project information

MDrakos:
https://www.pivotaltracker.com/story/show/173713612

### DIFF
--- a/internal/runners/show/show.go
+++ b/internal/runners/show/show.go
@@ -77,7 +77,7 @@ func (s *Show) Run(params Params) error {
 	if params.Remote != "" {
 		namespaced, fail := project.ParseNamespace(params.Remote)
 		if fail != nil {
-			return locale.WrapError(fail.ToError(), "err_show_parse_namespace", "Invalid remote argument, must be of the form <Owner>/<Project>")
+			return locale.WrapError(fail, "err_show_parse_namespace", "Invalid remote argument, must be of the form <Owner>/<Project>")
 		}
 
 		owner = namespaced.Owner
@@ -104,14 +104,14 @@ func (s *Show) Run(params Params) error {
 
 	remoteProject, fail := model.FetchProjectByName(owner, projectName)
 	if fail != nil && fail.Type.Matches(model.FailProjectNotFound) {
-		return locale.WrapError(fail.ToError(), "err_show_project_not_found", "Please run `state push` to synchronize this project with the ActiveState Platform.")
+		return locale.WrapError(fail, "err_show_project_not_found", "Please run `state push` to synchronize this project with the ActiveState Platform.")
 	} else if fail != nil {
 		return locale.WrapError(err, "err_show_get_project", "Could not get remote project details")
 	}
 
 	branch, fail := model.DefaultBranchForProjectName(owner, projectName)
 	if fail != nil {
-		return locale.WrapError(fail.ToError(), "err_show_get_default_branch", "Could not get project information from the platform")
+		return locale.WrapError(fail, "err_show_get_default_branch", "Could not get project information from the platform")
 	}
 	if branch.CommitID == nil {
 		return locale.NewError("err_show_commitID", "Remote project details are incorrect. Default branch is missing commitID")
@@ -202,7 +202,7 @@ func scriptsData(project *projectfile.Project, conditional *constraints.Conditio
 func platformsData(owner, project string, branchID strfmt.UUID) ([]string, error) {
 	remotePlatforms, fail := model.FetchPlatformsForCommit(branchID)
 	if fail != nil {
-		return nil, locale.WrapError(fail.ToError(), "err_show_get_platforms", "Could not get platform details for commit: {{.V0}}", branchID.String())
+		return nil, locale.WrapError(fail, "err_show_get_platforms", "Could not get platform details for commit: {{.V0}}", branchID.String())
 	}
 
 	platforms := make([]string, len(remotePlatforms))
@@ -219,7 +219,7 @@ func platformsData(owner, project string, branchID strfmt.UUID) ([]string, error
 func languagesData(owner, project string) ([]string, error) {
 	platformLanguages, fail := model.FetchLanguagesForProject(owner, project)
 	if fail != nil {
-		return nil, locale.WrapError(fail.ToError(), "err_show_get_langauges", "Could not get languages for project")
+		return nil, locale.WrapError(fail, "err_show_get_langauges", "Could not get languages for project")
 	}
 
 	languages := make([]string, len(platformLanguages))
@@ -240,7 +240,7 @@ func visibilityData(owner, project string, remoteProject *mono_models.Project) s
 func commitsData(owner, project string, commitID strfmt.UUID, localProject *project.Project) (string, error) {
 	latestCommit, fail := model.LatestCommitID(owner, project)
 	if fail != nil {
-		return "", locale.WrapError(fail.ToError(), "err_show_get_latest_commit", "Could not get latest commit ID")
+		return "", locale.WrapError(fail, "err_show_get_latest_commit", "Could not get latest commit ID")
 	}
 
 	if !authentication.Get().Authenticated() {
@@ -250,7 +250,7 @@ func commitsData(owner, project string, commitID strfmt.UUID, localProject *proj
 	if localProject != nil && localProject.Owner() == owner && localProject.Name() == project {
 		behind, fail := model.CommitsBehindLatest(owner, project, localProject.CommitID())
 		if fail != nil {
-			return "", locale.WrapError(fail.ToError(), "err_show_commits_behind", "Could not determine number of commits behind latest")
+			return "", locale.WrapError(fail, "err_show_commits_behind", "Could not determine number of commits behind latest")
 		}
 		if behind != 0 {
 			return fmt.Sprintf("%s (%d behind latest)", latestCommit, behind), nil
--- a/internal/runners/show/show.go
+++ b/internal/runners/show/show.go
@@ -87,9 +87,9 @@ func (s *Show) Run(params Params) error {
 			return locale.NewError("err_no_projectfile")
 		}
 
-		owner = s.project.Source().Owner
-		projectName = s.project.Source().Name
-		projectURL = s.project.Source().Project
+		owner = s.project.Owner()
+		projectName = s.project.Name()
+		projectURL = s.project.URL()
 
 		events, err = eventsData(s.project.Source(), s.conditional)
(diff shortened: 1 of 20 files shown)


pkg/platform/model/projects.go
```diff
@@ -33,6 +33,9 @@ var (
 
 	// FailProjectNameConflict is a failure due to a project name conflict
 	FailProjectNameConflict = failures.Type("model.fail.projectconflict")
+
+	// FailProjectNotFound is a fialure due to a project not being found
+	FailProjectNotFound = failures.Type("model.fail.projectnotfound")
```
MDrakos, line 38: This failure was added to satisfy the last AC.
> If project doesn't exist remotely yet then it should print a simple notice (along with the local information):
"Please run state push to synchronize this project with the ActiveState Platform."

I attempted to convert this function to returning an error however it would lead to a very large changeset

internal/runners/show/show.go
```diff
@@ -46,86 +63,102 @@ func New(prime primeable) *Show {
...
+		projectName = namespaced.Project
+	} else {
+		if s.project == nil {
+			return locale.NewError("err_no_projectfile")
 		}
 
-		projectFile, err := projectfile.Parse(projectFilePath)
+		owner = s.project.Source().Owner
+		projectName = s.project.Source().Name
+		projectURL = s.project.Source().Project
```
Naatan, line 92: I don't believe you need the source projectfile for any of these, just use the methods on `s.project`, eg. `s.project.Name()`.

internal/runners/show/show.go
```diff
@@ -46,86 +63,102 @@ func New(prime primeable) *Show {
...
 	}
 
-	scripts, err := scriptsData(src, s.conditional)
+	languages, err := languagesData(owner, projectName)
 	if err != nil {
-		return locale.WrapError(err, "err_show_scripts", "Could not parse scripts.")
+		return locale.WrapError(err, "err_show_langauges", "Could not retrieve language information")
+	}
+
+	commit, err := commitsData(owner, projectName, *branch.CommitID, s.project)
```
Naatan, line 134: This should be prioritizing the commit set in the projectfile, otherwise the "x behind" will always be zero.

internal/runners/show/show.go
```diff
@@ -165,3 +198,108 @@ func scriptsData(project *projectfile.Project, conditional *constraints.Conditio
...
+	if fail != nil {
+		return nil, locale.WrapError(fail.ToError(), "err_show_get_platforms", "Could not get platform details for commit: {{.V0}}", branchID.String())
+	}
+
+	platforms := make([]string, len(remotePlatforms))
+	for i, plat := range remotePlatforms {
+		if plat.Kernel == nil || plat.Kernel.Name == nil {
+			continue
+		}
+		platforms[i] = *plat.Kernel.Name
```
Naatan, line 213: Is that really the most appropriate field? I see they have a `DisplayName` field, shouldn't we use that? Also, if we have to construct the name ourselves like you are here please also include the version and architecture, because projects may be targeting multiple versions of eg. Linux.

internal/runners/show/show.go
```diff
@@ -165,3 +198,108 @@ func scriptsData(project *projectfile.Project, conditional *constraints.Conditio
...
+	}
+
+	return languages, nil
+}
+
+func visibilityData(owner, project string, remoteProject *mono_models.Project) string {
+	if remoteProject.Private {
+		return "Private"
+	}
+	return "Public"
```
Naatan, line 237: Localize

internal/runners/show/show.go
```diff
@@ -165,3 +198,108 @@ func scriptsData(project *projectfile.Project, conditional *constraints.Conditio
...
+		platforms[i] = *plat.Kernel.Name
+	}
+
+	return platforms, nil
+}
+
+func languagesData(owner, project string) ([]string, error) {
+	platformLanguages, fail := model.FetchLanguagesForProject(owner, project)
+	if fail != nil {
+		return nil, locale.WrapError(fail.ToError(), "err_show_get_langauges", "Could not get languages for project")
```
Naatan, line 222:
```suggestion
		return nil, locale.WrapError(fail.ToError(), "err_show_get_languages", "Could not get languages for project")
```

internal/runners/show/show.go
```diff
@@ -165,3 +198,108 @@ func scriptsData(project *projectfile.Project, conditional *constraints.Conditio
...
+	return "Public"
+}
+
+func commitsData(owner, project string, commitID strfmt.UUID, localProject *project.Project) (string, error) {
+	latestCommit, fail := model.LatestCommitID(owner, project)
+	if fail != nil {
+		return "", locale.WrapError(fail.ToError(), "err_show_get_latest_commit", "Could not get latest commit ID")
+	}
+
+	if !authentication.Get().Authenticated() {
```
Naatan, line 246: Pass via primer instead - 
http://github.com/ActiveState/cli/blob/d38e2365a1d64985e97c1895212defd478d3b591/internal/primer/primer.go#L78-L78

internal/runners/show/show.go
```diff
@@ -165,3 +198,108 @@ func scriptsData(project *projectfile.Project, conditional *constraints.Conditio
...
+		}
+		if behind != 0 {
+			return fmt.Sprintf("%s (%d behind latest)", latestCommit, behind), nil
+		}
+	}
+
+	return latestCommit.String(), nil
+}
+
+func secretsData(owner, project string) (map[string][]string, error) {
```
Naatan, line 263: Please use a typed struct, so something like:

```
type secretOutput {
  User []string
  Project []string
}
```

internal/runners/show/show.go
MDrakos: Updated the function for this

internal/runners/show/show.go
MDrakos: The `DisplayName` field is an amalgamation of OS, kernel, version, and architecture. There is an `operating_system` field with `name` and `version` however for `Linux` the name is `CentOS`. I went with this field as it's consistent with what `state platforms` uses. I've updated this code to use the same method that the platforms runner uses.

internal/runners/show/show.go
MDrakos: This is what I originally had, however it led to a tabbing issue when outputting the show information, hence the `map[string][]string`. I've converted it back. Here is an example of the output now:

```
Secrets:
User:
 - foox
 - BLAH
 - world
 - FOO
 - foo3
 - TooManyPizza
 - DOESTHISSHOWUP
Project:
 - foo
```

I spent some time in the output library but couldn't find a simple solution to fix this. If you know of a quick fix, or have another way you would like this presented let me know. Otherwise I will create a ticket to update the output library for this use case

internal/runners/show/show.go
```diff
@@ -46,86 +63,102 @@ func New(prime primeable) *Show {
...
-		Organization: pj.Owner(),
+		ProjectURL:   projectURL,
+		Namespace:    fmt.Sprintf("%s/%s", owner, projectName),
+		Name:         projectName,
+		Organization: owner,
+		Visibility:   visibilityData(owner, projectName, remoteProject),
+		Commit:       commit,
+		Languages:    languages,
+		Platforms:    platforms,
+		Secrets:      secrets,
```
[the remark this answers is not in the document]
MDrakos: This is addressed via moving using a struct over a map however I did add an update to the output library for this.

internal/runners/show/show.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
+	"github.com/ActiveState/cli/internal/runners/platforms"
```
Naatan, line 12: Runners are endpoints, they should never be imported outside of the cmdlet package

internal/runners/show/show.go
```diff
@@ -202,15 +215,14 @@ func scriptsData(project *projectfile.Project, conditional *constraints.Conditio
...
-	for i, plat := range remotePlatforms {
-		if plat.Kernel == nil || plat.Kernel.Name == nil {
-			continue
-		}
-		platforms[i] = *plat.Kernel.Name
+	plats := platforms.MakePlatformsFromModelPlatforms(remotePlatforms)
+
+	platforms := make([]string, len(plats))
+	for i, plat := range plats {
+		platforms[i] = fmt.Sprintf("%s %s %s-bit", plat.Name, plat.Version, plat.BitWidth)
```
Naatan, line 225: Based on your response, I'm confused why we're not using DisplayName, your description makes that sound like exactly the thing we'd want.

internal/runners/show/show.go
```diff
@@ -232,36 +244,37 @@ func languagesData(owner, project string) ([]string, error) {
 
 func visibilityData(owner, project string, remoteProject *mono_models.Project) string {
 	if remoteProject.Private {
-		return "Private"
+		return locale.T("show_visibility_private")
 	}
-	return "Public"
+	return locale.T("show_visibility_public")
```
Naatan: If you're not gonna use `locale.Tl` then make the key less explicit, just call it "private" and "public".

internal/runners/show/show.go
MDrakos: I hesitated to use it because I thought it might be too much information. But based on our discussions earlier re: platforms it's probably best that we just display with the API returns to us. This will mean outputs like:

```
Platforms:
 - macOS 10.12.6, Darwin 16.6.0, libSystem 1238.60.2 x86 64-bit
 - Windows 10 Version 1803, Windows 10.0.17134.1 x86 64-bit
 - CentOS 6.9, Linux 4.9.125, glibc 2.12 x86 64-bit
```

internal/runners/show/show.go
```diff
@@ -218,11 +217,9 @@ func platformsData(owner, project string, branchID strfmt.UUID) ([]string, error
...
 	}
 
-	plats := platforms.MakePlatformsFromModelPlatforms(remotePlatforms)
-
-	platforms := make([]string, len(plats))
-	for i, plat := range plats {
-		platforms[i] = fmt.Sprintf("%s %s %s-bit", plat.Name, plat.Version, plat.BitWidth)
+	platforms := make([]string, len(remotePlatforms))
+	for i, plat := range remotePlatforms {
+		platforms[i] = *plat.DisplayName
```
Naatan, line 222: Needs a nil check

internal/runners/show/show.go
Naatan: It's a bit verbose but at least if all clients use that then we have the ability to tweak it in a single location.

PR #822, opened 2020-07-16
Implemented artifact ID diffing for package management

Naatan:
https://www.pivotaltracker.com/story/show/173738952

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -87,7 +87,7 @@ const ArtifactFile = "artifact.json"
 // ArtifactArchiveName is the standardized name of an artifact archive
 const ArtifactArchiveName = "artifact.tar.gz"
 
-// ArtifactArchiveName is the standardized name of an artifact archive
+// ArtifactCacheFileName is the standardized name of an artifact cache file
 const ArtifactCacheFileName = "artifact_cache.json"
 
 // DefaultNamespaceDomain is the domain used when no namespace is given and one has to be constructed
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -87,6 +87,9 @@ const ArtifactFile = "artifact.json"
 // ArtifactArchiveName is the standardized name of an artifact archive
 const ArtifactArchiveName = "artifact.tar.gz"
 
+// ArtifactArchiveName is the standardized name of an artifact archive
+const ArtifactCacheFileName = "artifact_cache.json"
+
 // DefaultNamespaceDomain is the domain used when no namespace is given and one has to be constructed
 const DefaultNamespaceDomain = "github.com"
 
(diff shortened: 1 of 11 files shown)


internal/constants/constants.go
```diff
@@ -87,6 +87,9 @@ const ArtifactFile = "artifact.json"
 // ArtifactArchiveName is the standardized name of an artifact archive
 const ArtifactArchiveName = "artifact.tar.gz"
 
+// ArtifactArchiveName is the standardized name of an artifact archive
```
mdrohmann, line 90:
```suggestion
// ArtifactCacheFileName is the standardized name of an artifact cache file
```

pkg/platform/runtime/assembler.go
```diff
@@ -21,11 +21,11 @@ type Assembler interface {
 	EnvGetter
 	DownloadDirectoryProvider
 
-	// ArtifactsToDownloadAndUnpack returns the artifacts that need to be
+	// ArtifactsToDownload returns the artifacts that need to be
 	// downloaded and the archives that already exist and *only* need to be unpacked.
 	// The second return value is a map from the archive paths to the artifact
 	// instances.
```
mdrohmann, line 27:
```suggestion
	// ArtifactsToDownload returns the artifacts that need to be downloaded
```

pkg/platform/runtime/installer.go
```diff
@@ -195,9 +195,9 @@ func (installer *Installer) InstallArtifacts(runtimeAssembler Assembler) (envGet
...
 	}
 
-	downloadArtfs, unpackArchives := runtimeAssembler.ArtifactsToDownloadAndUnpack()
+	downloadArtfs := runtimeAssembler.ArtifactsToDownload()
 
-	if len(downloadArtfs) == 0 && len(unpackArchives) == 0 {
+	if len(downloadArtfs) == 0 {
 		// Already installed, no need to download or install
 		logging.Debug("Nothing to download")
 		return runtimeAssembler, false, nil
```
mdrohmann, line 203: Note that this block cancels *all* installation steps ie., also the `PreInstall()` where potentially artifacts are being removed.

So, if there is nothing to be downloaded but something is to be removed, we want to continue here...

pkg/platform/runtime/alternative.go
```diff
@@ -95,32 +101,53 @@ func (ar *AlternativeRuntime) BuildEngine() BuildEngine {
...
+	for _, v := range artifactsRequested {
+		if v.ArtifactID != nil && !funk.Contains(artifactCacheUuids, *v.ArtifactID) {
+			downloadArtfs = append(downloadArtfs, v)
 		}
 	}
-	return downloadArtfs, archives
+	return downloadArtfs
 }
 
 // IsInstalled checks if the merged runtime environment definition file exists
```
mdrohmann, line 153: I think that this function will also need to be adapted.  It probably needs to check if the requested artifacts are equal to the installed artifacts.

PR #828, opened 2020-07-17
Added rollbar logging to msi

Naatan:
https://www.pivotaltracker.com/story/show/173862464


installers/msi-language/ActiveState/Rollbar.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    public GenericException(string message) : base(message)
+    {
+        // This isn't working (stack is still empty) - leaving it here so we can iterate later
+        var stackTraceField = typeof(GenericException).BaseType
+            .GetField("_stackTraceString", BindingFlags.Instance | BindingFlags.NonPublic);
+
+        stackTraceField.SetValue(this, Environment.StackTrace);
+    }
+}
```
Naatan, line 88: I was unable to get this to actually produce a stack trace, all ears if you have any ideas.

mdrohmann: I actually don't know why this does not work.  The documentation says that `Environment.StackTrace` can be `Empty`, but it does not say when or why.

There should be a stack trace with every exception though.  Maybe that could be used somehow? https://docs.microsoft.com/en-us/dotnet/api/system.exception.stacktrace?view=netcore-3.1

PR #829, opened 2020-07-20
Ensure consistent versions

MDrakos:
https://www.pivotaltracker.com/story/show/173738908

### DIFF
--- a/internal/constants/version/version.go
+++ b/internal/constants/version/version.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"regexp"
 
+	"github.com/ActiveState/cli/internal/condition"
 	"github.com/ActiveState/cli/internal/environment"
 	"github.com/blang/semver"
 )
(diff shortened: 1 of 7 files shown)


scripts/version-generator/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+func run() {
+	flag.Parse()
+	if flag.NArg() != 2 {
+		printUsage()
+		os.Exit(0)
+	}
+	branch := flag.Arg(0)
+
+	stateURL := "https://s3.ca-central-1.amazonaws.com/cli-update/update/state/%s/linux-amd64.json"
```
Naatan, line 34: Please drop "linux", this is no longer platform specific. You can just go onto S3 and copy the linux file to "version.json".

internal/constants/version/version.go
```diff
@@ -88,20 +91,29 @@ func (v *Incrementation) Type() (string, error) {
...
+
+	rootPath, err := environment.GetRootPath()
 	if err != nil {
 		return "", err
 	}
-	defer resp.Body.Close()
 
-	type versionJSON struct {
-		Version string
+	versionFilePath := filepath.Join(rootPath, "build", "version.json")
```
Naatan, line 105: This should probably be `build/update/version.json`, see http://github.com/ActiveState/cli/blob/c26bb7df7cfda729e20290a02f5177c24673d613/activestate.yaml#L70-L70

PR #833, opened 2020-07-21
Update version file

MDrakos:
https://www.pivotaltracker.com/story/show/173738908

The `version.json` that we download has the current master version in it. Before this file can be deployed to s3 and overwrite the existing one, we have to update the files contents with the new version information.


MDrakos:
Can I create a story and address it after this is merged? We are still using the platform specific json files during update, in the install scripts, and potentially other places. Currently master will not increment without this fix.

PR #837, opened 2020-07-22
MSI language installer has language preset

mdrohmann:
https://www.pivotaltracker.com/story/show/173719976


activestate.yaml
```diff
@@ -137,41 +137,20 @@ scripts:
...
-        string=$1
 
-        # Use base length that avoids string overflow warnings from wix
-        # Subtract string length and brackets
-        padLength=$((246-${#string}-4))
-        printf -v pad '%*s' "$padLength"
-        padded=$(printf '%s' $1)$(echo ${pad// /-})
-        echo $padded
-      }
+      PROJECT_OWNER_NAME=${2:-ActiveState/ActivePerl-5.26}
```
MDrakos, line 141: Why do we have a default value for this argument? If the only argument passed was `Perl` wouldn't this script still fail when we try to run the Go script?

installers/msi-language/ActiveState/FileAssociation.cs
```diff
@@ -0,0 +1,85 @@
...
+        public string Extension { get; set; }
+        public string ProgId { get; set; }
+        public string FileTypeDescription { get; set; }
+        public string ExecutableFilePath { get; set; }
+
+        public bool SetAssociation()
+        {
+            bool madeChanges = false;
+            madeChanges |= SetKeyDefaultValue(@"Software\Classes\" + Extension, ProgId);
+            madeChanges |= SetKeyDefaultValue(@"Software\Classes\" + ProgId, FileTypeDescription);
```
MDrakos, line 25: When I was implementing this in the `wxs` file I didn't find that this registry key was necessary. What does this one do?

installers/msi-language/ActiveState/FileAssociation.cs
```diff
@@ -0,0 +1,85 @@
...
+            madeChanges |= SetKeyDefaultValue($@"Software\Classes\{ProgId}\shell\open\command", "\"" + ExecutableFilePath + "\" \"%1\" %*");
+            return madeChanges;
+        }
+
+        public bool DeleteAssociation()
+        {
+            // We only delete the ProgId entry, as the file extension entry could be modified by other programmes
```
MDrakos, line 32: We may want to confirm the default approach Windows uses when removing file type associations on uninstall. If I remember correctly it will delete the `Software\Classes\<Extension>` key as well

installers/msi-language/Uninstall/CustomAction.cs
```diff
@@ -33,7 +57,22 @@ public static ActionResult Uninstall(Session session)
                 return ActionResult.Failure;
             }
 
-            return Remove.EnvironmentEntries(session, installDir);
+            result = Remove.EnvironmentEntries(session, installDir);
+            if (result.Equals(ActionResult.Failure))
+            {
+                session.Log("Could not remove shortcuts directory");
+                ActiveState.RollbarHelper.Report("Could not remove shortcuts directory");
```
MDrakos, line 64: The remove above is not related to shortcuts.

Also nit: since we are creating the shortcuts in the preset logic it might also make sense to remove them there

scripts/msi-installer-builder/main.go
```diff
@@ -0,0 +1,189 @@
...
+		Preset:              Unknown.String(),
+		ProjectOwnerAndName: pad("PROJECT_OWNER_AND_NAME"),
+		ReleaseNotes:        pad("RELEASE_NOTES"),
+		ProjectName:         pad("PROJECT_NAME"),
+	}
+}
+
+func parseArgs(args []string) (*config, error) {
+	if len(os.Args) == 1 {
+		// empty configuration for quick testing
```
MDrakos, line 128: Can this be removed now?

installers/msi-language/Preset/CustomAction.cs
```diff
@@ -0,0 +1,214 @@
...
+            var exts = String.Join(";", oldPathExt.Split(';').Concat(PathExtensions).Distinct().ToArray());
+            if (exts != oldPathExt)
+            {
+                Environment.SetEnvironmentVariable("PATHEXT", exts, EnvironmentVariableTarget.Machine);
+            }
+
+            return ActionResult.Success;
+        }
+
+        private ActionResult PerlCriticShortcut()
```
MDrakos, line 115: Not related to this line directly however in the `Product.p.wxs` around line 44 we are creating a documentation shortcut that contains a hard coded `Perl` reference. Right now it's just the icon file but we may want to move that shortcut logic into here as well.

activestate.yaml
mdrohmann: This is to support the no-argument option that you also pointed out below.   You are right I can remove that.

installers/msi-language/ActiveState/FileAssociation.cs
mdrohmann: I filed a story for this: https://www.pivotaltracker.com/n/projects/2203557

As there could be several bindings to `.pl` or `.wpl` (from other programmes) too, I "believe" that the MSI installer logic is actually smarter here, and would somehow restore a still existing installation's binding.

installers/msi-language/ActiveState/FileAssociation.cs
mdrohmann: It shows up here: PerlDocument

installers/msi-language/Uninstall/CustomAction.cs
mdrohmann: I actually decided not to move it into the `Preset.Uninstall()` method, because the `shortCutDir` is something that we want to remove for every language probably...  We'll see. :)

PR #838, opened 2020-07-22
Retry download up to 5 times

Naatan:
https://www.pivotaltracker.com/story/show/173947294


installers/install.ps1
```diff
@@ -90,6 +90,35 @@ function errorOccured($suppress, $errMsg) {
     return $False, ""
 }
 
+function download([string] $url, [string] $out) {
+    $Stoploop = $false
```
MDrakos, line 94: This doesn't appear to be used anywhere

PR #840, opened 2020-07-23
MSI Private Projects

MDrakos:
https://www.pivotaltracker.com/story/show/173427276


installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -80,108 +80,58 @@ public static ActionResult InstallStateTool(Session session, out string stateToo
...
+            {
+                // Catch cancel and return
+                return runResult;
+            }
+            else if (runResult == ActionResult.Failure)
+            {
+                Record record = new Record();
+                session.Log(string.Format("Output: {0}", output));
+                var errorOutput = FormatErrorOutput(output);
+                record.FormatString = string.Format("Failed with error:\n{0}", errorOutput);
```
mdrohmann, line 119:
```suggestion
                record.FormatString = string.Format("Platform login failed with error:\n{0}", errorOutput);
```

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -80,108 +80,58 @@ public static ActionResult InstallStateTool(Session session, out string stateToo
...
-                    );
-                    return ActionResult.Failure;
-                }
+                session.Message(InstallMessage.Error | (InstallMessage)MessageBoxButtons.OK, record);
+                return runResult;
             }
-            catch (Exception objException)
+            // The auth command did not fail but the username we expected is not present in the output meaning
+            // another user is logged into the State Tool 
+            else if (!output.Contains(username))
```
mdrohmann, line 126: This is strange behavior by the state tool. If we cannot login with a requested user, the `state auth` command should fail IMO...

I created a story for this: https://www.pivotaltracker.com/story/show/173984323 and added an AC to ensure that the change would still work with the language MSI...

MDrakos:
@Naatan the rollbar errors generated from `ActiveState.Command.Run()` include the command string that failed. So we should be okay there.

PR #843, opened 2020-07-24
Improve error reporting

Naatan:


installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -63,6 +63,7 @@ public static ActionResult InstallStateTool(Session session, out string stateToo
             if (result.Equals(ActionResult.UserExit))
             {
                 // Catch cancel and return
+                ActiveState.RollbarHelper.Report("Installation exited prematurely due to UserExit (probably safe to ignore)");
```
MDrakos, line 66: nit: There's a couple more of these checks if you want to also update them [here](https://github.com/ActiveState/cli/blob/master/installers/msi-language/StateDeploy/CustomAction.cs#L109) and [here](https://github.com/ActiveState/cli/blob/master/installers/msi-language/StateDeploy/CustomAction.cs#L192)

Naatan: Will get to them in a followup, currently mainly interested in this function due to some strange errors.

PR #847, opened 2020-07-29
Fix enterprise project activation

MDrakos:
https://www.pivotaltracker.com/story/show/173948845

### DIFF
--- a/pkg/platform/model/recipe.go
+++ b/pkg/platform/model/recipe.go
@@ -58,7 +58,7 @@ func FetchRawRecipeForCommitAndPlatform(commitID strfmt.UUID, owner, project str
 
 // FetchRecipeIDForCommit returns a recipe ID for a project based on the given commitID and the current platform
 func FetchRecipeIDForCommit(commitID strfmt.UUID, owner, project, orgID string, private bool) (*strfmt.UUID, *failures.Failure) {
-	return fetchRecipeID(commitID, owner, project, orgID, private, nil)
+	return fetchRecipeID(commitID, owner, project, orgID, private, &HostPlatform)
 }
 
 // FetchRecipeIDForCommitAndPlatform returns a recipe ID for a project based on the given commitID and platform string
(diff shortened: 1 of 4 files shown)


pkg/platform/model/recipe.go
```diff
@@ -133,17 +138,20 @@ func commitToOrder(commitID strfmt.UUID, owner, project string, hostPlatform *st
...
+func fetchRecipeID(commitID strfmt.UUID, owner, project, orgID string, private bool, hostPlatform *string) (*strfmt.UUID, *failures.Failure) {
 	var err error
 	params := iop.NewSolveOrderParams()
 	params.Order, err = commitToOrder(commitID, owner, project, hostPlatform)
 	if err != nil {
 		return nil, FailOrderRecipes.Wrap(err)
 	}
+	if private {
+		params.OrganizationID = &orgID
+	}
```
Naatan, line 150: Any reason why we shouldn't just always set this?

pkg/platform/model/recipe.go
```diff
@@ -133,17 +138,20 @@ func commitToOrder(commitID strfmt.UUID, owner, project string, hostPlatform *st
 	return order, nil
 }
 
-func fetchRecipeID(commitID strfmt.UUID, owner, project string, hostPlatform *string) (*strfmt.UUID, *failures.Failure) {
+func fetchRecipeID(commitID strfmt.UUID, owner, project, orgID string, private bool, hostPlatform *string) (*strfmt.UUID, *failures.Failure) {
 	var err error
 	params := iop.NewSolveOrderParams()
 	params.Order, err = commitToOrder(commitID, owner, project, hostPlatform)
```
Naatan, line 144: We're still passing hostPlatform here, given the new additions for asserting the platform should this be removed?

pkg/platform/model/recipe.go
```diff
@@ -165,15 +173,21 @@ func fetchRecipeID(commitID strfmt.UUID, owner, project string, hostPlatform *st
 		}
 	}
 
-	// Because we filter platforms in the request we should only
-	// receive one recipe ID
-	if len(recipeID.Payload) != 1 {
-		return nil, FailOrderRecipes.New("err_recipe_payload")
+	if hostPlatform == nil {
+		hostPlatform = &HostPlatform
```
Naatan, line 177: I'd rather we rely on this always being passed in, so we're not making this function rely on globals.

pkg/platform/model/recipe.go
MDrakos: If this is set on public projects we do not get the correct recipeID back

pkg/platform/model/recipe.go
MDrakos: Originally `commitToOrder` was also filtering platforms from the returned order. It's no longer needed here, but is still needed in `fetchRawRecipes`, where it now lives until we address this: https://www.pivotaltracker.com/story/show/174076298.

PR #848, opened 2020-07-30
State export config command

MDrakos:
https://www.pivotaltracker.com/story/show/173905391


cmd/state/internal/cmdtree/export.go
```diff
@@ -109,3 +109,40 @@ func newAPIKeyCommand(prime *primer.Values) *captain.Command {
...
+				Value: &params.Filter,
+			},
+		},
+		[]*captain.Argument{},
+		func(ccmd *captain.Command, _ []string) error {
+			return config.Run(ccmd, params)
+		})
+}
+
+func newDirCommand(prime *primer.Values) *captain.Command {
```
Naatan, line 136: D'oh, we had two ACs covering the same use-case. Please drop the `dir` subcommand and use only the `--filter`. I'll update the ACs.

internal/runners/export/config.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return strings.Join(filterLookup, ",")
+}
+
+type Config struct {
+	out output.Outputer
+}
+
+type ConfigParams struct {
+	Filter string
```
Naatan, line 29: Let's use a FlagMarshaler for this

http://github.com/ActiveState/cli/blob/b65496236a6d4273d808ae85751fca49dc327e07/internal/captain/flag.go#L9-L9

Feel free to ping me if you'd like some pointers on how to use that.

internal/runners/export/config.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func NewConfig(prime primeable) *Config {
+	return &Config{prime.Output()}
+}
+
+func (c *Config) Run(cmd *captain.Command, params ConfigParams) error {
+	if params.Filter == FilterDir {
+		c.out.Print(config.ConfigPath())
```
Naatan, line 38: Rather than switching based on filter we should actually be filtering. Ie. the standard output is a struct with the "dir" key, and if filter=dir is given we actually filter to that level.

internal/runners/export/config.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil
+	}
+	return cmd.Usage()
```
Naatan, line 41: The AC says `If no filter is used it's outputted as json, eg. { "dir": "" }`

MDrakos: This stemmed from the overlapping ACs

internal/runners/export/config.go
```diff
@@ -0,0 +1,96 @@
...
+		if k != Unknown {
+			supported = append(supported, v)
+		}
+	}
+
+	return supported
+}
+
+func SupportedFilters() string {
+	return strings.Join(supportedFilters(), ", ")
```
Naatan, line 47: This logic belongs in the cmdlet imo, should just return the slice of strings here.

internal/runners/export/config.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Filter is the --filter flag for the export config command, it implements captain.FlagMarshaler
+type Filter int
+
+const (
+	Unknown Filter = iota
+	Dir
+)
+
+var lookup = map[Filter]string{
```
Naatan, line 21:
```suggestion
var filterLookup = map[Filter]string{
```

This packages isn't only about filters.

internal/runners/export/config.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+
+	"github.com/ActiveState/cli/internal/captain"
+	"github.com/ActiveState/cli/internal/config"
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+)
+
+// Filter is the --filter flag for the export config command, it implements captain.FlagMarshaler
+type Filter int
```
Naatan, line 14: Can we move all the filter stuff into a filter.go file?

internal/runners/export/config.go
```diff
@@ -0,0 +1,96 @@
+package export
```
Naatan, line 1: This one feels complex enough that we should put it in its own package.

internal/runners/export/config.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/ActiveState/cli/internal/config"
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+)
+
+// Filter is the --filter flag for the export config command, it implements captain.FlagMarshaler
+type Filter int
+
+const (
+	Unknown Filter = iota
```
Naatan, line 17: You don't seem to be using Unknown for anything that makes it necessary. ie. all logic relying on it could easily rely on the absence of it instead.

internal/runners/export/config.go
```diff
@@ -0,0 +1,96 @@
...
+	return &Config{prime.Output()}
+}
+
+func (c *Config) Run(cmd *captain.Command, params ConfigParams) error {
+	output := configOutput{config.ConfigPath()}
+
+	if params.Filter == Dir {
+		c.out.Print(output.Dir)
+		return nil
+	}
```
Naatan, line 87: This still isn't actually filtering. What happens if you give multiple filters?

Filter doesn't mean "only print this", it means filter down to the given keynames. If only one keyname is given then only print the value of that key.

internal/runners/export/config.go
MDrakos: I've updated it's name to `unset` and am now using it when creating the params

MDrakos: Nvm it can be removed 👍

internal/runners/export/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil {
+			return err
+		}
+		params.filters = append(params.filters, filter)
+	}
+
+	return nil
+}
+
+func NewConfig(prime primeable) *Config {
```
Naatan, line 39: This can now just be `New`

internal/captain/flag.go
```diff
@@ -47,6 +47,8 @@ func (c *Command) setFlags(flags []*Flag) error {
 			flagSetter().BoolVarP(
 				v, flag.Name, flag.Shorthand, *v, flag.Description,
 			)
+		case *[]string:
+			flagSetter().StringSliceVarP(v, flag.Name, flag.Shorthand, *v, flag.Description)
```
Naatan, line 51: You don't need this, use FlagMarshaler.Set().

internal/runners/export/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+type ConfigParams struct {
+	Filters []string
+	filters []*Filter
+}
+
+type primeable interface {
+	primer.Outputer
+}
+
+func prepare(params *ConfigParams) error {
```
Naatan, line 26: FlagMarshaler should be able to handle this logic

internal/runners/export/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	output := map[string]string{
+		Dir.String(): config.ConfigPath(),
+	}
+
+	if params.Filters == nil {
+		return c.printOutput(output)
+
+	}
```
Naatan, line 56: Rather than printing the output twice and depend on returns you can just add this conditional when you apply the filtering.

internal/runners/export/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if len(params.Filters) == 1 {
+		c.out.Print(output[params.filters[0].String()])
+		return nil
```
Naatan, line 60: This is assuming the key exists. Instead of this you could return early when you have the value inside the filter loop.

internal/runners/export/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return c.printOutput(filteredOutput)
+}
+
+func (c *Config) printOutput(output map[string]string) error {
+	data, err := json.Marshal(output)
+	if err != nil {
+		return locale.WrapError(err, "err_export_config_marshal", "Could not marshal config data")
+	}
```
Naatan, line 77: Don't need this, we have outputer for that (ie. --output json).

MDrakos: From the ACs:
> If no filter is used it's outputted as json, eg. `{ "dir": "" }`

Is there a preferred way to use the JSON outputter without the flag being passed? Or do we want to update the AC?

Naatan: Let's revisit that AC. If the invoker wants json they can use `--output`.

internal/runners/export/config/filter.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if len(f.terms) == 0 {
+		return locale.NewError("err_invalid_filter", value, strings.Join(SupportedFilters(), ", "))
+	}
+	return nil
+}
+
+func (f Filter) Type() string {
+	return "filters"
```
Naatan, line 69:
```suggestion
	return "filter"
```

internal/runners/export/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+	output := map[string]string{
+		Dir.String(): config.ConfigPath(),
+	}
+
+	filteredOutput := map[string]string{}
+	if params.Filter.terms == nil {
+		filteredOutput = output
+	}
+
+	for _, filter := range params.Filter.terms {
```
Naatan, line 39: Nit: You're iterating over terms, not filters (ie. var on left side should be `term`)

internal/runners/export/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (c *Config) Run(cmd *captain.Command, params *ConfigParams) error {
+	output := map[string]string{
+		Dir.String(): config.ConfigPath(),
+	}
+
+	filteredOutput := map[string]string{}
+	if params.Filter.terms == nil {
+		filteredOutput = output
+	}
```
Naatan, line 37: nit: I would invert this so it flows a little better, eg. `filteredOutput := output` is the initial definition, then the empty map and for loop happen inside an `!= nil` conditional.

internal/runners/export/config/config.go
MDrakos: Yep, missed that one

PR #849, opened 2020-07-30
Add `-c` flag to `state activate` so `state activate` can be used as a shell in vscode

Naatan:
https://www.pivotaltracker.com/story/show/174106083


internal/subshell/cmd/cmd.go
```diff
@@ -108,6 +114,9 @@ func (v *SubShell) Activate() *failures.Failure {
 	}
 
 	shellArgs := []string{"/K", v.rcFile.Name()}
+	if v.activateCommand != nil {
+		shellArgs = append(shellArgs, "/C", *v.activateCommand)
```
mdrohmann, line 118: In the `cmd` shell  `/K` and `/C` are mutually exclusive...  So that would need to be handled separately, I also don't know if you can both run a batch file and a custom command...

Naatan: Yeah just ran into this myself while testing. Good eye :)

PR #850, opened 2020-07-31
Add PPM conversion flow

mdrohmann:
https://www.pivotaltracker.com/story/show/174027892


internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,230 @@
...
+	visitedC chan string
+	once     sync.Once
+}
+
+// StartConversionFlowIfNecessary checks if the user is in a project directory.
+// If not, they are asked to create a project, and (in a wizard-kind-of way) informed about the consequences.
+func (p *Ppm) StartConversionFlowIfNecessary() error {
+	// start conversion flow only if we cannot find a project file
+	if p.project == nil {
+		c := make(chan os.Signal, 1)
```
Naatan, line 40: I'm concerned with the use of channels here, I don't see why this should be necessary. Is it to facilitate your tests? If so, those tests should really be integration tests, not unit tests.

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,230 @@
...
+			cf.createVirtualEnv()
+		}
+		return err
+	}
+	return nil
+}
+
+func newConversionFlow(survey surveySelectFunc, out output.Outputer, openURI func(string) error) *conversionFlow {
+	// Note: the buffer length needs to be big enough to hold the maximum number of visited items
+	visitedC := make(chan string, 20)
```
Naatan, line 59: This is trying to be too smart for its own good. Keep it dumb and easy to digest. I'd much rather see some redundant analytic event calls than layers of abstraction and channel magic so we can avoid a bit of redundancy.

internal/runners/ppm/convert_test.go
```diff
@@ -0,0 +1,183 @@
+package ppm
```
Naatan, line 1: Most of these tests read like integration tests to me, not unit tests. We invested heavily in writing an integration testing framework so please use that instead ;) Unit tests should be testing single functions. If you're relying on channels for tests it's a good signal you should be integration testing instead.

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -33,14 +36,18 @@ func newPpmCommand() *captain.Command {
...
+			if err != nil {
+				return err
+			}
 			return printMain()
 		},
 	)
 
 	var children []*captain.Command
-	children = addPackagesCommands(children)
+	children = addPackagesCommands(runner, children)
```
Naatan, line 48: Don't pass runners around. Runners are top level.

internal/runners/ppm/convert.go
mdrohmann: No, it is not for the tests.  It is for handling the interrupt which needs to be done in a go function listening to this `c` channel.

And then - to send the right analytics method in case an interrupt is happening - I have to also send the visited menu items via a channel...  It is quite awkward, but an integration test would really not help here, as I am mostly concerned about the correct flow and return values, but also that the analytics function is called with the correct arguments.  And especially the latter cannot be verified with integration tests...

internal/runners/ppm/convert.go
mdrohmann: It relates to the answer above.  But on second thought (considering what you sent on slack about the interrupts in the `survey` library), I could probably just handle interrupts through returned error codes from the survey library.  Then I would not need go functions and all this channel stuff that took me so long to get right. :scream: 

Yes, I will do that! :)  It will look much better afterwards!

internal/runners/ppm/convert.go
Naatan: Ok. I'll wait till the full PR then, but suffice to say I'd like the main conversion flow to look a lot dumber than it currently does. I think we can keep analytics and signal interrupts mostly separate from the rest of the conversion flow.

internal/runners/ppm/convert_test.go
mdrohmann: With the removal of the interrupt-signal-handling, it should have improved, and be clearer that these are actually unit tests.

cmd/state/internal/cmdtree/ppm.go
mdrohmann: As mentioned earlier, I would like to wait for @MDrakos to merge his blocking PR before I change this file, and keep the changes to it to a minimum to simplify merging.  I just added some placeholder to the commands that need the conversion logic, but that is temporary, of course...

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,191 @@
...
+type analyticsEventFunc func(string, string, string)
+
+// surveySelectFunc displays a menu with options that the user can select
+type surveySelectFunc func(message string, choices []string, defaultResponse string) (string, *failures.Failure)
+
+const (
+	askedWhy          string = "asked why"
+	seenStateToolInfo        = "visited state tool info"
+	seenPlatformInfo         = "visited platform info"
+	notConvinced             = "still wants ppm"
```
Naatan: These don't seem to be used

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,191 @@
...
+	openURI   func(string) error
+	eventFunc analyticsEventFunc
+}
+
+// StartConversionFlowIfNecessary checks if the user is in a project directory.
+// If not, they are asked to create a project, and (in a wizard-kind-of way) informed about the consequences.
+func (p *Ppm) StartConversionFlowIfNecessary() error {
+	// start conversion flow only if we cannot find a project file
+	if p.project == nil {
+		cf := newConversionFlow(p.prompt.Select, p.out, open.Run, analytics.EventWithLabel)
```
Naatan, line 46: Please remove the unit tests for this and simplify the code. If we really want to test this we can write integration tests, but for now let's just focus on landing this story.

If I got to write the `state deploy` code again I would dump a lot of the logic around unit tests and simplify great portions. The pain that code has caused me is far greater than that the unit tests actually solved any problem we were having.

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,191 @@
...
+
+		analytics.Event(analytics.CatPpmConversion, "run")
+		r, err := cf.runSurvey()
+		analytics.EventWithLabel(analytics.CatPpmConversion, "completed", r.String())
```
Naatan, line 50: Please log an error state and a successful state, eg. http://github.com/ActiveState/cli/blob/156a8d6a5b4649a459888c2c7a99d43c2d40b67d/cmd/state/internal/cmdtree/tutorial.go#L51-L51

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,191 @@
...
+	rejected
+	canceled
+)
+
+func (r conversionResult) String() string {
+	return []string{"accepted", "rejected", "canceled"}[r]
+}
+
+func (cf *conversionFlow) SendClickEvent(what string) {
+	cf.eventFunc(analytics.CatPpmConversion, "click", what)
```
Naatan, line 85: This does not justify having its own function, please just call the analytics code directly. Also please don't call it "click", there's no clicks in terminals..

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (cf *conversionFlow) runSurvey() (conversionResult, error) {
+	choices := []string{
+		convertAnswerCreate,
+		locale.Tl("ppm_convert_answer_why", "Why is this necessary? I Just want to manage packages."),
+	}
+	choice, fail := cf.survey(textutils.WordWrap(locale.Tl(
```
Naatan, line 94: I like the idea, but this does not belong in this story. If we are to handle this it would be handled from the output library. Feel free to file a story.

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,191 @@
...
+	cf.SendClickEvent("asked-why")
+	return cf.explainVirtualEnv(false, false)
+}
+
+func (cf *conversionFlow) createVirtualEnv() error {
+	// TODO: start wizard
+
+	cf.out.Print(textutils.WordWrap(locale.Tl(
+		"ppm_convert_after_tutorial",
+		"For your convenience you can continue to use ppm commands once you’ve activated your virtual runtime environment. We’ll give you handy tips on how your commands map to State Tool so you can learn as you go.",
```
Naatan, line 128:
```suggestion
		"For your convenience you can continue to use ppm commands once you've activated your virtual runtime environment. We'll give you handy tips on how your commands map to State Tool so you can learn as you go.",
```

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,191 @@
...
+	if !alreadySeenStateToolInfo {
+		choices = append(choices, stateToolInfo)
+	}
+	// add choice to open Platform marketing page (if not looked at before)
+	if !alreadySeenPlatformInfo {
+		choices = append(choices, platformInfo)
+	}
+	// always add choices to create virtual environment and to say no again
+	choices = append(choices, convertAnswerCreate, no)
+	explanation := textutils.WordWrap(locale.Tl("ppm_convert_explanation", "State Tool was developed from the ground up with modern software development practices in mind. Development environments with globally installed language runtime environments are increasingly shunned by modern development practices, and as a result the State Tool and the ActiveState Platform tries to do away with them entirely.\n"))
```
Naatan, line 149: This is a bit long to have in here, let's add it to en-us and use `locale.T`.

internal/runners/ppm/convert.go
mdrohmann: I am not quite convinced, as I believe that the unit-test could be very helpful here, as it can test things that are very difficult or impossible to test with integration tests, but I'll remove it.  It certainly makes the code a little simpler.

internal/runners/ppm/convert.go
mdrohmann: Story is filed here: https://www.pivotaltracker.com/story/show/174196140

And textutils.WordWrap is removed.

internal/runners/ppm/convert.go
mdrohmann: They should have been, maybe this was still on an older commit...

internal/runners/ppm/convert.go
Naatan: What wouldn't you be able to test with integration tests? Sounds like a gap in our integration testing framework that we should fill.

Personally my new standard for unit tests is they should be achievable with minimal disruption to your architecture. ie. if you are architecting your code in a very specific way JUST to facilitate unit tests at the cost of readability and maintainability then that code better be worth making the effort for.

mdrohmann: Well, I was thinking about the google analytics events.  In the unit tests, you can nicely check that they were triggered for each code path, and also check their payload.  It is not *impossible* to intercept these events, but probably not very feasible. But I understand your argument that it maybe is not worth the effort: After all state tool logic is never so complicated that we **have** to do unit tests...  

State Tool complexity comes more from the rather wide spectrum of functionality, and that can be managed with integration tests.

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// runSurvey is the entry point to the conversion survey
+func (cf *ConversionFlow) runSurvey() (conversionResult, error) {
+	choices := []string{
+		convertAnswerCreate,
+		locale.Tl("ppm_convert_answer_why", "Why is this necessary? I Just want to manage packages."),
+	}
+	choice, fail := cf.prompt.Select(locale.Tl(
+		"ppm_convert_create_question", "You need to create a runtime environment to proceed.\n"),
```
Naatan, line 100:
```suggestion
		"ppm_convert_create_question", "PPM functionality has been transferred to the State Tool. In order to continue managing your packages we need to create a virtual runtime environment for your project.\n"),
```

internal/runners/ppm/convert.go
```diff
@@ -0,0 +1,214 @@
...
+		choices, "")
+	if fail != nil {
+		return canceled, locale.WrapInputError(fail, "err_ppm_convert_interrupt", "Invalid response received.")
+	}
+	if choice == choices[0] {
+		return accepted, nil
+	}
+
+	analytics.EventWithLabel(analytics.CatPpmConversion, "selection", askedWhy)
```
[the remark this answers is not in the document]
mdrohmann: As with the other analytics events, I still fixed the names of the "labels" that are sent to GA, such that we are independent of changing the locale strings for the answers or adding a new locale language.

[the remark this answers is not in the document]
Naatan: ok, but `askedWhy` captures the question, not the answer.

PR #852, opened 2020-08-01
Paid tiers were still set to 'privateAllowed=false'

th3coop:
`privateProjects` was being set to false even when the user was on a paid tier.  We wouldn't break out of the loop when we found the tier we wanted so unless the tier was last  in the list, it would always get set to false.

 https://www.pivotaltracker.com/story/show/174131275

### DIFF
--- a/internal/runners/auth/auth.go
+++ b/internal/runners/auth/auth.go
@@ -111,7 +111,11 @@ func userToJSON(username string) ([]byte, *failures.Failure) {
 	tier := organization.Tier
 	privateProjects := false
 	for _, t := range tiers {
-		privateProjects = (tier == t.Name && t.RequiresPayment)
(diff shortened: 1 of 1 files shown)


internal/runners/auth/auth.go
```diff
@@ -111,7 +111,11 @@ func userToJSON(username string) ([]byte, *failures.Failure) {
 	tier := organization.Tier
 	privateProjects := false
 	for _, t := range tiers {
-		privateProjects = (tier == t.Name && t.RequiresPayment)
+		if pp := (tier == t.Name && t.RequiresPayment); pp {
+			privateProjects = pp
```
Naatan, line 115:
```suggestion
		if tier == t.Name && t.RequiresPayment {
			privateProjects = true
```

Makes it a bit easier to digest

PR #853, opened 2020-08-04
Shim popular PPM commands

MDrakos:
https://www.pivotaltracker.com/story/show/174027876


cmd/state/internal/cmdtree/ppm.go
```diff
@@ -2,128 +2,134 @@ package cmdtree
...
-	fmt.Println(strings.TrimSpace(locale.T("ppm_header_message")))
+func printDefault(prime *primer.Values) error {
+	prime.Output().Print(strings.TrimSpace(locale.T("ppm_header_message")))
 	return nil
 }
 
-func printMain() error {
-	fmt.Println(locale.T("ppm_print_main"))
+func printMain(prime *primer.Values) error {
+	prime.Output().Print(locale.T("ppm_print_main"))
```
Naatan, line 28: Just pass the output, no need to pass the whole `prime` structure.

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -171,25 +177,79 @@ func addOtherCommands(cmds []*captain.Command) []*captain.Command {
...
 		"prints ppm help message",
 		func(_ *captain.Command, _ []string) error {
-			return printMain()
+			return printMain(prime)
 		},
 	))
 }
+
+func shim(prime *primer.Values, intercepted, replaced, localeID string, args ...string) error {
+	err := shim(prime, intercepted, replaced, localeID, args...)
```
Naatan, line 204: `shim` is calling `shim`? Did you test this?

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -2,128 +2,134 @@ package cmdtree
...
 		nil, nil,
 		func(_ *captain.Command, args []string) error {
 			for _, arg := range args {
 				if arg == "--version" {
-					return printDefault()
+					return printDefault(prime)
 				}
 			}
-			return printMain()
+			return shim(prime, "ppm", "packages", "ppm_print_forward", args...)
```
Naatan, line 42: This `shim` stuff only works for printing out basic messages when invoking a command. This new stuff is getting much more advanced. For one we'll need a runner, and for another we need explicitly typed functions for eg. `ppk packages`.

The point of cmdtree is to establish the main command structure, it should not have any command logic, that's what the runner is for.

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -171,25 +177,79 @@ func addOtherCommands(cmds []*captain.Command) []*captain.Command {
...
 		"info",
 		"prints ppm help message",
 		func(_ *captain.Command, _ []string) error {
-			return printMain()
+			return printMain(prime)
 		},
 	))
 }
+
+func shim(prime *primer.Values, intercepted, replaced, localeID string, args ...string) error {
```
Naatan, line 203: Everything from this line onward belongs in a runner, not in the cmdtree.

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -171,25 +177,79 @@ func addOtherCommands(cmds []*captain.Command) []*captain.Command {
...
+	} else {
+		analytics.EventWithLabel(analytics.CatPPMShimCmd, "success", fmt.Sprintf("intercepted=%s, replaced=%s", intercepted, replaced))
+	}
+	return err
+}
+
+func executeShim(prime *primer.Values, intercepted, replaced, localeID string, args ...string) error {
+	pj, fail := projectfile.GetSafe()
+	if fail != nil && !fail.Type.Matches(projectfile.FailNoProject) {
+		return locale.WrapError(fail.ToError(), "err_ppm_get_projectfile", "Encountered unexpected error loading projectfile")
```
Naatan, line 216:
```suggestion
		return locale.WrapError(fail.ToError(), "err_ppm_get_projectfile", "Encountered unexpected error parsing your activestate.yaml, please manually correct the yaml or if the issue is out of your hands consider letting us know about your issue at {{.V0}}", constants.ForumsURL)
```

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -171,25 +177,79 @@ func addOtherCommands(cmds []*captain.Command) []*captain.Command {
...
+	}
+
+	forwarded := []string{"state", replaced}
+	forwarded = append(forwarded, replacedArgs...)
+	prime.Output().Print(locale.Tr(localeID, strings.Join(forwarded, " "), intercepted))
+
+	return invoke(replaced, replacedArgs...)
+}
+
+func invoke(command string, args ...string) error {
```
Naatan, line 239: Please be sure to replace this with my code once my PR is merged (before merging this one).

cmd/state/internal/cmdtree/ppm.go
MDrakos: Whoops that was a mistake. Fixed now in the runner

cmd/state/internal/cmdtree/ppm.go
MDrakos: Agreed. I've moved the shimming to a runner. I'm not sure if interpreted the typed functions comment correctly so please let me know.

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -1,135 +1,134 @@
...
-	children = addProjectCommands(prime, children)
-	children = addVersionCommand(prime, children)
-	children = addInfoCommand(prime, children)
-	children = addOtherCommands(prime, children)
+	children = addPackagesCommands(prime.Output(), children)
+	children = addRepositoryCommands(prime.Output(), children)
+	children = addProjectCommands(prime.Output(), children)
+	children = addVersionCommand(prime.Output(), children)
+	children = addInfoCommand(prime.Output(), children)
+	children = addOtherCommands(prime.Output(), children)
```
Naatan: I was only flagging this for your print functions, here it still makes sense to pass the whole prime structure as that's the convention.

cmd/state/internal/cmdtree/ppm.go
```diff
@@ -1,135 +1,134 @@
...
-	prime.Output().Print(strings.TrimSpace(locale.T("ppm_header_message")))
+func printDefault(out output.Outputer) error {
+	out.Print(strings.TrimSpace(locale.T("ppm_header_message")))
 	return nil
 }
 
-func printMain(prime *primer.Values) error {
-	prime.Output().Print(locale.T("ppm_print_main"))
+func printMain(out output.Outputer) error {
+	out.Print(locale.T("ppm_print_main"))
```
Naatan: Let's move this to the runner also, it looks out of place seeing this in the cmdtree.

internal/runners/ppm/shim.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/internal/runbits"
+	"github.com/ActiveState/cli/pkg/projectfile"
+)
+
+type Shim struct {
+	out         output.Outputer
+	intercepted string
+	replaced    string
+	localeID    string
```
Naatan, line 20: Please make these args to your shim function rather than struct parameters.

internal/runners/ppm/shim.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (s *Shim) shim(args ...string) error {
+	err := s.executeShim(args...)
+	if err != nil {
+		analytics.EventWithLabel(analytics.CatPPMShimCmd, "error", errs.Join(err, " :: ").Error())
+	} else {
+		analytics.EventWithLabel(analytics.CatPPMShimCmd, "success", fmt.Sprintf("intercepted=%s, replaced=%s", s.intercepted, s.replaced))
+	}
+	return err
+}
```
Naatan, line 72: Please get rid of this function and handle your execution and analytics directly from the invoker (one level higher). Each invocation should have its own analytics anyway, since right now you are loosing what the particular subcommand was.

Suggest:

category: CatPPMShimCmd
event: subcommand(s)
label: success | "error: %v" % errs.join()..

internal/runners/ppm/shim.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (s *Shim) executeShim(args ...string) error {
+	pj, fail := projectfile.GetSafe()
```
Naatan, line 75: This should be passed in through the prime structure.

internal/runners/ppm/shim.go
```diff
@@ -0,0 +1,107 @@
...
+	replacedArgs := args
+	if len(commands) > 1 {
+		s.replaced = commands[0]
+		replacedArgs = commands[1:]
+		replacedArgs = append(replacedArgs, args...)
+	}
+
+	forwarded := []string{"state", s.replaced}
+	forwarded = append(forwarded, replacedArgs...)
+	s.out.Print(locale.Tr(s.localeID, strings.Join(forwarded, " "), s.intercepted))
```
Naatan, line 100: This print belongs in the invoker, not here. Fine if it prints this and then errors out if an error happens, that'd just give the user more context if anything.

internal/runners/ppm/shim.go
MDrakos: Each invocation does have its own analytics, at least in terms of what's reported. The subcommand for ppm is captured in `s.intercepted`, which is now an argument.

PR #855, opened 2020-08-04
MSI cannot remove directory on uninstall

MDrakos:
https://www.pivotaltracker.com/story/show/173947659

It appears that this error can occur when there is an open handle to a file within the directory, though I wasn't able to reproduce. This could include: the directory being open in Explorer, an antivirus scanner using a file, backup software, and a number of other possibilities. This PR adds an error message to the user when we cannot remove a directory. We could also add a retry mechanism if we think it is necessary.

### DIFF
--- a/installers/msi-language/Uninstall/CustomAction.cs
+++ b/installers/msi-language/Uninstall/CustomAction.cs
@@ -47,7 +47,6 @@ namespace Uninstall
                 if (result.Equals(ActionResult.Failure))
                 {
                     session.Log("Could not remove installation directory");
-                    ActiveState.RollbarHelper.Report("Could not remove installation directory");
 
                     Record record = new Record();
                     record.FormatString = string.Format("Could not remove installation directory entry at: {0}, please ensure no files in the directory are currently being used and try again", installDir);
--- a/installers/msi-language/Uninstall/CustomAction.cs
+++ b/installers/msi-language/Uninstall/CustomAction.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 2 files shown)


installers/msi-language/Uninstall/CustomAction.cs
```diff
@@ -47,6 +48,11 @@ public static ActionResult Uninstall(Session session)
                 {
                     session.Log("Could not remove installation directory");
                     ActiveState.RollbarHelper.Report("Could not remove installation directory");
```
Naatan, line 50: Let's remove this since we're already logging it in Remove.Dir

installers/msi-language/Uninstall/CustomAction.cs
```diff
@@ -100,7 +106,7 @@ public static ActionResult Dir(Session session, string dir)
                 {
                     Directory.Delete(dir, true);
                 }
-                catch (IOException e)
+                catch (Exception e)
```
Naatan, line 109: I think the old one was better, any other exception we'd want to bubble up as we're not handling them..

https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.delete?view=netcore-3.1

MDrakos: This stemmed from my testing. When the `perl.exe` binary was being used and I tried to uninstall I got an `UnauthorizedAccessException`... I would expect an `IOException` in this case since that includes `being used by another process` but... Windows 🤷‍♂️

Naatan: huh ok, fair enough then

PR #856, opened 2020-08-04
only add force flag for symlink sub-command

mdrohmann:
https://www.pivotaltracker.com/story/show/173993679

### DIFF
--- a/installers/msi-language/CustomExitDlg.wxs
+++ b/installers/msi-language/CustomExitDlg.wxs
@@ -18,14 +18,14 @@
         <Control Id="StateToolWizardActionCheckBox" Type="CheckBox" X="130" Y="145" Width="10" Height="10" CheckBoxValue="1" Property="STATE_TOOL_WIZARD_CHECK_BOX" Hidden="yes">
           <Condition Action="show">NOT Installed</Condition>
         </Control>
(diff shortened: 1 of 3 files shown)


mdrohmann:
I am not quite sure how this ever worked ie., how we got to the error `"--force is not a valid option"`, because all the installation actions seemed to have not been executed during repair/reinstallation...

mdrohmann:
> Are you sure about this? I tested both the MSI on master and the one on your branch, the behavior on change was the same for both. I think `REINSTALL` is not the same as `modify`. There's a helpful cheat sheet here: https://resources.flexera.com/web/pdf/archive/is-chs-common-msi-conditions.pdf

Thanks for the link!  I think I understand what's going on now:  During my tests, I always re-ran the installer and when it presented the menu options ("Change", "Repair", "Remove"), I clicked on "Repair"(!) which translates to `REINSTALL~=ALL`.  *But* "Change" aka "Modify" does not translate to a `REINSTALL` condition.

Well, that is all interesting and explains the problems I encountered.  If I had clicked on "Change" instead everything would have looked just fine. But IMHO "Repair" should re-install everything, just like it does now with the changed conditionals.  What do you think?

MDrakos:
>But IMHO "Repair" should re-install everything, just like it does now with the changed conditionals. What do you think?

I think that makes sense. We don't present any options to the user for `change` so it appears to behave more like `repair`. We may want to remove the `change` option all together at some point, or actually support it.

PR #857, opened 2020-08-05
add project ID to run-command,activate GA event

mdrohmann:
https://www.pivotaltracker.com/story/show/173599100

### DIFF
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -251,9 +251,11 @@ func (c *Command) runner(cobraCmd *cobra.Command, args []string) error {
 	if outputFlag != nil && outputFlag.Changed {
 		analytics.CustomDimensions.SetOutput(outputFlag.Value.String())
 	}
-	subCommandString := strings.Join(c.subcommandNames(), " ")
-	analytics.Event(analytics.CatRunCmd, subCommandString)
-
+	// "activate" GA events are handled in the runners, in order that we can add project ID as label
+	if cobraCmd.Name() != "activate" {
+		subCommandString := strings.Join(c.subcommandNames(), " ")
+		analytics.Event(analytics.CatRunCmd, subCommandString)
+	}
 	// Run OnUse functions for non-persistent flags
 	c.runFlags(false)
 
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -27,9 +27,12 @@ type Command struct {
 	arguments []*Argument
 
 	execute func(cmd *Command, args []string) error
+
+	// deferAnalytics should be set if the command handles the GA reporting in its execute function
+	deferAnalytics bool
 }
 
-func NewCommand(name, description string, flags []*Flag, args []*Argument, executor Executor) *Command {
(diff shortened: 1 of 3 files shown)


internal/captain/command.go
```diff
@@ -27,9 +27,12 @@ type Command struct {
 	arguments []*Argument
 
 	execute func(cmd *Command, args []string) error
+
+	// deferAnalytics should be set if the command handles the GA reporting in its execute function
+	deferAnalytics bool
 }
 
-func NewCommand(name, description string, flags []*Flag, args []*Argument, executor Executor) *Command {
+func NewCommand(name, description string, flags []*Flag, args []*Argument, executor Executor, deferAnalytics ...bool) *Command {
```
Naatan, line 35: Please use a setter http://github.com/ActiveState/cli/blob/04e34ba136b9b2e3764188ac4022417a51049428/internal/captain/command.go#L190-L190

mdrohmann:
In the last commit, I removed the sending of the "build"/"project" GA event, as it will be moved/fixed by Mikes PR.  So, I will need to merge with his PR first...

internal/runners/activate/activate.go
```diff
@@ -81,7 +62,12 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
 		}
 	}
 
-	go sendProjectIDToAnalytics(params.Namespace, filepath.Join(targetPath, constants.ConfigFileName))
+	// Send google analytics event with label set to project namespace
+	names, fail := project.ParseNamespaceOrConfigfile(params.Namespace.String(), filepath.Join(targetPath, constants.ConfigFileName))
+	if fail != nil {
+		logging.Debug("error resolving namespace: %v", fail.ToError())
+	}
+	analytics.EventWithLabel(analytics.CatRunCmd, "activate", names.String())
```
Naatan, line 70: `params.Namespace.String()` is already what you want, no need to parse the config file again.

mdrohmann: I think that `params.Namespace.String()` can be empty as the namespace argument to the activate command is optional:
https://github.com/ActiveState/cli/blob/master/cmd/state/internal/cmdtree/activate.go#L46

So, if I just run `state activate` without any arguments, the namespace needs to be sourced from the as.yaml file.

internal/runners/activate/activate.go
```diff
@@ -81,7 +62,12 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
 		}
 	}
 
-	go sendProjectIDToAnalytics(params.Namespace, filepath.Join(targetPath, constants.ConfigFileName))
+	// Send google analytics event with label set to project namespace
+	names, fail := project.ParseNamespaceOrConfigfile(params.Namespace.String(), filepath.Join(targetPath, constants.ConfigFileName))
+	if fail != nil {
+		logging.Debug("error resolving namespace: %v", fail.ToError())
```
Naatan, line 68: Should have some sort of fallback mechanic, currently it will still call `names.String()` which will cause a nil pointer panic.

mdrohmann: Thanks for catching that.

PR #858, opened 2020-08-05
Clean up tutorial and PPM conversion UX

Naatan:


internal/runbits/invoke.go
```diff
@@ -29,6 +29,20 @@ func Invoke(out output.Outputer, args ...string) error {
...
+	out.Notice("[INFO]" + strings.Repeat("-", termWidth) + "[/RESET]\n")
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Invoke will invoke a state tool command with the given args and prints a friendly message indicating what we're doing
```
MDrakos, line 44: nit: this docstring isn't correct

locale/en-us.yaml
```diff
@@ -1573,19 +1573,23 @@ ppm_print_main:
...
 
-    Note that `{{.V0}}` is not a direct analog for `{{.V1}}` and your command may have failed as a result. 
-    Please run `{{.V0}} --help` for further information on how to use it for your use-case.
+    Note that `[BOLD][BLUE]{{.V0}}[/RESET]` is not a direct analog for `[BOLD][BLUE]{{.V1}}[/RESET]` and your command may have failed as a result.
+    Please run `[BOLD][BLUE]{{.V0}} --help[/RESET]` for further information on how to use it for your use-case.
 ppm_print_redundant:
   other: |
-    ppm area and its subcommands have been made redundant thanks to State Tool's use of virtual environments. 
-    You can simply start managing your packages, run `{{.V0}} --help` for more information.
+    `ppm area` and its subcommands have been made redundant thanks to State Tool's use of virtual environments.
```
MDrakos, line 1591: nit: this one is on me, but `ppm area` should probably not be hard coded in this locale entry incase we want to use it again.

Naatan: I'd say the problem here is the locale entry key, not the value.

PR #859, opened 2020-08-05
Track installation source

MDrakos:
https://www.pivotaltracker.com/story/show/173861281


installers/install.sh
```diff
@@ -382,6 +382,11 @@ update_rc_file() {
   echo "$pathenv" >> "$RC_FILE"
 }
 
+# Write install file
+STATEPATH=$INSTALLDIR/$STATEEXE
+CONFIGDIR=$($STATEPATH "export" "config" "--filter=dir")
+echo "install.sh" >> $CONFIGDIR/"install.txt"
```
mdrohmann, line 388: I think that we should probably replace the file here, as otherwise (if people install the state tool several times without cleaning the config directory) we could see installation sources "install.sh\ninstall.sh\install.sh".

```suggestion
echo "install.sh" > $CONFIGDIR/"install.txt"
```

installers/install.ps1
```diff
@@ -369,6 +369,13 @@ function install()
     }
     Move-Item (Join-Path $tmpParentPath $stateexe) $installPath
 
+    # Write install file
+    $StatePath = Join-Path -Path $installDir -ChildPath $script:STATEEXE
+    $Command = "$StatePath export config --filter=dir"
+    $ConfigDir = & Invoke-Expression $Command | Out-String
+    $InstallFilePath = Join-Path -Path $ConfigDir.Trim() -ChildPath "install.txt"
```
Naatan, line 376: Please name this `installsource.txt` or `installer.txt`, `install.txt` is too vague.

PR #862, opened 2020-08-06
Deprecation file fix

MDrakos:
https://www.pivotaltracker.com/story/show/174045570

The file not being present could be one of the causes of this report. If this continues we should have more actionable information once https://www.pivotaltracker.com/story/show/174219102 is merged.

### DIFF
--- a/internal/deprecation/deprecation.go
+++ b/internal/deprecation/deprecation.go
@@ -131,12 +131,15 @@ func (checker *Checker) check(versionNumber string) (*Info, *failures.Failure) {
 }
 
 func (checker *Checker) shouldFetch() bool {
+	if !fileutils.FileExists(checker.deprecationFile) {
+		return true
+	}
+
 	lastFetch := checker.config.GetTime(fetchKey)
-	if !lastFetch.IsZero() && time.Now().Before(lastFetch) && fileutils.FileExists(checker.deprecationFile) {
+	if !lastFetch.IsZero() && time.Now().Before(lastFetch) {
 		return false
 	}
 
-	checker.config.Set(fetchKey, time.Now().Add(15*time.Minute))
 	return true
 
 }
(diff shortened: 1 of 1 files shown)


internal/deprecation/deprecation.go
```diff
@@ -131,7 +132,7 @@ func (checker *Checker) check(versionNumber string) (*Info, *failures.Failure) {
 
 func (checker *Checker) shouldFetch() bool {
 	lastFetch := checker.config.GetTime(fetchKey)
-	if !lastFetch.IsZero() && time.Now().Before(lastFetch) {
+	if !lastFetch.IsZero() && time.Now().Before(lastFetch) && fileutils.FileExists(checker.deprecationFile) {
```
Naatan, line 135: This feels like the wrong place for this check. This conditional is about asserting whether it's time to do a check. Whether this file exists should not be part of this assertion. I think you need this check a little later in the process.

MDrakos: Fair enough. I moved the check for the file to before this timed check as it makes sense to fetch the info if the file doesn't exist regardless of the time of the last fetch.

internal/deprecation/deprecation.go
```diff
@@ -130,12 +131,15 @@ func (checker *Checker) check(versionNumber string) (*Info, *failures.Failure) {
...
+	if !fileutils.FileExists(checker.deprecationFile) {
+		return true
+	}
+
 	lastFetch := checker.config.GetTime(fetchKey)
 	if !lastFetch.IsZero() && time.Now().Before(lastFetch) {
 		return false
 	}
 
-	checker.config.Set(fetchKey, time.Now().Add(15*time.Minute))
```
Naatan, line 138: Rather than remove it here let's add it to the start with a defer. That way the behaviour remains centralized.

MDrakos: I would prefer to have it centralized as well, however adding a defer will mean that the `fetchKey` is set even when this function returns `false`. I could defer an anonymous func but that would be making this function more messy. I could also just call the `Set` function in two places here if keeping it centralized is what we want.

Naatan: hmm ok. How about we invert the conditional and only do the lastCheck stuff if the file exists? That's nearly the same as your first iteration, I know, except that GetTime doesn't get called and the logic is a bit more legible.

PR #863, opened 2020-08-06
Rollbar Fix

MDrakos:
https://www.pivotaltracker.com/story/show/174219726


internal/logging/rollbar.go
```diff
@@ -16,6 +16,7 @@ type delayedLog struct {
 var delayedLogs []delayedLog
 
 func SetupRollbar() {
+	rollbar.SetLogger(&rollbar.SilentClientLogger{})
```
Naatan, line 19: Nit: this wouldn't actually have made any difference.

PR #864, opened 2020-08-06
Fix integration tests

mdrohmann:
https://www.pivotaltracker.com/story/show/173824428


.github/workflows-src/jobs.lib.yml
```diff
@@ -80,6 +80,7 @@ steps:
   - #@ steps_setup()
   - #@ steps_preprocess()
   - #@ steps_build()
+  - #@ steps_genupdate()
```
mdrohmann, line 83: This needs to be run, such that the `TestUpdateBits` test works.

.github/workflows/build.yml
```diff
@@ -283,6 +287,7 @@ jobs:
       shell: bash
       run: |
         cd build
+        rm -Rf session-shared-build
```
mdrohmann, line 290: @Naatan It looks like you forgot to run `state run build-workflows` after a change.  So, this change ended up here...

internal/logging/rollbar.go
```diff
@@ -22,7 +22,6 @@ func SetupRollbar() {
 	if _, ok := rollbar.Custom()["UserID"]; !ok {
 		UpdateRollbarPerson("unknown", "unknown", "unknown")
 	}
-	fmt.Println("Using install source: ", config.InstallSource())
```
mdrohmann, line 25: Was this a debug line?  It broke all the JSON parser tests

internal/runners/languages/update.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/blang/semver"
+	goversion "github.com/hashicorp/go-version"
```
mdrohmann, line 8: This change was necessary, as we have a Python version on the Platform that is not semantic, namely `2.7.18.1`.

internal/testhelpers/e2e/session.go
```diff
@@ -81,7 +81,7 @@ func (s *Session) executablePath() string {
 	return exec
 }
 
-func New(t *testing.T, retainDirs bool) *Session {
+func New(t *testing.T, retainDirs bool, extraEnv ...string) *Session {
```
mdrohmann, line 84: I added functionality to define environment variables that are set during an entire session (for every `ConsoleProcess`)

.github/workflows-src/jobs.lib.yml
```diff
@@ -47,6 +47,7 @@ steps:
   - #@ steps_genupdate()
   - #@ steps_build_state_msi()
   - #@ steps_build_languages_msi()
+  - #@ steps_msi_inttest()
```
mdrohmann, line 50: The MSI language integration tests depend on the MSI, of course.  So I created a separate step for them and put them into the "build" job on Windows.  It's not very pretty, as a "test" is not a "build", which is why I filed a follow-up story: https://www.pivotaltracker.com/story/show/174260391

test/integration/edit_int_test.go
```diff
@@ -77,6 +77,9 @@ func (suite *EditIntegrationTestSuite) TestEdit() {
 }
 
 func (suite *EditIntegrationTestSuite) TestEdit_NonInteractive() {
+	if runtime.GOOS == "windows" && e2e.RunningOnCI() {
+		suite.T().Skip("Windows CI does not support ctrl-c interrupts.")
```
mdrohmann, line 81: Follow-up story is here: https://www.pivotaltracker.com/story/show/174260391

.github/workflows-src/steps.lib.yml
```diff
@@ -84,10 +84,18 @@ run: |
 #@ def steps_inttest():
 name: "Integration Tests"
 shell: bash
-run: go test -timeout 20m ./test/integration -v
+run: SHELL='' go test -timeout 20m -tags 'state_int_tests' ./test/integration -v
 env:
   INTEGRATION_TEST_USERNAME: ${{ secrets.INTEGRATION_TEST_USERNAME }}
   INTEGRATION_TEST_PASSWORD: ${{ secrets.INTEGRATION_TEST_PASSWORD }}
+  PLATFORM_API_TOKEN: ${{ secrets.PLATFORM_API_TOKEN }}
```
mdrohmann, line 91: @Naatan The secret still needs to be defined in Github.  But I do not have the permissions for that...

internal/runners/deploy/path_lin_win.go
```diff
@@ -28,11 +29,14 @@ func usablePath() (string, error) {
...
 
-		// Record result
+		// check for preferred PATHs
 		if funk.Contains(preferredPaths, path) {
 			return path, nil
 		}
-		result = path
+		// use the first available directory in PATH
+		if result == "" {
+			result = path
```
Naatan, line 38: This breaks the promise of the function; that it returns a "usable" path.

test/integration/deploy_int_test.go
```diff
@@ -86,29 +79,63 @@ func (suite *DeployIntegrationTestSuite) TestDeployPerl() {
...
+	cp.SendLine(errorLevel)
+	cp.Expect("0")
+
+	cp.SendLine("exit")
 	cp.ExpectExitCode(0)
 }
 
+func (suite *DeployIntegrationTestSuite) checkSymlink(name string, binDir, workDir string) {
+	// Linux symlinks to /usr/local/bin or the first write-able directory in PATH, so we can verify right away
+	if runtime.GOOS == "Linux" {
```
Naatan, line 120: Nit: just return early instead

internal/runners/deploy/path_lin_win.go
mdrohmann: I think it is still usable, no?

At this line, `path`  is a directory and writable ie., a path.  But now, I am only setting the *first* usable path to `result`.  Before, it was returning the *last* usable candidate, which created very strange results...  Usually, the first one on your PATH variable is the most important one...

.github/workflows-src/steps.lib.yml
Naatan: Done

internal/runners/deploy/path_lin_win.go
Naatan: Oh I see, I miss-interpreted this.

PR #865, opened 2020-08-06
Version increment fix

MDrakos:
https://www.pivotaltracker.com/story/show/174223623

In order to test this fix I will have to run the changes on master as everything is working correctly locally.

Confirmed that the `version.json` file is not being updated by downloading the session artifacts. This may have to do with each platform's build step updating and then uploading that file in parallel.


scripts/version-getter/main.go
```diff
@@ -0,0 +1,59 @@
+package main
```
Naatan, line 1: How is this different from 
```
wget -O base-version.json https://s3.ca-central-1.amazonaws.com/cli-update/update/state/version.json
```
 ?

PR #867, opened 2020-08-07
Rollbar install source

MDrakos:
https://www.pivotaltracker.com/story/show/174204075

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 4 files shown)


internal/config/config.go
```diff
@@ -48,6 +50,17 @@ func CachePath() string {
 	return defaultConfig.CachePath()
 }
 
+// InstallSource returns the source of the State Tool installation
+func InstallSource() string {
```
Naatan, line 54: Let's cache the result, since we're now accessing it in two places.

PR #870, opened 2020-08-10
Include commit ID in Rollbar logs

MDrakos:
https://www.pivotaltracker.com/story/show/174204147

### DIFF
--- a/scripts/msi-installer-builder/main.go
+++ b/scripts/msi-installer-builder/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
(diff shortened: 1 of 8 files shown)


scripts/msi-installer-builder/main.go
```diff
@@ -113,6 +121,16 @@ func normalize(preset languagePreset, c *config) (*config, error) {
 	return c, nil
 }
 
+func commitID() (string, error) {
+	cmd := exec.Command("git", "rev-parse", "HEAD")
```
Naatan, line 129: The `constants` package already provides this, please use that instead so we don't have multiple implementations of the same thing.

So `constants.RevisionHash`

PR #874, opened 2020-08-11
MSI logs only first critical error

MDrakos:
https://www.pivotaltracker.com/story/show/174204138

### DIFF
--- a/installers/msi-language/ActiveState/Rollbar.cs
+++ b/installers/msi-language/ActiveState/Rollbar.cs
@@ -73,13 +73,35 @@ namespace ActiveState
             person.UserName = userName;
             RollbarLocator.RollbarInstance.Config.Person = person;
         }
-
-        public static void Report(string message, IDictionary<string, object> customFields = null )
-        {
-            RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Error(new GenericException(message), customFields);
-        }
-
     }
+}
+
+public class RollbarReport
+{
+    private static RollbarReport _instance;
+    private static object syncLock = new object();
+
+    public static readonly TimeSpan RollbarTimeout = TimeSpan.FromSeconds(10);
+
+    protected RollbarReport() { }
+
+    public static void Critical(string message, IDictionary<string, object> customFields = null)
+    {
+        lock (syncLock)
+        {
+            if (_instance == null)
+            {
+                _instance = new RollbarReport();
+                RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Error(new GenericException(message), customFields);
+            }
+        }
+    }
+
+    public static void NonCritical(string message, IDictionary<string, object> customFields = null)
+    {
+        RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Error(new GenericException(message), customFields);
+    }
+
 }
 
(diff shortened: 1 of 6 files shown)


installers/msi-language/ActiveState/Rollbar.cs
```diff
@@ -73,13 +73,35 @@ private static void SetRollbarReportingUser(string id, string userName)
             person.UserName = userName;
             RollbarLocator.RollbarInstance.Config.Person = person;
         }
+    }
+}
+
+public class RollbarReport
+{
+    private static RollbarReport _instance;
```
Naatan, line 81: Why record the RollbarReport? Just make this a boolean called `NoFurtherReports` or something.

installers/msi-language/ActiveState/Rollbar.cs
```diff
@@ -73,13 +73,35 @@ private static void SetRollbarReportingUser(string id, string userName)
...
+            {
+                _instance = new RollbarReport();
+                RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Critical(new GenericException(message), customFields);
+            }
         }
+    }
 
+    public static void Error(string message, IDictionary<string, object> customFields = null)
+    {
+        RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Error(new GenericException(message), customFields);
```
Naatan, line 102: See AC:
>  When a critical error is encountered no further reports are send 

This goes for all errors.

installers/msi-language/Preset/CustomAction.cs
```diff
@@ -230,13 +230,13 @@ public static ActionResult InstallPreset(Session session)
...
                 {
-                    RollbarHelper.Report(string.Format("unexpected failure in Preset installation"));
+                    RollbarReport.Error(string.Format("unexpected failure in Preset installation"));
                 }
                 return res;
             }
             catch (Exception err)
             {
-                RollbarHelper.Report(string.Format("unknown error in language preset: {0}", err));
+                RollbarReport.Error(string.Format("unknown error in language preset: {0}", err));
```
Naatan, line 239: This should be critical. We want to record the source of the issue as critical, not the tail end, otherwise we still have the same problem.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -228,7 +229,7 @@ public static ActionResult InstallStateTool(Session session, out string stateToo
             {
                 string msg = string.Format("Could not update PATH. Attempted to set path to: {0}, encountered exception: {1}", newPath, e.ToString());
                 session.Log(msg);
-                ActiveState.RollbarHelper.Report(msg);
+                RollbarReport.Error(msg);
```
Naatan, line 232: This makes the install fail right? So should be critical

installers/msi-language/ActiveState/Rollbar.cs
MDrakos: Ahh okay, for some reason I interpreted this as no further _critical_ errors.

installers/msi-language/StateDeploy/CustomAction.cs
MDrakos: Yes, you're right. I think we updated this to specifically fail if we can't update PATH

PR #875, opened 2020-08-12
Win32 state tool installation

mdrohmann:
https://www.pivotaltracker.com/story/show/174224609

### DIFF
--- a/.github/workflows-src/jobs.lib.yml
+++ b/.github/workflows-src/jobs.lib.yml
@@ -1,4 +1,4 @@
-#@ load("steps.lib.yml", "steps_genversion", "steps_preprocess", "steps_build", "steps_genupdate", "steps_installgo", "steps_setup", "steps_test", "steps_inttest", "steps_uploadartifacts", "steps_uploadsessartifacts", "steps_downloadsessartifacts", "steps_deletesessartifacts", "steps_deploy", "steps_build_state_msi", "steps_build_languages_msi", "steps_cleanbuild", "steps_validate", "steps_sign", "steps_sign_installscript", "steps_build_installscripts", "steps_uploadsharedartifacts", "steps_downloadsharedartifacts", "steps_downloadallsessartifacts", "steps_sanitizeallsessartifacts")
+#@ load("steps.lib.yml", "steps_genversion", "steps_preprocess", "steps_build", "steps_build_32bit", "steps_genupdate", "steps_genupdate_32bit", "steps_installgo", "steps_setup", "steps_test", "steps_inttest", "steps_uploadartifacts", "steps_uploadsessartifacts", "steps_downloadsessartifacts", "steps_deletesessartifacts", "steps_deploy", "steps_build_state_msi", "steps_build_languages_msi", "steps_cleanbuild", "steps_validate", "steps_sign", "steps_sign_installscript", "steps_build_installscripts", "steps_uploadsharedartifacts", "steps_downloadsharedartifacts", "steps_downloadallsessartifacts", "steps_sanitizeallsessartifacts")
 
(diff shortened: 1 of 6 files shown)


activestate.yaml
```diff
@@ -39,7 +39,9 @@ scripts:
     value: |
       set -e
       $constants.SET_ENV
-      go build -tags "$GO_BUILD_TAGS" -o $constants.BUILD_TARGET $constants.CLI_BUILDFLAGS $constants.CLI_PKGS
+      export GOARCH=${1:-amd64}
+      export TARGET_SUFFIX=${1:-}
+      go build -tags "$GO_BUILD_TAGS" -o $constants.BUILD_TARGET${TARGET_SUFFIX} $constants.CLI_BUILDFLAGS $constants.CLI_PKGS
```
Naatan, line 44: I think this would produce a build target `state.exe386`, no?

activestate.yaml
```diff
@@ -39,7 +39,9 @@ scripts:
     value: |
       set -e
       $constants.SET_ENV
-      go build -tags "$GO_BUILD_TAGS" -o $constants.BUILD_TARGET $constants.CLI_BUILDFLAGS $constants.CLI_PKGS
+      export GOARCH=${1:-amd64}
```
Naatan, line 42: What's the dash for?

.github/workflows-src/jobs.lib.yml
```diff
@@ -41,10 +41,12 @@ steps:
   - #@ steps_setup()
   - #@ steps_preprocess()
   - #@ steps_build()
+  - #@ steps_build_32bit()
   - #@ steps_build_installscripts()
   - #@ steps_sign()
```
Naatan, line 46: This would need to be updated also

activestate.yaml
mdrohmann: It expands to "amd64" as a default value in case `$1` is empty or unset: https://i.stack.imgur.com/T2Fp8.png

activestate.yaml
mdrohmann: Yes, I agree that looks silly.  The update-generator scripts renames it to `windows-386.exe`, but I will change it, and make it an explicit `build32` script with a separate `BUILD_TARGET`...  Just a little verbose...

activestate.yaml
```diff
@@ -10,6 +10,12 @@ constants:
   - name: BUILD_TARGET
     if: ne .OS.Name "Windows"
     value: ./build/state
+  - name: BUILD_TARGET_32bit
+    if: eq .OS.Name "Windows"
+    value: ./build/state32.exe
+  - name: BUILD_TARGET_32bit
```
Naatan, line 16: nit: `bit` should be uppercase

activestate.yaml
```diff
@@ -67,6 +80,27 @@ scripts:
...
+  - name: sign-state-tool
+    if: eq .OS.Name "Windows"
+    language: bash
+    description: Signs the state tool with the signtool executable
+    value: |
+      set -e
+
+      # This assumes that a certificate file Cert.p12 file is unpacked in the current directory
+      signtool.exe sign -d "ActiveState State Tool" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ${constants.BUILD_TARGET}
+      signtool.exe sign -d "ActiveState State Tool (32bit)" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ${constants.BUILD_TARGET_32bit}
```
Naatan, line 103: This is using inputs defined only in the CI config, so a bit awkward to have here. Let's move it back to the CI config as I don't want to scope creep this story too much.

PR #880, opened 2020-08-13
Scheduled integration tests work again

MDrakos:
https://www.pivotaltracker.com/story/show/174364009

This PR ensures that the `integration tests` work on a scheduled basis.

Currently, the scheduled tests fail, because a `version.json` file needs to be generated.


.github/workflows-src/jobs.lib.yml
```diff
@@ -73,11 +73,13 @@ steps:
 name: "Integration Test"
 strategy: *strategy
 runs-on: *runs-on
+needs: [ prepare ]
```
mdrohmann, line 76: Do we still need the other changes, when integration test jobs are dependent on prepare?

.github/workflows/integration-test.yml
```diff
@@ -14,6 +14,8 @@ jobs:
         - macos-latest
         - windows-latest
     runs-on: ${{ matrix.platform }}
+    needs:
+    - prepare
```
mdrohmann, line 18: This workflow does not have a job called "prepare". Would that be necessary though?

MDrakos:
Closing this PR as the change set is captured in https://github.com/ActiveState/cli/pull/882

PR #882, opened 2020-08-13
Tag integration tests and allow to run subset only

mdrohmann:
https://www.pivotaltracker.com/story/show/174363982

This has been worked on during innovation days by Mike and Martin.  The following ACs are implemented:

- Every integration test is "tagged" with tags defining their scopes
- One tag is named "critical" and all tests that should always pass should be tagged with it
- If the environment variable `TEST_SUITE_TAGS` is set during a test run, only tests with the tags specified in the `TEST_SUITE_TAGS` value are run.  The value is `:`-separated list of tag names
- Tags can be set during CI by setting labels names `run-int-tests-only:<tagname>`


.github/workflows-src/steps.lib.yml
```diff
@@ -100,7 +100,11 @@ run: |
 #@ def steps_inttest():
 name: "Integration Tests"
 shell: bash
-run: SHELL='' go test -timeout 20m ./test/integration -v
+run: |
```
mdrohmann, line 103: Is it possible to make this smarter like this:

We look for all labels that are specified and start with `int-test-only: <label>` and then add the `<label>` to `TEST_SUITE_TAGS`? Potentially several ones?

.github/workflows-src/steps.lib.yml
```diff
@@ -100,7 +100,23 @@ run: |
...
+  echo "$LABELS"
+  IFS=',' read -r -a TESTS <<< "$LABELS"
+  for i in "${TESTS[@]}"
+  do
+    :
+    START=${i%:*}
+    if [ "$START" == "run-int-test-only" ]; then
+      TEST_SUITE=${i##*:}
+      echo $TEST_SUITE
+      export TEST_SUITE_TAGS=$TESTS_SUITE:$TEST_SUITE_TAGS
```
mdrohmann, line 114:
```suggestion
      export TEST_SUITE_TAGS=$TEST_SUITE:$TEST_SUITE_TAGS
```

test/integration/activate_int_test.go
```diff
@@ -16,28 +16,33 @@ import (
...
+	"github.com/ActiveState/cli/internal/testhelpers/tagsuite"
 )
 
 type ActivateIntegrationTestSuite struct {
-	suite.Suite
+	tagsuite.Suite
 }
 
 func (suite *ActivateIntegrationTestSuite) TestActivatePython3() {
+	suite.OnlyRunForTags("python", "activate", "critical")
```
Naatan, line 27: Can we centralize these tag values somewhere as constants and just use the constants here. Then we can both ensure that the tags are consistent as well as have a central place to check for which tags are available.

Naatan (changes requested):

internal/testhelpers/tagsuite/tagsuite.go
```diff
@@ -8,6 +8,45 @@ import (
 	"github.com/thoas/go-funk"
 )
 
+const (
+	Activate      = "activate"
+	Alternative   = "alternative"
+	Auth          = "auth"
+	CLIDeploy     = "cli-deploy"
```
Naatan, line 15: Why does this one have a cli prefix? Consider my OCD triggered :p

internal/testhelpers/tagsuite/tagsuite.go
```diff
@@ -8,6 +8,45 @@ import (
...
+	Pull          = "pull"
+	Push          = "push"
+	Python        = "python"
+	Run           = "run"
+	Scripts       = "scripts"
+	Secrets       = "secrets"
+	Shell         = "shell"
+	Shim          = "shim"
+	Show          = "show"
+	Update        = "Update"
```
Naatan, line 46:
```suggestion
	Update        = "update"
```

All the other ones are lowercase.

Naatan (changes requested):

internal/testhelpers/tagsuite/tagsuite.go
MDrakos: Because this is used in the `TestUpdateBits` integration test. So the tag is for integration tests that ensure the cli has deployed properly. I didn't add the tag but I think that was @mdrohmann's intention.

Naatan (approved):

PR #885, opened 2020-08-14
Support multiple activestate.yamls

Naatan:
https://www.pivotaltracker.com/story/show/174343681


pkg/projectfile/projectfile.go
```diff
@@ -437,17 +439,40 @@ var persistentProject *Project
...
+		}
+
+		secondaryProject, fail := parse(filepath.Join(projectDir, file.Name()))
+		if fail != nil {
+			return nil, fail
+		}
+		if err := mergo.Merge(project, *secondaryProject, mergo.WithAppendSlice); err != nil {
+			return nil, failures.FailMarshal.Wrap(err, locale.Tl("err_merge_project", "Could not merge {{.V0}} into your activestate.yaml", file.Name()))
+		}
 	}
```
daved, line 476: nit: If you gather the files using a glob, you can be sure that all the returned file names are intended to be merged.
https://github.com/ActiveState/cli/compare/dgreen/exp-asy_split#diff-6a2389bce029e4a55df04d9f53423c45R20-R46

This should be faster (Readdir is heavy-handed, imo) and the code should be simpler. Using split is safe because the glob guarantees that the filenames will have three sections, but I'm not strongly suggesting to not use regex.

pkg/projectfile/projectfile.go
```diff
@@ -437,17 +439,40 @@ var persistentProject *Project
...
-	project.path = configFilepath
+	project, fail := parse(configFilepath)
+	if fail != nil {
+		return nil, fail
+	}
 
-	if err != nil {
-		return nil, FailParseProject.New(locale.T("err_project_parse", map[string]interface{}{"Error": err.Error()}))
+	re, _ := regexp.Compile(`activestate.(\w+).yaml`)
+	for _, file := range files {
```
daved, line 454: nit: fileInfo and fileInfos would better reflect what type is being handled here.

PR #887, opened 2020-08-17
Fix State Tool wizard launch

MDrakos:
https://www.pivotaltracker.com/story/show/174224476

Discovered during testing of the MSI on Windows 7. The custom action being used to launch the State Tool wizard is immediate. This means that the property `STATE_TOOL_PATH` will always be `state.exe` and not the actual path set in the deferred custom actions. In the State Tool installation custom action we force the installation to `<AppDataDir>/ActiveState/bin`, so I've used that here.


MDrakos:
@mdrohmann I tried a few different approaches but admittedly did not investigate for long as I wanted to get something functional. I would prefer not to have the hard-coded state tool path in the product.wxs file.

I attempted scheduling the custom action as both `deferred` and `commit`. This custom action fires after the `InstallFinalize` step which produces an error in both cases as the custom action must be scheduled between `InstallInitialize` and `InstallFinalize` if it is not `immediate`. I did not try other sequences.

MDrakos:
Good point re: previously installed state tool. The property `STATE_TOOL_PATH` is properly set if the state tool is already installed, so we can use the original custom action there, and fallback to the new one if the state tool wasn't installed before the language installation.

PR #889, opened 2020-08-18
State pacakges --output=json has json output

mdrohmann:
https://www.pivotaltracker.com/story/show/174370744

### DIFF
--- a/internal/runners/packages/table.go
+++ b/internal/runners/packages/table.go
@@ -4,44 +4,69 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
 	"github.com/bndr/gotabulate"
 )
 
-type table struct {
-	headers []string
-	data    [][]string
+type packageTable struct {
+	rows        []packageRow
+	emptyOutput string // string returned when table is empty
 }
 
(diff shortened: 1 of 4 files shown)


internal/runners/packages/table.go
```diff
@@ -4,44 +4,69 @@ import (
...
+	headers := []string{
+		locale.T("package_name"),
+		locale.T("package_version"),
+	}
+
+	data := make([][]string, 0, len(t.rows))
+	for _, row := range t.rows {
+		data = append(data, []string{row.Pkg, row.Version})
+	}
+	tab := gotabulate.Create(data)
```
Naatan, line 59: We should not be using gotabulate here at all, the outputter facilitates table output.

eg. http://github.com/ActiveState/cli/blob/cf6e4835824104807fc163d8d79302b95214ffab/internal/runners/events/events.go#L27-L27

mdrohmann: I see.  I looked at the scripts implementation first: https://github.com/ActiveState/cli/blob/master/internal/runners/scripts/scripts.go#L45-L51  where we are still using gotabulate .

But yes, I will change it this way.  That's better for sure...

Naatan: I'll file a bug for scripts, thanks!

PR #894, opened 2020-08-19
Send GA pixel in language MSI

mdrohmann:
https://www.pivotaltracker.com/story/show/174391493

I tested it on Windows 10 and Windows 8.1.

Unfortunately, the google analytics web request has to be done synchronously - adding a short delay to the initialization of the dialog.  I set the web client timeout to a low limit for that reason, such that it does not get "too long".


installers/msi-language/GAPixel/GAPixelCA.cs
```diff
@@ -0,0 +1,256 @@
...
+                        continue;
+                    }
+                }
+            }
+            catch (Exception err)
+            {
+                session.Log(String.Format("Error getting unique ID {0}", err));
+            }
+            // fallback method
+            return Guid.NewGuid().ToString();
```
Naatan, line 53: Can we use something static as a fallback? That way we can identify if this is in fact not working.

installers/msi-language/GAPixel/GAPixelCA.cs
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+
+
+    // modified from https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#query-the-registry-using-code
+    // and from https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#query-the-registry-using-code-older-framework-versions
+    public class GetDotNetVersion
```
Naatan, line 61: This is doing way too much for the value it brings, please discard it. We just won't track this.

mdrohmann: Ok.

mdrohmann:
So, in standup we decided not to send GA event for Windows 10.  Turns out that there is a [VersionNT](https://docs.microsoft.com/en-us/windows/win32/msi/versionnt) property.  But I should have read the documentation properly:  It is not send for Windows 10 anymore. 🤦 That was not very effective.  [Here](https://support.microsoft.com/en-us/help/3202260/versionnt-value-for-windows-10-and-windows-server-2016) the write this madness out in words, and suggest to do a `setup.exe` instead. 😬 

Did I mention that everything that combines versioning and Microsoft drives me insane?  Maybe this is the reason why they do not have deprecation policies.  It is just to difficult to figure out for them what version of stuff they are dealing with.

Anyways:  I just added the check to the custom action.  And that seems to work.

Also interesting: I wrote this message twice now, because it turns out that when you update Visual Studio, the installer ends with a dialog that has a "Restart" button.  Clicking on it, does restart Visual Studio... and your computer.  I love Microsoft!!!

installers/msi-language/GAPixel/GAPixelCA.cs
```diff
@@ -74,6 +75,11 @@ protected override WebRequest GetWebRequest(Uri uri)
         public static ActionResult SendPixel(Session session)
         {
             var wv = System.Environment.OSVersion.VersionString;
+            if (System.Environment.OSVersion.Version.Major > 6)
+            {
+                session.Log("skipping GA tracking for Windows > 8.1 version is: {0} major version: {1}", wv, System.Environment.OSVersion.Version.Major);
+                return ActionResult.Success;
+            }
```
Naatan, line 82:
```suggestion
```

If we're already in the custom action we might as well fire it. The concern was not having any delays on Windows 10 which I now realize is moot anyway as it wouldn't be firing the event anyway.

Naatan (dismissed):

mdrohmann:
I had to merge with master first, because we now have a `CustomFatalError` dialog.

PR #896, opened 2020-08-19
Scripts support multiple languages

MDrakos:
https://www.pivotaltracker.com/story/show/174264578


internal/runners/run/run.go
```diff
@@ -76,53 +76,64 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
...
-		venv.OnInstallArtifacts(func() { print.Line(locale.T("installing_artifacts")) })
-
-		if fail := venv.Activate(); fail != nil {
-			logging.Errorf("Unable to activate state: %s", fail.Error())
-			return fail.WithDescription("error_state_run_activate")
+	// IsActivated is unreliable currently. We store the activated
+	// value in a varaible to ensure we are not activating multiple
+	// types in the loop below
+	activated := subshell.IsActivated()
+	for _, l := range script.Languages() {
```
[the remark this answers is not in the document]
MDrakos: That's what the `activated` bool is for, however the activation logic can be moved out of the loop all together

internal/runners/run/run.go
```diff
@@ -113,16 +99,31 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
...
 		subs.SetEnv(env)
 
-		// get the "clean" path (only PATHS that are set by venv)
+		// search the "clean" path first (PATHS that are set by venv)
 		env, err = venv.GetEnv(false, "")
 		if err != nil {
 			return err
 		}
-		path = env["PATH"]
+		path = env["PATH"] + string(os.PathSeparator) + path
```
Naatan, line 107: This doesn't seem related to this PR, and reading through you would just end up with PATH being in `path` twice.

internal/runners/run/run.go
```diff
@@ -113,16 +99,31 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
...
+			lang = l
+			break
+		}
+		attempted = append(attempted, l.String())
+	}
+
+	if !lang.Recognized() {
+		warning := locale.Tl(
+			"run_warn_deprecated_script_without_language",
+			"[YELLOW]DEPRECATION WARNING: Scripts without a recognized or defined language currently fall back to using the default shell for your platform. This fallback mechanic will soon stop working and a language will need to be explicitly defined for each script. Please configure the 'language' field with a valid option (one of {{.V0}})[/RESET]",
```
Naatan, line 121: You're handling two different conditions under the same warning.

Behaviour 1 = no language defined, print warning and default to something.

Behaviour 2 = language not valid, return error and exit nonzero.

internal/runners/run/run.go
```diff
@@ -132,10 +133,22 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
...
 	}
 	defer sf.Clean()
-
 	print.Info(locale.Tr("info_state_run_running", script.Name(), script.Source().Path()))
 	// ignore code for now, passing via failure
-	return subs.Run(sf.Filename(), args...)
+	err := subs.Run(sf.Filename(), args...)
+	if err != nil {
+		if len(attempted) > 0 {
+			return locale.WrapError(
```
Naatan, line 141: This should be an input error

internal/runners/run/run.go
```diff
@@ -145,11 +158,20 @@ func configCachePath() string {
 	return config.CachePath()
 }
 
-func pathProvidesExec(filterByPath, exec, path string) bool {
+func pathProvidesLang(path string, language language.Language) bool {
 	paths := splitPath(path)
-	if filterByPath != "" {
-		paths = filterPrefixed(filterByPath, paths)
```
Naatan, line 151: Wouldn't you still want this?

pkg/project/project.go
```diff
@@ -603,15 +603,23 @@ func (script *Script) SourceScript() *projectfile.Script { return script.script
...
+		languages[i] = language.MakeByName(strings.TrimSpace(lang))
+	}
+	return languages
 }
 
-// LanguageSafe returns the language of this script. The returned
-// language is guaranteed to be of a known scripting language
+// LanguageSafe returns the first language of this script. The
+// returned language is guaranteed to be of a known scripting language
 func (script *Script) LanguageSafe() language.Language {
```
Naatan, line 618: This should return a list as well. Otherwise you're implicitly giving users of this function limited / legacy behaviour.

pkg/projectfile/projectfile.go
```diff
@@ -391,14 +390,14 @@ func MakeEventsFromConstrainedEntities(items []ConstrainedEntity) (events []*Eve
...
-	Standalone  bool              `yaml:"standalone,omitempty"`
-	Language    language.Language `yaml:"language,omitempty"`
-	Conditional Conditional       `yaml:"if"`
-	Constraints Constraint        `yaml:"constraints,omitempty"`
+	Name        string      `yaml:"name"`
+	Description string      `yaml:"description,omitempty"`
+	Value       string      `yaml:"value"`
+	Filename    string      `yaml:"filename,omitempty"`
+	Standalone  bool        `yaml:"standalone,omitempty"`
+	Language    string      `yaml:"language,omitempty"`
```
Naatan, line 398: Why not `[]language.Language` ?

internal/runners/run/run.go
MDrakos: `GetEnv` is not inheriting the environment so the entry at `env["PATH"]` only contains the cache dir.

internal/runners/run/run.go
MDrakos: No not with the updates, though I've since updated the caller of this function and hopefully it's more clear now

pkg/projectfile/projectfile.go
MDrakos: This is to avoid breaking changes to the `activestate.yaml`. If we use `[]language.Language` then the field value has to become a list and we have to represent it with `[entry, entry]` or 

```
list:
  - entry
  - entry
```

internal/runners/run/run.go
```diff
@@ -104,21 +100,48 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
...
+			"The language for this script is not supported. Please configure the 'language' field with a valid option (one, or more, of {{.V0}})", strings.Join(language.RecognizedNames(), ", "),
+		)
 	}
 
-	if !lang.Recognized() {
+	if lang == language.Unset {
 		warning := locale.Tl(
 			"run_warn_deprecated_script_without_language",
-			"[YELLOW]DEPRECATION WARNING: Scripts without a recognized or defined language currently fall back to using the default shell for your platform. This fallback mechanic will soon stop working and a language will need to be explicitly defined for each script. Please configure the 'language' field with a valid option (one of {{.V0}})[/RESET]",
+			"[YELLOW]DEPRECATION WARNING: Scripts without defined language currently fall back to using the default shell for your platform. This fallback mechanic will soon stop working and a language will need to be explicitly defined for each script. Please configure the 'language' field with a valid option (one of {{.V0}})[/RESET]",
```
Naatan, line 144:
```suggestion
			"[YELLOW]DEPRECATION WARNING: Scripts without a defined language currently fall back to using the default shell for your platform. This fallback mechanic will soon stop working and a language will need to be explicitly defined for each script. Please configure the 'language' field with a valid option (one of {{.V0}})[/RESET]",
```

internal/runners/run/run.go
```diff
@@ -104,21 +100,48 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
...
-		attempted = append(attempted, l.String())
+		attempted = append(attempted, lang.String())
+	}
+
+	if script.Standalone() && !lang.Executable().Builtin() {
+		return FailStandaloneConflict.New("error_state_run_standalone_conflict")
+	}
+
+	if lang == language.Unknown {
+		return locale.NewError(
```
Naatan, line 135: Should be an input error

internal/runners/scripts/edit.go
```diff
@@ -94,7 +94,7 @@ func createScriptFile(script *project.Script, expand bool) (*scriptfile.ScriptFi
 		scriptBlock = script.Value()
 	}
 
-	f, fail := scriptfile.NewAsSource(script.LanguageSafe(), script.Name(), scriptBlock)
+	f, fail := scriptfile.NewAsSource(script.LanguageSafe()[0], script.Name(), scriptBlock)
```
Naatan, line 97: Do I even have to say it? :p

pkg/project/expander.go
```diff
@@ -177,7 +177,7 @@ func expandPath(name string, script *Script) (string, *failures.Failure) {
 		return script.cachedFile(), nil
 	}
 
-	sf, fail := scriptfile.NewEmpty(script.LanguageSafe(), name)
+	sf, fail := scriptfile.NewEmpty(script.LanguageSafe()[0], name)
```
Naatan, line 180: Same

internal/runners/scripts/edit.go
MDrakos: lol yeah I went back and forth on this one as it shouldn't be possible for `LanguageSafe` to return an empty slice. Updated the usage to fallback to the default scripting languages

Naatan (dismissed):

internal/runners/run/run.go
```diff
@@ -75,7 +75,7 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
 		return fail
 	}
 
-	path := os.Getenv("PATH")
+	var venvPath string
```
Naatan, line 78: Can we rename this to `venvExePath` and maybe add a comment explaining what it does. I realize you didn't introduce this but reviewing it it was a stumbling block figuring out what it was used for.

internal/runners/run/run.go
```diff
@@ -99,31 +99,41 @@ func run(out output.Outputer, subs subshell.SubShell, name string, args []string
...
 
-		var exec string
-		if lang.Executable().Available() {
-			exec = lang.Executable().Name()
+		lang = language.MakeByShell(subs.Shell())
+	}
+
+	var attempted []string
+	for _, l := range script.Languages() {
+		var path, exec string
```
Naatan, line 119: And let's rename this to `exePath`

Naatan (changes requested):

internal/runners/run/run.go
```diff
@@ -79,7 +79,11 @@ func run(out output.Outputer, subs subshell.SubShell, proj *project.Project, nam
 		return fail
 	}
 
-	var venvPath string
+	// venvExePath stores a virtual environment's PATH value. If the script
+	// requires activation this is the PATH we should be searching for
+	// executables in.
+	path := os.Getenv("PATH")
```
Naatan, line 85: Looks like you forgot to rename the actual variable (to `venvExePath`).

Naatan (changes requested):

test/integration/run_int_test.go
```diff
@@ -203,13 +208,63 @@ func (suite *RunIntegrationTestSuite) TestRun_DeprecatedLackingLanguage() {
...
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+	)
+	cp.Expect("Activating state: ActiveState-CLI/Python3")
+	cp.WaitForInput(120 * time.Second)
+
+	cp.SendLine(fmt.Sprintf("%s run helloWorldMultiple", cp.Executable()))
+	cp.Expect("3")
+	cp.SendLine("exit")
+	cp.ExpectExitCode(0)
+}
```
Naatan: I feel like we can cover this with our existing activation tests, no need to create new ones just to make this one single assertion.

ie. do this assertion in one of the tests we already have for running activations.

Naatan (changes requested):

Naatan (approved):

Naatan (approved):

PR #897, opened 2020-08-20
Report MSI stages to GA

mdrohmann:
https://www.pivotaltracker.com/story/show/174407605


installers/msi-language/ActiveState/Tracker.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                Category = category,
+                Action = action,
+                Label = label,
+                Value = value,
+            };
+
+            eventTrackingParameters.ClientId = this._cid;
+
+            return await this._tracker.TrackAsync(eventTrackingParameters);
```
Naatan, line 44: Nit: why use async if you're going to await anyway?

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -253,6 +245,35 @@ public static ActionResult InstallStateTool(Session session, out string stateToo
...
+                return ActionResult.Success;
+            }
+
+            Status.ProgressBar.StatusMessage(session, "Installing State Tool...");
+            Status.ProgressBar.Increment(session, 1);
+
+            var ret = _installStateTool(session, out stateToolPath);
+            if (ret == ActionResult.Success)
+            {
+                TrackerSingleton.Instance.TrackEventInBackground(session, "stage", "state-tool", "success");
```
Naatan, line 270: Let's also trigger this if it's installed but use the value "skipped".

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -253,6 +245,35 @@ public static ActionResult InstallStateTool(Session session, out string stateToo
...
+            }
+            else if (ret == ActionResult.Failure)
+            {
+                TrackerSingleton.Instance.TrackEventInBackground(session, "stage", "state-tool", "failure");
```
Naatan, line 274: Wouldn't this also amount to a "finished" "failure"?

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -386,8 +410,18 @@ public static ActionResult StateDeploy(Session session)
                         record.FormatString = String.Format("{0} failed with error:\n{1}", seq.Description, errorOutput);
 
                         MessageResult msgRes = session.Message(InstallMessage.Error | (InstallMessage)MessageBoxButtons.OK, record);
+                        if (!artifactsFetched)
+                        {
+                            TrackerSingleton.Instance.TrackEventSynchronously(session, "stage", "artifacts", "failure");
```
Naatan, line 415: Artifacts in this case refers to the whole deployment job. So you can call this for the entirety of the deploy process.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -398,6 +432,7 @@ public static ActionResult StateDeploy(Session session)
                 return ActionResult.Failure;
             }
 
+            TrackerSingleton.Instance.TrackEventSynchronously(session, "stage", "finished", "success");
```
Naatan, line 435: There's nothing else that can fail after this? Can't we track this a bit closer to the actual success state of the MSI installation?

installers/msi-language/StateDeploy/CustomAction.cs
mdrohmann: All "finished" commands are now triggered after the finishing MSI dialogs are displayed (CustomExitDialog => success, CustomFatalError => failure).

installers/msi-language/StateDeploy/CustomAction.cs
mdrohmann: All "finished" commands are now triggered after the finishing MSI dialogs are displayed (CustomExitDialog => success, CustomFatalError => failure).

installers/msi-language/ActiveState/Tracker.cs
mdrohmann: The short answer: `TrackAsync` is what the C# library provides.  

The long answer: Every async function is supposed to have an `await` in C#.  The way I understand the idea behind it: You can write your control flow, as if you were writing a synchronized control flow, but things are actually running in threads and in parallel with nicer functionality to handle issues that occur with concurrency, like cancellation and callbacks.  I think every time you hit an `await` statement, the thread that you are executing in, could also change...

PR #899, opened 2020-08-20
Replace `print` with `output`

Naatan:
https://www.pivotaltracker.com/story/show/174262113


pkg/platform/authentication/auth.go
```diff
@@ -241,7 +241,7 @@ func (s *Auth) Client() *mono_client.Mono {
 	if !s.Authenticated() {
 		if fail := s.Authenticate(); fail != nil {
 			logging.Error("Trying to get the Client while not authenticated")
-			print.Error(locale.T("err_api_not_authenticated"))
+			fmt.Fprint(os.Stderr, locale.T("err_api_not_authenticated"))
```
mdrohmann, line 244: Why `fmt` here? Is there a reason not to use the `output` library here?

state/secrets/get.go
```diff
@@ -53,11 +54,11 @@ func (cmd *Command) ExecuteGet(_ *cobra.Command, args []string) {
 			if secret.IsUser() {
 				err = "secrets_err_user_not_defined"
 			}
-			print.Error(locale.Tr(err, cmd.Args.Name))
+			fmt.Fprint(os.Stderr, locale.Tr(err, cmd.Args.Name))
```
mdrohmann, line 57: Can't we use the `outputer` library in this file?

test/integration/activate_int_test.go
```diff
@@ -246,8 +246,8 @@ func (suite *ActivateIntegrationTestSuite) TestInit_Activation_NoCommitID() {
 		e2e.WithArgs("activate"),
 		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
 	)
-	cp.Expect(locale.Tr("err_project_no_commit", url))
 	cp.ExpectExitCode(1)
+	suite.Contains(e2e.CleanOutput(cp.Snapshot()), locale.Tr("err_project_no_commit", url))
```
mdrohmann: This can now use `cp.ExpectLongString()`

test/integration/package_int_test.go
```diff
@@ -187,8 +186,8 @@ func (suite *PackageIntegrationTestSuite) TestPackage_searchWithExactTermWrongTe
 	suite.PrepareActiveStateYAML(ts)
 
 	cp := ts.Spawn("packages", "search", "xxxrequestsxxx", "--exact-term")
-	cp.Expect("Currently no package of the provided name is available on the ActiveState Platform")
 	cp.ExpectExitCode(0)
+	suite.Contains(e2e.CleanOutput(cp.Snapshot()), "Currently no package of the provided name is available on the ActiveState Platform")
```
mdrohmann, line 190: Can use `cp.ExpectLongString()`

test/integration/package_int_test.go
```diff
@@ -209,8 +208,8 @@ func (suite *PackageIntegrationTestSuite) TestPackage_searchWithWrongLang() {
 	suite.PrepareActiveStateYAML(ts)
 
 	cp := ts.Spawn("packages", "search", "numpy", "--language=perl")
-	cp.Expect("Currently no package of the provided name is available on the ActiveState Platform")
 	cp.ExpectExitCode(0)
+	suite.Contains(e2e.CleanOutput(cp.Snapshot()), "Currently no package of the provided name is available on the ActiveState Platform")
```
mdrohmann, line 212: Can use `cp.ExpectLongString()`

test/integration/push_int_test.go
```diff
@@ -36,8 +36,9 @@ func (suite *PushIntegrationTestSuite) TestInitAndPush() {
 
 	wd := filepath.Join(cp.WorkDirectory(), namespace)
 	cp = ts.SpawnWithOpts(e2e.WithArgs("push"), e2e.WithWorkDirectory(wd))
-	cp.Expect(fmt.Sprintf("Project created at https://%s/%s/%s with language %s", constants.PlatformURL, username, pname, "python3"))
 	cp.ExpectExitCode(0)
+	suite.Contains(e2e.CleanOutput(cp.TrimmedSnapshot()), fmt.Sprintf("Project created at"))
```
mdrohmann, line 40: Can use `cp.ExpectLongString()`

internal/testhelpers/e2e/clean.go
```diff
@@ -79,3 +80,8 @@ func deleteUser(name string) error {
 
 	return nil
 }
+
+func CleanOutput(v string) string {
```
mdrohmann, line 84: This should not be needed, as we know have the `cp.ExpectLongString()` method...

internal/output/editor0.go
```diff
@@ -21,6 +21,7 @@ func (f *EditorV0) Type() Format {
 
 func NewEditorV0(config *Config) (EditorV0, *failures.Failure) {
 	json, fail := NewJSON(config)
+	json.printNUL = false
```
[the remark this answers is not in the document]
Naatan: No, this was something added in the `output` package, which postdates Komodo work.

internal/output/plain.go
```diff
@@ -294,6 +307,10 @@ func sprintTable(slice []interface{}) (string, error) {
 		rows = append(rows, row)
 	}
 
+	if termWidth == 0 {
```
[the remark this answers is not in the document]
Naatan: In theory yes, but it could still be calculating a lower value on line 288. This is just an extra safeguard.

internal/runners/organizations/organizations.go
```diff
@@ -1,110 +1,80 @@
...
-		if fail != nil {
-			return fail.WithDescription("organizations_err_output")
-		}
-
-		print.Line(string(data))
-		return nil
-	default:
-		listOrganizations(orgs)
+	if len(orgs) == 0 {
+		out.Notice(locale.T("organization_no_orgs"))
```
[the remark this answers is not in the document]
Naatan: That's what `Notice` does; http://github.com/ActiveState/cli/blob/b271676c134d0ae2ff85ee37cbb4a31014af8b38/internal/output/json.go#L59-L59

pkg/platform/authentication/auth.go
Naatan: I was still trying to avoid having a global for `output` when I wrote this, now that I see it again I might as well rewrite it to use the global, least then we're aware of the legacy dependency.

state/secrets/get.go
Naatan: Secrets is entirely legacy, using outputer would require serious refactoring. I'd rather do this in a separate story in one go.

PR #901, opened 2020-08-21
State tutorial locale fix

MDrakos:
https://www.pivotaltracker.com/story/show/174185346

Tested this change manually with a few commands and the output appears as expected.

### DIFF
--- a/internal/locale/locale.go
+++ b/internal/locale/locale.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"text/template"
@@ -157,7 +158,8 @@ func Tt(translationID string, args ...interface{}) string {
 	// this very painful otherwise
 
 	// Replace newlines in yaml strings with space to avoid concatenated words
-	translation = strings.Replace(translation, "\n", " ", -1)
+	replaceRegex := regexp.MustCompile(`[\ ]?\n(\w|\{)`)
+	translation = replaceRegex.ReplaceAllString(translation, " $1")
 	translation = strings.Replace(translation, "{{BR}}", "\n", -1)
 	// Avoid indentation after newlines
 	translation = strings.Replace(translation, "\n ", "\n", -1)
--- a/internal/locale/locale.go
+++ b/internal/locale/locale.go
@@ -155,8 +155,12 @@ func Tt(translationID string, args ...interface{}) string {
 
 	// For templates we want to manually specify the linebreaks as the way YAML gets parsed makes
 	// this very painful otherwise
-	translation = strings.Replace(translation, "\n", "", -1)
+
+	// Replace newlines in yaml strings with space to avoid concatenated words
+	translation = strings.Replace(translation, "\n", " ", -1)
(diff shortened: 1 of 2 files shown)


internal/locale/locale.go
```diff
@@ -155,8 +156,13 @@ func Tt(translationID string, args ...interface{}) string {
 
 	// For templates we want to manually specify the linebreaks as the way YAML gets parsed makes
 	// this very painful otherwise
-	translation = strings.Replace(translation, "\n", "", -1)
+
+	// Replace newlines in yaml strings with space to avoid concatenated words
+	replaceRegex := regexp.MustCompile(`[\ ]?\n(\w|\{)`)
```
Naatan, line 161: think we could simplify this to `s/\s*\n/ /`

internal/locale/locale.go
```diff
@@ -155,8 +156,13 @@ func Tt(translationID string, args ...interface{}) string {
...
+	translation = replaceRegex.ReplaceAllString(translation, " $1")
 	translation = strings.Replace(translation, "{{BR}}", "\n", -1)
+	// Avoid indentation after newlines
+	translation = strings.Replace(translation, "\n ", "\n", -1)
```
Naatan, line 165: Can drop this if you use my suggestion

locale/en-us.yaml
```diff
@@ -1654,12 +1653,10 @@ ppm_convert_explanation:
     State Tool, like Perl 7, is being developed from the ground up with modern software development practices in mind.
     Best practices suggest one environment per project, which is the State Tool’s default option.
     State Tool was developed from the ground up with modern software development practices in mind.
-    [[BR]]
 ppm_convert_create_question:
   other: |
     PPM functionality is being replaced by the State Tool, which does things differently. For example, before you can
     start managing dependencies, you need to create a new virtual environment for your Perl project.
-    [[BR]]
```
Naatan, line 1662: These are here for a reason, don't remove them.

internal/locale/locale.go
MDrakos: I've updated this to `\s*\n` which works in my testings. `/s` matches to the characters `/s`, same with `/ /`.

internal/locale/locale.go
MDrakos: This is still required to avoid a space on the newlines we purposely added to the yaml strings

internal/locale/locale.go
```diff
@@ -155,8 +156,13 @@ func Tt(translationID string, args ...interface{}) string {
 
 	// For templates we want to manually specify the linebreaks as the way YAML gets parsed makes
 	// this very painful otherwise
-	translation = strings.Replace(translation, "\n", "", -1)
+
+	// Replace newlines in yaml strings with space to avoid concatenated words
+	replaceRegex := regexp.MustCompile(`\s*\n`)
+	translation = replaceRegex.ReplaceAllString(translation, " $1")
```
Naatan, line 162:
```suggestion
	translation = replaceRegex.ReplaceAllString(translation, " ")
```

There is no group

MDrakos: Whoops

PR #903, opened 2020-08-24
Write compatiblity warning for windows < 10

mdrohmann:
https://www.pivotaltracker.com/story/show/174268514

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -49,8 +49,7 @@ func main() {
 		osv, err := sysinfo.OSVersion()
 		if err != nil {
 			logging.Debug("Could not retrieve os version info: %v", err)
-		}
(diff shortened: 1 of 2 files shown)


cmd/state/main.go
```diff
@@ -43,6 +45,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	if runtime.GOOS == "windows" {
+		osv, err := sysinfo.OSVersion()
+		if err != nil {
+			logging.Debug("Could not retrieve os version info: %v", err)
+		}
+		if osv.Major < 10 {
```
Naatan, line 53: This will be a nil pointer panic if it errored above.

mdrohmann: Thanks for catching.

PR #904, opened 2020-08-24
Add custom dimension for GA events

mdrohmann:
https://www.pivotaltracker.com/story/show/174456774


installers/msi-language/ActiveState/Tracker.cs
```diff
@@ -21,15 +20,21 @@ public sealed class TrackerSingleton
...
             this._tracker = new SimpleTracker(GoogleAnalyticsUserAgent, simpleTrackerEnvironment);
             this._cid = GetInfo.GetUniqueId();
-  		}
+        }
 
-        public async Task<TrackingResult> TrackEventAsync(string category, string action, string label, long value = 1)
+        private string modifyVersionString(string versionProperty)
+	{
+            var versionParts = versionProperty.Split('.');
+            return String.Format("{0}.{1}-CE", versionParts[0], versionParts[1]);
```
Naatan, line 34: Let's just send the version property we have, the value doesn't have to be exactly as in the ACs, long as we know it's 5.26 or 5.28 that's fine.

Then this will also be compatible with future iterations that might not be CE's.

mdrohmann:
It seems to work: 
Selection_023

PR #909, opened 2020-08-25
Ensure state cmd avialable as alias if not on path

daved:
https://www.pivotaltracker.com/story/show/174389657


internal/subshell/sscommon/rcfile.go
```diff
@@ -199,3 +211,22 @@ func SetupProjectRcFile(templateName, ext string, env map[string]string) (*os.Fi
...
+
+func inPath(pathVal, dir string) bool {
+	paths := strings.Split(pathVal, string(os.PathListSeparator))
+	for _, path := range paths {
+		if path == dir {
+			return true
+		}
+	}
+	return false
+}
```
Naatan, line 232: I think you can dump both of these and use `fileutils.PathContainsParent` instead.

test/integration/activate_int_test.go
```diff
@@ -51,6 +51,23 @@ func (suite *ActivateIntegrationTestSuite) TestActivateWithoutRuntime() {
...
+	ts := e2e.NewNoPathUpdate(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.Spawn("activate", "ActiveState-CLI/Python3")
+	cp.Expect("Where would you like to checkout")
+	cp.SendLine(cp.WorkDirectory())
+	cp.Expect("activated state", 20*time.Second)
+	cp.WaitForInput(10 * time.Second)
+
+	cp.SendLine("alias state")
```
Naatan, line 64: This won't pass on Windows. Please run integration tests on this PR (add the label).

internal/subshell/sscommon/rcfile.go
daved: PathContainsParent checks if the provided path is at all matching or a child of the provided parent path. It doesn't deal with the PATH envvar.

test/integration/activate_int_test.go
daved: Sorry, I was still working out the doskey args. It helped me fix (hopefully) a bug where args were not passed to the aliased command.

internal/subshell/sscommon/rcfile.go
Naatan: Nor does inPath.. ? You'd have to iterate over PATH and then call `PathContainsParent `.

daved: I'm mostly sure that that would return false positives.

daved: I've renamed the function and variables to make the logic more clear. Manual testing does confirm the behavior I expect, and the added integration test should be working once I fully fix the doskey usage.

assets/shells/config.bat
```diff
@@ -13,6 +13,10 @@ set {{$K}}={{$V}}
 DOSKEY {{$K}}=state run {{$CMD}}
 {{end}}
 
+{{ if .ExecAlias }}
+DOSKEY state="{{.ExecAlias}}" $*
```
daved, line 17: @Naatan My fix here makes me think that the doskey usage above for scripts is broken. If it looks so to you as well, I'll file a bug.

Naatan: This is probably unique to DOSKEY right? I think alias does this implicitly. Feel free to address in this PR since it's such a small change.

internal/subshell/sscommon/rcfile.go
Naatan: @daved this isn't addressed

assets/shells/config.bat
daved: I successfully tested this manually as compared to what's on master.

internal/subshell/sscommon/rcfile.go
daved: Discussed in chat; Closing.

daved:
@Naatan Sorry for the additional delay. Because doskey macros cannot call other doskey macros, I had to use the full path to the exec for both the state macro, and the run macros.

Also, I added the ExecName key/val to the template map because I needed to use the command name (i.e. state) as a value enough times in the .bat file that I felt it made sense to have it be a variable rather than a magic number. Please let me know if you'd rather not have the var in use.

assets/shells/config.bat
```diff
@@ -9,14 +9,17 @@ set {{$K}}={{$V}}
 {{- end}}
 {{- end}}
 
-{{range $K, $CMD := .Scripts}}
-DOSKEY {{$K}}={{.Exec}} run "{{$CMD}}" $*
-{{end}}
+{{$execCmd := .ExecName}}
```
Naatan, line 12: I don't understand the point of this variable, It's never used as .ExecName, it will always be set to .ExecAlias on line 15? Seems you can just drop this and use .ExecAlias.

daved: The cmd string used should be "state" unless an alias is used. In that case, the absolute path is used.

PR #917, opened 2020-08-25
Remove works from MSI menu

mdrohmann:
https://www.pivotaltracker.com/story/show/174482706

### DIFF
--- a/installers/msi-language/Product.p.wxs
+++ b/installers/msi-language/Product.p.wxs
@@ -68,23 +68,23 @@
       </Component>
     </Feature>
 
-    <Property Id="REMEMBER">
+    <Property Id="REMEMBER" Secure="yes">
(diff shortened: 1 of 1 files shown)


mdrohmann:
At least on Windows 8.1, the properties that I now tagged as secure, were previously ignored with the log message: `Ignoring disallowed property XYZ`.  This has something to do with the client/server architecture of MSIs: only secure properties are allowed to be moved to the server.

I cannot say that I understand it, but this fixes the problem.

mdrohmann:
That's taken care of. :)

Naatan:
Did you verify that this fixes the issue on Windows 10? That's where I got the issue.

mdrohmann:
I could only reproduce it on Windows 8.1. On Windows 10 it always worked. @Naatan Can you try on your machine once this is merged?

PR #919, opened 2020-08-26
Fix google cloud library error

MDrakos:
https://www.pivotaltracker.com/story/show/174501822

There was a small update to the code in the google cloud library that may help is with this issue, so I've updated the library. Also added a recover to the function that uses that library


internal/ci/gcloud/secrets.go
```diff
@@ -18,6 +18,14 @@ func (e ErrNotAvailable) Error() string { return "" }
 
 // GetSecret accesses the payload for the given secret
 func GetSecret(name string) (string, error) {
+	// We have run into instances where secretmanager.NewClient can panic
+	// when it tries to determine if it is running on Google Compute Engine
+	defer func() {
+		if r := recover(); r != nil {
+			logging.Error("Recovered from panic attempting to get gcloud secret: %v", r)
```
Naatan, line 25: You need to name the return parameters and define them in your recovery function cause otherwise you don't know what it could be returning. Could easily cause another panic.

PR #920, opened 2020-08-26
MSI logging sent to rollbar

mdrohmann:
https://www.pivotaltracker.com/story/show/174501900


mdrohmann:
I tested that everything is still working on my Windows 10 machine.  And after forcing an error, the captured logs were printed back in the log file created by msiexec.  I did not verify yet, that they were actually sent to rollbar.

A note on the implementation:
All log files are written to a file in a temporary directory right now.  This way they can be shared between the custom actions.  That seems to work, but I would like to re-visit this in a follow-up story that I will file soon.

PR #923, opened 2020-08-26
Add retries to graphql calls

daved:
https://www.pivotaltracker.com/story/show/174515303


daved:
The second two AC items were moved here: https://www.pivotaltracker.com/n/projects/2203557/stories/174519013

The current setup uses the global default HTTP client within the net/http library. Other uses of retryablehttp will need to follow this setup in order to ensure that the retryable HTTP client has httpmock endpoints registered to it.


Extra info (feel free to ignore):
A more structured solution would be nice, but far too costly for the benefit. IIRC, we are likely moving away from faked HTTP endpoints, and (I'm mostly certain) there is no real concern for race conditions created by the current setup.

PR #925, opened 2020-08-27
Check for clean installation directory

mdrohmann:
https://www.pivotaltracker.com/story/show/174520337

### DIFF
--- a/internal/runners/deploy/deploy.go
+++ b/internal/runners/deploy/deploy.go
@@ -186,7 +186,7 @@ func ensurePathIsClean(path string) error {
 	}
 	empty, fail := fileutils.IsEmptyDir(path)
 	if fail != nil {
-		return errs.Wrap(fail, "Failed to check if {{.V0}} is empty.", path)
+		return errs.Wrap(fail, "Failed to check if %s is empty.", path)
 	}
 	if !empty {
 		return locale.NewError("deploy_install_path_not_empty", "The installation directory {{.V0}} needs to be empty.", path)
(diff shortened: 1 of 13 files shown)


Naatan (dismissed):

mdrohmann:
@Naatan I saw that you already approved this PR:  I was still adding the same functionality to the `state deploy` command.  With that change it would also fail early if the installation directory is not clean.  I think it is safer to do, as long as we try to remove the directory contents before we install.  If a user (by accident or deliberately) selects a directory like `~/.local/bin`, they probably would be sad, when they realize that the folder had been wiped after the installation...

You probably also noticed that I added an `.editorconfig` file and re-formatted a bunch of XML files.  So, I suggest to ignore whitespace changes when viewing this PR.  My hope is that these change will get rid of the formatting problems in future PRs...

internal/runners/deploy/deploy.go
```diff
@@ -177,7 +177,29 @@ func runStepsWithFuncs(targetPath string, force, userScope bool, namespace proje
...
+		return locale.NewError("deploy_install_path_not_empty", "The installation directory {{.V0}} needs to be empty.", path)
+	}
+	return nil
+}
+
 func install(path string, installer installable, out output.Outputer) (runtime.EnvGetter, error) {
+	// check that path is empty or does not exist yet
+	err := ensurePathIsClean(path)
+	if err != nil {
+		return nil, err
```
Naatan, line 201: Wrap with locale error, never just forward errors.

PR #929, opened 2020-08-27
Use file modification time as source of truth for update check

Naatan:
https://www.pivotaltracker.com/story/show/174520822

### DIFF
--- a/assets/ppm/ppm.bat
+++ b/assets/ppm/ppm.bat
@@ -1,3 +1,3 @@
 @echo off
 
-state.exe _ppm %*
+{{.exe}} _ppm %*
--- a/assets/ppm/ppm.sh
+++ b/assets/ppm/ppm.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-state _ppm $*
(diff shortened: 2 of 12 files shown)


mdrohmann (dismissed):
Now, I actually understand the question that you asked me yesterday on slack.  I think that your GitHub link was a couple lines off.  But yes, I think it is okay not to copy the executable, as there are only a couple of cases where the executable may be modified, and these are the update tests.  :)

We just need to be really careful, as previously when we shared the `state` executable between *all* tests on Windows, we ran into `AccessDenied` errors, and on all other OSes, tests failed or passed incorrectly, because the state tool executable under test, was suddenly updated.  That was difficult to debug...  But I am confident that the update tests are the only ones that are actually tampering with an executable.

I wonder though if it was better to rename the `UseDistinctExe` function to `UseDistinctStateTool` or something like this, because the State Tool executable is really the only one that gets copied here.  That was different from using the option `ReUseExecutables` where all executables (even those specified with `SpawnWithCmd` where affected.

PR #932, opened 2020-08-28
Add http retry to inventory and headchef calls (and centralize retry setup)

daved:
https://www.pivotaltracker.com/story/show/174520426


daved:
The inventory and headchef calls listed were covered. To give us a quick view of other possible high-value targets, a "best effort" list of calls is available here: https://docs.google.com/spreadsheets/d/1zr76Aq6zw9I7hpyTGZ4SK7xP2H2KXWju5QElV5Jm33Q/edit#gid=561405140

pkg/platform/apicalls
```diff
@@ -0,0 +1,61 @@
+#! /usr/bin/env bash
```
Naatan, line 1: Accidental commit?

daved: I found the usefulness of the script to be worth seeing if you'd be willing to keep it around. I can store it locally in my own dev environment if not.

Naatan: It would have to go into the `scripts` folder, but I think this is too situational. Don't think it makes sense to commit it.

PR #933, opened 2020-08-28
Infer the language instead of requiring the user to input it

Naatan:
https://www.pivotaltracker.com/story/show/173569440

### DIFF
--- a/internal/runners/packages/list.go
+++ b/internal/runners/packages/list.go
@@ -62,7 +62,7 @@ func (l *List) Run(params ListRunParams) error {
 		return fail.WithDescription("package_err_cannot_fetch_checkpoint")
 	}
 
-	table := newFilteredRequirementsTable(checkpoint, params.Name)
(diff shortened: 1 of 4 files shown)


internal/runners/packages/list.go
```diff
@@ -139,7 +139,7 @@ func fetchCheckpoint(commit *strfmt.UUID) (model.Checkpoint, *failures.Failure)
 		return nil, model.FailNoData.New(locale.T("package_no_data"))
 	}
 
-	return model.FilterCheckpointPackages(checkpoint), fail
+	return checkpoint, fail
```
MDrakos, line 142: This change doesn't seem to be related to importing packages. This will also include requirements that are not packages in the output of `state packages`

Naatan: It's related since otherwise I cannot filter out the language. I'll address the second point.

PR #938, opened 2020-08-31
Avoid rollbar logs for already updating tool

daved:
https://www.pivotaltracker.com/story/show/174558582

### DIFF
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/url"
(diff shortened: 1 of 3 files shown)


daved:
From a debug message within the tests:
```
file "/tmp/090935712/locked-keep" is already locked: cannot acquire lock: pid 20999 exists
file "/tmp/090935712/locked-remove" is already locked: cannot acquire exclusive lock| failed to lock file| resource temporarily unavailable
```

internal/updater/updater.go
```diff
@@ -188,6 +189,11 @@ func (u *Updater) update(out output.Outputer) error {
 	// same new state tool version several times.
 	_, err = pl.TryLock()
 	if err != nil {
+		if inProgErr := new(*osutils.AlreadyLockedError); errors.As(err, inProgErr) {
+			logging.Debug("Already updating: %s", errs.Join(*inProgErr, ": "))
```
mdrohmann, line 193: 👍

PR #940, opened 2020-08-31
Enhance rollbar logging 174555634

MDrakos:
https://www.pivotaltracker.com/story/show/174555634

A lot of the changes here are reverting the logging calls. Important changes are in `Product.p.wxs`, `Rollbar.cs`, and `Logging.cs`

An example report is here: https://rollbar.com/activestate/msi/items/341/occurrences/134369786239/. The logging is very verbose, we may want to reduce some of the logging if we find it too noisy.


mdrohmann (dismissed):
This all makes sense to me.   It will just require another merge with master.  I am glad that we can use the integrated logging mechanism...

PR #942, opened 2020-09-02
Update msi rollbar version format

daved:
https://www.pivotaltracker.com/story/show/174587605


installers/msi-language/StateCommandExists/CustomAction.cs
```diff
@@ -9,7 +9,7 @@ public class CustomActions
         [CustomAction]
         public static ActionResult CheckCommands(Session session)
         {
-            ActiveState.RollbarHelper.ConfigureRollbarSingleton(session["COMMIT_ID"]);
+            ActiveState.RollbarHelper.ConfigureRollbarSingleton(session["MSI_VERSION"]);
```
daved, line 12: I'm calling this as the existing code does, but it doesn't match `session.CustomActionData` in other files. Does this need to be different?

scripts/msi-installer-builder/main.go
```diff
@@ -114,6 +116,11 @@ func normalize(preset languagePreset, c *config) (*config, error) {
 	if err != nil {
 		return c, err
 	}
+
+	if c.MSIVersion == "" {
+		c.MSIVersion = msiVersionInfo()
+	}
```
Naatan, line 122: This feels dirty and pointless, we are always setting it to `msiVersionInfo()` one way or another.

Naatan (changes requested):

scripts/msi-installer-builder/main.go
daved: The apparent duplication seems less dirty, to me, than the function not doing it's job based on the special knowledge that we happen to set the field in current cases.

Naatan: So error out if it's empty. This is too dirty.

installers/msi-language/StateCommandExists/CustomAction.cs
daved: @MDrakos or @mdrohmann Would you, please, shed light on why this call is different than other session usages?

MDrakos: This custom action is scheduled as `immediate` so it has access to the session object. The custom actions that are deferred do not, so we must pass the data it needs via the `CustomActionData`. This is all done in the `Product.p.wxs` file.

Naatan (dismissed):

Naatan (approved):

PR #944, opened 2020-09-03
Send user network finish event

MDrakos:
https://www.pivotaltracker.com/story/show/174606570

Addresses AC:
 - We send the finish event with the value: "user_network"

### DIFF
--- a/installers/msi-language/StateDeploy/CustomAction.cs
+++ b/installers/msi-language/StateDeploy/CustomAction.cs
@@ -531,7 +531,7 @@ namespace StateDeploy
         [CustomAction]
         public static ActionResult GAReportFailure(Session session)
         {
-            session.Log("sending event about starting the MSI");
+            session.Log("sending event about MSI failure");
             TrackerSingleton.Instance.TrackEventSynchronously(session, session["SESSION_ID"], "stage", "finished", "failure", session["ProductVersion"]);
             return ActionResult.Success;
         }
@@ -539,7 +539,7 @@ namespace StateDeploy
         [CustomAction]
         public static ActionResult GAReportSuccess(Session session)
         {
-            session.Log("sending event about starting the MSI");
+            session.Log("sending event about MSI success");
             TrackerSingleton.Instance.TrackEventSynchronously(session, session["SESSION_ID"], "stage", "finished", "success", session["ProductVersion"]);
             return ActionResult.Success;
         }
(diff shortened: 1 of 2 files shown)


installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -567,6 +567,17 @@ public static ActionResult GAReportUserExit(Session session)
...
         }
 
+        /// <summary>
+        /// Reports a user network error event to google analytics
+        /// </summary>
+        [CustomAction]
+        public static ActionResult GAReportUserNetwork(Session session)
+        {
+            session.Log("sending user network error event");
+            TrackerSingleton.Instance.TrackEventSynchronously(session, session["SESSION_ID"], "stage", "finished", "network", session["ProductVersion"]);
```
Naatan, line 577: the AC says `user_network`.

Naatan (changes requested):

Naatan (approved):
I was just reviewing the AC, would still want Martin's eyes on this before we merge.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -616,6 +627,7 @@ public static ActionResult SetNetworkErrorProperties(Session session)
                 RollbarReport.Error(msg, session);
             }
 
+            session.DoAction("GAReportUserNetwork");
```
mdrohmann, line 630: Interesting!  So, you can run "immediate" actions from a "deferred" action, and then use the `session` variables.  Maybe we can use to reduce the list of variables that we set in the `CustomActionData`?

My concern with this is, that with this set-up both the `stage/finished/network` and the `stage/finished/failure` messages will be sent. @Naatan Is that ok?

mdrohmann (changes requested):
I believe that in case of a network error, we are still sending the `finished/failure` event.  That may be okay.  But the cleaner solution (although of course very difficult one, might be to change the `GAReportFailure` CustomAction and check if we have to send a failure or a network event...

installers/msi-language/StateDeploy/CustomAction.cs
MDrakos: This method is actually run as immediate. The flow is a little different, it's declared [here](https://github.com/ActiveState/cli/blob/master/installers/msi-language/Product.p.wxs#L224) and eventually called via the Custom fatal error dialog [here](https://github.com/ActiveState/cli/blob/master/installers/msi-language/CustomFatalError.wxs#L37). I had to do it this way as you can only schedule one action to occur with `OnExit="error"`

mdrohmann: Ah! Maybe I get it know.  So, `OnExit=error` we are just invoking the `SetNetworkErrorProperties` custom action which then is calling the `CustomFatalError` dialog.

However, the `CustomFatalError` dialog is still calling `GAReportFailure` when the user clicks the `"Finish"` button https://github.com/ActiveState/cli/blob/302fe08257775318e0602bd5348bb1b8024df085/installers/msi-language/Product.p.wxs#L130.

But don't we then invoke `GAReportUserNetwork` every time, even if `NETWORK_ERROR` is empty?

MDrakos: Yes you are right about the `GaReportUserNetwork`. That should be now fixed.

I've decoupled the `GAReportFailure` from the Finish button on the exit dialog and instead it is now triggered from `SetNetworkProperties` if there is no network error.

mdrohmann (approved):

PR #947, opened 2020-09-03
Allow user-defined commit id in activate checkout and deploy

daved:
https://www.pivotaltracker.com/story/show/174611979


locale/en-us.yaml
```diff
@@ -901,7 +901,7 @@ err_activate_namespace:
 err_activate_create_project:
   other: Could not create new project
 err_invalid_namespace:
-  other: "Invalid namespace: {{.V0}}. Should be in the format of ORGANISATION_NAME/PROJECT_NAME. Names should be alphanumeric and may contain dashes and periods."
+  other: "Invalid namespace: {{.V0}}. Should be in the format of ORGANISATION_NAME/PROJECT_NAME. This may include a suffix in the format of {namespace}#COMMIT_ID. Names and commit ID should be alphanumeric and may contain dashes and periods."
```
Naatan, line 904:
```suggestion
  other: "Invalid namespace: {{.V0}}. Should be in the format of ORGANISATION_NAME/PROJECT_NAME, and can optionally contain a #COMMIT_ID suffix. Names should be alphanumeric and may contain dashes and periods."
```

pkg/project/namespace.go
```diff
@@ -43,13 +45,18 @@ func (ns *Namespaced) String() string {
...
+	var sep, commitSep, commitID string
 	if ns.IsValid() {
 		sep = "/"
+		if ns.CommitID != nil {
+			commitSep = "#"
+			commitID = ns.CommitID.String()
+		}
 	}
-	return fmt.Sprintf("%s%s%s", ns.Owner, sep, ns.Project)
+	return fmt.Sprintf("%s%s%s%s%s", ns.Owner, sep, ns.Project, commitSep, commitID)
```
Naatan, line 56: This is outside of our AC's. It would affect a lot of cases that we're not looking to address with this story. eg. errors that mention the namespace (we don't want to show the commitID there).

Naatan (changes requested):

pkg/project/namespace.go
daved: Yeah, I'm thinking it makes most sense to have a specific type wrapping namespaced to add this behavior if it becomes a more used feature. For now, though, I'll make sure namespaced behaves pretty much exactly as it does before.

daved:
@Naatan If this is looking OK, it's likely easy enough to be worth adding an integration test. At least for activate, but maybe for deploy as well.

.github/workflows/build.yml
```diff
@@ -105,12 +105,7 @@ jobs:
     - name: Sign
       shell: bash
       if: runner.os == 'Windows'
-      run: "echo $MSI_CERT_BASE64 | base64 --decode > Cert.p12\nexport PATH=/c/Program\\
-        Files\\ \\(x86\\)/WiX\\ Toolset\\ v3.11/bin/:/c/Program\\ Files\\ \\(x86\\)/Windows\\
-        Kits/10/bin/10.0.16299.0/x86/:$PATH\n\nsigntool.exe sign -d \"ActiveState
-        State Tool\" -f \"Cert.p12\" -p ${CODE_SIGNING_PASSWD} ./build/state.exe \nsigntool.exe
-        sign -d \"ActiveState State Tool (32bit)\" -f \"Cert.p12\" -p ${CODE_SIGNING_PASSWD}
-        ./build/state32.exe \n"
+      run: "echo $MSI_CERT_BASE64 | base64 --decode > Cert.p12\nexport PATH=/c/Program\\ Files\\ \\(x86\\)/WiX\\ Toolset\\ v3.11/bin/:/c/Program\\ Files\\ \\(x86\\)/Windows\\ Kits/10/bin/10.0.16299.0/x86/:$PATH\n\nsigntool.exe sign -d \"ActiveState State Tool\" -f \"Cert.p12\" -p ${CODE_SIGNING_PASSWD} ./build/state.exe \nsigntool.exe sign -d \"ActiveState State Tool (32bit)\" -f \"Cert.p12\" -p ${CODE_SIGNING_PASSWD} ./build/state32.exe \n"
```
Naatan, line 108: Please undo, unrelated (check what you're committing)

Naatan (dismissed):

.github/workflows/build.yml
daved: I was aware, but since it's generated, I thought it was beneficial. Should I still remove the generated change?

Naatan: Yes

Naatan (approved):

PR #948, opened 2020-09-03
Handle SessionID edge cases

mdrohmann:
https://www.pivotaltracker.com/story/show/174603871

These are the edge cases:

- if UI=None or UI=basic, no session id is set in the beginning.  So, I set them at the start of the installation and save the session id in a registry key, so it can be retrieved for the `stage/final/*` events.
- if user cancels or installation fails right at the beginning before the session id is set (and the `stage/started` event is tracked), both has to be done at the end.

If there are still cases with `sessionID=unset` they should be reported as rollbar errors.


Naatan (changes requested):
Can we centralize this session ID logic into a single function and just always invoke it from the start of a custom action. This is getting way too messy to maintain, I now count 3 different ways that the session ID can be defined.

It's fine that there are 3 triggers for defining the session ID, but the logic for defining/getting the session ID should just be one single centralized function.

mdrohmann:
I centralized it, as much as I could.  

And I also realized that - as we now rely on saving temporary data in the registry anyways - that we could use this registry trick for other things (like discovering an existing state tool, and even setting the installation directory from a registry path).  
That would 
- shorten our parameter list for the [StateDeploy custom action](https://github.com/ActiveState/cli/blob/62cd9e84d4d9c99fd99d47dffffaf3a2740882f3/installers/msi-language/Product.p.wxs#L200) a little, and 
- make the MSI more resilient for non-ui installations, as it means to centralize more logic to the actual installation sequence (which should be executed for every sensible installation).  Which in the end also simplifies the logic.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -517,43 +542,99 @@ private static string BuildDeployCmd(Session session, string subCommand)
...
         {
             session.Log("sending event about MSI failure");
-            TrackerSingleton.Instance.TrackEventSynchronously(session, session["SESSION_ID"], "stage", "finished", "failure", session["ProductVersion"]);
+            var sessionID = GetSessionIDForExitAction(session);
+            if (sessionID == "unset")
+            {
+                // this can happen if an installation error happens before we could initialize the session id and send the start event
+
+                // So, we create a new session id,
+                sessionID = NewSessionID(session, session["ProductVersion"]);
```
MDrakos, line 571: It appears that `NewSessionID` now sends the started event but here it's being called from `GAReportFailure`. It's also called from `SetSessionID` which is scheduled in the `InstallUISequence` so I think this could lead to double reporting of the started event. 

It may be better to separate the two and specifically schedule both the `NewSessionID` and `SendStartedEvent` in the to be more explicit

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -563,7 +644,13 @@ public static ActionResult GAReportStart(Session session)
         public static ActionResult GAReportUserExit(Session session)
         {
             session.Log("sending user exit event");
-            TrackerSingleton.Instance.TrackEventSynchronously(session, session["SESSION_ID"], "stage", "finished", "cancelled", session["ProductVersion"]);
+            var sessionID = GetSessionIDForExitAction(session);
+            if (sessionID == "unset")
+            {
+                // This can happen, when the user cancelled on the Welcome Dialog, before the session id has been generated.
+                sessionID = NewSessionID(session, session["ProductVersion"]);
```
MDrakos, line 653: This is another instance where the started event may be reported twice

MDrakos (changes requested):

Naatan (changes requested):
As discussed on Slack, let's simplify this so we don't need to actually generate the sessionID

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -380,8 +380,33 @@ public InstallSequenceElement(string subCommand, string description)
         private static ActionResult run(Session session)
         {
             var sessionID = session.CustomActionData["SESSION_ID"];
+            var uiLevel = session.CustomActionData["UI_LEVEL"];
             var productVersion = session.CustomActionData["PRODUCT_VERSION"];
 
+            if (sessionID == "unset")
+            {
+                if (uiLevel != "2" /* no ui */ && uiLevel != "3" /* basic ui */)
```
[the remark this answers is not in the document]
mdrohmann: Yes, I guess that makes some of these changes obsolete.  Well, now it will be removed anyways.

But to answer your question.  UILEVEL==4, the `SetSessionID` immediate custom action was still executed.

installers/msi-language/StateDeploy/CustomAction.cs
mdrohmann: The idea was that when a new session id *must* be created, because it didn't exist before (was "unset"), that a start event cannot have been send yet.

But again that is obsolete now.

installers/msi-language/StateDeploy/CustomAction.cs
mdrohmann: Same as above...

Naatan (dismissed):

Naatan (approved):

MDrakos (approved):

PR #949, opened 2020-09-04
downgrade rollbar to 3.5.0

mdrohmann:
https://www.pivotaltracker.com/story/show/174664183


MDrakos (approved):
If you haven't already, we should trigger both an `Error` and `Critical` report just to ensure everything is still working as expected.

mdrohmann:
I just tested a critical error: https://rollbar.com/activestate/msi/items/364/  That is still working.

Will also trigger an `error`-error...

PR #951, opened 2020-09-04
Detect network issues from state commands

MDrakos:
https://www.pivotaltracker.com/story/show/174609354

### DIFF
--- a/installers/msi-language/ActiveState/Network.cs
+++ b/installers/msi-language/ActiveState/Network.cs
@@ -1,0 +1,69 @@
+﻿using System;
+using Microsoft.Win32;
+using Microsoft.Deployment.WindowsInstaller;
+
+
+namespace ActiveState
+{
+    public class Network
+    {
+        private const string networkErrorKey = "NetworkError";
+        private const string networkErrorMessageKey = "NetworkErrorMessage";
+
+        /// <summary>
+        /// ResetErrorDetails clears the registry entries for network errors.
+        /// This function must be run from a deferred custom action
+        /// </summary>
+        public static void ResetErrorDetails(Session session)
+        {
+            // Deferred custom actions run as administrator so we have to specifically set
+            // the registry key for the user using their SID in order for the value to
+            // be available in later immediate custom actions
+            string registryKey = string.Format("HKEY_USERS\\{0}\\SOFTWARE\\ActiveState\\{1}", session.CustomActionData["USERSID"], session.CustomActionData["PRODUCT_NAME"]);
+            RegistryValueKind registryEntryDataType = RegistryValueKind.String;
+            try
+            {
+                Registry.SetValue(registryKey, networkErrorKey, "false", registryEntryDataType);
+                Registry.SetValue(registryKey, networkErrorMessageKey, "", registryEntryDataType);
+            }
+            catch (Exception e)
+            {
+                string msg = string.Format("Could not delete network error registry keys. Exception: {0}", e.ToString());
+                session.Log(msg);
+                RollbarReport.Error(msg, session);
+            }
+        }
+
+        /// <summary>
+        /// SetErrorDetails writes the network error details to the user's registry.
+        /// This function must be run from a deferred custom action
+        /// </summary>
+        public static void SetErrorDetails(Session session, string msg)
+        {
+            SetErrorDetails(
+                session,
+                string.Format("HKEY_USERS\\{0}\\SOFTWARE\\ActiveState\\{1}", session.CustomActionData["USERSID"], session.CustomActionData["PRODUCT_NAME"]),
+                msg
+            );
+        }
+
+        /// <summary>
+        /// SetErrorDetails writes the network error details to the user's registry
+        /// </summary>
+        private static void SetErrorDetails(Session session, string registryKey, string msg)
+        {
+            RegistryValueKind registryEntryDataType = RegistryValueKind.String;
+            try
+            {
+                Registry.SetValue(registryKey, networkErrorKey, "true", registryEntryDataType);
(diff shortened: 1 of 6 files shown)


MDrakos:
This is an example of the output screen after these changes:

image

installers/msi-language/ActiveState/Network.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <summary>
+        /// ResetErrorDetails clears the registry entries for network errors.
+        /// This function must be run from a deferred custom action
+        /// </summary>
+        public static void ResetErrorDetails(Session session)
+        {
+            // Deferred custom actions run as administrator so we have to specifically set
+            // the registry key for the user using their SID in order for the value to
+            // be available in later immediate custom actions
+            string registryKey = string.Format("HKEY_USERS\\{0}\\SOFTWARE\\ActiveState\\{1}", session.CustomActionData["USERSID"], session.CustomActionData["PRODUCT_NAME"]);
```
Naatan, line 22: Is `USERSID` still valid given @mdrohmann's latest changes?

internal/retryhttp/client.go
```diff
@@ -71,19 +66,19 @@ func normalizeResponse(res *http.Response, err error) (*http.Response, error) {
 	if res != nil {
 		switch res.StatusCode {
 		case 408:
-			return res, locale.WrapInputError(&UserNetworkError{408}, "err_user_network_server_timeout", "Request failed due to timeout during communication with server. {{.V0}}", solutionLocale)
+			return res, locale.WrapInputError(&UserNetworkError{408}, "err_user_network_server_timeout", "Request failed due to timeout during communication with server.")
 		case 425:
-			return res, locale.WrapInputError(&UserNetworkError{425}, "err_user_network_tooearly", "Request failed due to retrying connection too fast. {{.V0}}", solutionLocale)
+			return res, locale.WrapInputError(&UserNetworkError{425}, "err_user_network_tooearly", "Request failed due to retrying connection too fast.")
 		case 429:
-			return res, locale.WrapInputError(&UserNetworkError{429}, "err_user_network_toomany", "Request failed due to too many requests. {{.V0}}", solutionLocale)
```
Naatan, line 78: These are here for the CLI use-case. Not the end of the world if we repeat this in the MSI.

Naatan (changes requested):

installers/msi-language/CustomFatalError.wxs
```diff
@@ -18,13 +18,13 @@
 				<Control Id="CustomTextFatalError" Type="Hyperlink" X="135" Y="110" Width="220" Height="20" Transparent="yes" Hidden="yes" Text="!(loc.FatalErrorPrimary)">
 					<Condition Action="show">NETWORK_ERROR = "false"</Condition>
 				</Control>
-				<Control Id="CustomTextNetworkError" Type="Text" X="135" Y="70" Width="220" Height="60" Transparent="yes" Hidden="yes" Text="!(loc.NetworkErrorPrimary)">
+				<Control Id="CustomTextNetworkError" Type="Text" X="135" Y="60" Width="220" Height="60" Transparent="yes" Hidden="yes" Text="!(loc.NetworkErrorPrimary)">
 					<Condition Action="show">NETWORK_ERROR = "true"</Condition>
 				</Control>
-				<Control Id="CustomTextNetworkErrorDetails" Type="Text" X="135" Y="130" Width="220" Height="40" Transparent="yes" Hidden="yes" Text="!(loc.NetworkErrorDetails)">
+				<Control Id="CustomTextNetworkErrorDetails" Type="Text" X="135" Y="120" Width="220" Height="60" Transparent="yes" Hidden="yes" Text="!(loc.NetworkErrorDetails)">
```
mdrohmann, line 24: I wonder if we can actually use a [ScrollableText control](https://docs.microsoft.com/en-us/windows/win32/msi/scrollabletext-control) here, just like in the EULA dialog.  That way we would ensure that we always have enough space for the error text.

installers/msi-language/ActiveState/Network.cs
mdrohmann: I believe that `USERSID` would still be the right choice here.

mdrohmann (changes requested):
As we cannot control the length of the error message very well, I thought it would be nice to use a ScrollableText element for it.

installers/msi-language/ActiveState/Network.cs
MDrakos: Yes, the `USERSID` stuff is still required here so we can pass the network error details from the deferred custom action to an immediate one

installers/msi-language/CustomFatalError.wxs
MDrakos: I spent some time trying different alternatives. The issue with a `ScrollableText` element is that it requires an `.rft` file at build time to populate the text box. Apparently it is possible to open the MSI database at runtime and replace that file, however after some investigation I did not go down that route. Come to think of it I don't think a deferred custom action can access the MSI database as it has to get access via the `session` object.

internal/retryhttp/client.go
Naatan: That's fine, this'll give them the information they need.

Naatan (dismissed):

installers/msi-language/CustomFatalError.wxs
mdrohmann: I actually also remember looking into that for some other dialog.  Well, another of those strange limitations.  Let's leave it like this.  There is quite a bit of space after all...

mdrohmann (dismissed):

mdrohmann (approved):

Naatan (approved):

PR #952, opened 2020-09-04
Catch and handle rollbar exceptions

mdrohmann:
https://www.pivotaltracker.com/story/show/174675582

### DIFF
--- a/installers/msi-language/ActiveState/Rollbar.cs
+++ b/installers/msi-language/ActiveState/Rollbar.cs
@@ -118,13 +118,20 @@ public class RollbarReport
 	    
             if (!criticalReported)
             {
-                if (level == Level.Critical)
+                try
                 {
-                    criticalReported = true;
-                    RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Critical(new GenericException(message), customFields);
-                } else
+                    if (level == Level.Critical)
+                    {
+                        criticalReported = true;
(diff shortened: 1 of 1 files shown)


installers/msi-language/ActiveState/Rollbar.cs
```diff
@@ -118,13 +118,20 @@ private static void Report(Level level, string message, Session session, IDictio
...
+                        RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Critical(new GenericException(message), customFields);
+                    }
+                    else
+                    {
+                        RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Error(new GenericException(message), customFields);
+                    }
+                } catch (System.Exception e)
                 {
-                    RollbarLocator.RollbarInstance.AsBlockingLogger(RollbarTimeout).Error(new GenericException(message), customFields);
+                    session.Log("Logging to rollbar failed with error: {0}", e);
```
MDrakos, line 134: Would it be worthwhile trying to send an S3/GA Event in this case? That we we can potentially get some information when this is happening.

MDrakos (changes requested):

installers/msi-language/ActiveState/Rollbar.cs
mdrohmann: That is a good idea.  @Naatan can I just decide on the event name?  I would choose category "error", label "rollbar".

Naatan: Sounds good

Naatan (approved):

MDrakos (approved):

PR #959, opened 2020-09-09
Retry download/get on body read errors

daved:
https://www.pivotaltracker.com/story/show/174678449


daved:
@Naatan This is not working. I've hit an issue with error/failure propagation. Before I spend more time, I'd appreciate having your direction.

daved:
Closed in favor of #963

daved:
Reopening after #963 was reverted by #973

daved:
Reopening after #963 was reverted by #973

internal/download/get.go
```diff
@@ -39,66 +40,80 @@ func SetMocking(useMocking bool) {
...
+	fn := func() error {
+		logging.Debug("Retrieving url: %s", url)
+		client := retryhttp.NewClientFromExisting(retryhttp.DefaultClient, 3)
+		resp, err := client.Get(url)
+		if err != nil {
+			code := -1
+			if resp != nil {
+				code = resp.StatusCode
+			}
+			fail := failures.FailNetwork.Wrap(err, locale.Tl("err_network_get", "Status code: {{.V0}}", strconv.Itoa(code)))
```
Naatan, line 59: You changed the function to return errors instead of failures, so return errors.

internal/download/get.go
```diff
@@ -39,66 +40,80 @@ func SetMocking(useMocking bool) {
...
+			return &retryfn.ControlError{
+				Cause: fail,
+				Type:  retryfn.Halt,
+			}
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != 200 {
+			fail := failures.FailNetwork.New("err_invalid_status_code", strconv.Itoa(resp.StatusCode))
```
Naatan, line 68: Use errors

internal/download/get.go
```diff
@@ -39,66 +40,80 @@ func SetMocking(useMocking bool) {
...
-	}
+		var total int
+		length := resp.Header.Get("Content-Length")
+		if length == "" {
+			total = 1
+		} else {
+			total, err = strconv.Atoi(length)
+			if err != nil {
+				logging.Debug("Content-length: %v", length)
+				return failures.FailInput.Wrap(err)
```
Naatan, line 83: Should now be an error

internal/download/get.go
```diff
@@ -39,66 +40,80 @@ func SetMocking(useMocking bool) {
...
-		total, err = strconv.Atoi(length)
+		bar := progress.AddByteProgressBar(int64(total))
+		src := bar.ProxyReader(resp.Body)
+		var dst bytes.Buffer
+
+		_, err = io.Copy(&dst, src)
 		if err != nil {
-			logging.Debug("Content-length: %v", length)
-			return nil, failures.FailInput.Wrap(err)
+			return failures.FailInput.Wrap(err)
```
Naatan, line 93: Error

internal/download/get.go
```diff
@@ -39,66 +40,80 @@ func SetMocking(useMocking bool) {
...
 }
 
-func _testHTTPGetWithProgress(url string, progress *progress.Progress) ([]byte, *failures.Failure) {
+func _testHTTPGetWithProgress(url string, progress *progress.Progress) ([]byte, error) {
 	return _testHTTPGet(url)
 }
 
 // _testHTTPGet is used when in tests, this cannot be in the test itself as that would limit it to only that one test
-func _testHTTPGet(url string) ([]byte, *failures.Failure) {
+func _testHTTPGet(url string) ([]byte, error) {
```
Naatan, line 116: This function is still returning a failure

internal/download/manager.go
```diff
@@ -4,18 +4,17 @@ import (
...
-	"github.com/ActiveState/cli/internal/failures"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/progress"
 )
 
 // Manager is our main download manager, it takes care of processing the downloads and communicating progress
 type Manager struct {
 	WorkerCount int
-	failure     *failures.Failure
+	err         error
```
Naatan, line 17: This was out of scope for your story. Since the work is done let's leave it in, but please scope yourself down next time.

internal/download/manager.go
```diff
@@ -55,32 +54,32 @@ func (m *Manager) Download() *failures.Failure {
...
+	m.err = fail.ToError()
+	if m.err != nil {
 		return
 	}
 
-	err := ioutil.WriteFile(entry.Path, bytes, 0666)
+	err = ioutil.WriteFile(entry.Path, bytes, 0666)
 	if err != nil {
-		m.failure = failures.FailIO.Wrap(err)
+		m.err = failures.FailIO.Wrap(err)
```
Naatan, line 82: This should be an error, not a failure

internal/retryfn/retryfn.go
```diff
@@ -0,0 +1,96 @@
...
+func (rf *RetryFn) Run() error {
+	if rf.fn == nil {
+		return nil
+	}
+
+	var err error
+	for i := rf.tries; i > 0; i-- {
+		rf.calls++
+
+		if err = rf.fn(); err != nil {
```
Naatan, line 73: Please invert your conditional to reduce the indent

internal/retryfn/retryfn.go
```diff
@@ -0,0 +1,96 @@
...
+			cerr := &ControlError{}
+			if errors.As(err, &cerr) {
```
Naatan, line 75: Invert

internal/download/get.go
```diff
@@ -39,66 +40,80 @@ func SetMocking(useMocking bool) {
...
+				Type:  retryfn.Halt,
+			}
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != 200 {
+			fail := failures.FailNetwork.New("err_invalid_status_code", strconv.Itoa(resp.StatusCode))
+			return &retryfn.ControlError{
+				Cause: fail,
+				Type:  retryfn.Halt,
```
Naatan, line 71: Wouldn't this cause it NOT to retry?

internal/retryhttp/client.go
```diff
@@ -124,3 +124,15 @@ func NewClient(timeout time.Duration, retries int) *Client {
 		Client: retryClient,
 	}
 }
+
+func NewClientFromExisting(c *Client, retries int) *Client {
```
Naatan, line 128: Should make `NewClient` use this to reduce the duplication of logic

Naatan (changes requested):

internal/download/manager.go
daved: It seemed to make sense at the time since I would have had to, otherwise, added failure wrapping to the errors returned by the download func.

internal/retryfn/retryfn.go
daved: Making this change made it apparent to me that I need to add a test and fix some behavior.

internal/download/get.go
daved: In the case of "statusCode != 200":
IIUC, if the flow reaches here, the server has responded with some kind of affirmative and retrying is unlikely to then produce a 200.

In the case of "client.Get()-err != nil":
I was trying to prevent trying so many times, but I think it makes sense now that we're setting max tries to 3x3, so that is updated.

internal/download/manager.go
Naatan: That would have been much simpler than refactoring all of this.

pkg/platform/runtime/installer.go
```diff
@@ -200,10 +200,10 @@ func (installer *Installer) InstallArtifacts(runtimeAssembler Assembler) (envGet
 		}
 
 		if len(downloadArtfs) > 0 {
-			archives, fail := installer.runtimeDownloader.Download(downloadArtfs, runtimeAssembler, progressBar)
-			if fail != nil {
+			archives, err := installer.runtimeDownloader.Download(downloadArtfs, runtimeAssembler, progressBar)
+			if err != nil {
 				progressBar.Cancel()
-				return nil, false, fail
+				return nil, false, failures.FailIO.Wrap(err)
```
Naatan, line 206: This should be reverted altogether, unrelated to your story/PR

Naatan (changes requested):

internal/retryfn/retryfn.go
```diff
@@ -70,20 +70,23 @@ func (rf *RetryFn) Run() error {
...
+		}
+
+		cerr := &ControlError{}
+		if !errors.As(err, &cerr) {
+			continue
+		}
+
+		switch cerr.Type {
+		case Unset, Unknown:
+			err = cerr.Cause
```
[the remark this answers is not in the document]
daved: Requiring retries to be called for explicitly seems contrary to the general purpose of a retry mechanism.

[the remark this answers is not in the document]
Naatan: True but your retry mechanism is wrapping a lot of logic, where a multitude of different errors can happen, not all of which warranting a retry. Should we go with this approach I would prefer a bit more intelligence around what cases justify a retry.

daved:
Closing in favor of #975

PR #961, opened 2020-09-09
Revert "state deploy checks for clean target directory"

Naatan:
This reverts commit 808dbb9b35ab7e1361ab548c55bc1d71d2bdf2bd.

# Conflicts:
#	internal/runners/deploy/deploy.go

https://www.pivotaltracker.com/story/show/174739175

### DIFF
--- a/internal/runners/deploy/deploy.go
+++ b/internal/runners/deploy/deploy.go
@@ -182,29 +182,7 @@ func runStepsWithFuncs(targetPath string, force, userScope bool, namespace proje
 
 type installFunc func(path string, installer installable, out output.Outputer) (runtime.EnvGetter, error)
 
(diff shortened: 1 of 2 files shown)


mdrohmann (approved):
I know that the main reason we have `state deploy` is the MSI, but do we have some monitoring if people are also using it without the MSI?  If there are a few of them, I would think we should probably re-visit this issue a little, because in several places we just plain delete everything: https://github.com/ActiveState/cli/blob/39329ebb7fc480c2bc245b905ea4088f25fbe3f1/pkg/platform/runtime/installer.go#L288

T

Naatan:
@mdrohmann filed a followup: https://www.pivotaltracker.com/story/show/174739618

PR #963, opened 2020-09-10
Download S3 urls using the S3 download manager

Naatan:
https://www.pivotaltracker.com/story/show/174721786


internal/download/get.go
```diff
@@ -31,81 +34,115 @@ func init() {
...
+		r.Region = "us-east-1"
+	} else if domain[1] == "s3" && domain[3] == "amazonaws" { // https://bucket-name.s3.Region.amazonaws.com/key-name
+		r.Bucket = domain[0]
+		r.Region = domain[2]
+	} else { // https://s3.Region.amazonaws.com/bucket-name/key-name
+		r.Bucket = strings.SplitN(url.Path, "/", 1)[0]
+		r.Region = domain[1]
+	}
+	return r
+}
```
daved, line 131: strong gritty nit:  The index bounds should be checked before access and error conditions should be returned. The improved error message content would provide a trail to follow in the (yes, unlikely) case an incompatible URL is passed in.

internal/download/get.go
```diff
@@ -31,81 +34,115 @@ func init() {
...
-		total, err = strconv.Atoi(length)
+		lengthInt, err := strconv.Atoi(res.Header.Get("Content-Length"))
 		if err != nil {
-			logging.Debug("Content-length: %v", length)
-			return nil, failures.FailInput.Wrap(err)
+			logging.Debug("Could not grab content-length: %v", err)
+		} else {
+			length = int64(lengthInt)
 		}
 	}
```
daved, line 71: Is it OK that the length is potentially 0? I think a basic default value (128?) would at least give some indication of action despite being wrong. Otherwise, 0 will show the bar at 100% and will be wrong, but it will show no action at all.

internal/download/get.go
```diff
@@ -31,81 +34,115 @@ func init() {
...
 	if err != nil {
-		return nil, failures.FailInput.Wrap(err)
+		return nil, locale.WrapError(err, "err_dl_s3", "Downloading failed due to underlying S3 error: {{.V0}}.", err.Error())
 	}
 
-	if !bar.Completed() {
-		// Failsafe, so we don't get blocked by a progressbar
-		bar.IncrBy(total)
-	}
+	bar.Abort(true) // ensure we don't get stuck on an incomplete bar
```
daved, line 106: note: This will interrupt rather than update to 100%.

daved (changes requested):

internal/download/manager.go
```diff
@@ -78,7 +87,7 @@ func (m *Manager) Job(entry *Entry) {
 		return
 	}
 
-	err := ioutil.WriteFile(entry.Path, bytes, 0666)
+	err = ioutil.WriteFile(entry.Path, bytes, 0666)
```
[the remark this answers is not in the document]
Naatan: This didn't actually change as part of this PR, and I don't want to risk breaking anything this PR isn't targeting.

internal/download/get.go
```diff
@@ -31,81 +34,115 @@ func init() {
...
 
-	src = bar.ProxyReader(resp.Body)
+	// Record progress
+	bar := progress.AddByteProgressBar(length)
+	cb := func(length int) {
+		if !bar.Completed() {
+			// Failsafe, so we don't get blocked by a progressbar
+			bar.IncrBy(length)
+		}
+	}
```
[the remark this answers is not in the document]
Naatan: The completion check is the same as it war before this PR and I will trust it was there for good reason, especially considering how shaky the progressbar stuff is.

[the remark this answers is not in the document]
daved: This is occurring during increments, though. The completion check should happen after the download is complete.

internal/download/get.go
Naatan: I falsely assumed `SplitN` would ensure the length of the resulting slice. Will fix.

daved: Just for reference, here is what I came up with trialing stuff yesterday (the logic may be off from what you intend, so this isn't necessarily meant to be a drop-in replacement): https://play.golang.org/p/r-qgMHUEmbi

internal/download/get.go
Naatan: Yeah, I tested with 0. It'll still show you it's pulling in stuff but you won't know how close it is to finishing, which in this scenario we can't know.

internal/download/get.go
Naatan: In theory it should "interrupt" when it's at 100% anyway. This is just a failsafe. Without this if for some reason it doesn't reach 100% due to some unforeseen bug it will just hang and the process won't exit.

internal/download/get.go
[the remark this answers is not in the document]
daved: Should be fine regardless. Marking resolved.

internal/download/get.go
```diff
@@ -31,81 +34,121 @@ func init() {
...
+		r.Region = constants.DefaultS3Region
+	} else if domain[1] == "s3" && domain[3] == "amazonaws" { // https://bucket-name.s3.Region.amazonaws.com/key-name
+		r.Bucket = domain[0]
+		r.Region = domain[2]
+	} else { // https://s3.Region.amazonaws.com/bucket-name/key-name
+		r.Bucket = strings.SplitN(url.Path, "/", 1)[0]
+		r.Region = domain[1]
+	}
+	return r, nil
 }
```
daved, line 137: This appears to be broken now: https://play.golang.org/p/11uiIA8xTON

daved (changes requested):

internal/download/get.go
Naatan: I don't see a problem?

daved: - https://bucket-name.s3.Region.amazonaws.com/key-name/extra
  - bucket: "/key-name/extra" region: "s3" key: "/key-name/extra"
    - bucket and region are incorrect
- https://bucket-name.s3.amazonaws.com/key-name/extra
  - bucket: "bucket-name" region: "us-east-1" key: "/key-name/extra"
    - looks good
- https://s3.Region.amazonaws.com/bucket-name/key-name/extra
  - bucket: "/bucket-name/key-name/extra" region: "Region" key: "/bucket-name/key-name/extra"
    - bucket and key are incorrect

Also, my version strips the slash prefix, but I'm not sure if that's needed.

daved (approved):

PR #964, opened 2020-09-10
Produce error code 11 on DNS errors

MDrakos:
https://www.pivotaltracker.com/story/show/174739637

### DIFF
--- a/internal/retryhttp/client.go
+++ b/internal/retryhttp/client.go
@@ -88,7 +88,7 @@ func normalizeResponse(res *http.Response, err error) (*http.Response, error) {
 
 	var dnsError *net.DNSError
 	if errors.Is(err, dnsError) {
-		return res, locale.WrapError(&UserNetworkError{}, "err_user_network_dns", "Request failed due to DNS error. {{.V0}}", solutionLocale)
+		return res, locale.WrapError(&UserNetworkError{}, "err_user_network_dns", "Request failed due to DNS error: {{.V0}}. {{.V1}}", err.Error(), solutionLocale)
 	}
 
 	return res, err
--- a/internal/retryhttp/client.go
+++ b/internal/retryhttp/client.go
@@ -1,6 +1,7 @@
 package retryhttp
 
 import (
+	"errors"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


internal/retryhttp/client.go
```diff
@@ -78,6 +85,12 @@ func normalizeResponse(res *http.Response, err error) (*http.Response, error) {
 			return res, locale.WrapInputError(&UserNetworkError{429}, "err_user_network_toomany", "Request failed due to too many requests. {{.V0}}", solutionLocale)
 		}
 	}
+
+	var dnsError *net.DNSError
+	if errors.Is(err, dnsError) {
+		return res, locale.WrapError(&UserNetworkError{}, "err_user_network_dns", "Request failed due to DNS error. {{.V0}}", solutionLocale)
```
Naatan, line 91:
```suggestion
		return res, locale.WrapError(&UserNetworkError{}, "err_user_network_dns", "Request failed due to DNS error: {{.V0}}. {{.V1}}", err.Error(), solutionLocale)
```

In this case I think we should actually include the error itself.

internal/retryhttp/client.go
```diff
@@ -67,6 +68,12 @@ func (c *Client) Do(req *retryablehttp.Request) (*http.Response, error) {
 	return normalizeResponse(c.Client.Do(req))
 }
 
+func (c *Client) StandardClient() *http.Client {
```
Naatan, line 71: What invokes this? And wouldn't this cause some sort of recursion in the retrying mechanic?

Naatan (changes requested):

internal/retryhttp/client.go
MDrakos: It's invoked [here](https://github.com/ActiveState/cli/blob/master/pkg/platform/api/graphql/graphql.go#L50) and various other places. Before this was using the `StandardClient` function from the retry library and not ours.

I don't think there should be any recursion issues as this is still using our client in the round tripper. I noticed in the retry library they use `sync.once` when initializing the round tripper and the retry client. We could do something similar here just to be sure.

Naatan (approved):

PR #967, opened 2020-09-10
We support relocation of artifacts produced by the alternative builder

mdrohmann:
https://www.pivotaltracker.com/story/show/174634293

### DIFF
--- a/pkg/platform/runtime/envdef/file_transform.go
+++ b/pkg/platform/runtime/envdef/file_transform.go
@@ -63,6 +63,7 @@ func (ft *FileTransform) relocateFile(fileBytes []byte, replacement string) ([]b
 	if len(*ft.PadWith) != 1 {
 		return fileBytes, errs.New("Padding character needs to have exactly one byte, got %d", len(*ft.PadWith))
 	}
+	pad := []byte(*ft.PadWith)[0]
 
 	// replacement should be shorter than search string
 	if len(replacementBytes) > len(findBytes) {
@@ -70,16 +71,18 @@ func (ft *FileTransform) relocateFile(fileBytes []byte, replacement string) ([]b
 		return fileBytes, locale.NewError("file_transform_replacement_too_long", "Replacement text cannot be longer than search text in a binary file.")
 	}
 
-	regexExpandBytes := []byte("${1}")
 	// Must account for the expand characters (ie. '${1}') in the
 	// replacement bytes in order for the binary paddding to be correct
+	regexExpandBytes := []byte("${1}")
 	replacementBytes = append(replacementBytes, regexExpandBytes...)
 
-	pad := []byte(*ft.PadWith)[0]
+	// paddedReplaceBytes is the replacement string plus the padding bytes added to the end
+	// It shall look like this: `<replacementBytes>${1}<padding>` with `len(replacementBytes)+len(padding)=len(findBytes)`
 	paddedReplaceBytes := bytes.Repeat([]byte{pad}, len(findBytes)+len(regexExpandBytes))
 	copy(paddedReplaceBytes, replacementBytes)
 
 	quoteEscapeFind := regexp.QuoteMeta(ft.Pattern)
+	// replacementRegex matches the search Pattern plus subsequent text up to the string termination character (pad, which usually is 0x00)
 	replacementRegex, err := regexp.Compile(fmt.Sprintf(`%s([^\\x%02x]*)`, quoteEscapeFind, pad))
 	if err != nil {
 		return fileBytes, errs.Wrap(err, "Failed to compile replacement regular expression.")
--- a/pkg/platform/runtime/envdef/file_transform.go
+++ b/pkg/platform/runtime/envdef/file_transform.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/fileutils"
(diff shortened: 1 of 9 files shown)


mdrohmann:
I tested the dummy artifacts both on linux and windows with the temporary integration test, and they showed the expected behavior - I believe.

I didn't manage to finish the unit tests.  I think it would be good to unit test the `relocateFile` and maybe the `applyConstTransforms` methods. Especially for a case where we have to relocate a binary file, because that is not happening with the dummy artifacts.

pkg/platform/runtime/envdef/file_transform.go
```diff
@@ -0,0 +1,144 @@
...
+	In      []string `json:"in"` // List of constants to apply this transform to
+	Pattern string   `json:"pattern"`
+	With    string   `json:"with"`
+}
+
+// applyConstTransforms applies the constant transforms to the Constants values
+func (ft *FileTransform) applyConstTransforms(constants Constants) (Constants, error) {
+	cs := constants
+	for _, ct := range ft.ConstTransforms {
+		fmt.Printf("pattern=%q\n", ct.Pattern)
```
Naatan, line 37: Remove debugging code

pkg/platform/runtime/envdef/file_transform.go
```diff
@@ -0,0 +1,144 @@
...
+	findBytes := []byte(ft.Pattern)
+	replacementBytes := []byte(replacement)
+
+	if ft.PadWith == nil {
+		return bytes.ReplaceAll(fileBytes, findBytes, replacementBytes), nil
+	}
+
+	// padding should be one byte
+	if len(*ft.PadWith) != 1 {
+		return fileBytes, errs.New("Padding character needs to have exactly one byte, got %d", len(*ft.PadWith))
```
Naatan, line 64: Is this what Shaun said? If not, can you get him to verify this.

pkg/platform/runtime/envdef/file_transform.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// replacement should be shorter than search string
+	if len(replacementBytes) > len(findBytes) {
```
Naatan, line 68: This is only really an issue if the file is a binary, which we can't know at this stage. Can you get Shaun to weigh in on how this should be handled with the alternate builder? One could argue we should be able to rely on the input of the builder (ie. they should have this error check, not us).

pkg/platform/runtime/envdef/file_transform.go
```diff
@@ -0,0 +1,144 @@
...
+		logging.Errorf("Replacement text too long: %s, original text: %s", ft.Pattern, replacement)
+		return fileBytes, locale.NewError("file_transform_replacement_too_long", "Replacement text cannot be longer than search text in a binary file.")
+	}
+
+	regexExpandBytes := []byte("${1}")
```
Naatan, line 73: Looking through the code I don't see any reason why this needs to be a slice, this is the only values it will ever hold. And none of the usage of this variable relies on slices.

pkg/platform/runtime/envdef/file_transform.go
```diff
@@ -0,0 +1,144 @@
...
+	// Must account for the expand characters (ie. '${1}') in the
+	// replacement bytes in order for the binary paddding to be correct
+	replacementBytes = append(replacementBytes, regexExpandBytes...)
+
+	pad := []byte(*ft.PadWith)[0]
+	paddedReplaceBytes := bytes.Repeat([]byte{pad}, len(findBytes)+len(regexExpandBytes))
```
Naatan, line 79: Why convert it from a byte and then to a byte again? Just use `string()` down below in your replacementRegex. Much more legible that way.

Naatan, line 79: `len(regexExpandBytes)` makes me think either I'm not understanding this code or you're using the wrong var. Wouldn't this -always- be 1?

pkg/platform/runtime/envdef/file_transform.go
```diff
@@ -0,0 +1,144 @@
...
+	copy(paddedReplaceBytes, replacementBytes)
+
+	quoteEscapeFind := regexp.QuoteMeta(ft.Pattern)
+	replacementRegex, err := regexp.Compile(fmt.Sprintf(`%s([^\\x%02x]*)`, quoteEscapeFind, pad))
```
Naatan, line 83: What is this pattern doing? Can you add a comment?

pkg/platform/runtime/envdef/file_transform.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	return res
+}
+
+// ApplyTransform applies a file transformation to all specified files
+func (ft *FileTransform) ApplyTransform(baseDir string, constants Constants) error {
+	// compute transformed constants
+	tcs, err := ft.applyConstTransforms(constants)
+	if err != nil {
+		fmt.Printf("%v\n", err)
```
Naatan, line 104: Drop debug statement

Naatan (changes requested):

pkg/platform/runtime/envdef/file_transform.go
mdrohmann: I put a question in the artifact document.  If the padding character is wider, it may be impossible to ensure that the relocated string has the same length as the original.

Naatan (approved):

PR #970, opened 2020-09-11
Path error dialog

MDrakos:
https://www.pivotaltracker.com/story/show/174760146

### DIFF
--- a/installers/msi-language/ActiveState/Error.cs
+++ b/installers/msi-language/ActiveState/Error.cs
@@ -41,6 +41,7 @@ namespace ActiveState
         {
             string registryKey = string.Format("HKEY_USERS\\{0}\\SOFTWARE\\ActiveState\\{1}", session.CustomActionData["USERSID"], session.CustomActionData["PRODUCT_NAME"]);
             RegistryValueKind registryEntryDataType = RegistryValueKind.String;
+            session.Log("Setting registry key: {0}", registryKey);
             try
             {
                 Registry.SetValue(registryKey, TypeRegistryKey, errorType, registryEntryDataType);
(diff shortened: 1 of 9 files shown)


installers/msi-language/ActiveState/Error.cs
```diff
@@ -35,35 +34,57 @@ public static void ResetErrorDetails(Session session)
...
+    }
+
+    public class NetworkError
+    {
+        private const string type = "Network";
+
+        public static void SetDetails(Session session, string msg)
+        {
+            session.Log("Setting network error details");
+            Error.SetDetails(session, "Network", msg);
```
Naatan, line 82:
```suggestion
            Error.SetDetails(session, type, msg);
```

installers/msi-language/ActiveState/Error.cs
```diff
@@ -35,35 +34,57 @@ public static void ResetErrorDetails(Session session)
...
+            Error.SetDetails(session, type, msg);
+        }
+
+        public static string Type()
+        {
+            return type;
+        }
+    }
+
+    public class NetworkError
```
Naatan, line 75: These 2 are basically identical aside from type. I'm sure C# has class inheritance that you could use here?

Naatan (changes requested):
Please ensure you've tested all error cases before merging

installers/msi-language/ActiveState/Error.cs
MDrakos: Yes we can. It changes how we use the errors but removes the duplicated code.

MDrakos:
I've tested the 3 error dialogs and the regular install. It all works as expected.

Naatan (approved):

PR #974, opened 2020-09-15
Disable repair and modify for MSI

Naatan:
https://www.pivotaltracker.com/story/show/174719572


MDrakos (approved):
Nice! That's easier than I expected it to be.

PR #975, opened 2020-09-16
Retry download body reads up to 3 times

Naatan:
https://www.pivotaltracker.com/story/show/174810912

I was getting uncomfortable with how much is going into https://github.com/ActiveState/cli/pull/959, so I took a quick stab at a solution and feels like this more concisely addresses the ACs.

If I'm missing something obvious we can discuss but otherwise I'd prefer we use this simpler approach for now. The other PR has too much going on in it.

### DIFF
--- a/internal/download/get.go
+++ b/internal/download/get.go
@@ -45,8 +45,12 @@ func httpGet(url string) ([]byte, *failures.Failure) {
 }
 
 func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *failures.Failure) {
-	logging.Debug("Retrieving url: %s", url)
-	client := retryhttp.NewClient(0 /* 0 = no timeout */, 5)
+	return httpGetWithProgressRetry(url, progress, 3)
+}
+
+func httpGetWithProgressRetry(url string, progress *progress.Progress, attempt int) ([]byte, *failures.Failure) {
+	logging.Debug("Retrieving url: %s, attempt: %d", url, attempt)
+	client := retryhttp.NewClient(0 /* 0 = no timeout */, 3)
 	resp, err := client.Get(url)
 	if err != nil {
 		code := -1
@@ -74,6 +78,13 @@ func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *fail
 	}
 
 	bar := progress.AddByteProgressBar(int64(total))
+	
+	// Ensure bar is always closed (especially for retries)
+	defer func() {
+		if !bar.Completed() {
+			bar.Abort(true)
+		}
+	}()
 
 	src := resp.Body
 	var dst bytes.Buffer
@@ -82,7 +93,11 @@ func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *fail
 
 	_, err = io.Copy(&dst, src)
 	if err != nil {
-		return nil, failures.FailInput.Wrap(err)
+		logging.Debug("Reading body failed: %s", err)
+		if attempt < 3 {
+			return httpGetWithProgressRetry(url, progress, attempt + 1)
+		}
+		return nil, failures.FailNetwork.Wrap(err)
 	}
 
 	if !bar.Completed() {
--- a/internal/download/get.go
+++ b/internal/download/get.go
@@ -45,7 +45,7 @@ func httpGet(url string) ([]byte, *failures.Failure) {
 }
 
 func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *failures.Failure) {
-	return httpGetWithProgressRetry(url, progress, 0, 3)
+	return httpGetWithProgressRetry(url, progress, 1, 3)
 }
 
 func httpGetWithProgressRetry(url string, progress *progress.Progress, attempt int, retries int) ([]byte, *failures.Failure) {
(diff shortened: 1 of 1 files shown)


internal/download/get.go
```diff
@@ -45,8 +45,12 @@ func httpGet(url string) ([]byte, *failures.Failure) {
 }
 
 func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *failures.Failure) {
-	logging.Debug("Retrieving url: %s", url)
-	client := retryhttp.NewClient(0 /* 0 = no timeout */, 5)
+	return httpGetWithProgressRetry(url, progress, 3)
```
daved, line 48: I believe that this will prevent any tries. There are two bits of information needed here, I think. `attempt` and `max`.

daved (changes requested):

internal/download/get.go
Naatan: D'oh, good eye.

internal/download/get.go
```diff
@@ -45,8 +45,12 @@ func httpGet(url string) ([]byte, *failures.Failure) {
...
 
 func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *failures.Failure) {
-	logging.Debug("Retrieving url: %s", url)
-	client := retryhttp.NewClient(0 /* 0 = no timeout */, 5)
+	return httpGetWithProgressRetry(url, progress, 0, 3)
+}
+
+func httpGetWithProgressRetry(url string, progress *progress.Progress, attempt int, retries int) ([]byte, *failures.Failure) {
+	logging.Debug("Retrieving url: %s, attempt: %d", url, attempt)
+	client := retryhttp.NewClient(0 /* 0 = no timeout */, retries)
```
daved, line 53: The quantity of http connection retries expanding exponentially here is not obvious to callers. It's probably fine for now being hidden away behind "httpGetWithProgress", but maybe a comment is warranted.

internal/download/get.go
```diff
@@ -75,14 +79,25 @@ func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *fail
...
 	if err != nil {
-		return nil, failures.FailInput.Wrap(err)
+		logging.Debug("Reading body failed: %s", err)
+		if attempt < 3 {
+			return httpGetWithProgressRetry(url, progress, attempt + 1, retries)
+		}
+		return nil, failures.FailNetwork.Wrap(err)
 	}
 
 	if !bar.Completed() {
```
daved, line 103: With this same logic being contained in a deferred function, it should be safe to remove this.

internal/download/get.go
```diff
@@ -75,14 +79,25 @@ func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *fail
...
 	src := resp.Body
 	var dst bytes.Buffer
 
 	src = bar.ProxyReader(resp.Body)
 
 	_, err = io.Copy(&dst, src)
 	if err != nil {
-		return nil, failures.FailInput.Wrap(err)
+		logging.Debug("Reading body failed: %s", err)
+		if attempt < 3 {
```
daved, line 97: Please use retries var instead of 3.

daved (changes requested):

internal/download/get.go
```diff
@@ -45,8 +45,12 @@ func httpGet(url string) ([]byte, *failures.Failure) {
 }
 
 func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *failures.Failure) {
-	logging.Debug("Retrieving url: %s", url)
-	client := retryhttp.NewClient(0 /* 0 = no timeout */, 5)
+	return httpGetWithProgressRetry(url, progress, 0, 3)
```
daved, line 48: Starting with attempt 0 will result in attempt counts showing as 0-2. It would make more sense in the logs if 0 was 1, and the conditional below was updated to `if attempt <= retries {`. For caution, httpGetWithProgressRetry can check for `<= 0` and bump to 1.

internal/download/get.go
Naatan: The thing under the conditional is different. This ones increasing, the other one is aborting.

internal/download/get.go
Naatan: Good call, missed that one.

internal/download/get.go
```diff
@@ -75,14 +79,25 @@ func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *fail
...
 	var dst bytes.Buffer
 
 	src = bar.ProxyReader(resp.Body)
 
 	_, err = io.Copy(&dst, src)
 	if err != nil {
-		return nil, failures.FailInput.Wrap(err)
+		logging.Debug("Reading body failed: %s", err)
+		if attempt <= retries {
+			return httpGetWithProgressRetry(url, progress, attempt + 1, retriesg)
```
daved, line 98: Looks like a typo `retriesg`

Naatan, line 98:
```suggestion
			return httpGetWithProgressRetry(url, progress, attempt + 1, retries)
```

internal/download/get.go
```diff
@@ -75,14 +79,25 @@ func httpGetWithProgress(url string, progress *progress.Progress) ([]byte, *fail
...
 	src := resp.Body
 	var dst bytes.Buffer
 
 	src = bar.ProxyReader(resp.Body)
 
 	_, err = io.Copy(&dst, src)
 	if err != nil {
-		return nil, failures.FailInput.Wrap(err)
+		logging.Debug("Reading body failed: %s", err)
+		if attempt <= retries {
```
daved, line 97: nit: It might be worth checking if the error is net timeout.

daved (approved):

PR #976, opened 2020-09-16
Ensure user_network event is properly captured

mdrohmann:
https://www.pivotaltracker.com/story/show/174783031

I ran the MSI without an internet connection, as this should certainly cause some `user_network` errrors.

Issues found and corrected:

- Dns errors were not correctly reported as (exit code 11 - user_network errors) in State Tool
- Sending a Rollbar report could crash the entire MSI if the geo-location service failed
- The GA/S3 tracking could block for ever on an offline computer (probably also one with a bad internet-connection)

I also cancelled the installation right at the start of the artifact installation, and observed the following:

- As the State Tool command was already inside the first HTTP request, it created the timeout error (after my cancellation).
- Instead of showing the cancellation screen, I got the user network error finishing screen.

All of this is probably good and means that it would be rather the other way round now: When the user cancels, there is still a chance that the `user_network` event gets reported instead of the cancel event.

### DIFF
--- a/installers/msi-language/ActiveState/Tracker.cs
+++ b/installers/msi-language/ActiveState/Tracker.cs
@@ -29,8 +29,9 @@ namespace ActiveState
             this._cid = GetInfo.GetUniqueId();
         }
 
-        public async Task<TrackingResult> TrackEventAsync(string sessionID, string category, string action, string label, string msiVersion, long value = 1)
+        public async Task<TrackingResult> TrackEventAsync(Session session, string sessionID, string category, string action, string label, string msiVersion, long value = 1)
         {
+            session.Log("Sending GA Event");
             var eventTrackingParameters = new EventTracking
             {
                 Category = category,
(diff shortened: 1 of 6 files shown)


installers/msi-language/ActiveState/Tracker.cs
```diff
@@ -110,10 +113,14 @@ public void TrackEventSynchronously(Session session, string msiLogFileName, stri
             session.Log("Sending event {0}/{1}/{2} for cid={3} (custom dimension 1: {4}, pid={5})", category, action, label, this._cid, langVersion, pid);
             var sessionID = computeSessionID(msiLogFileName);
             var t = Task.WhenAll(
-                TrackEventAsync(sessionID, category, action, label, langVersion, value),
+                TrackEventAsync(session, sessionID, category, action, label, langVersion, value),
                 TrackS3Event(session, sessionID, category, action, label)
             );
-            t.Wait();
+            var completed = t.Wait(TimeSpan.FromSeconds(15));
```
Naatan, line 119: Isn't this redundant with the client.Timeout?

Naatan (changes requested):

installers/msi-language/ActiveState/Tracker.cs
mdrohmann: As mentioned during stand-up, the MSI still blocks forever without this "outer" timeout.  Perhaps, because the GA Event blocks forever.  But I couldn't find a way to change the underlying http-client used by that library.

As discussed I suggest to leave the client timeout in there anyways, so we can use it when we add retries in a later story.

Naatan (approved):

PR #980, opened 2020-09-16
Fake progress bar during installation

MDrakos:
https://www.pivotaltracker.com/story/show/174777843

### DIFF
--- a/installers/msi-language/StateDeploy/CustomAction.cs
+++ b/installers/msi-language/StateDeploy/CustomAction.cs
@@ -153,14 +153,9 @@ namespace StateDeploy
                 return ActionResult.Success;
             });
 
-            Task waitTask = Task.Run(() =>
-            {
-                incrementTask.Wait();
-            });
-
             ActionResult result = downloadTask.Result;
             tokenSource.Cancel();
-            waitTask.Wait();
+            incrementTask.Wait();
             if (result.Equals(ActionResult.Failure))
             {
                 return result;
--- a/installers/msi-language/StateDeploy/CustomAction.cs
+++ b/installers/msi-language/StateDeploy/CustomAction.cs
@@ -129,19 +129,7 @@ namespace StateDeploy
 
             Task incrementTask = Task.Run(() =>
             {
-                if (token.IsCancellationRequested)
-                {
-                    token.ThrowIfCancellationRequested();
-                }
-                for (int i = 0; i <= 50; i++)
-                {
-                    if (token.IsCancellationRequested)
-                    {
-                        token.ThrowIfCancellationRequested();
-                    }
-                    Status.ProgressBar.Increment(session, 1);
-                    Thread.Sleep(150);
-                }
+                incrementProgressBar(session, 50, token);
             });
 
             Task<ActionResult> downloadTask = Task.Run(() =>
@@ -165,15 +153,14 @@ namespace StateDeploy
                 return ActionResult.Success;
             });
 
-            try
+            Task waitTask = Task.Run(() =>
             {
                 incrementTask.Wait();
-            } catch (OperationCanceledException)
-            {
-                session.Log("Increment progress bar was cancelled");
-            }
+            });
 
             ActionResult result = downloadTask.Result;
+            tokenSource.Cancel();
+            waitTask.Wait();
             if (result.Equals(ActionResult.Failure))
             {
                 return result;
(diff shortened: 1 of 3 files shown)


MDrakos:
Here is an example of the progress on my VM:
example

Naatan:
Can we speed it up by about 25%? I'm concerned the slow progress could itself cause cancellations. For most people a speedup of 25% would still not cause a hang at the end.

MDrakos:
@Naatan here is an example with the updated speed

example

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -122,21 +123,62 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
+                        token.ThrowIfCancellationRequested();
+                    }
+                    Status.ProgressBar.Increment(session, 1);
+                    Thread.Sleep(150);
+                }
+            });
+
+            Task<ActionResult> downloadTask = Task.Run(() =>
+            {
+                try
```
mdrohmann, line 149: Shouldn't this task request the cancellation on the token somewhere?

I assume that otherwise we always need to wait for the progressbar to finish, even if the download has finished before, no?

mdrohmann: A different way to implement this could also be [Task.WhenAny](
https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenany?view=netframework-4.5). When that returns, you can just always cancel the progress bar task, and wait for the download task.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -122,21 +123,62 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
+            try
+            {
+                incrementTask.Wait();
+            } catch (OperationCanceledException)
+            {
+                session.Log("Increment progress bar was cancelled");
             }
-            catch (WebException e)
+
+            ActionResult result = downloadTask.Result;
```
mdrohmann, line 176: I think here we need to add a `downloadTask.Wait()` first, in case the download task took longer than the `incrementTask`.

mdrohmann (changes requested):

installers/msi-language/StateDeploy/CustomAction.cs
MDrakos: From how I understand it, inspecting the `task.Result` is equivalent to a wait. From [this link](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1.result?view=netcore-3.1)
> Accessing the property's get accessor blocks the calling thread until the asynchronous operation is complete; it is equivalent to calling the Wait method.

So we wait for the download task to finish by inspecting the results and then cancel the progress bar. Though I could be misunderstanding how this works as this is the first time I've used threading in C#. Let me know if that makes sense or if there is something I've overlooked.

installers/msi-language/StateDeploy/CustomAction.cs
MDrakos: See comment above

installers/msi-language/StateDeploy/CustomAction.cs
mdrohmann: That is interesting.  I did not know that, or expected that to be the case... 

But what I still don't understand is how the cancellation would work: From what I understand the token(source) is never cancelled.  So, I suppose we will always block in line 170 until the progressbar has been updated 50 times... Or did I miss something?

mdrohmann (changes requested):

installers/msi-language/StateDeploy/CustomAction.cs
MDrakos: Okay I now see what you mean. I forgot to add the `token.Cance()` call. Further to that it would still have blocked on the wait call. I've updated now and tested to ensure that the progress bar incrementing is cancelled when the download has finished.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -165,15 +153,14 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
                 incrementTask.Wait();
-            } catch (OperationCanceledException)
-            {
-                session.Log("Increment progress bar was cancelled");
-            }
+            });
 
             ActionResult result = downloadTask.Result;
+            tokenSource.Cancel();
+            waitTask.Wait();
```
mdrohmann, line 163: Wouldn't it suffice to wait for the `incrementTask` here: `incrementTask.Wait()`?  Then lines 156-159 could be dropped.

mdrohmann (changes requested):

mdrohmann (approved):

PR #981, opened 2020-09-16
Ensure solveorder retries

daved:
https://www.pivotaltracker.com/story/show/174739083


daved:
@Naatan If this solution is acceptable, would you like this applied to other swagger params for retried calls within this story or another or not at all?

pkg/platform/model/recipe.go
```diff
@@ -144,6 +145,8 @@ func fetchRecipeID(commitID strfmt.UUID, owner, project, orgID string, private b
 	var err error
 	params := iop.NewSolveOrderParams()
 	params.SetHTTPClient(retryhttp.DefaultClient.StandardClient())
+	timeout := params.HTTPClient.Timeout * time.Duration(retryhttp.DefaultClient.RetryMax)
```
Naatan, line 148: Can you be sure `params.HTTPClient.Timeout` isn't `0` here?

And what's the rationale for multiplying it by `RetryMax`?

Makes sense applying this for other similar cases.

Naatan (changes requested):

pkg/platform/model/recipe.go
daved: It's safe to assume here because NewSolveOrderParams uses a global default value, but it's probably not safe to assume in most Params setups. The rationale for multiplying it is that if we limit downloads to, say 30 seconds, we would potentially need to accommodate a total duration of {retry_count} x 30 seconds to prevent premature cancellation.

I consider this a fairly vulgar solution, and it's not appearing to work without more digging into how the various timeouts and context deadlines work together. This is very much related to my concerns about http client resource reuse. Parallel to that, we have this issue in a number of places in different forms.

Naatan: Can you just add some simple logic to verify it's not 0, should be really cheap.

 Also I don't understand the rationale for multiplying. What do downloads have to do with this? This is for solving an order.

daved: Sure! As for multiplying, if that isn't done, then it could happen that the 5th try works after 29 seconds, then has only 1 second to complete. By extending it to the retryCt*timeout, we can be sure to give a fair amount of time during the connection once it's made. I've no strong feelings about it.

Naatan: Let's stick to just retrying for now. I see what you mean but we're not doing this anywhere else yet. Would prefer if we consider that to be an important aspect that we roll it out universally. Additionally still not sold on using RetryMax to multiply..

Naatan:
@daved btw did you test that adding these timeouts causes it to actually retry?

pkg/platform/api/headchef/headchef.go
```diff
@@ -132,10 +132,19 @@ func (b *BuildParams) WriteToRequest(req runtime.ClientRequest, reg strfmt.Regis
...
 
 func (r *Client) reqBuild(buildReq *headchef_models.V1BuildRequest, buildStatus *BuildStatus) {
+	defClient := retryhttp.DefaultClient
+	timeout := defClient.HTTPClient.Timeout
+	if timeout == 0 {
+		timeout = retryhttp.DefaultTimeout
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
```
Naatan, line 142: Can we bake all of this into retryhttp please, no need to see this chunk repeated for every call we make. We just need a HTTP client and a ctx, both of which retryhttp can provide.

Naatan (changes requested):

pkg/platform/api/headchef/headchef.go
```diff
@@ -132,10 +131,16 @@ func (b *BuildParams) WriteToRequest(req runtime.ClientRequest, reg strfmt.Regis
...
+
+	ctx, cancel := retryhttp.NewContext(nil, timeout)
+	defer cancel()
+
 	startParams := headchef_operations.StartBuildV1Params{
-		Context:      context.Background(),
+		Context:      ctx,
 		BuildRequest: buildReq,
-		HTTPClient:   retryhttp.DefaultClient.StandardClient(),
+		HTTPClient:   defClient.StandardClient(),
```
Naatan, line 143: This still looks more complicated than it needs to be in my eyes.

Can we simplify to something like:

```
	client := retryhttp.NewClient(60, 3)
	defer client.Close()

	startParams := headchef_operations.StartBuildV1Params{
		Context:      client.Context,
		BuildRequest: buildReq,
		HTTPClient:   client.HTTPClient,
	}
```

We want retryhttp to own as much of this as possible.

Naatan (changes requested):

pkg/platform/api/headchef/headchef.go
daved: The context.Context type is meant to be a per-request payload. For instance, in the net/http package, the Context type is not included as a field in the Client type, but is injected via a Request type field when calling the Client's Do method as opposed to the Get/Post/Head methods. I think it's very likely that those convenience methods would include a context.Context param if the context package had been part of the stdlib at that time net/http was authored.

My concern here is that we are adding to the complexity/conflation that is building as with the timeout wrangling.

daved: After speaking in Slack, we're going with creating a containing type called RetryHTTP to avoid any confusion about Request-specific data.

internal/testhelpers/srvstatus/srvstatus.go
```diff
@@ -0,0 +1,53 @@
+package srvstatus
```
Naatan, line 1: This should be in a test file

pkg/platform/api/headchef/headchef.go
```diff
@@ -132,10 +131,16 @@ func (b *BuildParams) WriteToRequest(req runtime.ClientRequest, reg strfmt.Regis
 }
 
 func (r *Client) reqBuild(buildReq *headchef_models.V1BuildRequest, buildStatus *BuildStatus) {
+	timeout := time.Second * 60
+	retries := 3
+
+	retry := retryhttp.New(retryhttp.NewClient(timeout, retries))
```
Naatan, line 137:
```suggestion
	retry := retryhttp.New(retryhttp.NewClient(time.Second * 60, 3))
```

Naatan: nit

pkg/platform/model/inventory.go
```diff
@@ -118,15 +118,19 @@ func SearchIngredientsStrict(language, name string) ([]*IngredientAndVersion, *f
...
+	defer retry.Close()
+
 	lim := int64(limit)
 
 	client := inventory.Get()
 
-	params := inventory_operations.NewGetNamespaceIngredientsParams()
+	params := inventory_operations.NewGetNamespaceIngredientsParamsWithContext(retry.Context)
+	params.SetHTTPClient(retry.Client.StandardClient())
+	params.SetTimeout(retry.Client.HTTPClient.Timeout)
```
Naatan, line 130: What's different about this `SetTimeout` ?

Naatan (approved):

internal/testhelpers/srvstatus/srvstatus.go
daved: Just tried and the retryhttp test can't import after: "no non-test Go files in srvstatus"

pkg/platform/model/inventory.go
daved: I am about >85% sure that the context.Context is passed via "client operation" here: https://github.com/ActiveState/cli/blob/f715b239c228d260274d91f58b1ec0b60a6f704c/pkg/platform/api/inventory/inventory_client/inventory_operations/inventory_operations_client.go#L2263-L2275

And then is the first encountered (and used) context value here: https://github.com/go-openapi/runtime/blob/d5a301537e5208a4b0c720f36ba75c0fa802c7df/client/runtime.go#L411-L439

I thought it safest to set both to the same timeout in case something was ever wired differently, or my following of the code was incorrect.

Naatan: I traced it down to http://github.com/ActiveState/cli/blob/77e2701270db9b9789e00e1274af701a5fa1443c/vendor/github.com/go-openapi/runtime/client/runtime.go#L384-L384

Please use `params.SetTimeout(60)` and throw away the retryhttp stuff we did for this PR. Seems just for addressing this story only a single line change was requested.

I highly recommend you invest some time in setting up your vim with proper code intelligence or using an IDE that provides you with this. It's a HUGE time saver.

Naatan (approved):

PR #984, opened 2020-09-17
Retry s3 pixel downloads up to 3 times

mdrohmann:
https://www.pivotaltracker.com/story/show/174827006

### DIFF
--- a/installers/msi-language/ActiveState/Tracker.cs
+++ b/installers/msi-language/ActiveState/Tracker.cs
@@ -61,14 +61,11 @@ namespace ActiveState
                 // retry up to 3 times to download the S3 pixel
                 RetryHelper.RetryOnException(session, 3, TimeSpan.FromSeconds(1), () =>
                 {
-                    Task.Run(async () =>
-                    {
-                        var client = new TimeoutWebClient();
-                        // tr tp complete an s3 tracking event in seven seconds or less.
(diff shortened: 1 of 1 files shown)


installers/msi-language/ActiveState/Tracker.cs
```diff
@@ -58,12 +58,20 @@ public async Task TrackS3Event(Session session, string sessionID, string categor
...
-                await client.DownloadStringTaskAsync(pixelURL);
+                // retry up to 3 times to download the S3 pixel
+                RetryHelper.RetryOnException(session, 3, TimeSpan.FromSeconds(1), () =>
+                {
+                    Task.Run(async () =>
+                    {
+                        var client = new TimeoutWebClient();
+                        // tr tp complete an s3 tracking event in seven seconds or less.
+                        client.Timeout = 7 * 1000;
+                        var res = await client.DownloadStringTaskAsync(pixelURL);
```
MDrakos, line 69: Could we just use `client.DownloadString` rather than this async method if we are just going to wait for it anyways?

MDrakos (changes requested):

installers/msi-language/ActiveState/Tracker.cs
mdrohmann: That is a good point.

MDrakos (approved):

PR #986, opened 2020-09-17
setting UILevel as third dimenstion in GA tracking

mdrohmann:
https://www.pivotaltracker.com/story/show/174865312


MDrakos (changes requested):
~Having the UI level in Rollbar is a good addition, however the AC in the story is to add UI level to Google Analytics~

Scratch that, sorry I missed some files

MDrakos (approved):

PR #994, opened 2020-09-23
set LAST_DIALOG for Maintenance mode

mdrohmann:
https://www.pivotaltracker.com/story/show/174936746

It turns out that the user still can run the MSI twice.  They will see the `MaintenanceWelcomeDlg` and then the `MaintenanceTypeDlg` looking like this:
 
remove_only_maintenance

So, they can *only* remove the MSI here, "Change" and "Repair" are greyed out.

This PR adds `LAST_DIALOG` messages for these cases making it clear where the user cancels in case they are in this mode. 

Note: If they select _Remove_, the `VerifyReadyDlg` is shown, and I made sure that `LAST_DIALOG` is distinct from the values during a regular installation: `VerifyReadyRemoveDlg` and `Post_VerifyReadyRemoveDlg`.

### DIFF
--- a/installers/msi-language/Product.p.wxs
+++ b/installers/msi-language/Product.p.wxs
@@ -23,8 +23,8 @@
 		<!-- End Configuration parameters -->
 
 		<!-- Disable Modification -->
-		<Property Id="ARPNOREPAIR" Value="yes" Secure="yes" />
-		<Property Id="ARPNOMODIFY" Value="yes" Secure="yes" />
+		<Property Id="ARPNOREPAIR" Value="1" Secure="yes" />
+		<Property Id="ARPNOMODIFY" Value="1" Secure="yes" />
 		<!-- End Disable Modification -->
 
 		<Property Id="LAST_DIALOG" Value="unset" Secure="yes" Admin="yes" />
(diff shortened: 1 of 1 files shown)


installers/msi-language/Product.p.wxs
```diff
@@ -124,17 +124,23 @@
...
 			<Publish Dialog="ProgressDlg" Control="Next" Property="LAST_DIALOG" Value="Post_ProgressDlg">1</Publish>
 			<Publish Dialog="WelcomeDlg" Control="Next" Property="LAST_DIALOG" Value="Post_WelcomeDlg">1</Publish>
 			<Publish Dialog="MaintenanceWelcomeDlg" Control="Next" Property="LAST_DIALOG" Value="Post_MaintenanceWelcomeDlg">1</Publish>
+			<Publish Dialog="MaintenanceTypeDlg" Control="RemoveButton" Property="LAST_DIALOG" Value="Post_RemoveDlg">1</Publish>
 			<Publish Dialog="LicenseAgreementDlg" Control="Next" Property="LAST_DIALOG" Value="Post_LicenseAgreementDlg">1</Publish>
 			<Publish Dialog="InstallDirDlg" Control="Next" Property="LAST_DIALOG" Value="Post_InstallDirDlg">1</Publish>
-			<Publish Dialog="VerifyReadyDlg" Control="Install" Property="LAST_DIALOG" Value="Post_VerifyReadyDlg">1</Publish>
+			<Publish Dialog="VerifyReadyDlg" Control="Install" Property="LAST_DIALOG" Value="Post_VerifyReadyDlg">NOT Installed</Publish>
+			<Publish Dialog="VerifyReadyDlg" Control="InstallNoShield" Property="LAST_DIALOG" Value="Post_VerifyReadyDlg">NOT Installed</Publish>
```
mdrohmann, line 141: I added this line, and the line for `RemoveNoShield` below, because that is the Control's name in the `VerifyReadyDlg` if no UAC is necessary ie., if the user is already an administrator.  So, we may not have correctly caught these cases before...

Cf., https://github.com/wixtoolset/wix3/blob/2ed6d7906c87bb65b75cfa0020e38a1b1dffa49b/src/ext/UIExtension/wixlib/VerifyReadyDlg.wxs#L19

Naatan (changes requested):
The point of this story was to address uninstalls affecting messing with our analytics. We're trying to optimize the installation flow, not the uninstallation flow. Let's close this story, I'll open a new one to address it via a dimension instead.

Naatan:
https://www.pivotaltracker.com/story/show/174956249

mdrohmann:
> The point of this story was to address uninstalls affecting messing with our analytics. We're trying to optimize the installation flow, not the uninstallation flow.

That makes sense to me and my intention was to have all uninstallation events separated from the Installation events such that they can be separated out in GA.  Of course, it is better to add a custom dimension, as it makes it easier to separate them out.  But why don't I do this in this PR, as the changes are kind of prerequisite for it?

PR #995, opened 2020-09-23
Track install type as custom dimension

mdrohmann:
https://www.pivotaltracker.com/story/show/174956249

### DIFF
--- a/installers/msi-language/ActiveState/Tracker.cs
+++ b/installers/msi-language/ActiveState/Tracker.cs
@@ -29,7 +29,7 @@ namespace ActiveState
             this._cid = GetInfo.GetUniqueId();
         }
 
-        public async Task<TrackingResult> TrackEventAsync(Session session, string sessionID, string category, string action, string label, string productVersion, string uilevel, long value = 1)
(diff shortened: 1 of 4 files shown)


installers/msi-language/ActiveState/Tracker.cs
```diff
@@ -81,15 +82,46 @@ public async Task TrackS3Event(Session session, string sessionID, string categor
...
+            private string getValueFromSession(Session session, string key1, string key2)
+            {
+                if (!session.GetMode(InstallRunMode.Scheduled))
+                {
+                    return session[key1];
+                }
+                if (session.CustomActionData.ContainsKey(key2)) {
+                    return session.CustomActionData[key2];
+                }
+                return "";
```
MDrakos, line 122: Might be worthwhile surrounding this in a try-catch to ensure we don't get an exception for a key error, even though that is very unlikely.

On that note I don't think we're sending the `INSTALL_MODE` property on uninstall. Is that a value we want to capture?

MDrakos (changes requested):

MDrakos (approved):

PR #996, opened 2020-09-23
Update default activestate.yaml

MDrakos:
https://www.pivotaltracker.com/story/show/174871451

### DIFF
--- a/internal/runners/activate/checkout.go
+++ b/internal/runners/activate/checkout.go
@@ -56,6 +56,11 @@ func (r *Checkout) Run(namespace string, targetPath string) error {
 		}
 	}
 
+	language, fail := model.DefaultLanguageForProject(ns.Owner, ns.Project)
+	if fail != nil {
+		return fail
+	}
+
 	// Create the config file, if the repo clone didn't already create it
 	configFile := filepath.Join(targetPath, constants.ConfigFileName)
 	if !fileutils.FileExists(configFile) {
@@ -64,6 +69,7 @@ func (r *Checkout) Run(namespace string, targetPath string) error {
 			Project:   ns.Project,
 			CommitID:  commitID,
 			Directory: targetPath,
+			Language:  language,
 		})
 		if fail != nil {
 			return fail
--- a/internal/runners/activate/checkout.go
+++ b/internal/runners/activate/checkout.go
@@ -2,8 +2,6 @@ package activate
 
 import (
 	"path/filepath"
-	"strconv"
-	"strings"
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/fileutils"
@@ -13,7 +11,6 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/ActiveState/cli/pkg/projectfile"
-	"github.com/blang/semver"
 )
 
 type CheckoutAble interface {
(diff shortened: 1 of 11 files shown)


pkg/projectfile/projectfile.go
```diff
@@ -816,8 +816,13 @@ func createCustom(params *CreateParams) (*Project, *failures.Failure) {
 	if runtime.GOOS == "windows" {
 		shell = "batch"
 	}
+
+	yaml := "sample_yaml"
+	if strings.ToLower(params.Language) == constants.PerlLanguageName {
```
Naatan, line 821: Please see if you can use http://github.com/ActiveState/cli/blob/77e2701270db9b9789e00e1274af701a5fa1443c/internal/language/language.go#L22-L22

locale/en-us.yaml
```diff
@@ -1068,6 +1068,27 @@ language_name_unknown:
   other: Unknown
 language_unknown_options:
   other: Unknown options
+sample_yaml_perl:
```
Naatan, line 1071: What about Python?

locale/en-us.yaml
```diff
@@ -1068,6 +1068,27 @@ language_name_unknown:
...
+          print "Your 'activated state' allows you to manage packages, scripts, secrets and more via the activestate.yaml file at the root of your project directory.\n\n";
+          print "To manage packages use the `state packages` command. For more information about package management, use `state packages --help`. For more information about the State Tool use `state help`.\n\n";
+          print "To access additional features of the ActiveState Platform and use a web-based interface for managing your project, you can visit https://platform.activestate.com/{{.Owner}}/{{.Project}}.\n\n";
+          print "Edit your activestate.yaml to remove this message.\n\n";
+    events:
+      # This is the ACTIVATE event, it will run whenever a new virtual environment is created (eg. by running `state activate`)
+      # On Linux and macOS this will be ran as part of your shell's rc file, so you can use it to set up aliases, functions, environment variables, etc.
+      - name: ACTIVATE
+        value: $scripts.activationMessage.path()
 sample_yaml:
```
Naatan, line 1092: We can drop this. It's either Perl or Python.

locale/en-us.yaml
```diff
@@ -1068,6 +1068,27 @@ language_name_unknown:
...
+    scripts:
+    # This script uses a secret. Note that you can define your own secrets at
+    # https://platform.activestate.com/{{.Owner}}/{{.Project}}/scripts
+      - name: helloWorld
+        language: {{.Shell}}
+        value: echo ${secrets.user.world}
+      - name: activationMessage
+        language: perl
+        value: |
+          print "You are now in an 'activated state', this will give you a virtual environment to work in that doesn't affect the rest of your system.\n\n";
```
Naatan, line 1082: I'm assuming Perl has a better solution than individual `print` statements for each line. Try to do it with a single print statement. 

The current implementation is awkward because of limitations we should no longer have.

Naatan (changes requested):

locale/en-us.yaml
MDrakos: I left it out due to the story points and the Perl specific title and description. It's added now.

internal/runners/activate/checkout.go
```diff
@@ -72,3 +81,21 @@ func (r *Checkout) Run(namespace string, targetPath string) error {
...
 }
+
+func getLanguage(owner, project string) (string, error) {
+	modelLanguage, fail := model.DefaultLanguageForProject(owner, project)
+	if fail != nil {
+		return "", fail
+	}
+
+	language := modelLanguage.Name
+	if strings.ToLower(modelLanguage.Name) == "python" {
```
Naatan, line 92: Hard coded "python"

internal/runners/activate/checkout.go
```diff
@@ -72,3 +81,21 @@ func (r *Checkout) Run(namespace string, targetPath string) error {
...
+		version, err := semver.Parse(modelLanguage.Version)
+		if err != nil {
+			return language, err
+		}
+		language = modelLanguage.Name + strconv.FormatUint(version.Major, 10)
```
Naatan, line 97: I think you basically just need http://github.com/ActiveState/cli/blob/77e2701270db9b9789e00e1274af701a5fa1443c/internal/language/language.go#L115-L115

Naatan (changes requested):

internal/runners/activate/checkout.go
MDrakos: The name value we get from the API for both versions of python is just `python`. The `MakeByName` function expects `python2` or `python3` for a successful lookup.

Naatan: I see. In that case let's keep this logic but move it next to `MakeByName` and name it `MakeByNameAndversion`

Naatan (approved):

PR #998, opened 2020-09-23
Start prepare command

daved:
https://www.pivotaltracker.com/story/show/174913214

This is the first part of the above story; To simplify development and testing of changes to installers, this should be merged first. The second part is #1000


cmd/state/internal/cmdtree/prepare.go
```diff
@@ -0,0 +1,26 @@
...
+		"for programmatic use only",
+		[]*captain.Flag{},
+		[]*captain.Argument{},
+		func(_ *captain.Command, _ []string) error {
+			return runner.Run()
+		},
+	)
+
+	cmd.SetHidden(true)
+	cmd.SetDeferAnalytics(true) // ?
```
Naatan, line 23:
```suggestion
```

cmd/state/internal/cmdtree/prepare.go
```diff
@@ -0,0 +1,26 @@
+package cmdtree
+
+import (
+	"github.com/ActiveState/cli/internal/captain"
+	"github.com/ActiveState/cli/internal/primer"
+	"github.com/ActiveState/cli/internal/runners/prepare"
+)
+
+func newPrepareCommand(prime *primer.Values) *captain.Command {
+	runner := prepare.New(prime.Output())
```
Naatan, line 10: Pass prime itself. Please follow the pattern.

internal/captain/command.go
```diff
@@ -172,6 +172,10 @@ func (c *Command) SetDeferAnalytics(value bool) {
 	c.deferAnalytics = value
 }
 
+func (c *Command) SetHidden(value bool) {
```
Naatan, line 175: We seem to already have something for this:

http://github.com/ActiveState/cli/blob/77e2701270db9b9789e00e1274af701a5fa1443c/internal/captain/option.go#L7-L7

Naatan (changes requested):
This is missing the MSI changes

internal/captain/command.go
daved: Apologies if this is obvious; Where do the options get set/processed?

Naatan: Hmm looks like it got dropped along the line somewhere, and I can't find where it originated from. Ok disregard my comment here. Feel free to remove that file if you like, seems both functions are unused.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -267,6 +267,25 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
+                record.FormatString = string.Format("state _prepare failed with error:\n{0}", errorOutput);
+
+                session.Message(InstallMessage.Error | (InstallMessage)MessageBoxButtons.OK, record);
+                return prepareRunResult;
+            }
+            else
+            {
+                session.Log(string.Format("Prepare Output: {0}", prepareOutput));
+            }
+
```
daved, line 288: @mdrohmann I'd like there to be logged info regardless of ActionResult value. Will this log the error output? Also, I was unsure about the need for rollbar reporting here, and very unsure about how to deal with the session.Message line. Any help is appreciated.

cmd/state/internal/cmdtree/prepare.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/ActiveState/cli/internal/primer"
+	"github.com/ActiveState/cli/internal/runners/prepare"
+)
+
+func newPrepareCommand(prime *primer.Values) *captain.Command {
+	runner := prepare.New(prime)
+
+	cmd := captain.NewCommand(
+		"_prepare",
+		"for programmatic use only",
```
Naatan, line 14:
```suggestion
		locale.Tl("Prepare environment for use with the State Tool."),
```

Naatan (changes requested):

cmd/state/internal/cmdtree/prepare.go
daved: I added this exact message, but provided a l10n key according to `Tl` usage.

Naatan (approved):

PR #999, opened 2020-09-25
Default installs symlink dir

MDrakos:
https://www.pivotaltracker.com/story/show/174871624


MDrakos:
I may have misinterpreted the ACs in this story so please let me know if any of these changes are in the wrong place

Naatan:
@MDrakos I'm afraid you missed a very crucial AC, the first one:

> Implement using state _prepare

The point of `state _prepare` is so we don't implement this redundantly in each installer using languages we're not comfortable with.

MDrakos:
Closing this as diff is showing incorrect changeset

PR #1000, opened 2020-09-25
Add prepare call to installers

daved:
https://www.pivotaltracker.com/story/show/174913214

This is the second part of the above story; To simplify development and testing of changes to installers, #998 should be merged first.


.github/workflows-src/steps.lib.yml
```diff
@@ -112,7 +113,10 @@ env:
 #@ def steps_inttest():
 name: "Integration Tests"
 shell: bash
-run: SHELL='' go test -timeout 20m ./test/integration -v
+run: |
+  export SHELL=''
+  mkdir -p logs
+  go test -timeout 20m ./test/integration -v
```
Naatan, line 119: Can we get rid of this stuff and instead just echo the log file after running the job. The job outputs ARE the logs. This is overcomplicating things.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -250,21 +250,44 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
+                {
+                    Environment.SetEnvironmentVariable("PATH", newPath, EnvironmentVariableTarget.Machine);
+                }
+                catch (Exception e)
+                {
+                    string msg = string.Format("Could not update PATH. Encountered exception: {0}", e.Message);
+                    session.Log(msg);
+                    new SecurityError().SetDetails(session, msg);
+                    return ActionResult.Failure;
+                }
```
Naatan, line 268: This shouldn't be necessary.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -250,21 +250,44 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
             }
-            catch (Exception e)
+
+            session.Log("Running prepare step...");
+            string prepareCmd = " --output=json" + " _prepare";
```
Naatan, line 272: Nit: lead with the command, not the flags.
```suggestion
            string prepareCmd = " _prepare --output=json";
```

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -250,21 +250,44 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
-                string msg = string.Format("Could not update PATH. Encountered exception: {0}", e.Message);
+                string msg = string.Format("`state _prepare` returned with error: {0}", prepareOutput);
                 session.Log(msg);
-                new SecurityError().SetDetails(session, msg);
-                return ActionResult.Failure;
+                RollbarReport.Critical(msg, session);
+
+                Record record = new Record();
+                var errorOutput = Command.FormatErrorOutput(prepareOutput);
+                record.FormatString = string.Format("state _prepare failed with error:\n{0}", errorOutput);
```
Naatan, line 283:
```suggestion
                record.FormatString = msg;
```

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -250,21 +250,44 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
-            catch (Exception e)
+
+            session.Log("Running prepare step...");
+            string prepareCmd = " --output=json" + " _prepare";
+            string prepareOutput;
+            ActionResult prepareRunResult = ActiveState.Command.Run(session, stateToolPath, prepareCmd, out prepareOutput);
+            if (prepareRunResult.Equals(ActionResult.Failure))
             {
-                string msg = string.Format("Could not update PATH. Encountered exception: {0}", e.Message);
+                string msg = string.Format("`state _prepare` returned with error: {0}", prepareOutput);
```
Naatan, line 277:
```suggestion
                string msg = string.Format("Preparing environment caused error: {0}", prepareOutput);
```

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -250,21 +250,44 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
-                new SecurityError().SetDetails(session, msg);
-                return ActionResult.Failure;
+                RollbarReport.Critical(msg, session);
+
+                Record record = new Record();
+                var errorOutput = Command.FormatErrorOutput(prepareOutput);
+                record.FormatString = string.Format("state _prepare failed with error:\n{0}", errorOutput);
+
+                session.Message(InstallMessage.Error | (InstallMessage)MessageBoxButtons.OK, record);
+                return prepareRunResult;
```
Naatan, line 286:
```suggestion
                return ActionResult.Failure;
```

Be explicit

Naatan (changes requested):

.github/workflows-src/steps.lib.yml
daved: I am mostly sure that that would mean a failure in the tests will result in no msi log output. Is there a simpler mechanism to ensure those logs will print? It's not clear if TRAP would work, and I'm out of other ideas at the moment.

Naatan: Just capture the exit code, do the print and then exit with the exit code? Should be way cheaper than all the stuff going on in this PR to produce logs as artifacts.

installers/msi-language/StateDeploy/CustomAction.cs
daved: This was existing. Can I file a bug if it is wrong/useless?

.github/workflows-src/steps.lib.yml
Naatan: Heck just a `set +e` might actually suffice

installers/msi-language/StateDeploy/CustomAction.cs
daved: I'm committing the suggestion, but I was trying to be cautious in the case that there was other data managed/propagated by that instance of the result.

.github/workflows-src/steps.lib.yml
daved: Yeah, I think that would work. I don't know why I thought that was already the case and some other mechanic was in play.

.github/workflows-src/steps.lib.yml
```diff
@@ -102,7 +102,11 @@ name: "Critical Integration Tests"
 shell: bash
 run: |
   export SHELL=''
+  set +e
   go test -timeout 10m github.com/ActiveState/cli/test/integration -test.v -test.run "^TestLanguageMsiActivePerl|TestDeployIntegrationTestSuite|TestInitIntegrationTestSuite|TestPushIntegrationTestSuite|TestUpdateIntegrationTestSuite|PrepareIntegrationTestSuite$"
+  testCode=$?
+  head -n-0 build/*log 2>/dev/null
```
Naatan, line 108: Curious, why `head` and not `cat`?

.github/workflows-src/steps.lib.yml
```diff
@@ -102,7 +102,11 @@ name: "Critical Integration Tests"
...
+  exit ${testCode}
```
Naatan, line 109: Doesn't `set +e` make this redundant?

ie. the only change required is `set +e` and `cat build/*.log` at the end

.github/workflows-src/steps.lib.yml
```diff
@@ -102,7 +102,11 @@ name: "Critical Integration Tests"
 shell: bash
```
Naatan, line 102: Doesn't look like this PR needs the state tool binaries rebuilt, please remove them so we don't needlessly balloon our repo size.

installers/msi-language/StateDeploy/CustomAction.cs
Naatan: Oh I see, I thought this was being done for our `_prepare` command execution. My bad, disregard.

installers/msi-language/StateDeploy/CustomAction.cs
```diff
@@ -250,22 +250,45 @@ private static ActionResult _installStateTool(Session session, out string stateT
...
+                    string msg = string.Format("Could not update PATH. Encountered exception: {0}", e.Message);
+                    session.Log(msg);
+                    new SecurityError().SetDetails(session, msg);
+                    return ActionResult.Failure;
+                }
             }
-            catch (Exception e)
+
+            session.Log("Running prepare step...");
+            string prepareCmd = " _prepare --output=json";
```
Naatan, line 272:
```suggestion
            string prepareCmd = " _prepare";
```

Naatan (changes requested):

.github/workflows-src/steps.lib.yml
daved: Head prints out the file name before each file body.

.github/workflows-src/steps.lib.yml
daved: We want the success/failure to be solely determined by the `go test` results; The `head` call will throw off the value of `$?`, so we have to explicitly set the exit code.

.github/workflows-src/steps.lib.yml
daved: This is necessary to bring in the _prepare cmd changes. I opted to do it in this PR since this PR introduces the calls to that cmd.

Naatan (approved):

PR #1001, opened 2020-09-25
Default installs directory

MDrakos:
https://www.pivotaltracker.com/story/show/174871624


internal/runners/prepare/prepare.go
```diff
@@ -1,30 +1,63 @@
...
+		out:      prime.Output(),
+		subshell: prime.Subshell(),
 	}
 }
 
 // Run executes the prepare behavior.
 func (r *Prepare) Run() error {
 	logging.Debug("ExecutePrepare")
 
+	binDir := filepath.Join(config.CachePath(), "prepareBin")
```
Naatan, line 41:
```suggestion
	binDir := filepath.Join(config.CachePath(), "bin")
```

internal/runners/prepare/prepare.go
```diff
@@ -1,30 +1,63 @@
...
+	fail := fileutils.Mkdir(binDir)
+	if fail != nil {
+		return locale.WrapError(fail.ToError(), "err_prepare_bin_dir", "Could not create global installations directory")
```
Naatan, line 44:
```suggestion
		return locale.WrapError(fail.ToError(), "err_prepare_bin_dir", "Could not create bin directory")
```

internal/runners/prepare/prepare.go
```diff
@@ -1,30 +1,63 @@
...
+	}
+
+	envUpdates := map[string]string{
+		"PATH": fmt.Sprintf("%s%s%s", binDir, string(os.PathListSeparator), os.Getenv("PATH")),
+	}
+
+	fail = r.subshell.WriteUserEnv(envUpdates, true)
```
Naatan, line 51: There's an issue with the current approach. `WriteUserEnv` has logic that cleans up previous writes to the env before writing the new one. The problem here is that it assumes specifically the `deploy` use-case. Which means that if you use both `state deploy` and `state activate --default` then one will override the other.

We need to update the mechanic here so the two can live side by side.

internal/runners/prepare/prepare.go
```diff
@@ -1,30 +1,63 @@
...
+	if fail != nil {
+		return locale.WrapError(fail.ToError(), "err_prepare_update_env", "Could not update user environment")
```
Naatan, line 53: This shouldn't cause a failure on Linux. Just a warning.

Naatan (changes requested):
Provided it doesn't make things too difficult can we also include an integration test for this command.

internal/runners/prepare/prepare.go
```diff
@@ -38,19 +37,20 @@ func New(prime primeable) *Prepare {
...
-		"PATH": fmt.Sprintf("%s%s%s", binDir, string(os.PathListSeparator), os.Getenv("PATH")),
+		"PATH": binDir,
 	}
 
-	fail = r.subshell.WriteUserEnv(envUpdates, true)
+	fail = r.subshell.WriteUserEnv(envUpdates, sscommon.Default, true)
 	if fail != nil {
-		return locale.WrapError(fail.ToError(), "err_prepare_update_env", "Could not update user environment")
+		logging.Debug("Encountered failure attempting to update user environment: %s", fail.ToError())
+		r.out.Notice("Could not update user environment")
```
Naatan, line 53: This should still fail on Mac and Windows. My review said "on Linux".

internal/subshell/sscommon/env.go
```diff
@@ -0,0 +1,49 @@
...
+	start string
+	stop  string
+	key   string
+}
+
+var lookup = [...]envData{
+	{
+		constants.RCAppendDeployStartLine,
+		constants.RCAppendDeployStopLine,
+		"user_deploy_env",
```
Naatan, line 22:
```suggestion
		"user_env",
```
Changing this would needlessly break older versions that relied on it.

internal/subshell/sscommon/env.go
```diff
@@ -0,0 +1,49 @@
+package sscommon
+
+import "github.com/ActiveState/cli/internal/constants"
+
+type EnvType int
+
+const (
+	Deploy EnvType = iota
+	Default
```
Naatan, line 9: I don't see the point in making this an iota that then relies on a lookup table. The values could just be the envData.

Naatan (changes requested):

internal/subshell/sscommon/env.go
MDrakos: The idea was to restrict users of `WriteUserEnv` and `WriteRCFile` so they are not passing their own start, stop, and config key information that could potential conflict with deploy and default, or not have any identifying information at all. It's arguable if this package should dictate that though.

I've left a couple `var`s to be helpful but exported the data type.

internal/runners/prepare/prepare.go
MDrakos: Whoops, missed that.

internal/runners/deploy/deploy.go
```diff
@@ -217,7 +218,12 @@ func configure(installpath string, envGetter runtime.EnvGetter, out output.Outpu
 
 	out.Notice(locale.Tr("deploy_configure_shell", sshell.Shell()))
 
-	fail := sshell.WriteUserEnv(env, sscommon.Deploy, userScope)
+	envData := sscommon.EnvData{
+		constants.RCAppendDeployStartLine,
+		constants.RCAppendDeployStopLine,
+		"user_env",
+	}
```
Naatan, line 225: The envData should still live in the file you had, so it's centralized and reusable. I was just saying the lookup stuff seemed excessive.

internal/runners/prepare/prepare.go
```diff
@@ -47,8 +48,16 @@ func (r *Prepare) Run() error {
...
+		constants.RCAppendDefaultStopLine,
+		"user_default_env",
+	}
+	fail = r.subshell.WriteUserEnv(envUpdates, data, true)
 	if fail != nil {
+		if runtime.GOOS != "linux" {
+			return locale.WrapError(fail.ToError(), "err_prepare_update_env", "Could not update user environment")
+		}
 		logging.Debug("Encountered failure attempting to update user environment: %s", fail.ToError())
 		r.out.Notice("Could not update user environment")
```
Naatan, line 62: Should be localized

Naatan (changes requested):

Naatan (approved):

PR #1002, opened 2020-09-25
add --default flag to `state activate`

mdrohmann:
https://www.pivotaltracker.com/story/show/174871470


internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+	out.Print(locale.Tl("default_shim", "Writing default installation to {{.V0}}.", targetPath))
+	for _, exePath := range exePaths {
+		shimPath := shimTargetPath(targetPath, exePath)
+
+		// The link should not exist as we are always rolling back old shims before we run this code.
+		if fileutils.TargetExists(shimPath) {
+			return locale.NewInputError(
+				"err_default_symlink_target_exists",
+				"Cannot create shim as the target already exists: {{.V0}}.", shimPath)
+		}
```
Naatan, line 226: Since we own this directory let's just overwrite the target and ensure the issue is logged to rollbar. Most likely the rollback just didn't run properly, we can still give it a best effort while logging the error so we can follow up internally.

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+
+		logging.Debug("Shimming %s at %s", exePath, shimPath)
+		if err := shim(exePath, shimPath); err != nil {
```
Naatan, line 229: Let's call this `createShimFile`

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+		shimPath := shimTargetPath(targetPath, exePath)
+
+		// The link should not exist as we are always rolling back old shims before we run this code.
+		if fileutils.TargetExists(shimPath) {
+			return locale.NewInputError(
+				"err_default_symlink_target_exists",
+				"Cannot create shim as the target already exists: {{.V0}}.", shimPath)
+		}
+
+		logging.Debug("Shimming %s at %s", exePath, shimPath)
```
Naatan, line 228: This debug statement seems like it should live inside the function you're calling below

test/integration/activate_int_test.go
```diff
@@ -352,6 +352,36 @@ func (suite *ActivateIntegrationTestSuite) TestActivate_Command() {
 	cp.ExpectExitCode(0)
 }
 
+func (suite *ActivateIntegrationTestSuite) TestActivate_Default() {
```
Naatan, line 355: Rather than running a whole new activation test which slows down our overall test progress let's just add `--default` to one of our existing `state activate` tests. This is just one additional assertion.

Naatan (changes requested):

internal/runners/activate/activation.go
```diff
@@ -90,6 +104,13 @@ func activate(proj *project.Project, out output.Outputer, subs subshell.SubShell
 		return false, locale.WrapError(fail, "error_could_not_activate_venv", "Could not activate project. If this is a private project ensure that you are authenticated.")
 	}
 
+	if setDefault {
```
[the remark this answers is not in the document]
mdrohmann: The `venv.Activate()` method here fails in case we already have an activated project.   So, I cannot go this route when I have an activated project and make it the default project.  But yes, you're right and this can be simplified.

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+
+func setDefaultProjectPath(d DefaultConfigurer, path string) {
+	d.Set("default_project_path", path)
+}
+
+type primable interface {
+	primer.Outputer
+}
+
+// gets the intended target for a shim
```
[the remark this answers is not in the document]
mdrohmann: I used this function to do some sanity check before rolling back the shims.  It should return the path that it wants to link to.

 But I think this makes only sense for the symlinks that I had implemented before, not for shims.  I think I will rewrite this function so it just returns whether the file is likely to be a shim.  So, we are less likely to remove files that are added for other reasons.  But of course, this is never perfect...

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// remove symlinks pointing to default project
+	files, err := ioutil.ReadDir(binDir)
+	if err != nil {
+		return err
+	}
+	for _, f := range files {
+		fn := f.Name()
+		target, err := shimTarget(fn)
+		if err != nil {
```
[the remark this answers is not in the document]
mdrohmann: Okay, I'm getting quite confused now.  Should the implementation allow to define parallel default projects?  Because I thought that you could only ever have one project be the default one ie., Python or Perl.

I added this logic and the `shimTarget` function to ensure that we are not deleting too much, but above I argued that it is probably cleaner for this iteration to just check if the file is probably a shim, and if that is the case, I remove it.  That is in line with your first comment here.  Would that still be okay?

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+			return exes, errs.Wrap(err, "Error while walking path")
+		}
+	}
+
+	return exes, nil
+}
+
+// UniqueExes filters the array of executables for those that would be selected by the command shell in case of a name collision
+func UniqueExes(exePaths []string, pathext string) ([]string, error) {
```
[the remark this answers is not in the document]
mdrohmann: That is correct to assume.

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+
+	err = ioutil.WriteFile(shimPath, []byte(shimStr), 0755)
+	if err != nil {
+		return errs.Wrap(err, "failed to write shim command %s", shimPath)
+	}
+	return nil
+}
+
+func needsRollback() bool {
+	return true
```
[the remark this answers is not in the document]
mdrohmann: Forgot to remove it, yes.

internal/runners/activate/activation.go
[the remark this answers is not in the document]
mdrohmann: I refactored this quite a bit, and I believe that it is clearer now.

internal/runners/activate/default.go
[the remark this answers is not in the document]
mdrohmann: I rewrote this function and added a doc-string.

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// gets the intended target for a shim
+func shimTarget(fn string) (string, error) {
+	contents, err := ioutil.ReadFile(fn)
+	if err != nil {
+		return "", err
+	}
+
+	targetRe := regexp.MustCompile("^ target: (.)$")
```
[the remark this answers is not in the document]
mdrohmann: That's basically what I am doing now.  And what I wanted to do before, but it was split up in two places, and the regexp was incorrect.  Should be fixed now.

internal/runners/activate/default.go
```diff
@@ -0,0 +1,269 @@
...
+		return err
+	}
+	for _, f := range files {
+		fn := f.Name()
+		target, err := shimTarget(fn)
+		if err != nil {
+			continue
+		}
+
+		// TODO: Decide if we need to do these checks...
```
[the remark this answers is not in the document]
mdrohmann: Changed the checks and moved them to `isShimAndTargetsDir`.  Looks reasonable to me now.

internal/runners/activate/activation.go
```diff
@@ -75,13 +75,27 @@ func activationLoop(out output.Outputer, subs subshell.SubShell, targetPath stri
...
-func activate(proj *project.Project, out output.Outputer, subs subshell.SubShell) (bool, error) {
+func activate(proj *project.Project, out output.Outputer, cfg DefaultConfigurer, subs subshell.SubShell, setDefault bool) (bool, error) {
 	projectfile.Reset()
 	venv := virtualenvironment.Get()
+
+	alreadyActivated := os.Getenv(constants.ActivatedStateEnvVarName) != ""
+
+	// if we are already activated and --default flag is set, just setup the default activation directory
+	if setDefault && alreadyActivated {
+		venv.ActivateRuntime()
```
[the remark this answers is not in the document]
mdrohmann: It is now called `venv.Setup()` after which the `virtualenvironment` is "set up" in such a way that we actually know what environment variables need to be set.

internal/runners/activate/default.go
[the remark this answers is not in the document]
Naatan: Yes it should allow parallel defaults for different languages. You are setting the default for the language of the project, not the default for any project ever.

I had talked about this while pointing and it's somewhat implied in the ACs, guess I could've made it more explicit.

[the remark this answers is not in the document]
mdrohmann: Yes, I remember it now.  Sorry, I forgot, and did not infer it from the ACs.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
+package defact
```
Naatan, line 1: `defact`?

Let's name this `globaldefault`. `default` feels to generic.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+	ext   string
+}
+
+func getDefaultProjectPath(d DefaultConfigurer) string {
+	return d.GetString("default_project_path")
+}
+
+func setDefaultProjectPath(d DefaultConfigurer, path string) {
+	d.Set("default_project_path", path)
+}
```
Naatan, line 41: Unused functions

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+
+type primable interface {
+	primer.Outputer
+}
+
+// isShimAndTargetsDir check if the specified filename is (probably) a shim and targets a file in the specified directory
+// This function is used during rollback to clean old shims
+func isShimAndTargetsDir(filename, dir string) bool {
```
Naatan, line 49: This took me multiple passes of the function and its contents to understand. The naming and description isn't explaining what's happening here very well.

Let's make this a lot clearer, name it something like

```
isShimFor(shimFile, runtimeDir string) bool
```

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+func setDefaultProjectPath(d DefaultConfigurer, path string) {
+	d.Set("default_project_path", path)
+}
+
+type primable interface {
+	primer.Outputer
+}
+
+// isShimAndTargetsDir check if the specified filename is (probably) a shim and targets a file in the specified directory
+// This function is used during rollback to clean old shims
```
Naatan, line 48: Let's rephrase this to something like

```
checks if the given shim points to the given runtime directory
```

The only undetermined part is whether it points to our runtime directory. We own the shims and the shim directory, so there are no "maybe's" there.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+func isShimAndTargetsDir(filename, dir string) bool {
+	contents, err := ioutil.ReadFile(filename)
+	if err != nil {
+		logging.Debug("Could not read file contents of shim candidate %s: %v", filename, err)
```
Naatan, line 52: This should be an error. We own the dir. Us not being able to read its files amounts to an error.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+		return false
+	}
+
+	targetRe := regexp.MustCompile("(?m)^(?:REM|#) State Tool Shim Target: (.*)$")
```
Naatan, line 56: This line is mentioned in 3 different places while being crucial to the execution of our code. Please make it a constant and use it in all 3 places.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+	target := targetRe.FindStringSubmatch(string(contents))
+
+	if len(target) != 2 {
+		return false
+	}
+
+	if dir == "" {
+		return true
+	}
```
Naatan, line 65: Are you using this function for two completely distinct use-cases? One to determine if a target is a shim and the other if the shim contains the dir we want? This is super confusing. Please keep it unixy; specialized functions that do one specific thing.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+
+	res, err := fileutils.PathContainsParent(target[1], dir)
```
Naatan, line 67: This doesn't make sense, we're just checking if the given path is in the cache dir? We should be checking if the path matched is the runtime directory that we are interested in. Whether it is in the cache dir is already determined by the fact that we know this is a shim and we know that the shim resides in our bin dir. We don't need an additional check to assert what we already know. We need to assert that this is a shim for our runtime dir.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	return res
+}
+
+// rollbackShims removes all shims in the global binary directory that target a previous default project
+// If the target of a shim does not exist anymore, the shim is also removed.
+func rollbackShims(cfg DefaultConfigurer) error {
+	binDir := config.GlobalBinPath()
+
+	// remove symlinks pointing to default project
```
Naatan, line 80: Let's stop calling them symlinks, they're shims.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+
+	for _, exe := range exeFiles {
+		result = append(result, exe.fpath)
+	}
+	return result, nil
+}
+
+// shimTargetPath returns the full path to the shim target (adds .bat on Windows)
+func shimTargetPath(targetDir string, path string) string {
+	target := filepath.Clean(filepath.Join(targetDir, filepath.Base(path)))
```
Naatan, line 168: Let's change `path` to `filename` and have the caller do `filepath.Base()`

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+	if rt.GOOS != "windows" {
+		return target
+	}
+
+	oldExt := filepath.Ext(target)
+	return target[0:len(target)-len(oldExt)] + ".bat"
```
Naatan, line 174: scriptfile and the language type should be the ones facilitating this

http://github.com/ActiveState/cli/blob/pkg-174895290/internal/scriptfile/scriptfile.go#L20-L20

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+	err = ioutil.WriteFile(shimPath, []byte(shimStr), 0755)
+	if err != nil {
+		return errs.Wrap(err, "failed to write shim command %s", shimPath)
+	}
+	return nil
+}
+
+// createShimFiles creates shims in the target path of all executables found in the bins dir
+// It overwrites existing files, if the overwrite flag is set.
+// On Windows the same executable name can have several file extensions,
```
Naatan, line 209: This is true on Mac/Linux too. `.sh` vs no extension.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+		shimPath := shimTargetPath(targetPath, exePath)
+
+		// The link should not exist as we are always rolling back old shims before we run this code.
+		if fileutils.TargetExists(shimPath) {
+			logging.Error("Cannot create shim as target already exists: %s.", shimPath)
+			continue
+		}
+
+		if err := createShimFile(exePath, shimPath); err != nil {
+			return err
```
Naatan, line 224: Wrap

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+// target (with the same or a different extension) from a different directory.
+// Also: Only the executable with the highest priority according to pathExt is shimmed.
+func createShimFiles(targetPath string, exePaths []string) error {
+	for _, exePath := range exePaths {
+		shimPath := shimTargetPath(targetPath, exePath)
+
+		// The link should not exist as we are always rolling back old shims before we run this code.
+		if fileutils.TargetExists(shimPath) {
+			logging.Error("Cannot create shim as target already exists: %s.", shimPath)
+			continue
```
Naatan, line 220: Return the error if we can't do the thing we were tasked to do.

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+
+		if err := createShimFile(exePath, shimPath); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// SetupDefaultActivation sets symlinks in the global bin directory to the currently activated runtime
```
Naatan, line 231: Not symlinks (I'll stop saying this now, please find/replace all).

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+func SetupDefaultActivation(cfg DefaultConfigurer, output output.Outputer, envGetter runtime.EnvGetter) error {
+	env, err := envGetter.GetEnv(false, "")
+	if err != nil {
+		return err
+	}
+
+	// roll back old symlinks
+	rollbackShims(cfg)
```
Naatan, line 239: No error check

internal/defact/default.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// Remove duplicate executables as per PATH and PATHEXT
+	exes, err = UniqueExes(exes, os.Getenv("PATHEXT"))
+	if err != nil {
+		return locale.WrapError(err, "err_unique_exes", "Could not detect unique executables, make sure your PATH and PATHEXT environment variables are properly configured.")
+	}
+
+	binDir := config.GlobalBinPath()
+
+	output.Print(locale.Tl("default_shim", "Writing default installation to {{.V0}}.", binDir))
```
Naatan, line 260: This should be in the runner logic, not here. I don't think we need to pass output into this package at all.

internal/runners/activate/activate.go
```diff
@@ -49,7 +53,28 @@ func (r *Activate) Run(params *ActivateParams) error {
...
-	logging.Debug("Activate %v, %v", params.Namespace, params.PreferredPath)
+	logging.Debug("Activate with namespace=%v, path=%v", params.Namespace, params.PreferredPath)
+
+	activeProjectDir := os.Getenv(constants.ActivatedStateEnvVarName)
+	alreadyActivated := activeProjectDir != ""
+
+	// handle case, if we are already activated
+	if alreadyActivated {
+		if params.Default && params.Namespace.IsValid() {
+			return locale.NewError("err_default_with_name_already_active", "Trying to set {{.V0}} as default project while in activated environment.  Please de-activate the current runtime.", params.Namespace.String())
```
Naatan, line 64: We should still set the default. This is a legitimate use-case we want to facilitate.

The AC's specifically state this:

> When called in an already activated environment, only set up the default (don't activate again or error out about already being activated)

internal/runners/activate/activate.go
```diff
@@ -49,7 +53,28 @@ func (r *Activate) Run(params *ActivateParams) error {
...
+		}
+		if !params.Default {
```
Naatan, line 66: Why would this only happen if not using Default?

internal/runners/activate/activation.go
```diff
@@ -75,19 +78,34 @@ func activationLoop(out output.Outputer, subs subshell.SubShell, targetPath stri
...
+
 	venv.OnDownloadArtifacts(func() { out.Notice(locale.T("downloading_artifacts")) })
 	venv.OnInstallArtifacts(func() { out.Notice(locale.T("installing_artifacts")) })
 	venv.OnUseCache(func() { out.Notice(locale.T("using_cached_env")) })
-	fail := venv.Activate()
-	if fail != nil {
-		return false, locale.WrapError(fail, "error_could_not_activate_venv", "Could not activate project. If this is a private project ensure that you are authenticated.")
+
+	logging.Debug("Setting up virtual Environment")
+	if strings.ToLower(os.Getenv(constants.DisableRuntime)) == "true" {
```
Naatan, line 94: This doesn't make any sense, you are skipping venv setup when in fact the constant only calls for skipping the runtime activation.

Naatan (changes requested):

internal/defact/default.go
mdrohmann: I don't think that you can type `script` and it would run `script.sh` instead.  The file endings in Mac/Linux are solely to give some orientation to the user.

internal/runners/activate/activate.go
mdrohmann: Yes, this only for the case where a name has been explicitly specified ie.,

You are in an activated state for project `testOrg/A` and then you run `state activate testOrg/B --default`.  As it is specifically requested, probably `testOrg/B` should have been selected, but after a few initial trials it became to confusing too me how to ensure that going forward we don't use `testOrg/A`.  As this, I think, is also a target of your refactoring, it will hopefully be easier after the refactoring.

If you just run `state activate --default` in an activated project, it would work as requested in the AC.

internal/runners/activate/activate.go
mdrohmann: This addresses the above AC:  If you run `state activate --default` it is never an error - even if already activated.  Otherwise, the code below will create the error message.

internal/runners/activate/activation.go
mdrohmann: Well, maybe the term `runtime activation` is well defined, but what we were and still are skipping, is the download of the artifacts and all parsing of the environment variables.  

That made sense to me, as the second part is only possible if the first one completed.  We could make the logic more complicated here if `DisableRuntime` is `"true"`), such that the environment variables are still parsed if they are available ie., if the artifact have been downloaded before. But that is basically the same behavior as `DiableRuntime=="false"` and the artifacts installed.  

I believe that this is correct `DisableRuntime==true` implies the disabling the parsing of the environment variables.  That's how it has been, BTW.  I tried to stay compatible here, not messing with existent logic...

internal/defact/default.go
mdrohmann: I think that it is always possible that the user adds their own binaries or other files there.  I guess it is not too bad to delete these files, as we "own" the directory, but I usually attempt to be "nice" in these cases. :)

internal/defact/default.go
mdrohmann: The error will trigger when the user adds a file that we do not have a write permissions for.  The behavior is okay, of course.  I'm just pointing it out here.

internal/defact/default.go
Naatan: I don't think being "nice" helps us or the user here. If the user goes off the reservation then stuff will break. We shouldn't try to facilitate unsupported use-cases.

internal/runners/activate/activate.go
```diff
@@ -1,116 +1,136 @@
...
 	if err != nil {
 		return locale.WrapError(err, "err_activate_projecttouse", "Could not figure out what project to use.")
 	}
 
 	// Run checkout if no project was given
-	if projectToUse == nil {
-		if !params.Namespace.IsValid() {
-			return locale.WrapError(err, "err_namespace_invalid", "Invalid namespace: {{.V0}}.", params.Namespace.String())
+	if proj == nil {
+		if params.Namespace == nil {
```
mdrohmann, line 78: This condition actually can never be true. It is always initialized: https://github.com/ActiveState/cli/blob/2b34f6b82e208668ea493c9a1fb432d7ffc0483a/cmd/state/internal/cmdtree/activate.go#L15
The `!params.Namespace.IsValid()` was correct.

internal/globaldefault/default.go
```diff
@@ -0,0 +1,190 @@
...
+		return locale.WrapError(err, "err_createshims", "Could not create shim files to set up the default runtime environment.")
+	}
+
+	cfg.Set(constants.GlobalDefaultPrefname, projectPath)
+
+	return nil
+}
+
+// cleanup removes all shims in the global binary directory that target a previous default project
+// If the target of a shim does not exist anymore, the shim is also removed.
```
mdrohmann, line 111: This comment should be removed.  The function now just removes all files in the directory.

internal/globaldefault/default.go
```diff
@@ -0,0 +1,190 @@
...
+		return errs.Wrap(err, "Could not read through global bin dir")
+	}
+	for _, f := range files {
+		if f.IsDir() {
+			continue
+		}
+
+		fn := filepath.Join(binDir, f.Name())
+
+		// remove shim if it links to old project path or target does not exist anymore
```
mdrohmann, line 130: same for this comment:  It does not reflect the reality anymore. :)

internal/runners/prepare/prepare.go
```diff
@@ -37,29 +34,18 @@ func New(prime primeable) *Prepare {
...
-	envUpdates := map[string]string{
-		"PATH": binDir,
-	}
-
-	fail = r.subshell.WriteUserEnv(envUpdates, sscommon.Default, true)
-	if fail != nil {
+	if err := globaldefault.Prepare(r.subshell); err != nil {
 		if runtime.GOOS != "linux" {
-			return locale.WrapError(fail.ToError(), "err_prepare_update_env", "Could not update user environment")
+			return locale.WrapError(err, "err_prepare_update_env", "Could not prepare environment.")
```
mdrohmann, line 39: Shouldn't that be handled in the `Prepare` function?  If this function errors, it really just means that the preparation step failed, I think.

pkg/platform/runtime/runtime.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// Retrieve artifact binary directory
+	var bins []string
+	if p, ok := envMap["PATH"]; ok {
+		bins = strings.Split(p, string(os.PathListSeparator))
+	}
+
+	result := []*language.Language{}
+	for _, bin := range bins {
```
mdrohmann, line 75: I still wish this could be more explicit, and not by searching through a bunch of executables. What do you think of storing this information when we download the artifacts?  Then we can extract them from the list of packages in the recipe by looking for ingredients in the "language" namespace. These hardcoded code bits are so difficult to maintain, and hopefully we'll add a lot of languages at some point.

mdrohmann (commented):
Nice cleanup.  It is getting better. :)

internal/runners/prepare/prepare.go
mdrohmann, line 39: The error key `err_prepare_update_env` has already been used inside of the `globaldefault.Prepare` function...

Naatan:
@mdrohmann ready for re-review

Naatan (approved):

PR #1004, opened 2020-09-30
State shim

MDrakos:
https://www.pivotaltracker.com/story/show/174994170


cmd/state/internal/cmdtree/shim.go
```diff
@@ -0,0 +1,43 @@
...
+	cmd := captain.NewCommand(
+		"shim",
+		locale.T("shim_description"),
+		[]*captain.Flag{
+			{
+				Name:        locale.T("flag_state_shim_language"),
+				Description: locale.T("flag_state_shim_language_description"),
+				Value:       &params.Language,
+			},
+		},
```
Naatan, line 23: The ACs don't call for any flags. And flags wouldn't work for this command anyway since we want to support `--`.

cmd/state/internal/cmdtree/shim.go
```diff
@@ -0,0 +1,43 @@
...
+		func(ccmd *captain.Command, args []string) error {
+			if len(args) > 0 {
+				args = args[1:]
+			}
+			return runner.Run(params, args...)
+		},
+	)
+	cmd.SetSkipUpdate(true)
+	cmd.SetSkipDeprecationCheck(true)
```
Naatan, line 40: This needs a `cmd.SetDisableFlagParsing(true)` so `--` can work (https://unix.stackexchange.com/a/11382)

internal/captain/command.go
```diff
@@ -172,6 +177,14 @@ func (c *Command) SetDeferAnalytics(value bool) {
 	c.deferAnalytics = value
 }
 
+func (c *Command) SetSkipUpdate(value bool) {
+	c.skipUpdate = value
+}
+
+func (c *Command) SetSkipDeprecationCheck(value bool) {
+	c.skipDeprecationCheck = value
+}
```
Naatan, line 186: Let's unify these under `SetSkipChecks`

internal/captain/command.go
```diff
@@ -218,6 +240,38 @@ func (c *Command) AddLegacyChildren(children ...cobraCommander) {
...
+
+func (c *Command) findNext(next string) *Command {
+	for _, cmd := range c.commands {
+		if cmd.cobra.Use == next {
+			return cmd
+		}
+	}
+	return nil
+}
```
Naatan, line 273: Why not use `cobra.Find`? This needs a better solution regardless, we're doing too much here.

cmd/state/main.go
```diff
@@ -165,7 +147,26 @@ func run(args []string, out output.Outputer) (int, error) {
 
 	// Run the actual command
 	cmds := cmdtree.New(primer.New(pj, out, authentication.Get(), prompter, sshell, conditional))
-	err = cmds.Execute(args[1:])
 
+	// Auto update to latest state tool version, only runs once per day
```
Naatan, line 151: Let's do the cmd check here. The AutoUpdate or Check code shouldn't need to have the command passed to them. (See my comment about unifying the checks also).

cmd/state/internal/cmdtree/shim.go
```diff
@@ -0,0 +1,43 @@
...
+			},
+		},
+		func(ccmd *captain.Command, args []string) error {
+			if len(args) > 0 {
+				args = args[1:]
+			}
+			return runner.Run(params, args...)
+		},
+	)
+	cmd.SetSkipUpdate(true)
```
Naatan, line 39: Should set this on the update command as well so we can get rid of http://github.com/ActiveState/cli/blob/201911db318226fdbbb9c81cfced9dcb321690d9/cmd/state/autoupdate.go#L23-L23

internal/path/path.go
```diff
@@ -0,0 +1,61 @@
+package path
```
Naatan, line 1: This feels more like `fileutils` to me.

internal/runners/shim/shim.go
```diff
@@ -0,0 +1,95 @@
...
+	} else {
+		for _, l := range project.Get().Languages() {
+			// Use first language found
+			lang = language.MakeByName(l.Name())
+			break
+		}
+	}
+	if !lang.Recognized() {
+		return locale.NewError("err_shim_language", "Unsupported language")
+	}
```
Naatan, line 53: Language is irrelevant. We're just inheriting the environment. Whether the user wants to run `perl` on that or `whoami` does not matter to us.

internal/runners/shim/shim.go
```diff
@@ -0,0 +1,95 @@
...
+	if !path.ProvidesExecutable("", lang.Executable().Name(), envPath) {
+		return locale.NewError("err_shim_exec", "Path does not contain: {{.V0}}", lang.Executable().Name())
+	}
+
+	scriptBlock, fail := fileutils.ReadFile(params.Script)
+	if fail != nil {
+		return locale.WrapError(fail.ToError(), "err_shim_read_script", "Could not read script file at: {{.V0}}", params.Script)
+	}
+
+	sf, fail := scriptfile.New(lang, filepath.Base(params.Script), string(scriptBlock))
```
Naatan, line 89: `lang` in this case would be `bash` or `batch` depending on what shell we're running in (ie. `batch` if on cmd, otherwise `bash`)

internal/runners/shim/shim.go
```diff
@@ -0,0 +1,95 @@
...
+	subshell subshell.SubShell
+}
+
+type primeable interface {
+	primer.Outputer
+	primer.Subsheller
+}
+
+type Params struct {
+	Script   string
```
Naatan, line 30:
```suggestion
	Command   string
```

Copy/paste much? ;)

test/integration/shim_int_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	scriptBlock := `
+print("Hello World!")
+	`
+
+	testScript := filepath.Join(fmt.Sprintf("%s/%s.pl", ts.Dirs.Work, suite.T().Name()))
+	fail := fileutils.WriteFile(testScript, []byte(scriptBlock))
+	suite.Require().NoError(fail.ToError())
+
+	cp := ts.Spawn("shim", "--language", "python3", fmt.Sprintf("%s", testScript))
```
Naatan, line 38: I'm not sure what went wrong here. The first AC says:

> `state shim -- perl foo.pl` runs `perl foo.pl` inside our runtime environment

But what you're testing here is

```
state shim --language python3 "Hello World!"
```

Please stick to the AC's.

Naatan: Here's some examples of things that should work with `state shim`, maybe that helps:

```
state shim -- whoami
state shim -- perl foo.pl
state shim -- perlcritic foo.pl
state shim -- find "foo" | grep bar
```

Note the double hyphen is crucial. You don't HAVE to use it, but we HAVE to support it.

Naatan (changes requested):

internal/captain/command.go
MDrakos: For the purpose of the skip checks we are setting properties on `captain.Command`. `cobra.Find` will return us a `cobra.Command` and we do not have a way to convert a cobra command to one of ours.

test/integration/shim_int_test.go
MDrakos: Thanks, the examples helped.

Naatan: It means we have to support both `state shim whoami` and `state shim -- whoami`. I believe it's simply a matter of setting `cmd.SetDisableFlagParsing(true)`

internal/captain/command.go
Naatan: I think a less involved and more scalable solution would be to keep a mapping of cobra <-> our commands and use that in tandem with `cobra.Find`

test/integration/shim_int_test.go
```diff
@@ -0,0 +1,121 @@
...
+print("Hello World!")
+	`
+
+	testScript := filepath.Join(fmt.Sprintf("%s/%s.pl", ts.Dirs.Work, suite.T().Name()))
+	fail := fileutils.WriteFile(testScript, []byte(scriptBlock))
+	suite.Require().NoError(fail.ToError())
+
+	cp := ts.Spawn("shim", "--", "python3", fmt.Sprintf("%s", testScript))
+	cp.Expect("Hello World!")
+	cp.ExpectExitCode(0)
```
Naatan, line 40: If this is working there's something wrong with the exit stuff. This is effectively running:

```shim -- python3 "print("Hello World!")"```

Note the double quotes.

**Review by Naatan — CHANGES_REQUESTED** [review]

File: test/integration/shim_int_test.go [hunk 13]
```diff
@@ -0,0 +1,121 @@
...
+	fail := fileutils.WriteFile(testScript, []byte(scriptBlock))
+	suite.Require().NoError(fail.ToError())
+
+	cp := ts.Spawn("shim", "--", "python3", fmt.Sprintf("%s", testScript))
+	cp.SendLine("ActiveState")
+	cp.Expect("Hello ActiveState!")
+	cp.ExpectExitCode(0)
+}
+
+func TestShimIntegrationTestSuite(t *testing.T) {
```
Naatan, line 119: I don't think any of our tests are actually asserting whether the environment is properly shimmed. ie. they'd be successful if someone already had python3 installed that isn't provided by us.

test/integration/shim_int_test.go
MDrakos: Turns out we have to leave the flag parsing in order for cobra to handle this

test/integration/shim_int_test.go
MDrakos: Added a test that will only run if there is already an existing python3 installation

test/integration/shim_int_test.go
MDrakos: I'm not sure I'm following. `testScript` is the path of the file so this should be running 

`state shim -- python3 /path/to/script.py`

And the subsequent shell script that the subshell runs should be 

`python3 /path/to/script.py`

The contents of the script file should not be read and passed as a string.

cmd/state/main.go
```diff
@@ -147,12 +148,14 @@ func run(args []string, out output.Outputer) (int, error) {
 	// Run the actual command
 	cmds := cmdtree.New(primer.New(pj, out, authentication.Get(), prompter, sshell, conditional))
 
-	var skipChecks bool
-	child := cmds.Command().Find(args)
-	if child != nil {
-		skipChecks = child.SkipChecks()
+	var child *captain.Command
+	if len(args) != 0 {
```
Naatan, line 152: `args` will never be 0, since `args[0]` is always the executable that was invoked.

internal/captain/command.go
```diff
@@ -231,27 +234,12 @@ func (c *Command) AddLegacyChildren(children ...cobraCommander) {
...
-		return c
+func (c *Command) Find(args []string) (*Command, error) {
+	foundCobra, _, err := c.cobra.Find(args)
+	if err != nil {
+		return nil, err
 	}
-
-	found, _ := innerFind(c, args[1:])
-	return found
+	return cobraMapping[foundCobra], nil
```
Naatan, line 242: Should still check if the mapping exists.

internal/captain/command.go
```diff
@@ -231,27 +234,12 @@ func (c *Command) AddLegacyChildren(children ...cobraCommander) {
...
-		}
-		return innerCmd, innerArgs
-	}
-
-	if len(args) < 2 {
-		return c
+func (c *Command) Find(args []string) (*Command, error) {
+	foundCobra, _, err := c.cobra.Find(args)
+	if err != nil {
+		return nil, err
```
Naatan, line 240: Wrap

internal/runners/shim/shim.go
```diff
@@ -54,14 +59,16 @@ func (s *Shim) Run(args ...string) error {
 	}
 
 	lang := language.Bash
+	scriptArgs := fmt.Sprintf("%s \"$@\"", args[0])
```
Naatan, line 62:
```suggestion
	scriptArgs := fmt.Sprintf(`%s "$@"`, args[0])
```

test/integration/shim_int_test.go
```diff
@@ -31,11 +32,14 @@ func (suite *ShimIntegrationTestSuite) TestShim_Basic() {
...
 
-	testScript := filepath.Join(fmt.Sprintf("%s/%s.pl", ts.Dirs.Work, suite.T().Name()))
+	testScript := filepath.Join(fmt.Sprintf("%s/%s.py", ts.Dirs.Work, suite.T().Name()))
 	fail := fileutils.WriteFile(testScript, []byte(scriptBlock))
 	suite.Require().NoError(fail.ToError())
 
-	cp := ts.Spawn("shim", "--", "python3", fmt.Sprintf("%s", testScript))
+	cp := ts.SpawnWithOpts(
+		e2e.WithArgs("shim", "--", "python3", fmt.Sprintf("%s", testScript)),
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
```
Naatan, line 41: Wonder if we can make these tests a bit cheaper by disabling runtime and just asserting that the environment was sourced some other way (eg. just check `ACTIVESTATE_ACTIVATED_ID` or something -- and ensure its value is different from the one that was passed to the current process that's running the tests)

test/integration/shim_int_test.go
```diff
@@ -106,16 +116,41 @@ name = input("Enter your name: ")
...
+	suite.Require().NoError(fail.ToError())
+
+	cp := ts.Spawn("shim", "--", "python3", testScript)
+	cp.Expect("Hello World!")
+	cp.ExpectExitCode(0)
+}
+
 func TestShimIntegrationTestSuite(t *testing.T) {
 	suite.Run(t, new(ShimIntegrationTestSuite))
 }
```
Naatan, line 156: Can we add an integration test that tests the variant without `--`. eg. `state shim python3 foo.py`

Naatan (changes requested):

cmd/state/internal/cmdtree/shim.go
```diff
@@ -0,0 +1,30 @@
...
+		[]*captain.Flag{},
+		[]*captain.Argument{},
+		func(ccmd *captain.Command, args []string) error {
+			return runner.Run(args...)
+		},
+	)
+	cmd.SetSkipChecks(true)
+
+	if !strings.Contains(strings.Join(prime.Args(), " "), " -- ") {
+		cmd.SetDisableFlagParsing(true)
```
MDrakos, line 27: This changed was prompted by running `state shim python -c ...`

If we leave flag parsing on Cobra will handle the `--` correctly, however since we have to support `state shim` without it we need to disable flag parsing when the delimiter isn't present so Cobra doesn't try to interpret flags for the shimmed command

internal/primer/primer.go
```diff
@@ -18,15 +18,17 @@ type Values struct {
 	prompt      prompt.Prompter
 	subshell    subshell.SubShell
 	conditional *constraints.Conditional
+	args        []string
 }
 
-func New(project *project.Project, output output.Outputer, auth *authentication.Auth, prompt prompt.Prompter, subshell subshell.SubShell, conditional *constraints.Conditional) *Values {
+func New(project *project.Project, output output.Outputer, auth *authentication.Auth, prompt prompt.Prompter, subshell subshell.SubShell, conditional *constraints.Conditional, args []string) *Values {
```
Naatan, line 24: Primer is for passing persistent variables around. It's not a catchall for passing arguments. You shouldn't need to pass args this way at all, captain/cobra already has access to this. It's already part of the architecture.

internal/captain/command.go
```diff
@@ -237,9 +237,12 @@ func (c *Command) AddLegacyChildren(children ...cobraCommander) {
 func (c *Command) Find(args []string) (*Command, error) {
 	foundCobra, _, err := c.cobra.Find(args)
 	if err != nil {
-		return nil, err
+		return nil, locale.WrapError(err, "err_captain_cobra_find", fmt.Sprintf("Could not find child Command with args: %s", args))
```
Naatan, line 240: `fmt.Sprintf` is working around the locale mechanic. This should be something like this:

```
return nil, locale.WrapError(err, "err_captain_cobra_find", "Could not find child Command with args: {{.V0}}", args)
```

But this probably shouldn't be a localized error at all.

internal/captain/command.go
```diff
@@ -237,9 +237,12 @@ func (c *Command) AddLegacyChildren(children ...cobraCommander) {
...
 	}
-	return cobraMapping[foundCobra], nil
+	if cmd, ok := cobraMapping[foundCobra]; ok {
+		return cmd, nil
+	}
+	return nil, locale.NewError("err_captain_cmd_find", fmt.Sprintf("Could not find child Command with args: %s", args))
```
Naatan, line 245: Pass arguments via locale, not Sprintf

Naatan (changes requested):

internal/primer/primer.go
MDrakos: The problem with accessing the args from captain/cobra is that by that time the flag are already parsed. I've removed the args from primer and moved the delimiter check.

cmd/state/main.go
```diff
@@ -145,13 +146,21 @@ func run(args []string, out output.Outputer) (int, error) {
...
 		logging.Debug("Could not find child command, error: %v", err)
 	}
 
+	// Cobra will handle the `--` delimiter if flag parsing is enabled.
+	// If the delimeter is not present we have to disable flag parsing
+	// to ensure flags are passed to the shimmed command rather than
+	// parsed as a flag for `state shim`
+	if !strings.Contains(strings.Join(args, " "), " -- ") {
+		child.SetDisableFlagParsing(true)
+	}
```
Naatan: We can still do this inside cmdtree, just don't pass the arg as part of `prime`. Having it in here does feel dirty.

Naatan (changes requested):

cmd/state/main.go
MDrakos: It appears that cobra will use `os.Args` if `cmd.SetArgs` is not called on the command. After this it does a lot of parsing and processing to them. The args I'm using here are before all of the processing so I can inform Cobra on how to process them.

internal/captain/command.go
Naatan: Why'd you remove this?

Naatan (changes requested):

internal/captain/command.go
MDrakos: Because of this https://github.com/ActiveState/cli/pull/1004#discussion_r500563763. Specifically:
> But this probably shouldn't be a localized error at all.

Is that not what you meant?

Naatan: That doesn't mean it shouldn't be wrapped, it just means to use `errs.Wrap` instead of `locale.WrapError`

internal/captain/command.go
MDrakos: Makes sense, it's wrapped now.

Naatan (approved):

PR #1005, opened 2020-09-30
compute cid based on MAC address and Windows Product Key

mdrohmann:
https://www.pivotaltracker.com/story/show/174959900

### DIFF
--- a/installers/msi-language/GAPixel/GAPixelCA.cs
+++ b/installers/msi-language/GAPixel/GAPixelCA.cs
@@ -21,12 +21,74 @@ namespace GAPixel
             }
         }
 
+        public static string DecodeProductKeyWin8AndUp(byte[] digitalProductId)
+        {
+            var key = String.Empty;
+            const int keyOffset = 52;
+            var isWin8 = (byte)((digitalProductId[66] / 6) & 1);
+            digitalProductId[66] = (byte)((digitalProductId[66] & 0xf7) | (isWin8 & 2) * 4);
+
+            // Possible alpha-numeric characters in product key.
+            const string digits = "BCDFGHJKMPQRTVWXY2346789";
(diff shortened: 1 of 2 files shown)


installers/msi-language/GAPixel/GAPixelCA.cs
```diff
@@ -5,59 +5,47 @@
...
-        public static string GetUniqueId(Session session=null)
+        public static string GetOrCreateNewCid(Session session)
         {
             try
             {
-                var oMClass = new ManagementClass("Win32_NetworkAdapterConfiguration");
-                var colMObj = oMClass.GetInstances();
-                foreach (var objMO in colMObj)
+                var baseKey = Registry.CurrentUser;
+                if (session.GetMode(InstallRunMode.Scheduled) && session.CustomActionData.ContainsKey("USERSID"))
```
MDrakos, line 19: This will only get the registry key if this method is called from a deferred custom action. I _think_  this can be called from an immediate custom action via the `GAReport` CAs. If that's the case we can still try to get the registry key here we just have to use `Registry.CurrentUser`

MDrakos (changes requested):

installers/msi-language/GAPixel/GAPixelCA.cs
mdrohmann: In the non-scheduled (immediate?) mode, `baseKey` is actually set to `Registry.CurrentUser` from which it then goes to the sub-key.  So, if I understand correctly, it is already implemented the way you describe it.

MDrakos: Oh I see what I missed now. 👍

MDrakos (approved):

PR #1006, opened 2020-10-01
Ensure activate uses project.GetSafe by default

daved:
https://www.pivotaltracker.com/story/show/175082480

### DIFF
--- a/internal/runners/activate/activate.go
+++ b/internal/runners/activate/activate.go
@@ -9,7 +9,6 @@ import (
 	"github.com/ActiveState/cli/internal/failures"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/subshell"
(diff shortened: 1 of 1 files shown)


internal/runners/activate/activate.go
```diff
@@ -97,25 +96,27 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
...
-		targetPath, err = osutils.Getwd()
+		proj, fail = project.GetSafe()
 	}
-	if err != nil {
-		return "", err
-	}
-
-	proj, fail := project.FromPath(targetPath)
 	if fail != nil {
 		return targetPath, fail
```
daved, line 121: I think this line causes some errors to be missed. The correct behavior seems to be occurring, so I didn't attempt any debugging. If `return targetPath, fail` is exchanged with `return targetPath, fail.ToError()`, the outcome was different. This might be due to me misunderstanding the ToError/fail/error interactions.

Naatan: Please feel free to change it to `ToError()`, you're right this looks bugged.

Naatan (changes requested):
It's super unclear to me what this code is fixing and how these changes have an effect on the global project mechanic. Please explain the problem a little and either comment the code so it's clear why things are happening this way, or find a better way to address the underlying issue.

I feel like this fix is working around something rather than actually fixing it.

internal/runners/activate/activate.go
```diff
@@ -97,27 +96,29 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
...
 	switch {
 	// Checkout via namespace (eg. state activate org/project) and set resulting path
 	case namespace != "":
-		targetPath, err = r.namespaceSelect.Run(namespace, preferredPath)
+		namesPath, err := r.namespaceSelect.Run(namespace, preferredPath)
+		if err != nil {
+			return "", err
+		}
+		proj, fail = project.FromPath(namesPath)
+		targetPath = namesPath
```
Naatan, line 111: It doesn't look like we actually still need `targetPath`?

Naatan (changes requested):

Naatan (approved):

PR #1008, opened 2020-10-02
Wait for build to complete

Naatan:
https://www.pivotaltracker.com/story/show/174895290


pkg/platform/api/buildlogstream/streamer.go
```diff
@@ -0,0 +1,155 @@
...
+		err = connectErr
+	}
+
+	logging.Debug("connecting to head-chef websocket")
+	r.socket.Connect()
+	if err != nil {
+		return errs.Wrap(err, "Could not connect to websocket")
+	}
+
+	time.Sleep(time.Second)
```
daved, line 78: Sleeps like this are a bright redish-orange flag for me. Martin's concern about not using OnConnect seem quite relevant; The control flow is not as direct as it appears.

Without spending more time, I'm not exactly sure what `gowebsocket` is really offering us here since it's using `gorilla/websocket` underneath. Nevertheless, I think that layer is hiding the control flow and making it a bit more difficult to rationalize the control flow.

Naatan: This is WIP code, not ready for any sort of review. The sleep was to test a race condition.

daved: Understood... still, not using OnConnect and SendBinary/SendText would mean that it's possible to be interacting with connection in a poor state.

pkg/platform/model/recipe.go
```diff
@@ -151,9 +165,6 @@ func fetchRecipeID(commitID strfmt.UUID, owner, project, orgID string, private b
 	if err != nil {
 		return nil, FailOrderRecipes.Wrap(err)
 	}
-	if private {
```
mdrohmann, line 154: Why is that not needed anymore?  Can we then also remove the arguments `private` and `orgID`?

mdrohmann (changes requested):

mdrohmann:
Looks good to me.  Just had a couple questions about some of the logic...

pkg/platform/api/buildlogstream/streamer.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+
+func (r *Request) responseReader(conn *websocket.Conn, readErr chan error) {
+	artifactMap, err := artifactMap(r.recipeID)
+	if err != nil {
+		readErr <- errs.Wrap(err, "Could not generate artifact map")
+	}
+
+	total := len(artifactMap)
```
mdrohmann, line 83: I noticed the AC: 
> Exclude artifacts with namespace builder and builder-lib from the progress indicator

That will need to be implemented still.

internal/runners/packages/remove.go
```diff
@@ -37,16 +37,13 @@ func (r *Remove) Run(params RemoveRunParams) error {
 
 	// Commit the package
 	pj := project.Get()
-	fail = model.CommitPackage(pj.Owner(), pj.Name(), model.OperationRemoved, params.Name, "")
+	_, fail = model.CommitPackage(pj.Owner(), pj.Name(), model.OperationRemoved, params.Name, "")
```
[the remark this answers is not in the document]
Naatan: I wrongly assumed this was running the same logic as add/update. Will fix, thanks!

pkg/platform/api/buildlogstream/streamer.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+
+	return artifactMap, nil
+}
+
+func artifactDescription(artifactID strfmt.UUID, artifactMap map[strfmt.UUID]artifactMapping) string {
+	v, ok := artifactMap[artifactID]
+	if !ok || v.Name == nil {
+		return locale.Tl("unknown_artifact_description", "Artifact {{.V0}}", artifactID.String())
```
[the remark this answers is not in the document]
Naatan: No since this is a legitimate use-case that in my tests happens at least once per invocation. I've told Stephen about this, I think he's going to work on implementing a name for this use-case.

pkg/platform/model/recipe.go
Naatan: They removed it from the inventory API. I'll follow up on the greater refactoring in another PR.

pkg/platform/api/buildlogstream/streamer.go
Naatan: That's here: https://github.com/ActiveState/cli/pull/1008/files#diff-4bf0b2d5fe8f37d210ff6cd68690150eR132

Though I see now that I missed `builder-lib`. Will add that.

mdrohmann (approved):

PR #1010, opened 2020-10-02
Refactor projectfile  detection to fix bugs and be less error prone

Naatan:
https://www.pivotaltracker.com/story/show/175098414

### DIFF
--- a/internal/runners/activate/activate.go
+++ b/internal/runners/activate/activate.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ActiveState/cli/internal/subshell"
 	"github.com/ActiveState/cli/internal/virtualenvironment"
 	"github.com/ActiveState/cli/pkg/project"
+	"github.com/ActiveState/cli/pkg/projectfile"
 )
 
 type Activate struct {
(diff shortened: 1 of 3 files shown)


internal/runners/activate/activate.go
```diff
@@ -90,33 +110,29 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
...
+		return preferredPath, nil
 	default:
-		proj, fail = project.GetSafe()
-	}
-	if fail != nil {
-		return "", fail.ToError()
+		// Get path from working directory
+		var fail *failures.Failure
+		targetPath, fail := projectfile.GetProjectFilePath()
+		return filepath.Dir(targetPath), fail.ToError()
```
daved, line 128: An error return along with useful value is quite abnormal. This would usually be solved by returning the useful value as part of a specific error return so that callers know they are in a special condition. Barring that for whatever reason, I suggest including a comment on the function clarifying the unusual behavior.

If the return here is misleading me, I suggest explicitly returning empty string w/error or processed string with nil.

daved (changes requested):

pkg/projectfile/projectfile.go
```diff
@@ -656,35 +656,56 @@ func setCommitInYAML(data []byte, commitID string) ([]byte, *failures.Failure) {
 
 // GetProjectFilePath returns the path to the project activestate.yaml
 func GetProjectFilePath() (string, *failures.Failure) {
+	lookup := []func() (string, *failures.Failure){
+		getProjectFilePathFromEnv,
+		getProjectFilePathFromWd,
+		getProjectFilePathFromDefault,
+	}
```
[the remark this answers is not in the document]
daved: *it's going to get called at least twice in succession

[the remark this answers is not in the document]
Naatan: It costs us nothing. This way it's conveniently in one spot and doesn't require the reader to jump through the file any more than they already do.

internal/runners/activate/activate.go
Naatan: This function has no opinion on whether the value is valid, it's just returning the error from `GetProjectFilePath()`. The fail.ToError() stuff is just obscuring things, but that will go away when we get rid of failures.

I don't think this needs to be any more verbose than it already is. Once failures go away this'll be nice and neat, until then it seems like a pretty minor stumbling block.

daved: My only significant concern here is clarifying if callers are to be interacting with the useful value when there is also an error return. It's clean and readable enough otherwise, IMO.

Naatan: Error always takes precedence. Callers don't need any special logic.

daved (approved):

PR #1011, opened 2020-10-05
State packages bundles support

MDrakos:
https://www.pivotaltracker.com/story/show/174918218


internal/runners/packages/remove.go
```diff
@@ -37,7 +37,17 @@ func (r *Remove) Run(params RemoveRunParams) error {
 
 	// Commit the package
 	pj := project.Get()
-	fail = model.CommitPackage(pj.Owner(), pj.Name(), model.OperationRemoved, params.Name, "")
+	language, fail := model.DefaultLanguageNameForProject(pj.Owner(), pj.Name())
+	if fail != nil {
+		return fail.WithDescription("err_fetch_languages")
```
Naatan, line 42: This function returns errors, not failures.

internal/runners/packages/remove.go
```diff
@@ -37,7 +37,17 @@ func (r *Remove) Run(params RemoveRunParams) error {
...
+	}
+
+	ingredient, err := model.IngredientWithLatestVersion(language, params.Name)
+	if err != nil {
+		return locale.WrapError(err, "err_remove_get_package", "Could not find package to remove")
+	}
+
+	fail = model.CommitPackage(pj.Owner(), pj.Name(), model.OperationRemoved, params.Name, ingredient.Namespace, "")
 	if fail != nil {
 		return fail.WithDescription("err_package_removed")
```
Naatan, line 52: Not part of your PR, but may as well fix, this should be an error.

Naatan: There's one of these on line 35 as well. Can you fix that one too? They just need a `.ToError()`

pkg/platform/model/vcs.go
```diff
@@ -64,8 +64,11 @@ const (
 	// NamespaceLanguageMatch is the namespace used for language requirements
 	NamespaceLanguageMatch = `^language$`
 
-	// NamespacePackageMatch is the namespace used for package requirements
-	NamespacePackageMatch = `^language\/\w+$`
+	// NamespaceLanguagePackageMatch is the namespace used for language package requirements
+	NamespaceLanguagePackageMatch = `^language\/\w+$`
```
Naatan, line 68: Please undo. This is matching a package namespace, adding language to the name just obscures things

pkg/platform/model/vcs.go
```diff
@@ -64,8 +64,11 @@ const (
...
+
+	// NamespaceBundlesPackageMatch is the namespace used for bundle package requirements
+	NamespaceBundlesPackageMatch = `^bundles\/\w+$`
```
Naatan, line 71: This should be `NamespaceBundlesMatch`. Bundles aren't packages.

pkg/platform/model/vcs.go
```diff
@@ -87,8 +90,8 @@ func NamespaceMatch(query string, namespace NamespaceMatchable) bool {
 type Namespace string
 
 // NamespacePackage creates a new package namespace
-func NamespacePackage(language string) Namespace {
-	return Namespace(fmt.Sprintf("language/%s", language))
+func NamespacePackage(namespace, language string) Namespace {
+	return Namespace(fmt.Sprintf("%s/%s", namespace, language))
```
Naatan, line 94: This makes no sense, this function is meant to explicitly construct a package namespace. Making the namespace configurable breaks that contract and renders this function pointless. I think what you want is a second function `NamespaceBundle`. These two concepts are not married.

Naatan (changes requested):

pkg/platform/model/vcs.go
MDrakos: That's good to know. Not obvious since `state packages` is now handling both and neither of the namespaces have any mention of `package`

pkg/platform/model/vcs.go
MDrakos: This stemmed from not knowing that the `bundles` namespace does not contain packages whereas the `languages` namespace does.

internal/runners/packages/remove.go
```diff
@@ -32,14 +32,14 @@ func NewRemove(prime primeable) *Remove {
...
-		return fail.WithDescription("err_activate_auth_required")
+		return fail.WithDescription("err_activate_auth_required").ToError()
 	}
 
 	// Commit the package
 	pj := project.Get()
 	language, fail := model.DefaultLanguageNameForProject(pj.Owner(), pj.Name())
 	if fail != nil {
-		return fail.WithDescription("err_fetch_languages")
+		return locale.WrapError(fail.ToError(), "err_fetch_languages")
```
Naatan, line 42: fwiw you don't need to use `ToError()` when wrapping, the wrapper handles that.

pkg/platform/model/inventory.go
```diff
@@ -28,11 +29,22 @@ var (
...
+// Platform is a sane version of whatever the hell it is go-swagger thinks it's doing
 type Platform = inventory_models.V1Platform
 
+const (
+	// PackageNamespacePrefix is the namespace prefix for packages
+	PackageNamespacePrefix = "language"
+
+	// BundlesNamespacePrefix is the namespace prefix for bundles
+	BundlesNamespacePrefix = "bundles"
+)
```
Naatan, line 46: Can we put this in http://github.com/ActiveState/cli/blob/84104c044fd08c96b63c769ab9b657879e26901a/pkg/platform/model/vcs.go#L1-L1

So it's nice and central

Naatan (changes requested):

Naatan (approved):

PR #1012, opened 2020-10-05
Start commit-only ASY project handling

daved:
https://www.pivotaltracker.com/story/show/175018982


pkg/platform/runtime/download.go
```diff
@@ -121,27 +121,39 @@ func (r *Download) fetchRecipeID() (strfmt.UUID, *failures.Failure) {
...
+	var recipeID strfmt.UUID
+	orgID := strfmt.UUID("00000000-0000-0000-0000-000000000000")
+	projectID := strfmt.UUID("00000000-0000-0000-0000-000000000000")
+	var fail *failures.Failure
 
-	recipeID, fail := r.fetchRecipeID()
-	if fail != nil {
-		return nil, fail
-	}
+	if r.owner != "" || r.projectName != "" {
```
Naatan, line 130: This looks like it was bugged before. If a commitID is given then it should override the owner/projectName.

pkg/project/project.go
```diff
@@ -232,6 +240,10 @@ func (p *Project) CommitUUID() strfmt.UUID {
 	return strfmt.UUID(p.commitID)
 }
 
+func (p *Project) HasPlatformPath() bool {
```
Naatan, line 243:
```suggestion
func (p *Project) IsHeadless() bool {
```

Naatan (commented):
Looks ok to me so far, but let me know if you encounter any rabbit holes. If this is all it takes then I'd expect this story to land by the end of the day.

pkg/platform/runtime/download.go
```diff
@@ -123,33 +123,47 @@ func (r *Download) fetchRecipeID() (strfmt.UUID, *failures.Failure) {
...
 // The first return argument specifies whether we are dealing with an alternative build
 func (r *Download) FetchArtifacts() (*FetchArtifactsResult, *failures.Failure) {
-	platProject, fail := model.FetchProjectByName(r.owner, r.projectName)
-	if fail != nil {
-		return nil, fail
-	}
-	r.orgID = platProject.OrganizationID.String()
-	r.private = platProject.Private
+	commitID := strfmt.UUID(constants.ValidZeroUUID)
+	recipeID := strfmt.UUID(constants.ValidZeroUUID)
```
Naatan, line 127: recipeID can never be zero

pkg/platform/runtime/download.go
```diff
@@ -123,33 +123,47 @@ func (r *Download) fetchRecipeID() (strfmt.UUID, *failures.Failure) {
...
+		for _, branch := range platProject.Branches {
+			if branch.Default {
+				branchCommitID = branch.CommitID
+				break
+			}
+		}
+		if branchCommitID == nil {
+			return nil, FailNoCommitID.New("fetch_err_runtime_no_commitid")
 		}
+		commitID = *branchCommitID
```
Naatan, line 155: I just realized this code is bugged / out of date (as in, the code you're updating). There is no scenario in which we don't feed the commitID into the runtime downloader. So if the commitID was not given this would effectively amount to an error.

You can drop this whole chunk in favour of line 133.

pkg/project/project.go
```diff
@@ -232,6 +240,10 @@ func (p *Project) CommitUUID() strfmt.UUID {
 	return strfmt.UUID(p.commitID)
 }
 
+func (p *Project) IsHeadless() bool {
+	return p.commitID != "" && (p.name == "" || p.owner == "")
```
Naatan, line 244: Let's instead just check if the url contains `/commit/` (ie. match the CommitURLRe). Technically this logic here is sound but I prefer we be a bit more explicit.

Naatan (changes requested):

pkg/platform/runtime/download.go
daved: I dropped most of this function. Even setting the orgID and projectID may be useless, but I suspect it would be worth passing that into the build request if we can.

Naatan (approved):

PR #1013, opened 2020-10-06
State packages add/update/remove can enter headless-commit state

mdrohmann:
https://www.pivotaltracker.com/story/show/175019250


internal/runners/packages/packages.go
```diff
@@ -3,26 +3,44 @@ package packages
...
+	if !authentication.Get().Authenticated() {
+		anonymousOk, fail := prompt.Confirm(locale.T("prompt_headless_anonymous"), true)
+		if fail != nil {
+			// TODO: Maybe ignore on interrupt?
+			return errs.Wrap(fail.ToError(), "Error prompting to proceed anonymously during headless commit.")
+		}
+		isHeadless = anonymousOk
+	}
+
+	// Note: User also lands here if answering No to the question about anonymous commit.
```
mdrohmann, line 37: This was not explicitly mentioned in the ACs.  Alternatively we could do nothing here and write an error message.

Naatan:
@mdrohmann please merge master

pkg/project/project.go
```diff
@@ -236,8 +236,17 @@ func (p *Project) CommitID() string {
...
+	if ok := strfmt.Default.Validates("uuid", p.commitID); !ok {
+		return "", locale.NewError("invalid_uuid_val", "Invalid commit ID {{.V0}} in activestate.yaml.  You could replace it with 'latest'", p.commitID)
+	}
+
+	var uuid strfmt.UUID
+	if err := uuid.UnmarshalText([]byte(p.commitID)); err != nil {
+		return "", locale.WrapError(err, "err_commit_id_unmarshal", "Failed to unmarshal the commit id {{.V0}} read from activestate.yaml.", p.commitID)
+	}
+
+	return uuid, nil
```
Naatan, line 249: Let's do this parsing in projectfile.Parse(). So this function doesn't need to return any error.

internal/runners/packages/packages.go
```diff
@@ -3,49 +3,79 @@ package packages
...
 )
 
 const latestVersion = "latest"
 
 func executePackageOperation(out output.Outputer, prompt prompt.Prompter, language, name, version string, operation model.Operation) error {
-	fail := auth.RequireAuthentication(locale.T("auth_required_activate"), out, prompt)
-	if fail != nil {
-		return fail.WithDescription("err_activate_auth_required")
+	isHeadless := false
+	if !authentication.Get().Authenticated() {
```
Naatan, line 24: This should be passed in via the primer

internal/runners/packages/packages.go
```diff
@@ -3,49 +3,79 @@ package packages
...
+		anonymousOk, fail := prompt.Confirm(locale.T("prompt_headless_anonymous"), true)
+		if fail != nil {
+			return locale.WrapInputError(fail.ToError(), "Failed to determine if it is okay to proceed without authorization.")
```
Naatan, line 27:
```suggestion
			return locale.WrapInputError(fail.ToError(), "Authentication cancelled.")
```

internal/runners/packages/packages.go
```diff
@@ -3,49 +3,79 @@ package packages
...
+		commitID, fail = model.CommitPackage(parentCommitID, operation, name, ingredientNamespace, version)
+		if fail != nil {
+			return locale.WrapError(fail.ToError(), "err_package_"+string(operation))
+		}
+	} else {
+		var fail *failures.Failure
+		commitID, fail = model.CommitPackageInBranch(pj.Owner(), pj.Name(), operation, ingredientNamespace, name, version)
+		if fail != nil {
+			return fail.WithDescription("err_package_" + string(operation)).ToError()
+		}
```
Naatan, line 78: Let's clean this up. We should have one model function that creates the commit, and another model function that updates the branch. Both should be called from here, but updating the branch should only be called if `!isHeadless`.

internal/runners/packages/packages.go
```diff
@@ -3,49 +3,79 @@ package packages
...
 	}
 
 	if strings.ToLower(version) == latestVersion {
 		version = ""
 	}
 
 	// Verify that the provided package actually exists (the vcs API doesn't care)
-	var ingredient *model.IngredientAndVersion
+	ingredientNamespace := ""
```
Naatan, line 45: There's no situation in which a namespace should ever be empty

internal/runners/packages/remove.go
```diff
@@ -33,23 +32,12 @@ func NewRemove(prime primeable) *Remove {
...
-
-	// Commit the package
-	pj := project.Get()
-	language, fail := model.DefaultLanguageNameForProject(pj.Owner(), pj.Name())
-	if fail != nil {
-		return locale.WrapError(fail, "err_fetch_languages")
-	}
-
-	return executePackageOperation(r.out, r.Prompter, language, params.Name, "", model.OperationRemoved)
+	return executePackageOperation(r.out, r.Prompter, "", params.Name, "", model.OperationRemoved)
```
Naatan, line 42: Language can't be empty.. ?

locale/en-us.yaml
```diff
@@ -1248,6 +1248,14 @@ package_version_updated:
   other: "Package updated: {{.V0}}@{{.V1}}"
 package_ingredient_err:
   other: Failed to resolve an ingredient named {{.V0}}.
+package_headless_added:
+  other: "Package added: {{.V0}}"
+package_headless_updated:
+  other: "Package updated: {{.V0}}"
+package_headless_removed:
+  other: "Package removed: {{.V0}}"
```
Naatan, line 1256: Are these still correct with us relying on the swagger?

Naatan (changes requested):

internal/runners/packages/remove.go
mdrohmann: For remove, we do not need to know the language.

locale/en-us.yaml
mdrohmann: Confirmed: https://github.com/ActiveState/cli/blob/master/pkg/platform/api/mono/mono_models/commit_change_editable.go#L72-L78

internal/runners/packages/packages.go
mdrohmann: Hmm... It turns out that we did not need it when I started working on this: https://github.com/ActiveState/cli/blob/9cfa094060a8cb33ea629bbf24b25b49c1d7204e/internal/runners/packages/remove.go#L45

But I think that has actually changed by now, because we have bundles support now.  So, we'll need it for the remove case and then we also needs the `language` again... :) All of this should be fixed now.

internal/runners/packages/packages.go
```diff
@@ -19,20 +19,14 @@ import (
...
-func executePackageOperation(out output.Outputer, prompt prompt.Prompter, language, name, version string, operation model.Operation) error {
-	// Use our own interpolation string since we don't want to assume our swagger schema will never change
-	var operationStr = "added"
-	if operation == model.OperationUpdated {
-		operationStr = "updated"
-	} else if operation == model.OperationRemoved {
-		operationStr = "removed"
-	}
+func executePackageOperation(out output.Outputer, authentication *authentication.Auth, prompt prompt.Prompter, language, name, version string, operation model.Operation) error {
+	pj := project.Get()
```
Naatan, line 23: Use value passed via primer.

internal/runners/packages/remove.go
```diff
@@ -25,12 +30,25 @@ func NewRemove(prime primeable) *Remove {
...
 func (r *Remove) Run(params RemoveRunParams) error {
 	logging.Debug("ExecuteRemove")
+	err := r.run(params)
+	headless.Notify(r.out, r.proj, err, "packages")
+	return err
+}
+
+func (r *Remove) run(params RemoveRunParams) error {
+	// Commit the package
+	pj := project.Get()
```
Naatan, line 47: Use value passed via primer.

Naatan (changes requested):

Naatan (approved):

PR #1014, opened 2020-10-06
Update commit ID, and account for non-alternate builds

Naatan:
https://www.pivotaltracker.com/story/show/174895290

I realized after pushing the other PR that it breaks camel based package changes. And it also doesn't set the commitID in the activestate.yaml despite updating the runtime, which will cause the runtime to think it's not up to date.


mdrohmann (approved):
Ha!  Yes, that's better... :)

PR #1015, opened 2020-10-07
Support extended support Python versions

MDrakos:
https://www.pivotaltracker.com/story/show/175062513

### DIFF
--- a/internal/language/language.go
+++ b/internal/language/language.go
@@ -2,10 +2,12 @@ package language
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/locale"
+	"github.com/blang/semver"
 )
 
 // Language tracks the languages potentially used.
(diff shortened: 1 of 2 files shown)


internal/language/language_test.go
```diff
@@ -116,3 +116,59 @@ func TestRecognizedSupporteds(t *testing.T) {
...
+			"Invalid version",
+			args{"python", ""},
+			Unknown,
+			true,
+		},
+		{
+			"Valid Perl version",
+			args{"perl", "5.28.1"},
+			Perl,
+			false,
```
Naatan, line 159: Please assert the versions also

internal/language/language.go
```diff
@@ -128,11 +126,11 @@ func MakeByName(name string) Language {
...
 	if strings.ToLower(name) == Python2.Requirement() {
-		version, err := semver.Parse(version)
-		if err != nil {
-			return Unknown, err
+		if len(version) == 0 {
+			return Unknown, locale.NewError("err_language_make_version", "Invalid version number, should be of format <major>.<minor>.<patch>")
 		}
-		name = name + strconv.FormatUint(version.Major, 10)
+		parts := strings.Split(version, ".")
+		name = name + parts[0]
```
Naatan, line 133: This is too specific to our current use-case. Instead let's just drop everything after 0.0.0 and send that to semver.

eg. `1.2.3.4` becomes `1.2.3`.

Naatan (changes requested):

internal/language/language_test.go
MDrakos: The version isn't returned from the `MakeByNameAndVersion` function so I'm not sure how to assert that. It's implicitly asserted in the Python tests but not for Perl

Naatan (approved):

PR #1016, opened 2020-10-07
Add headless notices at end of cmd exec

daved:
https://www.pivotaltracker.com/story/show/175020204


cmd/state/internal/headless/headless.go
```diff
@@ -0,0 +1,22 @@
...
+func Notify(prime *primer.Values, err error, cmdNames ...string) {
+	if err != nil || !prime.Project().IsHeadless() {
+		return
+	}
+
+	for _, cmd := range cmdNames {
+		prime.Output().Error(locale.T("message_headless_" + cmd))
+	}
+
+	prime.Output().Error(locale.T("message_headless"))
```
Naatan, line 21: Use `Notice()` not `Error()`

Naatan (changes requested):
I'd rather see this in the runner. cmdtree should just be dealing with the command registration and routing, it should not be printing anything.

Other than that this looks ok for a first iteration. Not sure about the headless package, but I can't think of anything better off the top of my head.

cmd/state/internal/headless/headless.go
daved: :facepalm: I didn't even think about whether notice was already going to stderr.

internal/runners/packages/import.go
```diff
@@ -48,25 +49,34 @@ func NewImportRunParams() *ImportRunParams {
...
 		prime.Output(),
+		prime.Project(),
 		prime.Prompt(),
 	}
 }
 
 // Run executes the import behavior.
 func (i *Import) Run(params ImportRunParams) error {
+	err := i.run(params)
+	headless.Notify(i.out, i.proj, err)
```
Naatan, line 75: missing cmdNames

internal/runners/packages/search.go
```diff
@@ -19,18 +19,26 @@ type SearchRunParams struct {
...
-		out: prime.Output(),
+		out:  prime.Output(),
+		proj: prime.Project(),
 	}
 }
 
 // Run is executed when `state packages search` is ran
 func (s *Search) Run(params SearchRunParams) error {
+	err := s.run(params)
+	headless.Notify(s.out, s.proj, err)
```
Naatan, line 37: Missing cmdNames

internal/runners/packages/list.go
```diff
@@ -23,18 +23,26 @@ type ListRunParams struct {
...
-		out: prime.Output(),
+		out:  prime.Output(),
+		proj: prime.Project(),
 	}
 }
 
 // Run executes the list behavior.
 func (l *List) Run(params ListRunParams) error {
+	err := l.run(params)
+	headless.Notify(l.out, l.proj, err)
```
Naatan, line 41: missing cmdNames

Naatan (changes requested):

internal/runners/packages/import.go
daved: Theses are supposed to be missing the cmdNames; They should only print a general message, and not a command-specific message.

Naatan: Where do you get that from? Looking at the AC's again I don't see these commands in there at all.

daved: hmm. I might have misunderstood. Pete had stated that all commands run headless should get a generic message, but some should get a more specific message. I'll review the AC and what Pete said in Slack and update the AC, then ping you for approval.

Naatan: Please do not update ACs without my say so. Slack does not trump ACs for stories that are in our sprint commit.

Naatan (approved):

PR #1018, opened 2020-10-08
Add --replace option to state activate

mdrohmann:
https://www.pivotaltracker.com/story/show/175020315


internal/runners/activate/activate.go
```diff
@@ -81,12 +95,28 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
...
+		if err != nil {
+			return locale.WrapError(err, "err_activate_replace_write_commit", "Failed to write new namespace to activestate.yaml.")
+		}
+	} else {
+		var fail *failures.Failure
+		names, fail = project.ParseNamespaceOrConfigfile(params.Namespace.String(), filepath.Join(projectPath, constants.ConfigFileName))
+		if fail != nil {
+			names = &project.Namespaced{}
+			logging.Debug("error resolving namespace: %v", fail.ToError())
+		}
```
Naatan, line 117: This is way too much going on for this function. Please split it off into a separate function.

internal/runners/activate/activate.go
```diff
@@ -51,9 +53,21 @@ func (r *Activate) Run(params *ActivateParams) error {
...
+		if fail != nil {
+			return locale.WrapError(fail.ToError(), "err_activate_replace_pathtouse", "Could not determine path to activestate.yaml.")
+		}
+		pathToUse = filepath.Dir(targetPath)
+	} else {
+		var err error
+		pathToUse, err = r.pathToUse(params.Namespace.String(), params.PreferredPath)
+		if err != nil {
+			return locale.WrapError(err, "err_activate_pathtouse", "Could not figure out what path to use.")
+		}
```
Naatan, line 70: This shouldn't be necessary. The regular behaviour of pathToUse is still valid.

pkg/projectfile/projectfile.go
```diff
@@ -606,15 +607,37 @@ func (p *Project) save(path string) *failures.Failure {
 	return nil
 }
 
+func (p *Project) SetNamespaceAndCommit(owner, project, commitID string) error {
```
Naatan, line 610: This is doing too much. Just change this to `SetNamespace`, then cal `SetCommit` separately. We're not trying to shave off milliseconds of IO.

pkg/projectfile/projectfile.go
```diff
@@ -606,15 +607,37 @@ func (p *Project) save(path string) *failures.Failure {
...
+	if err != nil {
+		return err
+	}
+
+	namespace := fmt.Sprintf("%s/%s", owner, project)
+
+	if commitID == "" {
+		return errs.New("commitID must not be empty")
+	}
+	commitQryParam := []byte(fmt.Sprintf("project: https://%s/%s?commitID=%s", constants.DefaultAPIHost, namespace, commitID))
```
Naatan, line 621: This shouldn't have any opinion on the API host.

pkg/projectfile/projectfile.go
```diff
@@ -606,15 +607,37 @@ func (p *Project) save(path string) *failures.Failure {
 	return nil
 }
 
+func (p *Project) SetNamespaceAndCommit(owner, project, commitID string) error {
+	data, err := ioutil.ReadFile(p.path)
+	if err != nil {
+		return err
```
Naatan, line 613: Wrap

pkg/projectfile/projectfile.go
```diff
@@ -606,15 +607,37 @@ func (p *Project) save(path string) *failures.Failure {
...
+	if !strings.Contains(string(out), commitID) {
+		return locale.NewError(
+			"err_set_namespace_and_commit_id", "Failed to set namespace {{.V0}} in activestate.yaml.", namespace)
+	}
+
+	if err := ioutil.WriteFile(p.path, out, 0664); err != nil {
+		return failures.FailOS.Wrap(err)
+	}
+
+	Reset()
```
Naatan, line 633: David made a good point in that this should be `p.Reload()`

Naatan (changes requested):

internal/runners/activate/activate.go
```diff
@@ -51,7 +53,14 @@ func (r *Activate) Run(params *ActivateParams) error {
 func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc) error {
 	logging.Debug("Activate %v, %v", params.Namespace, params.PreferredPath)
 
-	pathToUse, err := r.pathToUse(params.Namespace.String(), params.PreferredPath)
+	nSpace := params.Namespace.String()
+	path := params.PreferredPath
+	if params.Replace {
+		nSpace = ""
+		path = ""
+	}
```
Naatan, line 61: This isn't necessary. Just drop it. We don't want special behaviour based on some implicit flag.

internal/runners/activate/activate.go
```diff
@@ -113,6 +120,37 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
...
+	projectPath := filepath.Dir(prj.Source().Path())
+
+	var fail *failures.Failure
+	nspace, fail := project.ParseNamespaceOrConfigfile(names.String(), filepath.Join(projectPath, constants.ConfigFileName))
+	if fail != nil {
+		nspace = &project.Namespaced{}
+		logging.Debug("error resolving namespace: %v", fail.ToError())
+	}
+	return nspace, nil
+}
```
Naatan, line 152: This is doing too much. It's doing both replace AND figuring out the namespace. The namespace stuff can stay where it was for all I care. It was the replace stuff that should have its own function.

Keep it unixy! ;)

Naatan (changes requested):

internal/runners/activate/activate.go
mdrohmann: It actually *is* necessary.  If I specify a namespace, `pathToUse` believes that I want to activate a new project and hence need a new path where I can check it out.  So, it runs the `NamespaceSelect.Run()` which asks me for a new project path.

It is ugly either way, because `params.Replace` requires quite a bit of a different behavior:  It makes us re-interpret the namespace argument very differently, and maybe that is wrong in the first place.  As I can see that being very confusing to the user as well.  We could say that `--replace` needs to have a string argument that then will be a completely different parameter `params.ReplaceNamespace`?  Shall I do that? Note, however, that `state activate new-org/new-project --replace` would be invalid then...

Naatan: I see. Good point! I think the cleaner solution is what you proposed: make it an argument. And that also cleans up the UX (makes it less implicit). From a product perspective it shouldn't matter all that much since this is all invoked via the protocol. Which brings up the next problem: if we update this we also have to update the protocol handling. Would you mind doing that for this PR? Shouldn't be much of a detour.

Naatan (changes requested):
See my reply to your earlier review comment

cmd/state/internal/cmdtree/activate.go
```diff
@@ -47,6 +54,21 @@ func newActivateCommand(prime *primer.Values) *captain.Command {
 			},
 		},
 		func(_ *captain.Command, _ []string) error {
+			if params.ReplaceWith.IsValid() {
+				if params.PreferredPath != "" {
+					return locale.NewInputError(
+						"activate_flag_replace_and_path_incompatible",
+						"The flags --path and --replace-with are mutually exclusive.",
```
Naatan, line 61:
```suggestion
						"The flags --path and --replace are mutually exclusive.",
```

internal/runners/activate/activate.go
```diff
@@ -54,7 +55,10 @@ func (r *Activate) Run(params *ActivateParams) error {
 func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc) error {
 	logging.Debug("Activate %v, %v", params.Namespace, params.PreferredPath)
 
-	pathToUse, err := r.pathToUse(params.Namespace.String(), params.PreferredPath)
+	nSpace := params.Namespace.String()
+	path := params.PreferredPath
+
+	pathToUse, err := r.pathToUse(nSpace, path)
```
Naatan, line 61:
```suggestion
	pathToUse, err := r.pathToUse(params.Namespace.String(), params.PreferredPath)
```

internal/runners/activate/activate.go
```diff
@@ -116,6 +124,32 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
 	return activatorLoop(r.out, r.subshell, projectPath, activate)
 }
 
+func updateProject(prjFile *projectfile.Project, names *project.Namespaced) (*project.Project, error) {
+	if names.CommitID == nil || *names.CommitID == "" {
+		latestID, fail := model.LatestCommitID(names.Owner, names.Project)
+		if fail != nil {
+			return nil, locale.WrapInputError(fail.ToError(), "err_set_namespace_retrieve_commit", "Could not retrieve the latest commit for the specified project {{.V0}}.", names.String())
+		}
+		names.CommitID = latestID
```
Naatan, line 133: Feels a little dirty and error-prone using the namespace reference to record the commitID that you want to end up using. Suggest using a local var for this instead.

internal/runners/activate/activate.go
```diff
@@ -116,6 +124,32 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
 	return activatorLoop(r.out, r.subshell, projectPath, activate)
 }
 
+func updateProject(prjFile *projectfile.Project, names *project.Namespaced) (*project.Project, error) {
```
Naatan, line 127: This shouldn't be returning the project. That's making this function do more than it's tasked to.

Naatan (changes requested):

internal/runners/activate/activate.go
```diff
@@ -89,10 +86,16 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
 	// on --replace, replace namespace and commit id in as.yaml
 	if params.ReplaceWith.IsValid() {
 		var err error
-		projectToUse, err = updateProject(projectToUse.Source(), params.ReplaceWith)
+		err = updateProjectFile(projectToUse.Source(), params.ReplaceWith)
```
Naatan: Nit: these 3 lines can now be simplified to

```
if err := updateProjectFile(projectToUse.Source(), params.ReplaceWith); err != nil {
```

internal/runners/activate/activate.go
```diff
@@ -89,10 +86,16 @@ func (r *Activate) run(params *ActivateParams, activatorLoop activationLoopFunc)
...
 		if err != nil {
 			return locale.WrapError(err, "err_activate_replace_write", "Could not update the project file with new namespace.")
 		}
+		var fail *failures.Failure
+		// re-read the project file and return the updated project
+		projectToUse, fail = project.Parse(projectToUse.Source().Path())
+		if fail != nil {
+			return locale.WrapError(fail.ToError(), "err_activate_reload", "Could not reload project.")
+		}
```
Naatan: You shouldn't need this at all. We're reloading the project when it's changed. If this is necessary something in the project writing logic is broken.

Naatan (changes requested):

internal/runners/activate/activate.go
mdrohmann: I updated it, such that the project gets updated with the `SetNamespace` and `SetCommitID` calls that were only on the `projectfile.Project` struct.  I believe it is better to do these modifications on that level, as we do not only want to change the project *file* representation but also the actual representation of that file that we are using in the lines below.

pkg/project/project.go
```diff
@@ -51,6 +51,27 @@ type Project struct {
...
+
+// SetCommit updates the commit ID for the project and also updates the underlying project file
+func (p *Project) SetCommit(commitID string) error {
+	fail := p.Source().SetCommit(commitID)
+	if fail != nil {
+		return errs.Wrap(fail.ToError(), "Failed to write commitID to projectfile")
+	}
+	p.commitID = commitID
+	return nil
+}
```
Naatan, line 73: This is only really needed because we are recording these fields to the project struct, which looking at it now it doesn't really need to do at all. Let's instead fix this by sourcing these fields from the sourced projectfile.

Naatan (changes requested):

mdrohmann:
As discussed, `owner`, `project(name)` and `commitID` are now retrieved from the parsed URL string in the YAML file.  This should make things more clear in the future.

Naatan (changes requested):
Looks great! Just one minor change.

Naatan (approved):

mdrohmann:
Some unit tests failed because they were creating a projectfile without initializing the `parsedURL` structure.  As it was happening in several unit tests, my current change parses the URL on demand and then memoizes the result.

On second thought: This makes the `projectfile` implementation more robust, but also more complicated just to support some tests.  Let me know what you think:  Of course, I could also revert the latest commit and initialize the `parsedURL` in every failing test manually...

Naatan (changes requested):
This is making things too involved and error-prone just for the benefit of unit tests. I'd prefer we fix the unit tests instead.

Naatan (approved):

PR #1021, opened 2020-10-09
install.sh does not require State Tool to be on PATH if --activate

mdrohmann:
https://www.pivotaltracker.com/story/show/174782116

This is okay, as we set aliases entries to the current State
Tool.

 https://www.pivotaltracker.com/story/show/174782116


installers/install.sh
```diff
@@ -419,3 +397,11 @@ else
...
   fi
 fi
+
+if [ -n "${ACTIVATE}" ]; then
+  # switch this shell to interactive mode
+  set -i
+  # control flow of this script ends with this line: replace the shell with the activated project's shell
+  exec $STATEPATH activate ${ACTIVATE}
+fi
+info "You may now start using the '$STATEEXE' program."
```
daved, line 407: Are you developing on Windows? If not, how did the lack of newline happen?!

daved (approved):

installers/install.sh
daved: I marked approved thinking of this as a nit, but on Linux/Darwin, the newline is kinda/sorta required.

daved (changes requested):

daved (approved):

PR #1024, opened 2020-10-13
Add title mechanism to captain command

daved:
https://www.pivotaltracker.com/story/show/174261853

 https://www.pivotaltracker.com/story/show/174261853


daved:
I'm not sure if commands like _prepare, _protocol, and shim, should our should not have titles. For this PR, I generally favored noisiness, but it may be worth keeping an eye out for commands that should be quieter.

internal/output/txtstyle/title.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+const (
+	// DefaultTitlePadding is the padding character length.
+	DefaultTitlePadding = 3
+)
+
+// Title represents the config of a styled title. It does not, currently,
+// support combining diactics (more info: https://play.golang.org/p/VmHyq3JJ7On).
+type Title struct {
```
Naatan, line 14: Let's have this implement output.Marshaller

http://github.com/ActiveState/cli/blob/d657478e6235fe89e8a7bc401c63e60d83349ac3/internal/output/mediator.go#L8-L8

And then only have it output data for the plain outputter, eg.

```
func (t *Title) MarshalOutput(format output.Format) interface{} {
	if format == output.PlainFormatName {
		return t.String()
	}
	return nil
}
```

Naatan: We may need to implement some type to mean "don't output this". I think `nil` will just cause it to print something "empty". Could have some type `output.Suppress` that's handled here:

http://github.com/ActiveState/cli/blob/d657478e6235fe89e8a7bc401c63e60d83349ac3/internal/output/mediator.go#L12-L12

internal/output/txtstyle/title.go
```diff
@@ -0,0 +1,66 @@
...
+		// IS not in first AND not in last line AND start or end of line
+		if i > lineLen-1 && i < lineLen*4-1 && (i%lineLen == 0 || i%lineLen == lineLen-2) {
+			rs[i] = '║'
+			continue
+		}
+
+		rs[i] = ' '
+	}
+
+	return string(rs[:len(rs)-1]) // drop ending newline
```
Naatan, line 65: Very cool! My only criticism; can we turn things like `lineLen-2` into variables like `posTopRightCorner`. That would instantly make this whole function a lot easier to grok.

internal/output/txtstyle/title_test.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func TestTitleString(t *testing.T) {
+	tests := []struct {
+		test  string
+		title string
+		want  string
+	}{
+		{test: "basic non-empty", title: "Platforms", want: nonemptyBasic},
+		{test: "basic empty", title: "", want: emptyBasic},
```
Naatan, line 15: I think if it's empty we should just not print anything.

Naatan (changes requested):

Naatan (approved):

PR #1025, opened 2020-10-13
Table output

MDrakos:
https://www.pivotaltracker.com/story/show/174262049

 https://www.pivotaltracker.com/story/show/174262049


internal/table/table.go
```diff
@@ -0,0 +1,33 @@
...
+
+type Table struct {
+	Info        string      `opts:"hideKey"`
+	Rows        interface{} `opts:"hideKey"`
+	emptyOutput string
+}
+
+func (t *Table) MarshalOutput(format output.Format) interface{} {
+	if format == output.PlainFormatName {
+		value := reflect.ValueOf(t.Rows)
```
MDrakos, line 17: I would rather not use `reflect` here. Alternatively we could not handle the empty output.

internal/runners/scripts/scripts.go
```diff
@@ -32,39 +30,19 @@ type scriptLine struct {
...
-	t.SetAlign("left")
-
-	return t.Render("simple")
-}
-
-func (f outputFormat) scriptsTable() (hdrs []string, rows [][]string) {
-	for _, s := range f {
-		row := []string{
-			s.Name, s.Description,
+func (t scriptsTable) MarshalOutput(format output.Format) interface{} {
```
Naatan, line 37: You don't need `MarshalOutput` here. This is plain controller logic, no custom marshaller required.

internal/output/plain.go
```diff
@@ -320,8 +320,7 @@ func sprintTable(slice []interface{}) (string, error) {
 	// Don't print whitespace lines
 	t.SetHideLines([]string{"betweenLine", "top", "aboveTitle", "LineTop", "LineBottom", "bottomLine"})
 	t.SetAlign("left")
-
-	render := t.Render("simple")
+	render := t.Render("standard")
```
Naatan, line 323: Rather than bake our layout into tabulate and marry us to this fork, why not instead fix the issue that stopped you from using the original and open a PR? We can use the fork in the interim.

So in other words: export the fields that prevented you from creating our own custom layout.

Naatan (changes requested):

internal/output/plain.go
MDrakos: Fair point. The library appears to be fairly stale but I think this is from lack of issues and PRs, not neglect from the author so I will create a PR there with the changes we need.

Naatan (changes requested):
You didn't address 

https://github.com/ActiveState/cli/pull/1025#discussion_r504977461

I'm asking for that change to be reflected here. And then yeah; also open a PR with the original repo. But the main action item there is to make those changes to the code.

internal/output/plain.go
```diff
@@ -320,6 +320,21 @@ func sprintTable(slice []interface{}) (string, error) {
 	// Don't print whitespace lines
 	t.SetHideLines([]string{"betweenLine", "top", "aboveTitle", "LineTop", "LineBottom", "bottomLine"})
 	t.SetAlign("left")
+
+	// Set our own custom table format
+	tabulate.TableFormats["standard"] = tabulate.TableFormat{
+		LineTop:         tabulate.Line{"", "-", "", ""},
+		LineBelowHeader: tabulate.Line{"", "\u2500", "", ""},
+		LineBottom:      tabulate.Line{"", "-", "", ""},
+		HeaderRow:       tabulate.Row{"[BOLD]", "", "[/RESET]"},
```
Naatan, line 329:
```suggestion
		HeaderRow:       tabulate.Row{"[HEADING]", "", "[/RESET]"},
```

Martin JUST pushed the change that removes `BOLD`.

Naatan (dismissed):

Naatan (approved):

PR #1027, opened 2020-10-14
remove explicit colors in favor of semantic formatters

mdrohmann:
https://www.pivotaltracker.com/story/show/174262019


internal/runners/run/run.go
```diff
@@ -83,7 +83,7 @@ func run(out output.Outputer, subs subshell.SubShell, proj *project.Project, nam
 	if !lang.Recognized() {
 		warning := locale.Tl(
 			"run_warn_deprecated_script_without_language",
-			"[YELLOW]DEPRECATION WARNING: Scripts without a defined language currently fall back to using  the default shell for your platform. This fallback mechanic will soon stop working and a language will need to be explicitly defined for each script. Please configure the 'language' field with a valid option (one of {{.V0}})[/RESET]",
+			"[NOTICE][HIGHLIGHT]DEPRECATION WARNING:[/RESET][NOTICE] Scripts without a defined language currently fall back to using  the default shell for your platform. This fallback mechanic will soon stop working and a language will need to be explicitly defined for each script. Please configure the 'language' field with a valid option (one of {{.V0}})[/RESET]",
```
Naatan, line 86:
```suggestion
			"[NOTICE][HIGHLIGHT]DEPRECATION WARNING:[/RESET][NOTICE] Scripts without a defined language currently fall back to using  the default shell for your platform. This fallback mechanic will soon stop working and a language will need to be explicitly defined for each script. Please configure the 'language' field with a valid option (one of {{.V0}})[/RESET]",
```

Is this covered in the UX mocks? ie. using HIGHLIGHT here at all.

internal/output/color.go
```diff
@@ -12,12 +12,62 @@ var colorRx *regexp.Regexp
...
+
+// Heading switches to bold and bright foreground
+func (dct defaultColorTheme) Heading(writer io.Writer) {
+	ct.Foreground(writer, ct.White, true)
+	ct.ChangeStyle(writer, ct.Bold)
+}
+
+// Notice switches to bright foreground
+func (dct defaultColorTheme) Notice(writer io.Writer) {
+	ct.Foreground(writer, ct.White, true)
```
Naatan, line 41: Isn't this also setting it as white? The AC just calls for "bright".

Naatan (changes requested):

internal/output/color.go
mdrohmann: Unfortunately, it is not possible to just set the "bright" flag, you always have to specify a color, and white is the default color.  When I set "white+bright" on my terminal which has a solarized color theme, it actually shows up as bright black (or grey-ish).  So, I think that's the best we can do here.

internal/runners/run/run.go
mdrohmann: You are probably right. It just says

> great for highlights and actionable content

But it is not really actionable.  So, I think I will remove the HIGHLIGHT.

internal/output/color.go
Naatan: I see. In that case let's merge `HIGHLIGHT` into `NOTICE`. So let's drop `HIGHLIGHT` and instead make all those instances use `NOTICE`.

Naatan (changes requested):
See my response to your comment

internal/output/color.go
mdrohmann: Just to make sure that I undestand correctly:  Would you like to remove the HIGHLIGHT formatter completely (although it is using a different color `teal`)?

I exchanged some uses of `[BOLD] a url or something [/RESET]` with highlight which seems sensible to me.  Do you want me to change these to `[NOTICE]` or remove them completely?

mdrohmann: Mostly `[HIGHLIGHT]` is currently used for State Tool commands that could be copied and pasted to the command line.

Naatan: Ok, let's maybe just make the difference more distinct then. Let's rename `HIGHLIGHT` to `ACTIONABLE`.

internal/output/color.go
```diff
@@ -107,7 +107,7 @@ func colorize(ct ColorTheme, writer io.Writer, arg string) {
 		ct.Error(writer)
 	case `DISABLED`:
 		ct.Disabled(writer)
-	case `HIGHLIGHT`:
+	case `ACTIONABLE`:
 		ct.Highlight(writer)
```
Naatan, line 111: Can we rename this too please

Naatan (changes requested):

internal/output/color.go
mdrohmann: Ah! Thanks for catching.

Naatan (approved):

PR #1034, opened 2020-10-19
Improved User Experience

Naatan:


cmd/state/errors.go
```diff
@@ -5,17 +5,77 @@ import (
...
+	}
+	return "\n" + strings.Join(outLines, "\n")
+}
+
+func trimError(msg string) string {
+	if strings.Count(msg, ".") > 1 || strings.Count(msg, ",") > 1 {
+		return msg // Don't trim dots if we have multiple sentences.
+	}
+	return strings.TrimRight(msg, " .")
+}
```
mdrohmann, line 77: I don't quite understand what the `trimError` function is supposed to do. 
Could you please explain?

 I also noticed that it is currently skipping the trimming when there are at least *2* dots or commas in the message.   Should it be *1* instead?

mdrohmann (changes requested):
It looks pretty cool.

cmd/state/errors.go
Naatan: This is making the errors consistently not have a period at or whitespace at the end UNLESS the sentence includes a comma or multiple dots (ie. is more than one sentence).

mdrohmann: Should it then be: ``` strings.Count(msg, ",") > 0 ``` ?

cmd/state/errors.go
```diff
@@ -5,17 +5,77 @@ import (
...
+	// Print tips
+	outLines = append(outLines, output.Heading(locale.Tl("err_more_help", "Need More Help?")).String())
+	for _, tip := range errorTips {
+		outLines = append(outLines, fmt.Sprintf(" [DISABLED]-[/RESET] %s", trimError(tip)))
+	}
+	return "\n" + strings.Join(outLines, "\n")
+}
+
+func trimError(msg string) string {
+	if strings.Count(msg, ".") > 1 || strings.Count(msg, ",") > 1 {
```
Naatan, line 73:
```suggestion
	if strings.Count(msg, ".") > 1 || strings.Count(msg, ",") > 0 {
```

cmd/state/errors.go
Naatan: Yes, you're right! fixed.

mdrohmann (approved):

PR #1037, opened 2020-10-21
Ensure project checkouts are unique

mdrohmann:
https://www.pivotaltracker.com/story/show/175148779


mdrohmann:
I ran a bunch of integration tests.  Which revealed this bug https://www.pivotaltracker.com/story/show/175391329 that breaks the `state activate --default` tests, so I eventually removed `run-int-test-only: activate` from the tags list.

internal/fileutils/fileutils.go
```diff
@@ -772,6 +772,26 @@ func ResolvePath(path string) (string, error) {
 	return evalPath, nil
 }
 
+// ResolveUniquePath gets the absolute location of the provided path
+// with the best effort attempt to produce the same result for all possible paths to the
+// given target.
+func ResolveUniquePath(path string) (string, error) {
+	evalPath, err := ResolvePath(filepath.Clean(path))
```
Naatan, line 779: Wouldn't it be safer to put `filepath.Clean()` at the end of this function? Otherwise we're assuming both ResolvePath and GetLongPathName will only ever produce cleaned paths themselves.

internal/fileutils/path_helpers_lin_mac.go
```diff
@@ -0,0 +1,15 @@
+// +build !windows
+
+package fileutils
+
+// GetShortPathName returns the Windows short path
+// This function does not alter the path in any way on Linux or MacOS
+func GetShortPathName(path string) (string, error) {
```
Naatan, line 7: Nit: I think it would be cleaner if you made `ResolveUniquePath` be OS specific, and made these functions just things used only on the Windows implementation.

internal/runbits/namespace.go
```diff
@@ -0,0 +1,28 @@
...
+func AvailableProjectPaths(c ConfigAble, namespace string) []string {
+	key := fmt.Sprintf("project_%s", namespace)
+	paths := c.GetStringSlice(key)
+	paths = funk.FilterString(paths, func(path string) bool {
+		return fileutils.FileExists(filepath.Join(path, constants.ConfigFileName))
+	})
+	paths = funk.UniqString(paths)
+	c.Set(key, paths)
+	return paths
+}
```
Naatan, line 28: This should be part of the project package. Not runbits.

Runbits is for controller logic that's shared between controllers. ie. code that prints.

pkg/platform/runtime/runtime.go
```diff
@@ -20,15 +23,20 @@ type Runtime struct {
...
 }
 
-func NewRuntime(commitID strfmt.UUID, owner string, projectName string, msgHandler MessageHandler) *Runtime {
-	var installPath string
-	if runtime.GOOS == "darwin" {
-		// mac doesn't use relocation so we can safely use a longer path
-		installPath = filepath.Join(config.CachePath(), owner, projectName)
-	} else {
-		installPath = filepath.Join(config.CachePath(), hash.ShortHash(owner, projectName))
+func NewRuntime(projectDir string, commitID strfmt.UUID, owner string, projectName string, msgHandler MessageHandler) (*Runtime, error) {
```
Naatan, line 26: `projectDir` should probably be `cacheDir`

pkg/platform/runtime/runtime.go
```diff
@@ -20,15 +23,20 @@ type Runtime struct {
...
+	var resolvedProjectDir string
+	if projectDir != "" {
+		var err error
+		projectDir = strings.TrimSuffix(projectDir, constants.ConfigFileName)
+		resolvedProjectDir, err = fileutils.ResolveUniquePath(projectDir)
+		if err != nil {
+			return nil, errs.Wrap(err, "Failed to resolve unique file path to project dir")
```
Naatan, line 33: Seems like a good place to localize the error where we have some context about what actually went wrong.

Naatan (changes requested):

internal/fileutils/fileutils.go
mdrohmann: I fundamentally agree and had it like this in the beginning, but then on MacOS the unit tests fail.  I don't understand it still, although I put quite some effort in trying to understand it, but if I put `/private/var/temp/./targetdir` in there, `ResolveUniquePath` does not resolve the symlink to `/var/temp/targetdir`.  If I put the cleaned version `/private/var/temp/targetdir` in, it does.

Cf., https://github.com/ActiveState/cli/runs/1289415663#step:7:37

Of course, I could also add a `filepath.Clean` at the end again.  Maybe I'll do that, just to be (more) sure.

internal/runbits/namespace.go
mdrohmann: I was a little unsure on where to put it.  `project` makes sense.

pkg/platform/runtime/runtime.go
mdrohmann: I think `projectDir` is a better name here, because it actually points to the directory where the `activestate.yaml` file is located.  We create the `cacheDir` from that `projectDir` (by hashing it), as the AC requested.

internal/fileutils/fileutils.go
Naatan: Yeah I think that works; call it twice. 

I remember struggling with the macOS tempdir issue way back, never quite figured it out either.

pkg/platform/runtime/runtime.go
Naatan: ahh gotcha

Naatan (approved):

PR #1040, opened 2020-10-22
Reorganize help output

MDrakos:
https://www.pivotaltracker.com/story/show/173974781


MDrakos:
Here is a sample output:

```
» ./build/state --help
The ActiveState CLI allows you to easily switch between your ActiveState environments

Find more information at: https://docs.activestate.com/platform/state/

Package Management:
  add         Add a new package to your project
  remove      Remove a package from your project
  import      Import packages from a list of dependencies
  search      Search for available packages that can be added to your project
  packages      Manage packages used in your project

Environment Management:
  activate      Activate a project
  init          Initialize a new project
  clean         Clean caches, configuration files, or completely remove the state tool
  deploy        Deploys the given runtime environment on the local system

Platform Tools:
  auth          Authenticate against the ActiveState Platform
  invite        Invite new members to an organization
  organizations List member organizations on the ActiveState Platform
  platforms     Manage platforms used in your project
  projects      Manage your projects
  show          Show information about a project

Automation:
  events        Manage project events
  scripts       Show project scripts

Version Control:
  fork          Fork an existing ActiveState Platform project
  pull          Pull in the latest version of your project from the ActiveState Platform
  push          Push your latest changes to the platform
  history       View history of the active or given project

Utilities:
  update        Updates the State Tool to the latest available version
  export        Print information based on the provided subcommand

Usage:
state [flags]
state [command]

Flags:
      --confirm-exit-on-error   prompts the user to press enter before exiting, when an error occurs
  -h, --help                    help for state
  -l, --locale string           Set the localisation
      --mono                    Force monochrome text output
  -o, --output string           Set the output method, possible values: plain, json, editor
  -v, --verbose                 Verbose output
      --version                 Show the version of our state executable

Use "state [command] --help" for more information about a command.
```

Note that the inconsistent spacing under `Package Management` is due to the `add`, `remove`, `import`, and `search` commands having a different parent command than the rest. This is what cobra uses to determine the padding. Once the command are restructured the output should be spaced correctly.

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -113,6 +113,68 @@ func New(prime *primer.Values, args ...string) *CmdTree {
 
 	applyLegacyChildren(stateCmd, globals)
 
+	groups := []captain.CommandGroup{
```
Naatan, line 116: The group registration should live with the command itself. We don't want to start registering commands in multiple different ways.

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -113,6 +113,68 @@ func New(prime *primer.Values, args ...string) *CmdTree {
...
+		{
+			Message: "Package Management:",
```
Naatan, line 118: This isn't localized. And the colon at the end is presentation, not something that we should be concerned with at this level.

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -113,6 +113,68 @@ func New(prime *primer.Values, args ...string) *CmdTree {
...
+				stateCmd.FindSafe([]string{"export"}),
+			},
+		},
+	}
+
+	templater := captain.Templater{
+		Cmd:           stateCmd,
+		CommandGroups: groups,
+	}
+	stateCmd.SetUsageFunc(templater.RootCmdUsageFunc())
```
Naatan, line 176: cmdtree is for registering command structure, captain is for representing the commands. Setting the template should happen in captain.

internal/captain/template.go
```diff
@@ -0,0 +1,93 @@
...
+	"os"
+	"strings"
+	"text/template"
+	"unicode"
+
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/spf13/cobra"
+)
+
+type Templater struct {
```
Naatan, line 14: I'm not sure I understand what this struct is used. None of the methods under it seems to actually work together in any meaningful way. It just seems like you're trying to save yourself passing the struct fields as arguments to the functions? 

Naming it something very generic like "Templater" also adds to the confusion.

internal/captain/template.go
```diff
@@ -0,0 +1,93 @@
+package captain
```
Naatan, line 1: This file is kind of all over the place. Can you please organize your code a little. I'll try to make notes to call out particular things but that's just to hopefully give some idea on what can be improved. Really this just needs some architecture put into it. At the moment this looks like prototype code.

internal/captain/template.go
```diff
@@ -0,0 +1,93 @@
...
+		cmds := []string{cmdGroup.Message}
+		for _, cmd := range cmdGroup.Commands {
+			if cmd.cobra.IsAvailableCommand() {
+				cmds = append(cmds, fmt.Sprintf("  %s %s", rpad(cmd.Use(), cmd.cobra.NamePadding()), cmd.Description()))
+			}
+		}
+		groups = append(groups, strings.Join(cmds, "\n"))
+	}
+	return strings.Join(groups, "\n\n")
+}
```
Naatan, line 50: Imagine we can loose this function entirely by moving the group registration to the command itself, but some concerns just for future reference;

 - You're hardcoding a lot of templating logic here, that logic should be in the template
 - Try to return/continue early. eg. in your cmd loop invert the conditional and turn it into a `continue`, then your append logic lives a level higher and you're not 5 indentation levels deep. (nitpick)

internal/captain/template.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (t *Templater) cmdGroups(c *cobra.Command) []CommandGroup {
+	if len(t.CommandGroups) == 0 {
+		return nil
+	}
+	return t.CommandGroups
```
Naatan, line 56: This seems like pointless busywork. Aren't you going to have to nil check the result of this function anyway?

internal/captain/template.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (t *Templater) templateFuncs() template.FuncMap {
+	return template.FuncMap{
+		"CmdGroupsString":         t.cmdGroupsString,
+		"trimTrailingWhitespaces": trimTrailingWhitespaces,
+		"rpad":                    rpad,
```
Naatan, line 63: OCD trigger. SnakeCase AND camelCase?

internal/captain/template.go
```diff
@@ -0,0 +1,93 @@
...
+func localizedTemplate(c *Command, template string) string {
+	localizedArgs := []map[string]string{}
+	for _, arg := range c.Arguments() {
+		req := ""
+		if arg.Required {
+			req = "1"
+		}
+		localizedArgs = append(localizedArgs, map[string]string{
+			"Name":        locale.T(arg.Name),
+			"Description": locale.T(arg.Description),
```
Naatan, line 76: These fields are already localized

Naatan (changes requested):

cmd/state/internal/cmdtree/cmdtree.go
Naatan: fwiw, I'd expect to see something like this:

```
Index: internal/captain/command.go
===================================================================
--- internal/captain/command.go	(revision ee63cd9bb4a9a38ccbd267bea9645c4004e3c8eb)
+++ internal/captain/command.go	(date 1603403020369)
@@ -42,7 +42,13 @@
 	out output.Outputer
 }
 
-func NewCommand(name, title, description string, out output.Outputer, flags []*Flag, args []*Argument, executor Executor) *Command {
+type CommandGroup string
+
+var (
+	PackagesGroup CommandGroup = "packages"
+)
+
+func NewCommand(name, title, description string, groups []CommandGroup, out output.Outputer, flags []*Flag, args []*Argument, executor Executor) *Command {
 	// Validate args
 	for idx, arg := range args {
 		if idx > 0 && arg.Required && !args[idx-1].Required {
```

internal/captain/template.go
MDrakos: If was for convenience and was initially doing more than it is here. But you're right, I can just pass the arguments.

internal/captain/template.go
MDrakos: Agreed. What's left in this file _should_ just be related to templates

internal/captain/template.go
MDrakos: Yep, it had an extra conditional before but after I removed it I did not remove the function. It's gone now.

cmd/state/internal/cmdtree/cmdtree.go
MDrakos: We were already setting the template in this file though I'm not sure why. I've removed it.

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -137,6 +137,55 @@ func newStateCommand(globals *globalOptions, prime *primer.Values) *captain.Comm
...
+				},
+			},
+			{
+				Message: locale.Tl("utilities_group_message", "Utilities"),
+				Commands: []string{
+					"update",
+					"export",
+				},
+			},
+		},
```
MDrakos, line 188: I realize this strays a bit from what we discussed. However it feels in line with how we create our new commands. Additionally this does not require us to keep a mapping of group names to commands. We can search for the commands as we encounter them in the template.

Naatan (changes requested):
Still missing my point I think. The command registration itself should be calling out the group. There shouldn't be a single command that registers all the grouping.

So when we register the "activate" command, that call should also be setting the group that it belongs to.

MDrakos (commented):
Looks like there is a failing unit test here: https://github.com/ActiveState/cli/blob/master/internal/locale/locale_test.go#L12

Otherwise looks good to me.

PR #1041, opened 2020-10-22
fix regression: nested activations

mdrohmann:
https://www.pivotaltracker.com/story/show/175359617

https://www.pivotaltracker.com/story/show/175391329

### DIFF
--- a/internal/runners/activate/activate.go
+++ b/internal/runners/activate/activate.go
@@ -66,7 +66,7 @@ func (r *Activate) run(params *ActivateParams) error {
 			return locale.NewInputError("err_already_activated", "You cannot activate a new project when you are already in an activated state.")
 		}
 		if params.Namespace == nil || params.Namespace.IsValid() {
-			return locale.NewInputError("err_conflicting_default_while_activated", "Cannot set {{.V0}} as the global default project while in an activated state.", params.Namespace.String())
+			return locale.NewInputError("err_conflicting_default_while_activated", "Cannot set [NOTICE]{{.V0}}[/RESET] as the global default project while in an activated state.", params.Namespace.String())
 		}
 	}
 
@@ -119,7 +119,7 @@ func (r *Activate) run(params *ActivateParams) error {
 			return locale.WrapError(err, "err_activate_default", "Could not configure your project as the default.")
 		}
 
-		r.out.Notice(locale.Tl("global_default_set", "Successfully configured {{.V0}} as the global default project.", proj.Namespace().String()))
+		r.out.Notice(locale.Tl("global_default_set", "Successfully configured [NOTICE]{{.V0}}[/RESET] as the global default project.", proj.Namespace().String()))
 
 		if alreadyActivated {
 			return nil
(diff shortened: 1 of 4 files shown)


mdrohmann:
I spend some time getting the `activate` integration tests to run again, following some red herrings until I finally realized that they were running into this bug: https://www.pivotaltracker.com/story/show/175418057  which I think is pretty severe, but also unrelated to this story.  So, once again, I de-activated the integration tests for `activate`.

internal/virtualenvironment/virtualenvironment_test.go
```diff
@@ -38,19 +37,6 @@ func teardown() {
...
-	defer teardown()
-
-	os.Setenv(constants.ActivatedStateEnvVarName, "test")
-
-	venv := New(nil)
-	failure := venv.Activate()
-	require.NotNil(t, failure, "expected a failure")
-	assert.Equal(t, FailAlreadyActive, failure.Type)
-	assert.Equal(t, locale.Tr("err_already_active"), failure.Error())
-}
```
Naatan, line 52: Seeing as this broke and we weren't aware; can we ensure that we're still testing this? Just in the right place. Probably an integration test.

internal/runners/activate/activate.go
```diff
@@ -60,6 +60,16 @@ func (r *Activate) Run(params *ActivateParams) error {
 func (r *Activate) run(params *ActivateParams) error {
 	logging.Debug("Activate %v, %v", params.Namespace, params.PreferredPath)
 
+	alreadyActivated := subshell.IsActivated()
+	if alreadyActivated {
+		if !params.Default {
+			return locale.NewInputError("err_already_activated", "You cannot activate a new project when you are already in an activated state.")
+		}
+		if params.Namespace == nil || params.Namespace.IsValid() {
+			return locale.NewInputError("err_conflicting_default_while_activated", "Cannot set {{.V0}} as the global default project while in an activated state.", params.Namespace.String())
```
Naatan, line 69:
```suggestion
			return locale.NewInputError("err_conflicting_default_while_activated", "Cannot set [NOTICE]{{.V0}}[/RESET] as the global default project while in an activated state.", params.Namespace.String())
```

internal/runners/activate/activate.go
```diff
@@ -108,6 +118,12 @@ func (r *Activate) run(params *ActivateParams) error {
 		if err != nil {
 			return locale.WrapError(err, "err_activate_default", "Could not configure your project as the default.")
 		}
+
+		r.out.Notice(locale.Tl("global_default_set", "Successfully configured {{.V0}} as the global default project.", proj.Namespace().String()))
```
Naatan, line 122:
```suggestion
		r.out.Notice(locale.Tl("global_default_set", "Successfully configured [NOTICE]{{.V0}}[/RESET] as the global default project.", proj.Namespace().String()))
```

Naatan (changes requested):

Naatan (approved):

PR #1047, opened 2020-10-26
State auth signup fails if already authenticated

MDrakos:
https://www.pivotaltracker.com/story/show/174172289

### DIFF
--- a/pkg/cmdlets/auth/signup.go
+++ b/pkg/cmdlets/auth/signup.go
@@ -45,6 +45,10 @@ type signupInput struct {
 func Signup(out output.Outputer, prompt prompt.Prompter) error {
 	input := &signupInput{}
 
+	if authentication.Get().Authenticated() {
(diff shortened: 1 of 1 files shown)


pkg/cmdlets/auth/signup.go
```diff
@@ -45,6 +45,10 @@ type signupInput struct {
 func Signup(out output.Outputer, prompt prompt.Prompter) error {
 	input := &signupInput{}
 
+	if authentication.Get().Authenticated() {
+		return locale.NewInputError("err_auth_authenticated", "Currently authenticated as: {{.V0}}. Please log out run `state auth signup` again", authentication.Get().WhoAmI())
```
Naatan, line 49:
```suggestion
		return locale.NewInputError("err_auth_authenticated", "You are already authenticated as: {{.V0}}. You can log out by running `state auth logout`.", authentication.Get().WhoAmI())
```

Naatan (changes requested):

Naatan (approved):

PR #1049, opened 2020-10-26
Fix activate_int_test on Windows

mdrohmann:
https://www.pivotaltracker.com/story/show/169467800


Naatan (dismissed):

mdrohmann:
The story was so old that I am unsure if the bug actually still persists.  I added an integration test on Windows, and fixed a bunch of problems such that all of the tests pass again:
- state shim always resolves the path to the executable in the default project (otherwise it may run itself in an infinite loop)
  - Note: This can still happen if the shimmed executable is not part of the downloaded artifacts
- state shim arguments are forwarded with `"$@"` instead of `$*`, as otherwise `-c 'import sys; print(sys.copyright)'` arrived as arguments `["-c", "import", "sys;", ...]` at the shim command.
- `cmd` shell does not know about shebangs in script files, so we have to specify the executable to run the script with, explicitly.
- When running tests with `ACTIVESTATE_CLI_DISABLE_RUNTIME`, the activation message is only printed if the script language is available on the machine.  Windows CI only has python2.  So, I moved these tests from `python3` to `python2`.

There is also a follow-up story: https://www.pivotaltracker.com/story/show/175468062

internal/virtualenvironment/virtualenvironment.go
```diff
@@ -78,6 +79,23 @@ func (v *VirtualEnvironment) Setup(installIfNecessary bool) *failures.Failure {
...
+// FindProgramOnPath tries to find the cmd on the PATH environment variable
+func FindProgramOnPath(cmd string, envMap map[string]string) string {
+	var bins []string
+	if p, ok := envMap["PATH"]; ok {
+		bins = strings.Split(p, string(os.PathListSeparator))
+	}
+
+	for _, b := range bins {
+		cmdPath := filepath.Join(b, cmd)
+		if fileutils.IsExecutable(cmdPath) {
```
Naatan, line 91: This only works if the user typed the fully qualified executable name. (eg. "perl.exe" on windows, rather than just "perl").

Why do we need the full path though? We're setting up the environment for them, their shell should be doing this, not us.

test/integration/activate_int_test.go
```diff
@@ -46,7 +46,7 @@ func (suite *ActivateIntegrationTestSuite) TestActivateWithoutRuntime() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
-	cp := ts.Spawn("activate", "ActiveState-CLI/Python3")
+	cp := ts.Spawn("activate", "ActiveState-CLI/Python2")
```
Naatan, line 49: What's the purpose of changing these to Python 2?

test/integration/activate_int_test.go
```diff
@@ -335,7 +347,7 @@ func (suite *ActivateIntegrationTestSuite) TestActivate_FromCache() {
 	defer ts.Close()
 
 	cp := ts.SpawnWithOpts(
-		e2e.WithArgs("activate", "ActiveState-CLI/small-python", "--path", ts.Dirs.Work),
+		e2e.WithArgs("activate", "ActiveState-CLI/Python2", "--path", ts.Dirs.Work),
```
Naatan, line 350: Why change these? I'm a little concerned we're fixing tests just so they pass without ensuring that it actually fixes the user facing issue.

Naatan (changes requested):

internal/virtualenvironment/virtualenvironment.go
mdrohmann: That is a good point:  So, we would also need to take the `PATHEXT` environment variable into account.

On you question why we need this:
When we start a program in the subshell, we parse our own environment variables and existing `.bashrc`, `.zshrc` etc files in.  And possibly they are merged with an existing environment.  In the tests that created a situation, where `<cachedir>/bin` always appeared early in the `PATH`.  So, when I call 

1. I call `<cachedir>/bin/python3`
1. that calls `state shim -- python3`
1. after running it through `subshell.Run()` we are calling `<cachedir>/bin/python3` again!
1. continue at 1.

This was my suggestion to resolve this issue, but I am now realizing that this is incorrect for two reasons:
1. the `PATHEXT` variable needs to be taken into account on Windows.
1. `state shim` should just ensure that it is not calling itself, maybe by removing the global bin dir from the path.  As this here would fail, if we "shim" an executable that is not in the runtime environment.

mdrohmann: Also I am realizing:  As I de-activated the `state activate --default` tests, this is not necessary anymore to ensure that the tests are actually working, so how about I revert these changes, and write up a story with what I think needs to be done?

test/integration/activate_int_test.go
mdrohmann: It is ugly. No debate here...

- The `state activate` command now always runs an activation event, which is interpreted by a python script.
- To speed these tests up, we are running with `ACTIVESTATE_CLI_DISABLE_RUNTIME=true`.  => We do not provide a `python` executable!  So, we rely on the `python` executable that is provided by the system.
- On Windows CI, we only have `python2` but not `python3`.

As the activation event is not really what we are testing here, I chose to do it this way.

The alternative but slower solution is to enable the runtime for these tests.

mdrohmann:
This PR is now superseded by #1063 and #1064 .

PR #1052, opened 2020-10-27
Add before/after cmd event handling

daved:
https://www.pivotaltracker.com/story/show/175386922

 https://www.pivotaltracker.com/story/show/175386922


cmd/state/internal/cmdtree/internal/middleware/befaft/befaft.go
```diff
@@ -0,0 +1,37 @@
+package befaft
```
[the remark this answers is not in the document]
daved: I'll surely make it longer, but this is due to the admonishment to keep pkg names short like "bufio" instead of "bufferinputoutput". :trollface: 

Also, general names like "events" leads to packages with expansive and ever-expanding dependencies. This is why "runners" contains specific packages like "run". In this case, the corollary to "runners" is "middleware".

internal/runners/run/event.go
```diff
@@ -0,0 +1,22 @@
+package run
```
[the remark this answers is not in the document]
daved: IIUC, the events call a script in the same exact manner as run calls a script, with the difference being the trigger. I felt that calling runners without the entanglements of the original cobra/command logic was a major benefit of the current design, so I've not understood your aversion to doing so. The caveat that jumps out to me is if a runner does something with that can create instability if called "out of order" (which we should probably identify and add warnings to the relevant package comments).

From here, if you still don't agree, I suppose I'd put this in another package and duplicate some of the runner logic.

daved:
The scriptrun unit tests are failing at the moment, so that will need to be fixed. I assume that means some artifact tests will be failing as well. Once the overall direction of this PR is settled, I'll address the failing tests and mark as non-WIP.

activestate.yaml
```diff
@@ -159,3 +162,6 @@ events:
   - name: file-changed
     scope: [ "locale" ]
     value: packr
+  - name: after-command
+    scope: ["scripts"]
+    value: display
```
Naatan, line 167: Remove your testing code from the activestate.yaml please

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -1,6 +1,7 @@
 package cmdtree
 
 import (
+	"github.com/ActiveState/cli/cmd/state/internal/cmdtree/internal/intercepts/cmdcall"
```
Naatan, line 4: I think you may have done something wrong here. 2 internals?

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,52 @@
...
+	return &CmdCall{
+		primer: p,
+	}
+}
+
+// InterceptExec handles the before command logic, calls the next ExecuteFunc,
+// and then handles the after command logic.
+func (c *CmdCall) InterceptExec(next captain.ExecuteFunc) captain.ExecuteFunc {
+	return func(cmd *captain.Command, args []string) error {
+		cc := cmdcall.New(c.primer, cmd.UseFull())
```
Naatan, line 27: What's with the recursion? We are already in a CmdCall.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,52 @@
...
+
+		if err := cc.Run(project.BeforeCmd); err != nil {
+			return locale.WrapError(
+				err, "err_intercept_cmdcall_before",
+				"before-command event run failure",
```
Naatan, line 32: This is not a friendly user facing message.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,52 @@
...
+			return locale.WrapError(
+				err, "err_intercept_cmdcall_next",
+				"before/after-command next func failure",
+			)
+		}
+
+		if err := cc.Run(project.AfterCmd); err != nil {
+			return locale.WrapError(
+				err, "err_intercept_cmdcall_after",
+				"after-command event run failure",
```
Naatan, line 46: This is not a friendly user facing message.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,52 @@
...
+			return locale.WrapError(
+				err, "err_intercept_cmdcall_before",
+				"before-command event run failure",
+			)
+		}
+
+		if err := next(cmd, args); err != nil {
+			return locale.WrapError(
+				err, "err_intercept_cmdcall_next",
+				"before/after-command next func failure",
```
Naatan, line 39: For this one let's just return the `err`. We'll sort the locale wrapping issue separately.

cmd/state/internal/cmdtree/push.go
```diff
@@ -20,7 +20,10 @@ func newPushCommand(prime *primer.Values) *captain.Command {
 		[]*captain.Flag{},
 		[]*captain.Argument{},
 		func(ccmd *captain.Command, args []string) error {
-			return pushRunner.Run()
+			if fail := pushRunner.Run(); fail != nil {
+				return fail
+			}
+			return nil
```
Naatan, line 26: You won't need this after you have the `next()` call just return the error. The error handling in `main` accounts for runners that return failures.

internal/captain/command.go
```diff
@@ -260,14 +267,30 @@ func (c *Command) Flags() []*Flag {
...
 	return c.arguments
 }
 
+func (c *Command) SetInterceptChain(fns ...InterceptFunc) {
+	c.interceptChain = fns
+}
+
+func (c *Command) interceptFunc() InterceptFunc {
+	return func(fn ExecuteFunc) ExecuteFunc {
+		for i := len(c.interceptChain) - 1; i >= 0; i-- {
```
Naatan, line 284: What's wrong with `for _, intercept := range c.interceptChain` ?

This is a nit really. Just seems needlessly verbose.

internal/events/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,94 @@
...
+		out:      p.Output(),
+		proj:     p.Project(),
+		subshell: p.Subshell(),
+		cmdList:  cmdList,
+		p:        p,
+	}
+}
+
+// Run executes the event handling logic by running any relevant scripts.
+func (cc *CmdCall) Run(t project.EventType) error {
```
Naatan, line 43: Use your words. Rename `t` please.

internal/events/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,94 @@
...
+
+	var events []*project.Event
+	for _, event := range cc.proj.Events() {
+		if event.Name() != string(t) {
+			continue
+		}
+
+		scopes, err := event.Scope()
+		if err != nil {
+			return err // TODO: this
```
Naatan, line 58: Resolve TODO please

internal/events/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	if len(events) == 0 {
+		return nil
+	}
+
+	for _, event := range events {
+		val, err := event.Value()
+		if err != nil {
+			return err // TODO: this
```
Naatan, line 76: Resolve TODO please

internal/events/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		ss := strings.Split(val, " ")
+		if len(ss) == 0 {
+			return errors.New("no script defined") // TODO: this
```
Naatan, line 81: Resolve TODO please

internal/events/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		if !scriptrun.ProjectHasScript(cc.proj, ss[0]) {
```
Naatan, line 84: I feel like `RunScript` should be doing this?

internal/events/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,94 @@
...
+			continue
+		}
+
+		if err := scriptrun.RunScript(cc.out, cc.subshell, cc.proj, ss[0], ss[1:]); err != nil {
+			return err // TODO: this
```
Naatan, line 89: Resolve TODO please

internal/scriptrun/scriptrun.go
```diff
@@ -0,0 +1,197 @@
...
+	FailStandaloneConflict = failures.Type("run.fail.standaloneconflict", failures.FailUser)
+	// FailExecNotFound indicates when the builtin language exec is not available
+	FailExecNotFound = failures.Type("run.fail.execnotfound", failures.FailUser)
+)
+
+// ProjectHasScript is a helper function to determine if a project contains a
+// script by name.
+func ProjectHasScript(proj *project.Project, name string) bool {
+	script := proj.ScriptByName(name)
+	return script != nil
```
Naatan, line 36: This is pointless busywork. Please delete this function and just use `proj.ScriptByName(name) != nil`.

internal/scriptrun/scriptrun.go
```diff
@@ -0,0 +1,197 @@
...
+		return locale.WrapError(err, "err_run_scriptval", "Could not get script value.")
+	}
+
+	sf, fail := scriptfile.New(lang, script.Name(), scriptBlock)
+	if fail != nil {
+		return fail.WithDescription("error_state_run_setup_scriptfile")
+	}
+	defer sf.Clean()
+
+	out.Notice(output.Heading(locale.Tl("script_output", "Script Output")))
```
Naatan, line 144: So it's still going to print all this stuff even when it's just running events?

Naatan (changes requested):

cmd/state/internal/cmdtree/cmdtree.go
daved: The reason for this is because there's two important scopes here - the command call event handling, and the construction of the command call event intercept function. Keeping this in a separate package is my preference to help keep tabs on dependencies/isolation. My second preference is to have the type in the cmdtree package since that's the only place that will use the intercepts.

My third preference (with a significant preference against) is to have it in the internal/events/cmdcall package. Having the intercept there would be like having a captain command in a runner.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
daved: Same as other concern. https://github.com/ActiveState/cli/pull/1052#discussion_r515408992

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
daved: The reason these are not informative is that they should not be user-facing at all, really. This is a workaround since the current errors handling panics if no l10n is set.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
daved: This will panic as mentioned here: https://github.com/ActiveState/cli/pull/1052#discussion_r515409315

internal/captain/command.go
daved: The loop must iterate in reverse or the nesting order will be backward.

internal/events/cmdcall/cmdcall.go
daved: Yes, it also does that check. I did this for now to see if we want to skip when the script isn't present, or error out. For instance, if there are 3 events with different scripts, do we want to halt on script 2 if it doesn't exist, or should 1 and 3 be run with 2 being skipped?

internal/scriptrun/scriptrun.go
daved: This is needed by the run runner so that it only prints the title if it is somewhat reasonable that the script will run correctly.

internal/scriptrun/scriptrun.go
daved: I wasn't sure about this part - what should be printed for different cases. Depending on the level of detail needed, we will end up needing to maintain two very similar functions.  Alternately, we can provide a silent outputer, or accept the output (wouldn't the user want to know the script output?).

cmd/state/internal/cmdtree/cmdtree.go
Naatan: Just put it in `"github.com/ActiveState/cli/cmd/state/internal/cmdtree/intercepts/cmdcall"`

The double internal is the issue.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
Naatan: I'm not sure I understand why the location of the package, which is addressed in the comment you linked, is related to my concern here?

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
Naatan: This is an error raised by event handling specifically isn't it? Not sure I understand why this would not be user facing.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
Naatan: Why? It would either return the error from the runner or the error from the before/after event handling, either of which would be localized.

internal/events/cmdcall/cmdcall.go
Naatan: We should halt as soon as there is an error. I don't see what this check has to do with that though.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
daved: The cmdcall.New is setting up a command call event handler, and the cmdcall struct in this package is the command call intercept context.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
daved: It shouldn't be user-facing - this scope is only relevant to our own understanding of control flow. This issue exists apart from event handling and is due to our error handling panicking if a localization is not found. I've not investigated the error l10n handling to figure out what is happening.

internal/events/cmdcall/cmdcall.go
daved: In the run runner, if the script is not able to be found, it never prints the title. This is a similar check that, minimally, ensures the script exists before running the script itself. Since we want to error out, the check is useless now and will be removed.

cmd/state/internal/cmdtree/internal/intercepts/cmdcall/cmdcall.go
daved: Marking as resolved to be discussed in other threads.

cmd/state/internal/cmdtree/push.go
daved: This is needed or a nil return from pushRunner.Run will be a typed nil and pollute/confuse inspection down the line.

activestate.yaml
Naatan: It's back again..

cmd/state/internal/cmdtree/projects.go
```diff
@@ -20,7 +20,10 @@ func newProjectsCommand(prime *primer.Values) *captain.Command {
 		[]*captain.Flag{},
 		[]*captain.Argument{},
 		func(ccmd *captain.Command, args []string) error {
-			return runner.Run()
+			if fail := runner.Run(); fail != nil {
+				return locale.NewError("tmp_err_runner_projects", fail.Error())
+			}
+			return nil
```
Naatan, line 26: You shouldn't need this. The error handling in the main package handles both types. If the middleware is depending on this that should be handled in the middleware because there'll likely be a few more runners that are returning failures.

internal/scriptrun/scriptrun.go
```diff
@@ -82,14 +80,15 @@ func (s *ScriptRunner) PrepareVirtualEnv() error {
...
+// Run executes the script after ensuring the environment is prepared.
+func (s *ScriptRun) Run(script *project.Script, args []string) error {
 	if s.project == nil {
 		return locale.NewInputError("err_no_projectfile")
 	}
 
 	// Determine which project script to run based on the given script name.
 	if script == nil {
-		return locale.NewInputError("error_state_run_unknown_name", "Script does not exist: {{.V0}}.", script.Name())
+		return locale.NewInputError("error_state_run_unknown_name", "Script was not provided.")
```
Naatan, line 91:
```suggestion
		return locale.NewInputError("error_state_run_unknown_name", "Requested script does not exist.")
```

cmd/state/internal/cmdtree/push.go
```diff
@@ -20,7 +20,10 @@ func newPushCommand(prime *primer.Values) *captain.Command {
 		[]*captain.Flag{},
 		[]*captain.Argument{},
 		func(ccmd *captain.Command, args []string) error {
-			return pushRunner.Run()
+			if fail := pushRunner.Run(); fail != nil {
+				return locale.NewError("tmp_err_runner_push", fail.Error())
+			}
+			return nil
```
Naatan, line 26: Again, shouldn't need this.

internal/events/cmdcall/cmdcall.go
```diff
@@ -0,0 +1,104 @@
...
+		proj:      p.Project(),
+		subshell:  p.Subshell(),
+		cmdList:   cmdList,
+		p:         p,
+		scriptrun: scriptrun.New(p.Output(), p.Subshell(), p.Project()),
+	}
+}
+
+// Run executes the event handling logic by running any relevant scripts.
+func (cc *CmdCall) Run(eventType project.EventType) error {
```
Naatan, line 45: Please include the event name in your localized errors here. Localized error should ideally be actionable. The current errors are very vague and won't tell the user much.

Naatan (changes requested):

cmd/state/internal/cmdtree/push.go
daved: Consolidating with other thread.

cmd/state/internal/cmdtree/projects.go
daved: The issue with error chain handling that affected localized errors, also affects failures. The difference being that dealing with failures will be large enough of an issue, I highly suggest we go with this workaround and file a bug for smoothing this out in the short-term. (Projects and Push are the only two runners that return failures.)

I have two ideas (for the separate story):
1. Since failures can be shown to the user, failures should implement locale.ErrorLocalizer and possibly have an Unwrap method.
2. The main package handling of failures needs to reworked a bit to be able to unlink the first failure from the chain (though this would complicate the control flow of the current unwrapError function regarding existing localized error wrapping).

The above are just my current thoughts. I'm not sure how "in opposition" the above are, or if they may be needed in some combination.

cmd/state/internal/cmdtree/intercepts/cmdcall/cmdcall.go
```diff
@@ -31,6 +32,15 @@ func (c *CmdCall) InterceptExec(next captain.ExecuteFunc) captain.ExecuteFunc {
...
 
 		if err := next(cmd, args); err != nil {
+			// check can be removed when no runners return failures and,
+			// possibly, when main pkg error handling digs through the error
+			// chain to get to nested failures
+			if fail, ok := err.(*failures.Failure); ok {
+				if fail == nil {
+					return nil
+				}
+				return fail // wrapping would break error handling higher in stack
```
Naatan, line 42: Why? The code in `main` know how to handle failures.

internal/events/cmdcall/cmdcall.go
```diff
@@ -59,7 +59,8 @@ func (cc *CmdCall) Run(eventType project.EventType) error {
 		if err != nil {
 			return locale.WrapError(
 				err, "cmdcall_event_err_get_scope",
-				"Cannot obtain scopes from event",
+				"Cannot obtain scopes for event '{{.V0}}'",
```
Naatan, line 62:
```suggestion
				"Cannot obtain scopes for event '[NOTICE]{{.V0}}[/RESET]'",
```

internal/events/cmdcall/cmdcall.go
```diff
@@ -80,22 +81,26 @@ func (cc *CmdCall) Run(eventType project.EventType) error {
 		if err != nil {
 			return locale.WrapError(
 				err, "cmdcall_event_err_get_value",
-				"Cannot get triggered event value",
+				"Cannot get triggered event value for event '{{.V0}}'",
```
Naatan, line 84:
```suggestion
				"Cannot get triggered event value for event '[NOTICE]{{.V0}}[/RESET]'",
```

internal/events/cmdcall/cmdcall.go
```diff
@@ -80,22 +81,26 @@ func (cc *CmdCall) Run(eventType project.EventType) error {
...
+				event.Name(),
 			)
 		}
 
 		ss := strings.Split(val, " ")
 		if len(ss) == 0 {
 			return locale.NewError(
 				"cmdcall_event_err_get_value",
-				"Triggered event value is empty",
+				"Triggered event value is empty for event '{{.V0}}'",
```
Naatan, line 93:
```suggestion
				"Triggered event value is empty for event '[NOTICE]{{.V0}}[/RESET]'",
```

internal/events/cmdcall/cmdcall.go
```diff
@@ -80,22 +81,26 @@ func (cc *CmdCall) Run(eventType project.EventType) error {
...
 			)
 		}
 
-		if err := cc.scriptrun.Run(cc.proj.ScriptByName(ss[0]), ss[1:]); err != nil {
+		scriptName, scriptArgs := ss[0], ss[1:]
+		if err := cc.scriptrun.Run(cc.proj.ScriptByName(scriptName), scriptArgs); err != nil {
 			return locale.WrapError(
 				err, "cmdcall_event_err_script_run",
-				"Failure running defined script",
+				"Failure running defined script '{{.V0}}' for event '{{.V1}}'",
```
Naatan, line 102:
```suggestion
				"Failure running defined script '[NOTICE]{{.V0}}[/RESET]' for event '[NOTICE]{{.V1}}[/RESET]'",
```

Naatan (changes requested):

cmd/state/internal/cmdtree/intercepts/cmdcall/cmdcall.go
daved: The main error handling can handle a failure passed directly back to it, but it does not handle nested failures. I was correct in my suspicion that failures could implement locale.ErrorLocalizer and things would be eased.

The "ToError" calls I added are the shortest forms of ensuring a typed-nil is not propagated up the stack.

The auth client methods had some knock-on effects, but I capped them. The similar package-level function still uses the old functionality for now.

daved: The use of FailureType.Wrap() makes it too awkward to treat failures much like errors, or even equals in satisfying our own interfaces. This means I'm stuck at the main pkg error handling not handling nested failures correctly (especially with multiple failures in a chain).

Naatan: > but it does not handle nested failures

This is not true. It does handle nested failures. If you are running into a case where it doesn't then that's a bug and I'd like to hear more about it.

daved: FTR; Communicated in Slack with nothing to report for now.

Naatan (approved):

PR #1053, opened 2020-10-28
Update anonymous commit id for as.yaml

mdrohmann:
https://www.pivotaltracker.com/story/show/175462168


mdrohmann:
I am not sure if I fixed the bug correctly, because the description said, that there was a situation where the URL still included the `owner/name` namespace but the `commitID` was updated.  I could not re-produce that.  It was either no update or the correct update.

But with this change the commitID will be updated when a build is still in progress.

Naatan (changes requested):
This is changing behaviour and does not address the bug. If you can't repro please bounce it back at me.

Let's close this PR. I'll see if I can get a better repro.

mdrohmann:
Okay.  I'll close this, add a comment to the story and unstart it.

PR #1055, opened 2020-10-28
State revert

MDrakos:
https://www.pivotaltracker.com/story/show/172213823


test/integration/package_int_test.go
```diff
@@ -313,29 +314,37 @@ func (suite *PackageIntegrationTestSuite) TestPackage_operation() {
...
+	// cp.ExpectRe("(?:Package removed|The project is currently building)")
+	// cp.Wait()
+
+	cp = ts.Spawn("revert", firstCommit)
+	cp.SendLine("y")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("history")
+	cp.ExpectExitCode(0)
+	cp.Expect("Description: Reverting to commit")
```
mdrohmann, line 347: I believe you will have to switch these two lines.

When you wait for the exit code first, all the output will have been parsed already.

pkg/cmdlets/commit/commit.go
```diff
@@ -0,0 +1,67 @@
+package commit
```
MDrakos, line 1: I'm not confident in the naming of this package or where it lives. Let me know if this shared printing logic should be moved

test/integration/package_int_test.go
```diff
@@ -313,29 +314,38 @@ func (suite *PackageIntegrationTestSuite) TestPackage_operation() {
...
+	cp.ExpectExitCode(0)
 
-	suite.Run("packages update", func() {
-		cp := ts.Spawn("packages", "update", "dateparser@0.7.6")
-		cp.ExpectRe("(?:Package updated|The project is currently building)")
-		cp.Wait()
-	})
+	// Get the first commitID we find, which should be the first commit for the project
+	commitRe := regexp.MustCompile(`[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}`)
+	firstCommit := commitRe.FindString(cp.TrimmedSnapshot())
```
MDrakos, line 328: Went this route for getting the commit to revert to. Created a story here: https://www.pivotaltracker.com/story/show/175487029 to follow up on the output of `state history` and now `state revert`

Naatan (dismissed):
Looks good from an architectural pov. I'll leave the in-depth review to martin seeing as he was already flagged.

internal/runners/revert/revert.go
```diff
@@ -0,0 +1,107 @@
...
+	if !containsCommitID(history, commitID) {
+		return locale.NewError("err_revert_invalid_commit_id", "Commit ID: {{.V0}} does not belong to the project: {{.V1}}", params.CommitID, r.project.Namespace().String())
+	}
+
+	count, fail := model.CommitsBehindLatest(r.project.Owner(), r.project.Name(), r.project.CommitID())
+	if fail != nil {
+		return locale.WrapError(fail.ToError(), "err_revert_commits_behind", "Could not determine if local project is synchronized with platform")
+	}
+	if count > 0 {
+		return locale.NewInputError("err_revert_behind_latest", "Your project is {{.V0}} commit(s) behind. Please run `state pull` to syncronize your project and run `state revert` again", strconv.Itoa(count))
```
mdrohmann, line 73: ``
```suggestion
		return locale.NewInputError("err_revert_behind_latest", "Your project is {{.V0}} commit(s) behind. Please run `state pull` to synchronize your project and run `state revert` again", strconv.Itoa(count))
```

internal/runners/revert/revert.go
```diff
@@ -0,0 +1,107 @@
...
+		return locale.WrapError(fail.ToError(), "err_revert_get_history", "Could not get project commit history")
+	}
+	if !containsCommitID(history, commitID) {
+		return locale.NewError("err_revert_invalid_commit_id", "Commit ID: {{.V0}} does not belong to the project: {{.V1}}", params.CommitID, r.project.Namespace().String())
+	}
+
+	count, fail := model.CommitsBehindLatest(r.project.Owner(), r.project.Name(), r.project.CommitID())
+	if fail != nil {
+		return locale.WrapError(fail.ToError(), "err_revert_commits_behind", "Could not determine if local project is synchronized with platform")
+	}
```
mdrohmann, line 71: Just wondering:  Do we actually need to run all these checks?  I would expect them to happen on the backend already.  This, of course, makes it easier to write out helpful and localized messages, and before asking the user to confirm. So, I am not opposed to the choice.  I am just wondering.  Is the dashboard doing the same?

internal/runners/history/history_test.go
```diff
@@ -22,7 +21,7 @@ func Test_printCommits(t *testing.T) {
 		name        string
 		args        args
 		wantStrings []string
-		wantFailure *failures.Failure
+		wantFailure error
```
mdrohmann, line 24: nit: maybe we can get rid of the `wantFailure` argument, as it is unused in this test...

mdrohmann (changes requested):
Looks great, just found a typo and had a question.

internal/runners/revert/revert.go
MDrakos: Yes the backend will return an error in both cases I've covered above. I thought this would be a good way to provide more contextual errors but I think it's doing too much. I've updated the error handling. If a user gives a commit ID that covers the two cases above (and possibly more) they will receive:

```
Could not revert to commit: <commitID> please ensure that the local project is synchronized with the platform and that the given commit ID belongs to the current project
Error occurred while trying to update a branch: Commit parent doesn't match branch commit
```

This is a fairly long error string but provides some remedies.

internal/runners/history/history_test.go
MDrakos: `wantFailure` is used, if it is nil in a test that we want no errors in we continue with the output checks

PR #1057, opened 2020-10-28
Languages update

MDrakos:
https://www.pivotaltracker.com/story/show/175405071

### DIFF
--- a/internal/runners/languages/install.go
+++ b/internal/runners/languages/install.go
@@ -9,20 +9,25 @@ import (
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/project"
 )
 
 type Update struct {
-	out output.Outputer
+	out     output.Outputer
+	project *project.Project
 }
 
-func NewUpdate(prime primer.Outputer) *Update {
-	return &Update{prime.Output()}
+type primeable interface {
+	primer.Projecter
+	primer.Outputer
+}
+
+func NewUpdate(prime primeable) *Update {
+	return &Update{prime.Output(), prime.Project()}
 }
 
 type UpdateParams struct {
-	Owner       string
-	ProjectName string
-	Language    string
+	Language string
 }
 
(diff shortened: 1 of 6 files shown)


internal/runners/languages/install.go
```diff
@@ -53,10 +51,7 @@ func (u *Update) Run(params *UpdateParams) error {
 
 func parseLanguage(langName string) (*model.Language, error) {
 	if !strings.Contains(langName, "@") {
-		return &model.Language{
-			Name:    langName,
-			Version: "",
-		}, nil
+		return nil, locale.NewInputError("err_languages_parameter", "Invalid parameter for language argument, must be of the form <language>@<version>")
```
Naatan, line 54: `@version` is still optional. Remember the point of this story is consistent behaviour with `state install`, in which it is also optional.

Naatan: I think what you're touching on here is the difference in behaviour when installing vs when updating. When updating you should supply a specific version. But when you're just installing it should just detect the version if not supplied.

I think we should probably just say "X is already installed" if you run `install ..` without specifying a version. @rawktron thoughts?

Naatan (changes requested):

internal/runners/languages/install.go
MDrakos: It seems like `state install python` should say `python<version> is already installed` if run in a python project. If run in a project with a different language it will replace it with the latest python?

And then `state install python@version` will update the version of python if in a python project. Otherwise it will install(replace) with that specific version.

Just want to confirm what the desired behaviour is before I make the changes.

Naatan: Yes that sounds right to me. But for the first one it should also suggest that they can update, eg.

```
python is already installed, you can update it by running python@<version>
```

MDrakos: The dashboard does not allow switching the language of a project, and we already have an existing check for this. I have left that part as is.

Now `state install python` in a python project will return the error you suggested @Naatan. 
`state install python@<version>` will install the version requested.
`state install python` in a Perl project will return an error stating that you cannot switch the language of a project.

Naatan (approved):

PR #1062, opened 2020-10-29
Update state platforms add argument

MDrakos:
https://www.pivotaltracker.com/story/show/175462736

### DIFF
--- a/internal/runners/platforms/platforms.go
+++ b/internal/runners/platforms/platforms.go
@@ -92,9 +92,9 @@ func splitNameAndVersion(input string) (string, string) {
 }
 
 func prepareLatestVersion(params Params) (Params, error) {
-	platformUUID, fail := model.HostPlatformToPlatformID(params.Name)
-	if fail != nil {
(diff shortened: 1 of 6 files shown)


internal/runners/platforms/platforms.go
```diff
@@ -71,9 +75,77 @@ func prepareParams(ps Params) (Params, error) {
...
+}
+
+// platformSemver covers the semvers used by the platform that do not conform
+// to the semver standard. (ie. they have an extra numerical value after patch)
+type platformSemver struct {
+	semver     semver.Version
+	additional int
+}
+
+func latestPlatformVersion(name string) (string, error) {
```
Naatan, line 107: This is doing too much. This type of logic belongs on the solver. For now if they don't supply a version just use the constants here:

http://github.com/ActiveState/cli/blob/c91da98bb5a4f39f6823f713ac381d68b3581745/internal/constants/constants.go#L184-L184

Naatan (changes requested):

Naatan (approved):

PR #1063, opened 2020-10-29
State shim should not call itself recursively

mdrohmann:
I disabled the removal of the global bin path, because it is actually not
necessary anymore if the --default flag specifies the associated project
explicitly.

 https://www.pivotaltracker.com/story/show/175487411

### DIFF
--- a/internal/runners/shim/shim.go
+++ b/internal/runners/shim/shim.go
@@ -5,6 +5,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/failures"
 	"github.com/ActiveState/cli/internal/language"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
@@ -31,6 +33,10 @@ type primeable interface {
 	primer.Projecter
 }
 
+type Params struct {
+	Path string
+}
+
 func New(prime primeable) *Shim {
 	return &Shim{
 		prime.Subshell(),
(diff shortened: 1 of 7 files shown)


internal/runners/shim/shim.go
```diff
@@ -39,35 +45,50 @@ func New(prime primeable) *Shim {
...
+	if err != nil {
+		return err
+	}
+	logging.Debug("Trying to shim %s on PATH=%s", args[0], env["PATH"])
+	// Ensure that we are not calling the shim recursively
+	oldval, ok := env[constants.ShimEnvVarName]
+	if ok && oldval == args[0] {
+		return locale.NewError("err_shim_recursive_loop", "Could not resolve shimmed executable {{.V0}}", args[0])
+	}
+	env[constants.ShimEnvVarName] = args[0]
```
Naatan, line 88: Seems we can drop this

Naatan (changes requested):

internal/runners/shim/shim.go
mdrohmann: I would advocate for leaving it in.  I can still think of cases, where we could run into this issue eg.,

1. Project at path `p` provides executable `e`
1. User runs `state activate --default` => `e` is shimmed at `<bindir>/e`
1. User updates project by removing the artifact that provides `e`
1. User calls `e` resolving to `<bindir>/e` => using project path `p`, but the actual executable `e` is removed => we try `<bindir>/e`

That is only one case that I could think of right now.   There may be others, and this check is fairly simple and effective.

Naatan (approved):

PR #1065, opened 2020-10-30
Convert headless commit to project

mdrohmann:
It is still blocked by an API change, but I figured we can start reviewing it already...

 https://www.pivotaltracker.com/story/show/175404422


internal/runners/initialize/init.go
```diff
@@ -32,22 +33,24 @@ type RunParams struct {
...
 func New(prime primeable) *Initialize {
-	return &Initialize{prime.Output()}
+	return &Initialize{prime.Output(), prime.Project()}
 }
 
 func prepare(params *RunParams) error {
 	if params.Language == "" {
 		// Manually check for language requirement, because we need to fallback on the --language flag to support editor.V0
-		return failures.FailUserInput.New(locale.T("err_init_no_language"))
+		return locale.NewInputError("err_init_no_language", "You need to supply the [NOTICE]language[/RESET] argument, run '[ACTIONABLE]state init --help[/RESET]' for more information.")
```
Naatan, line 53:
```suggestion
		return locale.NewInputError("err_init_no_language", "You need to supply the [NOTICE]language[/RESET] argument, run `[ACTIONABLE]state init --help[/RESET]` for more information.")
```

internal/runners/initialize/init.go
```diff
@@ -103,11 +106,55 @@ func prepare(params *RunParams) error {
...
+	return path, nil
+}
+
+func run(params *RunParams, out output.Outputer, proj *project.Project) (string, error) {
+	// try to initialize project at --path if provided
+	if params.Path != "" {
+		var fail *failures.Failure
+		proj, fail = project.Parse(filepath.Join(params.Path, constants.ConfigFileName))
+		if fail != nil {
+			logging.Debug("No project file parsed at path: %v", fail.ToError())
```
Naatan, line 150: You are assuming a failure means that the file doesn't exist. Please be explicit.

internal/runners/initialize/init.go
```diff
@@ -103,11 +106,55 @@ func prepare(params *RunParams) error {
...
+		if err != nil {
+			return "", errs.Wrap(err, "Could not determine current working directory.")
+		}
+	}
+
+	out.Notice(locale.Tr(
+		"init_success",
+		params.Namespace.Project, params.Namespace.Owner, path))
+	return path, nil
+}
```
Naatan, line 142: I don't think this needs its own function. The logic is 90% the same, the only difference with the headless commit is that you seed a commit ID. So the logic between headless and not headless should just be the same with one single if conditional that sets the commit.

internal/runners/push/push.go
```diff
@@ -64,28 +66,36 @@ func (r *Push) Run() *failures.Failure {
 		return nil
 	}
 
-	lang, langVersion, fail := r.languageForProject(r.project)
-	if fail != nil {
-		return fail
-	}
+	// try to create the project at the given commit id.
+	if r.project.CommitID() != "" {
+		err := model.CreateProjectAtCommit(r.project.Owner(), r.project.Name(), r.project.Private(), r.project.CommitUUID())
```
Naatan, line 71: Rather than having two specialized functions can we de-specialize the existing one. Same as before the logic here is pretty much identical. Just make a function that creates a project and another one that attaches a commit. This diff should be tiny.

Naatan (changes requested):

internal/runners/initialize/init.go
mdrohmann: IMO the other initialization branch is quite different, as it does not have a project file yet, so it needs to be created from scratch.  As a result, there is not need to modify the project file later (as I do in the `runHeadless` function by changing namespace and commitID here).

The only part that I would re-use is the output of the success message in the end.  But I decided that it was not worth it, to introduce a big `if headless { } else { }` block in the `run` method.  Do you want me to do that here though?

Naatan (changes requested):
See comments

mdrohmann:
@Naatan It is pretty much ready for review now.  The integration test is run as part of the "Critical Integration Test" jobs, and fails at the point where it should: when the API disallows to set the commitID.

internal/runners/initialize/init.go
```diff
@@ -47,22 +46,24 @@ func New(prime primeable) *Initialize {
...
+		params.language = language.Supported{language.MakeByName(langParts[0])}
+		if !params.language.Recognized() {
+			return language.NewUnrecognizedLanguageError(
+				params.language.String(),
+				language.RecognizedSupportedsNames(),
+			)
+		}
 	}
 
 	// Fail if target dir already has an activestate.yaml
```
mdrohmann: The block below would then need to be de-activated `if headless`

Naatan (changes requested):
I pushed some changes that get it more in the direction I was thinking, which helped me realize it's difficult to achieve without significantly more refactoring. So I landed it at a middle ground. The way we're handling params in this runner and in project creations is very awkward, but let's not go down that rabbit hole now.

Please see if these latest changes make sense.

internal/runners/initialize/init.go
```diff
@@ -111,79 +115,62 @@ func (r *Initialize) Run(params *RunParams) error {
...
 
-	if convertHeadless {
-		headlessCommitID := proj.CommitID()
+	logging.Debug("Init: %s/%s %v", params.Namespace.Owner, params.Namespace.Project, params.Private)
 
-		err := proj.Source().SetNamespace(params.Namespace.Owner, params.Namespace.Project)
-		if err != nil {
-			return "", errs.Wrap(err, "Could not set namespace in project file.")
-		}
+	createParams := &projectfile.CreateParams{
```
mdrohmann: Interesting!  But do we really want to replace the entire `activestate.yaml` with a vanilla one at this point?  The user might have added changes during the headless edits that they want to keep...

internal/runners/push/push.go
```diff
@@ -35,76 +34,106 @@ func NewPush(config *viper.Viper, prime primeable) *Push {
...
 	}
 
 	if r.project != nil && r.project.IsHeadless() {
-		return failures.FailUserInput.Wrap(locale.NewInputError("err_push_headless", "You must first create a project. Please visit {{.V0}} to create your project.", r.project.URL()))
+		return errs.AddTips(locale.NewInputError(
+			"err_push_headless",
+			"You must first create a project."),
+			locale.Tl("push_headless_push_tip_conversion_url", "Visit [ACTIONABLE]{{.V0}}[/RESET] to create a project in the browser.", r.project.URL()),
+			locale.Tl("push_headless_push_tip_state_init", "Run [ACTIONABLE]state init[/RESET] to create a project with the State Tool.."),
+		)
```
Naatan, line 48: Very nice!

internal/runners/push/push.go
```diff
@@ -35,76 +34,106 @@ func NewPush(config *viper.Viper, prime primeable) *Push {
...
+		return locale.WrapError(fail.ToError(), "push_project_create_empty_err", "Failed to create an project {{.V0}}.", r.project.Namespace().String())
 	}
+	if r.project.CommitID() != "" {
+		// try to create the project at the given commit id.
+		err := model.UpdateProjectBranchCommit(pjm, r.project.CommitUUID())
+		if err != nil {
+			return locale.WrapError(err, "push_project_branch_commit_err", "Failed to update new project {{.V0}} to current commitID.", r.project.Namespace().String())
+		}
+	} else {
+		_, commitID, fail := model.InitializeProject(r.project.Owner(), r.project.Name(), model.HostPlatform, lang, langVersion)
```
Naatan, line 87: You already did this on line 76. What should happen here is creating the commit. The update branch call (currently on line 82) would ALWAYS happen regardless of headless state.

Naatan (changes requested):

internal/runners/push/push.go
mdrohmann: I pushed a change to address that.  Note, that `CreateEmptyProject` does not call `InitializeProject` anymore, as you suggested in your comment.

internal/runners/initialize/init.go
Naatan: `prepare` isn't called in `if headless`.

internal/runners/initialize/init.go
Naatan: I think you were looking at an earlier commit. Please refresh.

pkg/platform/model/projects.go
```diff
@@ -120,13 +120,8 @@ func DefaultBranchForProject(pj *mono_models.Project) (*mono_models.Branch, *fai
...
 
-// CreateProject will create the project on the platform
-func CreateProject(owner, name, hostPlatform string, lang *language.Supported, langVersion string, private bool) (*mono_models.Project, strfmt.UUID, *failures.Failure) {
-	_, fail := CreateEmptyProject(owner, name, private)
-	if fail != nil {
-		return nil, "", fail
-	}
-
+// InitializeProject will create the project on the platform
+func InitializeProject(hostPlatform string, lang *language.Supported, langVersion string) (strfmt.UUID, *failures.Failure) {
```
Naatan, line 124: This function should be dropped now. The requirement stuff can live in the init function. CommitInitial is what should be called from init.

Naatan (changes requested):

internal/runners/initialize/init.go
```diff
@@ -125,7 +125,9 @@ func run(params *RunParams, out output.Outputer) (string, error) {
 	logging.Debug("Init: %s/%s %v", params.Namespace.Owner, params.Namespace.Project, params.Private)
 
 	if isHeadless {
+		commitID := proj.CommitID()
 		proj.Source().SetNamespace(params.Namespace.Owner, params.Namespace.Project)
+		proj.Source().SetCommit(commitID, false)
```
Naatan, line 130: Why? The commit ID would already be set. Is `SetNamespace` overwriting the commitID? If so that should be addressed.

Naatan (changes requested):

Naatan (approved):

Naatan (approved):

PR #1066, opened 2020-10-30
add instrumentation for Carlisle's journey

mdrohmann:
https://www.pivotaltracker.com/story/show/175405120


Naatan (changes requested):
Please implement this in captain. Similar to http://github.com/ActiveState/cli/blob/28b561de163121caff52997572c3cbfc5b4b8ace/internal/captain/command.go#L384-L384

Naatan (changes requested):
I was only asking for the analytics to be in captain. Not refactor the main error handling to be part of captain. Please keep it simple. This story should maybe have 10 lines of diff.

mdrohmann:
But how am I supposed to get the exit code for the analytics in captain?

mdrohmann:
I have to unwrap it from the error, and so I am reusing the function that we already had.  I didn't change anything in there, just moved it from `cmd/state` to `internal/captain`.

Naatan:
@mdrohmann let's move `unwrapExitCode` (and nothing else) into the `errs` package. Then we can call that from both captain and main.

We can look into combining those calls long term, but for now the overhead isn't really a concern.

mdrohmann:
I left it in `internal/captain/errors.go`, because when I move it to `internal/errs` I get a cyclical dependency error.

Naatan (commented):
Pushed some small changes to get it landed since this is a blocker for release and it's late on Friday.

Naatan (approved):

PR #1068, opened 2020-11-02
Fix forwarding durning update

MDrakos:
https://www.pivotaltracker.com/story/show/175208661


mdrohmann (dismissed):
LGTM!

internal/updater/check.go
```diff
@@ -98,6 +99,9 @@ func AutoUpdate(pjPath string, out output.Outputer) (updated bool, resultVersion
 		if os.IsPermission(errs.InnerError(err)) {
 			out.Error(locale.T("auto_update_permission_err"))
 		}
+		if errors.As(err, new(*osutils.AlreadyLockedError)) {
+			out.Notice(locale.T("auto_update_already_updating"))
```
Naatan, line 103: Auto update is implicit. The user did not ask to be updated. Showing them this notice would only add friction.

Let's instead ensure that the error is logged if it isn't an error we can recognize, but otherwise not expose this to the user.

Naatan (changes requested):

internal/updater/check.go
MDrakos: The error will be logged here if it is not a permissions error, in which case the error is reported to the user. Do we want to remove this behaviour as well?

I've left the existing log line in `updater.go` as it could provide some value information in debugging.

Naatan: Yeah let's not log it (as an error anyway), since this is "expected" behaviour.

MDrakos: Okay changed it to a debug log

internal/updater/check.go
```diff
@@ -98,7 +98,7 @@ func AutoUpdate(pjPath string, out output.Outputer) (updated bool, resultVersion
 		if os.IsPermission(errs.InnerError(err)) {
 			out.Error(locale.T("auto_update_permission_err"))
 		}
-		logging.Error("Unable to self update: %s", err)
+		logging.Debug("Unable to self update: %s", err)
```
Naatan, line 101: Now all errors are disregarded. I was saying you still identify the cause of the error, and only if you cannot identify the cause do you log it as an error.

Naatan (changes requested):

internal/updater/check.go
```diff
@@ -95,10 +96,15 @@ func AutoUpdate(pjPath string, out output.Outputer) (updated bool, resultVersion
 	logging.Debug("Self-updating.")
 	err = update.Run(out)
 	if err != nil {
+		log := logging.Error
 		if os.IsPermission(errs.InnerError(err)) {
 			out.Error(locale.T("auto_update_permission_err"))
+			log = logging.Debug
```
Naatan, line 102:
```suggestion
```

Naatan (changes requested):

Naatan (approved):

PR #1069, opened 2020-11-03
Cache only activation

mdrohmann:
https://www.pivotaltracker.com/story/show/175447071

### DIFF
--- a/assets/activestate.yaml.python.tpl
+++ b/assets/activestate.yaml.python.tpl
@@ -2,6 +2,7 @@ scripts:
   - name: activationMessage
     language: {{.Language}}
     value: |
+      # -*- coding: utf-8 -*-
       import textwrap
       print(textwrap.dedent("""
         You are now in an activated state, which is like a virtual environment to work
--- a/pkg/platform/runtime/alternative.go
+++ b/pkg/platform/runtime/alternative.go
@@ -20,7 +20,6 @@ import (
 	"github.com/ActiveState/cli/internal/hash"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/rtutils"
 	"github.com/ActiveState/cli/internal/unarchiver"
 	"github.com/ActiveState/cli/pkg/platform/runtime/envdef"
 )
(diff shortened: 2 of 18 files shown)


assets/activestate.yaml.python.tpl
```diff
@@ -2,6 +2,7 @@ scripts:
   - name: activationMessage
     language: {{.Language}}
     value: |
+      # -*- coding: utf-8 -*-
```
mdrohmann, line 5: This needs to be here for python2.  It otherwise barks when we try to print non-ascii characters, and doesn't print the activation message.

pkg/platform/runtime/alternative.go
```diff
@@ -25,42 +25,73 @@ import (
...
 	ArtifactID strfmt.UUID
 	Files      []string
 }
 
-// NewAlternativeRuntime returns a new alternative runtime assembler
+// NewAlternativeEnv returns a new alternative runtime environment
+func NewAlternativeEnv(cacheDir string) (*AlternativeEnv, *failures.Failure) {
+	if rtutils.BuiltViaCI {
+		return nil, failures.FailRuntime.New("Alternative builds are not yet supported. Please contact support if you are seeing this error.")
+	}
```
Naatan, line 55: We should probably drop this. Not sure why we're not already running into this.

pkg/platform/runtime/installer.go
```diff
@@ -130,23 +170,28 @@ func (installer *Installer) Assembler() (Assembler, *failures.Failure) {
...
 }
 
 // InstallArtifacts installs all artifacts provided by a runtime assembler
 func (installer *Installer) InstallArtifacts(runtimeAssembler Assembler) (envGetter EnvGetter, freshInstallation bool, fail *failures.Failure) {
+	err := installer.runtime.StoreBuildEngine(runtimeAssembler.BuildEngine())
+	if err != nil {
+		return nil, false, failures.FailRuntime.Wrap(err, "TODO")
+	}
+
 	if runtimeAssembler.IsInstalled() {
```
Naatan, line 195: I think we'll probably want to mark the installation as complete here as well, so projects that were activated before this change get picked up.

Naatan (changes requested):

Naatan (dismissed):

Naatan (approved):

PR #1075, opened 2020-11-03
Ensure that we have the activation events necessary to measure success

Naatan:
New event category: activation
Possible event actions:
 - start -- right at the start of activation, before anything happens
 - before-subshell -- right before the subshell is created (this would be our “success state”, as we cannot reliably appropriate subshell activations to successes)
 - error - an error happened during activation that warrants further investigation (via rollbar, not actionable on analytics beyond signalling the type of exit)
 - user-input-error -- activation failed due to user input (eg. they cancelled, entered an invalid path, etc.)
 - user-exit-error -- subshell exited with non-zero. This isn’t an issue we need to look into (eg. user ran “exit 1")

New dimension: projectName (eg. ActiveState/Perl-5.32)

Using this data you can assert eg.

Succesful activations - cat: activation, action: before-subshell, project dimension: ActiveState/Perl-5.32
Activation failures - # of activation starts minus # of before-subshell events

### DIFF
--- a/cmd/state/internal/cmdtree/activate.go
+++ b/cmd/state/internal/cmdtree/activate.go
@@ -1,6 +1,10 @@
 package cmdtree
 
 import (
+	"errors"
+	"os/exec"
+
+	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/captain"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/primer"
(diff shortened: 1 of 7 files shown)


cmd/state/internal/cmdtree/activate.go
```diff
@@ -68,10 +72,28 @@ func newActivateCommand(prime *primer.Values) *captain.Command {
...
+						analytics.Event(analytics.CatActivationFlow, "user-exit-error")
+					}
+				}
+			}
+
+			return err
 		},
 	)
 	cmd.SetGroup(EnvironmentGroup)
-	cmd.SetDeferAnalytics(true)
```
mdrohmann, line 75: We still have this analytics event in the activate runner: https://github.com/ActiveState/cli/blob/master/internal/runners/activate/activate.go#L121

So, it would be send twice.  We should probably remove that now, no?

internal/captain/command.go
```diff
@@ -416,7 +417,7 @@ func (c *Command) runner(cobraCmd *cobra.Command, args []string) error {
 	err := c.execute(c, args)
 	if !c.deferAnalytics {
 		exitCode := errs.UnwrapExitCode(failures.ToError(err))
-		analytics.EventWithValue(analytics.CatCommandExit, subCommandString, int64(exitCode))
+		analytics.EventWithLabel(analytics.CatCommandExit, subCommandString, strconv.Itoa(exitCode))
```
mdrohmann, line 420: Does `EventWithValue` not work at all?

mdrohmann (changes requested):

internal/captain/command.go
Naatan: It does, I was mistaken. There's 4 levels: category, action, label, value. Value isn't really useful for exit codes, since we want to group by these results and "Value" fields aren't really made for that.

cmd/state/internal/cmdtree/activate.go
Naatan: Good point! Will remove that.

internal/captain/command.go
mdrohmann: Ah! That makes sense...

mdrohmann (approved):

PR #1077, opened 2020-11-04
Ensure build_engine file exists after installation

mdrohmann:
https://www.pivotaltracker.com/story/show/175579507

### DIFF
--- a/pkg/platform/runtime/installer.go
+++ b/pkg/platform/runtime/installer.go
@@ -187,16 +187,17 @@ func (installer *Installer) Assembler() (Assembler, *failures.Failure) {
 
 // InstallArtifacts installs all artifacts provided by a runtime assembler
 func (installer *Installer) InstallArtifacts(runtimeAssembler Assembler) (envGetter EnvGetter, freshInstallation bool, fail *failures.Failure) {
-	err := installer.runtime.StoreBuildEngine(runtimeAssembler.BuildEngine())
-	if err != nil {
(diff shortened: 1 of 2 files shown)


mdrohmann:
It turns out that the `build_engine` file was created too early, at least for a camel-type deploy.  During the installation process it was removed.  So, I wrote it another time.  But it should never have been a user-facing error, because when an activation from cache failed, it always tried a regular installation:

- https://github.com/ActiveState/cli/blob/d493b5876ff84ce1a7950bdfacae1f0d2ad15ed3/pkg/platform/runtime/installer.go#L100-L104
- https://github.com/ActiveState/cli/blob/d493b5876ff84ce1a7950bdfacae1f0d2ad15ed3/pkg/platform/runtime/installer.go#L115-L118

All rollbar errors seem to have gone through that route logging ["Failed to retrieve cached assembler"](https://rollbar.com/activestate/state-tool/items/1224/?item_page=0&item_count=100&#instances)

Naatan:
@mdrohmann do we still need both `installer.runtime.StoreBuildEngine` calls?

mdrohmann:
I was looking into that too right now.  We could move the first one down into the `if runtimeAssembler.Installed()` block, as it is really only there for the situation that I was concerned about before:  The user had already a functioning cache, but no `build_engine` file yet.  Yes, I think that would be the correct way to do it.

Naatan (approved):

PR #1080, opened 2020-11-05
check for existence of project for package operations

mdrohmann:
https://www.pivotaltracker.com/story/show/175600293

### DIFF
--- a/locale/en-us.yaml
+++ b/locale/en-us.yaml
@@ -61,8 +61,6 @@ err_language_not_yet_supported:
   other: "The Platform responded with an artifact that is not supported by this version of the State Tool, artifact name: {{.V0}}. Is your State Tool up to date?"
 error_missing_arg:
   other: "Argument missing: [NOTICE]{{.V0}}[/RESET]"
-err_no_projectfile:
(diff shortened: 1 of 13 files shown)


locale/en-us.yaml
```diff
@@ -1574,3 +1574,5 @@ prepare_protocol_warning:
   other: Could not add state protocol
 confirm:
   other: Please Confirm
+package_operation_no_project:
+  other: Could not find a project. You can create a new project with [ACTIONABLE]state init[/RESET] or run this command from a directory that is in an existing project.
```
Naatan, line 1578:
```suggestion
  other: Could not find a project. You can create a new project with [ACTIONABLE]state init[/RESET] or run this command from a directory that is in an existing project. You can also mark a project as the default to use by running `[ACTIONABLE]state activate --default[/RESET]` inside your project directory.
```

Naatan (changes requested):

mdrohmann:
@Naatan I just realized that the same problem occurs with `state run`.  This seems to be a common mistake that we make.  Is it okay if I review all our runners for this story and unify the error message for them?

Naatan (approved):

PR #1081, opened 2020-11-05
Table output

MDrakos:
https://www.pivotaltracker.com/story/show/175505266


MDrakos:
@Naatan it's still in a rough state and does not address all of the ACs yet but some of the structure is there.

internal/table/table.go
```diff
@@ -0,0 +1,167 @@
...
+type FormatFunc func(string, ...interface{}) string
+
+type Table struct {
+	rows   [][]string
+	header []string
+	// TODO: Make column width part of a row struct?
+	columnWidths []int
+	padding      int
+	// TODO: Add logic to deal with max column size
+	maxSize int
```
Naatan, line 25: These seem like values that should be passed through functions rather than recorded to the struct

internal/table/table.go
```diff
@@ -0,0 +1,167 @@
...
+		b.WriteString(str)
+	}
+	return b
+}
+
+func (b *writeBuffer) String() string {
+	return b.String()
+}
+
+func Create(data interface{}) *Table {
```
Naatan, line 50: Why is this taking an `interface{}` value? Converting types is the job of the plain outputter. This package should just be dealing with opinionated data.

internal/table/table_test.go
```diff
@@ -0,0 +1,28 @@
...
+		ErrWriter:   os.Stderr,
+		Colored:     true,
+		Interactive: true,
+	})
+	if fail != nil {
+		t.Fatal(fail)
+	}
+	table := Create([][]string{{"some", "stuff"}})
+	table.AddHeader([]string{"First", "Second"})
+	table.WithHeaderFormat("[HEADING]%s[/RESET]")
```
Naatan, line 22: This feels a little awkward to me. The rows and columns are a bit implicit and the headers are added only later on. I'd prefer something a bit more structured like:

```
package main

type Table struct {
	Headers []string
	Rows []Row
	Padding int
}

type Row struct {
	Columns []string
}

func main() {
	_ = Table{
		[]string{"Project", "Owner"},
		[]Row{
			Row{[]string{"cli", "ActiveState"}},
			Row{[]string{"/tmp/cli"}}, // note only one column, we should infer that therefore it would span
		},
		5,
	}
}
```

Naatan (changes requested):

Naatan:
@mdrohmann I'd like to have spent more time with this but it's near EOD on Friday and my vacation is starting. Nothing particularly "wrong" with the code, just would have liked to clean it up more. Your math skills might see some easy wins here, please feel free to nitpick this PR more than you otherwise would.

internal/runners/projects/projects.go
```diff
@@ -16,7 +17,31 @@ import (
...
+		r = append(r, projectOutputPlain{v.Name, v.Organization, strings.Join(checkouts, "\n")})
+	}
+	return r
+}
+
+type projectOutputPlain struct {
+	Name           string
+	Organization   string
+	LocalCheckouts string `locale:"local_checkouts,Local Checkouts" opts:"emptyNil,separateLine"`
 }
```
MDrakos, line 45: Can you move this type definition into the `MarshalOutput` function? It doesn't appear to be used anywhere else

internal/table/table.go
```diff
@@ -0,0 +1,190 @@
...
+	for _, row := range t.rows {
+		out += renderRow(row.columns, colWidths) + linebreak
+	}
+
+	return strings.TrimRight(out, linebreak)
+}
+
+func (t *Table) calculateWidth(maxTotalWidth int) ([]int, int) {
+	// Calculate required width of each column
+	minWidth := padding * 2
```
MDrakos, line 60: We are using this `padding * 2` value in a lot of places. Why not just have the padding be 4?

internal/table/table.go
```diff
@@ -0,0 +1,190 @@
...
+					colWidths[n] = rowValueSize // Set width according to column size
+				}
+			}
+			if spanWidth > minWidth {
+				minWidth = spanWidth
+			}
+		}
+
+		// Add padding and update the total width so far
+		colWidths[n] += (padding * 2)
```
MDrakos, line 89:
```suggestion
		colWidths[n] += padding * 2
```

internal/table/table.go
```diff
@@ -0,0 +1,190 @@
...
+		total = minWidth
+	}
+
+	// Limit to max width
+	if total > maxTotalWidth {
+		total = maxTotalWidth
+	}
+
+	// Calculate column widths according to the total width
+	calculatedTotal := 0
```
MDrakos, line 115: It looks like this isn't being used. Maybe we could use it to verify the `calulcatedTotal` is the same, or within tolerance of the original total. That or just remove it.

internal/table/table.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+		}
+
+		// Add padding and update the total width so far
+		colWidths[n] += (padding * 2)
+		columnTotal += colWidths[n]
+	}
+
+	// Equalize widths by 20% of average width
+	// This is to prevent columns that are much larger than others from taking up most of the table width
```
MDrakos, line 94: This function is starting to get pretty long and harder to grok. Is there any way we could split up some of this work?

internal/table/table.go
```diff
@@ -0,0 +1,190 @@
...
+	// don't want to modify the provided slice
+	columns := make([]string, len(providedColumns))
+	copy(columns, providedColumns)
+
+	result := ""
+
+	// Keep rendering lines until there's no column data left to render
+	for len(strings.Join(columns, "")) != 0 {
+		// Iterate over the columns by their line sizes
+		for n, maxlen := range colWidths {
```
MDrakos, line 141: nit: `maxLen`

MDrakos (commented):
Overall looks good to me and makes sense. Just some minor comments. I can't approve so I will leave it up to you.

internal/table/table.go
mdrohmann: It is chosen that way because the padding happens on both sides of the text (left and right).  At some point we might want to change it that the first and/or last column are handled differently actually (such that they are padded on side only)...

mdrohmann:
@MDrakos I believe that I addressed all your comments.  Please let me know if you find anything else, or if I missed something.

MDrakos:
@mdrohmann LGTM!

PR #1082, opened 2020-11-05
Fix panic when command has typo

MDrakos:
https://www.pivotaltracker.com/story/show/175546722

### DIFF
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -17,6 +17,10 @@ constants:
       GOPATH="${GOPATH:=`go env GOPATH`}"
       GOROOT="${GOROOT:=`go env GOROOT`}"
       PATH="${GOROOT}/bin:${GOPATH}/bin:${PATH}"
+      if [ "$CI" == "true" ]
+      then
+        GO_BUILD_TAGS="CI"
+      fi
 scripts:
   - name: preprocess
     language: bash
--- a/cmd/state/errors.go
+++ b/cmd/state/errors.go
@@ -95,9 +95,7 @@ func unwrapError(err error) (int, error) {
 	// unwrap exit code before we remove un-localized wrapped errors from err variable
 	code := errs.UnwrapExitCode(err)
 
-	// Account for low level erros returned from 3rd party libraries that are not localized
-	unwrappedErrs := locale.UnwrapError(err)
-	if !locale.HasError(err) && isErrs && !hasMarshaller && len(unwrappedErrs) != 0 {
+	if !locale.HasError(err) && isErrs && !hasMarshaller {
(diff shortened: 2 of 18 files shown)


cmd/state/errors.go
```diff
@@ -95,7 +95,9 @@ func unwrapError(err error) (int, error) {
 	// unwrap exit code before we remove un-localized wrapped errors from err variable
 	code := errs.UnwrapExitCode(err)
 
-	if !locale.HasError(err) && isErrs && !hasMarshaller {
+	// Account for low level erros returned from 3rd party libraries that are not localized
+	unwrappedErrs := locale.UnwrapError(err)
+	if !locale.HasError(err) && isErrs && !hasMarshaller && len(unwrappedErrs) != 0 {
```
Naatan, line 100: This makes no sense. The first condition says `err` is has no localization in its chain, but the last condition is saying it needs to have at least one localization in its chain. I'm not sure what problem this is trying to solve, but please revisit.

activestate.yaml
```diff
@@ -17,6 +17,10 @@ constants:
       GOPATH="${GOPATH:=`go env GOPATH`}"
       GOROOT="${GOROOT:=`go env GOROOT`}"
       PATH="${GOROOT}/bin:${GOPATH}/bin:${PATH}"
+      if [ "$CI" == "true" ]
+      then
+        GO_BUILD_TAGS="CI"
+      fi
```
Naatan, line 23: How is this related to this PR?

Naatan (changes requested):

activestate.yaml
MDrakos: Indirectly based on the comment here: https://github.com/ActiveState/cli/blob/master/cmd/state/errors.go#L101. Since this error was reported from a state tool version that was built on CI.

cmd/state/errors.go
MDrakos: The underlying issue comes from adding tips to all errors. If the error is from cobra, adding a tip changes the error type and `isErrs` is now true.

I've added a fix that doesn't add tips to errors when there is no child command found since that will happen in this case, however other errors returned form cobra will have this problem.

We could add a check that the error does not implement our `errs.Error` interface before adding tips, but I was not sure if that would be appropriate.

Naatan (changes requested):
Seems the issue here is that we're not wrapping cobra errors properly. We have a function for that here:

http://github.com/ActiveState/cli/blob/c5b3649adfe08bfb238b648394da507ac97fa3e5/internal/captain/command.go#L480-L480

The current fix is too detached from the root cause. If I can't tell from the fixed code what the problem was then odds are we're not fixing it the right way.

PR #1083, opened 2020-11-05
Install.sh should not fail inside activated environment

mdrohmann:
The problem here was that the install scripts called `state export config
--filter=dir`.

Whether needed or not (`state export config` does not need it), we are always
trying to open a project file, and in an activated environment it first tries
to resolve the project file from the environment variable `ACTIVESTATE_PROJECT`
and fails with the "corrupted environment" error that bubbled up in the installation scripts - if that file does not exist.

This PR changes that behavior and attempts other ways to resolve a project file
(from working directory and the default directory)

https://www.pivotaltracker.com/story/show/175600566


pkg/projectfile/projectfile.go
```diff
@@ -824,7 +824,9 @@ func GetProjectFilePath() (string, *failures.Failure) {
 	for _, getProjectFilePath := range lookup {
 		path, fail := getProjectFilePath()
 		if fail != nil {
-			return "", fail
+			// log error and try the next function
+			logging.Error("Failed to get project file: %v", fail.ToError())
+			continue
```
Naatan, line 829: The contract here is that a failure should be raised. If the issue is considered benign then the `getProjectFile..` function would not be returning an error at all. This change is adding complexity to solve a problem that's unrelated to this function.

I think the current behaviour is proper; if your environment is corrupted it should be raised to the user so they can fix it. The problem we're fixing is that this might happen during installation. Maybe for the `state export` call we could ensure in the installer it's called with a clean environment and from a temp path that won't have a project file?

Naatan (changes requested):

pkg/projectfile/projectfile.go
mdrohmann: That sounds good.

installers/install.sh
```diff
@@ -18,7 +18,9 @@ Flags:
  --activate-default <project>    Activate a project and make it the system default
  -h                              Show usage information (what you're currently reading)
 EOF
-`
+
+# ignore project file if we are already in an activated environment
+export ACTIVESTATE_PROJECT=""
```
Naatan, line 23:
```suggestion
unset ACTIVESTATE_PROJECT
```

Naatan (changes requested):

Naatan (dismissed):

Naatan (approved):

PR #1086, opened 2020-11-05
Remove pkg add integration test as the results are too unreliable for integration tests

Naatan:


test/integration/push_int_test.go
```diff
@@ -73,15 +73,6 @@ func (suite *PushIntegrationTestSuite) TestCarlisle() {
 	cp = ts.Spawn("auth", "logout")
 	cp.ExpectExitCode(0)
 
-	// anonymous commit
-	wd := filepath.Join(cp.WorkDirectory(), namespace)
```
MDrakos, line 77: Think we need to keep this line, judging by the CI errors

MDrakos (approved):

PR #1088, opened 2020-11-06
WIP: Update deferred analytics handling

daved:
https://www.pivotaltracker.com/story/show/175546571

 https://www.pivotaltracker.com/story/show/175546571


daved:
If I understand correctly, we'll be dropping the pageview type. That should make marshaling/unmarshaling easier since there will only be one datatype to deal with in the stored list.

The deferred/immediate sending logic is sketched out in client.go for now and the current review requested is for general oversight with that sketch being central to my thoughts/concerns.

There is a lot of code changes in analytics.go to ensure the control flow is protected, but the general logic is the same and calling code should mostly look the same.

PR #1091, opened 2020-11-06
Add tip to 'cannot activate activated' error

daved:
https://www.pivotaltracker.com/story/show/175600387

 https://www.pivotaltracker.com/story/show/175600387

### DIFF
--- a/internal/runners/activate/activate.go
+++ b/internal/runners/activate/activate.go
@@ -72,8 +72,17 @@ func (r *Activate) run(params *ActivateParams) error {
 	alreadyActivated := subshell.IsActivated()
 	if alreadyActivated {
 		if !params.Default {
-			return locale.NewInputError("err_already_activated", "You cannot activate a new project when you are already in an activated state.")
+			err := locale.NewInputError("err_already_activated",
+				"You cannot activate a new project when you "+
+					"are already in an activated state.",
+			)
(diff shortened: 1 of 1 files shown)


internal/runners/activate/activate.go
```diff
@@ -72,8 +72,17 @@ func (r *Activate) run(params *ActivateParams) error {
...
-			return locale.NewInputError("err_already_activated", "You cannot activate a new project when you are already in an activated state.")
+			err := locale.NewInputError("err_already_activated",
+				"You cannot activate a new project when you "+
+					"are already in an activated state.",
+			)
+			tipMsg := locale.Tl("err_tip_exit_activated",
+				"To exit your activated state simply close your "+
+					"current shell by running [ACTIONABLE]exit[/RESET].",
+			)
+			return errs.AddTips(err, tipMsg)
```
Naatan, line 83: The tip and the error are not the same:

https://www.pivotaltracker.com/n/projects/2203557/stories/175600387

Naatan (changes requested):

internal/runners/activate/activate.go
daved: Sorry, I read that wrong. Fixed!

Naatan (approved):

PR #1093, opened 2020-11-06
wait for analytics calls to finish in background

mdrohmann:
https://www.pivotaltracker.com/story/show/175624819


internal/analytics/analytics.go
```diff
@@ -80,10 +81,18 @@ func (d *customDimensions) toMap() map[string]string {
...
+var (
+	eventWaitGroup sync.WaitGroup
+)
+
 func init() {
 	setup()
 }
 
+func WaitForAllEvents() {
+	eventWaitGroup.Wait()
```
Naatan, line 93: Could we add a timeout to this? Let's wait for at most a second. I don't want slow connections or firewalls to cause commands to just hang. And loosing out on events from those outliers should be pretty negligible.

Naatan (changes requested):

internal/analytics/analytics.go
mdrohmann: To add the timeout, I had to change our google-analytics library for this PR: 

So, we would need to merge this one first: https://github.com/ActiveState/go-ogle-analytics/pull/1, and then update the go.mod dependencies here.

internal/analytics/deferred.go
```diff
@@ -21,6 +21,18 @@ type deferredData struct {
 
 const deferredCfgKey = "deferrer_analytics"
 
+func SetDeferred(da bool) {
```
Naatan, line 24: @daved ran into this while addressing this story so added it here; `sendDeferred` was being called for commands where defer=true.

cmd/state/main.go
```diff
@@ -82,6 +84,7 @@ func main() {
 		}
 	}
 
+	analytics.WaitForAllEvents(time.Second * 1)
```
daved, line 87: nit: I suggest putting this at the end of the captain command runner wrap that calls SetDefault (just to centralize the logic).

daved: Also nit: 2 or 3 more seconds should cover most circumstances where there is a long request hanging, but should not feel too bad to users (it should rarely happen).

internal/analytics/analytics.go
```diff
@@ -82,11 +84,31 @@ func (d *customDimensions) toMap() map[string]string {
...
 	CustomDimensions = &customDimensions{}
 	setup()
 }
 
+// WaitForAllEvents waits for all events to return
+func WaitForAllEvents(t time.Duration) {
+	wg := make(chan bool)
+	go func() {
+		eventWaitGroup.Wait()
+		wg <- true
```
daved, line 101: nit/not-nit:
```suggestion
		defer close(wg)
		eventWaitGroup.Wait()
```
This ensures resource cleanup.

internal/analytics/deferred.go
```diff
@@ -21,6 +21,18 @@ type deferredData struct {
...
+	deferAnalytics = da
+	if deferAnalytics {
+		return
+	}
+	go func() {
+		if err := sendDeferred(sendEvent); err != nil {
+			logging.Errorf("Could not send deferred events: %v", err)
+		}
+	}()
```
daved, line 33:
```suggestion
	eventWaitGroup.Add(1)
	go func() {
		defer eventWaitGroup.Done()
		if err := sendDeferred(sendEvent); err != nil {
			logging.Errorf("Could not send deferred events: %v", err)
		}
	}()
```
To be thorough (take as a nit if you'd like).

daved (changes requested):

internal/analytics/analytics.go
daved: Also, it's idiomatic to use a `chan struct{}` since `struct{}` is a zero-sized primitive.

daved (approved):

PR #1094, opened 2020-11-06
Fix shim error looking for projectfile

MDrakos:
On Windows if a user were to run `state activate some/project --default`, leave that window open in an activated state, and then run a shimmed command in another cmd shell they would encounter the `Project file not found` error. After exiting the activated state in the original window the command would work as expected.

This may be due to the file handle being open by the original command prompt so when we search for the config file and eventually call `stat` on it we get an error. 

We already have successfully retrieved the project at this point so we can use it.

 https://www.pivotaltracker.com/story/show/175607429


mdrohmann (approved):

Naatan (changes requested):
Hang on. What is this fixing? This looks like a workaround not a fix.

Naatan:
I'll elaborate; something appears to be broken with `projectfile.Get()` that is affecting shim. Let's find what that something is and fix it. 

Again if I can't tell from your code alone what problem it is trying to fix then you're probably not fixing the right problem.

internal/globaldefault/default.go
```diff
@@ -105,7 +106,7 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 
 	cfg.Set(constants.GlobalDefaultPrefname, projectPath)
 
-	return nil
+	return cfg.WriteConfig()
```
Naatan, line 109: Wrap

pkg/projectfile/projectfile.go
```diff
@@ -822,17 +822,20 @@ func GetProjectFilePath() (string, *failures.Failure) {
...
+			logging.Debug("Failed to find project file path: %v", fail)
 		}
+
 		if path != "" {
 			return path, nil
 		}
 	}
 
-	return "", FailNoProject.New(locale.T("err_no_projectfile"))
+	return "", FailNoProject.Wrap(fail, locale.T("err_no_projectfile"))
```
Naatan, line 838: This would cause red herrings. The failure you're wrapping is not what caused the error.

Naatan (changes requested):

pkg/projectfile/projectfile.go
```diff
@@ -822,17 +822,20 @@ func GetProjectFilePath() (string, *failures.Failure) {
...
 		getProjectFilePathFromDefault,
 	}
+	var path string
+	var fail *failures.Failure
 	for _, getProjectFilePath := range lookup {
-		path, fail := getProjectFilePath()
+		path, fail = getProjectFilePath()
 		if fail != nil {
-			return "", fail
+			logging.Debug("Failed to find project file path: %v", fail)
```
mdrohmann, line 830: This would change our current control flow:  If an ACTIVESTATE_PROJECT environment variable is set, but the file is not found, it would not be an error anymore.  @Naatan convinced me in a different PR, that we probably shouldn't do that.

So, is the error actually because we didn't write the default config path early enough to the config file?  In that case, adding the `cfg.WriteConfig()` to the global default setup routine should already take care of the problem, no?

internal/globaldefault/default.go
```diff
@@ -104,6 +105,10 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 	}
 
 	cfg.Set(constants.GlobalDefaultPrefname, projectPath)
+	err = cfg.WriteConfig()
+	if err != nil {
+		return locale.WrapError(err, "err_write_config", "Could not write to configuration file")
+	}
```
Naatan, line 111: Don't need this one, the one in the activation runner will suffice.

Naatan (changes requested):

pkg/projectfile/projectfile.go
MDrakos: Yes, confirmed that updating `cfg.WriteConfig()` with the original loop will work.

Naatan (approved):

PR #1101, opened 2020-11-10
Fix activation after update

daved:
https://www.pivotaltracker.com/story/show/175607307

 https://www.pivotaltracker.com/story/show/175607307


daved:
@Naatan I ended up adding a new structure (process.Activation) in order to minimize the work required by callers.

The scripts included for testing the updates saved me a tremendous amount of time, so I've included them (fupdstor was written on my own time, and setmod-by-hrs was a script I already had and added a minor update to).

Also, I ran into an issue where the file generated by the PIDLock type is not removed if the process is killed via SIGKILL. I'll file a bug if the source of this issue is not within the current changeset.

internal/updater/check.go
```diff
@@ -56,23 +56,23 @@ type UpdateResult struct {
...
-	// Check for an update, but timeout after one second.
+	// Check for an update, but timeout
 	logging.Debug("Checking for updates.")
 	update := Updater{
 		CurrentVersion: constants.Version,
 		APIURL:         constants.APIUpdateURL,
 		CmdName:        constants.CommandName,
 	}
-	seconds := 1
+	seconds := 2
```
Naatan, line 75: Undo this please. Guess you did this for your tests, but I did not ask for this to be increased nor would I want it to. If there's an issue here please file a bug.

Naatan (changes requested):
Looking good! Just one minor issue.

Naatan (approved):

PR #1103, opened 2020-11-10
manage progress bar width based on terminal width

mdrohmann:
https://www.pivotaltracker.com/story/show/175650853

### DIFF
--- a/internal/progress/progress.go
+++ b/internal/progress/progress.go
@@ -10,9 +10,12 @@ package progress
 import (
 	"context"
 	"io"
+	"os"
 
+	"github.com/ActiveState/cli/internal/logging"
 	"github.com/vbauerster/mpb/v4"
 	"github.com/vbauerster/mpb/v4/decor"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 // FileSizeCallback can be called by a task to report that a sub-task of length `fileSize` (in bytes) has been finished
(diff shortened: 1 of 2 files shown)


internal/progress/progress.go
```diff
@@ -47,6 +50,21 @@ func WithDebugOutput(w io.Writer) mpb.ContainerOption {
...
 func New(options ...mpb.ContainerOption) *Progress {
+	tw, _, err := terminal.GetSize(int(os.Stdout.Fd()))
+	if err != nil {
+		logging.Error("Could net get terminal size: %v", err)
+		tw = 120
+	}
+
+	// ensure that we have enough space by modifying the progress bar width
+	// we subtract 26 from the terminal width
+	// * 20 for name and counter to the left of bar
```
MDrakos, line 61: Will the name and counter always be 20 characters?

internal/progress/progress.go
```diff
@@ -47,6 +50,21 @@ func WithDebugOutput(w io.Writer) mpb.ContainerOption {
...
+	// * 5 for percentag
```
MDrakos, line 62:
```suggestion
	// * 5 for percentage
```

MDrakos (approved):

internal/progress/progress.go
mdrohmann: It is a good point. I misunderstood how this works.  It is dependent on the longest on the string.  So, it could change with localization. :(

mdrohmann: I think that I will add functionality to limit the name size based on the terminal.

MDrakos (approved):

PR #1106, opened 2020-11-10
Install scripts activate project even if state tool is installed

mdrohmann:
https://www.pivotaltracker.com/story/show/175650839


Naatan (changes requested):
I want our installers to skip over the install logic and just get to the success state. I don't want a whole separate flow if install doesn't need to take place. Just skip over the install flow and get to the success bit. 

In effect; we should only run the activate call in one place. 

If that creates too much need for refactoring / rearchitecting please let me know, that's not the intend here.

mdrohmann:
It is quite a bit of refactoring. Basically it needs to be organized like the install.ps1 script.  But it should be okay...

Naatan (changes requested):
I asked you to come talk to me if this required significant refactoring. Please go back to the original commit. I'm not against refactoring but if we're going to invest in that we're going to do it right. All this would do now is land us in an intermediary state with its own set of bugs that will then itself have to be refactored again when we decide to make that investment.

Naatan (approved):

PR #1108, opened 2020-11-11
Ignore ExitCode error during state activate

mdrohmann:
https://www.pivotaltracker.com/story/show/175651734

### DIFF
--- a/internal/errs/unwrap.go
+++ b/internal/errs/unwrap.go
@@ -26,9 +26,10 @@ func unwrapExitError(err error) (bool, int) {
 // UnwrapExitCode checks if the given error is a failure of type FailExecCmdExit and
 // returns the ExitCode of the process that failed with this error
 func UnwrapExitCode(errFail error) int {
-	isExitError, code := unwrapExitError(errFail)
+	var eerr interface{ ExitCode() int }
+	isExitError := errors.As(errFail, &eerr)
 	if isExitError {
-		return code
+		return eerr.ExitCode()
 	}
 
 	// failure might be in the error stack
@@ -42,9 +43,9 @@ func UnwrapExitCode(errFail error) int {
 	}
 
 	err := fail.ToError()
-	isExitError, code = unwrapExitError(err)
+	isExitError = errors.As(err, &eerr)
 	if isExitError {
-		return code
+		return eerr.ExitCode()
 	}
 
 	return 1
--- a/internal/errs/unwrap.go
+++ b/internal/errs/unwrap.go
@@ -6,23 +6,6 @@ import (
 	"github.com/ActiveState/cli/internal/failures"
 )
 
-func unwrapExitError(err error) (bool, int) {
-	var eerr interface{ ExitCode() int }
-	isExitError := errors.As(err, &eerr)
-	if isExitError {
-		// If exit error happened in activated shell, do not forward exit code, but return 0
-		var activatedErr interface{ IsFromActivatedShell() }
-		isFromActivatedShell := errors.As(err, &activatedErr)
-		if isFromActivatedShell {
-			return true, 0
-		}
-
-		return true, eerr.ExitCode()
-	}
-
-	return false, 0
-}
-
 // UnwrapExitCode checks if the given error is a failure of type FailExecCmdExit and
 // returns the ExitCode of the process that failed with this error
 func UnwrapExitCode(errFail error) int {
--- a/internal/errs/unwrap.go
+++ b/internal/errs/unwrap.go
@@ -6,13 +6,29 @@ import (
 	"github.com/ActiveState/cli/internal/failures"
 )
 
+func unwrapExitError(err error) (bool, int) {
+	var eerr interface{ ExitCode() int }
+	isExitError := errors.As(err, &eerr)
+	if isExitError {
+		// If exit error happened in activated shell, do not forward exit code, but return 0
+		var activatedErr interface{ IsFromActivatedShell() }
+		isFromActivatedShell := errors.As(err, &activatedErr)
+		if isFromActivatedShell {
+			return true, 0
+		}
+
+		return true, eerr.ExitCode()
+	}
+
+	return false, 0
+}
+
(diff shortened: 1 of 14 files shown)


internal/runners/activate/activation.go
```diff
@@ -19,6 +20,20 @@ import (
...
+	error
+}
+
+func (ae activationError) IsFromActivatedShell() {}
+
+// Is Silent returns true if the wrapped error is an exit code error
+func (ae activationError) IsSilent() bool {
+	var eerr interface{ ExitCode() int }
+	return errors.As(ae, &eerr)
+}
```
Naatan, line 33: This seems too presumptuous. Because an error specifies an ExitCode it is therefore "silent"? This doesn't feel right. Also why do we need both `IsFromActivatedShell` and `IsSilent`?

internal/runners/activate/activation.go
```diff
@@ -19,6 +20,20 @@ import (
 	"github.com/ActiveState/cli/pkg/projectfile"
 )
 
+type activationError struct {
+	error
+}
+
+func (ae activationError) IsFromActivatedShell() {}
```
Naatan, line 27: This feels too specific to our current use-case. Can we name this according to the behaviour it should have, not according to the use-case of the behaviour.

internal/errs/unwrap.go
```diff
@@ -6,13 +6,29 @@ import (
...
+func unwrapExitError(err error) (bool, int) {
+	var eerr interface{ ExitCode() int }
+	isExitError := errors.As(err, &eerr)
+	if isExitError {
+		// If exit error happened in activated shell, do not forward exit code, but return 0
+		var activatedErr interface{ IsFromActivatedShell() }
+		isFromActivatedShell := errors.As(err, &activatedErr)
+		if isFromActivatedShell {
+			return true, 0
+		}
```
Naatan, line 18: This chunk feels unnecessary. We already know it's an exit error. We don't need to know whether it came from an activated state. This just seems to override the exitCode, which shouldn't be necessary; the activate runner should convey this. eg. wrap it in an ExitCode=0 error.

Naatan (changes requested):
Please move the error handling changes (failures > error) into a separate PR. Normally I'd be welcoming of this type of change provided it's not too big, which in this case I don't feel qualifies.

internal/errs/unwrap.go
mdrohmann: That is a good idea.  If you haven't yet, please also consider my comment to the PR: https://www.pivotaltracker.com/story/show/175651734/comments/219641816

After implementing this, I believe that it would be better not to change the error code to 0, but to just silent the error message.  After all, that was what the complaint during the demo was about, and changing the error code changes the scripts behavior.  I am especially concerned about the `state activate -c` use case, that we have in the visual studio code extension.

mdrohmann: > 
> 
> Please move the error handling changes (failures > error) into a separate PR. Normally I'd be welcoming of this type of change provided it's not too big, which in this case I don't feel qualifies.

I actually wanted to explain myself here, before I requested a review: I found these changes warranted, because the exit code handling that was duplicated for failures and errors complicated things quite a bit.  So, I was happy that I could move it all to failures.

Naatan: > That is a good idea. If you haven't yet, please also consider my comment to the PR: https://www.pivotaltracker.com/story/show/175651734/comments/219641816
> 
> After implementing this, I believe that it would be better not to change the error code to 0, but to just silent the error message. After all, that was what the complaint during the demo was about, and changing the error code changes the scripts behavior. I am especially concerned about the `state activate -c` use case, that we have in the visual studio code extension.

I'm on board with this. Agree forwarding the exit code isn't really the issue here, it's the output we provide around it.

Naatan: > I actually wanted to explain myself here, before I requested a review: I found these changes warranted, because the exit code handling that was duplicated for failures and errors complicated things quite a bit. So, I was happy that I could move it all to errors.
> 
> Please let me know how you decide on the PT story comments, and I will implement it without the part that removes the failure handling.

Ok. Let's go with it then, but please tripple check that you didn't miss any failures. Be mindful also of functions that called your function that now returns an error rather than a failure.

internal/runners/activate/activation.go
mdrohmann: I think you are right that we could instead just implement another `ErrorCode()` function that returns zero.

The logic is necessary to implement the "implied" AC that we should not print error messages on non-zero exit codes in the subshell.  Other subshell errors (that are unlikely but could be caused by not finding the shell command or something like this) should IMO still produce error messages.

internal/errs/unwrap.go
```diff
@@ -6,6 +6,23 @@ import (
 	"github.com/ActiveState/cli/internal/failures"
 )
 
+func unwrapExitError(err error) (bool, int) {
```
Naatan, line 9: Forgot to delete this function

Naatan (changes requested):

internal/runners/activate/activation.go
```diff
@@ -67,9 +67,9 @@ func (r *Activate) activateAndWait(proj *project.Project, venv *virtualenvironme
 
 	analytics.Event(analytics.CatActivationFlow, "before-subshell")
 
-	fail := <-r.subshell.Failures()
-	if fail != nil {
-		return locale.WrapError(fail, "error_in_active_subshell", "Failure encountered in active subshell")
+	err = <-r.subshell.Errors()
+	if err != nil {
+		return locale.WrapError(err, "error_in_active_subshell", "Failure encountered in active subshell")
```
mdrohmann, line 72: This is the only place where the "failures" -> "error" change is used in the State Tool.  As I renamed the function, I can't have missed anything.  And therefore, I am 99.99% sure that I am not accidentally wrapping an error in a failure. :)

Naatan (changes requested):
Think you missed my comment? 

https://github.com/ActiveState/cli/pull/1108#discussion_r522338717

Naatan (approved):

PR #1110, opened 2020-11-12
Update prompt color on Windows

MDrakos:
https://www.pivotaltracker.com/story/show/175652184

### DIFF
--- a/internal/prompt/helpers.go
+++ b/internal/prompt/helpers.go
@@ -2,8 +2,10 @@ package prompt
 
 import (
 	"errors"
+	"fmt"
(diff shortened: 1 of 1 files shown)


MDrakos:
On Windows the default color code for cyan (`\e[0;36m`) is darker than what we are used to on Unix systems. See output of ColorTool here:
image
So we can use the higher intensity cyan on Windows to get the colors more in-line

Naatan (changes requested):
Let's just use `cyan+h` on both.

Naatan (approved):

PR #1114, opened 2020-11-12
install scripts: de-activate some instruction messages on activation

mdrohmann:
https://www.pivotaltracker.com/story/show/175605937


installers/install.sh
```diff
@@ -405,7 +409,11 @@ STATEPATH=$INSTALLDIR/$STATEEXE
 CONFIGDIR=$($STATEPATH "export" "config" "--filter=dir")
 echo "install.sh" > $CONFIGDIR/"installsource.txt"
 
-$STATEPATH _prepare || exit $?
+if [ ( -n "${ACTIVATE}" ) -or ( -n "${ACTIVATE_DEFAULT}") ]; then
+  $STATEPATH _prepare 1>/dev/null || exit $?
```
Naatan, line 413: Why /dev/null this? If it exits zero there wouldn't be any output right?

Naatan (changes requested):

installers/install.sh
mdrohmann: It suppresses the message: 

```
To add the global installation directory to your environment please add the line `...` to your RC file
```

mdrohmann: But actually that does not make any sense! https://github.com/ActiveState/cli/blob/master/internal/runners/prepare/prepare.go#L58-L62

This message is printed after we have already tried to do that.  This message should be part of the warning message in case, we cannot fail to the rc file...

mdrohmann: I changed the output in the prepare runner such that it makes more sense IMO.

Naatan (approved):

PR #1116, opened 2020-11-13
Fixing install.sh one-liner

mdrohmann:
https://www.pivotaltracker.com/story/show/175722288


MDrakos (approved):
I've made this mistake as well. That's a tricky variable name. Might be worth a rename or adding an underscore. I'll leave it up to you.

mdrohmann:
I decided to merge right away, as some users are running into that already...  But it is a good idea to add an underscore.

PR #1118, opened 2020-11-13
add a global signal handler for interrupts

mdrohmann:
https://www.pivotaltracker.com/story/show/175721061

### DIFF
--- a/internal/sighandler/sighandler.go
+++ b/internal/sighandler/sighandler.go
@@ -3,44 +3,73 @@ package sighandler
 import (
 	"os"
 	"os/signal"
-	"time"
 
-	"github.com/ActiveState/cli/internal/analytics"
-	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/errs"
 )
 
-var sigCh chan os.Signal
-var ignoreCh chan bool
+// signalStack maintains a stack of signal handlers only the most recent one is active
+var signalStack []signalStacker
 
-// Init initializes the global signal interrupt handler
-func Init(subCommand string) {
-	sigCh = make(chan os.Signal, 1)
-	signal.Notify(sigCh, os.Interrupt)
-	ignoreCh = make(chan bool, 0)
+// Push adds a signal handler to the stack of signal handler
+// It stops all signal handlers that are lower in the stack and makes the current one the only active one.
+func Push(s signalStacker) {
+	// stop currently active signal-handler
+	if len(signalStack) > 0 {
+		signalStack[len(signalStack)-1].Stop()
+	}
+	// add new signal-handler
+	signalStack = append(signalStack, s)
 
-	go func(subCommand string) {
-		defer close(sigCh)
-		defer close(ignoreCh)
-		ignore := false
-		for {
-			select {
-			case <-sigCh:
-				if !ignore {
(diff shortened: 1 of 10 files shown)


mdrohmann:
@Naatan This is how I think it can be done.  As mentioned earlier it is not pretty, as it relies on some globally created channels right now.

There is still an open question on how to implement the `activate, user-interrupt-error` event (AC 4) though: https://www.pivotaltracker.com/story/show/175721061/comments/219706870

internal/subshell/sscommon/sscommon.go
```diff
@@ -177,33 +175,13 @@ func binaryPathCmd(env []string, name string) (string, error) {
 	return split[0], nil
 }
 
-func ignoreInterrupts(ctx context.Context) {
```
Naatan, line 180: I don't think we should be removing this. IIRC it is required so as not to bubble stuff up from the subshell context. For this story we're concerned with interrupts happening outside of subshell.

internal/sighandler/sighandler.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/ActiveState/cli/internal/analytics"
+	"github.com/ActiveState/cli/internal/logging"
+)
+
+var sigCh chan os.Signal
+var ignoreCh chan bool
+
+// Init initializes the global signal interrupt handler
+func Init(subCommand string) {
```
Naatan, line 16: Can we use `New` and return a struct that can be closed. We don't want to still be capturing this after the captain runner has finished.

internal/sighandler/sighandler.go
```diff
@@ -0,0 +1,46 @@
...
+		defer close(ignoreCh)
+		ignore := false
+		for {
+			select {
+			case <-sigCh:
+				if !ignore {
+					logging.Debug("captured ctrl-c event")
+					analytics.EventWithLabel(analytics.CatCommandExit, subCommand, "interrupt")
+					analytics.WaitForAllEvents(time.Second)
+					os.Exit(1)
```
Naatan, line 32: Why are you calling `os.Exit`? Big red flag. We want to bubble up interrupts as errors.

Naatan, line 32: Can we handle this outside this package.

Something like:

```
sh := sighandler.New()
defer sh.Close()
sh.OnInterrupt(func () { ... })
```

Naatan (changes requested):

internal/sighandler/sighandler.go
mdrohmann: I did it this way, because that is the default behavior of `ctrl-c`.  It exits the programme with a non-zero exit code.  Here we just inject our analytics code before.

I could add another channel, and run the `c.execute()` function in a go function, such that I could listen for what returns first: the main path (`c.execute` or the signal handler).  That should be a little nicer.

internal/sighandler/sighandler.go
```diff
@@ -0,0 +1,75 @@
...
+	return nil
+}
+
+type signalStacker interface {
+	Stop()
+	Resume()
+	Close() error
+}
+
+type resumeableSignaler interface {
```
mdrohmann, line 49: remove

daved (approved):

swagile1 (approved):

PR #1119, opened 2020-11-16
Install.sh needs to work in bash

mdrohmann:
https://www.pivotaltracker.com/story/show/175745799

### DIFF
--- a/installers/install.sh
+++ b/installers/install.sh
@@ -377,7 +377,7 @@ manual_installation_instructions() {
 manual_update_instructions() {
   info "State Tool installation complete."
   # skip instruction to source rc file when we are activating
-  if [ -n "${ACTIVATE}" ] || [ "${ACTIVATE_DEFAULT}" ]; then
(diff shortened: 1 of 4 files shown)


installers/install.sh
```diff
@@ -377,7 +377,7 @@ manual_installation_instructions() {
 manual_update_instructions() {
   info "State Tool installation complete."
   # skip instruction to source rc file when we are activating
-  if [ ( -n "${ACTIVATE}" ) -or ( -n "${ACTIVATE_DEFAULT}" ) ]; then
+  if [ -n "${ACTIVATE}" -or -n "${ACTIVATE_DEFAULT}" ]; then
```
MDrakos, line 380:
```suggestion
  if [ -n "${ACTIVATE}" ] || [ "${ACTIVATE_DEFAULT}" ]; then
```
Tested with this an it appears to work as expected.

MDrakos (changes requested):

MDrakos (dismissed):

MDrakos (approved):

PR #1120, opened 2020-11-16
Install via feature name

MDrakos:
https://www.pivotaltracker.com/story/show/175709644

### DIFF
--- a/pkg/platform/model/inventory.go
+++ b/pkg/platform/model/inventory.go
@@ -52,7 +52,7 @@ func IngredientByNameAndVersion(language, name, version string, prefix Namespace
 
 	for _, ingredient := range results {
 		for _, feature := range ingredient.LatestVersion.ProvidedFeatures {
-			if feature.Feature == nil || strings.ToLower(*feature.Feature) != strings.ToLower(name) {
+			if feature.Feature == nil || *feature.Feature != name {
 				continue
 			}
 
@@ -84,7 +84,7 @@ func IngredientWithLatestVersion(language, name string, prefix NamespacePrefix) 
 
 	for _, res := range results {
 		for _, feature := range res.LatestVersion.ProvidedFeatures {
-			if feature.Feature == nil || strings.ToLower(*feature.Feature) != strings.ToLower(name) {
+			if feature.Feature == nil || *feature.Feature != name {
 				continue
 			}
 
--- a/pkg/platform/model/inventory.go
+++ b/pkg/platform/model/inventory.go
@@ -51,17 +51,19 @@ func IngredientByNameAndVersion(language, name, version string, prefix Namespace
 	}
 
 	for _, ingredient := range results {
-		if ingredient.Ingredient.Name == nil || *ingredient.Ingredient.Name != name {
-			continue
-		}
+		for _, feature := range ingredient.LatestVersion.ProvidedFeatures {
+			if feature.Feature == nil || *feature.Feature != name {
+				continue
+			}
 
-		for _, ver := range ingredient.Versions {
-			if ver.Version == version {
-				return &IngredientAndVersion{
-					ingredient.V1SearchIngredientsResponseIngredientsItems,
-					ver.Version,
-					ingredient.Namespace,
-				}, nil
+			for _, ver := range ingredient.Versions {
+				if ver.Version == version {
+					return &IngredientAndVersion{
+						ingredient.V1SearchIngredientsResponseIngredientsItems,
+						ver.Version,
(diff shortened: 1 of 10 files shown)


pkg/platform/model/inventory.go
```diff
@@ -51,17 +51,19 @@ func IngredientByNameAndVersion(language, name, version string, prefix Namespace
 	}
 
 	for _, ingredient := range results {
-		if ingredient.Ingredient.Name == nil || *ingredient.Ingredient.Name != name {
-			continue
-		}
+		for _, feature := range ingredient.LatestVersion.ProvidedFeatures {
```
mdrohmann, line 54: Can we assume that the provided features are the same for all versions of an ingredient?  Or do we have to iterate over all versions and features?

test/integration/package_int_test.go
```diff
@@ -348,26 +348,29 @@ func (suite *PackageIntegrationTestSuite) TestPackage_operation() {
 
 	suite.Run("install", func() {
 		cp := ts.Spawn("install", "dateparser@0.7.2")
-		cp.ExpectRe("(?:Package added|project is currently building)")
+		cp.ExpectRe("(?:Package added|project is currently building)", 60*time.Second)
```
mdrohmann, line 351: Maybe we can use a package with less dependencies here?  Maybe `JSON` and `JSON@4.02`?

test/integration/push_int_test.go
```diff
@@ -82,6 +82,7 @@ func (suite *PushIntegrationTestSuite) TestCarlisle() {
 	cp.Expect("(Y/n)")
 	cp.SendLine("y")
 	cp.Wait()
+	cp.ExpectExitCode(0)
```
mdrohmann, line 85: It makes sense to wait for the exit code here, but then the `cp.Wait()` needs to be removed.
```suggestion
	cp.ExpectExitCode(0)
```

test/integration/push_int_test.go
```diff
@@ -90,7 +91,7 @@ func (suite *PushIntegrationTestSuite) TestCarlisle() {
 	ts.LoginAsPersistentUser()
 
 	// convert to real project
-	cp = ts.SpawnWithOpts(e2e.WithArgs("init", namespace), e2e.WithWorkDirectory(wd))
+	cp = ts.SpawnWithOpts(e2e.WithArgs("init", namespace, "python3"), e2e.WithWorkDirectory(wd))
```
mdrohmann, line 94: When transformed from a headless commit, the language should be inferred.  No need to specify it.

mdrohmann (changes requested):

test/integration/push_int_test.go
MDrakos: Yes you're correct, this is from debugging

pkg/platform/model/inventory.go
```diff
@@ -51,17 +51,19 @@ func IngredientByNameAndVersion(language, name, version string, prefix Namespace
...
+			if feature.Feature == nil || strings.ToLower(*feature.Feature) != strings.ToLower(name) {
```
mdrohmann, line 55: Maybe, let's not yet add the case-insensitive part here...  I have just opened a PR that we can work on after this one lands: https://github.com/ActiveState/cli/pull/1121  I had to park that on Friday because the Ctrl-C story was prioritized.

It involves one more step: preferring the exact casing if both options are available.

pkg/platform/model/inventory.go
```diff
@@ -81,16 +83,17 @@ func IngredientWithLatestVersion(language, name string, prefix NamespacePrefix)
 	}
 
 	for _, res := range results {
-		if res.Ingredient.Name == nil || *res.Ingredient.Name != name {
-			continue
-		}
+		for _, feature := range res.LatestVersion.ProvidedFeatures {
+			if feature.Feature == nil || strings.ToLower(*feature.Feature) != strings.ToLower(name) {
```
mdrohmann, line 87: same here.

pkg/platform/model/inventory.go
MDrakos: `provided_features` is only available via `latest_version` in the search response. `latest_version` should represent the full ingredient version data model as noted in the response here: https://platform.activestate.com/sv/inventory-api-v1/docs/#operation/searchIngredients

test/integration/package_int_test.go
MDrakos: Good idea. I opted for `urllib3` as this is a python project

mdrohmann (approved):

PR #1121, opened 2020-11-17
Allow for case-insensitive package names

mdrohmann:
https://www.pivotaltracker.com/story/show/175469208

### DIFF
--- a/internal/testhelpers/e2e/session.go
+++ b/internal/testhelpers/e2e/session.go
@@ -193,6 +193,7 @@ func (s *Session) SpawnCmdWithOpts(exe string, opts ...SpawnOptions) *termtest.C
 			RetainWorkDir:  true,
 			ObserveExpect:  observeExpectFn(s),
 			ObserveSend:    observeSendFn(s),
+			ExtraOpts:      []expect.ConsoleOpt{expect.WithStdout(os.Stdout)},
 		},
 		NonWriteableBinDir: false,
 	}
--- a/internal/testhelpers/e2e/session.go
+++ b/internal/testhelpers/e2e/session.go
@@ -193,6 +193,7 @@ func (s *Session) SpawnCmdWithOpts(exe string, opts ...SpawnOptions) *termtest.C
 			RetainWorkDir:  true,
 			ObserveExpect:  observeExpectFn(s),
 			ObserveSend:    observeSendFn(s),
+			// uncomment the next line to print all output to stdout eg., if tests fail with a hard timeout causing a panic
 			// ExtraOpts:      []expect.ConsoleOpt{expect.WithStdout(os.Stdout)},
 		},
 		NonWriteableBinDir: false,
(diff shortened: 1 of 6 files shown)


internal/testhelpers/e2e/session.go
```diff
@@ -193,6 +193,7 @@ func (s *Session) SpawnCmdWithOpts(exe string, opts ...SpawnOptions) *termtest.C
 			RetainWorkDir:  true,
 			ObserveExpect:  observeExpectFn(s),
 			ObserveSend:    observeSendFn(s),
+			// ExtraOpts:      []expect.ConsoleOpt{expect.WithStdout(os.Stdout)},
```
MDrakos, line 196: Let's leave a comment here to explain why this code is commented out and why you may want to enable it

pkg/platform/model/inventory.go
```diff
@@ -50,25 +52,59 @@ func IngredientByNameAndVersion(language, name, version string, prefix Namespace
...
+						})
 				}
 			}
 		}
 	}
 
-	return nil, locale.NewInputError("inventory_ingredient_version_not_available", "Version {{.V0}} is not available for package {{.V1}} on the ActiveState Platform", version, name)
+	bestMatch, err := FilterForBestIngredientMatch(candidates, name)
+	if err != nil {
+		return nil, locale.NewInputError("inventory_ingredient_version_not_available", "Version {{.V0}} is not available for package {{.V1}} on the ActiveState Platform", version, name)
```
MDrakos, line 78: The error from `FilterForBestIngredientMatch` is being thrown away here. Shouldn't we just return the error, or wrap again?

pkg/platform/model/inventory.go
```diff
@@ -50,25 +52,59 @@ func IngredientByNameAndVersion(language, name, version string, prefix Namespace
...
+		for _, feature := range c.LatestVersion.ProvidedFeatures {
+			if *feature.Feature == name {
+				return c, nil
+			}
+		}
+	}
+
+	if len(candidates) == 0 {
+		return nil, locale.NewInputError("inventory_ingredient_no_version_available", "No versions are available for package {{.V0}} on the ActiveState Platform", name)
+	}
```
MDrakos, line 97: nit: I think this check makes more sense at the beginning of this function

pkg/platform/model/inventory.go
```diff
@@ -82,21 +118,33 @@ func IngredientWithLatestVersion(language, name string, prefix NamespacePrefix)
...
+		return nil, errs.Wrap(err, "Could not retrieve a match for ingredient %s.", name)
+	}
 
-	return nil, locale.NewInputError("inventory_ingredient_no_version_available", "No versions are available for package {{.V0}} on the ActiveState Platform", name)
+	latest := &IngredientAndVersion{
+		bestMatch.V1SearchIngredientsResponseIngredientsItems,
+		*bestMatch.LatestVersion.Version,
+		bestMatch.Namespace,
+	}
+	return latest, nil
```
MDrakos, line 147: Can't we just return the `bestMatch` at this point?

MDrakos (changes requested):
Looks good! Just a few minor questions/concerns

pkg/platform/model/inventory.go
mdrohmann: thanks for catching that.

pkg/platform/model/inventory.go
mdrohmann: yes, indeed!

MDrakos (approved):

PR #1126, opened 2020-11-19
state push can push to existing projects

mdrohmann:
https://www.pivotaltracker.com/story/show/175651094

### DIFF
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -1,11 +1,14 @@
 package push
 
 import (
+	"path/filepath"
+
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/projectfile"
 
 	"github.com/spf13/viper"
 
@@ -17,12 +20,17 @@ import (
 
 type configGetter interface {
 	GetString(key string) string
+	GetStringMapStringSlice(key string) map[string][]string
 }
 
 type Push struct {
 	config configGetter
 	output.Outputer
 	project *project.Project
+}
+
+type PushParams struct {
+	Namespace *project.Namespaced
 }
 
 type primeable interface {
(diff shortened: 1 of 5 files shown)


pkg/projectfile/projectfile.go
```diff
@@ -1185,6 +1185,29 @@ func (p *Project) Persist() {
...
+	if projects == nil {
+		projects = make(map[string][]string)
+	}
+
+	for name, paths := range projects {
+		if name == "/" {
+			continue
+		}
+		for _, path := range paths {
+			if isEqual, _ := fileutils.PathsEqual(projectPath, path); isEqual {
```
MDrakos, line 1203: We should at least be logging the error here, if not returning it.

internal/runners/push/push.go
```diff
@@ -34,52 +42,89 @@ func NewPush(config *viper.Viper, prime primeable) *Push {
...
+	owner := r.project.Owner()
+	name := r.project.Name()
+	if r.project.IsHeadless() {
+		namespace := params.Namespace
+		if !namespace.IsValid() {
+			names := projectfile.GetProjectNameForPath(r.config, filepath.Dir(r.project.Source().Path()))
+			if names == "" {
+				return errs.AddTips(
+					locale.NewInputError("push_needs_namespace", "Could not find out what project to push to."),
+					locale.Tl("push_add_namespace_tip", "You can specify a project name as an argument for [NOTICE]state push[/RESET]."),
```
MDrakos, line 67: We may want to use [ACTIONABLE] here. Something like `You can specify a project by running `state push <project name>`. Just a suggestion

internal/runners/push/push.go
```diff
@@ -34,52 +42,89 @@ func NewPush(config *viper.Viper, prime primeable) *Push {
...
+			return locale.NewInputError("push_invalid_arg_namespace", "")
+		}
+	}
+
+	// Get the project remotely if it already exists
+	pjm, fail := model.FetchProjectByName(owner, name)
 	if fail != nil {
-		if !fail.Type.Matches(model.FailProjectNotFound) {
+		if fail.Type.Matches(model.FailProjectNotFound) {
+			if r.project.IsHeadless() {
```
MDrakos, line 88: We should be able to combine these ifs and remove the nesting

MDrakos (changes requested):
Looks good. Though your comment in stand up was correct, there are a lot of branches in push now. It might make sense to go over and try to move some logic to helper functions. Maybe a set a functions to prepare for push if in a headless state, etc... I'll leave it up to you to see if that makes sense.

The only other thing I could think of adding is an integration test if that is feasible.

pkg/projectfile/projectfile.go
mdrohmann: I logged it.

mdrohmann:
I decided against splitting up the runners function, as last time I started splitting it up by use-case, Nathan asked me to unify the control flow.  He might think differently now, that the complexity has grown, but I think I'll ask him when he is back...

internal/runners/push/push.go
```diff
@@ -34,23 +42,51 @@ func NewPush(config *viper.Viper, prime primeable) *Push {
...
+			var fail *failures.Failure
+			namespace, fail = project.ParseNamespace(names)
+			if fail != nil {
+				return errs.Wrap(fail.ToError(), "Could not parse namespace %s to push headless commit to", name)
+			}
+		}
+		owner = namespace.Owner
+		name = namespace.Project
+	} else {
+		if params.Namespace.IsValid() {
```
MDrakos, line 79:
```suggestion
		if !params.Namespace.IsValid() {
```

internal/runners/push/push.go
```diff
@@ -59,35 +95,43 @@ func (r *Push) Run() error {
...
+		}
+		if branch.CommitID != nil && branch.CommitID.String() == r.project.CommitID() {
+			r.Outputer.Notice(locale.T("push_up_to_date"))
+			return nil
+		}
+	} else {
+		r.Outputer.Notice(locale.Tl("push_creating_project", "Creating project [NOTICE]{{.V1}}[/RESET] under [NOTICE]{{.V0}}[/RESET] on the ActiveState Platform", owner, name))
+		pjm, fail = model.CreateEmptyProject(owner, name, r.project.Private())
+		if fail != nil {
+			return locale.WrapError(fail.ToError(), "push_project_create_empty_err", "Failed to create an project {{.V0}}.", r.project.Namespace().String())
```
MDrakos, line 121:
```suggestion
			return locale.WrapError(fail.ToError(), "push_project_create_empty_err", "Failed to create a project {{.V0}}.", r.project.Namespace().String())
```

MDrakos (changes requested):

internal/runners/push/push.go
mdrohmann: Well, here we really have to check for `Namespace.IsValid()` to fulfill this AC: 

> If you provide a namespace while not in a headless state we should error out

But I noticed that I didn't specify the error message yet.  Ooops! Will fix.

MDrakos (approved):

PR #1128, opened 2020-11-20
Ensure verbose flag is working

daved:
https://www.pivotaltracker.com/story/show/175586318

 https://www.pivotaltracker.com/story/show/175586318

### DIFF
--- a/internal/logging/defaults.go
+++ b/internal/logging/defaults.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/rollbar/rollbar-go"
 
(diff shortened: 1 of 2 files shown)


internal/logging/defaults.go
```diff
@@ -20,18 +21,35 @@ import (
 // Logger describes a logging function, like Debug, Error, Warning, etc.
 type Logger func(msg string, args ...interface{})
 
+type safeBool struct {
+	mu sync.Mutex
```
mdrohmann, line 25: I would argue against using a mutex in the logger, as that is quite expensive to call.  My suggestions would be to 
- either use a regular boolean variable, as it would not be the end of the world if the verbose flag was incorrect in (and that is also a very unlikely situation) OR
- use an atomic bool like here: https://gist.github.com/mahan/6256149  That is much more efficient than a mutex...  It just ensures that we are not reading the value from a cached CPU register with an old value

mdrohmann (changes requested):

internal/logging/defaults.go
daved: I opted not to use an atomic because it's more difficult to read the implementation and the performance improvement is measured in billionths of a second. Personally, I'd prefer the atomic op anyway, but felt strongly that Nathan wouldn't. Should I still proceed with the atomic?

daved: Regarding not using a guard: I strongly think we should not accept races as "benign". The concern is not simply incorrect data, it's torn data which is "undefined" behavior and can pop the application.

mdrohmann: Let's just leave the mutex. I always strive to reduce locks if it is easy, but it is not the end of the world - especially in our use case.

mdrohmann (approved):

PR #1129, opened 2020-11-20
Powershell intstaller encoding

MDrakos:
I've updated the encoding to `UTF-8 with BOM`. 

The file was originally encoded as `POSIX shell script, ASCII text executable, with CRLF line terminators` and then we switched it to `UTF-16` which, as we know, does not play well with Git.

According to Microsoft the default encoding for Powershell 6+ is just `UTF-8` however with that encoding I was not able to display the text boxes. This is also the case with the default `Windows 1252` encoding. Changing to `UTF-8 with BOM` fixed this.

I think the above is related to the last paragraph here: https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8

>Microsoft compilers[9] and interpreters, and many pieces of software on Microsoft Windows such as Notepad treat the BOM as a required magic number rather than use heuristics. These tools add a BOM when saving text as UTF-8, and cannot interpret UTF-8 unless the BOM is present or the file contains only ASCII. Windows PowerShell (up to 5.1) will add a BOM when it saves UTF-8 XML documents. However, PowerShell Core 6 has added a -Encoding switch on some cmdlets called utf8NoBOM so that document can be saved without BOM. Google Docs also adds a BOM when converting a document to a plain text file for download.

I've tested the installer with this encoding on both a Windows 10 and Windows 7. On both there are no issues with displaying the characters required for the text box.

 https://www.pivotaltracker.com/story/show/175424110


mdrohmann (approved):
:+1:

PR #1130, opened 2020-11-20
explain and fix admin default installations

mdrohmann:
https://www.pivotaltracker.com/story/show/175672636

I created a follow-up story to discuss if this is really the behavior that we want for administrator default installations: https://www.pivotaltracker.com/story/show/175826240

### DIFF
--- a/assets/.gitattributes
+++ b/assets/.gitattributes
@@ -1,0 +1,4 @@
+* text=auto
+
+*.bat text eol=crlf
+*.ps1 text eol=crlf
--- a/internal/globaldefault/default.go
+++ b/internal/globaldefault/default.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/gobuffalo/packr"
+	"github.com/thoas/go-funk"
 
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
@@ -55,10 +56,11 @@ func isBinDirOnWindowsUserPath(binDir string) bool {
 		return false
 	}
 
-	for _, p := range strings.Split(path, string(os.PathListSeparator)) {
-		if p == binDir {
-			return true
-		}
+	if funk.ContainsString(
+		strings.Split(path, string(os.PathListSeparator)),
+		binDir,
+	) {
+		return true
 	}
 
 	return false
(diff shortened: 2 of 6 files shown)


internal/globaldefault/default.go
```diff
@@ -34,6 +38,34 @@ func BinDir() string {
...
+	if fail != nil {
+		logging.Error("Failed to get user PATH")
+		return false
+	}
+
+	for _, p := range strings.Split(path, string(os.PathListSeparator)) {
+		if p == binDir {
+			return true
+		}
+	}
```
MDrakos, line 64: nit: I think we could use `funk.Contains` here?

internal/globaldefault/default.go
```diff
@@ -61,6 +98,31 @@ func Prepare(subshell subshell.SubShell) error {
...
+			logging.Error("Failed to determine current user.")
+			return
+		}
+		out.Notice(locale.Tl(
+			"default_admin_activation_warning",
+			"[NOTICE]The default activation is added to the environment of user {{.V0}}.  The project may be inaccessible when run with Administrator privileges or authenticated as a different user.[/RESET]",
+			u.Username,
+		))
+	}
+}
```
MDrakos, line 124: While this needs issue definitely needs to be discussed and addressed I feel this code is well outside the scope of this story. We are introducing new messaging to the user that should be vetted by both Nathan and Pete before it goes in. It may be best to add this to the ticket description [here](https://www.pivotaltracker.com/n/projects/2203557/stories/175826240) as a possible solution or stop-gap until we know how we want to handle it.

If you like to get this in on this PR we can involve Pete and see what he thinks.

MDrakos (changes requested):

internal/globaldefault/default.go
mdrohmann: Good point.  I'll remove it, and bring it up during stand-up anyways.

assets/.gitattributes
```diff
@@ -0,0 +1,4 @@
+* text=auto
```
mdrohmann, line 1: I added this file, because the bat files in our assets had just LF endings, and when I was running `state activate ActiveState/Perl-5.32 --default` as an Administrator the activation script was not interpreted correctly.  But *only* as an Administrator user.  It worked otherwise.  As debugging the issues also meant fixing it, I decided to put it into this PR. Let me know if you want me to put it in a separate bug instead...

MDrakos (approved):

PR #1132, opened 2020-11-23
Fix pkg count in build streamer msg handling

daved:
https://www.pivotaltracker.com/story/show/175619749

 https://www.pivotaltracker.com/story/show/175619749


mdrohmann (approved):
LGTM

PR #1133, opened 2020-11-23
Update commit printing

MDrakos:
https://www.pivotaltracker.com/story/show/175487029

### DIFF
--- a/internal/table/table.go
+++ b/internal/table/table.go
@@ -170,16 +170,14 @@ func renderRow(providedColumns []string, colWidths []int) string {
 				end = maxLen
 			}
 
+			skipLineBreak := 0
 			if breakpos := runeSliceIndexOf(colValue, linebreakRune); breakpos != -1 && breakpos < end {
 				end = breakpos + 1
+				skipLineBreak = 1
 			}
 
 			suffix := strings.Repeat(" ", maxLen-end)
-			if len(colValue) > end && colValue[end-1] == linebreakRune {
-				result += pad(string(colValue[0:end-1]) + suffix)
-			} else {
-				result += pad(string(colValue[0:end]) + suffix)
-			}
+			result += pad(string(colValue[0:end-skipLineBreak]) + suffix)
 
 			columns[n] = string(colValue[end:])
 		}
--- a/internal/table/table.go
+++ b/internal/table/table.go
@@ -175,7 +175,12 @@ func renderRow(providedColumns []string, colWidths []int) string {
 			}
 
 			suffix := strings.Repeat(" ", maxLen-end)
-			result += pad(string(colValue[0:end]) + suffix)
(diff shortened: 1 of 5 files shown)


internal/table/table.go
```diff
@@ -175,7 +175,12 @@ func renderRow(providedColumns []string, colWidths []int) string {
 			}
 
 			suffix := strings.Repeat(" ", maxLen-end)
-			result += pad(string(colValue[0:end]) + suffix)
+			if len(colValue) > end && colValue[end-1] == linebreakRune {
+				result += pad(string(colValue[0:end-1]) + suffix)
+			} else {
+				result += pad(string(colValue[0:end]) + suffix)
+			}
+
```
MDrakos, line 183: I'm unsure if this is the correct way to handle this, though it did fix my problem. Please let me know if there is a better approach

pkg/cmdlets/commit/commit.go
```diff
@@ -13,26 +13,62 @@ import (
...
+	}, nil
+}
+
+func shortHash(commitID string) string {
+	split := strings.Split(commitID, "-")
+	if len(split) == 0 {
+		return ""
+	}
+	return split[0]
+}
```
MDrakos, line 69: I had to use a short hash as the formatting can be a little odd when we use the full hash. I will file a bug with details of the output with long values in the commit and message columns

MDrakos: Story here: https://www.pivotaltracker.com/story/show/175863427

internal/table/table.go
mdrohmann: Maybe we could do something like this:

```
			skipLineBreak := 0
			if breakpos := runeSliceIndexOf(colValue, linebreakRune); breakpos != -1 && breakpos < end {
				end = breakpos + 1
				skipLineBreak = 1
			}
 			suffix := strings.Repeat(" ", maxLen-end)
 			result += pad(string(colValue[0:end-skipLineBreak]) + suffix) 

```

mdrohmann: That would get rid of the second if-clause

internal/table/table_test.go
```diff
@@ -209,7 +209,7 @@ func Test_renderRow(t *testing.T) {
 				colWidths:       []int{8, 8},
 			},
 			"  abcd    abcd  \n" +
-				"          e\n    \n" +
```
mdrohmann, line 212: Well, it looks like this unit-test was broken all along.  Sorry about that.

mdrohmann (changes requested):

internal/table/table_test.go
MDrakos: No worries, and thank you for confirming that this was actually broken. I thought the two line breaks seemed a little suspect

mdrohmann (approved):

PR #1137, opened 2020-11-24
Check execution policy

MDrakos:
https://www.pivotaltracker.com/story/show/175878953

### DIFF
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -269,6 +269,18 @@ function test-64bitPtr() {
     return [IntPtr]::size -eq 8
 }
 
+function checkExecutionPolicy() {
+    $executionPolicy = Get-ExecutionPolicy -Scope Process
+    Write-Host "Execution Policy: $executionPolicy"
+    if ($executionPolicy -eq "Restricted" -Or $executionPolicy -eq "Undefined") {
+        Write-Warning "Execution Policy is set to $executionPolicy."
+        Write-Warning "In order to run scripts please update the Execution Policy."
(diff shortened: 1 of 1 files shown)


mdrohmann (commented):
Does this actually address the problem that Francesca had?  The error message appeared during parsing the profile.ps1 file, which is not necessary for our script, is it?

I would assume that if there was an actual problem executing the `install.ps1` script itself, we would not even be able to get to the point where we could check it.  But I may be wrong, in which case we should leave it in.

But to address the story's bug:  How about just adding the `-NoProfile` option to the one-liner?

MDrakos:
That's a good point. I have since been able to reproduce this. It appears that our one-liner bypasses the execution policy. This may be because we are passing script contents to Powershell which is noted [here](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7.1) as a way to circumvent the execution policy. 

When the execution policy is restricted and the user has a `profile.ps1` they will get the error message Francesca encountered every time they open a Powershell prompt. This is why we see this error when we invoke Powershell from cmd. Since this is an issue with the users' setup and it doesn't affect the script I think we should not make any changes to address this. Adding the `-NoProfile` flag could also have unforeseen consequences for users with unique setups that rely on the profile.

rawktron:
@MDrakos I agree -- this is a configuration issue on the user's machine and not an issue in the tool. If possible we should tell users that they need to do this -- but at the same time, the OS is clearly already doing this.

PR #1141, opened 2020-11-25
State projects shows no projects

MDrakos:
This is the sequence of events that can lead to `state projects` showing no projects, even though there is a default activation:

1. `state activate --default` a project (This was done using the one-liner)
2. Install a package
3. Uninstall the state tool (or just remove the config directory)
3. `state activate --default` again with the same project in the same directory

Since we added a package in step 2 the project is now in a headless state. When we add the project mapping to the config we get:

```
projects:
  /:
  - C:\Users\Administrator\ActiveState\Perl-5.32
```

This fix writes the mapping during activation using the namespace parameter. There is also a minor update to the project regex. I've also filed a story [here](https://www.pivotaltracker.com/story/show/175901661) to update our handling of headless projects in the config and with `state projects`

 https://www.pivotaltracker.com/story/show/175880344

### DIFF
--- a/internal/runners/activate/activate.go
+++ b/internal/runners/activate/activate.go
@@ -175,6 +175,8 @@ func (r *Activate) run(params *ActivateParams) error {
 			return locale.WrapError(err, "err_activate_default", "Could not configure your project as the default.")
 		}
 
+		projectfile.StoreProjectMapping(params.Namespace.String(), filepath.Dir(proj.Source().Path()))
+
 		r.out.Notice(output.Heading(locale.Tl("global_default_heading", "Global Default")))
 		r.out.Notice(locale.Tl("global_default_set", "Successfully configured [NOTICE]{{.V0}}[/RESET] as the global default project.", proj.Namespace().String()))
 
--- a/pkg/projectfile/projectfile.go
+++ b/pkg/projectfile/projectfile.go
@@ -72,7 +72,7 @@ var (
 )
 
 var (
-	urlProjectRegexStr = fmt.Sprintf(`https:\/\/[\w\.]+\/([\w_.-]*)\/([\w_.-]*)(?:\?commitID=)*(.*)`)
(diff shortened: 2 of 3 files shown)


mdrohmann (changes requested):
So in that case - even with the changes from this PR - the `state projects` command would still not show any projects, because the directory is head-less, no?

Does storing the project configuration during `state activate --default` actually change anything then?  I would think that it should be early enough, as we did it before which was happening whe reading the `activestate.yaml` IIUC.

MDrakos:
In my testing this does fix the issue. We end up with two entries in the project mapping. The incorrect entry will be written every time the project file is parsed/saved and the correct one is written when a default project is activated. Dealing with the incorrect entry is for the other story I filed. This PR is admittedly more of a quick fix until we know how we want to represent headless projects in the config file.

```
projects:
 /:
  - C:\Users\Administrator\ActiveState\Perl-5.32
 activestate/perl-5.32:
  - C:\Users\Administrator\ActiveState\Perl-5.32
```

mdrohmann (changes requested):
Interesting.  But is then maybe sufficient to add the line that writes the configuration line early ie., this one: https://github.com/ActiveState/cli/pull/1141/files#diff-0927148580d0b8620ad94bebab7330dc24b7b2d940c17f245f247fb7f51dda85R1238-R1241

If I understand correctly that would be run during an activation with the correct namespace even without the addition of the line in `activate.go`, no?

MDrakos:
Unfortunately no, it will run during activation but in the scenario outlined above the project is already in a headless state so the key written to the config will still be `/`. 

Thinking about this further I think the correct solution is to have a proper way to represent headless projects in the config and therefore with `state projects`. Maybe we should leave this story in favour of the other one I filed. We just need to settle on a representation.

mdrohmann:
I think that's what we should do. Let's discuss that in standup today.

MDrakos:
Closing this as per discussion from standup today. This is a known limitation and a very unique set of circumstances to reproduce.

PR #1143, opened 2020-11-27
Get auth status from scripts

MDrakos:
https://www.pivotaltracker.com/story/show/175861261


Naatan (changes requested):
I'm sorry but those AC's were way too loosely defined and I was not involved with the planning of the story. I'd like to better understand the use-case and then define those ACs.

cmd/state/main.go
```diff
@@ -146,11 +146,13 @@ func run(args []string, out output.Outputer) (int, error) {
 	}
 	// Set up conditional, which accesses a lot of primer data
 	sshell := subshell.New()
-	conditional := constraints.NewPrimeConditional(pjOwner, pjName, pjNamespace, sshell.Shell())
+	auth := authentication.Get()
+	conditional := constraints.NewPrimeConditional(auth, pjOwner, pjName, pjNamespace, sshell.Shell())
 	project.RegisterConditional(conditional)
+	project.RegisterExpander("mixin", project.NewMixin(auth).Expander)
```
Naatan, line 152: Since this is all contained within the project package there's no reason to register this externally. Just hook it up via http://github.com/ActiveState/cli/blob/8182de4063d5524abb67c1fefb10da02178d8050/pkg/project/registry.go#L20-L20

internal/constraints/constraints.go
```diff
@@ -27,9 +28,22 @@ type Conditional struct {
...
 	c := &Conditional{map[string]interface{}{}, map[string]interface{}{}}
 
+	c.RegisterFunc("Mixin", func() map[string]interface{} {
+		res := map[string]string{
+			"name":  "",
+			"email": "",
+		}
+		if a.Authenticated() {
+			res["name"] = a.WhoAmI()
+			res["email"] = a.Email()
```
Naatan, line 41: The convention for constraints seems to be to uppercase these

pkg/project/expander.go
```diff
@@ -193,6 +194,35 @@ func expandPath(name string, script *Script) (string, error) {
...
+func NewMixin(auth *authentication.Auth) *Mixin {
+	return &Mixin{auth}
+}
+
+// Expander expands mixin variables
+func (m *Mixin) Expander(_ string, name string, meta string, _ bool, _ *Project) (string, error) {
+	if name == "user" {
+		return userExpander(m.auth, meta), nil
+	}
+	return "", nil
```
Naatan, line 223: This should probably be an error

Naatan (changes requested):

cmd/state/main.go
mdrohmann: In that case I should probably not initialize with `auth`, as this would be done in the package's `init()` function.  That is the reason I put it here - to avoid using the `authentication.Get()` call...

pkg/project/expander.go
mdrohmann: The other expander functions follow this convention.  If we cannot resolve the name, we assume that it is not our variable.

cmd/state/main.go
Naatan: Ok, let's leave it here for now. We can revisit once we refactor the project package.

Naatan (approved):

PR #1144, opened 2020-11-30
Populate anonID field on headless commits

MDrakos:
https://www.pivotaltracker.com/story/show/175861123

### DIFF
--- a/internal/runners/history/history.go
+++ b/internal/runners/history/history.go
@@ -93,7 +93,9 @@ func printCommits(out output.Outputer, commits []*mono_models.Commit, orgs []gmo
 func authorIDsForCommits(commits []*mono_models.Commit) []strfmt.UUID {
 	authorIDs := []strfmt.UUID{}
 	for _, commit := range commits {
-		authorIDs = append(authorIDs, *commit.Author)
+		if commit.Author != nil {
+			authorIDs = append(authorIDs, *commit.Author)
+		}
 	}
 	return authorIDs
 }
--- a/internal/runners/history/history.go
+++ b/internal/runners/history/history.go
@@ -93,7 +93,7 @@ func printCommits(out output.Outputer, commits []*mono_models.Commit, orgs []gmo
 func authorIDsForCommits(commits []*mono_models.Commit) []strfmt.UUID {
 	authorIDs := []strfmt.UUID{}
 	for _, commit := range commits {
-		authorIDs = append(authorIDs, commit.Author)
+		authorIDs = append(authorIDs, *commit.Author)
 	}
 	return authorIDs
 }
--- a/internal/runners/revert/revert.go
+++ b/internal/runners/revert/revert.go
@@ -58,6 +58,9 @@ func (r *Revert) Run(params *Params) error {
 	if err != nil {
 		return locale.WrapError(err, "err_revert_get_commit", "Could not fetch commit details for commit with ID: {{.V0}}", params.CommitID)
 	}
+	if revertCommit.Author == nil {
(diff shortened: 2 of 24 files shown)


internal/runners/history/history.go
```diff
@@ -93,7 +93,7 @@ func printCommits(out output.Outputer, commits []*mono_models.Commit, orgs []gmo
 func authorIDsForCommits(commits []*mono_models.Commit) []strfmt.UUID {
 	authorIDs := []strfmt.UUID{}
 	for _, commit := range commits {
-		authorIDs = append(authorIDs, commit.Author)
+		authorIDs = append(authorIDs, *commit.Author)
```
Naatan, line 96: Nil check

pkg/platform/model/vcs.go
```diff
@@ -235,13 +235,20 @@ func CommitsBehindLatest(ownerName, projectName, commitID string) (int, *failure
...
 		Message:        commitMessage,
 		ParentCommitID: parentCommitID,
-	})
+	}
+
+	if isHeadless {
+		commit.AnonID = logging.UniqID()
+	}
+
+	params.SetCommit(commit)
```
Naatan, line 251: This is too much specialized knowledge that we shouldn't be dealing with at this level. Instead of passing `isHeadless` just pass the anonymous ID.

Naatan (changes requested):
Lots of missing nil checks.

Let's also move the source of truth for the UniqID to the auth package. Logging was historically the only package that used it, but that doesn't mean it's the most logical one to own it now that we start using it elsewhere.

pkg/platform/api/graphql/model/projectconv.go
```diff
@@ -41,6 +41,10 @@ func (p *Project) ToMonoProject() (*mono_models.Project, *failures.Failure) {
 		}
 	}
 
+	if p.RepoURL == nil {
+		return nil, failures.FailMarshal.Wrap(ErrMissingRepoURL)
+	}
```
MDrakos, line 46: I tried to update this method to return an error but it quickly snowballed so opted to add a failure here instead.

MDrakos:
I opted to create a new package for this as moving it to the auth package created an import cycle.

internal/runners/revert/revert.go
```diff
@@ -58,8 +58,11 @@ func (r *Revert) Run(params *Params) error {
 	if err != nil {
 		return locale.WrapError(err, "err_revert_get_commit", "Could not fetch commit details for commit with ID: {{.V0}}", params.CommitID)
 	}
+	if revertCommit.Author == nil {
+		return locale.NewError("err_nil_author", "Revert commit has no author")
```
Naatan, line 62: Why does this matter for this use-case?

pkg/cmdlets/commit/commit.go
```diff
@@ -46,7 +46,11 @@ func PrintCommits(out output.Outputer, commits []*mono_models.Commit, orgs []gmo
 }
 
 func commitDataFromCommit(commit *mono_models.Commit, orgs []gmodel.Organization) (commitData, error) {
-	username, err := usernameForID(commit.Author, orgs)
+	if commit.Author == nil {
+		return commitData{}, locale.NewError("err_commit_no_author", "Commit has no author")
+	}
+
+	username, err := usernameForID(*commit.Author, orgs)
```
Naatan, line 53: We should just use an "anonymous" username rather than fail outright. We support anonymous commits, so this code shouldn't be failing if you're anonymous.

Naatan (changes requested):

Naatan (approved):

PR #1145, opened 2020-11-30
termtest CI does not recognize some prompt inputs

mdrohmann:
https://www.pivotaltracker.com/story/show/175706773

Apparently, pressing the Enter key even on Windows only sends the newline character `\n` to the terminal, not `\r\n`.  By doing the latter, the second byte can be instead sent buffered and be send to the following prompt - leading to the unexpected behavior that we had seen.


Naatan (changes requested):
Can we make this the default for `.Send()` and introduce a `.SendUnterminated()` (or whatever you want to name that).

Feel like this is too error prone.

mdrohmann:
I think you are right, but I actually changed the `conproc.SendLine` function so that it always just sends a simple newline character.  There is now also a function to add an OS aware newline separator `SendOSLine`.

mdrohmann:
The integration tests are not running right now, because Github Actions are degraded.  But I tested it locally, and believe that they will eventually pass this way...

Naatan (approved):

PR #1148, opened 2020-12-02
Clean up table math to improve maintainability

Naatan:


internal/table/table.go
```diff
@@ -55,86 +55,81 @@ func (t *Table) Render() string {
...
-				minWidth = spanWidth
+			columnSize := utf8.RuneCountInString(columnValue)
+
+			// Detect spanned column info
+			rowHasSpannedColumn := len(row.columns) < len(t.headers)
+			spannedColumnIndex := len(row.columns) - 1
+
+			if rowHasSpannedColumn && n == spannedColumnIndex {
+				// Record total row size as minTableWidth
+				minTableWidth = mathutils.MaxInt(minTableWidth, colWidths[n]+columnSize+(padding*2))
```
mdrohmann, line 81:
```suggestion
				minTableWidth = mathutils.MaxInt(minTableWidth, colWidthsCombined+columnSize+(padding*2))
```

internal/table/table.go
```diff
@@ -55,86 +55,81 @@ func (t *Table) Render() string {
...
+		colWidths[n] += int((averageWidth - colWidth) * multiplier)
 	}
 
-	colWidths = rescaleColumns(colWidths, columnTotal, targetTotal)
-	logging.Debug("Table column widths: %v, total: %d", colWidths, targetTotal)
+	// Account for floats that got rounded
+	colWidths[len(colWidths)-1] += int(total) - mathutils.Total(colWidths...)
+}
+
+func rescaleColumns(colWidths []int, targetTotal int) {
```
mdrohmann, line 123: This is okay, and certainly simpler to read.  I rewrote in a way such that never more than one extra column would be added to the last column.  If we have six columns and the first five columns were all truncated down by .8, the last column what get an extra four characters (4 = 5 * 0.8).  That seems excessive. :) But of course, it is also a total edge case, and we will rarely have to work with small column widths where this makes a difference.

internal/table/table_test.go
```diff
@@ -96,7 +98,7 @@ func TestTable_colWidths(t1 *testing.T) {
 				},
 				100,
 			},
-			[]int{23, 23, 27, 27},
+			[]int{22, 22, 27, 29},
```
mdrohmann, line 101: This is also a good example for how the previous `rescaleColumns` function adds less jitter.  Before your changes, the last two columns were equal length (as the had the same contents), now the last column is a little wider.  But that's probably just something that I am concerned about. :)

mdrohmann (changes requested):
You can ignore my two comments about the `rescaleColumns()` function if you like.

internal/table/table.go
Naatan: This would include the columns this column is spanning. But good point, it should include all preceding columns. Will fix, thanks!

internal/table/table.go
Naatan: Yeah I'd favour readability over accounting for a few units that don't make much of a difference in the grand scheme of things.

internal/table/table_test.go
Naatan: Yeah totally get it :) But I think this only registers with small numbers like this. In practice this won't really affect the UX and keeps the code readability. I'm trying to weigh all sides, and since readability is already problematic here the math can suffer the compromise here.

mdrohmann (approved):

PR #1149, opened 2020-12-03
Use ResolveRecipes instead of SolveOrder

Naatan:
https://www.pivotaltracker.com/story/show/176002052

 https://www.pivotaltracker.com/story/show/176002052


mdrohmann (approved):
:heavy_check_mark:

PR #1151, opened 2020-12-04
Add matching features list to pkgs search output

daved:
https://www.pivotaltracker.com/story/show/175750825

 https://www.pivotaltracker.com/story/show/175750825


daved:
The API, currently, returns all features provided for a found ingredient. A followup story has been created to ensure only matching modules are displayed: https://www.pivotaltracker.com/story/show/176004973

daved:
Also, a related story was filed to fix and/or allow control over column width: https://www.pivotaltracker.com/story/show/176005063

internal/output/plain.go
```diff
@@ -324,3 +330,30 @@ func trimValue(value string, size int) string {
...
+		}
+	}
+
+	return 0
+}
+
+func columns(offset int, value string) []string {
+	if offset < 0 {
+		offset *= -1
+		logging.Errorf("Negative shiftCols values are not handled, using postive value %d", offset)
```
Naatan, line 354: If we encounter this case we should just use 0, if the intend is to fail as gracefully as possible.

internal/output/plain.go
```diff
@@ -324,3 +330,30 @@ func trimValue(value string, size int) string {
 	}
 	return value
 }
+
+func shiftColsVal(opts []string) int {
+	for _, opt := range opts {
+		a := strings.TrimPrefix(opt, string(ShiftColsPrefix))
+		if len(a) < len(opt) {
```
Naatan, line 337: This conditional seems error prone and pointless. Granted it would look ridiculous if we passed an int that's longer in string size than the prefix, but that doesn't take away that this is unnecessary. Just drop the conditional here.

internal/output/plain.go
```diff
@@ -324,3 +330,30 @@ func trimValue(value string, size int) string {
...
+			n, err := strconv.Atoi(a)
+			if err != nil {
+				logging.Errorf("Cannot get shiftCols value: %v", err)
```
Naatan, line 340:
```suggestion
				logging.Errorf("Cannot get shiftCols value: %s, error: %v", a, err)
```

internal/runners/packages/search.go
```diff
@@ -86,6 +80,56 @@ func targetedLanguage(languageOpt string) (string, *failures.Failure) {
 	return model.LanguageForCommit(proj.CommitUUID())
 }
 
+type modules []string
+
+func makeModules(normalizedName string, pack *model.IngredientAndVersion) modules {
```
Naatan, line 85: Wouldn't `newModules` be the idiomatic naming?

internal/runners/packages/search.go
```diff
@@ -86,6 +80,56 @@ func targetedLanguage(languageOpt string) (string, *failures.Failure) {
...
+	var ms modules
+	for _, x := range pack.LatestVersion.ProvidedFeatures {
+		if x.Feature != nil && *x.Feature != normalizedName {
+			ms = append(ms, *x.Feature)
+		}
```
Naatan, line 90:
```suggestion
	for _, pf := range pack.LatestVersion.ProvidedFeatures {
		if pf.Feature != nil && *pf.Feature != normalizedName {
			ms = append(ms, *pf.Feature)
		}
```

I'm sorry this is nitpicky but this is a new level of variable naming that just triggers my OCD. Short vars are one thing. Short vars that have zero relation to their value in their naming is a whole other thing.

Naatan (changes requested):

internal/runners/packages/search.go
daved: Not in this case; The "new" and "make" prefix correlate to the "new" and "make" builtin functions. The make prefix is less common in general because return values tend to be pointers-to, but also because using the make prefix for non-pointer-to returns is not as widely known.

daved: Using the "make" prefix is practiced in the standard library, but not as consistently as "new": https://golang.org/src/cmd/internal/src/pos.go?s=1216:1306#L28

internal/output/plain.go
daved: This ensures that a potential value was found. The length of "a" would be the remaining string-integer except when no prefix is trimmed (i.e. no shiftCols option was set).

internal/runners/packages/search.go
daved: Not nitpicky! I will fix this - it's a scratch var name because I didn't know the name of the type/field that was needed when I first wrote the code. :facepalm:

internal/output/plain.go
daved: This was what was done initially, but the output was redundant with the strconv.Atoi error message. Is it OK to leave as-is?

internal/output/plain.go
daved: In order to improve readability, the var name has been updated and a comment was added to the conditional line.

internal/output/plain.go
Naatan: This is nitpicky at best, but I prefer my errors not to make any assumptions about what they are wrapping.

Naatan (approved):

PR #1152, opened 2020-12-04
Table word wrapping handles colour codes

MDrakos:
https://www.pivotaltracker.com/story/show/175727287

 https://www.pivotaltracker.com/story/show/176045571


internal/table/table.go
```diff
@@ -133,51 +140,120 @@ func rescaleColumns(colWidths []int, targetTotal int) {
 	colWidths[len(colWidths)-1] += targetTotal - mathutils.Total(colWidths...)
 }
 
+func getCroppedText(text string, maxLen int) []entry {
+	entries := make([]entry, 0)
+	pos := 0
+	matches := colorize.ColorRx.FindAllSubmatchIndex([]byte(text), -1)
+	runeText := []rune(text)
+
+	for pos < len(runeText) {
```
Naatan, line 149: I feel like we can do this in a single for loop, rather than the two. You're already asserting this conditional on line 159. And the loop on line 152 is basically just cutting things off at the line marker, which you could also do with a conditional.

I think it's probably most sensible to drop the for loop on 152 and just start a new entry when count==maxLen.

internal/table/table.go
```diff
@@ -1,9 +1,11 @@
 package table
```
Naatan, line 1: This should not live in the table package. It's use won't be exclusive to tables.

internal/table/table.go
```diff
@@ -133,51 +140,120 @@ func rescaleColumns(colWidths []int, targetTotal int) {
 	colWidths[len(colWidths)-1] += targetTotal - mathutils.Total(colWidths...)
 }
 
+func getCroppedText(text string, maxLen int) []entry {
```
Naatan, line 143: This is not accounting for any existing linebreaks within the text.

eg.

```
getCroppedText("foo\nbar", 4)
```

Should result in 

```
[]entry{{"foo", 3}, {"bar", 3}}
```

Whereas your current implementation would produce

```
[]entry{{"foo\n", 4}, {"bar", 3}}
```

internal/table/table.go
```diff
@@ -133,51 +140,120 @@ func rescaleColumns(colWidths []int, targetTotal int) {
...
+	entries := make([]entry, 0)
+	pos := 0
+	matches := colorize.ColorRx.FindAllSubmatchIndex([]byte(text), -1)
+	runeText := []rune(text)
+
+	for pos < len(runeText) {
+		buffer := bytes.Buffer{}
```
Naatan, line 150: Why make a bytes buffer if you're just going to send it nothing but strings, and then convert it to a string at the end?

internal/table/table.go
```diff
@@ -133,51 +140,120 @@ func rescaleColumns(colWidths []int, targetTotal int) {
...
+		if entry.length != 0 {
+			continue
+		}
+
+		text := entries[i].line
+		if i >= 1 {
+			entries[i-1].line = entries[i-1].line + text
+		}
+		entries = entries[0:i]
+	}
```
Naatan, line 189: This feels dirty. Why not just scan ahead during your loop above, see if you have a match right after where you're about to cut off the entry.

internal/table/table.go
```diff
@@ -133,51 +140,120 @@ func rescaleColumns(colWidths []int, targetTotal int) {
...
+
+func matchIndex(pos int, matches [][]int) (int, bool) {
+	for _, match := range matches {
+		start, stop := match[0], match[1]
+		if pos == start {
+			return stop, true
+		}
+	}
+	return -1, false
+}
```
Naatan, line 202: http://github.com/ActiveState/cli/blob/master/internal/sliceutils/sliceutils.go#L9-L9

internal/table/table.go
```diff
@@ -133,51 +140,120 @@ func rescaleColumns(colWidths []int, targetTotal int) {
...
 
-			skipLineBreak := 0
-			if breakpos := runeSliceIndexOf(colValue, linebreakRune); breakpos != -1 && breakpos < end {
-				end = breakpos + 1
-				skipLineBreak = 1
-			}
+	totalRows := 0
+	for _, columnEntries := range entries {
+		if len(columnEntries) > totalRows {
+			totalRows = len(columnEntries)
```
Naatan, line 237: This is setting the number of totalRows to the maximum number of columns found?

internal/table/table.go
```diff
@@ -133,51 +140,120 @@ func rescaleColumns(colWidths []int, targetTotal int) {
...
+	// Render each row
+	for i := 0; i < totalRows; i++ {
+		for n, columnEntry := range entries {
+			maxLen := widths[n]
+			text := ""
+			suffix := strings.Repeat(" ", maxLen)
+
+			if len(columnEntry) > i {
+				repeat := maxLen - columnEntry[i].length
+				suffix = strings.Repeat(" ", repeat)
```
Naatan, line 250: Should guard this against panics (if repeat is negative). Shouldn't happen, but I'd rather we have a slightly messed up looking table than a panic, in the event that there is a bug in this code.

Naatan (changes requested):

internal/table/table.go
MDrakos: I've moved it to the `colorize` package as it's dealing with the tags that that package defines. Let me know if there is a better spot

internal/table/table.go
MDrakos: Yes, not necessary

internal/table/table.go
MDrakos: This is not necessary now that I have implemented some the changes you requested above

internal/table/table.go
MDrakos: This function is doing a little more than the one you linked, specifically it's giving me the ending position of the tag if the index matches the start. I've removed this function and added the loop into `GetCroppedText`

internal/table/table.go
MDrakos: I originally had this, but removed it as I did not encounter it in any of the tests. I have added it back.

internal/table/table.go
MDrakos: This is setting the number of total rows to the maximum length of columns. Each `columnEntry` is a `[]entry` which represents all of the entries in that column, so the length is the number of rows in that column.

For the table:
```
one    two    three
1      2      3
```

The column entries would be
```
[[{one 3} {1 1}] [{two 3} {2 1}] [{three 5} {3 1}]]
```

Naatan: > This is setting the number of total rows to the maximum length of columns.

I think you might want to rename your variable or something? This is very confusing. Rows are not columns.

internal/colorize/cropped.go
```diff
@@ -0,0 +1,58 @@
...
+		// then we write the whole tag, otherwise write by rune
+		for _, match := range matches {
+			start, stop := match[0], match[1]
+			if pos == start {
+				entryText += string(runeText[pos:stop])
+				pos = stop
+			}
+		}
+
+		if pos > len(runeText)-1 {
```
Naatan, line 27: The for loop already handles this

internal/colorize/cropped.go
```diff
@@ -0,0 +1,58 @@
...
+			entries = append(entries, Entry{entryText, count})
+			break
+		}
+
+		if count == maxLen {
```
Naatan, line 32:
```suggestion
		// Reached end of line
		if count == maxLen {
```

internal/colorize/cropped.go
```diff
@@ -0,0 +1,58 @@
...
+			entries = append(entries, Entry{entryText, count})
+			count = 0
+			entryText = ""
+			continue
+		}
+
+		if runeText[pos] == '\n' {
```
Naatan, line 39:
```suggestion
		// Text already has line ending, so terminate the line here
		if runeText[pos] == '\n' {
```

internal/colorize/cropped.go
```diff
@@ -0,0 +1,58 @@
...
+type Entry struct {
+	Line   string
+	Length int
+}
+
+func GetCroppedText(text string, maxLen int) []Entry {
+	entries := make([]Entry, 0)
+	entryText := ""
+	pos := 0
+	count := 0
```
Naatan, line 12: The difference between these is not clear from their name. Please name them a little more explicitly.

internal/colorize/cropped.go
Naatan, line 32: You're not accounting for the color code on its own on a line anymore?

internal/table/table.go
```diff
@@ -228,13 +166,15 @@ func renderRow(providedColumns []string, colWidths []int) string {
...
 
 	totalRows := 0
-	for _, columnEntries := range entries {
-		if len(columnEntries) > totalRows {
-			totalRows = len(columnEntries)
+	for _, entry := range entries {
+		// Each entry represents a column. The longest column determines
+		// the number of rows the table will have.
+		if len(entry) > totalRows {
+			totalRows = len(entry)
```
Naatan, line 177: Can we rename "rows" to "lines". I think that's what I'm tripping over here. A row can have multiple lines.

internal/table/table.go
```diff
@@ -246,9 +186,12 @@ func renderRow(providedColumns []string, colWidths []int) string {
 			suffix := strings.Repeat(" ", maxLen)
 
 			if len(columnEntry) > i {
-				repeat := maxLen - columnEntry[i].length
+				repeat := maxLen - columnEntry[i].Length
+				if repeat < 0 {
+					repeat = padding
+				}
```
Naatan, line 192: To simplify you could use http://github.com/ActiveState/cli/blob/master/internal/mathutils/mathutils.go#L9-L9

Naatan (changes requested):

internal/table/table.go
MDrakos: Ahh okay yeah I can see where the confusion comes from

internal/colorize/cropped.go
MDrakos: This check is from the matching that happens above. If `stop` (the of the reset tag) is at the end of the string then the index is out of bounds

internal/table/table.go
MDrakos: Do you mean to use the max of `maxLen - columnEntry[i].Length` and `padding`? If so there are cases where the value we want will be the smaller of those two.

Naatan: I mean 

```
repeat := mathutils.MaxInt(0, maxLen - columnEntry[i].Length)
```

MDrakos: Ahh okay. Thanks. For some reason I wasn't seeing this.

internal/colorize/cropped.go
MDrakos: With the changes you suggested above we don't end up with entries that are just colour tags. Though this _could_ happen if there is a newline character right before or after a colour tag but in that case it would be respecting the existing line ending. Do we want to do anything special/differently there?

internal/colorize/cropped.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func GetCroppedText(text string, maxLen int) []Entry {
+	entries := make([]Entry, 0)
+	entryText := ""
+	currentPosition := 0
+	runesWritten := 0
+	matches := colorRx.FindAllStringSubmatchIndex(text, -1)
+	runeText := []rune(text)
+	last := len(runeText) - 1
```
Naatan, line 15:
```suggestion
func GetCroppedText(text []rune, lineSize int) []Entry {
        var string line // the line we're building, including the color codes
        var int plainTextPosition // plainText = minus color codes
        var int plainTextWritten
        
	entries := make([]Entry, 0)
	colorCodes := colorRx.FindAllStringSubmatchIndex(text, -1)
```

Trying to make your variables a bit more consistent and easier to digest.

This renames (make sure to update the code):

- maxLen > lineSize
- entryText > line
- currentPosition > plainTextPosition
- runesWritten > plainTextWritten
- matches > colorCodes
- last > dropped, just do `len(runeText) - 1` in your code

Open to other suggestions, but end goal is this code overall still needs to improve in readability. Variable names are perfect for telling the story of your code, so be mindful of how you name them.

internal/table/table.go
```diff
@@ -140,44 +146,59 @@ func renderRow(providedColumns []string, colWidths []int) string {
...
+		// Each entry represents a column. The longest column determines
+		// the number of rows the table will have.
+		if len(entry) > lines {
+			lines = len(entry)
+		}
+	}
 
-			suffix := strings.Repeat(" ", maxLen-end)
-			result += pad(string(colValue[0:end-skipLineBreak]) + suffix)
+	// Render each row
```
Naatan, line 181:
```suggestion
	// Render each line
```

internal/table/table.go
```diff
@@ -140,44 +146,59 @@ func renderRow(providedColumns []string, colWidths []int) string {
...
-				skipLineBreak = 1
-			}
+	lines := 0
+	for _, entry := range entries {
+		// Each entry represents a column. The longest column determines
+		// the number of rows the table will have.
+		if len(entry) > lines {
+			lines = len(entry)
+		}
+	}
```
Naatan, line 179: This logic still seems wrong, and a bit redundant. I think we should just iterate over `entries` until we reach an iteration where not a single column had data.

internal/table/table.go
```diff
@@ -140,44 +146,59 @@ func renderRow(providedColumns []string, colWidths []int) string {
...
+		for n, columnEntry := range entries {
+			maxLen := widths[n]
+			text := ""
+			suffix := strings.Repeat(" ", maxLen)
+
+			if len(columnEntry) > i {
+				end := len(columnEntry[i].Line)
+				if end > maxLen {
+					end = maxLen
+				}
```
Naatan, line 192: This should never happen, the cropped text function should have ensured that.

internal/table/table.go
```diff
@@ -140,44 +146,59 @@ func renderRow(providedColumns []string, colWidths []int) string {
...
-			result += pad(string(colValue[0:end-skipLineBreak]) + suffix)
+	// Render each row
+	for i := 0; i < lines; i++ {
+		for n, columnEntry := range entries {
+			maxLen := widths[n]
+			text := ""
+			suffix := strings.Repeat(" ", maxLen)
+
+			if len(columnEntry) > i {
+				end := len(columnEntry[i].Line)
```
Naatan, line 189: The entry literally has a length property for this purpose. Now you're just including the color codes again.

Naatan (changes requested):

internal/table/table.go
MDrakos: This and the check you commented on above had to be added as we were calculating column widths with the colour tags. I've removed that in the width calculation and now we can use `entry.Length` as we should.

internal/table/table.go
MDrakos: As per our discussion yesterday the entires represent columns. I've updated the naming to (hopefully) be more clear. You are right though that this is redundant. I've updated `renderRow` and removed this loop.

Naatan (approved):

PR #1156, opened 2020-12-07
remove environment backup logic

mdrohmann:
https://www.pivotaltracker.com/story/show/176040319

 https://www.pivotaltracker.com/story/show/176040319

### DIFF
--- a/internal/subshell/cmd/env.go
+++ b/internal/subshell/cmd/env.go
@@ -34,6 +34,33 @@ func getEnvironmentPath(userScope bool) string {
 	return `SYSTEM\ControlSet001\Control\Session Manager\Environment`
 }
 
+// unsetUserEnv clears a state cool configured environment variable
+// It only does this if the value equals the expected value (meaning if we can verify that state tool was in fact
+// responsible for setting it)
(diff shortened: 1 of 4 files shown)


internal/subshell/cmd/env.go
```diff
@@ -34,47 +34,6 @@ func getEnvironmentPath(userScope bool) string {
 	return `SYSTEM\ControlSet001\Control\Session Manager\Environment`
 }
 
-// unsetUserEnv clears a state cool configured environment variable
-// It only does this if the value equals the expected value (meaning if we can verify that state tool was in fact
-// responsible for setting it)
-func (c *CmdEnv) unset(name, ifValueEquals string) *failures.Failure {
```
Naatan, line 40: Wouldn't this still be relevant? Restoring a backup is only part of the logic here.

Naatan (changes requested):

internal/subshell/cmd/env.go
mdrohmann: Good point.  I mis-understood that function initially.

Naatan (approved):

PR #1158, opened 2020-12-09
Clean up crop and renderRow logic

Naatan:
https://www.pivotaltracker.com/story/show/175727287

@MDrakos there was too much going on to do in a review, so I went ahead and did a PR instead.

### DIFF
--- a/internal/colorize/cropped.go
+++ b/internal/colorize/cropped.go
@@ -1,61 +1,44 @@
 package colorize
 
-type Entry struct {
+type CroppedLines []CroppedLine
+
+type CroppedLine struct {
 	Line   string
 	Length int
 }
 
-func GetCroppedText(text string, maxLen int) []Entry {
-	var line string // the line we're building, including the color codes
-	var plainTextPosition int
-	var plainTextWritten int
+func GetCroppedText(text []rune, maxLen int) CroppedLines {
+	entries := make([]CroppedLine, 0)
+	colorCodes := colorRx.FindAllStringSubmatchIndex(string(text), -1)
 
-	entries := make([]Entry, 0)
-	colorCodes := colorRx.FindAllStringSubmatchIndex(text, -1)
-	runeText := []rune(text)
+	entry := CroppedLine{}
+	for pos := 0; pos < len(text); pos++ {
+		inColorTag := inRange(pos, colorCodes)
+		amend := text[pos]
+		lineEnd := amend == '\n'
 
-	for plainTextPosition < len(runeText) {
-		// If we reach an index that we recognize (ie. the start of a tag)
-		// then we write the whole tag, otherwise write by rune
-		for _, match := range colorCodes {
-			start, stop := match[0], match[1]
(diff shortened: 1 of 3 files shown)


internal/colorize/cropped.go
```diff
@@ -1,61 +1,44 @@
...
+			entry.Line += string(amend)
+			if !inColorTag {
+				entry.Length++
+			}
 		}
 
-		if plainTextPosition == len(runeText)-1 {
-			line += string(runeText[plainTextPosition])
-			entries = append(entries, Entry{line, plainTextWritten + 1})
+		if !inRange(pos+1, colorCodes) && (entry.Length == maxLen || lineEnd || pos == len(text)-1) {
```
MDrakos, line 27: Nit: This or a similar comment may be helpful
```suggestion
                // Ensure the next position is not within a color tag and check conditions that would end this entry
		if !inRange(pos+1, colorCodes) && (entry.Length == maxLen || lineEnd || pos == len(text)-1) {
```

internal/colorize/cropped_test.go
```diff
@@ -13,52 +13,52 @@ func Test_GetCroppedText(t *testing.T) {
 	tests := []struct {
 		name string
 		args args
-		want []Entry
+		want []CroppedLine
```
MDrakos, line 16: I think this is why the unit tests are failing
```suggestion
		want CroppedLines
```

internal/table/table.go
```diff
@@ -142,62 +142,46 @@ func rescaleColumns(colWidths []int, targetTotal int) {
...
-		for n, column := range columns {
-			columnMaxWidth := widths[n]
-			suffixRepeat := columnMaxWidth
-
-			// If this column has an entry for the current row we use it
-			var currentColumnText string
-			if len(column) > currentRow {
-				suffixRepeat = mathutils.MaxInt(0, columnMaxWidth-column[currentRow].Length)
-				currentColumnText = column[currentRow].Line
+	for lineNo := 0; rendered; lineNo++ {
```
MDrakos, line 162: Cool!

internal/table/table.go
```diff
@@ -142,62 +142,46 @@ func rescaleColumns(colWidths []int, targetTotal int) {
...
 
-		// If this row is empty then no columns had data for it and we are done rendering
-		if strings.TrimSpace(currentRowText) == "" {
-			emptyRow = true
-		} else {
-			result += currentRowText + linebreak
-			currentRow++
+			// Add padding and fill up missing whitespace
+			prefix := strings.Repeat(" ", padding)
+			suffix := strings.Repeat(" ", padding+(widths[columnNo]-columnLine.Length-(padding*2)))
```
MDrakos, line 174: I think we can remove that `pad` function at 187 with this change

MDrakos (changes requested):
Looks good. Just requesting changes to get the unit test fix in.

MDrakos (approved):

PR #1160, opened 2020-12-09
Add Simple outputter

MDrakos:
https://www.pivotaltracker.com/story/show/176004324


Naatan (changes requested):
This is more complicated than it needs to be, check out

http://github.com/ActiveState/cli/blob/350dd1055d72f51c166110b3b73c47c8a03aad21/internal/output/editor.go#L16-L16

You can do the same as what Editor is doing, and just override the Notice method.

Naatan (approved):

PR #1161, opened 2020-12-09
update UX for `state show`

mdrohmann:
https://www.pivotaltracker.com/story/show/176004314


internal/output/plain.go
```diff
@@ -22,6 +22,42 @@ import (
...
+
+// NewStackedOutput wraps a slice and ensures that all elements are interpreted individually but not prepended with bullet points
+func NewStackedOutput(v []interface{}) *StackedOutput {
+	return &StackedOutput{v}
+}
+
+// Stacks returns the wrapped slice
+func (s *StackedOutput) Stacks() []interface{} {
+	return s.v
+}
```
Naatan, line 59: I'm confused what this is doing? We specifically did the vertical tables story so that stories like this one could just use that mechanic.

internal/runners/show/show.go
```diff
@@ -158,24 +162,71 @@ func (s *Show) Run(params Params) error {
...
+		"",
+		output.Heading(locale.Tl("state_show_scripts_header", "Scripts")),
+		formatScripts(scripts),
+		"",
+		output.Heading(locale.Tl("state_show_platforms_header", "Platforms")),
+		platforms,
+		"",
+		output.Heading(locale.Tl("state_show_languages_header", "Languages")),
+		languages,
+	}))
```
Naatan, line 201: This is neat but totally goes against convention. Please stick to our existing patterns.

Definitely reaching the point where we need to take another look at how we handle output, but this isn't the story for that.

Naatan (changes requested):

internal/output/plain.go
mdrohmann: As I have an array/slice of different blocks that I have to print.  So, I print the output as a slice, but slices are always interpreted as lists (with bullet points before them).  So, I added this construct that allows us to create such a stacked output.

I had a different idea, that I will create a commit for.  But I think I like this one better...

internal/output/plain.go
```diff
@@ -22,6 +22,27 @@ import (
...
+
+// PrependString returns the prepended string
+func (po PrependedSliceItem) PrependString() string {
+	return po.prepend
+}
+
+// Value returns the wrapped value
+func (po PrependedSliceItem) Value() interface{} {
+	return po.v
+}
```
Naatan: This is again adding new convention and it seems like all this is doing is prepending a value to a string? I don't want to see any updates to the output package for this PR.

internal/output/plain.go
```diff
@@ -217,14 +238,21 @@ func sprintSlice(value interface{}) (string, error) {
...
 			return "", err
 		}
 
 		// prepend if stringValue does not represent a slice
-		if !isSlice(v) {
-			stringValue = " - " + stringValue
+		if vp, ok := v.(interface{ PrependString() string }); ok {
+			stringValue = vp.PrependString() + stringValue
+		} else if !isSlice(v) {
+			stringValue = "• " + stringValue
```
Naatan, line 234: These are all interesting solutions to a problem we need to solve; but not in this story. This warrants further discussion. I don't want to just accept this to unblock the story without giving this some proper thought first. And as I said on Slack this is ripe for a refactor, so rather than adding more throwaway code that's going to make it harder to replace I'd rather we just use the tools we have.

mdrohmann:
This would work, but honestly I would favor the approach that @daved suggested.  It would look like this: https://github.com/ActiveState/cli/compare/variant-david

internal/runners/show/show.go
```diff
@@ -158,24 +162,71 @@ func (s *Show) Run(params Params) error {
...
+		}{&rd},
+	)
+	plainPrint(output.Heading(locale.Tl("state_show_events_header", "Events")))
+	plainPrint(events)
+	plainPrint(output.Heading(locale.Tl("state_show_scripts_header", "Scripts")))
+	plainPrint(formatScripts(scripts))
+	plainPrint(output.Heading(locale.Tl("state_show_platforms_header", "Platforms")))
+	plainPrint(platforms)
+	plainPrint(output.Heading(locale.Tl("state_show_languages_header", "Languages")))
+	plainPrint(languages)
```
Naatan: This is overly complicated. Just tack a `MarshalOutput` onto `outputData` and `out.Print(outputData)`. Then in your marshal function do your plain prints if this is a plain output.

`output.NewFormatter` is intended more for when you want to make a really simple tweak to the output.

Naatan (changes requested):

internal/output/plain.go
mdrohmann: Okay, I have filed a story for this: https://www.pivotaltracker.com/story/show/176127475

internal/runners/show/show.go
mdrohmann: If we are okay with the `MarshalOutput` function calling output.Print instead of just returning what it should print, that is a fine solution...

Naatan (dismissed):

mdrohmann:
@Naatan I had to merge once more with our master branch.  The Critical Integration Tests are also failing because of the issue that we identified earlier.  If you approve, feel free to merge it, as I cannot do it right now...

PR #1167, opened 2020-12-11
State install presents alternatives

MDrakos:
https://www.pivotaltracker.com/story/show/176077867


internal/runners/packages/packages.go
```diff
@@ -50,7 +50,7 @@ func executePackageOperation(pj *project.Project, out output.Outputer, authentic
 		_, err = model.IngredientByNameAndVersion(name, version, ns)
 	}
 	if err != nil {
-		return locale.WrapError(err, "package_ingredient_err", "Failed to resolve an ingredient named {{.V0}}.", name)
+		return addSuggestions(err, ns, name)
```
Naatan, line 53: This is assuming that the error means there are no results. You should assert the error type.

internal/runners/packages/packages.go
```diff
@@ -127,6 +127,28 @@ func executePackageOperation(pj *project.Project, out output.Outputer, authentic
 	return nil
 }
 
+func addSuggestions(err error, ns model.Namespace, name string) error {
```
Naatan, line 130: This shouldn't be needing any error as its input.

internal/runners/packages/packages.go
```diff
@@ -127,6 +127,28 @@ func executePackageOperation(pj *project.Project, out output.Outputer, authentic
...
+	results, searchFail := model.SearchIngredients(ns, name)
+	if searchFail != nil {
```
Naatan, line 132:
```suggestion
	results, err := model.SearchIngredients(ns, name)
	if err != nil {
```

internal/runners/packages/packages.go
```diff
@@ -127,6 +127,28 @@ func executePackageOperation(pj *project.Project, out output.Outputer, authentic
...
+		// Log and return generic error if search failed
+		logging.Error("Failed to search for ingredients with namespace: %s and name: %s, got error: %v", ns.String(), name, searchFail)
```
Naatan, line 134:
```suggestion
```

internal/runners/packages/packages.go
```diff
@@ -127,6 +127,28 @@ func executePackageOperation(pj *project.Project, out output.Outputer, authentic
...
+	}
+
+	maxResults := 5
+	if len(results) > maxResults {
+		results = results[:maxResults]
+	}
+
+	suggestions := make([]string, maxResults)
+	for i := range results {
+		suggestions[i] = fmt.Sprintf(" - %s", *results[i].Ingredient.Name)
```
Naatan, line 145: Should just use a slice of strings to print these. The plain outputter will translate that into a readable list.

Naatan (changes requested):

PR #1170, opened 2020-12-15
replace validation for package installation

mdrohmann:
https://www.pivotaltracker.com/story/show/176076302

### DIFF
--- a/internal/runners/packages/packages.go
+++ b/internal/runners/packages/packages.go
@@ -143,7 +143,7 @@ func addSuggestions(ns model.Namespace, name string) error {
 	for i := range results {
 		suggestions[i] = fmt.Sprintf(" - %s", *results[i].Ingredient.Name)
 	}
-	suggestions = append(suggestions, locale.Tl("ingredient_alternatives_more", " - .. (to see more results run `state search {{.V0}}`)", name))
+	suggestions = append(suggestions, locale.Tr(fmt.Sprintf("%s_ingredient_alternatives_more", ns.Type()), name))
 
 	return locale.NewInputError("package_ingredient_alternatives", "Could not match {{.V0}}. Did you mean:\n\n{{.V1}}", name, strings.Join(suggestions, "\n"))
 }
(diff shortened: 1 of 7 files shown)


internal/runners/packages/packages.go
```diff
@@ -82,6 +70,16 @@ func executePackageOperation(pj *project.Project, out output.Outputer, authentic
 
 	logging.Debug("Order changed: %v", orderChanged)
 
+	// Verify that the provided package actually exists (the vcs API doesn't care)
+	_, err = model.FetchRecipe(commitID, pj.Owner(), pj.Name(), &model.HostPlatform)
+	if err != nil {
+		rerr := &inventory_operations.ResolveRecipesBadRequest{}
+		if errors.As(err, &rerr) {
```
mdrohmann, line 77: Note, that this check is rather unspecific.  The `ResolveRecipesBadRequest` error could happen for other reasons.  Its payload has nothing that specifically says that an ingredient was missing: https://activestate.slack.com/archives/C7PEK0LLU/p1608058443447900

One thought that I had to work around this, is update the suggestions function in a way such that it also checks if the name and version that are provided are actually not matching any of the suggestions.  But then we still need to know if the features are case-sensitive or not, which is probably knowledge that should live in the solver.  So, maybe we need a follow-up story for pb?

Naatan: I think for this story we'd want to just make sure that you still report the error value of FetchRecipe.

pkg/platform/model/recipe.go
```diff
@@ -160,11 +160,11 @@ func FetchRecipe(commitID strfmt.UUID, owner, project string, hostPlatform *stri
 
 		orderBody, _ := json.Marshal(params.Order)
 		switch rrErr := err.(type) {
-		case *iop.SolveOrderDefault:
+		case *iop.ResolveRecipesDefault:
 			msg := *rrErr.Payload.Message
 			logging.Error("Could not solve order, error: %s, order: %s", msg, string(orderBody))
 			return nil, locale.WrapInputError(err, "err_solve_order", "", msg)
-		case *iop.SolveOrderBadRequest:
```
Naatan, line 167: Nice catch!

Naatan (changes requested):
Looks good! I think all that's missing is ensuring the FetchRecipe error is always reported (even when suggesting alternatives).

Naatan (dismissed):

Naatan (approved):

PR #1172, opened 2020-12-15
Start info command

daved:
https://www.pivotaltracker.com/story/show/176004293

 https://www.pivotaltracker.com/story/show/176004293


internal/runners/packages/info.go
```diff
@@ -0,0 +1,238 @@
...
+		if iv.Ingredient.Name != nil {
+			res.name = *iv.Ingredient.Name
+		}
+
+		if iv.Ingredient.Description != nil {
+			res.Description = *iv.Ingredient.Description
+		}
+
+		if iv.Ingredient.Links != nil && iv.Ingredient.Links.Self != nil {
+			res.PkgDetailsTable.Link = iv.Ingredient.Links.Self.String()
```
Naatan, line 116: This is an API link that shows JSON information. We want the link to the package itself (eg. the github page). If the API doesn't provide this then defer to the AC:

> Any information that's not provided by the search API can be disregarded but should be reported to Pete

internal/runners/packages/info.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+	}
+
+	if iv.LatestVersion != nil {
+		if iv.LatestVersion.Version != nil {
+			res.latestVersion = *iv.LatestVersion.Version
+		}
+
+		if iv.LatestVersion.CopyrightText != nil {
+			res.PkgDetailsTable.Copyright = *iv.LatestVersion.CopyrightText
```
Naatan, line 126: Why is this printing as a JSON array? eg.

```
["Copyright (c) Django Software Foundation and individual contributors."]  
```

internal/runners/packages/info.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		if iv.LatestVersion.LicenseExpression != nil {
+			res.PkgDetailsTable.License = *iv.LatestVersion.LicenseExpression
```
Naatan, line 130: Also printing as a JSON array;

```
["BSD-3-Clause"]  
```

Naatan (changes requested):

internal/runners/packages/info.go
daved: Discussed in Slack: We're not showing this information to users at this time while changes are made in the backend. (Same for the License field.)

Naatan (approved):

PR #1181, opened 2020-12-17
Add --set-channel flag and move `update --lock` to `lock` subcommand

Naatan:
https://www.pivotaltracker.com/story/show/176151454


internal/runners/update/update.go
```diff
@@ -40,75 +38,14 @@ func New(prime primeable) *Update {
...
-	}
-
-	u.out.Print(locale.Tl("version_lock_updated", "Locked version updated to {{.V0}}", constants.Version))
-	return nil
-}
-
-func (u *Update) runUpdateGlobal() error {
 	u.out.Notice(locale.Tl("updating_version", "Updating State Tool to latest version available."))
 
+	channel := fetchChannel(params.Channel, true)
```
mdrohmann, line 43: So, here are ignoring the channel name configured in the activestate.yaml file, and instead just update or switch the "global" channel of the installed state tool. That confused me for a while, but it makes sense!  I actually like this separation! :+1:

mdrohmann (approved):
Took a while to understand all the changes, but it LGTM!

PR #1183, opened 2020-12-17
log to rollbar when on release or beta branch

mdrohmann:
https://www.pivotaltracker.com/story/show/176212180


internal/constants/constants.go
```diff
@@ -116,10 +116,13 @@ const DefaultRSABitLength int = 4096
...
-const StableBranch = "stable"
+// ReleaseBranch is the branch used for release builds
+const ReleaseBranch = "release"
 
-// UnstableBranch is the branch used for unstable builds
+// BetaBranch is the branch used for beta builds
+const BetaBranch = "beta"
+
+// UnstableBranch is the branch used for unstable builds.  Note that this branch is not used anymore
 const UnstableBranch = "unstable"
```
Naatan, line 126: This can be dropped

Naatan (changes requested):

internal/constants/constants.go
mdrohmann: Shouldn't we leave it in for a little while in case there are installation with a locked version on the unstable branch?

mdrohmann: nvmd. They wouldn't be running this state tool... I am going to remove it.

Naatan (approved):

PR #1186, opened 2020-12-17
Ensure clean uninstall cmd removes all config files

daved:
https://www.pivotaltracker.com/story/show/176096297

 https://www.pivotaltracker.com/story/show/176096297


daved:
@Naatan Initially, I moved all config directory removal logic to the config package. However, that resulted in a very large diff. Worse, the cascading changes caused the control flow to be even harder to control than it is currently (i.e. debug logging was showing up and not able to be controlled well due to logging being called from init and init-related functions).

I reverted those changes, and produced a diff that will, for this story, ensure that all but the current log file is removed from the config directory. Once the logging and config packages are refactored to better be a part of the dependency tree, we will more easily be able to remove all files.

Naatan (approved):

Naatan:
Nice work keeping it simple!

PR #1191, opened 2020-12-22
abstract viper configuration

mdrohmann:
https://www.pivotaltracker.com/story/show/165094030


mdrohmann:
I added a follow-up story: https://www.pivotaltracker.com/story/show/176253321

internal/config/instance.go
```diff
@@ -29,21 +33,112 @@ type Instance struct {
...
+// Get returns the default configuration instance
+func Get() *Instance {
+	if defaultConfig == nil {
+		defaultConfig = New()
+	}
+	return defaultConfig
+}
+
+// WriteConfig writes the configuration
+func (i *Instance) WriteConfig() error {
```
Naatan, line 98: I think we can drop this, from a quick search the code that uses `viper.WriteConfig` should probably be using `config.Save`.

Naatan (changes requested):
This looks great! Just one minor nit that stood out to me which we might as well address in this PR.

Naatan (approved):

PR #1193, opened 2020-12-22
unify project mapping store

mdrohmann:
https://www.pivotaltracker.com/story/show/176225844


internal/runners/activate/namespace.go
```diff
@@ -51,12 +54,6 @@ func (r *NamespaceSelect) Run(namespace string, preferredPath string) (string, e
 		return "", err
 	}
 
-	// Save path for future use
```
mdrohmann, line 54: The project mapping is added whenever we source a project file, therefore also during the initial activation. And that suffices.  I tested that

```
state activate ActiveState/Perl-5.32 --path /tmp/a-new-location
state projects
```

shows the new location of Perl-5.32.

Naatan (approved):
Looking good!

PR #1196, opened 2020-12-23
Start story-cleanup script

daved:
https://www.pivotaltracker.com/story/show/174891360

### DIFF
--- a/scripts/story-cleanup/story-cleanup.py
+++ b/scripts/story-cleanup/story-cleanup.py
@@ -19,7 +19,7 @@ PT_DAYS_AGED = 190 if 'PT_DAYS_AGED' not in os.environ else int(os.getenv('PT_DA
 if PT_DAYS_AGED < 30:
     PT_DAYS_AGED = 30
 
-def params(base, added):
+def mergeDicts(base, added):
     ps = base.copy()
     ps.update(added)
     return ps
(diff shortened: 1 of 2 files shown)


daved:
Notes:
1. The amount of stories handled per run is limited by the API or a hardcoded value of 50 (whichever is less). I picked this value because it seems like it would cover most weekly/monthly needs, but minimize affected stories if some mistake occurs.
2. The secret which defines the primary project id will need to be updated for the script to work: `project.PT_PROJID_PRIMARY`

daved:
Testing:
1. The gathering logic was tested on our current project to ensure only relevant stories were collected.
2. The move logic was tested using a throwaway project to avoid harming our current project.

If there is some way to back-up the project before running this, I'd feel better. Otherwise, the limit can be decreased for a first run.

scripts/story-cleanup/story-cleanup.py
```diff
@@ -0,0 +1,54 @@
...
+
+from requests.exceptions import HTTPError
+
+PT_PROJID_PRI = os.getenv('PT_PROJID_PRIMARY')
+PT_PROJID_STR = os.getenv('PT_PROJID_STORAGE')
+PT_URLPFX_FMT = 'https://www.pivotaltracker.com/services/v5/projects/{}'
+PT_URLPFX_PRI = PT_URLPFX_FMT.format(PT_PROJID_PRI)
+PT_HDRS_BASE = {'X-TrackerToken':os.getenv('PT_API_TOKEN')}
+PT_PRMS_BASE = {'date_format':'millis', 'limit':50}
+PT_DAYS_AGED = 190 if 'PT_DAYS_AGED' not in os.environ else int(os.getenv('PT_DAYS_AGED'))
```
Naatan, line 17: Can you please rename these to make them a little more humanly readable.

eg. `PT_PROJID_STR` > `PT_PROJID_STORAGE`

scripts/story-cleanup/story-cleanup.py
```diff
@@ -0,0 +1,54 @@
...
+    response = requests.put(PT_URLPFX_PRI+'/stories/'+storyID, headers=PT_HDRS_BASE, json=payload)
+    response.raise_for_status()
+
+def main():
+    try:
+        agedStories = getAgedStories()
+        for story in agedStories:
+            moveStory(story)
+    except HTTPError, e:
+        print(e)
```
Naatan, line 51: Then why even catch it? :p

Naatan (changes requested):

scripts/story-cleanup/story-cleanup.py
daved: I fixed this a little, but the main reason is that the printed output is more easily readable than the trace.

Naatan (approved):

PR #1197, opened 2021-01-05
state secrets UX update

MDrakos:
https://www.pivotaltracker.com/story/show/176004310

### DIFF
--- a/locale/en-us.yaml
+++ b/locale/en-us.yaml
@@ -631,9 +631,11 @@ secrets_header_description:
 secrets_header_usage:
   other: Usage
 secrets_row_value_set:
-  other: Defined
+  other: "[SUCCESS]Defined[/RESET]"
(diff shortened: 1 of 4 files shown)


locale/en-us.yaml
```diff
@@ -631,9 +631,11 @@ secrets_header_description:
 secrets_header_usage:
   other: Usage
 secrets_row_value_set:
-  other: Defined
+  other: "[SUCCESS]Defined[/RESET]"
 secrets_row_value_unset:
-  other: Undefined
+  other: "[ERROR]Not defined[/RESET]"
```
MDrakos, line 636: These are different from the mock as the output was empty with the special characters. Filed a story here: https://www.pivotaltracker.com/story/show/176372179

Naatan (approved):

PR #1198, opened 2021-01-05
State pull can replace project url

mdrohmann:
https://www.pivotaltracker.com/story/show/176209787

### DIFF
--- a/internal/runners/pull/pull.go
+++ b/internal/runners/pull/pull.go
@@ -80,7 +80,7 @@ func (p *Pull) Run(params *PullParams) error {
 	if params.SetProject != "" {
 		related, err := areCommitsRelated(*target.CommitID, p.project.CommitUUID())
 		if !related && !params.Force {
-			confirmed, err := p.prompt.Confirm(locale.T("confirm"), locale.Tl("confirm_unrelated_pull_set_project", "If you switch to {{.V0}}, you may lose changes to your project. Are you sure you want to do this?", target.String()), false)
+			confirmed, err := p.prompt.Confirm(locale.T("confirm"), locale.Tl("confirm_unrelated_pull_set_project", "If you switch to {{.V0}}, you may lose changes to your project. Are you sure you want to do this?", target.String()), new(bool))
 			if err != nil {
 				return locale.WrapError(err, "err_pull_confirm", "Failed to get user confirmation to update project")
 			}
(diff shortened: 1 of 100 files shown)


internal/runners/pull/pull.go
```diff
@@ -48,30 +59,49 @@ func (f *outputFormat) MarshalOutput(format output.Format) interface{} {
...
+	target, err := targetProject(p.project, params.SetProject)
 	if err != nil {
-		return locale.WrapInputError(err, "err_pull_commit", "Could not retrieve the latest commit for your project.")
+		return errs.Wrap(err, "Unable to determine target project")
+	}
+
+	if params.SetProject != "" {
+		related, err := areCommitsRelated(*target.CommitID, p.project.CommitUUID())
+		if !related && !params.Force {
+			confirmed, err := p.prompt.Confirm(locale.T("confirm"), locale.Tl("confirm_unrelated_pull_set_project", "If you switch to {{.V0}}, you may lose changes to your project.", target.String()), false)
```
Naatan, line 80: This is missing the "Are you sure you want to do this?" Otherwise we're just prompting them for yes/no without actually giving them a prompt.

internal/runners/pull/pull.go
```diff
@@ -48,30 +59,49 @@ func (f *outputFormat) MarshalOutput(format output.Format) interface{} {
...
+			if err != nil {
+				return locale.WrapError(err, "err_pull_confirm", "Failed to get user confirmation to update project")
+			}
+			if !confirmed {
+				return locale.NewInputError("err_pull_aborted", "Pull aborted by user")
+			}
+		}
+
+		err = p.project.Source().SetNamespace(target.Owner, target.Project)
```
Naatan, line 89: This seems to be redundant with line 97

Naatan (changes requested):

internal/runners/pull/pull.go
mdrohmann: Very good point!

internal/runners/pull/pull.go
mdrohmann: Yes, thanks for catching.  Line 97 should never have been there...

Naatan (approved):

mdrohmann:
I had to merge with master to update the confirmation prompt, and also noticed that I accidentally added a commit from my other PR.  That is reverted now.

Naatan (approved):

PR #1199, opened 2021-01-07
remove os.Exit from config

mdrohmann:
https://www.pivotaltracker.com/story/show/176253321


cmd/state/main.go
```diff
@@ -100,7 +100,10 @@ func run(args []string, isInteractive bool, out output.Outputer) (int, error) {
 	verbose := os.Getenv("VERBOSE") != "" || argsHaveVerbose(args)
 	logging.CurrentHandler().SetVerbose(verbose)
 
-	cfg := config.Get()
+	cfg, err := config.Get()
```
Naatan, line 103: We should rename this to `New()`. `Get()` to me conveys the deprecated "global" way of doing things.

internal/analytics/analytics.go
```diff
@@ -202,11 +208,15 @@ func sendEventAndLog(category, action, label string, dimensions map[string]strin
 
 func sendEvent(category, action, label string, dimensions map[string]string) error {
 	if deferAnalytics {
-		if err := deferEvent(category, action, label, dimensions); err != nil {
+		// TODO: figure out a way to pass configuration
+		cfg, err := config.Get()
```
Naatan, line 212: Was this too much for this PR? If so please file a story.

internal/deprecation/deprecation.go
```diff
@@ -70,7 +71,11 @@ func Check() (*Info, error) {
 
 // CheckVersionNumber will run a Checker.Check with defaults
 func CheckVersionNumber(versionNumber string) (*Info, error) {
-	checker := NewChecker(DefaultTimeout, config.Get())
+	cfg, err := config.Get()
```
Naatan, line 74: This should be fed in.

internal/keypairs/local.go
```diff
@@ -92,8 +102,12 @@ func DeleteWithDefaults() error {
 }
 
 // LocalKeyFilename returns the full filepath for the given key name
-func LocalKeyFilename(keyName string) string {
-	return filepath.Join(config.Get().ConfigPath(), keyName+".key")
+func LocalKeyFilename(keyName string) (string, error) {
+	cfg, err := config.Get()
```
Naatan, line 106: Should be fed in

internal/locale/locale.go
```diff
@@ -32,7 +33,14 @@ var exit = os.Exit
 func init() {
 	logging.Debug("Init")
 
-	config.Get().SetDefault("Locale", "en-US")
+	locale := getLocaleFlag()
+	cfg, err := config.Get()
```
Naatan, line 37: Should be fed in

internal/locale/locale.go
```diff
@@ -88,17 +91,21 @@ func getLocaleFlag() string {
...
-		exit(1)
-		return
+		return errs.New("Locale does not exist: %s", localeName)
 	}
 
 	translateFunction, _ = i18n.Tfunc(localeName)
 	_ = translateFunction
 
-	config.Get().Set("Locale", localeName)
+	cfg, err := config.Get()
```
Naatan, line 102: Should be fed in. I understand this is problematic since locale itself is "global". But we can set the cfg param inside this package from the main package.

internal/logging/defaults.go
```diff
@@ -132,8 +134,17 @@ func init() {
 	handler := &fileHandler{DefaultFormatter, nil, safeBool{}}
 	SetHandler(handler)
 
+	cfg, err := config.Get()
```
Naatan, line 137: Should be fed in

internal/logging/rollbar.go
```diff
@@ -47,8 +47,16 @@ func SetupRollbar() {
 		data["platform_os"] = runtime.GOOS
 	})
 
+	cfg, err := config.Get()
```
Naatan, line 50: Should be fed in

internal/machineid/machineid.go
```diff
@@ -17,12 +17,18 @@ func uniqID(machineIDGetter func() (string, error), uuidGetter func() string) st
 		return machID
 	}
 
-	machineID := config.Get().GetString("machineID")
+	cfg, err := config.Get()
```
Naatan, line 20: Should be fed in

pkg/cmdlets/auth/signup.go
```diff
@@ -88,7 +88,12 @@ func downloadTOS() (string, error) {
 	}
 	defer resp.Body.Close()
 
-	tosPath := filepath.Join(config.Get().ConfigPath(), "platform_tos.txt")
+	cfg, err := config.Get()
```
Naatan, line 91: Should be fed in

pkg/platform/authentication/auth.go
```diff
@@ -38,19 +38,32 @@ type Auth struct {
...
+		persist = New(cfg)
 	}
 	return persist
 }
 
 // Client is a shortcut for calling Client() on the persisted auth
 func Client() *mono_client.Mono {
-	return Get().Client()
+	a := Get()
+	return a.Client()
```
Naatan, line 66: Pointless change?

pkg/platform/runtime/runtime.go
```diff
@@ -35,7 +35,11 @@ func NewRuntime(projectDir string, commitID strfmt.UUID, owner string, projectNa
 		logging.Debug("In NewRuntime: resolved project dir is: %s", resolvedProjectDir)
 	}
 
-	installPath := filepath.Join(config.Get().CachePath(), hash.ShortHash(resolvedProjectDir))
+	cfg, err := config.Get()
```
Naatan, line 38: Should be fed in

pkg/projectfile/projectfile.go
```diff
@@ -709,7 +714,12 @@ func (p *Project) save(path string) error {
 		return errs.Wrap(err, "f.Write %s failed", path)
 	}
 
-	storeProjectMapping(fmt.Sprintf("%s/%s", p.parsedURL.Owner, p.parsedURL.Name), filepath.Dir(p.Path()))
+	cfg, err := config.Get()
```
Naatan, line 717: Should be fed in

pkg/projectfile/projectfile.go
```diff
@@ -852,7 +862,11 @@ func getProjectFilePathFromWd() (string, error) {
 }
 
 func getProjectFilePathFromDefault() (string, error) {
-	defaultProjectPath := config.Get().GetString(constants.GlobalDefaultPrefname)
+	cfg, err := config.Get()
```
Naatan, line 865: Should be fed in

Naatan (changes requested):
I think there should really only be one instance of `config.New()` (ie. `.Get()` as you currently have it). In the main package. Disregarding tests of course.

cmd/state/main.go
mdrohmann: `Get()` is currently still a "global" function as it returns a persistent configuration instance - if that exists.  It just now also returns an error.
I didn't remove that behavior yet, because

- `analytics` and `locale` use the configuration inside of their `init()` functions
- Changing it in the authentication package, means updating every `authentication.Get()`, `authentication.Client()` and `authentication.ClientAuth()` call, and that happens in 127 (!) places. I forgot to make a story for that though, which I did now: https://www.pivotaltracker.com/story/show/176419803

I think that I will feed the configuration instance into the remaining places first.  Then I will see if I could try to change the `locale` and `analytics` packages, as you suggested... by adding a `SetConfig()` method from the main function.

I could also give the `auth` refactor a try, of course...   But maybe it is better to first refactor the `authentication` package, such that it does not have the globally persisting object anymore.

Naatan: Ok, I guess it makes sense to scope this down into another story.

Naatan: https://www.pivotaltracker.com/story/show/176422561

mdrohmann:
I removed `config.Get` from a few more packages, but it is still in `analytics`, `locale` and `logging`, because IMO it seems better to do this as part of a different PR, so it is simpler to review, and might be slightly more complicated.

It is also still in `projectfile` and `authentication`, because these are used so wildly, and IMO we should remove their `Get` functions first.

internal/logging/rollbar.go
```diff
@@ -8,17 +8,21 @@ import (
...
+	InstallSource() string
+}
+
 type delayedLog struct {
 	level string
 	msg   interface{}
 }
 
 var delayedLogs []delayedLog
+var rollbarCfg Configurable
```
Naatan, line 25: Nit: Why are we calling this "rollbarCfg"? This is just config, and rollbar is free to use it. Seems weird to appropriate it to rollbar by name.

Naatan (dismissed):

internal/logging/rollbar.go
mdrohmann: Good point.

internal/logging/rollbar.go
```diff
@@ -64,8 +64,8 @@ func SendToRollbarWhenReady(level string, msg interface{}) {
 	delayedLogs = append(delayedLogs, delayedLog{level, msg})
 }
 
-func UpdateRollbarConfig(cfg Configurable) {
-	rollbarCfg = cfg
+func UpdateRollbarConfig(c Configurable) {
```
Naatan, line 67: Should rename this function as well then.

Naatan (changes requested):

Naatan (approved):

PR #1202, opened 2021-01-12
State help formatting fix

MDrakos:
https://www.pivotaltracker.com/story/show/175984479


Naatan (changes requested):
I don't think we should be constructing the data in the template. Construct the data outside of the template, then print it inside. We can use some helper functions but they should be minimal and geared towards formatting, not data construction.

Naatan (approved):

PR #1203, opened 2021-01-12
add vscode integration tests

mdrohmann:


Naatan (changes requested):
Looks good! But it's missing state init: https://github.com/ActiveState/vscode-activestate-platform/blob/restructure/src/model/state/init.ts#L4
-

mdrohmann:
> But it's missing state init: https://github.com/ActiveState/vscode-activestate-platform/blob/restructure/src/model/state/init.ts#L4

The init test goes together with the one for `state push` here: https://github.com/ActiveState/cli/pull/1203/files#diff-68d1e25c0a3052c0c38745dd240c544bbbc9f321a687b98e7654df9ec0ec9dc8R21-R27  I changed its name for clarification.

Naatan (dismissed):

Naatan (approved):

PR #1207, opened 2021-01-14
Update GetCroppedText to better handle multi-byte characters

MDrakos:
https://www.pivotaltracker.com/story/show/176372179

### DIFF
--- a/internal/colorize/cropped.go
+++ b/internal/colorize/cropped.go
@@ -7,14 +7,13 @@ type CroppedLine struct {
 	Length int
 }
 
-func GetCroppedText(text []rune, maxLen int) CroppedLines {
+func GetCroppedText(text string, maxLen int) CroppedLines {
 	entries := make([]CroppedLine, 0)
-	colorCodes := colorRx.FindAllStringSubmatchIndex(string(text), -1)
+	colorCodes := colorRx.FindAllStringSubmatchIndex(text, -1)
 
 	entry := CroppedLine{}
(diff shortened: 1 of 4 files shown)


Naatan:
@mdrohmann I'm still not entirely clear on how runes work. Can you double check this won't have any adverse effects?

internal/colorize/cropped.go
```diff
@@ -7,14 +7,13 @@ type CroppedLine struct {
 	Length int
 }
 
-func GetCroppedText(text []rune, maxLen int) CroppedLines {
+func GetCroppedText(text string, maxLen int) CroppedLines {
 	entries := make([]CroppedLine, 0)
-	colorCodes := colorRx.FindAllStringSubmatchIndex(string(text), -1)
+	colorCodes := colorRx.FindAllStringSubmatchIndex(text, -1)
```
mdrohmann, line 12: This makes sense.  It's certainly correct to use `[]rune` as the argument type here. 

`len("✔ab") == 4`, `len([]rune("✔ab")) == 3`   As we are trying to limit the "visible width", we are interested in the latter number here.

BTW:  This will still not be enough, if we every support weird East-Asian locales, as these runes may be two characters wide(!).  Then we'll have to use something [like this](https://github.com/mattn/go-runewidth)

But I think that this is a problem for our future selves...

An example of runes that are two characters wide:
```
つつ
1234
```

mdrohmann (approved):

PR #1208, opened 2021-01-14
Add Support for Completions (only for bash, zsh, fish)

Naatan:
https://www.pivotaltracker.com/story/show/176471025

Lots of vendor changes because I upgraded cobra and ran `mod tidy` and `mod vendor`


internal/subshell/bash/bash.go
```diff
@@ -45,14 +50,39 @@ func (v *SubShell) SetBinary(binary string) {
...
+		return errs.Wrap(err, "RcFile failure")
 	}
 
 	env = sscommon.EscapeEnv(env)
-	return sscommon.WriteRcFile("bashrc_append.sh", filepath.Join(homeDir, ".bashrc"), envType, env)
+	return sscommon.WriteRcFile("bashrc_append.sh", rcFile, envType, env)
+}
+
+func (v *SubShell) WriteCompletionScript(completionScript string) error {
+	dir := "/usr/local/etc/bash_completion.d/"
```
mdrohmann, line 64: Are there no directories relative to the HOME directory where we can write to?  This would probably generate an error unless run as a super-user...

mdrohmann (changes requested):
Looks great.  But maybe we can write the bash completions in a user directory...

internal/subshell/bash/bash.go
Naatan: There's `~/.bash_completion` but that's a single file that's used for ALL completions. From my research I couldn't see it being used much.

Fact is bash completions don't even work unless you install the bash-completion plugin for bash, which I can't imagine is very widely used. It just feels like a waste of effort to do more than what is being done here.

mdrohmann (approved):

PR #1213, opened 2021-01-25
Build success predicitons

mdrohmann:
This is a proof of concept PR for how to add build success predictions to the
State Tool.

What I learned:

- Code to compute the artifact DAG had to be copied from the Head-chef.  So, this part should probably be exposed server-side in the head-chef.
- Currently, the build-status for every artifact is queried individually, slowing down the processing significantly.  This can be speed up by putting the code in the head-chef as well, where it can be handled with a signle SQL query.
- Best value seems to be in the State Tool or on the Dashboard as part of the `state install` (adding a package) flow. 

Suggestions:

- Create a head-chef endpoint to return artifact BuildState values for a bulk of artifact ids.
- (Possibly) add a GraphQL query to return prediction information (for a set of orders)
- Design a new `state install` flow, that informs the user about the predicted build success before the build is tried.
- Design a `state install` wizard, that provides suggestions in case a build is known to fail.  Based on which artifacts failed to build, we can suggest to select different versions for these artifacts.


mdrohmann:
This PR has just been created for documentation purposes.

PR #1214, opened 2021-01-25
junk: Experiment with adding scripts by directory/filename/path-with-glob

daved:
_This PR is for reference only. Please do not merge without additional communication._

Based on a number of requests and my own desire, these changes add a new block the AS.Y file so that `state run`able scripts can be added by filename/path-with-glob.

Usage notes:
- This adds a new block titled "script_sources"
- An entry may refer to one or more locations
  - A location may refer to one file or many via glob (i.e. one entry may grab many files)
- Interpreter type must be delineated
- Filename extensions can be, optionally, trimmed
- Constants may be set per entry (i.e. set of files)

Dev notes:
-  This hits projectfile and project package entanglements
- virtualenv structures and flow should be improved/simplified
  - Note that constants needed to be injected into PrepareVirtualEnv
    - This dependency can, likely, be more cleanly modeled/abstracted
- We may need to rethink how scripts and other "objects" are handled
  - Consider lazy loading files and/or using caching mechanisms to minimize process duration

Suggestions (It would be nice to):
- best-guess the interpreter and notify users of scripts operating without 100% confidence
- provide language-level libraries to easily call functionality like `state secrets`
- consider a different data structure for constants so that the yaml is more expressive
  - Maybe, instead of Name/Value structs, the projectfile representation could just be a map[string]string
    - This seems like a great example where the projectfile and project packages could show their distinction

### DIFF
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -21,11 +21,16 @@ constants:
         GO_BUILD_TAGS="CI"
       fi
 script_sources:
-  - name: cleanit
+  - name: story-cleanup
+    description: "Runs Python script to move old stories from primary project to storage project"
     language: python3
     strip_ext: true
-    locations:
-      - ./scripts/pys/*
+    locations: [./scripts/story-cleanup/story-cleanup.py]
+    constants:
+      - {name: PT_API_TOKEN,      value: $secrets.project.PT_API_TOKEN}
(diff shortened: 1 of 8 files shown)


daved:
Closed for archival as requested by @Naatan

daved:
Reopening to change targeted branch name.

daved:
Closing in favor of #1216

PR #1215, opened 2021-01-25
add ArgMarshaler for package@version args

mdrohmann:
https://www.pivotaltracker.com/story/show/176194323


mdrohmann:
This unifies the behavior for arguments of `language@version`, `platform@version`, `package@version` and `channel@version`.

internal/captain/nameversion.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (nv *NameVersion) Set(arg string) error {
+	nameArg := strings.Split(arg, "@")
+	nv.name = nameArg[0]
+	if len(nameArg) == 2 {
+		nv.version = nameArg[1]
+	}
+	if len(nameArg) > 2 {
+		return errors.New("invalid format")
```
MDrakos, line 23: I think this needs localization and also could be a more descriptive error

internal/config/config_test.go
```diff
@@ -108,6 +108,7 @@ func (suite *ConfigTestSuite) testNoHomeRunner() {
 }
 
 func (suite *ConfigTestSuite) TestNoHome() {
+	suite.T().Skip()
```
MDrakos, line 111: Was this for debugging?

MDrakos (changes requested):
I wonder if `captain` is the right place for `nameversion.go`. That or maybe the types are too specific to our commands? I'm not sure if it's appropriate to give `captain` knowledge of what a `PackageVersion` is. Though I will defer to @Naatan for his thoughts.

Other than that I like how it's used in the runners and only noted two minor comments to address.

internal/config/config_test.go
mdrohmann: Thanks for catching!

mdrohmann:
I think you are right about the location for the `NameVersion` specializations.  I moved the specializations to their respective runners.

MDrakos (dismissed):

internal/runners/packages/packages.go
```diff
@@ -21,6 +22,18 @@ import (
...
 )
 
+type PackageVersion struct {
+	captain.NameVersion
+}
+
+func (pv *PackageVersion) Set(arg string) error {
+	err := pv.NameVersion.Set(arg)
+	if err != nil {
+		return locale.NewError("err_package_format", "The package and version provided is not formatting correctly, must be in the form of <package>@<version>")
```
Naatan, line 32: Should be an input error

internal/captain/nameversion.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (nv *NameVersion) Set(arg string) error {
+	nameArg := strings.Split(arg, "@")
+	nv.name = nameArg[0]
+	if len(nameArg) == 2 {
+		nv.version = nameArg[1]
+	}
+	if len(nameArg) > 2 {
+		return locale.NewError("name_version_format_err", "Invalid format: Should be <name@version>")
```
Naatan, line 22: Should be input error

internal/runners/platforms/platforms.go
```diff
@@ -7,11 +7,24 @@ import (
...
 )
 
+type PlatformVersion struct {
+	captain.NameVersion
+}
+
+func (pv *PlatformVersion) Set(arg string) error {
+	err := pv.NameVersion.Set(arg)
+	if err != nil {
+		return locale.NewError("err_platform_format", "The platform and version provided is not formatting correctly, must be in the form of <platform>@<version>")
```
Naatan, line 23: Input error

Naatan (changes requested):

Naatan (changes requested):
You're also loosing the original error I just noticed. Should be WrapInputError not NewInputError.

Naatan (approved):

PR #1218, opened 2021-01-25
Building remotely progress fix

MDrakos:
https://www.pivotaltracker.com/story/show/176210040

@mdrohmann your suggestion was correct. Initially, I had the call to `BuildFinished()` in the wrong place. After moving the call I was unable to reproduce the bug again.


mdrohmann (approved):

Naatan (changes requested):
This still feels very fragile. I think we should ensure BuildFinished is ALWAYS called when the error channel is written.

Naatan (approved):

PR #1222, opened 2021-01-26
Analytics S3 pixel

MDrakos:
https://www.pivotaltracker.com/story/show/176649630


internal/analytics/analytics.go
```diff
@@ -206,6 +208,15 @@ func sendEventAndLog(category, action, label string, dimensions map[string]strin
 }
 
 func sendEvent(category, action, label string, dimensions map[string]string) error {
+	err := sendGAEvent(category, action, label, dimensions)
```
mdrohmann, line 211: I think that `sendGAEvent` should only be [this part](https://github.com/ActiveState/cli/pull/1222/files#diff-895f0dd442920ec2aa8a34401e292e8d73c29054d75da4a2541f10380e901eafL223-L238), and the handling of deferred analytics event should be handled by both the S3 Pixel and GA ie., in this method.

Right now, the deferred events are always sent to S3, even when they should be deferred.

mdrohmann: No wait!  The GA Events are also sent! 
@Naatan Shouldn't there be a `return nil` after line 220, as we do *not* want to send anything if `deferAnalytics` is `true`?

If I understand this correctly, we are still sending a GA event with every `state shim` command immediately, but *also* as a deferred bulk operation later...

internal/download/get.go
```diff
@@ -77,30 +77,31 @@ func httpGetWithProgressRetry(url string, progress *progress.Progress, attempt i
...
+			}
+		}()
+	}
 
 	_, err = io.Copy(&dst, src)
 	if err != nil {
 		logging.Debug("Reading body failed: %s", err)
 		if attempt <= retries {
-			return httpGetWithProgressRetry(url, progress, attempt+1, retries)
+			return httpGetWithProgressRetry(url, prog, attempt+1, retries)
```
mdrohmann, line 99: I wonder how the progress bar actually behaves for retries here.  The deferred `bar.Abort()` call will not have been issued at this point.  So, the (not really) aborted `ByteProgressBar` that we added above is still active somehow.  I suppose it is a little too much for this PR (which isn't about this anyways), but maybe we could have a follow-up story to investigate this...

mdrohmann (changes requested):

internal/analytics/analytics.go
Naatan: I think you're right.. we need to actually cancel out when deferring.

internal/download/get.go
MDrakos: Good point, if this is an issue then it existed before the changes here and we haven't seen any reports or noticed any issues with the progress bar on retries. I agree we should investigate to be sure and potentially update this function to use our retry library.

Story is here: https://www.pivotaltracker.com/story/show/176683389

mdrohmann (approved):

Naatan (changes requested):
As noted in stand we should just use the native networking library to call this pixel, not our download package.

Also I think we can get away with just doing a HEAD request (should be faster).

Naatan (changes requested):
Let's run both `sendGAEvent` and `sendS3Pixel` in goroutines so they don't block one another.

Naatan (changes requested):
Why is this adding so many changes? The waitgroups seem pointless since it doesn't ever actually wait. I think this whole change could've just been adding `go ` in front of 2 calls?

MDrakos:
It's my understanding that any currently running goroutines will be killed when the main program completes https://groups.google.com/g/golang-nuts/c/jHhXrbb1jC8. The waitgroups ensure that we don't potentially lose analytic events. 

The other change was to have `sendEvent` no longer return an error as we can handle the logging there.

Naatan:
> It's my understanding that any currently running goroutines will be killed when the main program completes https://groups.google.com/g/golang-nuts/c/jHhXrbb1jC8. The waitgroups ensure that we don't potentially lose analytic events.
> 
> The other change was to have `sendEvent` no longer return an error as we can handle the logging there.

But that's not what this story is about. Analytics are already running in subroutines, that isn't new here. And as it stands it doesn't look like the problem of ensuring the analytics are sent is actually solved here. Good that you found that issue though, let's log a bug for it and address it in its own PR.

MDrakos:
>But that's not what this story is about. Analytics are already running in subroutines, that isn't new here. And as it stands it doesn't look like the problem of ensuring the analytics are sent is actually solved here. Good that you found that issue though, let's log a bug for it and address it in its own PR.

Right, I didn't check if the other analytics gorountines are being waited on. I've reverted my initial changes and simplified things. Here is the story to address events potentially being missed: https://www.pivotaltracker.com/story/show/176710543

mdrohmann:
@Naatan I think @MDrakos solution worked, as we have a function that waits for the `eventWaitGroup` to finish: https://github.com/ActiveState/cli/blob/master/internal%2Fanalytics%2Fanalytics.go#L101-L114

It is called in the `captain.Command.runner()` function.

But we could remove the go routines [here](https://github.com/ActiveState/cli/blob/master/internal%2Fanalytics%2Fanalytics.go#L177-L180) and [here](https://github.com/ActiveState/cli/blob/master/internal%2Fanalytics%2Fanalytics.go#L190-L193), as it is  superfluous to run the two go routines in another go routine.

mdrohmann:
I think if we do not wait for the go routines to finish anymore, we should remove the `WaitForAllEvents()` function completely, as it is not doing anything useful after https://github.com/ActiveState/cli/pull/1222/commits/5c74a983e28a265d7b3952726018cb15d43d551a

Naatan:
I see, thanks @mdrohmann

@MDrakos can you add back just the waitgroup stuff? Should be no more than 3 new lines by my count.

Naatan (approved):

PR #1225, opened 2021-01-27
send deferred analytics at least every 24 hours

mdrohmann:
https://www.pivotaltracker.com/story/show/176649521


internal/analytics/deferred.go
```diff
@@ -18,10 +26,53 @@ type deferredData struct {
 }
 
 const deferredCfgKey = "deferrer_analytics"
+const deferrerFileName = "deferrer"
+
+func deferrerTimeStamp(cfg Configurable) string {
+	return filepath.Join(cfg.ConfigPath(), deferrerFileName)
+}
```
MDrakos, line 33: This function is returning a file path, not a time stamp. Could it be modified to stat the file and return that timestamp? That or just change the name of the function.

internal/analytics/deferred.go
```diff
@@ -18,10 +26,53 @@ type deferredData struct {
...
+	if err != nil {
+		logging.Errorf("Could not determine State Tool executable: %v", err)
+		exe = "state"
+	}
+	cmd := exec.Command(exe, "--version")
+	cmd.SysProcAttr = osutils.SysProcAttrForNewProcessGroup()
+	cmd.Env = append(os.Environ(), fmt.Sprintf("%s=true", constants.DisableUpdates))
+	err = cmd.Start()
+	if err != nil {
+		return errs.Wrap(err, "Failed to run state --version in background")
```
MDrakos, line 57: Since we are do not return if we encounter an error calling `os.Executable()` we could include the value of `exe` in the error message here.

internal/analytics/deferred.go
```diff
@@ -18,10 +26,53 @@ type deferredData struct {
...
+
+func runNonDeferredStateToolCommand(cfg Configurable) error {
+	exe, err := os.Executable()
+	if err != nil {
+		logging.Errorf("Could not determine State Tool executable: %v", err)
+		exe = "state"
+	}
+	cmd := exec.Command(exe, "--version")
+	cmd.SysProcAttr = osutils.SysProcAttrForNewProcessGroup()
+	cmd.Env = append(os.Environ(), fmt.Sprintf("%s=true", constants.DisableUpdates))
```
MDrakos, line 54: Do we not need to route `stdout` and `stderr` here to avoid it printing the results of `state --version` to the user's terminal?

MDrakos (changes requested):
Looks good! Just a couple minor comments and a question.

internal/analytics/deferred.go
mdrohmann: Good point.  Changing the names.

internal/analytics/deferred.go
mdrohmann: They are set to nil by default, which should means `/dev/null`, `/dev/zero`...  But I think I will make that explicit, just to make sure...

internal/analytics/deferred.go
```diff
@@ -18,10 +26,57 @@ type deferredData struct {
...
+	cmd.Stdin = nil
+	cmd.Stdout = nil
```
MDrakos, line 56: `cmd.Stderr` as well?

internal/updater/updater_lin_mac_test.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 func TestUpdaterNoError(t *testing.T) {
+	t.SkipNow()
```
MDrakos, line 23: I imagine this is related to #1226? Can it be removed once that PR is merged?

MDrakos (approved):

MDrakos (approved):

Naatan (approved):

PR #1227, opened 2021-01-28
Update analytics error code reporting

MDrakos:


internal/errs/unwrap.go
```diff
@@ -13,5 +13,8 @@ func UnwrapExitCode(err error) int {
 		return eerr.ExitCode()
 	}
 
+	if err == nil {
```
mdrohmann, line 16: nit: maybe this should go to the top of the function?

mdrohmann (approved):

internal/errs/unwrap.go
MDrakos: Yeah I agree, feels a bit awkward there.

mdrohmann (approved):

Naatan (approved):

PR #1230, opened 2021-01-29
log wrapped error to rollbar in analytics.go

mdrohmann:
https://www.pivotaltracker.com/story/show/176706112

This does not fix the issue, but will report the wrapped error containing the actual problem to rollbar.  My plan is to unstart the bug-report first, and add a blocker that we'll have to wait for new (more detailed) error messages first.

### DIFF
--- a/internal/errs/errs.go
+++ b/internal/errs/errs.go
@@ -125,3 +125,19 @@ func Matches(err error, target interface{}) bool {
 	}
 	return false
 }
+
+func JoinWrappedErrors(err error) string {
+	errs := []string{err.Error()}
+	for {
+		uerr, ok := err.(interface{ Unwrap() error })
+		if !ok {
+			break
+		}
+		err = uerr.Unwrap()
+		if err == nil {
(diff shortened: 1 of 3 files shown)


MDrakos (approved):

internal/errs/errs.go
```diff
@@ -125,3 +125,19 @@ func Matches(err error, target interface{}) bool {
 	}
 	return false
 }
+
+func JoinWrappedErrors(err error) string {
```
Naatan, line 129: This functionality already exists: http://github.com/ActiveState/cli/blob/f054c636e6eab88e5fb7dd38a3d693aa8eda6222/internal/errs/errs.go#L75-L75

Naatan (changes requested):

internal/errs/errs.go
mdrohmann: look at that...

Naatan (approved):

PR #1232, opened 2021-01-29
Guard viper configuration with RWLock

mdrohmann:
https://www.pivotaltracker.com/story/show/176721598

### DIFF
--- a/internal/analytics/deferred.go
+++ b/internal/analytics/deferred.go
@@ -122,19 +122,18 @@ func sendDeferred(cfg Configurable, sender func(string, string, string, map[stri
 	if err != nil {
 		return errs.Wrap(err, "Could not load events on send")
 	}
-	if len(deferred) == 0 {
-		return nil
-	}
-	for n, event := range deferred {
-		if err := sender(event.Category, event.Action, event.Label, event.Dimensions); err != nil {
-			return errs.Wrap(err, "Could not send deferred event")
+	if len(deferred) > 0 {
+		for n, event := range deferred {
+			if err := sender(event.Category, event.Action, event.Label, event.Dimensions); err != nil {
+				return errs.Wrap(err, "Could not send deferred event")
+			}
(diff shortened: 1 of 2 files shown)


mdrohmann:
It is hard to tell, but I am fairly sure that if this fixed the problem.  I could re-produce it once more without these changes (took a lot of tries), but now it does not seem to happen anymore...

internal/analytics/deferred.go
```diff
@@ -122,6 +122,9 @@ func sendDeferred(cfg Configurable, sender func(string, string, string, map[stri
 	if err != nil {
 		return errs.Wrap(err, "Could not load events on send")
 	}
+	if len(deferred) == 0 {
```
daved, line 125: Suggestion: It might be worth, instead, using `if len(deferred) > 0` and placing it around the cfg.Save below so that the "deferrerFilePath" is cleaned up.

daved (dismissed):

daved (approved):

internal/analytics/deferred.go
```diff
@@ -122,17 +122,19 @@ func sendDeferred(cfg Configurable, sender func(string, string, string, map[stri
 	if err != nil {
 		return errs.Wrap(err, "Could not load events on send")
 	}
-	for n, event := range deferred {
-		if err := sender(event.Category, event.Action, event.Label, event.Dimensions); err != nil {
-			return errs.Wrap(err, "Could not send deferred event")
+	if len(deferred) > 0 {
```
Naatan, line 125: Nit: This is an unnecessary change.

Naatan (approved):

internal/analytics/deferred.go
mdrohmann: It skips the save of the configuration file, which I believe should be avoided if nothing changed.  Of course, it doesn't affect this bug, but as we have seen that viper also has issues with concurrent writes between processes...

PR #1237, opened 2021-02-03
Add analytics calls for runtime install logic

daved:
https://www.pivotaltracker.com/story/show/176748963


pkg/platform/runtime/installer.go
```diff
@@ -67,18 +81,29 @@ func NewInstaller(runtime *Runtime) *Installer {
...
+		analytics.EventWithLabel(catRuntime, actFailure, lblAssembler)
 		return nil, false, err
 	}
-	return installer.InstallArtifacts(assembler)
+
+	eg, fresh, err := installer.InstallArtifacts(assembler)
+	if err != nil {
+		analytics.EventWithLabel(catRuntime, actFailure, lblArtifacts)
+	}
+	analytics.Event(catRuntime, actSuccess)
```
Naatan, line 105: You're missing an `else` here.

Naatan (changes requested):

pkg/platform/runtime/installer.go
daved: ugh. embarrassing

Naatan (approved):

PR #1238, opened 2021-02-03
Move change dir command before setting environment variables

mdrohmann:
https://www.pivotaltracker.com/story/show/176754000

### DIFF
--- a/assets/shells/bashrc.sh
+++ b/assets/shells/bashrc.sh
@@ -6,6 +6,8 @@ if [ -z "$PROMPT_COMMAND" ]; then
 fi
 {{end}}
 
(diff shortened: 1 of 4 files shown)


mdrohmann:
It turns out that the problem here was very specific to my own `zsh` configuration:  The line `cd <workdirectory>` in the rc file parsed during activation triggered another source of the `~/.zshrc` file.  This PR fixes this issue, but it was a red herring when trying to figure out what @autarch experienced during his demo.  So, I created a follow-up story: https://www.pivotaltracker.com/story/show/176793096

MDrakos (approved):

Naatan (approved):

PR #1239, opened 2021-02-03
Do not parse verbose flags in shim args

mdrohmann:
https://www.pivotaltracker.com/story/show/176667499

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -207,6 +207,10 @@ func run(args []string, isInteractive bool, out output.Outputer) (int, error) {
 
 func argsHaveVerbose(args []string) bool {
 	for _, arg := range args {
+		// Skip looking for verbose args after --, eg. for `state shim -- perl -v`
+		if arg == "--" {
+			return false
+		}
 		if arg == "--verbose" || arg == "-v" {
 			return true
 		}


MDrakos (changes requested):
I believe cobra should be handling this for us here https://github.com/ActiveState/cli/blob/master/cmd/state/internal/cmdtree/shim.go#L43 but perhaps it's not being called any longer? I noticed that shim has a `--path` flag so disabling flag parsing may interfere with that. I'm not sure that `shim` logic should live in main. Can you take a look into this and see if `disableFlagParsing` is working as intended?

mdrohmann:
The issue here, is that the check for verbose arguments is done even before the cmdtree is initialized.  So, this parsing is completely separate from cobra:


Here is where the cmdtree is initialized:
https://github.com/ActiveState/cli/blob/master/cmd%2Fstate%2Fmain.go#L168
And the verbose check is on line 102 above that.

mdrohmann:
@Naatan The Windows shim also uses the double dash separator: https://github.com/ActiveState/cli/blob/master/assets%2Fshim%2Fshim.bat#L5

MDrakos (approved):
>The issue here, is that the check for verbose arguments is done even before the cmdtree is initialized. So, this parsing is completely separate from cobra

I see what you mean. I will defer to @Naatan on this one. I still feel that we should be handling the flags all in one place, but that should probably be it's own ticket.

Naatan (approved):

PR #1240, opened 2021-02-05
Add state cve report

mdrohmann:
https://www.pivotaltracker.com/story/show/176449977

### DIFF
--- a/cmd/state/internal/cmdtree/cve.go
+++ b/cmd/state/internal/cmdtree/cve.go
@@ -42,7 +42,7 @@ func newReportCommand(prime *primer.Values) *captain.Command {
 		[]*captain.Flag{},
 		[]*captain.Argument{
 			{
-				Name:        locale.Tl("cve_report_namespace_arg", "organization/project"),
+				Name:        locale.Tl("cve_report_namespace_arg", "Organization/Project"),
 				Description: locale.Tl("cve_report_namespace_arg_description", "The project for which the report is created"),
 				Value:       params.Namespace,
 			},
--- a/cmd/state/internal/cmdtree/cve.go
+++ b/cmd/state/internal/cmdtree/cve.go
@@ -5,6 +5,7 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runners/cve"
+	"github.com/ActiveState/cli/pkg/project"
 )
 
 func newCveCommand(prime *primer.Values) *captain.Command {
@@ -12,8 +13,8 @@ func newCveCommand(prime *primer.Values) *captain.Command {
 
 	cmd := captain.NewCommand(
 		"cve",
-		locale.Tl("cve_title", "CVE Summary"),
-		locale.Tl("cve_description", "Show a summary of CVE vulnerabilities"),
+		locale.Tl("cve_title", "Vulnerability Summary"),
+		locale.Tl("cve_description", "Show a summary of project vulnerabilities"),
 		prime.Output(),
 		prime.Config(),
 		[]*captain.Flag{},
@@ -28,12 +29,14 @@ func newCveCommand(prime *primer.Values) *captain.Command {
 
 func newReportCommand(prime *primer.Values) *captain.Command {
 	report := cve.NewReport(prime)
-	params := cve.ReportParams{}
+	params := cve.ReportParams{
+		Namespace: &project.Namespaced{},
+	}
 
 	return captain.NewCommand(
 		"report",
-		locale.Tl("cve_report_title", "Print a vulnerability report"),
-		locale.Tl("cve_report_cmd_description", "Print a vulnerability report"),
+		locale.Tl("cve_report_title", "Vulnerability Report"),
+		locale.Tl("cve_report_cmd_description", "Show a detailed report of project vulnerabilities"),
 		prime.Output(),
 		prime.Config(),
 		[]*captain.Flag{},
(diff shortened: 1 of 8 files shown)


cmd/state/internal/cmdtree/cve.go
```diff
@@ -25,3 +26,29 @@ func newCveCommand(prime *primer.Values) *captain.Command {
...
+	return captain.NewCommand(
+		"report",
+		locale.Tl("cve_report_title", "Vulnerability Report"),
+		locale.Tl("cve_report_cmd_description", "Show a detailed report of project vulnerabilities"),
+		prime.Output(),
+		prime.Config(),
+		[]*captain.Flag{},
+		[]*captain.Argument{
+			{
+				Name:        locale.Tl("cve_report_namespace_arg", "organization/project"),
```
MDrakos, line 45:
```suggestion
				Name:        locale.Tl("cve_report_namespace_arg", "Organization/Project"),
```

internal/runners/cve/cve.go
```diff
@@ -120,6 +120,14 @@ type SeverityCountOutput struct {
 	Severity string `locale:"severity,Severity"`
 }
 
+func (od *outputDataPrinter) printFooter() {
+	od.output.Print("")
+	od.output.Print([]string{
+		locale.Tl("cve_hint_report", "To view a detailed report for this runtime, run [ACTIONABLE]state cve report[/RESET]"),
+		locale.Tl("cve_hint_specific_report", "For a specific runtime, run [ACTIONABLE]state cve report [orgName/projectName][/RESET]"),
```
MDrakos, line 127:
```suggestion
		locale.Tl("cve_hint_specific_report", "For a specific runtime, run [ACTIONABLE]state cve report [Organization/Project][/RESET]"),
```

internal/runners/cve/report.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+			rd.output.Print(fmt.Sprintf("  %s %-10s [ACTIONABLE]%s[/RESET]", bar, severity, d.CveId))
+		}
+		rd.output.Print("")
+	}
+
+	rd.output.Print("")
+	rd.output.Print([]string{
+		locale.Tl("cve_report_hint_cve", "To view a specific CVE, run [ACTIONABLE]state cve open [cve-id][/RESET]."),
+	})
```
MDrakos, line 205: Wrapping this in a `[]string` will prepend a ` - ` to it. Is this the desired output here?

test/integration/cve_int_test.go
```diff
@@ -26,16 +26,38 @@ func (suite *CveIntegrationTestSuite) TestCveSummary() {
...
 
+func (suite *CveIntegrationTestSuite) TestCveReport() {
+	suite.OnlyRunForTags(tagsuite.Cve)
+
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	ts.LoginAsPersistentUser()
+
+	cp := ts.Spawn("cve", "report", "ActiveState/ActivePython-3.7")
```
MDrakos, line 47: This project isn't owned by us, and could be updated to patch these vulnerabilities. You may want to fork this project, or create your own so this test does not end up randomly breaking

MDrakos (approved):
Makes sense. My comments are just nits and suggestions.

internal/runners/cve/report.go
mdrohmann: Yes, in the mock these are also presented as a list.

cmd/state/internal/cmdtree/cve.go
```diff
@@ -25,3 +26,29 @@ func newCveCommand(prime *primer.Values) *captain.Command {
 	cmd.SetGroup(PlatformGroup)
 	return cmd
 }
+
+func newReportCommand(prime *primer.Values) *captain.Command {
+	report := cve.NewReport(prime)
+	params := cve.ReportParams{
+		Namespace: &project.Namespaced{},
```
Naatan, line 33: Why initialize an empty namespace? Shouldn't this be nil if it wasn't provided?

internal/runners/cve/report.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+			cves[ve.Version] = append(cves[ve.Version], ve)
+		}
+
+		for ver, vuls := range cves {
+			packageVulnerabilities = append(packageVulnerabilities, DetailedByPackageOutput{
+				v.Name, ver, vuls,
+			})
+		}
+	}
```
Naatan, line 107: Can we move this into the model?

internal/runners/cve/report.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+		// Remove this block with story https://www.pivotaltracker.com/story/show/176508772
+		// filter double entries
+		if _, ok := visited[v.Name]; ok {
+			continue
+		}
+		visited[v.Name] = struct{}{}
+
+		cves := make(map[string][]medmodel.Vulnerability)
```
Naatan, line 94: If we're going to do post-processing on these types at least alias the type.

Naatan (changes requested):

cmd/state/internal/cmdtree/cve.go
mdrohmann: This is how we did in [other places](https://github.com/ActiveState/cli/blob/master/cmd%2Fstate%2Finternal%2Fcmdtree%2Factivate.go#L19-L20) to set the namespace as an argument, by calling its `Set()` method as required by the `ArgMarshaler` interface.

Naatan (approved):

PR #1242, opened 2021-02-08
Reduce dependency on projectfile package

MDrakos:
https://www.pivotaltracker.com/story/show/176753605

Here is a link to the corresponding doc: https://docs.google.com/document/d/1vMKL94rigmbYfnikBqPtwZkmBudXy-jxnlsJYYnO89E/edit?usp=sharing. I used this to keep track of where I was but it also contains some notes/suggestions about the files still containing `projectfile` references.

I attempted to fix some of the integration tests, but there are quite a few failures that I do not believe are related to my changes. The critical integration tests and unit tests are passing.


cmd/state/forward.go
```diff
@@ -36,7 +36,7 @@ func forwardFn(bindir string, args []string, out output.Outputer, pj *project.Pr
 	}
 
 	// Retrieve the version info specified in the activestate.yaml
-	versionInfo, err := projectfile.ParseVersionInfo(pj.Source().Path())
+	versionInfo, err := project.ParseVersionInfo(pj.Source().Path())
```
Naatan, line 39: This one doesn't make sense. It takes the file of an activestate.yaml as the source, which is decidedly a `projectfile` responsibility.

cmd/state/main.go
```diff
@@ -128,11 +128,7 @@ func run(args []string, isInteractive bool, out output.Outputer) (int, error) {
 	// Set up project (if we have a valid path)
 	var pj *project.Project
 	if pjPath != "" {
-		pjf, err := projectfile.FromPath(pjPath)
-		if err != nil {
-			return 1, err
-		}
-		pj, err = project.New(pjf, out)
+		pj, err = project.FromPath(pjPath)
```
Naatan, line 131: Again, this one is receiving a path so should remain in `projectfile`.

pkg/project/project.go
```diff
@@ -252,6 +262,46 @@ func (p *Project) Namespace() *Namespaced {
...
+	return projectfile.GetProjectMapping(config)
+}
+
+func CleanProjectMapping(config ConfigGetter) {
+	projectfile.CleanProjectMapping(config)
+}
+
+func GetProjectNameForPath(config ConfigGetter, path string) string {
+	return projectfile.GetProjectNameForPath(config, path)
+}
```
Naatan, line 303: Sorry, I think this is just obfuscating the problem. In many cases the new equivalent of these would either be automated at a lower level or have a different API altogether. Please revert these.

Naatan (changes requested):
I'm sorry, this is moving in the wrong direction. What I wanted to get out of this story was reduce reliance on `projectfile` where there is a better equivalent in `project`. What this is currently doing is creating some aliases to `projectfile` and not much else. This is only going to make refactoring harder because now the connections are more detached. Think calls to `projectfile.Get()` and `project.Source()` (which returns the `projectfile` struct).

Let's close this story for now. I don't think it makes sense to invest more time on this for the next sprint.

PR #1243, opened 2021-02-08
Fasttrack detailed error reporting to release branch

mdrohmann:
We still needed to move this extra error to the release branch, in order to understand what is actually going on with the rollbar error that we receive.

- Should the  CHANGELOG also be updated?

 https://www.pivotaltracker.com/story/show/176706112


Naatan (approved):
This isn't user facing so no changelog update required.

mdrohmann:
> This isn't user facing so no changelog update required.

That's what I though.  I think you need to merge this PR though.

The tests finally pass, but I am still not authorized.

PR #1246, opened 2021-02-09
Add listing and creation of branches

MDrakos:
Original story: https://www.pivotaltracker.com/story/show/176449860

Current story: https://www.pivotaltracker.com/story/show/176755676

Note that the PR that allows us to create a branch from an old commit is not yet merged: https://github.com/ActiveState/TheHomeRepot/pull/5093


cmd/state/internal/cmdtree/branch.go
```diff
@@ -0,0 +1,49 @@
...
+		"add",
+		locale.Tl("add_title", "Adding branch"),
+		locale.Tl("add_description", "Add a branch to your project"),
+		prime.Output(),
+		prime.Config(),
+		[]*captain.Flag{},
+		[]*captain.Argument{
+			{
+				Name:        locale.Tl("branch_arg_name", "name"),
+				Description: locale.Tl("branch_arg_name_description", "Branch name to be created"),
```
Naatan, line 41:
```suggestion
				Description: locale.Tl("branch_arg_name_description", "Branch to be created"),
```

cmd/state/internal/cmdtree/branch.go
```diff
@@ -0,0 +1,49 @@
...
+				Value:       &params.Label,
+				Required:    true,
+			},
+		},
+		func(_ *captain.Command, _ []string) error {
+			return runner.Run(params)
+		}).SetGroup(PlatformGroup)
```
Naatan, line 48:
```suggestion
		})
```

Not applicable to sub-commands.

internal/runners/branch/add.go
```diff
@@ -0,0 +1,59 @@
...
+		project: prime.Project(),
+	}
+}
+
+func (a *Add) Run(params AddParams) error {
+	logging.Debug("ExecuteAdd")
+
+	project, err := model.FetchProjectByName(a.project.Owner(), a.project.Name())
+	if err != nil {
+		return locale.WrapError(err, "branch_list_proejct_err", "Could not get project details for project: {{.V0}}", a.project.Namespace().String())
```
Naatan, line 32: This should be reusing an existing error. We for sure are doing this elsewhere.

internal/runners/branch/add.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	branchID, err := model.AddBranch(project.ProjectID, params.Label)
+	if err != nil {
+		return locale.WrapError(err, "err_add_branch", "Could not add branch")
```
Naatan, line 37: We should try to assert here if this is because the branch label is already used. Also can you make sure that the error from the API bubbles up here?

internal/runners/branch/add.go
```diff
@@ -0,0 +1,59 @@
...
+		return locale.WrapError(err, "err_add_branch_no_default", "Could not retrieve branch information for: {{.V0}}", localBranch)
+	}
+
+	err = model.UpdateBranchTracking(*branchID, a.project.CommitUUID(), branch.BranchID, model.TrackingIgnore)
+	if err != nil {
+		logging.Debug("Unable to update tracking information, attempting to delete branch")
+		derr := model.DeleteBranch(*branchID)
+		if err != nil {
+			logging.Debug("Could not delete branch %s, got error: %v", params.Label, derr)
+		}
```
Naatan, line 52:
```suggestion
		derr := model.DeleteBranch(*branchID)
		if err != nil {
			logging.Debug("Could not delete branch %s, got error: %v", params.Label, derr)
		}
```

internal/runners/branch/add.go
```diff
@@ -0,0 +1,59 @@
...
+		return locale.WrapError(err, "err_add_branch", "Could not add branch")
+	}
+
+	localBranch := a.project.BranchName()
+	branch, err := model.BranchForProjectByName(project, localBranch)
+	if err != nil {
+		return locale.WrapError(err, "err_add_branch_no_default", "Could not retrieve branch information for: {{.V0}}", localBranch)
+	}
+
+	err = model.UpdateBranchTracking(*branchID, a.project.CommitUUID(), branch.BranchID, model.TrackingIgnore)
```
Naatan, line 46: `*branchID` wasn't nil checked.

internal/runners/branch/list.go
```diff
@@ -0,0 +1,45 @@
...
+		project: prime.Project(),
+	}
+}
+
+func (l *List) Run() error {
+	logging.Debug("ExecuteList")
+
+	project, err := model.FetchProjectByName(l.project.Owner(), l.project.Name())
+	if err != nil {
+		return locale.WrapError(err, "branch_list_project_err", "Could not get details for project: {{.V0}}", l.project.Namespace().String())
```
Naatan, line 34: Reuse error from elsewhere

internal/runners/branch/list.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	var branches []string
+	for _, branch := range project.Branches {
+		branches = append(branches, branch.Label)
```
Naatan, line 39: Should be able to add highlighting here?

test/integration/branch_int_test.go
```diff
@@ -0,0 +1,65 @@
...
+	defer ts.Close()
+
+	suite.PrepareActiveStateYAML(ts)
+
+	cp := ts.Spawn("branch")
+	expectations := []string{
+		"firstbranchchild",
+		"secondbranch",
+		"firstbranch",
+		"main",
```
Naatan, line 28: This seems badly sorted, can we sort by the date they were added or something? We should probably also show the parent/child structure, but I'll file a followup for that.

test/integration/branch_int_test.go
```diff
@@ -0,0 +1,65 @@
...
+	defer ts.Close()
+
+	username := ts.CreateNewUser()
+	namespace := fmt.Sprintf("%s/%s", username, "branch-test")
+
+	cp := ts.Spawn("fork", "ActiveState-CLI/Platforms", "--org", username, "--name", "platform-test")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("activate", namespace, "--path="+ts.Dirs.Work, "--output=json")
+	cp.ExpectExitCode(0)
```
Naatan, line 48: You can drop the fork and use a dedicated project simply by randomly generating the branch name. And I don't see why you need to activate?

Naatan (changes requested):

internal/runners/branch/add.go
MDrakos: Github won't let me commit this one as there are no changes

internal/runners/branch/add.go
MDrakos: Confirmed that the API error bubbles up here. The output is below and seems fairly clear. Do we want to assert the error here so we can add more context to the error? 

```
╔═══════════════╗
║ Adding branch ║
╚═══════════════╝

Something Went Wrong
────────────────────
 x Could not add branch
 x This project already contains a branch with that name

Need More Help?
───────────────
 • Run → `state add --help` for general help
 • Visit the Forum → https://community.activestate.com/c/state-tool/
```

internal/runners/branch/add.go
MDrakos: I opted to update the return type from `AddBranch`

test/integration/branch_int_test.go
MDrakos: There is no timestamp on the branch type, but I can sort it alphabetically by label

internal/runners/branch/list.go
MDrakos: Yes, there is no preference in the story here: https://www.pivotaltracker.com/story/show/176470500 as to what colour the highlight should be. I used `ACTIONABLE` but can change it to whatever is more appropriate.

internal/runners/branch/add.go
Naatan: Just want to remove the whole DeleteBranch logic. Reaching this code is unlikely and means something is broken, don't want to start maintaining code to deal with broken API's or something.

internal/runners/branch/list.go
Naatan: Let's use `NOTICE` and also mark it with an asterisk (`*`) in case the user disabled colors.

`ATIONABLE` is more for links, commands, etc. Stuff that's directly actionable.

test/integration/branch_int_test.go
Naatan: Can we at least sort the default branch on top?

internal/runners/branch/list.go
```diff
@@ -31,13 +34,22 @@ func (l *List) Run() error {
 
 	project, err := model.FetchProjectByName(l.project.Owner(), l.project.Name())
 	if err != nil {
-		return locale.WrapError(err, "branch_list_project_err", "Could not get details for project: {{.V0}}", l.project.Namespace().String())
+		return locale.WrapError(err, "err_fetch_project", l.project.Namespace().String())
```
Naatan, line 37:
```suggestion
		return locale.WrapError(err, "err_fetch_project", "", l.project.Namespace().String())
```

Third arg is the fallback string.

internal/runners/branch/list.go
```diff
@@ -31,13 +34,22 @@ func (l *List) Run() error {
...
 	}
 
+	localBranch := l.project.BranchName()
+
 	var branches []string
 	for _, branch := range project.Branches {
-		branches = append(branches, branch.Label)
+		branchName := branch.Label
+		if branchName == localBranch {
+			branchName = fmt.Sprintf("[ACTIONABLE]%s[/RESET]", branchName)
```
Naatan, line 46:
```suggestion
			branchName = fmt.Sprintf("[NOTICE]%s[/RESET] *", branchName)
```

test/integration/branch_int_test.go
```diff
@@ -38,25 +39,32 @@ func (suite *BranchIntegrationTestSuite) TestBranch_Add() {
...
 
-	cp := ts.Spawn("fork", "ActiveState-CLI/Platforms", "--org", username, "--name", "platform-test")
+	cp := ts.Spawn("pull")
+	cp.ExpectLongString("Your project in the activestate.yaml has been updated")
 	cp.ExpectExitCode(0)
 
-	cp = ts.Spawn("activate", namespace, "--path="+ts.Dirs.Work, "--output=json")
+	cp = ts.SpawnCmd("cat", "activestate.yaml")
 	cp.ExpectExitCode(0)
+	fmt.Println(cp.TrimmedSnapshot())
```
Naatan, line 53: Debug statement?

Naatan (changes requested):

internal/runners/branch/add.go
MDrakos: Makes sense. Do we have a story for `state branch delete`? If there is an issue with `state branch add` at this point a user may not be able to use the same branch label unless they can delete it. If there is no story I can create one.

test/integration/branch_int_test.go
MDrakos: Whoops

internal/runners/branch/add.go
Naatan: Created it just now: https://www.pivotaltracker.com/story/show/176926065

Naatan (approved):

PR #1247, opened 2021-02-09
Only include language fields with state init

MDrakos:
https://www.pivotaltracker.com/story/show/176871876

### DIFF
--- a/assets/activestate.yaml.tpl
+++ b/assets/activestate.yaml.tpl
@@ -1,5 +1,5 @@
 project: {{.Project}}
-{{if .IncludeLanguageField }}
+{{if and (ne .LanguageName "") (eq .CommitID "") }}
 languages: # Please run 'state push' to create your language runtime, once you do the language entry here will be removed
   - name: {{.LanguageName}}
     version: {{.LanguageVersion}}
--- a/assets/activestate.yaml.tpl
+++ b/assets/activestate.yaml.tpl
@@ -1,5 +1,5 @@
 project: {{.Project}}
-{{if ne .LanguageName ""}}
+{{if .IncludeLanguageField }}
 languages: # Please run 'state push' to create your language runtime, once you do the language entry here will be removed
   - name: {{.LanguageName}}
     version: {{.LanguageVersion}}
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,41 +1,41 @@
-# #!/bin/sh
-# gofiles=$(git diff --cached --name-only --diff-filter=ACM | grep '.go$' | grep -vE '^vendor/')
-# [ -z "$gofiles" ] && exit 0
+#!/bin/sh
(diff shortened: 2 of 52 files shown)


Naatan (changes requested):
This is a regression bug. Meaning something broke it that needs fixing. The current changeset suggests we're working around this fact rather than addressing it.

Not saying this isn't necessarily a good solution, just want to make sure we find out what caused the regression.

mdrohmann:
Poking around the code, it seems like this is introduced when we added specialized projectfile contents for different languages.  Then, the `CreateParams` needed a language argument, so the if clause in the `template` file did not work as expected anymore, because suddenly we provided a language argument for normal activations where an as.yaml needed to be created.

I think the solution is okay, but I would suggest changing the variable name in the `CreateParams` struct to something more descriptive like `projectNeedsPush`.  And maybe it is better to move the logic from the template file to the projectfile implementation ie., only set `LanguageName` [here](https://github.com/ActiveState/cli/blob/602eb4a234275648488c1ff065d915bfd24fcaa5/pkg%2Fprojectfile%2Fprojectfile.go#L1057) if necessary.

MDrakos:
@mdrohmann I came to the same conclusion after investigating the code. 

I don't think the `projectfile` package should have any knowledge of what `push` is and why it would need it. Using that name would tailor the field to our one current use case.

I agree it would be nice to have the logic live only in the `projectfile` Go code rather than in the template. I can add an if statement in the Go code and only update `data` when necessary if that helps with clarity, but we will still need the `if` statement in the template to retain the `language` field structure. At least I'm not sure of a way to remove the `if` from the template and have those fields excluded when `LanguageName` is nil.

mdrohmann (approved):
Okay.  That makes sense.

assets/activestate.yaml.tpl
```diff
@@ -1,5 +1,5 @@
 project: {{.Project}}
-{{if ne .LanguageName ""}}
+{{if .IncludeLanguageField }}
```
Naatan, line 2:
```suggestion
{{if ne .Language "" }}
```

I don't think we need a special field for this.

Naatan (changes requested):

assets/activestate.yaml.tpl
MDrakos: I think we do because in both cases (`activate` and `init`) that language field is not the empty string. We have to pass it for `activate` so the `activationMessage` script has a language and `init` for the `push` case. 

We could revert the if statement in the template to what you have here and then only populate the language field in the Go code if `IncludeLanguageField` is true. I'm not sure if it's more clear to have two conditionals rather than the one we have now.

Naatan: I'd prefer for the calling logic not to have to worry about this detail. I think we can just make the conditional `{{if and (ne .Language "") (eq .CommitID "") }}`

MDrakos: Makes sense. The conditional works with a minor change to the project file creation code.

Naatan (approved):

PR #1250, opened 2021-02-10
Fix Komodo integration test

MDrakos:
This was introduced when we updated the `state secrets` UX.  In order to use the vertical table output we have to wrap the output data in a struct to include the `veritcalTable` option. Also the `HasValue` field is not compatible with what Komodo expects. This PR updates the output to do some processing before printing in the `editor.V0` case.

 https://www.pivotaltracker.com/story/show/176807110


daved (approved):

Naatan (approved):

MDrakos:
It appears that another Komodo integration test is not failing consistently. Created a ticket here: https://www.pivotaltracker.com/story/show/176930769

PR #1251, opened 2021-02-10
Don't show forward error if the forward itself didn't fail

Naatan:


cmd/state/errors.go
```diff
@@ -20,6 +20,8 @@ type ErrorTips interface {
 	ErrorTips() []string
 }
 
+type SilencedError struct{ error }
```
daved, line 23: We already have a "silent/silence" error mechanism:
https://github.com/ActiveState/cli/blob/b189da4be3c077090572288c6e31639f48a1d298/cmd/state/errors.go#L131-L136
with an implementation here:
https://github.com/ActiveState/cli/blob/b189da4be3c077090572288c6e31639f48a1d298/internal/subshell/sscommon/sscommon.go#L16-L27

If the need for this story requires a different behavior, it might be worth naming the type added in this PR something like IgnoreError. Also, embedding an error like this breaks the chain/unwrapping so that detection/inspection is not possible.

daved (changes requested):

cmd/state/errors.go
Naatan: Nop, I had just forgotten about that. Thanks!

daved (approved):
The amount of logic and structure expressed in the main package regarding forwarding makes me think it is reasonable to create a package just for that need: https://www.pivotaltracker.com/story/show/176911571

PR #1254, opened 2021-02-12
Add state branch switch

MDrakos:
https://www.pivotaltracker.com/story/show/176649650

### DIFF
--- a/internal/runbits/refresh.go
+++ b/internal/runbits/refresh.go
@@ -1,0 +1,40 @@
+package runbits
+
+import (
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/pkg/platform/runtime"
+	"github.com/ActiveState/cli/pkg/project"
+	"github.com/go-openapi/strfmt"
+)
+
+// RefreshRuntime should be called after runtime mutations. A nil arg for "req"
+// means that the message handler will not print output for "a single
+// requirement". For example, if multiple requirements are affected, nil is the
+// appropriate value.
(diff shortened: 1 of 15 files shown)


internal/runners/branch/switch.go
```diff
@@ -0,0 +1,65 @@
...
+		config:  prime.Config(),
+	}
+}
+
+func (s *Switch) Run(params SwitchParams) error {
+	logging.Debug("ExecuteSwitch")
+
+	project, err := model.FetchProjectByName(s.project.Owner(), s.project.Name())
+	if err != nil {
+		return locale.WrapError(err, "err_fetch_project", s.project.Namespace().String())
```
Naatan, line 39: Third argument is fallback. Pass empty string if you don't want to use it. (This goes for any such changes in this PR).

internal/runbits/refresh.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/ActiveState/cli/pkg/platform/runtime"
+	"github.com/ActiveState/cli/pkg/project"
+	"github.com/go-openapi/strfmt"
+)
+
+// RefreshRuntime should be called after runtime mutations. A nil arg for "req"
+// means that the message handler will not print output for "a single
+// requirement". For example, if multiple requirements are affected, nil is the
+// appropriate value.
+func RefreshRuntime(out output.Outputer, req *RequestedRequirement, proj *project.Project, cachePath string, commitID strfmt.UUID, changed bool) error {
```
Naatan, line 15: Feels a little awkward here, filed this followup: https://www.pivotaltracker.com/story/show/176930119

Naatan (changes requested):

internal/runners/branch/switch.go
MDrakos: Right, I keep forgetting that. I think I've got them all

Naatan (approved):

PR #1256, opened 2021-02-16
Drop shim e2e tests (system python and doubledash)

daved:
https://www.pivotaltracker.com/story/show/176807063


daved:
@mdrohmann I'm not entirely sure about the changes I made, the AC is not clear about the desired outcome. The previous behavior seems to be restored by this change, but may not be desirable. Any guidance you can offer is appreciated.

mdrohmann (approved):
This makes sense to me.

mdrohmann (approved):

Naatan (approved):

PR #1258, opened 2021-02-16
use context.WithTimeout for fetchInfo timeouts

mdrohmann:
https://www.pivotaltracker.com/story/show/176812654

### DIFF
--- a/cmd/state/forward.go
+++ b/cmd/state/forward.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os/exec"
 	"path/filepath"
@@ -128,7 +129,7 @@ func ensureForwardExists(binary string, versionInfo *projectfile.VersionInfo, ou
 		DesiredVersion: desiredVersion,
 	}
 
-	info, err := up.Info()
+	info, err := up.Info(context.Background())
 	if err != nil {
 		return errs.Wrap(err, "Info failed")
 	}
--- a/internal/runners/update/update.go
+++ b/internal/runners/update/update.go
@@ -1,6 +1,7 @@
 package update
 
 import (
+	"context"
 	"os"
 
 	"github.com/ActiveState/cli/internal/constants"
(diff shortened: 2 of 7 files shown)


cmd/state/forward.go
```diff
@@ -128,7 +129,7 @@ func ensureForwardExists(binary string, versionInfo *projectfile.VersionInfo, ou
 		DesiredVersion: desiredVersion,
 	}
 
-	info, err := up.Info()
+	info, err := up.Info(context.Background())
```
daved, line 132: Not for this story, but... We have a greater issue; We need to wrangle our dependency on an HTTP client. In order to get the benefit of connection reuse, as well as putting a backstop timeout at the transport level, we should be passing around a common client that can have more aggressive per-request timeouts applied via context.Context as you've done here. Anyway, this is all to say that it feels wrong to me that we are making any HTTP calls without some timeout given. Again, for another story.

daved (approved):
Great work eliminating LOC!

cmd/state/forward.go
mdrohmann: It would be kind of nice if we had more timeouts.  On the other hand, it is (still) a console application, so the user can usually interrupt with Ctrl-C, unless we are in an automated setting, of course.  I agree that we should discuss that more...

As this is a bug story, I just made sure that the existing behavior is preserved ie., a very short timeout for the auto-update check.

daved (approved):

Naatan (approved):
For future reference please avoid big changesets like this for bugs. Bugs means fixes to our code, so not refactoring of code or introducing a lot of new code (at least not without discussion).

mdrohmann:
> For future reference please avoid big changesets like this for bugs. Bugs means fixes to our code, so not refactoring of code or introducing a lot of new code (at least not without discussion).

When starting with this I thought that this change would be still be okay, as it just gets rid of a complicated function which has a bug.  But I see your point, as it turned out to be several changes.  I will be more cautious which such a change for a bug in the future, and start a discussion first.

PR #1259, opened 2021-02-17
Notify log out/in rather than force logout

daved:
https://www.pivotaltracker.com/story/show/176928819

### DIFF
--- a/pkg/platform/authentication/auth.go
+++ b/pkg/platform/authentication/auth.go
@@ -145,8 +145,8 @@ func (s *Auth) AuthenticateWithModel(credentials *mono_models.Credentials) error
 	if err != nil {
 		tips := []string{
 			locale.T("relog_tip", "If you're seemingly logged in already, try logging out and in again."),
-			locale.T("logout_tip", "Logout with[ACTIONABLE]`state auth logout`[/RESET]."),
-			locale.T("logout_tip", "Login with[ACTIONABLE]`state auth`[/RESET]."),
+			locale.T("logout_tip", "Logout with [ACTIONABLE]`state auth logout`[/RESET]."),
+			locale.T("logout_tip", "Login with [ACTIONABLE]`state auth`[/RESET]."),
 		}
 
 		switch err.(type) {
--- a/pkg/platform/authentication/auth.go
+++ b/pkg/platform/authentication/auth.go
@@ -143,20 +143,20 @@ func (s *Auth) AuthenticateWithModel(credentials *mono_models.Credentials) error
 
 	loginOK, err := mono.Get().Authentication.PostLogin(params)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Naatan (commented):
I like the idea of adding a tip here, but the tips themselves are meant to be suuuper short and to the point, eg. "Logout with `state auth logout`".

daved:
@Naatan This will produce an effect only in some cases since we generally reach this code using `authentication.Get()`. In other words, the error w/tip will not propagate in most cases. I'm not sure how else to resolve this without touch many file.

Naatan:
> @Naatan This will produce an effect only in some cases since we generally reach this code using `authentication.Get()`. In other words, the error w/tip will not propagate in most cases. I'm not sure how else to resolve this without touch many file.

That's fine. Removing `auth.Get()` is its own story.

pkg/platform/authentication/auth.go
```diff
@@ -140,18 +140,23 @@ func (s *Auth) Authenticate() error {
...
 	params.SetCredentials(credentials)
-	loginOK, err := mono.Get().Authentication.PostLogin(params)
 
+	loginOK, err := mono.Get().Authentication.PostLogin(params)
 	if err != nil {
-		s.Logout()
+		tips := []string{
+			locale.T("relog_tip", "If you're seemingly logged in already, try logging out and in again."),
+			locale.T("logout_tip", "Logout with[ACTIONABLE]`state auth logout`[/RESET]."),
+			locale.T("logout_tip", "Login with[ACTIONABLE]`state auth`[/RESET]."),
```
Naatan, line 149:
```suggestion
			locale.T("logout_tip", "Logout with [ACTIONABLE]`state auth logout`[/RESET]."),
			locale.T("logout_tip", "Login with [ACTIONABLE]`state auth`[/RESET]."),
```

pkg/platform/authentication/auth.go
```diff
@@ -140,18 +140,23 @@ func (s *Auth) Authenticate() error {
 func (s *Auth) AuthenticateWithModel(credentials *mono_models.Credentials) error {
 	params := authentication.NewPostLoginParams()
 	params.SetCredentials(credentials)
-	loginOK, err := mono.Get().Authentication.PostLogin(params)
 
+	loginOK, err := mono.Get().Authentication.PostLogin(params)
 	if err != nil {
-		s.Logout()
+		tips := []string{
+			locale.T("relog_tip", "If you're seemingly logged in already, try logging out and in again."),
```
Naatan, line 147:
```suggestion
			locale.T("relog_tip", "If you're having trouble authenticating try logging out and logging back in again."),
```

Naatan (changes requested):

Naatan (approved):

PR #1261, opened 2021-02-18
Signup password mismatch permits retries, and related error output is improved

daved:
https://www.pivotaltracker.com/story/show/176705897

### DIFF
--- a/pkg/cmdlets/auth/signup.go
+++ b/pkg/cmdlets/auth/signup.go
@@ -22,6 +22,10 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 )
 
+const (
+	maxMatchTries = 3
+)
+
 type signupInput struct {
 	Name      string
 	Email     string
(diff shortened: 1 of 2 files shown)


pkg/cmdlets/auth/signup.go
```diff
@@ -156,22 +161,23 @@ func promptForSignup(input *signupInput, out output.Outputer, prompter prompt.Pr
...
-	err = passwordValidator(input.Password2)
-	if err != nil {
-		return errs.Wrap(err, "InvalidPassword failure")
+		if input.Password2 == input.Password {
+			break
+		}
+
+		locErr := locale.NewError("err_password_confirmation_failed")
+		if i < matchTries-1 {
+			out.Notice(locErr.UserError())
```
Naatan, line 177:
```suggestion
		if i < matchTries-1 {
			out.Notice(locale.T("err_password_confirmation_failed"))
```

Naatan (approved):
Looks good, just one nit.

pkg/cmdlets/auth/signup.go
daved: Yeah, I had wondered if you'd be OK with usage of UserError. It feels right to avoid localizing twice, but it's not common in our code. I opted to store the localization id as a var to use in both places.

Naatan (approved):

PR #1263, opened 2021-02-19
Replace Viper

MDrakos:


internal/config/instance.go
```diff
@@ -216,39 +242,74 @@ func (i *Instance) InstallSource() string {
...
 	}
+	i.data = configData
 	return nil
 }
 
-// Save saves the config file
-func (i *Instance) Save() error {
-	if i.noSave {
-		return nil
+func (i *Instance) configFileData() (map[string]interface{}, error) {
```
Naatan, line 253: Can we name this `readInConfigFile` or something that makes it explicit that this entails IO operations.

internal/config/instance.go
```diff
@@ -113,65 +118,86 @@ func Get() (*Instance, error) {
...
+		}
+		if attempts--; attempts > 0 {
+			time.Sleep(interval)
+			return retryLock(pl, attempts, interval)
+		}
+		return err
+	}
+
+	return nil
+}
```
Naatan, line 156: I think this should be part of our pidlock mechanic. Maybe a `WaitForLock(timeout)` function.

internal/config/instance.go
```diff
@@ -113,65 +118,86 @@ func Get() (*Instance, error) {
...
+
+func (i *Instance) get(key string) interface{} {
 	i.rwLock.RLock()
 	defer i.rwLock.RUnlock()
```
Naatan, line 160: We can drop this now.

internal/config/instance.go
```diff
@@ -113,65 +118,86 @@ func Get() (*Instance, error) {
...
-	return i.viper.GetString(key)
+	return i.data[strings.ToLower(key)]
```
Naatan, line 161: This is going to panic if the key doesn't exist, no?

internal/config/instance.go
```diff
@@ -216,39 +242,74 @@ func (i *Instance) InstallSource() string {
...
-	// Prepare viper, which is a library that automates configuration
-	// management between files, env vars and the CLI
-	i.viper.SetConfigName(i.Name())
-	i.viper.SetConfigType(i.Type())
-	i.viper.AddConfigPath(i.configDir.Path)
-	i.viper.AddConfigPath(".")
-
-	if err := i.viper.ReadInConfig(); err != nil {
-		return errs.Wrap(err, "Cannot read config.")
+	configData, err := i.configFileData()
```
Naatan, line 245: Why do we have a separate `configFileData()` function? Any reason why we'd be reading the config and not populating `i.data`? I can see in the `Set()` function you are updating it right away, but it doesn't cost much to populate a variable twice in a row.

internal/config/instance.go
```diff
@@ -316,6 +377,22 @@ func (i *Instance) ensureCacheExists() error {
 	return nil
 }
 
+func (i *Instance) getLockFile() string {
+	if i.lockFile == "" {
+		i.lockFile = filepath.Join(i.configDir.Path, "config-lock")
```
Naatan, line 382:
```suggestion
		i.lockFile = filepath.Join(i.configDir.Path, "config.lock")
```

Naatan (changes requested):

internal/config/instance.go
MDrakos: I think we still need this, otherwise we could encounter a panic attempting to read from a map that is currently being written to

internal/config/instance.go
MDrakos: No, this will just return nil

internal/config/instance.go
MDrakos: You're right this function isn't necessary.

internal/config/instance.go
Naatan: How? The only way that that could happen is if a subroutine is writing to i.data, in which case it would be more appropriate to use a mutex. But I think that mutex probably makes more sense when writing than when reading.

internal/osutils/lockfile/pidlock.go
```diff
@@ -94,6 +96,28 @@ func (pl *PidLock) Close(keepFile ...bool) error {
...
+		if err != nil {
+			if !errors.As(err, &lockedErr) {
+				return errs.Wrap(err, "Unexpected error attempting to acquire lock file")
+			}
+
+			select {
+			case <-timer.C:
+				return err
+			default:
+				time.Sleep(1 * time.Second)
```
Naatan, line 114: A lock should generally clear in milliseconds. This is wasting an entire second waiting to try again. Let's run this every 100ms.

internal/osutils/lockfile/pidlock.go
```diff
@@ -94,6 +96,28 @@ func (pl *PidLock) Close(keepFile ...bool) error {
...
 
+// WaitForLock will attempt to acquire the lock for the duration given
+func (pl *PidLock) WaitForLock(timeout time.Duration) error {
+	lockedErr := &AlreadyLockedError{}
+	timer := time.NewTimer(timeout)
+	for {
+		_, err := pl.TryLock()
+		if err != nil {
+			if !errors.As(err, &lockedErr) {
+				return errs.Wrap(err, "Unexpected error attempting to acquire lock file")
```
Naatan, line 107: I think we should localize this and say something like "Could not write config as another process appears to be using it". Least that will give some indication to the user as to what might be the problem.

internal/osutils/lockfile/pidlock.go
```diff
@@ -94,6 +96,28 @@ func (pl *PidLock) Close(keepFile ...bool) error {
...
 }
 
+// WaitForLock will attempt to acquire the lock for the duration given
+func (pl *PidLock) WaitForLock(timeout time.Duration) error {
+	lockedErr := &AlreadyLockedError{}
+	timer := time.NewTimer(timeout)
+	for {
+		_, err := pl.TryLock()
+		if err != nil {
+			if !errors.As(err, &lockedErr) {
```
Naatan, line 106:
```suggestion
			if !errors.As(err, &lockedErr) {
```

Use `errs.Matches(err, &AlreadyLockedError{})`.

Naatan (changes requested):

internal/osutils/lockfile/pidlock.go
MDrakos: I updated the error message here and also in the config package as I didn't want to mention the config file in this package.

internal/config/instance.go
MDrakos: The scenario you describe is also the one I'm thinking of. It is unlikely to happen in the same instance of the state tool but still possible I think. The `rwLock` is a mutex which is used in both `Set` and `Get`

internal/osutils/lockfile/pidlock.go
```diff
@@ -94,6 +95,28 @@ func (pl *PidLock) Close(keepFile ...bool) error {
...
+
+			select {
+			case <-timer.C:
+				return err
+			default:
+				time.Sleep(100 * time.Millisecond)
+			}
+		}
+		return nil
+	}
```
daved, line 117: This is a nice design, but I think the channel can be dropped for a simple time.Time.After call: https://play.golang.org/p/AZ9VI4e4HHv

internal/config/instance.go
daved: A small aside: In almost all cases a RWLock costs more than the optimization it provides. In other words, it requires quite a bit of contention before it's worth using.

internal/config/instance.go
```diff
@@ -113,65 +117,68 @@ func Get() (*Instance, error) {
...
-	i.viper.SetDefault(key, value)
+	return cast.ToStringMapStringSlice(i.get(key))
 }
 
 // GetBool retrieves a boolean value for a given key
 func (i *Instance) GetBool(key string) bool {
-	i.rwLock.RLock()
-	defer i.rwLock.RUnlock()
-	return i.viper.GetBool(key)
+	return cast.ToBool(i.get(key))
```
daved, line 166: This isn't on you, @MDrakos (I think), but not returning an error from these functions seems haphazard. Since this is being worked on now, I suggest that it should be added. If not now, then a story filed. Calling code can continue ignoring the error for a while, with another story (or multiple stories) to handle upgrading those call sites.

internal/config/instance.go
```diff
@@ -113,65 +117,68 @@ func Get() (*Instance, error) {
...
-	return i.viper.GetStringSlice(key)
+	return cast.ToStringSlice(i.get(key))
 }
 
 // GetTime retrieves a time instance for a given key
 func (i *Instance) GetTime(key string) time.Time {
-	i.rwLock.RLock()
-	defer i.rwLock.RUnlock()
-	return i.viper.GetTime(key)
+	return cast.ToTime(i.get(key))
```
daved, line 176: Another aside (may be worth thinking about now): The API surface can be reduced while increasing capability using this form: https://play.golang.org/p/sCtajzdwwTc

Many types can be added to the switch so that any reasonable values are handled, which is the compromise: On one hand, the methods can reveal what types are supported, but adding types adds noise to the API. On the other hand, many types can be added without any extra noise, but it is not fully clear what types are supported.

If this is interesting, but not for this story, please add a story or let me know and I can do so.

daved (changes requested):

internal/config/instance.go
```diff
@@ -231,6 +231,10 @@ func (i *Instance) ReadInConfig() error {
 
 	err = pl.WaitForLock(5 * time.Second)
 	if err != nil {
+		lockedErr := &lockfile.AlreadyLockedError{}
+		if errs.Matches(err, &lockedErr) {
```
Naatan, line 235:
```suggestion
		if errs.Matches(err, &lockfile.AlreadyLockedError{}) {
```

The whole point of `errs.Matches` is that you don't need to do this silly trickery. If the above doesn't work you're probably not constructing the error properly.

internal/config/instance.go
```diff
@@ -231,6 +231,10 @@ func (i *Instance) ReadInConfig() error {
...
+			return errs.Wrap(err, "Could not write config as another process appears to be using it")
```
Naatan, line 236: Why did you just detect the error type just to wrap it with a regular error? Did you intend to localize this?

internal/osutils/lockfile/pidlock.go
```diff
@@ -103,15 +102,15 @@ func (pl *PidLock) WaitForLock(timeout time.Duration) error {
 	for {
 		_, err := pl.TryLock()
 		if err != nil {
-			if !errors.As(err, &lockedErr) {
-				return errs.Wrap(err, "Unexpected error attempting to acquire lock file")
+			if !errs.Matches(err, &lockedErr) {
```
Naatan, line 105: Same here, drop the extra reference.

Naatan (changes requested):

internal/config/instance.go
MDrakos: No, importing `locale` in this package causes an import cycle. The intent here was to add some more context to the error. Do these wrapped messages not bubble up to the user in the same way as the localized errors? If so, I can remove this if it won't be helpful.

daved: At a glance, I think it makes sense to not return localized errors from this package, and then have handlers provide localized errors.

If that's not acceptable, I suggest that locale is doing too much by affecting persistent application state.

internal/config/instance.go
MDrakos: I agree that ignoring the error here could cause issues. One of the ACs of this story is that the config API remains unchanged. I've already violated this AC on the `Set` calls but that was after discussion with Nathan and concluding that we are not using `Set` in many places. Updating the get functions would be a larger undertaking that I don't think should be in the scope of this story but something worth visiting in the future.

internal/config/instance.go
MDrakos: This would also be another fairly drastic change to the config API. I'm not against it but do not think it's appropriate for this story.

MDrakos: Story here: https://www.pivotaltracker.com/story/show/177194225. This also covers your error handling proposal.

internal/osutils/lockfile/pidlock.go
```diff
@@ -94,6 +95,27 @@ func (pl *PidLock) Close(keepFile ...bool) error {
...
+			if !errs.Matches(err, &AlreadyLockedError{}) {
+				return errs.Wrap(err, "Could not acquire lock")
+			}
+
+			select {
+			case <-timer.C:
+				return err
+			default:
+				time.Sleep(100 * time.Millisecond)
+			}
```
daved, line 113: This should work fine - just noting that you had thumbed-up my suggestion to use time.Time.After instead.

daved (approved):

internal/osutils/lockfile/pidlock.go
daved: Oh, I missed this... Using a timer requires more detailed usage: https://github.com/golang/go/issues/27169#issuecomment-415393233

I do think avoiding the channel by using time.Time.After is worthwhile here.

MDrakos: Whoops, forgot to push the commit where I use `time.Now().After()`

internal/config/instance.go
Naatan: Non localized errors do not bubble up to the user. And the error chain is its own context. I think we can drop this.

Naatan (approved):

Naatan (approved):

PR #1264, opened 2021-02-19
Implement EnvironProvider for alternative builds

mdrohmann:
https://www.pivotaltracker.com/story/show/176968032


mdrohmann:
@Naatan @MDrakos I merged all my PRs in the runtime-refactor branch.

I hope that you find the code in a helpful state.  One thing that I forgot to mention is that I added a `testhelper` directory with some simple functions to load JSON files that contain actual responses from our platform.  The JSON files have been all downloaded programmatically by `go run ./pkg/platform/runtime2/testhelper/cmd`.
You can see how I used them in the `artifact_test.go` function.  I like this approach as then we can unit-test with "real" data, and that already turned out to be helpful for the tests in the `artifact.go` file, but it might also be helpful in the future as it provides a simple way to add regression tests.

See you next sprint!

PR #1273, opened 2021-03-03
Update branch listing

MDrakos:
https://www.pivotaltracker.com/story/show/176891945


cmd/state/internal/cmdtree/branch.go
```diff
@@ -12,7 +12,7 @@ func newBranchCommand(prime *primer.Values) *captain.Command {
 
 	return captain.NewCommand(
 		"branch",
-		locale.Tl("branch_title", "Listing branches"),
+		locale.Tl("branch_title", "Project Branches"),
```
Naatan, line 15: Let's undo this? I don't think adding a Project suffix here is helpful, and it may in some cases be plain inaccurate (headless state).

internal/runners/branch/list.go
```diff
@@ -33,35 +30,24 @@ func NewList(prime primeable) *List {
 func (l *List) Run() error {
 	logging.Debug("ExecuteList")
 
+	l.out.Print(
+		locale.Tl(
+			"branch_info",
+			"\nBranches allow you to create runtimes with different packages sets depending on your use case. Here are the branches in your current project.\n",
+		),
+	)
+
```
Naatan, line 39: Was this in the mock? Let's drop this. NONE of our other commands do this.

internal/runners/branch/list.go
```diff
@@ -33,35 +30,24 @@ func NewList(prime primeable) *List {
...
-	}
-	branches = append([]string{mainBranchLabel}, branches...)
-
-	l.out.Print(branches)
+	tree := NewBranchTree()
+	tree.AddLocalBranch(l.project.BranchName())
+	tree.AddBranchFormatting("[NOTICE]%s[/RESET]")
+	tree.AddLocalBranchFormatting("[ACTIONABLE]%s[/RESET] [DISABLED](Current)[/RESET]")
+	tree.BuildFromBranches(project.Branches)
+	l.out.Print(tree.String())
```
Naatan, line 50: This doesn't seem very JSON friendly.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+	bt.branchFormatting = formatting
+}
+
+func (bt *BranchTree) AddLocalBranch(branch string) {
+	bt.localBranch = branch
+}
+
+func (bt *BranchTree) AddLocalBranchFormatting(formatting string) {
+	bt.localBranchFormatting = formatting
+}
```
Naatan, line 112: You're calling these `Add*` but in effect they're just `Set`-ing. `Add` implies it's cumulative.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+	rootBranches          mono_models.Branches
+	localBranch           string
+	branchFormatting      string
+	localBranchFormatting string
+}
+
+const (
+	edgeLink string = "│  "
+	edgeMid  string = "├─"
+	edgeEnd  string = "└─"
```
Naatan, line 32: It's kinda strange that you have functions to set formatting but then this part is hardcoded. Either hardcore or don't (I'd say for our use case we should just hard code as this will be thrown away when we revisit formatting).

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/ActiveState/cli/pkg/platform/api/mono/mono_models"
+	"github.com/go-openapi/strfmt"
+)
+
+type branchNode struct {
+	Label string
+	ID    string
+}
+
+type tree map[branchNode]tree
```
Naatan, line 18: @daved is this "proper" Go? Feels really weird how it's self referencing.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (bt *BranchTree) AddLocalBranch(branch string) {
+	bt.localBranch = branch
+}
+
+func (bt *BranchTree) AddLocalBranchFormatting(formatting string) {
+	bt.localBranchFormatting = formatting
+}
+
+func (bt *BranchTree) String() string {
```
Naatan, line 114: It's a bit awkward how this basically won't work in mysterious ways unless you call `Build`.

Suggest renaming this to `Marshal()` and combining the build stuff into it.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+func NewBranchTree() *BranchTree {
+	return &BranchTree{tree: make(tree)}
+}
+
+func (bt *BranchTree) BuildFromBranches(branches mono_models.Branches) {
+	bt.rootBranches = getRootBranches(branches)
+	for _, branch := range bt.rootBranches {
+		bt.tree[branchNode{branch.Label, branch.BranchID.String()}] = buildBranchTree(branch, branches)
```
Naatan, line 42: I don't think you need to be storing `rootBranches` and `tree`, just pass them around instead. With the current implementation the state of your `struct` may at times be practically corrupted unless you just "know" which method to call next. It's best if the struct itself is never in a state where it will implicitly do something unexpected.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+}
+
+func buildBranchTree(currentBranch *mono_models.Branch, branches mono_models.Branches) tree {
+	t := getChildren(currentBranch, branches)
```
Naatan, line 47: As a rule of thumb I'd suggest naming your vars unless you're only using it within a ~2-3 line span.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+	for _, branch := range branches {
+		// Discard any branches without tracking information as we are only interested
+		// in child branches of the current branch
+		if branch.Tracks == nil {
+			continue
+		}
```
Naatan, line 53: This seems like unnecessary busywork, won't the next conditional already filter these out anyway?

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+
+		// Check that this branch is a child branch and recursively build its tree
+		if _, ok := t[branchNode{branch.Label, branch.BranchID.String()}]; ok {
+			t[branchNode{branch.Label, branch.BranchID.String()}] = buildBranchTree(branch, branches)
```
Naatan, line 57: You're constructing `branchNode` twice here.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+	for _, b := range branches {
+		if b.Tracks == nil {
+			continue
+		}
+		if b.Tracks.String() == branch.BranchID.String() {
+			children[branchNode{b.Label, b.BranchID.String()}] = make(tree)
+		}
+	}
+	return children
+}
```
Naatan, line 100: These should all go into the models package instead.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+func containsBranch(id *strfmt.UUID, branches mono_models.Branches) bool {
+	for _, branch := range branches {
+		if branch.BranchID.String() == id.String() {
+			return true
+		}
+	}
+	return false
+}
+
+func getChildren(branch *mono_models.Branch, branches mono_models.Branches) tree {
```
Naatan, line 85: I think we can move this to the model and have it return a model Branch instead of a tree. Keep the tree transforms local.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (bt *BranchTree) AddLocalBranchFormatting(formatting string) {
+	bt.localBranchFormatting = formatting
+}
+
+func (bt *BranchTree) String() string {
+	w := new(bytes.Buffer)
+	var levelsCompleted []int
+	bt.print(w, bt.tree, 0, levelsCompleted)
+	return w.String()
```
Naatan, line 118: Why is this struct concerning itself with printing? It should just be returning data. Passing a writer only makes sense if your output is buffered in some way.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (bt *BranchTree) print(w io.Writer, currentTree tree, currentLevel int, levelsCompleted []int) {
+	// Sort keys at current level to ensure consistent output
+	var nodes []branchNode
+	for k := range currentTree {
+		nodes = append(nodes, k)
+	}
```
Naatan, line 126: This is really signalling to me that you're abusing the `tree` type a bit. You don't actually care about the keys, the keys were only about quickly finding entries. I think you should keep that logic centralized to the functions that care about it. Now the rest of your logic is handicapped in using a type that you only really needed for one specific use case.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+	sort.Slice(nodes, func(i, j int) bool {
+		return nodes[i].Label < nodes[j].Label
+	})
+
+	for i, node := range nodes {
+		edge := edgeMid
```
Naatan, line 132: Can we name `edge` something more descriptive like `prefix` or `listBullet` or something else that doesn't require me to cross reference what this var contains.

internal/runners/branch/tree.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+}
+
+func (bt *BranchTree) printNode(w io.Writer, node branchNode, currentLevel int, levelsCompleted []int, edge string) {
+	// Print necessary edge links for current depth
+	for i := 0; i < currentLevel; i++ {
+		// Apply spacing for completed levels
+		// Do not print edge links for projects with multiple root-level branches
+		if isCompleted(levelsCompleted, i) || (i == 0 && len(bt.rootBranches) > 1) {
```
Naatan, line 153: I don't think you need `levelsCompleted` at all. All you need is `currentLevel`, and this conditional would just be `currentLevel == 0`.

Naatan (changes requested):

internal/runners/branch/tree.go
daved: Yep! It's not something I've needed to do often, so I can't offer much in the way of guidance regarding infinite recursion except that (IIRC) I've tended toward checking that on insertion. Though, I suppose that may be swayed by read versus write priority.

internal/runners/branch/list.go
MDrakos: Yes, again from the mock

internal/runners/branch/list.go
MDrakos: You're right. Updating this to handle output other than `plain` led to more refactoring.

internal/runners/branch/tree.go
MDrakos: I tried to convey how this should be used by having exported and private functions but since this is all in the same package I can see where the confusion comes from.

internal/runners/branch/tree.go
MDrakos: It will as long as it also has a nil check

internal/runners/branch/tree.go
MDrakos: Passing the writer was to help with recursively building the output but this can also be done with just a string as long as the recursive functions return a string

internal/runners/branch/tree.go
MDrakos: Sorry this comment wasn't complete. Checking `levelsCompleted` also determines if we should print the link prefix (ie. `|`) for any levels that are still incomplete. Your suggestion would work if it was just spacing as the comment alluded to.

internal/runners/branch/tree.go
MDrakos: The data that is contained in the keys is as much part of the output as the structure of the map. The keys are the nodes so they are not just for lookup. Leveraging the structure of the map along with the keys is how we build the output.

This sorting is to ensure that for each level of the tree the nodes are in alphabetical order. Similar to how we were sorting the branches when the output was based on a slice.

internal/runners/branch/tree.go
```diff
@@ -17,108 +16,87 @@ type branchNode struct {
...
-		if branch.Tracks == nil {
-			continue
-		}
-
-		// Check that this branch is a child branch and recursively build its tree
-		if _, ok := t[branchNode{branch.Label, branch.BranchID.String()}]; ok {
-			t[branchNode{branch.Label, branch.BranchID.String()}] = buildBranchTree(branch, branches)
-		}
+func (bt *BranchOutput) MarshalOutput(format output.Format) interface{} {
+	if format != output.PlainFormatName {
```
Naatan, line 41:
```suggestion
	if format != output.PlainFormatName && format != output.SimpleFormatName {
```

internal/runners/branch/tree.go
```diff
@@ -17,108 +16,87 @@ type branchNode struct {
...
+		return branchListing(bt.branches, bt.localBranch)
```
Naatan, line 42:
```suggestion
		return bt.branches
```

You can get rid of `branchListing`.

internal/runners/branch/tree.go
```diff
@@ -17,108 +16,87 @@ type branchNode struct {
...
+	tree := make(tree)
+	for _, branch := range model.GetRootBranches(branches) {
+		tree[branchNode{branch.Label, branch.BranchID.String()}] = buildBranchTree(branch, branches)
 	}
 
-	for _, b := range branches {
-		if b.Tracks == nil {
+	var levelsCompleted []int
+	rootBranches := model.GetRootBranches(branches)
+	return treeString(tree, rootBranches, localBranch, 0, levelsCompleted)
```
Naatan, line 68: This is doing too much. Be a little more functional about your logic. A function that marshals to a string should ONLY be doing that.

internal/runners/branch/tree.go
```diff
@@ -17,108 +16,87 @@ type branchNode struct {
...
-}
-
-func (bt *BranchTree) AddLocalBranch(branch string) {
-	bt.localBranch = branch
-}
-
-func (bt *BranchTree) AddLocalBranchFormatting(formatting string) {
-	bt.localBranchFormatting = formatting
-}
+func getChildren(branch *mono_models.Branch, branches mono_models.Branches) tree {
```
Naatan, line 89: Nit: I would call this `getChildTree`, bit more grokkable in terms of behaviour.

internal/runners/branch/tree.go
```diff
@@ -17,108 +16,87 @@ type branchNode struct {
...
-	bt.print(w, bt.tree, 0, levelsCompleted)
-	return w.String()
+	for _, child := range children {
+		childTree[branchNode{child.Label, child.BranchID.String()}] = make(tree)
+	}
+	return childTree
 }
 
-func (bt *BranchTree) print(w io.Writer, currentTree tree, currentLevel int, levelsCompleted []int) {
+func treeString(currentTree tree, rootBranches mono_models.Branches, localBranch string, currentLevel int, levelsCompleted []int) string {
```
Naatan, line 99: `treeString` and `nodeString` below feel like they should be `.String()` methods on a struct. And as said before at the moment they do way too much.

Naatan (changes requested):

Naatan (approved):

PR #1276, opened 2021-03-08
Fix runtime usage only being reported when installing changes

Naatan:
https://www.pivotaltracker.com/story/show/177255462


mdrohmann (approved):
LGTM

daved (approved):

PR #1280, opened 2021-03-09
Ensure errors are always unwrapped

Naatan:
Ran into this while working on something.

### DIFF
--- a/cmd/state/autoupdate.go
+++ b/cmd/state/autoupdate.go
@@ -40,9 +40,9 @@ func autoUpdate(args []string, out output.Outputer, pjPath string) (bool, error)
 	out.Notice(locale.Tr("auto_update_to_version", constants.Version, resultVersion))
 	code, err := relaunch()
 	if err != nil {
-		err = errs.WrapExitCode(err, code)
+		return true, errs.WrapExitCode(err, code)
 	}
-	return true, err
+	return true, nil
 }
 
 // When an update was found and applied, re-launch the update with the current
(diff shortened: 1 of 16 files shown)


internal/errs/errs.go
```diff
@@ -52,12 +58,21 @@ func (e *WrapperError) Stack() *stacktrace.Stacktrace {
...
 }
 
+func (e *WrapperError) ExitCode() int {
+	return e.code
+}
+
+func (e *WrapperError) SetExitCode(code int) {
+	e.code = code
+}
+
```
daved, line 68: IMO, this type is doing too much; I suggest adding behavior to error chains using structure rather than an OOP-like "superclass":
```go
type ExitError struct {
	Code int
	Err error
}

func (e *ExitError) Unwrap() error {
	return e.Err
}

func (e *ExitError) ExitCode() int {
	return e.Code
}
```
This can be easily added to error chains that should propagate exit codes:
```go
	return nil, &errs.ExitError{42, err}
```
Then, the interface can be freed up:
```go
type ExitCoder interface {
	ExitCode() int
}
```

daved (changes requested):

cmd/state/autoupdate.go
```diff
@@ -10,35 +10,39 @@ import (
...
 	defer updater.Cleanup()
 
 	out.Notice(output.Heading(locale.Tl("auto_update_title", "Auto Update")))
 	out.Notice(locale.Tr("auto_update_to_version", constants.Version, resultVersion))
 	code, err := relaunch()
-	return true, code, err
+	if err != nil {
+		err = errs.WrapExitCode(err, code)
+	}
+	return true, err
```
daved, line 45: nit suggestion:
```suggestion
		return true, errs.WrapExitCode(err, code)
	}
	return true, nil
```

internal/errs/unwrap.go
```diff
@@ -11,7 +11,7 @@ func UnwrapExitCode(err error) int {
 		return 0
 	}
 
-	var eerr interface{ ExitCode() int }
+	var eerr ExitCodeable
 	isExitError := errors.As(err, &eerr)
 	if isExitError {
```
daved, line 16: nit suggestion ninja refactor:
```suggestion
	if errors.As(err, &eerr) {
```

daved (approved):

daved (approved):

PR #1282, opened 2021-03-10
Runtime refactor activates Alternative build

mdrohmann:
https://www.pivotaltracker.com/story/show/177212331


pkg/platform/runtime2/artifact/download.go
```diff
@@ -15,8 +14,12 @@ type ArtifactDownload struct {
 	Checksum    string
 }
 
+type urlSigner interface {
+	SignS3URL(uri *url.URL) (*url.URL, error)
+}
+
 // NewDownloadsFromBuild extracts downloadable artifact information from the build status response
-func NewDownloadsFromBuild(buildStatus *headchef_models.BuildStatusResponse) ([]ArtifactDownload, error) {
+func NewDownloadsFromBuild(model urlSigner, buildStatus *headchef_models.BuildStatusResponse) ([]ArtifactDownload, error) {
```
Naatan, line 22: Nit: I'd prefer not doing this again unless we really have to. Just have `NewDownloadsFromBuild` call a new function and then do the signing with the results of that function. Then you'd just have to unit test the new function that has nothing to do with signing.

pkg/platform/runtime2/runtime.go
```diff
@@ -55,7 +61,16 @@ func (r *Runtime) Update(msgHandler setup.MessageHandler) error {
 }
 
 func (r *Runtime) Environ(inherit bool) (map[string]string, error) {
-	return r.store.Environ(inherit)
+	env, err := r.store.Environ(inherit)
+	if !r.envAccessed {
+		if err != nil {
+			analytics.EventWithLabel(setup.CatRuntime, setup.ActFailure, setup.LblEnv)
```
Naatan, line 67: I'd like the dependency on `setup` to be razor thin. Can we either move these constants to the analytics package or just create some shared types package for the runtime stuff?

pkg/platform/runtime2/setup/buildlog/buildlog.go
```diff
@@ -66,6 +65,7 @@ func New(artifactMap map[artifact.ArtifactID]artifact.ArtifactRecipe, conn Build
 		defer close(ch)
 		defer close(errCh)
 		defer messageHandler.BuildFinished()
+		defer conn.Close()
```
Naatan, line 68: This doesn't belong here imo. The connection was passed into this function, it should be closed by whatever created it.

Naatan: Another option is to actually create the connection here, in this function.

pkg/platform/runtime2/setup/setup.go
```diff
@@ -113,20 +117,57 @@ func NewWithModel(target Targeter, msgHandler MessageHandler, model ModelProvide
 	return &Setup{model, target, msgHandler, nil}
 }
 
-// Update installs the runtime locally (or updates it if it's already partially installed)
+// updateStepError attaches a label to an error returned during the Update() function.
+// The label will be used in the analytics event send during setup failures.
```
Naatan, line 121: This is too much complexity for something we care very little about. Let's keep it simple, just make the label more generic.

Naatan (changes requested):

pkg/platform/runtime2/artifact/download.go
mdrohmann: I like doing the signing outside of this function yes.

pkg/platform/runtime2/runtime.go
mdrohmann: I'll move it to the analytics package.  Most of our event strings live there already...

pkg/platform/runtime2/setup/buildlog/buildlog.go
mdrohmann: Yes, you are right.  It should be closed outside.  I prefer providing the connection as an argument, so we can test the logic with a "connection" that does not have to talk to server.

pkg/platform/runtime2/setup/setup.go
mdrohmann: I liked this approach, as it ensures that every error in the `update()` method triggers an analytics event.  But are you suggesting to just send one label for any error that the `update()` method returns?

pkg/platform/runtime2/artifact/download.go
mdrohmann: Turns out, we *had* to sign later, because if we use the build log streamer, the URL were never signed as it was!  Good that we caught it. :)

pkg/platform/runtime2/setup/setup.go
Naatan: Yes, for now I think that's fine. The reason we put it in the analytics is specifically because it was cheap to do so. It's already a redundancy because the actual error should be reported and analyzed on rollbar, not GA.

So any sort of hoop jumping to get this into the GA label is too much effort.

mdrohmann: I noticed that we would still do double reporting.  If you look at the deploy runner, you will see, that there are four occurrences of `runtime.New()` each of which will send a `"runtime", "start"` event.  

To do this cleanly (and not with lots of hard-to-grok) state variables and if-clauses inside the runtime package, we probably need a separate struct (maybe `RuntimeAnalytics`) that we instantiate somewhere at the runner level, and then can forward to `runtime.New()` and `runtime.Update()` calls. This struct could ensure that every Event is sent only once during an instances lifetime.

Another option would be to add `analytics.EventOnce()` and `analytics.EventWithLabelOnce` calls that keep track of what events have been reported already (in some struct).  That is more magical though.

I could do either of these things (or something else) in this PR, or create a follow-up story... If it was up to me, I would just go for the first approach in this PR.  But I let you decide, of course.

pkg/platform/runtime2/runtime.go
```diff
@@ -29,7 +29,11 @@ func IsNeedsUpdateError(err error) bool {
 	return errs.Matches(err, &NeedsUpdateError{})
 }
 
-func New(target setup.Targeter) (*Runtime, error) {
+func new(target setup.Targeter, noSetupYet bool) (*Runtime, error) {
+	if noSetupYet {
```
mdrohmann, line 33: I realized that `"runtime", "start"` also needs to be called during `runtime.New()` not only during the `Update()` method.  But unfortunately, the `runtime.New()` method is called by the `runtime.Update()` function as well.  And we do not want to double-send the event.  So, I had to add this (rather awkward) flag to check if we were called from the `Update()` function. :(

mdrohmann: This of course, could be removed if we had a smarter analytics call, that ensures that Events are called only once.

pkg/platform/runtime2/runtime.go
```diff
@@ -42,17 +46,23 @@ func New(target setup.Targeter) (*Runtime, error) {
 		return rt, &NeedsUpdateError{errs.New("Runtime requires setup.")}
 	}
 
-	analytics.Event(setup.CatRuntime, setup.ActCache)
+	if noSetupYet {
+		analytics.Event(analytics.CatRuntime, analytics.ActRuntimeCache)
+	}
```
mdrohmann, line 51: Similar issue as above.  But this is even more tricky, as it is not only about sending an event only once.  When called from the `Update()` function, we don't want to send this event ever, even if it was the first time, we tried...

mdrohmann: Actually, what do you think about adding a `runtime.NewWithUpdate()` function that does all the things: trying to get a runtime from cache, and if not available it calls the `Update()` method (which then could become private).  Those two functions would be more straight-forward to instrument without this awkward `noSetupYet` flag.  It would also reduce code duplication, as we essentially have the `NewWithUpdate()` function written explicitly in several runners right now...

pkg/platform/runtime2/runtime.go
mdrohmann: :facepalm: :+1:

mdrohmann: @Naatan It is fixed.  And I suppose you want me to file a story similar to what I proposed here: 

> To do this cleanly (and not with lots of hard-to-grok) state variables and if-clauses inside the runtime package, we probably need a separate struct (maybe RuntimeAnalytics) that we instantiate somewhere at the runner level, and then can forward to runtime.New() and runtime.Update() calls. This struct could ensure that every Event is sent only once during an instances lifetime.

pkg/platform/runtime2/setup/setup.go
Naatan: Yeah I think what we want is an `analytics.New()` that returns a struct which can be configured to only accept (certain?) events once. It'd be nice if all of the complexities regarding analytics are contained within the analytics package.

Naatan: That's a separate story altogether though, like we discussed in stand.

pkg/platform/runtime2/runtime.go
Naatan: Yes please, by all means file a story for that :)

Naatan (approved):

PR #1283, opened 2021-03-10
Merge environment definition in the correct order

mdrohmann:
https://www.pivotaltracker.com/story/show/177214086

### DIFF
--- a/pkg/platform/runtime2/store/store.go
+++ b/pkg/platform/runtime2/store/store.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/pkg/platform/api/headchef/headchef_models"
 	"github.com/ActiveState/cli/pkg/platform/api/inventory/inventory_models"
+	"github.com/ActiveState/cli/pkg/platform/runtime2/artifact"
 	"github.com/ActiveState/cli/pkg/platform/runtime2/envdef"
 	"github.com/ActiveState/cli/pkg/platform/runtime2/model"
 )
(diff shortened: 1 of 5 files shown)


pkg/platform/runtime2/store/store.go
```diff
@@ -139,8 +140,10 @@ func (s *Store) StoreRecipe(recipe *inventory_models.Recipe) error {
 	return nil
 }
 
-func (s *Store) Artifacts() ([]StoredArtifact, error) {
-	stored := []StoredArtifact{}
+// Artifacts loads artifact information collected during the installation.
+// It includes the environment definition configuration and files installed for this artifact.
+func (s *Store) Artifacts() (map[artifact.ArtifactID]StoredArtifact, error) {
+	stored := make(map[artifact.ArtifactID]StoredArtifact)
```
Naatan, line 146: Nit: Should make this a type.

pkg/platform/runtime2/store/store.go
```diff
@@ -205,20 +209,25 @@ func (s *Store) Environ(inherit bool) (map[string]string, error) {
 	return envDef.GetEnv(inherit), nil
 }
 
-func (s *Store) UpdateEnviron() error {
+func (s *Store) UpdateEnviron(buildStatus *headchef_models.BuildStatusResponse) error {
```
Naatan, line 212: Nit: a little awkward that this is dealing with the full build response. Might be cleaner if we introduce a model function that just returns the data we want. Then it should also be much easier to unit test this.

Naatan (approved):

Naatan (approved):
Much better!

PR #1285, opened 2021-03-15
Windows systray

MDrakos:
https://www.pivotaltracker.com/story/show/177209445


cmd/state-tray/internal/open/open_win.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/ActiveState/cli/internal/locale"
+)
+
+// Prompt will open the a command prompt and execute the given command string
+func Prompt(command string) error {
+	// start will open an instance of the given executable. The first parameter
+	// of start is the title. The /k flag tells the cmd prompt to stay open
+	// after the command has completed execution.
+	cmd := exec.Command("cmd.exe", "/c", "start", "", "cmd.exe", "/k", command)
```
Naatan, line 16: What are the 4th and 5th arguments for? So the empty one and the second "cmd.exe".

cmd/state-tray/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+	mQuit := systray.AddMenuItem("Exit", "")
+
+	for {
+		select {
+		case <-mAbout.ClickedCh:
+			logging.Debug("About event")
+			err := open.Prompt("state --version")
+			if err != nil {
+				handleError(errs.Wrap(err, "Could not open command prompt"))
```
Naatan, line 44: This is awkward. This function should just be returning errors. Then the calling function can have one single call to `handleError`, or just handle the error itself. The point is to unify the error handling, not to send it into a black hole.

cmd/state-tray/main.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+func onExit() {
+	// Not implemented
+}
+
+func handleError(err error) {
+	fmt.Fprintln(os.Stderr, err.Error())
+	logging.Error("Systray encountered an error: %v", err)
```
Naatan, line 72: If an error occurred we should also exit non-zero.

cmd/state-tray/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/getlantern/systray"
+	"github.com/gobuffalo/packr"
+)
+
+func main() {
+	systray.Run(onReady, onExit)
+}
+
+func onReady() {
```
Naatan, line 18: None of the user facing strings here are localized.

Naatan (changes requested):

cmd/state-tray/internal/open/open_win.go
MDrakos: The empty string is the title. In this case, when a new command prompt is open that will be the string in the title bar. The second `cmd.exe` is the program to start. I've updated the comment slightly

cmd/state-tray/main.go
MDrakos: I've updated this to exit. Though I'm not sure what's worse on error: a system tray program that seemingly crashes or one that does nothing. Both would prompt me to look for a log file somewhere.

cmd/state-tray/main.go
MDrakos: I'm not convinced that logging and error and printing it to `stderr` is a black hole. However, I see what you mean about unifying the error handling. I've wrapped this function and handled the errors in one place.

cmd/state-tray/main.go
MDrakos: Whoops, done now.

cmd/state-tray/main.go
Naatan: It's a black hole from the perspective of the function. I'd rather be handling errors "the go way" and then have one piece of code handle the reporting. By just throwing errors into this function at any point that you would otherwise returning an error it become a bit of a "black hole".

cmd/state-tray/main.go
```diff
@@ -16,32 +17,56 @@ func main() {
...
+	mQuit := systray.AddMenuItem(locale.Tl("tray_exit", "Exit"), "")
 
 	for {
 		select {
 		case <-mAbout.ClickedCh:
 			logging.Debug("About event")
 			err := open.Prompt("state --version")
 			if err != nil {
-				handleError(errs.Wrap(err, "Could not open command prompt"))
+				return errs.Wrap(err, "Could not open command prompt")
```
Naatan, line 69: Looking at this again now, this feels like it should just be logging the error. Since we don't want to end things just cause the "About" click caused an error.

cmd/state-tray/main.go
```diff
@@ -16,32 +17,56 @@ func main() {
 }
 
 func onReady() {
+	err := run()
```
Naatan, line 20: So is the ONLY error that can happen at the moment the "About" click? The systray stuff just doesn't throw errors?

Naatan (changes requested):

cmd/state-tray/main.go
MDrakos: Correct, nothing in the `systray` library returns an error. I'm not sure if this `run` function is necessary anymore if we are going to be logging the errors when they occur. Do you still want it?

cmd/state-tray/main.go
MDrakos: Makes sense. What do you think about setting logging to verbose by default here so that any calls to `logging.Error` also print? That way we don't have to do the additional call to `fmt.Fprintln`

Naatan: Sure, I think that makes sense.

cmd/state-tray/main.go
Naatan: Let's leave `run` where it is, that at least establishes the contract we want to keep long term.

Naatan (dismissed):

Naatan (approved):

PR #1288, opened 2021-03-16
Add camel runtime setup

mdrohmann:
https://www.pivotaltracker.com/story/show/176968140


pkg/platform/runtime2/runtime.go
```diff
@@ -79,7 +94,39 @@ func (r *Runtime) Environ(inherit bool) (map[string]string, error) {
 		}
 		r.envAccessed = true
 	}
-	return env, err
+	return injectProjectDir(env, projectDir), err
+}
+
+// injectProjectDir replaces {{.ProjectDir}} with the current project in environment variables
```
Naatan, line 100: For our (my) sanity, let's move this into a file called `legacy.go`.

pkg/platform/runtime2/setup/implementations/alternative/artifact.go
```diff
@@ -45,3 +45,7 @@ func (as *ArtifactSetup) Move(tmpDir string) error {
 func (as *ArtifactSetup) Unarchiver() unarchiver.Unarchiver {
 	return unarchiver.NewTarGz()
 }
+
+func (as *ArtifactSetup) InstallerExtension() string {
```
Naatan, line 49: Let's remove this and have the Unarchiver instead return an interface that implements` unarchiver.Unarchiver` as well as `Ext() string`.

pkg/platform/runtime2/stories.md
```diff
@@ -0,0 +1,40 @@
+## A Load runtime2.Runtime with runtime installed by old runtime package
```
Naatan, line 1: Make sure you remove this before the merge.

Naatan (changes requested):

pkg/platform/runtime2/runtime.go
```diff
@@ -47,6 +54,9 @@ func new(target setup.Targeter) (*Runtime, error) {
 
 // New attempts to create a new runtime from local storage.  If it fails with a NeedsUpdateError, Update() needs to be called to update the locally stored runtime.
 func New(target setup.Targeter) (*Runtime, error) {
+	if strings.ToLower(os.Getenv(constants.DisableRuntime)) == "true" {
+		return DisabledRuntime, nil
+	}
```
mdrohmann, line 59: I wanted to have the check for `ACTIVESTATE_DISABLE_RUNTIME` in one place only, so it needs to be here.  I did not want to return `nil`, as then the caller would have to check for that.  With the `DisabledRuntime` we will also get better error messages, in case we mess something up...

pkg/platform/runtime2/setup/implementations/camel/ape_installer.go
```diff
@@ -0,0 +1,75 @@
+package camel
```
mdrohmann, line 1: This file is basically moved from `pkg/platform/runtime/ape_installer.go`.

pkg/platform/runtime2/setup/implementations/camel/apy_installer.go
```diff
@@ -0,0 +1,73 @@
+package camel
```
mdrohmann, line 1: This file is basically moved from `pkg/platform/runtime/apy_installer.go`.

pkg/platform/runtime2/setup/implementations/camel/metadata.go
```diff
@@ -0,0 +1,150 @@
+package camel
```
mdrohmann, line 1: Changes compared to `pkg/platform/runtime/metadata.go`:

```
--- pkg/platform/runtime2/setup/implementations/camel/metadata.go	2021-03-18 14:30:14.202666220 -0700
+++ pkg/platform/runtime/metadata.go	2021-03-18 10:08:20.022254760 -0700
@@ -1,10 +1,9 @@
-package camel
+package runtime
 
 import (
 	"encoding/json"
 	"os"
 	"path/filepath"
-	"strings"
 
@@ -26,8 +25,8 @@
 
 // MetaData is used to parse the metadata.json file
 type MetaData struct {
-	// InstallDir is the root directory of the artifact files that we need to copy on the user's machine
-	InstallDir string
+	// Path is the directory containing the meta file
+	Path string
 
 	// AffectedEnv is an environment variable that we should ensure is not set, as it might conflict with the artifact
 	AffectedEnv string `json:"affected_env"`
@@ -60,9 +59,9 @@
 }
 
 // InitMetaData will create an instance of MetaData based on the metadata.json file found under the given artifact install dir
-func InitMetaData(rootDir string) (*MetaData, error) {
+func InitMetaData(installDir string) (*MetaData, error) {
 	var metaData *MetaData
-	metaFile := filepath.Join(rootDir, "support", constants.RuntimeMetaFile)
+	metaFile := filepath.Join(installDir, constants.RuntimeMetaFile)
 	if fileutils.FileExists(metaFile) {
 		contents, err := fileutils.ReadFile(metaFile)
 		if err != nil {
@@ -81,20 +80,8 @@
 		metaData.Env = map[string]string{}
 	}
 
-	var relInstallDir string
-	installDirs := strings.Split(constants.RuntimeInstallDirs, ",")
-	for _, dir := range installDirs {
-		if fileutils.DirExists(filepath.Join(rootDir, dir)) {
-			relInstallDir = dir
-		}
-	}
-
-	if relInstallDir == "" {
-		logging.Debug("Did not find an installation directory relative to metadata file.")
-	}
-
-	metaData.InstallDir = relInstallDir
-	err := metaData.Prepare(filepath.Join(rootDir, relInstallDir))
+	metaData.Path = installDir
+	err := metaData.Prepare()
 	if err != nil {
 		return nil, err
 	}
@@ -120,11 +107,11 @@
 	return metaData, nil
 }
 
-func (m *MetaData) hasBinaryFile(root string, executable string) bool {
+func (m *MetaData) hasBinaryFile(executable string) bool {
 	for _, dir := range m.BinaryLocations {
 		parent := ""
 		if dir.Relative {
-			parent = root
+			parent = m.Path
 		}
 		bin := filepath.Join(parent, dir.Path, executable)
 		if fileutils.FileExists(bin) {
@@ -142,10 +129,6 @@
 		logging.Debug("Not setting PYTHONPATH as the user already has it set")
 	}
 
-	// This is broken for two reasons:
-	// 1. Checking in the OS environment will only happen on installation, but at a later point, the OS environment might have changed, and we will overwrite the user's choice here
-	// 2. python code does not need to depend on PYTHONIOENCODING as pointed out here: https://stackoverflow.com/a/9942822
-	// Follow up story is here: https://www.pivotaltracker.com/story/show/177407383
 	if os.Getenv("PYTHONIOENCODING") == "" {
 		m.Env["PYTHONIOENCODING"] = "utf-8"
 	} else {
```

pkg/platform/runtime2/store/store.go
```diff
@@ -253,13 +253,6 @@ func (s *Store) updateEnviron(orderedArtifacts []artifact.ArtifactID, artifacts
 		}
 	}
 
-	cnst := envdef.NewConstants(s.InstallPath())
-	rtGlobal = rtGlobal.ExpandVariables(cnst)
-	err := rtGlobal.ApplyFileTransforms(s.InstallPath(), cnst)
-	if err != nil {
-		return nil, locale.WrapError(err, "runtime_alternative_file_transforms_err", "", "Could not apply necessary file transformations after unpacking")
-	}
-
```
mdrohmann, line 262: I moved this block into the `setupArtifact` function for these reasons:

- makes this function more unixy. :)
- the relocation can and should be parallelized (can be applied for each artifact individually)
- in the `setupArtifact` function the relocation is more isolated, as it is run inside of the temporary installation directory where only one artifact is unpacked to.

Naatan (approved):

PR #1291, opened 2021-03-24
add runtime msg handler

mdrohmann:
https://www.pivotaltracker.com/story/show/176968116

Camel installation:
camel

Installing a big project:
big-deploy2

Installing with the build log streamer:
buildlog-streamer


internal/runbits/progressbar.go
```diff
@@ -0,0 +1,222 @@
...
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+type artifactStepBar struct {
+	started time.Time
+	bar     *mpb.Bar
+}
+
+// progressBar receives update events and modifies a global state accordingly
+type progressBar struct {
```
Naatan, line 23: I feel like this name is a bit missleading. Maybe instead call it `runtimeProgress` or something?

internal/runbits/progressbar.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (pb *progressBar) buildStepName(step events.ArtifactSetupStep, artifactName string) string {
+	var prefix string
+	switch step {
+	case events.Download:
+		prefix = "D.."
+	case events.Unpack:
+		prefix = ".U."
+	case events.Install:
+		prefix = "..I"
```
Naatan, line 151: I don't see why we need to be abbreviating stuff. I know you're working without a mock but try to maximize the user experience. This won't make sense to anyone but us.

internal/runbits/progressbar.go
Naatan: Additionally looking at the methods on `progressBar` this feels really messy atm. It's not clear what `progressBar`'s responsibility is.

At the moment it does at least 3 different thing:

 - Handle runtime messages
 - Track progressbar state
 - Provide helper utilities (like trimName)

It would be nice if you could clean this up significantly. I'd like to be able to look at just the method names and get a good understanding of how it operates.

internal/runbits/proxyreader/proxyreader.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+type ByIncrementer interface {
+	IncrBy(int)
+}
+
+var _ io.Reader = &ProxyReader{}
+
+// ProxyReader wraps around a reader and calls the incrementer function on every read reporting the number of bytes read
+type ProxyReader struct {
```
Naatan, line 16: This does not seem like it belongs in `runbits`.

internal/runbits/runtimemsg.go
```diff
@@ -3,61 +3,92 @@ package runbits
...
+		ar.Name, strconv.Itoa(len(addedDependencies)), strconv.Itoa(len(changed.Added)),
+	))
+	for i, dep := range addedDependencies {
+		depMapping, ok := artifacts[dep]
+		if !ok {
+			logging.Error("Could not find dependency %s in artifactsMap", dep)
+			continue
+		}
+		var depCount string
+		recDeps := artifact.RecursiveDependenciesFor(dep, artifacts)
```
Naatan, line 57: This should probably be passed to `ChangeSummary` as an argument instead.

pkg/platform/runtime/setup/events/events.go
```diff
@@ -0,0 +1,204 @@
+package events
```
Naatan, line 1: I'm not sure I understand why this package exists. It seems to contain a ton of boilerplate and very little actual logic. Why do we need this?

Naatan (changes requested):
This feels very proof of concept to me. Please spend a little time reviewing your own code. There's a bunch of code that feels like it's in the wrong place or under the wrong struct type. The event stuff in particular is a big red flag to me, but I can't say for certain without having the full context. As it stands it's not easy to grok what its purpose is.

internal/runbits/progressbar.go
mdrohmann: Fair point...

internal/runbits/runtimemsg.go
mdrohmann: It would be quite expensive to compute this for every artifact.  That's why I preferred the on-demand version.

pkg/platform/runtime/setup/events/events.go
mdrohmann: To address your overall criticism of the package organization, let me explain a little more what my motivations were for the organization as it is:

As you saw the logic is split into the `runbits` and the `runtime/setup/events` packages.  
- My intention is to keep the generation of the events in the `events` package, because I felt that this has nothing to do with "code that prints"
- The same is true for the `RuntimeEventConsumer` that consumes the events, and carries and manipulates  some state with it.  - As a result, the `ProgressBar` in the `runbits` package does not have to deal with events anymore and is more target to producing output, hence it is in `runbits`
- If we want to do our own progress bar implementation, we could just replace the `runbits/progressbar` implementation with something that does not depend on `vbauerster/mpb`

So, in short the main distinction is: `runbits` has structs that deal with outputs, and `setup/events` with events as the are reported by the setup update function.

internal/runbits/progressbar.go
mdrohmann: As explained in my comment on the `events` package, this file (as it is in the runbits package) is focused on the output part, so to actually print the progress bars.  Therefore, it needs to keep the `mpb.Bar` pointers, which hold output-related state.
It is a little unfortunate, that I had to add the `artifactStepState` (I should rename it into `artifactStepBar`), because the `mpb.Bar` does not capture the time when it started, but wants me to tell it with every update.  I believe that this optimally should be stored in the `mpb.Bar` itself, so it is done here.  

Other Action items based on your comments:
- I will split out the `trimWidth` and `addTotalBar`, `addProgressBar` implementations into separate file, as they are more specific, but I think that they should still be part of the same struct.  As the entire file describes how to display the runtime progress.
- I will remove the dependency on `events.ArtifactsSetupStep` by moving the translation from the `Step` to a "title" into the `events` package.  That is certainly cleaner.

internal/runbits/proxyreader/proxyreader.go
mdrohmann: I moved it to `internal/proxyreader`.  I couldn't really think of any other package that it could fit with...

pkg/platform/runtime/setup/events/events.go
mdrohmann: Also if you refer to the `RuntimeProducer` for the boilerplate code: It is indeed not doing much more than creating events and sending them to the encapsulated channel.  However, I found it valuable to have a struct that *does* the encapsulation, such that we do not have to expose the channel anywhere.

Writing these simple functions also serves the purpose of implementing interfaces like the `BuildLogMessageHandler` interface.  So, we can easily replace the producer with a mock in tests...

pkg/platform/runtime/runtime.go
```diff
@@ -66,9 +65,26 @@ func New(target setup.Targeter) (*Runtime, error) {
 	return r, err
 }
 
-func (r *Runtime) Update(msgHandler setup.MessageHandler) error {
+type UpdateMessageHandler interface {
+	HandleUpdateEvents(eventCh <-chan events.BaseEventer)
```
Naatan, line 69: This is breaking the promise of the runtime message handler imo. Why do we have callbacks as well as channels? Use one or the other. Preferrably callbacks, leave the channels inside the runtime package.

internal/runbits/runtimemsg.go
```diff
@@ -3,61 +3,97 @@ package runbits
...
+		}
+		rmh.out.Notice(fmt.Sprintf("  [DISABLED]%s[/RESET] %s%s", prefix, depMapping.Name, depCount))
+	}
+	return nil
 }
 
-func (r RuntimeMessageHandler) BuildFinished() {
-	r.out.Notice(fmt.Sprintf("Build Finished"))
+// HandleUpdateEvents prints output based on runtime events received on the eventCh
+func (rmh *RuntimeMessageHandler) HandleUpdateEvents(eventCh <-chan events.BaseEventer) {
```
Naatan, line 73: This is essentially creating a whole new message handler inside the thing that is already a message handler. Please simplify.

internal/runbits/runtimemsg.go
```diff
@@ -3,61 +3,97 @@ package runbits
...
 	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
-	"github.com/ActiveState/cli/pkg/platform/runtime/setup"
+	"github.com/ActiveState/cli/pkg/platform/runtime/setup/events"
 )
 
 type RuntimeMessageHandler struct {
 	out output.Outputer
 }
 
 func NewRuntimeMessageHandler(out output.Outputer) *RuntimeMessageHandler {
```
Naatan, line 27: Looking at the code again in my IDE I now see that you're mainly just using the message handler to pass a channel and then do most of your logic via the channel. This pretty much nillifies the purpose of the message handler. Please commit to one or the other. (Use callbacks or use channels, not both). My preference goes to callbacks because it's much easier to work with. Write dumb code.

Naatan (changes requested):
I'm sorry but I am not sold on the events package. It seems like you just moved most of the runtime message handler logic into the events package and are now exposing a much harder to work with channel to consumers. Please dumb it down.

mdrohmann:
Changes:
- I added lots of documentation and comments
- moved the changeSummary function to its own package
- revisited the progressbar.go implementation
- simplified channel handling in `runtime.Runtime.Update` (it should be clearer now, that I just introduced the channel wrapped by the `Producer`
- cleaned up the runtime message handler

I couldn't quite figure out how to reduce the number of lines in the `events` package...

pkg/platform/runtime/setup/setup.go
```diff
@@ -163,7 +170,15 @@ func (s *Setup) update() error {
...
-	s.deleteOutdatedArtifacts(changedArtifacts, storedArtifacts)
+	if buildResult.BuildEngine == model.Camel {
+		// for camel builds we have to wipe previous installations
+		os.RemoveAll(s.store.InstallPath())
+	} else {
+		err := s.deleteOutdatedArtifacts(changedArtifacts, storedArtifacts)
+		if err != nil {
+			return locale.WrapError(err, "err_delete_outdated", "Could not remove outdated artifacts files in {{.V0}}.  You may try to remove the entire directory manually.", s.store.InstallPath())
+		}
+	}
```
Naatan, line 181: This should be happening in the camel package. Add an interface method if you need to, I don't want to start sprinkling this into the non-builder specific code.

The ideal of keeping that interface as light as possible does not mean we should cheat our way around it.

pkg/platform/runtime/setup/setup.go
```diff
@@ -232,17 +251,23 @@ func (s *Setup) deleteOutdatedArtifacts(changeset artifact.ArtifactChangeset, st
...
 	if err != nil {
 		return errs.Wrap(err, "Could not fetch artifacts to download.")
 	}
+	s.msgHandler.TotalArtifacts(len(downloads))
 	for _, a := range downloads {
 		func(a artifact.ArtifactDownload) {
 			wp.Submit(func() {
-				if err := s.setupArtifact(buildResult.BuildEngine, a.ArtifactID, a.UnsignedURI); err != nil {
+				// 'bundle' makes sense for the single camel download
+				name := "bundle"
```
Naatan, line 266: This is going to bite us. Please just feed in something that signals a painpoint, eg. "Unknown".

pkg/platform/runtime/setup/setup.go
```diff
@@ -271,13 +298,21 @@ func (s *Setup) installFromBuildLog(buildResult *model.BuildResult, artifacts ma
...
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
+		wp := workerpool.New(MaxConcurrency)
+		defer wp.StopWait()
 		for a := range buildLog.BuiltArtifactsChannel() {
 			func(a artifact.ArtifactDownload) {
 				wp.Submit(func() {
-					if err := s.setupArtifact(buildResult.BuildEngine, a.ArtifactID, a.UnsignedURI); err != nil {
+					name := "bundle"
```
Naatan, line 311: Use "Unknown"

pkg/platform/runtime/setup/setup.go
```diff
@@ -289,7 +324,8 @@ func (s *Setup) installFromBuildLog(buildResult *model.BuildResult, artifacts ma
 		return err
 	}
 
-	wp.StopWait()
+	// wait for the build artifacts to be processed
+	wg.Wait()
```
Naatan, line 328: Why was this necessary?

pkg/platform/runtime/setup/setup.go
```diff
@@ -313,21 +349,34 @@ func (s *Setup) setupArtifact(buildEngine model.BuildEngine, a artifact.Artifact
...
-	if err := s.downloadArtifact(unsignedURI, archivePath); err != nil {
-		return errs.Wrap(err, "Could not download artifact %s", unsignedURI)
+	downloadProgress := events.NewIncrementalProgress(s.msgHandler, events.Download, a, artifactName)
+	if err := s.downloadArtifact(unsignedURI, archivePath, downloadProgress); err != nil {
+		err := errs.Wrap(err, "Could not download artifact %s", unsignedURI)
+		s.msgHandler.ArtifactStepFailed(events.Download, a, err.Error())
+		return err
 	}
-	s.msgHandler.ArtifactDownloadCompleted(a)
+	s.msgHandler.ArtifactStepCompleted(events.Download, a)
```
Naatan, line 358: Why are we still calling it a msgHandler here? Should be `s.events...` I would think.

pkg/platform/runtime/runtime.go
```diff
@@ -66,19 +65,43 @@ func New(target setup.Targeter) (*Runtime, error) {
...
+		rt, err := new(r.target)
+		if err != nil {
+			setupErr = errs.Wrap(err, "Could not reinitialize runtime after update")
+			return
+		}
+		*r = *rt
+	}()
+
+	// ... and handle the runtime events in the main thread
+	err := msgHandler.HandleEvents(prod.Events())
```
Naatan, line 92: Is `HandleEvents` blocking? Would be good if we convey that in the method name, or have an additional `msgHandler.Wait()`.

Naatan (changes requested):

pkg/platform/runtime/setup/setup.go
mdrohmann: Thinking about it again, I think I'll need to make this also build engine specific.  For camel it should be "bundle", for alternative builds "unknown".

pkg/platform/runtime/setup/setup.go
mdrohmann: It is an unlikely situation, but likely enough such that I ran into it during testing:

Here is how it was.


```
Thread A                        Thread B
receive last artifact download (line 278 / 307)
                                       buildLog.Wait() returns (line 288/323)
                                       wp.StopWait() (line 292) closes the workerpool channel
wp.Submit() (line 279/310) causes a "closed channel"-panic.
```

The new version ensures that `wp.StopWait()` is called after all `wp.Submit()` calls are issued.

pkg/platform/runtime/setup/setup.go
mdrohmann: Good point.  `events` seems good.

pkg/platform/runtime/runtime.go
mdrohmann: Yes, I was thinking about that name for a long time, but wasn't really clear on what needed to be conveyed.

I like `WaitForAllEvents()`, `HandleAllEvents` or `HandleAllEventsAndWait()`.

The second would be my favorite.

pkg/platform/runtime/setup/setup.go
Naatan: Maybe add some comments explaining this so someone doesn't accidentally "clean this up" (ie. me.. :p)

pkg/platform/runtime/runtime.go
Naatan: I don't think `HandleAllEvents` conveys that it blocks. Out of these options I would favor `WaitForAllEvents()`.

Naatan (approved):

PR #1292, opened 2021-03-25
Move the runtime files to final destination synchronously

mdrohmann:


pkg/platform/runtime/setup/setup.go
```diff
@@ -247,27 +296,31 @@ func (s *Setup) deleteOutdatedArtifacts(changeset artifact.ArtifactChangeset, st
 	return nil
 }
 
-func (s *Setup) installFromBuildResult(buildResult *model.BuildResult, artifacts map[artifact.ArtifactID]artifact.ArtifactRecipe) error {
+// setupArtifactSubmitFunction returns a function that sets up an artifact and can be submitted to a workerpool
+func (s *Setup) setupArtifactSubmitFunction(a artifact.ArtifactDownload, buildResult *model.BuildResult, artifacts artifact.ArtifactRecipeMap, errors []error, secondStageCh chan<- func() error) func() {
+	return func() {
+		// 'bundle' makes sense for the single camel download
```
Naatan, line 302: This feels wrong, why are we making an assumption about camel downloads outside of the camel package?

pkg/platform/runtime/setup/setup.go
```diff
@@ -373,6 +414,22 @@ func (s *Setup) setupArtifact(buildEngine model.BuildEngine, a artifact.Artifact
 		return errs.Wrap(err, "Could not collect env info for artifact")
 	}
 
+	cnst := envdef.NewConstants(s.store.InstallPath())
```
Naatan, line 417: I thought we were doing this as one singular step at the end? ie. combine all the envdefs and then apply them.

pkg/platform/runtime/setup/setup.go
```diff
@@ -373,6 +414,22 @@ func (s *Setup) setupArtifact(buildEngine model.BuildEngine, a artifact.Artifact
...
+	envDef = envDef.ExpandVariables(cnst)
+	err = envDef.ApplyFileTransforms(filepath.Join(unpackedDir, envDef.InstallDir), cnst)
+	if err != nil {
+		return locale.WrapError(err, "runtime_alternative_file_transforms_err", "", "Could not apply necessary file transformations after unpacking")
+	}
+
+	// return s.moveToInstallPath(a, artifactName, unpackedDir, envDef, numFiles)
+	secondStageCh <- func() error { return s.moveToInstallPath(a, artifactName, unpackedDir, envDef, numFiles) }
```
Naatan, line 425: Maybe a slightly cleaner and more readable solution: https://github.com/faiface/mainthread/

Then we don't need to be passing any channel around either.

Naatan (changes requested):

pkg/platform/runtime/setup/setup.go
mdrohmann: Well, I guess it also makes sense for `alternative` artifacts, but that of course shouldn't happen.  We should always be able to resolve that to a "real" name.

It is just that for camel builds, there is no such name, as all artifacts are combined in one bundle.  I'll try to improve the comment.

pkg/platform/runtime/setup/setup.go
mdrohmann: The applying of the file transforms can also be parallelized safely when it is just done inside the temporary installation directory.  So, I implemented it this way.

I also just realized that we are actually not merging the transforms. So, the global environment definition will not contain some of the transformations that we need to apply.

pkg/platform/runtime/setup/setup.go
mdrohmann: I can do that...

pkg/platform/runtime/setup/setup.go
Naatan: The beauty of doing it all at the end is that it's automatically compatible with incremental updates. Is that still the case?

pkg/platform/runtime/setup/setup.go
```diff
@@ -222,41 +222,24 @@ func (s *Setup) installArtifacts(buildResult *model.BuildResult, artifacts artif
...
-	if firstStageError != nil {
-		// merge with artifact setup error if possible
-		var ase *ArtifactSetupErrors
-		if errors.As(firstStageError, &ase) {
-			artfErrs = append(artfErrs, ase.Errors()...)
-		} else {
-			artfErrs = append(artfErrs, firstStageError)
+			mainthread.Call(func() {
+				artfErrs = append(artfErrs, err)
+			})
```
Naatan, line 240: I don't think we should do this for updating a slice of errors. For that we should just use channels. Reserve `mainthread` for logic that is otherwise complicated to facilitate with channels. eg. moving a bunch of files from different artifacts without needing to combine them all into a single channel when that's not how the data flows.

Naatan (changes requested):

pkg/platform/runtime/setup/setup.go
mdrohmann: It is, yes: Every artifact's transformations are going to be applied (whether pulled in incrementally or not), but inside their own sandbox, before the files are moved to the final destination.

mdrohmann: Another advantage (next to speed), is its predictability.  Theoretically, there could be two artifacts that install the same file, but with differing transformation directives.  When we do merge the transformations in the end, we might end up applying the wrong transformation directive, or too many or in the wrong order...
With this approach, we'll only apply the correct transformations.

pkg/platform/runtime/setup/setup.go
mdrohmann: I pushed a solution using channels, it required another go routine though...

pkg/platform/runtime/setup/setup.go
```diff
@@ -223,7 +223,25 @@ func (s *Setup) installArtifacts(buildResult *model.BuildResult, artifacts artif
...
+		for err := range artfErrs {
+			errs = append(errs, err)
+		}
+
+		if len(errs) > 0 {
+			installErr <- &ArtifactSetupErrors{errs}
+		} else {
+			installErr <- nil
+		}
+	}()
```
Naatan: This looks a bit messy, see https://github.com/ActiveState/cli/pull/1295

Naatan (changes requested):

Naatan (changes requested):
Looking better! But can we move this into the `installFromBuildResult` and `installFromBuildLog` functions? Those functions block so they might as well return the errors rather than take a error channel argument.

Naatan (approved):

PR #1293, opened 2021-03-29
Systray shows locally checked out projects

MDrakos:
https://www.pivotaltracker.com/story/show/177258941


cmd/state-tray/internal/menu/projects.go
```diff
@@ -0,0 +1,73 @@
...
+func (u *LocalProjectsUpdater) Reload(projects []*graph.Project) {
+	for _, item := range u.items {
+		item.remove()
+	}
+
+	u.items = make([]localProjectsMenuItem, 0)
+	for _, project := range projects {
+		u.addLocalProject(project)
+	}
+}
```
MDrakos, line 37: This is more heavy-handed than I would like. I tried to re-use the goroutines and `menuItems` however the `systray` library does not facilitate reordering the menu items. To ensure the order returned form the daemon is reflected in the menu we have to "remove" the original entries and create new ones. If this order is not important, it can be modified to re-use existing menu entries.

cmd/state-tray/main.go
```diff
@@ -57,28 +72,56 @@ func run() error {
 
 	systray.AddSeparator()
 
+	mProjects := systray.AddMenuItem(locale.Tl("tray_projects_title", "Local Projects"), "")
+	mReload := mProjects.AddSubMenuItem("Reload", "Reload the local projects listing")
+	localProjectsUpdater := menu.NewLocalProjectsUpdater(mProjects)
```
MDrakos, line 77: I'm not sure about this approach. I think it would be better to wrap the `systray` library and return our own menu items that can handle specialized behavior. Something like:

```
mProjects := menu.NewLocalProjectsMenu(localProjects)
...
case <-mReload.ClickedCh:
	logging.Debug("Projects event")
	localProjects, err = model.LocalProjects()
	if err != nil {
		logging.Error("Could not get local projects listing: %v", err)
	}
	mProjects.Reload()
```

This could be done in this PR but would lead to different ways of creating a new menu item. Also, wrapping `systray` felt out of scope. I've created a ticket here: https://www.pivotaltracker.com/story/show/177551480

cmd/state-tray/internal/open/open_win.go
```diff
@@ -22,3 +23,8 @@ func Prompt(command string) error {
 
 	return nil
 }
+
+// Browser will open the default browser with the given URL
+func Browser(url string) error {
+	return open.Run(url)
```
Naatan, line 29: This doesn't seem like it needs to be in a windows only file.

pkg/platform/model/svc.go
```diff
@@ -24,11 +24,28 @@ func NewSvcModel(ctx context.Context, cfg *config.Instance) (*SvcModel, error) {
 	return &SvcModel{ctx, client}, nil
 }
 
+type versionResponse struct {
+	Version graph.Version `json:"version"`
+}
```
Naatan, line 29: Let's put this in the `graph` package in a file that we maintain ourselves. Only the one file in there is generated, the package itself is open for us to add our own stuff to.

Naatan (changes requested):

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -31,3 +39,30 @@ func (r *Resolver) Version(ctx context.Context) (*graph.Version, error) {
...
+	logging.Debug("Projects resolver")
+	config, err := config.New()
+	if err != nil {
+		return nil, locale.WrapError(err, "err_resolver_get_config", "Could not get new config instance")
+	}
+
+	var projects []*graph.Project
+	localConfigProjects := config.GetStringMapStringSlice(projectfile.LocalProjectsConfigKey)
+	for ns, locations := range localConfigProjects {
+		_, err := project.ParseNamespace(ns)
```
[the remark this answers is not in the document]
MDrakos: We are not storing projects with invalid namespaces. This was just another check but is most likely unnecessary.

cmd/state-tray/internal/menu/projects.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/ActiveState/cli/cmd/state-tray/internal/open"
+	"github.com/ActiveState/cli/internal/graph"
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/getlantern/systray"
+)
+
+type LocalProjectsUpdater struct {
+	menuItem *systray.MenuItem
+	items    []localProjectsMenuItem
```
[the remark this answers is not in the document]
MDrakos: You can have both pointer and non-pointer receivers. The pointer receivers are used when the method modifies the receiver in some way and to sometimes be a little bit quicker.

cmd/state-tray/internal/menu/projects.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	u.items = make([]localProjectsMenuItem, 0)
+	for _, project := range projects {
+		u.addLocalProject(project)
+	}
+}
+
+func (u *LocalProjectsUpdater) addLocalProject(project *graph.Project) {
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
MDrakos: The context is passed to the `wait` function and the `cancel` is used when a menu entry is "removed"

cmd/state-svc/internal/resolver/resolver.go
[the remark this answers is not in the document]
Naatan: Ok, please drop it then. If we need validation we should implement it in a centralized manner.

cmd/state-tray/internal/menu/projects.go
```diff
@@ -0,0 +1,73 @@
...
+
+	mProject := u.menuItem.AddSubMenuItem(project.Namespace, "")
+	u.items = append(u.items, localProjectsMenuItem{mProject, project, cancel})
+
+	if len(project.Locations) == 1 {
+		go wait(ctx, mProject, project.Namespace, project.Locations[0])
+		return
+	}
```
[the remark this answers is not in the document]
MDrakos: This check is a bit awkward but I think it needs to stay as I have to create sub-menu items if the length of the project locations slice is greater than 1 entry.

cmd/state-tray/internal/menu/projects.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (u *LocalProjectsUpdater) Update(projects []*graph.Project) {
+	for _, item := range u.items {
+		item.remove()
+	}
+
+	u.items = make([]*localProjectsMenuItem, len(projects))
+	for i, project := range projects {
+		item := newLocalProjectMenuItem(project, u.menuItem)
+		u.items[i] = item
```
Naatan, line 37: Nit: little awkward to depend on the index like this. Why not just use `append()`?

cmd/state-tray/internal/menu/projects.go
```diff
@@ -0,0 +1,83 @@
...
+		item.startEventLoop()
+	}
+}
+
+func newLocalProjectMenuItem(project *graph.Project, menuItem *systray.MenuItem) *localProjectsMenuItem {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	mProject := menuItem.AddSubMenuItem(project.Namespace, "")
```
Naatan, line 45: Nit: Might make your code a bit cleaner and better separate concerns if you add a `AddChild()` to `LocalProjectsUpdater` and pass `localProjectsMenuItem` to it. Effectively that would call `menuItem.AddSubMenuitem()` from `AddChild()` while keeping `localProjectsMenuItem` focused strictly on the menuitem itself.

So `LocalProjectsUpdater` is the thing that ties everything together.

cmd/state-tray/main.go
```diff
@@ -57,28 +72,56 @@ func run() error {
...
+				logging.Error("Could not open support url: %v", err)
+			}
 		case <-mAccount.ClickedCh:
 			logging.Debug("Account event")
 			// Not implemented
+		case <-mReload.ClickedCh:
+			logging.Debug("Projects event")
+			localProjects, err = model.LocalProjects()
+			if err != nil {
+				logging.Error("Could not get local projects listing: %v", err)
```
Naatan, line 122: Should return here, else line 124 will likely lead to a panic.

Naatan (changes requested):

cmd/state-tray/internal/menu/projects.go
MDrakos: Because I know the length of the slice. https://github.com/ActiveState/cli/pull/1293#discussion_r603638661

cmd/state-tray/internal/menu/projects.go
```diff
@@ -0,0 +1,83 @@
...
+	u.items = make([]*localProjectsMenuItem, len(projects))
+	for i, project := range projects {
+		item := newLocalProjectMenuItem(project, u.menuItem)
+		u.items[i] = item
+		item.startEventLoop()
+	}
+}
+
+func newLocalProjectMenuItem(project *graph.Project, menuItem *systray.MenuItem) *localProjectsMenuItem {
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
MDrakos: Agreed, context is too much here and it works fine with just a channel

cmd/state-tray/internal/menu/projects.go
MDrakos: I like this idea however I think it would require `AddChild()` to modify the passed in `localProjectsMenuItem` as it needs the returned type from `menuItem.AddSubMenuItem()`.

cmd/state-tray/main.go
MDrakos: The response on error will be `nil` and line 124 will not panic. I've tested it with a `nil` response and when it's running without the service, it works in both cases.

cmd/state-tray/internal/menu/projects.go
Naatan: That doesn't mean you can't use `append()`, to my knowledge?

cmd/state-tray/internal/menu/projects.go
Naatan: Why does it need to modify it? Sounds like it just needs to read a property.

changelog.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## 0.27.1
+
+### Added
+
+- System tray executable now populates list of local projects
+  ([PR #1293](https://github.com/ActiveState/cli/pull/1293))
+
```
Naatan, line 16: Let's leave this off since we're not actually distributing anything. ie. version 0.27.1 (actually we're at 0.28.0 now..) won't ship with the systray yet, even though it's part of the codebase.

cmd/state-tray/internal/menu/projects.go
Naatan: Would still like to see this addressed. The menuitem being passed this way is awkward.

cmd/state-tray/internal/menu/projects.go
```diff
@@ -16,9 +15,11 @@ type LocalProjectsUpdater struct {
 }
 
 type localProjectsMenuItem struct {
-	menuItem *systray.MenuItem
-	project  *graph.Project
-	cancel   context.CancelFunc
+	menuItem  *systray.MenuItem
+	namespace string
+	location  string
+	done      chan struct{}
```
Naatan, line 21: Nit: name this `close`. `done` conveys the menuitem itself is doing something that it needs to finish.

cmd/state-tray/internal/menu/projects.go
```diff
@@ -30,44 +31,71 @@ func (u *LocalProjectsUpdater) Update(projects []*graph.Project) {
...
-	if len(project.Locations) == 1 {
-		go waitForClick(ctx, mProject, project.Namespace, project.Locations[0])
-		return
+func newLocalProjectMenuItem(project *graph.Project, menuItem *systray.MenuItem) *localProjectsMenuItem {
+	done := make(chan struct{})
+	mProject := menuItem.AddSubMenuItem(project.Namespace, "")
+	item := &localProjectsMenuItem{
+		menuItem:  mProject,
+		namespace: project.Namespace,
+		done:      done,
```
Naatan, line 48: Nit: pointless to create a intermediary var for `done`

cmd/state-tray/internal/menu/projects.go
```diff
@@ -30,44 +31,71 @@ func (u *LocalProjectsUpdater) Update(projects []*graph.Project) {
...
 	}
+
+	return item
 }
 
-func (u *localProjectsMenuItem) remove() {
-	u.cancel()
-	u.menuItem.Hide()
+func (i *localProjectsMenuItem) addChild(menuItem *systray.MenuItem, location string) {
+	i.children = append(i.children, &localProjectsMenuItem{
```
Naatan, line 66: Use your constructor rather than manually creating a `localProjectsMenuItem{}` again.

If your constructor can't service the use case you can create a wrapper constructor. Key point being that the type is only constructed in one single function unless there's some strong reason to do otherwise.

This will likely force you to refactor your constructor, but that's not a bad thing.

Naatan (changes requested):

Naatan (approved):

Naatan (approved):

PR #1295, opened 2021-03-29
Try to clean up error handling logic

Naatan:
@mdrohmann Felt like this would be easier than doing it in the review. Haven't tested this, but key point is I feel like we can clean up that error handling logic significantly.

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -220,44 +220,27 @@ func (s *Setup) installArtifacts(buildResult *model.BuildResult, artifacts artif
 	if !buildResult.BuildReady && buildResult.BuildEngine == model.Camel {
 		return locale.NewInputError("build_status_in_progress", "", apimodel.ProjectURL(s.target.Owner(), s.target.Name(), s.target.CommitUUID().String()))
 	}
-	// Artifacts are installed in two stages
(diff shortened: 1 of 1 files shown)


pkg/platform/runtime/setup/setup.go
```diff
@@ -220,44 +220,27 @@ func (s *Setup) installArtifacts(buildResult *model.BuildResult, artifacts artif
...
-			artfErrs <- err
+			installErrs <- s.installFromBuildLog(buildResult, artifacts, installErrs)
 		}
 	})
 
-	return <-installErr
+	var errs []error
+	for err := range installErrs {
+		errs = append(errs, err)
+	}
```
mdrohmann, line 238: I think we'll never get here, because the `installErrs` channel will be blocked inside the `Run()` method which then doesn't return.

mdrohmann: You can try it.  It stops forever on the same error, or when all artifacts are installed.

Naatan: Why is it blocked inside the `Run()` method?

Anyway, don't mean to turn your PR into my PR, key point here is I think we can improve the error handling here. The current PR has too much going on in it for my comfort..

mdrohmann (changes requested):
This solution blocks forever unfortunately, as explained below...

PR #1294, opened 2021-03-29
state-svc auto starts with systray

Naatan:
https://www.pivotaltracker.com/story/show/177209361

 https://www.pivotaltracker.com/story/show/177209361

PR #1296, opened 2021-03-30
We install a state-tray start menu shortcut on Windows

Naatan:
Review blocked until https://github.com/ActiveState/cli/pull/1294 is merged.

 https://www.pivotaltracker.com/story/show/177447441


internal/osutils/shortcut/shortcut_windows.go
```diff
@@ -0,0 +1 @@
+package shortcut
```
MDrakos, line 1: It appears that the `shortcut` package is only being used from Windows-specific files. So either this file is no longer necessary or the contents of `shortcut.go` should be here.

internal/runners/prepare/prepare_windows.go
```diff
@@ -0,0 +1,115 @@
...
+		if err := prepareStartShortcut(); err != nil {
+			r.reportError(locale.Tr("err_prepare_shortcut", "Could not create start menu shortcut, error received: {{.V0}}.", err.Error()), err)
+		}
+	}
+
+	return nil
+}
+
+func prepareStartShortcut() error {
+	dir := filepath.Join(os.Getenv("USERPROFILE"), "AppData", "Roaming", "Microsoft", "Windows", "Start Menu", "Programs", "ActiveState")
```
MDrakos, line 41: Nit: you can use `APPDATA` here to get directly to the `Roaming` directory so:
```suggestion
	dir := filepath.Join(os.Getenv("APPDATA"), "Microsoft", "Windows", "Start Menu", "Programs", "ActiveState")
```

Can also use `APPDATA` in `autostart_windows.go`

internal/runners/prepare/prepare_windows.go
```diff
@@ -0,0 +1,115 @@
...
+
+		if err := prepareStartShortcut(); err != nil {
+			r.reportError(locale.Tr("err_prepare_shortcut", "Could not create start menu shortcut, error received: {{.V0}}.", err.Error()), err)
+		}
+	}
+
+	return nil
+}
+
+func prepareStartShortcut() error {
```
MDrakos, line 40: Nit: it may be helpful to have this name mention the systray as that is the shortcut it's preparing (ie. `prepareTrayStartShortcut`

MDrakos (changes requested):

internal/osutils/shortcut/shortcut_windows.go
Naatan: That's just because the current iteration is only Windows focused. The distinction is necessary because the windows specific code won't compile outside of windows. And if the package has no Go files on other platforms then building it will fail outright.

internal/runners/prepare/prepare_windows.go
Naatan: I'm a little concerned with using `APPDATA` as there's some mixed messages around its availability and value, eg. https://stackoverflow.com/questions/37946282/why-does-appdata-in-windows-7-seemingly-points-to-wrong-folder

MDrakos (approved):

PR #1299, opened 2021-04-01
Update deferrer stat error

MDrakos:
https://www.pivotaltracker.com/story/show/177207913

### DIFF
--- a/internal/analytics/deferred.go
+++ b/internal/analytics/deferred.go
@@ -35,14 +35,10 @@ func isDeferralDayAgo(cfg Configurable) bool {
 	df := deferrerFilePath(cfg)
 	stat, err := os.Stat(df)
 	if os.IsNotExist(err) {
-		tErr := fileutils.Touch(deferrerFilePath(cfg))
-		if tErr != nil {
-			logging.Errorf("Failed to create deferrer time stamp file: %v", tErr)
-		}
 		return false
 	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mdrohmann (dismissed):

internal/analytics/deferred.go
```diff
@@ -34,10 +34,18 @@ func deferrerFilePath(cfg Configurable) string {
 func isDeferralDayAgo(cfg Configurable) bool {
 	df := deferrerFilePath(cfg)
 	stat, err := os.Stat(df)
+	if os.IsNotExist(err) {
+		tErr := fileutils.Touch(deferrerFilePath(cfg))
+		if tErr != nil {
+			logging.Errorf("Failed to create deferrer time stamp file: %v", tErr)
+		}
```
Naatan, line 41: I don't think there's any reason we should be creating this file for the purpose of this function.

internal/analytics/deferred.go
```diff
@@ -34,10 +34,18 @@ func deferrerFilePath(cfg Configurable) string {
...
+		return false
+	}
 	if err != nil {
-		logging.Errorf("Could not stat file: %s, error: %v", df, err)
+		logging.Debug("Could not stat deferrer file: %s, error: %v", df, err)
```
Naatan, line 45: With the fix we have the issue at hand should be addressed. I'd still want to hear about these errors if they occur for some other reason.

Naatan (changes requested):

Naatan (approved):

PR #1301, opened 2021-04-02
Resolve symlinks when identifying configured SHELL

mdrohmann:
At least in Ubuntu, it is configurable what `/bin/sh` can point to.  By default it is `/bin/dash` (a POSIX compatible shell) which is a subset of `bash`.  (Maybe we should support it, or ensure that our bash scripts run with just POSIX commands.

I also changed the logging level from Error to Debug, as the message does not give us much actionable information.  It is just interesting to know what shells people are using.

 https://www.pivotaltracker.com/story/show/177207884

### DIFF
--- a/internal/subshell/subshell.go
+++ b/internal/subshell/subshell.go
@@ -2,6 +2,7 @@ package subshell
 
 import (
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -9,6 +10,7 @@ import (
 	"github.com/thoas/go-funk"
 
 	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/output"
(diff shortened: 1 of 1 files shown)


internal/subshell/subshell.go
```diff
@@ -84,10 +86,20 @@ func New() SubShell {
...
 	}
 
-	logging.Debug("Detected SHELL: %s", binary)
+	// try to find the binary on the PATH
+	binaryPath, err := exec.LookPath(binary)
+	if err == nil {
+		// if we found it, resolve all symlinks, for many Linux distributions the SHELL is "sh" but symlinked to a different default shell like bash or zsh
+		resolved, err := fileutils.ResolvePath(binaryPath)
+		if err != nil {
+			logging.Debug("Failed to resolve path to shell binary %s", binaryPath)
```
MDrakos, line 95: Nit: We're losing the error here and it could be helpful to have this in the logs

MDrakos (approved):

MDrakos (dismissed):

internal/subshell/subshell.go
```diff
@@ -108,7 +120,7 @@ func New() SubShell {
 	case "cmd":
 		subs = &cmd.SubShell{}
 	default:
-		logging.Errorf("Unsupported shell: %s, defaulting to OS default.", name)
+		logging.Debug("Unsupported shell: %s, defaulting to OS default.", name)
```
mdrohmann, line 123: I wonder if it is worthwhile creating an analytics event for this one.  So, we can find out what the most prevalent shells are that we are not supporting...

Naatan: Good idea! Please file a ticket :)

Naatan:
> Maybe we should support it

Sounds like a good idea, especially if it's the default on Ubuntu. Please file a ticket :)

internal/subshell/subshell.go
```diff
@@ -84,10 +86,20 @@ func New() SubShell {
...
-	logging.Debug("Detected SHELL: %s", binary)
+	// try to find the binary on the PATH
+	binaryPath, err := exec.LookPath(binary)
+	if err == nil {
+		// if we found it, resolve all symlinks, for many Linux distributions the SHELL is "sh" but symlinked to a different default shell like bash or zsh
+		resolved, err := fileutils.ResolvePath(binaryPath)
+		if err != nil {
+			logging.Debug("Failed to resolve path to shell binary %s: %v", binaryPath, err)
+		}
+		binary = resolved
```
Naatan, line 97: This should probably be in an `else`, since the value of `resolved` is not guaranteed if `err == nil`.

Naatan (changes requested):

internal/subshell/subshell.go
mdrohmann: Good point.

internal/subshell/subshell.go
mdrohmann: https://www.pivotaltracker.com/story/show/177629477

Naatan (approved):

PR #1303, opened 2021-04-05
Some minor bugfixes for the runtime installation

mdrohmann:
This PR contains some minor bugfixes that I ran into while testing the new runtime branch on Friday:

1. When a build fails, we were not properly forwarding the error message from the solver (which is designed to be helpful.)
2. The InstallationStarted event was sent with more than once during an installation
3. The build progress bar needs to count to the (number of expected artifacts + 1), because we are also waiting for the terminal artifact build to finish.

@Naatan Should I create bug reports on PT for this PR?  And is it okay to keep all of them in one PR, or should I split them up?

https://www.pivotaltracker.com/story/show/177631360
https://www.pivotaltracker.com/story/show/177631379

### DIFF
--- a/pkg/platform/runtime/setup/buildlog/buildlog.go
+++ b/pkg/platform/runtime/setup/buildlog/buildlog.go
@@ -58,7 +58,9 @@ func New(artifactMap map[artifact.ArtifactID]artifact.ArtifactRecipe, conn Build
 	errCh := make(chan error)
 
 	total := len(artifactMap)
-	events.BuildStarting(total)
+	// We set the total to one extra artifact, as we always have to wait for
+	// the terminal artifact to finish before the build is actually done.
+	events.BuildStarting(total + 1)
 
 	go func() {
 		defer close(ch)
--- a/pkg/platform/runtime/setup/buildlog/buildlog.go
+++ b/pkg/platform/runtime/setup/buildlog/buildlog.go
@@ -58,9 +58,7 @@ func New(artifactMap map[artifact.ArtifactID]artifact.ArtifactRecipe, conn Build
 	errCh := make(chan error)
 
 	total := len(artifactMap)
-	// We set the total to one extra artifact, as we always have to wait for
-	// the terminal artifact to finish before the build is actually done.
-	events.BuildStarting(total + 1)
+	events.BuildStarting(total)
 
 	go func() {
 		defer close(ch)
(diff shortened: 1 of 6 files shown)


Naatan:
> @Naatan Should I create bug reports on PT for this PR? And is it okay to keep all of them in one PR, or should I split them up?

Let's create just one so we can keep track of the PR in case it runs longer. Generally speaking I would favor we file individual bugs before we start working on them.

pkg/platform/runtime/setup/buildlog/buildlog.go
```diff
@@ -58,7 +58,9 @@ func New(artifactMap map[artifact.ArtifactID]artifact.ArtifactRecipe, conn Build
 	errCh := make(chan error)
 
 	total := len(artifactMap)
-	events.BuildStarting(total)
+	// We set the total to one extra artifact, as we always have to wait for
+	// the terminal artifact to finish before the build is actually done.
+	events.BuildStarting(total + 1)
```
Naatan, line 63: This feels error prone, is there no way we can get an accurate count via the API, without making assumptions?

Naatan (changes requested):

pkg/platform/runtime/setup/buildlog/buildlog.go
mdrohmann: That is a very good point!

The issue that I wanted to fix was: I encountered a build, where the installation was done, the build progress bar *indicated* that it was done, but the terminal artifact was still building, and so the buildlog instance was still waiting for it - the installation just halted.

I can think of two better ways to handle this:
1. We use the total number of artifacts that need to be built from the head-chef's `BuildStatusResponse` (not from the Recipe as we do now) which has all the artifact IDs that are actually being built.  We would not be able to send the `artifactName` with every build event anymore, but our current progressbar implementation is not doing anything with that information.
2. We update the progressbar such that it prints "Waiting for build to finish" (or something like that) after all expected artifacts have been built, but before the `BuildFinished` event has been received.

I personally favor Option 1, as it seems more honest, and is also much simpler to implement. Also: If we really ever want to display the names of currently building artifacts, there are other ways to achieve that.

But I should probably take this part out of this PR, and file a follow-up story...?

mdrohmann: @Naatan ^

Naatan: Yeah let's take it out. Or maybe address it with option 2 for now and we'll iterate with option 1 later. I do think we should show the names of the artifacts that are building, but that can be part of the iteration.

mdrohmann: Just to clarify:  Currently our progressbar is *not* printing the names of artifacts that are *building*; we do that only for the *download*, *unpack* and *install* steps.

I'll take it out, because Option 2 is actually difficult to implement with the `mpb.Progress` package.

mdrohmann: https://www.pivotaltracker.com/story/show/177631597

Naatan (approved):

PR #1304, opened 2021-04-07
Uninstall fix

MDrakos:
It appears that with newer versions of `python` simply having a directory in `site_packages` will allow you to import it. In the case of `rich` all of the contents of the directory were removed (except for a `__pycache__` directory) on `state uninstall`. You could still import this module but it would be unusable.

```
>>> import rich
>>> print(rich.__file__)
None
```

https://www.pivotaltracker.com/story/show/177550106

### DIFF
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -442,6 +442,21 @@ func IsEmptyDir(path string) (bool, error) {
 	return (len(files) == 0), nil
 }
 
+// IsEmptyDirs returns true if the directory at the provide path has no files
+// within it or any of its sub directories
+func IsEmptyDirs(path string) bool {
+	empty := true
+	filepath.Walk(path, func(p string, _ os.FileInfo, _ error) error {
+		if !IsDir(p) {
+			empty = false
+		}
+
+		return nil
+	})
+
+	return empty
+}
+
 // MoveAllFilesCallback is invoked for every file that we move
 type MoveAllFilesCallback func(fromPath, toPath string)
 
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -442,21 +442,6 @@ func IsEmptyDir(path string) (bool, error) {
 	return (len(files) == 0), nil
 }
 
-// IsEmptyDirs returns true if the directory at the provide path has no files
-// within it or any of its sub directories
-func IsEmptyDirs(path string) bool {
-	empty := true
-	filepath.Walk(path, func(p string, _ os.FileInfo, _ error) error {
-		if !IsDir(p) {
-			empty = false
-		}
-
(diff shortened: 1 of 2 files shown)


internal/fileutils/fileutils.go
```diff
@@ -442,6 +442,21 @@ func IsEmptyDir(path string) (bool, error) {
 	return (len(files) == 0), nil
 }
 
+// IsEmptyDirs returns true if the directory at the provide path has no files
+// within it or any of its sub directories
+func IsEmptyDirs(path string) bool {
+	empty := true
+	filepath.Walk(path, func(p string, _ os.FileInfo, _ error) error {
+		if !IsDir(p) {
```
mdrohmann, line 450: I think, even if `p` is a directory, we should set `empty = false`.  But also: the callback can just always *return* `true`.  Once it is called once for a file in that directory, we know that it is not empty.

pkg/platform/runtime/alternative.go
```diff
@@ -216,6 +217,18 @@ func (ai *AlternativeInstall) PreInstall() error {
 				return locale.WrapError(err, "err_rm_artf", "", "Could not remove old package file at {{.V0}}.", file)
 			}
 		}
+
+		for _, dir := range v.Dirs {
```
mdrohmann, line 221: We probably want to iterate over the directories in reverse alphabetical order, such that you look `a/b/c` before `a/b`.  Otherwise you might only remove `the first one, but not the latter one.

pkg/platform/runtime/alternative.go
```diff
@@ -216,6 +217,18 @@ func (ai *AlternativeInstall) PreInstall() error {
...
+			if !fileutils.DirExists(dir) {
+				continue
+			}
+			if fileutils.IsEmptyDirs(dir) {
```
mdrohmann, line 225: I still think it would be good to add a check for whether that `dir` is also in any of the artifacts that we have to keep.  But as that is such an edge case, it would probably also be okay to put this in a follow-up story.

mdrohmann (changes requested):

pkg/platform/runtime/alternative.go
MDrakos: Nice! `IsEmptyDirs` was to address this but sorting also works.

internal/fileutils/fileutils.go
MDrakos: This function is no longer needed due to your suggestion below 👍

pkg/platform/runtime/alternative.go
MDrakos: This is easy enough to put in this PR

mdrohmann (approved):
:+1:

Naatan (approved):

PR #1306, opened 2021-04-09
Update icon files

daved:
https://www.pivotaltracker.com/story/show/177585280


daved:
The SVG is saved exactly as it was provided. The PNG 64x64 and compressed using `pngcrush`. The ICO contains the sizes 16,24,32,48,64,128.

Naatan (approved):

PR #1307, opened 2021-04-12
Update changelog

MDrakos:
https://www.pivotaltracker.com/story/show/177550106


Naatan (changes requested):
version.txt says 0.27.0

Naatan (approved):
I meant the entry in the changelog was wrong :p We haven't released a 0.27.0, so there wouldn't be a 0.27.1.

PR #1308, opened 2021-04-13
Deploy/activate uses executor

Naatan:
https://www.pivotaltracker.com/story/show/177635510


test/integration/deploy_int_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	executor2 "github.com/ActiveState/cli/pkg/platform/runtime/executor"
```
mdrohmann, line 12: The `executor2` should probably be removed now...

mdrohmann (changes requested):
Looks good.  Just very minor things that I found.

mdrohmann (dismissed):

internal/appinfo/appinfo.go
```diff
@@ -17,14 +17,14 @@ type AppInfo struct {
 func TrayApp() *AppInfo {
 	return &AppInfo{
 		constants.TrayAppName,
-		absolute(filepath.Join(filepath.Dir(os.Args[0]), "state-tray") + osutils.ExeExt),
+		exePath(filepath.Join(filepath.Dir(os.Args[0]), "state-tray") + osutils.ExeExt),
```
mdrohmann: I noticed that this needs to look like this now:

 ```
exePath("state-tray")
```

File: internal/appinfo/appinfo.go [hunk 3]
```diff
@@ -17,14 +17,14 @@ type AppInfo struct {
...
 	}
 }
 
 func StateApp() *AppInfo {
 	return &AppInfo{
 		constants.StateAppName,
-		absolute(filepath.Join(filepath.Dir(os.Args[0]), "state") + osutils.ExeExt),
+		exePath(filepath.Join(filepath.Dir(os.Args[0]), "state") + osutils.ExeExt),
```
mdrohmann: same here:
```
exePath("state")
```

mdrohmann (commented):
I think it is this problem!

internal/appinfo/appinfo.go
Naatan: Thanks, just noticed that also -.-

PR #1310, opened 2021-04-13
Systray is runnable as a MacOS app

MDrakos:
https://www.pivotaltracker.com/story/show/177447481


cmd/state-tray/internal/startup/startup.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/fileutils"
+)
+
+func StartStateService() error {
+	svcExePath, err := getStateServicePath()
```
Naatan, line 13: Use the `appinfo` package for this.

cmd/state-tray/internal/startup/startup_win.go
```diff
@@ -0,0 +1,14 @@
...
+package startup
+
+import (
+	"os/exec"
+	"syscall"
+)
+
+func newStartServiceCommand(path string) *exec.Cmd {
+	cmd := exec.Command(path, "start")
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: 0x08000000} // CREATE_NO_WINDOW
```
Naatan, line 12: Check with Martin, I think he has a better solution to this problem in his PR.

internal/constants/constants.go
```diff
@@ -276,5 +276,8 @@ const SvcConfigPid = "svc-pid"
 // TrayAppName is the name we give our systray application
 const TrayAppName = "ActiveState Platform"
 
+// LaunchFileMacOS is the .plist file that holds the auto launch information on MacOS
+const LaunchFileMacOS = "com.activestate.platform.state-tray.plist"
```
Naatan, line 280: Just add this const to the package it's used in, since it's only used once and the value in question isn't something common that requires centralization.

internal/osutils/autostart/autostart_mac.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+func (a *App) Enable() error {
+	if a.IsEnabled() {
+		return nil
+	}
+
+	path, err := launchFilePath()
+	if err != nil {
+		return err
```
Naatan, line 24: Wrap

internal/osutils/autostart/autostart_mac.go
```diff
@@ -0,0 +1,60 @@
...
+	return nil
+}
+
+func (a *App) Disable() error {
+	if !a.IsEnabled() {
+		return nil
+	}
+	path, err := launchFilePath()
+	if err != nil {
+		return err
```
Naatan, line 41: Wrap

internal/osutils/autostart/autostart_mac.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	return os.Remove(path)
+}
+
+func (a *App) IsEnabled() bool {
+	path, err := launchFilePath()
+	if err != nil {
+		logging.Error("Could not get launch file: %v", err)
```
Naatan, line 49: No reason this function can't return an error.

Naatan (changes requested):

internal/osutils/autostart/autostart_mac.go
MDrakos: Okay, wasn't sure about that change as it would change the signature on all of the other platforms.

cmd/state-tray/main.go
```diff
@@ -84,7 +90,11 @@ func run() error {
 
 	systray.AddSeparator()
 	mAutoStart := systray.AddMenuItem(locale.Tl("tray_autostart", "Start on Login"), "")
-	if autostart.New().IsEnabled() {
+	enabled, err := autostart.New().IsEnabled()
```
Naatan, line 93: You're using an instance of autostart below on line 154, should those be using the same instance?

Naatan (dismissed):

cmd/state-tray/main.go
MDrakos: Yes, we can use the same instance there. 👍

Naatan (approved):

PR #1311, opened 2021-04-13
Generate multi-file updates

mdrohmann:
https://www.pivotaltracker.com/story/show/177210769

Once, this PR merges to master, we should have install scripts that can download our new updates.  For this reason, it also implements https://www.pivotaltracker.com/story/show/177785111.  This would have broken the installation of release updates otherwise.

I noticed that `state-svc` does not always automatically start when it is required by the `state update` call: https://www.pivotaltracker.com/story/show/177865635

Also: I removed the defunct confirmation for changing the environment in the new install.sh script.  A story explaining the details is here: https://www.pivotaltracker.com/story/show/177865692


.github/workflows-src/steps.lib.yml
```diff
@@ -1,20 +1,64 @@
...
-run: state run build
+run: state run build-svc
+#@ end
+---
+#@ def steps_build_tray_linux():
+name: Build Tray App (linux)
+shell: bash
+if: runner.os == 'Linux'
+run: |
+  sudo apt-get install libgtk-3-dev libappindicator3-dev -y
```
Naatan, line 18: We should probably change our target ubuntu version to the latest LTS. At the moment it's just using "latest", which might not actually be compatible with the latest LTS.

activestate.generators.yaml
```diff
@@ -65,23 +59,37 @@ scripts:
       $constants.SET_ENV
 
       export GOARCH=${1:-amd64}
-      export TARGET_SUFFIX=${1:-}
-      mkdir -p build/update
-      go run scripts/update-generator/main.go -o build/update $constants.BUILD_TARGET${TARGET_SUFFIX}
-  - name: generate-update-32bit
+      mkdir -p ./build/update
+      go run scripts/update-generator/main.go -o ./build/update $BUILD_TARGET_DIR/$constants.BUILD_INSTALLER_TARGET $BUILD_TARGET_DIR/$constants.BUILD_DAEMON_TARGET $BUILD_TARGET_DIR/$constants.BUILD_SYSTRAY_TARGET $BUILD_TARGET_DIR/$constants.BUILD_TARGET
```
Naatan, line 63: Can you add some `\\n` separators here so it's easier to grok and update.

activestate.windows.yaml
```diff
@@ -24,7 +21,14 @@ scripts:
     if: eq .OS.Name "Windows"
     description: Copies install scripts to build dir, this script exists for transparency with CI behavior
     value: |
-      cp installers/install.ps1 build/install.ps1
+
+      BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+      INSTALLERS_DIR=build/installers/${BRANCH#release}
+      mkdir -p $INSTALLERS_DIR
```
Naatan, line 28: We'd still want this to go to the old location if it's the `release` branch.

internal/constants/constants.go
```diff
@@ -297,3 +297,7 @@ const SvcAppName = "State Service"
 
 // StateAppName is the name we give our state cli executable
 const StateAppName = "State Tool"
+
+// ToplevelInstallArchiveDir is the top-level directory for files in an installation archive
+// Cf., https://www.pivotaltracker.com/story/show/177781411
```
Naatan, line 302: +1

Naatan (changes requested):
Looks mostly ok to me except for one minor issue (install.sh|ps1 still needing to go to the old location on the release channel).

Note though that I did not evaluate all the code in detail, there's a lot going on here.

activestate.windows.yaml
mdrohmann: Yes, That's why i have this `${BRANCH#release}` part, so in the `release` case, it is copied to `./build/installers/install.sh|install.ps1` and from there to the same location where it used to go `s3://cli-update/update/state/install.sh|install.ps1`

mdrohmann: At least I hope that this will happen, we should confirm after the first deployment to `master` that the scripts are uploaded to `s3://cli-update/update/state/master/`.  After that is clear, I am confident, the release install script will end up in the correct place.

Naatan (approved):

mdrohmann:
@Naatan Can you force-merge this PR please.  It is failing on the TestCarlisle integration test that is still broken on Windows since we added the executor "fix".

PR #1313, opened 2021-04-15
Multi-file installer mechanism

mdrohmann:
https://www.pivotaltracker.com/story/show/177209528


internal/runners/update/update.go
```diff
@@ -43,46 +47,23 @@ func (u *Update) Run(params *Params) error {
...
 
-	if err = up.Run(u.out, false); err != nil {
-		if os.IsPermission(errs.InnerError(err)) {
-			return locale.WrapError(err, "err_update_failed_due_to_permissions", "Update failed due to permission error.  You may have to re-run the command as a privileged user.")
-		}
-		return locale.WrapError(err, "err_update_failed", "Update failed, please try again later or try reinstalling the State Tool.")
-	}
-
-	u.out.Print(locale.Tl("version_updated", "Version updated to {{.V0}}@{{.V1}}", channel, info.Version))
+	u.out.Print(locale.Tl("version_updating_deferred", "Version updating to {{.V0}}@{{.V1}} in the background.\nRefer to logfile {{.V2}} for progress.", up.Channel, up.Version, up.Logfile))
```
Naatan, line 63:
```suggestion
	u.out.Print(locale.Tl("version_updating_deferred", "Version update to {{.V0}}@{{.V1}} has started and should complete in seconds.\nRefer to log file [ACTIONABLE]{{.V2}}[/RESET] for progress.", up.Channel, up.Version, up.Logfile))
```

internal/subshell/subshell.go
```diff
@@ -142,3 +134,35 @@ func New() SubShell {
 
 	return subs
 }
+
+func DetectShellBinary(cfg *config.Instance) (binary string) {
+	defer func() {
```
Naatan, line 139: This is going to cause config writes even when we received the shell from the config in the first place.

internal/testhelpers/installation/testcmd/main.go
```diff
@@ -0,0 +1,63 @@
+package main
```
Naatan, line 1: I'm not a fan of these, this is too much to facilitate a unit test. Just cover it with an integration test, or test the units you're actually interested in.

Naatan (changes requested):
I'm not seeing the auto-forwarding to older versions of the state tool which we discussed we'll still need for the initial version?

internal/runners/update/lock.go
```diff
@@ -34,70 +29,12 @@ type LockParams struct {
...
-	if err != nil {
-		return err
-	}
-
-	if !confirmed {
-		return locale.NewInputError("err_update_lock_noconfirm", "Cancelling by your request.")
-	}
-
-	return nil
+	return locale.NewInputError("locking_unsupported", "This version of the State Tool does not support version locking anymore.")
```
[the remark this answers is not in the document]
mdrohmann: Yes, I will re-introduce the functionality.

cmd/state-installer/installer.go
```diff
@@ -0,0 +1,156 @@
...
+		if exitCode != 0 {
+			return errs.New("Stopping %s exited with code %d", svcInfo.Name(), exitCode)
+		}
+	}
+
+	tmpDir := filepath.Dir(exe)
+	// clean-up temp directory when we are done.
+	defer os.RemoveAll(tmpDir)
+
+	inst, err := installation.New(filepath.Join(tmpDir, "bin"), installPath)
```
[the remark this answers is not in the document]
mdrohmann: Well, that's exactly the package that code was in on Thursday morning, until I decided that it could go here.  But, thinking about it, I agree with you...

internal/fileutils/fileutils.go
```diff
@@ -508,6 +508,61 @@ func MoveAllFilesRecursively(fromPath, toPath string, cb MoveAllFilesCallback) e
...
+			tmpToPath := fmt.Sprintf("%s.new", toPath)
+			err := CopyFile(fromPath, tmpToPath)
+			if err != nil {
+				return errs.Wrap(err, "failed to copy %s -> %s", fromPath, tmpToPath)
+			}
+			err = os.Chmod(tmpToPath, fileInfo.Mode())
+			if err != nil {
+				return errs.Wrap(err, "failed to set file permissions for %s", tmpToPath)
+			}
+			err = os.Rename(tmpToPath, toPath)
```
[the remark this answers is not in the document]
mdrohmann: Yes, it does.  But in our installation procedure we would have backed up the original files before.  But I think I should add a unit test for this case, as we are depending on that functionality (at least on Windows).

internal/testhelpers/installation/testcmd/main.go
mdrohmann: I think I can now remove these test scripts and instead move the test cases to the `update_int_test.go` file, but I will do that in #1311, as they depend on functioning update files.

mdrohmann: These tests were super helpful though, to stay confident about what I needed without having to have a full update yet...  And it will be easy to transfer them now...

cmd/state-installer/installer.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/machineid"
+	"github.com/ActiveState/cli/internal/osutils"
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/internal/subshell"
+	"github.com/ActiveState/cli/internal/subshell/sscommon"
+)
+
+func main() {
+	exitCode := run()
```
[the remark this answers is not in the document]
mdrohmann: As Mike pointed out, an `os.Exit(1)` prevents the execution of deferred functions (which in this case would at least be the `defer rollbar.Close()` function.  So, I will add two layers of `run` functions, the outer one of which will just do the `os.Exit()`.

cmd/state-installer/installer.go
```diff
@@ -24,63 +26,58 @@ import (
...
-	exitCode := run()
-	if exitCode != 0 {
-		os.Exit(exitCode)
+	if !_run() {
+		os.Exit(1)
 	}
 }
 
-func run() int {
+func _run() bool {
```
Naatan: This feels really dirty just to facilitate a `defer`, which we don't HAVE to use. It's there for your convenience, so if it's not convenient just don't use it.

I think what you want is for the main function to call rollbar.Close() before you os.Exit(), either by putting all that in a central exit function or just by actually calling `rollbar.Close()` right before the exit. 

I'm not married to that suggestion, I just want a solution that doesn't make you scratch your head when you look at it.

cmd/state-installer/installer.go
```diff
@@ -24,63 +26,58 @@ import (
...
 	logging.SetupRollbar(constants.StateInstallerRollbarToken)
 	defer rollbar.Close()
 
+	out := mustOutputer()
+	err := run(out)
+	if err != nil {
+		errMsg := fmt.Sprintf("%s failed with error: %s", filepath.Base(os.Args[0]), errs.Join(err, ": "))
+		logging.Error(errMsg)
+		out.Error(errMsg)
+		out.Print(fmt.Sprintf("To retry run %s", strings.Join(os.Args, " ")))
```
Naatan, line 47:
```suggestion
		out.Error(fmt.Sprintf("To retry run %s", strings.Join(os.Args, " ")))
```

cmd/state-installer/installer.go
```diff
@@ -135,11 +130,13 @@ func install(installPath string, cfg *config.Instance, out output.Outputer) erro
...
 
-	rcFile, err := shell.RcFile()
-	if err == nil {
-		out.Notice(fmt.Sprintf("Please either run 'source %s' or start a new login shell in order to start using the State Tool executable.", rcFile))
-	} else {
-		out.Notice("Please start a new login shell in order to start using the State Tool executable.")
+	if !funk.Contains(strings.Split(os.Getenv("PATH"), string(os.PathListSeparator)), installPath) {
+		rcFile, err := shell.RcFile()
+		if err == nil {
+			out.Notice(fmt.Sprintf("Please either run 'source %s' or start a new login shell in order to start using the State Tool executable.", rcFile))
```
Naatan, line 136:
```suggestion
			out.Notice("Please start a new shell in order to start using the State Tool executable.")
```

Let's stick with telling them to restart their shell. This is assuming everyone has their rcfile set up in a way that you can re-source it without adverse effects.

cmd/state-installer/installer.go
```diff
@@ -154,3 +151,17 @@ func install(installPath string, cfg *config.Instance, out output.Outputer) erro
...
+		OutWriter:   os.Stdout,
+		ErrWriter:   os.Stderr,
+		Colored:     true,
+		Interactive: false,
+	})
+	if err != nil {
+		logging.Error("This shouldn't happen.  Err should never be != nil.")
+	}
+	return out
+}
```
Naatan, line 167: This makes unnecessary assumptions about how `output` works. If `output.New` errors then the program errors.

cmd/state-installer/internal/installer/installation.go
```diff
@@ -22,7 +23,17 @@ func backupFiles(targetFiles []string) ([]string, error) {
...
 		if fileutils.TargetExists(t) {
 			newName := fmt.Sprintf("%s.bak", t)
-			err := os.Rename(t, newName)
+			if fileutils.TargetExists(newName) {
+				_ = os.Remove(newName)
+			}
+			var err error
+			if runtime.GOOS == "windows" {
+				// Note: We could also rename the files here, in which case the installation would always pass on Windows (even if the executable is still running), but we would not be able to remove the .bak file after the installation (if the executable is still running).
+				// We opted for this solution, as it is more compliant with other Windows installations, where the installation requires all existing programmes to be stopped prior to updates.
```
Naatan, line 32: I'd rather we try our best effort to make things work consistently across platforms, which means that if we can make it work while the executable is running we should.

People being able to use tooling without issues is more important than a few stale backup files that we can clean up on a later call.

Naatan (changes requested):

Naatan (changes requested):
Looks good! Please just verify though that if the backup files failed to remove the next update won't fail because they still exist (ie. it should just overwrite them).

Naatan (dismissed):

mdrohmann:
> Looks good! Please just verify though that if the backup files failed to remove the next update won't fail because they still exist (ie. it should just overwrite them).

I thought about that.  There is a possibility that some-one is running the State Tool for a very, very long time.  That is maybe more likely since we have our executors now.  So, it could be eg., some Python programme that runs as a server and is wrapped by the state tool at version A.
Then, we update that State Tool to version B, but end up with a backup pointing to version A (we could not remove it because of the running python server).
If now, the State Tool is again updated to version C and the python server is *still* running, we would fail at the step where we try to backup the files (ie., overwrite `state.bak`) with an AccessDenied error.

This is super unlikely (also because Windows forces you to restart so often), but absolutely possible.

Of course, we could make our backup system more intelligent, and move files to `.bak.x` with `x` a number that increments until it works, but I would recommend that for a follow-up story...

Naatan:
> There is a possibility that some-one is running the State Tool for a very, very long time

I was talking about the more likely scenario of the removal failing cause windows is windows and then a few days later needing to update and the update failing simply because those files exist (not because they're in use).

Long as the update install doesn't fail simply because the backup files exist I think we're good.

cmd/state-installer/internal/installer/installation_test.go
```diff
@@ -0,0 +1,139 @@
...
+	tests := []struct {
+		Name                      string
+		SimulateAdminInstallation bool
+		ExpectSuccess             bool
+	}{
+		{"successful", false, true},
+		{
+			"update-without-permissions",
+			true,
+			// On Windows, the installation will succeed even if the existing files have no write permissions, because we are allowed to rename them and delete them (we are just not allowed to overwrite them!).
```
mdrohmann, line 99: Just wanted to bring this to your attention.  That changed after I changed the installation behavior for Windows (as we are now removing the existing files first)

.github/workflows/build.yml
```diff
@@ -261,7 +261,7 @@ jobs:
       run: |
         export SHELL=''
         set +e
-        go test -timeout 10m github.com/ActiveState/cli/test/integration -test.v -test.run "^TestLanguageMsiActivePerl|TestDeployIntegrationTestSuite|TestInitIntegrationTestSuite|TestPushIntegrationTestSuite|TestUpdateIntegrationTestSuite|PrepareIntegrationTestSuite|InstallScriptsIntegrationTestSuite|ShowIntegrationTestSuite$"
```
mdrohmann, line 264: For this PR I will remove the UpdateIntegrationTests from the critical Path.  They are re-introduced in PR #1311 .

cmd/state-installer/internal/installer/installation_test.go
Naatan: Lol, oh windows. I guess this works to our advantage.

Naatan (approved):

mdrohmann:
As mentioned on slack, two critical intergration tests are failing, but it has become clear that this has nothing to do with this PR anymore...

PR #1314, opened 2021-04-16
Uninstall state tray and state service

MDrakos:
https://www.pivotaltracker.com/story/show/177448021


cmd/state-tray/main.go
```diff
@@ -49,20 +51,32 @@ func run() error {
...
+		return errs.Wrap(err, "Could not save pid")
+	}
+
+	stateSvcExe := filepath.Join(filepath.Dir(os.Args[0]), "state-svc")
+	if runtime.GOOS == "windows" {
+		stateSvcExe = stateSvcExe + ".exe"
+	}
+	if !fileutils.FileExists(stateSvcExe) {
+		return errs.New("Could not find: %s", stateSvcExe)
+	}
```
Naatan, line 70: This is already being done (properly - use `appinfo`) below on line 72.

cmd/state-tray/pkg/autostart/autostart.go
```diff
@@ -5,13 +5,14 @@ import (
...
 )
 
-func New() *autostart.App {
+func New(cfg *config.Instance) *autostart.App {
 	suffix := ""
 	if runtime.GOOS == "windows" {
 		suffix = ".exe"
 	}
-	return autostart.New("activestate-desktop", filepath.Join(filepath.Dir(os.Args[0]), "state-tray"+suffix))
+	return autostart.New("activestate-desktop", filepath.Join(filepath.Dir(os.Args[0]), "state-tray"+suffix), cfg)
```
Naatan, line 17: I realize this isn't part of this PR but you might as well update this to `appinfo.TrayApp().Exec()`

internal/constants/constants.go
```diff
@@ -287,3 +290,9 @@ const SvcAppName = "State Service"
 
 // StateAppName is the name we give our state cli executable
 const StateAppName = "State Tool"
+
+// AutoStartPath is the config key that holds the path to the auto start file
+const AutoStartPath = "auto_start_path"
```
Naatan, line 295: atm you're using `AutoStartPath` in platform specific logic to set up the autostart logic, but then in the destroy logic you just assume all platforms use this entry. This doesn't seem necessary, just have create and destroy functions that each platform can implement as they see fit.

internal/constants/constants.go
```diff
@@ -287,3 +290,9 @@ const SvcAppName = "State Service"
...
+
+// InstallPath is the config key that holds the path to the installation dir
+const InstallPath = "install_path"
```
Naatan, line 298: This is being covered in Martin's PR for the new update mechanism. There'll be a new package called `installation` that provides this information.

Naatan (changes requested):
Started reviewing but this has way too much going in for a simple stopgap PR that's meant to fix glaring uninstall issues long enough for us to get to the proper uninstall mechanism.

When I said "logic doesn't have to be high grade" I should have also said "but centralize the low grade code so it's easily removed". This is just sprinkling too much workaround type code all over the place.

Please revisit and implement this in a way that it's fully centralized. This shouldn't have any changes to constants.go or things like that because you're basically implementing high level systems to address a hacky solution that's meant to just carry us over until we have the proper uninstall system.

I'd expect to see maybe a new file inside the `clean uninstall` runner which stops the state-tray and then runs the uninstall logic.

Note for stopping the systray there's this in Martin's PR:

https://github.com/ActiveState/cli/pull/1313/files#diff-88e28b2772b130951314953db95d53a1f0e46819962485beaea32ceb9b8d3f4dR11

PR #1315, opened 2021-04-20
Update MacOS installation directory

MDrakos:
https://www.pivotaltracker.com/story/show/177749041


cmd/state-installer/internal/installer/installation_darwin.go
```diff
@@ -2,7 +2,58 @@
...
+				return errs.Wrap(err, "Could not write icon file")
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return errs.Wrap(err, "Could not create application directory")
+	}
+
+	trayInfo := appinfo.TrayApp()
```
mdrohmann, line 38: So, the tray app is the only one, we need to symlink now?  That is good...

cmd/state-installer/internal/installer/installation_darwin.go
```diff
@@ -2,7 +2,58 @@
...
+)
+
+const appDir = "/Applications"
+
 // InstallSystemFiles installs files in the /Application directory
 func InstallSystemFiles(installDir string) error {
+	box := packr.NewBox("../../../../assets/macOS/app")
+	err := box.Walk(func(path string, _ packr.File) error {
+		if fileutils.IsDir(path) {
+			err := fileutils.Mkdir(filepath.Join(appDir, path))
```
mdrohmann, line 22: I think in an update, this would error out, because the directory already exists.  We somehow need to make this function able to run twice without error. The easiest way is probably to just remove the entire target directory "ActiveState Desktop.app" at the beginning.

Of course it would be nice to also have a rollback function for this, but that is probably better done in a follow-up story, as my base PRs haven't landed yet.  And it will be easier once that has happened.

mdrohmann (changes requested):
I realized that the function needs to be idempotent.  So, we should clean up the app dir before we write files in there.

cmd/state-installer/internal/installer/installation_darwin.go
MDrakos: Yes, `appinfo` takes care of that so we only need to like the `state-tray` binary

cmd/state-installer/internal/installer/installation_darwin.go
MDrakos: Actually, this function will just return `nil` if the directory already exists. But I think it's better to just remove the directory and all of its contents rather than overwriting everything so I've applied this suggestion.

Is there a follow-up story for rollback functionality in our update/install mechanism already? If not I can create one.

mdrohmann (dismissed):

cmd/state-installer/internal/installer/installation_darwin.go
MDrakos: Ticket is here: https://www.pivotaltracker.com/story/show/177873388

cmd/state-installer/internal/installer/installation_darwin.go
```diff
@@ -2,7 +2,66 @@
...
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return errs.Wrap(err, "Could not create application directory")
+	}
+
+	trayInfo := appinfo.TrayApp()
+	err = createNewSymlink(filepath.Join(installDir, filepath.Base(trayInfo.Exec())), filepath.Join(appDir, appName, "Contents", "MacOS", filepath.Base(trayInfo.Exec())))
```
Naatan, line 47: These paths need variables so this is more readable. There's way too much happening in these args.

cmd/state-installer/internal/installer/installation_darwin.go
```diff
@@ -2,7 +2,66 @@
...
+	if err != nil {
+		return errs.Wrap(err, "Could not create state-tray symlink")
+	}
+
+	return nil
+}
+
+func createNewSymlink(target, filename string) error {
+	if fileutils.FileExists(filename) {
```
Naatan, line 56: This should use `TargetExists`. Symlinks aren't technically "files" (ie. this won't be true if the `filename` is a symlink).

cmd/state-installer/internal/installer/installation_darwin.go
```diff
@@ -2,7 +2,66 @@
...
+)
+
 // InstallSystemFiles installs files in the /Application directory
 func InstallSystemFiles(installDir string) error {
+	err := os.RemoveAll(filepath.Join(appDir, appName))
+	if err != nil {
+		return errs.Wrap(err, "Could not remove old app directory")
+	}
+
+	box := packr.NewBox("../../../../assets/macOS/app")
```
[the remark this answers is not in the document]
mdrohmann: Mike and I discussed that a little.  I initially liked the idea to *only* ship the executables in the archive.  On the other hand, we already noticed that we'll need a customary rollback mechanism for this.

So, thinking this through again, it is probably better if we stick with one system as @Naatan suggested.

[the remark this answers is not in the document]
Naatan: Ultimately I'll defer to you on this @mdrohmann since you're owning the update process and I can see this significantly delaying the PR. Could potentially be a later iteration also. But yeah I do think it would be better if this wasn't embedded in the binary.

cmd/state-installer/internal/installer/installation_darwin.go
mdrohmann: The latest `appinfo` implementation on `master` should have an option to set a custom base directory, so you can say:

```
fromTray := appinfo.TrayApp(installDir)
toTray := appinfo.TrayApp(appDir)
createSymink(fromTray.Exec(), toTray.Exec())
```

scripts/update-generator/main.go
```diff
@@ -129,10 +146,23 @@ func createUpdate(targetPath string, channel, version, platform string, installe
...
+			err := copyDirToDir(f, sysTempDir)
+			if err != nil {
+				return errs.Wrap(err, "Failed to copy directory %s", f)
+			}
+		} else {
+			err := copyFileToDir(f, binTempDir, true)
+			if err != nil {
+				return errs.Wrap(err, "Failed to copy binary file %s", f)
+			}
 		}
```
mdrohmann, line 166: This works, but it is rather implicit logic: files are binaries => `./bin` and directories are system files that go in the `./system` directory.  Can we maybe provide the system files as a separate argument?  Then we would need to introduce a flag for this, something like `-systemApplicationDir`...

mdrohmann (changes requested):

mdrohmann:
I realized that for these integration tests to work: https://github.com/ActiveState/cli/runs/2423391294#step:14:88, you'll probably need to add the app dir to the `generate-test-update` script too.  You should be able to test them locally on your own machine, BTW...

mdrohmann (approved):

mdrohmann:
Well, I noticed that the Critical Integration Tests on MacOS are still failing. I know they are a little bit difficult to fix, but I found this workflow useful:

- Change the "retainDirectories" flag to `true` (eg., [here](https://github.com/ActiveState/cli/blob/master/test/integration/update_int_test.go#L281-L281)
- Run the test locally
- Read the installation log file that appears in the test output

As it only fails on MacOS now, it probably fails because of some issue with the added functionality.

test/integration/update_int_test.go
```diff
@@ -63,6 +63,15 @@ func (suite *UpdateIntegrationTestSuite) BeforeTest(suiteName, testName string)
...
 
+	// if runtime.GOOS == "darwin" {
+	// 	home, err := homedir.Dir()
+	// 	suite.NoError(err)
+	// 	if !fileutils.DirExists(home) {
+	// 		err = fileutils.Mkdir(home)
+	// 		suite.NoError(err)
+	// 	}
+	// }
+
```
mdrohmann, line 74: nit: I suppose that is a leftover from debugging

mdrohmann (approved):
Looks great. 

And https://github.com/ActiveState/cli/runs/2459251679#step:15:196 is interesting.  Is it possible though that for some users, we'll need to actually create the /Users/<user>/Application directory.  Maybe, because we are the first application in that directory?

 Well, as this happened on CI, we  certainly want to have an install option to skip the Tray installation, as a user will otherwise run into the same problem when they want to use the State Tool on CI...

activestate.macos.yaml
```diff
@@ -0,0 +1,70 @@
...
+  - name: build-systray-app
+    language: bash
+    standalone: true
+    if: eq .OS.Name "MacOS"
+    description: Copies the MacOS application directory to the build dir
+    value: |
+      mkdir -p build/${constants.SYSTRAY_APP}
+
+      cp -r installers/${constants.SYSTRAY_APP} build/
+  - name: generate-update
```
Naatan, line 15: This seems almost identical to the non-macos version. Can we instead use a constant to tweak the flags that are different per OS?

activestate.macos.yaml
```diff
@@ -0,0 +1,70 @@
+constants:
```
Naatan, line 1: This file has a lot of stuff in it for very little change compared to the runners they're overriding. Please try to dial this back to the point that it only contains essentials.

internal/installation/paths.go
```diff
@@ -18,6 +18,13 @@ func InstallPath() (string, error) {
 	return defaultInstallPath()
 }
 
+func SystemInstallPath() (string, error) {
```
Naatan, line 21: Can we name this `LauncherInstallPath()`? `System` confuses me.

activestate.macos.yaml
```diff
@@ -0,0 +1,70 @@
...
+    standalone: true
+    if: eq .OS.Name "MacOS"
+    description: Generate update files
+    value: |
+      set -e
+      $constants.SET_ENV
+
+      export GOARCH=${1:-amd64}
+      mkdir -p ./build/update
+      go run scripts/update-generator/main.go --systemAppDir $BUILD_TARGET_DIR/$constants.SYSTRAY_APP -o ./build/update \
```
Naatan, line 26: Isn't the default systemAppDir already correct? I don't think this runner needs anything macOS specific.

activestate.macos.yaml
```diff
@@ -0,0 +1,70 @@
...
+      $BUILD_TARGET_DIR/$constants.BUILD_INSTALLER_TARGET \
+      $BUILD_TARGET_DIR/$constants.BUILD_DAEMON_TARGET \
+      $BUILD_TARGET_DIR/$constants.BUILD_SYSTRAY_TARGET \
+      $BUILD_TARGET_DIR/$constants.BUILD_TARGET \
+  - name: generate-test-update
```
Naatan, line 31: @mdrohmann this is making me increasingly uncomfortable. Can you please file a followup for cleaning up the testing of updates? This is way too much mocking.

cmd/state-installer/internal/installer/installation_darwin.go
```diff
@@ -9,42 +9,27 @@ import (
...
 const (
-	appDir  = "/Applications"
 	appName = "ActiveState Desktop.app"
 )
 
 // InstallSystemFiles installs files in the /Application directory
-func InstallSystemFiles(installDir string) error {
-	err := os.RemoveAll(filepath.Join(appDir, appName))
+func InstallSystemFiles(fromDir, binaryDir, systemInstallPath string) error {
+	err := os.RemoveAll(filepath.Join(systemInstallPath, appName))
```
Naatan, line 20: This would produce an error if doesn't exist, I think?

cmd/state-installer/internal/installer/installation_test.go
```diff
@@ -24,6 +26,7 @@ var updatedTestFileContent []byte = []byte("#!/bin/bash\necho updated;")
 func init() {
 	if runtime.GOOS == "windows" {
 		stateToolTestFile = "state.exe"
+		stateTrayTestFile = "state-tray.exe"
 		otherTestFile = "other.exe"
 		installerTestFile = "state-installer.exe"
```
Naatan, line 31: You could use this in the definitions above, then drop this whole block:

http://github.com/ActiveState/cli/blob/dd597af58d16cdac0f3cf0694338598543ec15eb/internal/exeutils/exeutils_windows.go#L3-L3

mdrohmann (approved):

scripts/update-generator/main.go
```diff
@@ -93,14 +94,31 @@ func copyFileToDir(filePath, dir string, isExecutable bool) error {
...
+		return errs.Wrap(err, "Could not create target path %s", targetPath)
+	}
+
+	err = fileutils.CopyFiles(fromDir, targetPath)
+	if err != nil {
+		return errs.Wrap(err, "Could not copy directory %s -> %s", fromDir, targetPath)
+	}
+
+	return nil
+}
```
Naatan, line 112: I think you can drop this and use fileutils instead:

http://github.com/ActiveState/cli/blob/dd597af58d16cdac0f3cf0694338598543ec15eb/internal/fileutils/fileutils.go#L537-L537

scripts/update-generator/main.go
```diff
@@ -129,10 +147,23 @@ func createUpdate(targetPath string, channel, version, platform string, installe
...
+	for _, f := range files {
+		err := copyFileToDir(f, binTempDir, true)
 		if err != nil {
-			return errs.Wrap(err, "Failed to copy binary file %s", bf)
+			return errs.Wrap(err, "Failed to copy binary file %s", f)
+		}
+	}
+
+	if systemAppDir != "" {
+		sysTempDir := filepath.Join(tempDir, "system")
```
Naatan, line 158: use `ioutil.TempDir()`

Naatan (changes requested):

activestate.macos.yaml
mdrohmann: https://www.pivotaltracker.com/story/show/177957411

IMO this is still the best way to test the new update mechanism reliably, as it doesn't rely on "new" updates being available on S3 yet, which they are not (for the most part).  So, until the release channel has some of those updates, I would like to keep them.

I would also like to point out two advantages of these tests:
1. They test that `./scripts/update-generator` generates an installable update.  So, without these tests, we may be uploading some broken update and will only find out - after the fact.  That, of course, may not be so bad, as we should catch it on the beta-channel.
2. They make the tests much faster, as we do not have to update over the internet.

Naatan: I'm fine with us still doing some mocking. But if I broke these tests and I had to fix this looking at this code would make me scream.. This looks daunting. I just want to get it in a place where it doesn't make me do that :p

cmd/state-installer/internal/installer/installation_darwin.go
MDrakos: No, `os.RemoveAll` will return nil if the path doesn't exist.

scripts/update-generator/main.go
MDrakos: It's probably better to use that function. And in trying to use it I discovered that it would error out if there was a directory in the source path. However a lot of the logic of this function will remain as it's just creating the target path.

scripts/update-generator/main.go
MDrakos: The installer expects a directory named `system` in its archive. I don't think that `ioutil.TempDir()` can do that as it appends a random string to the end of the filename

activestate.macos.yaml
MDrakos: I'm not following what the default systemAppDir is that you're referring to. The value being passed in here is the one copied from the `installers` directory to the build dir

scripts/update-generator/main.go
Naatan: you can still use `ioutil.TempDir()` for the value of `tempDir`. My worry with the current approach is that the value is constant, so you could run into conflicts.

MDrakos: Fair enough, I doubled checked and `tempDir` is using `ioutil.TempDir()` https://github.com/ActiveState/cli/blob/master/scripts/update-generator/main.go#L110

internal/fileutils/fileutils.go
```diff
@@ -572,6 +578,17 @@ func CopyAndRenameFiles(fromPath, toPath string) error {
...
+func copyPath(info os.FileInfo, fromPath, toPath string) error {
+	if info.IsDir() {
+		err := MkdirUnlessExists(toPath)
+		if err != nil {
+			return errs.Wrap(err, "MkdirUnlessExists %s failed", toPath)
+		}
+		return CopyFiles(fromPath, toPath)
+	}
+	return CopyFile(fromPath, toPath)
+}
```
Naatan, line 590: This looks dangerous. The `CopyAndRenameFiles` already enumerates all the files it needs to copy, and you're calling `CopyFiles` here, essentially turning all subdirs into a `CopyFiles` instruction, but while also still processing them through `CopyAndRenameFiles`.

What are you trying to solve here?

MDrakos: When I started to use `CopyAndRenameFiles` in the update generator, this change was necessary as the function currently does not handle directories in the `fromPath`. I can revert these changes and file a follow-up story as I'm no longer using `CopyAndRenameFiles` in the update generator. It is only being used in the installer and specifically for copying the contents of the `bin` directory.

Naatan: Yes, please revert and file a story. I'm not fully grokking what's happening here, it's looking more complicated than it probably needs to be. Either way not point holding up this PR over if you don't need it.

Naatan (changes requested):
Looks good, but as noted in the other comment let's revert the fileutils changes not required for this PR.

internal/fileutils/fileutils.go
MDrakos: Story filed here: https://www.pivotaltracker.com/story/show/178026455

Naatan (approved):

scripts/update-generator/main.go
```diff
@@ -24,11 +22,11 @@ import (
 
 var exit = os.Exit
 
-var outputDirFlag, platformFlag, branchFlag, versionFlag *string
+var outputDirFlag, platformFlag, branchFlag, versionFlag, systemAppDirFlag *string
```
mdrohmann, line 25:
```suggestion
var outputDirFlag, platformFlag, branchFlag, versionFlag *string
```

scripts/update-generator/main.go
```diff
@@ -175,7 +123,7 @@ func run() error {
 
 	installerPath := flag.Arg(0)
 
-	binaries := flag.Args()[1:]
+	target := flag.Args()[0]
```
mdrohmann, line 126: I think the `installerPath` is not used anymore, so it can be removed here, and also in the `createUpdate` function.

mdrohmann (dismissed):
Looks good.  Just found two things to clean up...

mdrohmann (dismissed):

mdrohmann (approved):

Naatan (approved):

PR #1316, opened 2021-04-20
Fix target dir not existing, and `perl.exe` not having an executor

Naatan:
https://www.pivotaltracker.com/story/show/177844160

I'm not a fan of these fixes, but I need a quick fix to hotfix this issue on the beta channel. As commented in the code a better fix should replace much of the shitty code introduced in this PR.

 https://www.pivotaltracker.com/story/show/177844160

### DIFF
--- a/pkg/platform/runtime/executor/executor.go
+++ b/pkg/platform/runtime/executor/executor.go
@@ -7,8 +7,10 @@ import (
 	rt "runtime"
 	"strings"
 
-	"github.com/ActiveState/cli/pkg/platform/runtime/envdef"
 	"github.com/gobuffalo/packr"
+
+	"github.com/ActiveState/cli/internal/exeutils"
+	"github.com/ActiveState/cli/pkg/platform/runtime/envdef"
 
 	"github.com/ActiveState/cli/internal/appinfo"
 	"github.com/ActiveState/cli/internal/errs"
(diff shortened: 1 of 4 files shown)


pkg/platform/runtime/executor/executor.go
```diff
@@ -107,6 +120,16 @@ func (f *Executor) createExecutor(exe string) error {
 	name := NameForExe(filepath.Base(exe))
 	target := filepath.Clean(filepath.Join(f.executorPath, name))
 
+	if strings.HasSuffix(exe, exeutils.Extension+exeutils.Extension) {
+		// This is super awkward, but we have a double .exe to temporarily work around an issue that will be fixed
```
mdrohmann, line 124: So, is the issue that the directory where we want to write our "executables"/shims to is the same one that we link to? So, we cannot name it "perl.exe", because that file already exists there?

Wouldn't that also be a problem on Linux or Mac, then?

mdrohmann: Nevermind.  I get it now...  This may be the solution then, unless we want to use some actual symlinking (which requires administrator privileges) or deploy actual executables instead of .bat files... :(

mdrohmann (approved):
Oh, Windows is such a mess...

pkg/platform/runtime/executor/executor.go
Naatan: No, [over here](https://github.com/ActiveState/cli/pull/1316/files#diff-e42b086cd919fa036e474c09ef5b0d22e1793581c88e0ca98320c2aa0f65b794R66) I'm adding an additional .exe so that when it replaces it with `.bat` it still has the first .exe, so the filename ends up being eg, `perl.exe.bat`. The problem this creates is that it will set the actual target exe as `perl.exe.exe`, which is what the line you commented on is addressing.

This is SUPER hacky and awkward and that's exactly what the comments and followup tickets are about.

PR #1317, opened 2021-04-21
Get tray working in Ubuntu

daved:
https://www.pivotaltracker.com/story/show/177447608


daved:
@mdrohmann I requested your review because I'm unsure about the build flags required to ensure the C code is working. I have no idea what can or should be removed, so don't take my changes as clear-minded.

cmd/state-tray/internal/open/open_lin.go
```diff
@@ -2,12 +2,29 @@
...
-}
+	shellData, err := exec.Command("bash", "-c",
+		"grep ^$(id -un): /etc/passwd | cut -d: -f7-",
+	).Output()
+	if err != nil {
+		return locale.WrapError(err,
+			"err_determine_default_shell", "Could not determine default shell",
+		)
+	}
+	shell := string(shellData[:len(shellData)-1]) // trim newline
```
Naatan, line 21: What, Bash couldn't do this? :p

cmd/state-tray/internal/open/open_lin.go
```diff
@@ -2,12 +2,29 @@
...
+
+	command = fmt.Sprintf("%s;%s", command, shell)
+	cmd := exec.Command("x-terminal-emulator", "-e", shell, "-c", command)
```
Naatan, line 24: Why do we need to feed the shell into this at all? Shouldn't the terminal emulator pick that up?

cmd/state-tray/libc.c
```diff
@@ -0,0 +1,42 @@
+/*
```
Naatan, line 1: This needs a comment explaining what the file is for. Also can we name it something more descriptive than "libc"?

internal/osutils/autostart/autostart_lin.go
```diff
@@ -2,17 +2,127 @@
...
+		if err != nil {
+			return "", errs.Wrap(err, "Could not open application file")
+		}
+		err = file.Chmod(0770)
+		file.Close()
+		if err != nil {
+			return "", errs.Wrap(err, "Could not make file executable")
+		}
+
+		cmd := exec.Command("gio", "set", appPath, "metadata::trusted", "true")
```
Naatan, line 121: What does this do, and do we have to rely on an external tool (`gio`) to do this?

internal/osutils/autostart/autostart_lin.go
```diff
@@ -2,17 +2,127 @@
...
+}
+
+func ensuredAppPath() (string, error) {
+	appDir, err := dirPath(applicationDir)
+	if err != nil {
+		return "", errs.Wrap(err, "Could not get application file")
+	}
+	appPath := filepath.Join(appDir, launchFile)
+
+	if !fileutils.FileExists(appPath) {
```
Naatan, line 94: Just return early if the file exists. No need to indent 30 lines of code.

internal/osutils/autostart/autostart_lin.go
```diff
@@ -2,17 +2,127 @@
...
+
+func dirPath(dir string) (string, error) {
+	homeDir, err := homedir.Dir()
+	if err != nil {
+		return "", errs.Wrap(err, "Could not get home directory")
+	}
+	return filepath.Join(homeDir, dir), nil
+}
+
+func ensuredAppPath() (string, error) {
```
Naatan, line 87: `ensuredAppPath` does not convey to me what it does. Please name this something more appropriate or add a comment.

internal/osutils/autostart/autostart_lin.go
```diff
@@ -2,17 +2,127 @@
...
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/gobuffalo/packr"
+	"github.com/mitchellh/go-homedir"
+)
+
 func (a *App) Enable() error {
-	panic("Not implemented")
+	appPath, err := ensuredAppPath()
```
Naatan, line 17: `ensuredAppPath` is writing a bunch of files, but you are calling it before even checking if we're already enabled.

Granted `ensuredAppPath` doesn't write when the main app file already exists, but that logic of "don't do this if it already exists" is exactly what should be happening in this function. Your `ensuredAppPath` function should just be in charge of creating the files, not worrying about whether it should.

assets/state-tray.desktop
```diff
@@ -0,0 +1,14 @@
+[Desktop Entry]
+Name=ActiveState Desktop
+GenericName=Language Runtime Manager
+Type=Application
+Comment=Manage ActiveState Platform Projects
+Exec="state-tray"
```
Naatan, line 6: Please use the full path. We don't want to have to rely on PATH for this, especially considering we'll be wanting to test multiple branches side by side.

Naatan (changes requested):

cmd/state-tray/internal/open/open_lin.go
```diff
@@ -2,12 +2,29 @@
...
+
+	"github.com/ActiveState/cli/internal/locale"
+)
 
 func Prompt(command string) error {
-	return errors.New("not implemented")
-}
+	shellData, err := exec.Command("bash", "-c",
+		"grep ^$(id -un): /etc/passwd | cut -d: -f7-",
+	).Output()
```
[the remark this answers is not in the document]
daved: This is the Ubuntu "default" approach to get the user . Or do you mean that we'd just make the user always use Bash for terminals we pull up for them?

This also works with "sh" instead of "bash", but these deps are very much expected to be in any Ubuntu install (bash, grep, cut).

cmd/state-tray/internal/open/open_lin.go
daved: Why, yes. Yes, it can.

cmd/state-tray/internal/open/open_lin.go
daved: When the terminal is brought up without a shell, it will execute the command and collapse immediately. The user will not see the output unless it is held it open. `x-terminal-emulator --hold` does work to display the output, but then the process hangs until it's signaled in some way. This signal cannot be initiated by users using Ctrl-{C,X,Z,D} from within the opened terminal.

The code here is the first reliable way I found to nicely open a terminal with the shell that the user configured. I could dig more for an alternative if you'd like.

cmd/state-tray/internal/open/open_lin.go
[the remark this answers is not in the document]
Naatan: At the very least use Go to do this. Read /etc/passwd and parse the part you're interested in. There's no reason for you to be invoking bash commands like this.

cmd/state-tray/internal/open/open_lin.go
daved: This is resolved by the resolution of another thread.

cmd/state-tray/internal/open/open_lin.go
```diff
@@ -3,22 +3,22 @@
...
 
 	"github.com/ActiveState/cli/internal/locale"
 )
 
+// Prompt brings up the user's preferred shell within a new terminal.
 func Prompt(command string) error {
-	shellData, err := exec.Command("bash", "-c",
-		"grep ^$(id -un): /etc/passwd | cut -d: -f7-",
-	).Output()
+	shell, err := preferredShellWithFallback("/bin/bash")
```
Naatan, line 18: Does it need to use an absolute path? I imagine `bash` should be on PATH.

cmd/state-tray/internal/open/open_lin.go
```diff
@@ -28,3 +28,38 @@ func Prompt(command string) error {
 
 	return nil
 }
+
+func preferredShellWithFallback(fallback string) (string, error) {
+	currentUser, err := user.Current()
+	if err != nil {
+		return "", locale.WrapError(err, "err_user_unknown", "Cannot get current user")
```
Naatan, line 35: So the fallback is only used if we make it through to reading the /etc/passwd? We should use it for any of these errors.

Suggest making this function just error out, and if it errors, log it from the consumer and fall back to bash from the consumer.

cmd/state-tray/internal/open/open_lin.go
```diff
@@ -28,3 +28,38 @@ func Prompt(command string) error {
...
+	}
+
+	f, err := os.Open("/etc/passwd")
+	if err != nil {
+		return "", locale.WrapError(err, "err_open_passwd", "Cannot open passwd file")
```
Naatan, line 40: This should explain why we're looking at the passwd file, otherwise this is a very scary error to get.

cmd/state-tray/internal/open/open_lin.go
```diff
@@ -28,3 +28,38 @@ func Prompt(command string) error {
...
+	for sc.Scan() {
+		line := sc.Text()
+		if !strings.HasPrefix(line, prefix) {
+			continue
+		}
+		parts := strings.Split(line, ":")
+		shell = parts[len(parts)-1]
+	}
+	if err := sc.Err(); err != nil {
+		return "", locale.WrapError(err, "err_scan_passwd", "/etc/passwd file scan failed")
```
Naatan, line 57: Explain why we're looking at this.

internal/osutils/autostart/autostart_lin.go
```diff
@@ -58,6 +63,18 @@ func (a *App) Disable() error {
...
+	if err != nil {
+		return false, errs.Wrap(err, "Could not get application file")
+	}
+	appPath := filepath.Join(appDir, launchFile)
+
+	if !fileutils.FileExists(appPath) {
+		if err = setupAppFile(appPath); err != nil {
+			return false, errs.Wrap(err, "Could not setup app file")
+		}
+	}
```
Naatan, line 76: `IsEnabled()` is for checking if it's enabled.... that's it. This function should not be creating anything.

internal/osutils/autostart/autostart_lin.go
```diff
@@ -84,45 +101,73 @@ func dirPath(dir string) (string, error) {
...
-		return "", errs.Wrap(err, "Could not get application file")
+		return errs.Wrap(err, "Could not parse desktop file template")
 	}
-	appPath := filepath.Join(appDir, launchFile)
 
-	if !fileutils.FileExists(appPath) {
-		box := packr.NewBox("../../../assets")
+	buf := &bytes.Buffer{}
+	trayExec := appinfo.TrayApp().Exec()
+	fmt.Println(trayExec)
```
Naatan, line 113: Debugging line?

Naatan (changes requested):

internal/osutils/autostart/autostart_lin.go
daved: :facepalm:

cmd/state-tray/internal/open/open_lin.go
daved: This package does no other logging, and I think it's not worth adding that. I suggest that we just use the fallback and don't worry about it at this time.

cmd/state-tray/internal/open/open_lin.go
daved: I changed the error message to not be so alarming.

cmd/state-tray/internal/open/open_lin.go
daved: Also changed this one to not be as alarming.

daved: I did add a comment for anyone reading the code, however.

internal/osutils/autostart/autostart_lin.go
daved: The type that is holding the context (the method receiver `*App`) does not have a part of its surface that makes setting up the desktop file make sense. I thought that the desktop file should be added in the installation process, but I was not sure since the AC of my story includes it. Can I make a story to later move the logic into the installer?

cmd/state-tray/internal/open/open_lin.go
mdrohmann: But shouldn't we return the fallback anyways in that case?  I don't think the user should ever be bothered with that...

cmd/state-tray/internal/open/open_lin.go
mdrohmann: Here I would also return the fallback.  Maybe this function should not return an error at all?  Just log problems via `logging.Debug`...

mdrohmann (changes requested):

cmd/state-tray/internal/open/open_lin.go
Naatan: I'd like to have the logging so we can gauge whether this is actually a sensible mechanic to use.

internal/osutils/autostart/autostart_lin.go
Naatan: `Enable()` seems like a pretty sensible place to me? Much more sensible than `IsEnabled()`.

cmd/state-tray/internal/open/open_lin.go
daved: From mobile: This looks like the code reviewed is from an older commit (or I didn't push!). Does this still apply?

internal/osutils/autostart/autostart_lin.go
daved: IsEnabled is run more often in the control flow rather than strictly when autostart is enabled manually. Using "Enable" for this logic means the application will not show up in the launcher until the application is enabled.

The desktop shortcut file creation should probably be an installation step, and not even in this code. Given the existing code and AC, I still suggest using IsEnabled along with a story to organize/model these things better.

cmd/state-tray/internal/open/open_lin.go
daved: Does this mean logging to Rollbar in order for us to see it?

internal/osutils/autostart/autostart_lin.go
Naatan: Oh I see what you're saying. You're mixing the launcher into the autostart behaviour. For the launcher a more appropriate place would be:

https://github.com/ActiveState/cli/blob/master/internal/runners/prepare/prepare_windows.go#L40-L40

I thought I already had a followup story to remove `state _prepare` in favour of the new installer, but can't seem to find it. Logged a new one:

https://www.pivotaltracker.com/story/show/177913672

cmd/state-tray/internal/open/open_lin.go
mdrohmann: Yes, I think so.  I believe that we should always use the fallback from this function if an error occurs.  Never forward the error, even if it looks less sketchy. :)

Once you add rollbar logging, you can still log the errors there...

cmd/state-tray/internal/libc-compatwrap/compatwrap_linux.go
```diff
@@ -0,0 +1,5 @@
+package compatwrap
```
Naatan, line 1: Does this really need to be in its own package? Wouldn't it be sufficient to have the OS suffix?

internal/osutils/autostart/autostart_lin.go
```diff
@@ -33,13 +30,35 @@ func (a *App) Enable() error {
...
+	path := filepath.Join(dir, constants.TrayLaunchFileName)
 
-	if err = os.Symlink(appPath, path); err != nil {
-		return errs.Wrap(err, "Could not create symlink")
+	scut, err := shortcut.New(appinfo.TrayApp().Exec(), path)
+	if err != nil {
+		return errs.Wrap(err, "Could not construct autostart shortcut")
 	}
+
+	iconsDir, err := dirPath(constants.IconsDir)
```
Naatan, line 42: dirPath is not a great name, not obvious what it does

internal/osutils/shortcut/shortcut_linux.go
```diff
@@ -0,0 +1,116 @@
...
+
+type ShortcutSaveOpts struct {
+	GenericName string
+	Comment     string
+	Keywords    string
+	IconData    []byte
+	IconPath    string
+}
+
+func (s *Shortcut) Save(name string, opts ShortcutSaveOpts) error {
```
Naatan, line 41: the target and save path go to `New()`, a number of opts go to `ShortcutSaveOpts` and then the Name goes to `Save()`. Frankly I don't follow why these parameters are being cut up this way. Pick one?

Additionally I don't want this API to be completely different depending on what OS you're on. Look at the Windows version and align it with that, or update both. But keep things simple. We don't need three different ways of passing parameters.

internal/osutils/shortcut/shortcut_linux.go
```diff
@@ -0,0 +1,116 @@
...
+	t := template.New("")
+	t, err := t.Parse(desktopFileTmpl)
+	if err != nil {
+		return errs.Wrap(err, "Could not parse desktop file template")
+	}
```
Naatan, line 46: fyi; http://github.com/ActiveState/cli/blob/9faf61f1e772249d29da4b95fadae575faf1c7a5/internal/strutils/strutils.go#L17-L17

internal/osutils/shortcut/shortcut_linux.go
```diff
@@ -0,0 +1,116 @@
...
+	file.Close()
+	if err != nil {
+		return errs.Wrap(err, "Could not make file executable")
+	}
+
+	// set the executable as trusted so users do not need to do it manually
+	// gio is "Gnome input/output"
+	cmd := exec.Command("gio", "set", s.path, "metadata::trusted", "true")
+	if err := cmd.Run(); err != nil {
+		return errs.Wrap(err, "Could not set desktop file as trusted")
```
Naatan, line 86: I think we should log this instead and not fail, because this is the only part I can see that's truly distro specific. The rest could theoretically work on various linux distro's.

internal/runners/prepare/prepare_linux.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/ActiveState/cli/internal/osutils/shortcut"
+	"github.com/ActiveState/cli/internal/rtutils"
+	"github.com/gobuffalo/packr"
+	"github.com/mitchellh/go-homedir"
+)
+
+func (r *Prepare) prepareOS() error {
+	if rtutils.BuiltViaCI { // disabled while we're still testing this functionality
+		return nil
+	}
```
Naatan, line 20: I think we can drop this now?

internal/runners/prepare/prepare_linux.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if err := r.prepareDesktopShortcut(); err != nil {
+		r.reportError(locale.Tr("err_prepare_shortcut", "Could not create start menu shortcut, error received: {{.V0}}.", err.Error()), err)
+	}
+
+	return nil
+}
+
+func (r *Prepare) prepareDesktopShortcut() error {
```
Naatan, line 33: This name seems wrong, we're not creating a "desktop" shortcut.

Naatan (changes requested):

cmd/state-tray/internal/libc-compatwrap/compatwrap_linux.go
daved: I think so. I was not able to find another way to ensure the C file was only picked up by Linux. I did try adding a `main_linux.go` file with the C import, but the compiler kept trying to add the C file to Windows and Darwin builds regardless and it returned errors "must import C".

internal/osutils/shortcut/shortcut_linux.go
daved: The params type required data to be structured more magically than necessary, so I swapped from map[string]interface{} to interface{}. Other callers should not be affected by this change.

internal/runners/prepare/prepare_linux.go
daved: Sure, should it also be removed from the equivalent Windows logic?

internal/osutils/autostart/autostart_lin.go
daved: There were two separate packages using this name, and both have been made more clear.

internal/runners/prepare/prepare_linux.go
Naatan: Yeah let's remove it.

mdrohmann:
I just noticed that this PR, installs the "App" and shortcuts via `state _prepare` whereas PR #1315 has that baked into the installer.  Nothing wrong with that, especially as we do not have to create a complete "Application" as on MacOS, but it is probably worth-while to revisit these separations during our architecture meeting this week.

cmd/state-tray/internal/libc-compatwrap/compatwrap_linux.go
Naatan: That's strange.. almost sounds like a bug in Go?

daved:
A story has been filed (https://www.pivotaltracker.com/story/show/177973533) to ensure the shortcut package Windows implementation more closely resembles the more sleek Linux implementation. On that note, the current Linux implementation includes a filename return in anticipation of the need in Windows to know what the shortcut filename is.

Naatan (approved):

mdrohmann (approved):

PR #1318, opened 2021-04-21
Deploy does not use executors

MDrakos:
https://www.pivotaltracker.com/story/show/177860990


Naatan (changes requested):
Looks good to me, except you don't need to pass the namespace. It was only user by virtualenv.

MDrakos:
The namespace being passed to `configure` is used by `SetupShellRCFile`. I'm not seeing it being passed anywhere else.

Naatan (approved):

PR #1319, opened 2021-04-21
Refresh runtime on pull

MDrakos:
https://www.pivotaltracker.com/story/show/177602458


test/integration/pull_int_test.go
```diff
@@ -18,15 +19,29 @@ func (suite *PullIntegrationTestSuite) TestPull() {
...
+	ts.PrepareActiveStateYAML(`project: "https://platform.activestate.com/ActiveState-CLI/Pull?commitID=b490d8f1-9e0c-4294-a9a1-88c5957665b5"`)
 
-	cp := ts.Spawn("pull")
+	cp := ts.SpawnWithOpts(
+		e2e.WithArgs("activate"),
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+	)
+	cp.ExpectLongString("default project?")
+	cp.Send("n")
+	cp.WaitForInput(120 * time.Second)
```
mdrohmann, line 30: I suppose this will work better if you do a `cp.Expect("You're Activated")` before this line.

The `WaitForInput` assertion, sends a message that is supposed to be interpreted by the starting shell, but if that happens to early (before the shell starts), it may land in some terminal nirvana...  By adding the proposed expectation you will shorten the time between sending the string and the shell starting up...

mdrohmann (approved):

Naatan (approved):

PR #1322, opened 2021-04-23
Use language filename

MDrakos:
https://www.pivotaltracker.com/story/show/177894983

Tested on Windows and I am able to run scripts without any language errors and the value of the language executable in a generated `activestate.yaml` has the correct `.exe` extenstion.


Naatan (dismissed):

MDrakos:
The integration test failure for `TestCarlisle` could be related to this: https://www.pivotaltracker.com/story/show/177628756.

We may also want to find a better solution for the unit tests on Ubuntu.

internal/scriptrun/scriptrun.go
```diff
@@ -196,6 +198,10 @@ func applySuffix(suffix string, paths []string) []string {
 }
 
 func isExecutableFile(name string) bool {
+	if !fileutils.FileExists(name) {
+		name = executor.NameForExe(name)
+	}
```
Naatan, line 203: I'm not a fan of including executor code here. I think the fix for this function should be to check if the ext of `name` is on PATHEXT.

Naatan (changes requested):

internal/scriptrun/scriptrun.go
MDrakos: In this case `language.Executable().Filename()` will return `perl.exe` so that extension should always be on `PATHEXT`. Since we are using executors we need to search for `perl.bat` or `perl.exe.bat`. We could potentially update the `language` package to return the executor filename for languages that we don't allow 3rd party usage of (ie. if [this](https://github.com/ActiveState/cli/blob/master/internal/language/language.go#L252) field is false).

Naatan: My point is that .bat should be on PATHEXT (right?) and so searching for `perl.bat` should return that it is executable.

Naatan (approved):

MDrakos:
@Naatan Can you merge this? Not sure why it's expecting another CI run. All of the checks have passed on the latest run.

PR #1325, opened 2021-04-26
MoveAllFilesRecursively() works the same on all platforms

mdrohmann:
https://www.pivotaltracker.com/story/show/177628756

These fix ensures that we can copy artifact files in the target installation directory, even if they already exist (with read-only permissions on Windows).

### DIFF
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -486,30 +486,33 @@ func MoveAllFilesRecursively(fromPath, toPath string, cb MoveAllFilesCallback) e
 			if fileInfo.Mode() != toInfo.Mode() {
 				logging.Warning(locale.Tr("warn_move_incompatible_modes", "", subFromPath, subToPath))
 			}
+
+			// If we are moving to an existing directory, call function recursively to overwrite and add files in that directory
+			if toInfo.IsDir() {
+				err := MoveAllFilesRecursively(subFromPath, subToPath, cb)
+				if err != nil {
+					return err
+				}
+				// source path should be empty now
+				err = os.Remove(subFromPath)
+				if err != nil {
+					return errs.Wrap(err, "os.Remove %s failed", subFromPath)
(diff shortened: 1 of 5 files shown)


MDrakos (approved):

internal/fileutils/fileutils.go
```diff
@@ -498,6 +498,12 @@ func MoveAllFilesRecursively(fromPath, toPath string, cb MoveAllFilesCallback) e
 				return errs.Wrap(err, "os.Remove %s failed", subFromPath)
 			}
 		} else {
+			// If the subToPath file exists, we remove it first - to ensure compatibility between in platforms:
+			// On Windows, the renaming step can otherwise fail if subToPath is read-only (file removal is allowed)
+			if toPathExists {
+				err = os.Remove(subToPath)
+				logging.Error("Failed to remove destination file %s: %v", subToPath, err)
+			}
```
Naatan, line 506: Can we take this chance to clean up the conditional flow here a little? I'm confused why this isn't in the conditional on line 482, I'm guessing it's got something to do with it being a file, but the flow of this logic is now making it really hard to grok why the code lives where it lives.

Not asking for any major refactoring, just clean up the conditionals so it's more self-explanatory.

Naatan (changes requested):

internal/fileutils/fileutils.go
mdrohmann: Yes, the block starting on line 482 is still handling with files AND directories, just checking whether the arguments are compatible.  But, yes I think I can flatten this logic a little...

Naatan (approved):
Much better! This function is still RIPE for a cleanup, but let's not get too distracted for this story :p

mdrohmann:
@Naatan Could you please merge this to master?  It is still failing on the TestCarlisle integration test, but that should work again once Mike's PR is also merged...

PR #1328, opened 2021-04-27
Install scripts can install a specific version

MDrakos:
https://www.pivotaltracker.com/story/show/177825507


mdrohmann (changes requested):
Looks great to me.

Only: Can we maybe add an integration test for that case?  For now it would need to install a new version from master, as that is the only branch with deployed "new" updates...

mdrohmann (approved):

Naatan (approved):
Nice work!

PR #1330, opened 2021-04-28
First draft for update warning

mdrohmann:
https://www.pivotaltracker.com/story/show/177827736

It currently looks like this: 
Selection_030

It is a different from what is asked for [here](https://www.pivotaltracker.com/story/show/177824358/comments/223624674).  IMO the suggested output needs to be a little revised.  We will see what Akash thinks about it.

It would be great if you could give some comments on the general design already.  I wasn't sure if this is how you envisioned the data to flow.  There certainly are several ways to do it.


cmd/state-update-dialog/app.go
```diff
@@ -30,13 +32,13 @@ type App struct {
...
 	a.wails = wails.CreateApp(&wails.AppConfig{
-		Width:            600,
-		Height:           400,
-		Title:            "ActiveState Desktop - Update Available",
-		HTML:             html,
-		JS:               js,
-		CSS:              css,
-		Colour:           "#FFF", // Wails uses this to detect dark mode
+		Width:  800,
+		Height: 600,
```
Naatan, line 36:
```suggestion
		Width:  600,
		Height: 400,
```

Don't worry about the design for this PR.

Naatan (changes requested):
This looks good to me. The design needs some tweaking but I'll do that in my PR.

The only thing I'm missing is the part where it checks that the lock is for "older" versions. Ie. we only want to warn about projects that are locked to versions where auto-forwarding was still a thing. Again we can just hardcode that particular version number as a constant.

mdrohmann:
> The only thing I'm missing is the part where it checks that the lock is for "older" versions. Ie. we only want to warn about projects that are locked to versions where auto-forwarding was still a thing. Again we can just hardcode that particular version number as a constant.

I added that check, but I wonder if it is correct... Projects with locks for multi-file updates also get unusable, as they will receive an error message that the project is locked to a version that is not installed anymore.  And for now, the single-file updates are actually forwarded, so they would still work.  So, maybe that check needs to be the other way round?  Or even get removed completely?

Naatan:
That auto-forwarding will only work for one version after we deploy, after that it won't work anymore. So those would still need the warning.

As for projects with locks for multi-file updates; I'm confused, we discussed this. They will get an error explaining how to install an older version.

The warning is about informing people that this thing that they've gotten used to is going to change. It doesn't change any of our decisions regarding how the locking works.

Naatan (approved):

mdrohmann:
> As for projects with locks for multi-file updates; I'm confused, we discussed this. They will get an error explaining how to install an older version.
I know that. But I understood it this way: *This* warning is exactly for that use-case: To inform the user, that they might have to do the installation again (installing an older State Tool version) to use those projects.  So, that's what confuses me...

Naatan:
> I know that. But I understood it this way: _This_ warning is exactly for that use-case: To inform the user, that they might have to do the installation again (installing an older State Tool version) to use those projects. So, that's what confuses me...

I could get on board with that, if it simplifies things. So this warning would show and instruct them how to install the older version? Or I guess link to the docs that would explain it? And as a result we wouldn't need the auto-forward mechanic in this update at all?

PR #1331, opened 2021-04-29
Facilitate incremental download of packages

mdrohmann:
https://www.pivotaltracker.com/story/show/177290892

### DIFF
--- a/pkg/platform/runtime/setup/implementations/alternative/runtime.go
+++ b/pkg/platform/runtime/setup/implementations/alternative/runtime.go
@@ -25,7 +25,9 @@ func NewSetup(store *store.Store, artifacts artifact.ArtifactRecipeMap) *Setup {
 	return &Setup{store: store, artifacts: artifacts}
 }
 
-func (s *Setup) DeleteOutdatedArtifacts(changeset artifact.ArtifactChangeset, storedArtifacted store.StoredArtifactMap) error {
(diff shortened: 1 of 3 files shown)


pkg/platform/runtime/setup/implementations/alternative/runtime.go
```diff
@@ -25,7 +25,9 @@ func NewSetup(store *store.Store, artifacts artifact.ArtifactRecipeMap) *Setup {
 	return &Setup{store: store, artifacts: artifacts}
 }
 
-func (s *Setup) DeleteOutdatedArtifacts(changeset artifact.ArtifactChangeset, storedArtifacted store.StoredArtifactMap) error {
+func (s *Setup) DeleteOutdatedArtifacts(changeset artifact.ArtifactChangeset, storedArtifacted store.StoredArtifactMap) ([]artifact.ArtifactID, error) {
+	var alreadyInstalled []artifact.ArtifactID
```
Naatan, line 29: Not going to request changes on this now, but this seems like something we should fix in the future. It's awkward that a function responsible for deleting outdated artifacts is returning artifacts that you want to keep. Either this function needs to be renamed or its logic needs to be split up into separate functions.

Naatan (approved):
Looks ok to me. Did you consider how this affects the env merging at the end?

pkg/platform/runtime/setup/implementations/alternative/runtime.go
mdrohmann: I agree.  I pushed a new version, that also gets rid of the go-funk dependency...

Naatan (approved):
Much better!

PR #1332, opened 2021-04-29
Add transitional updater

mdrohmann:
https://www.pivotaltracker.com/story/show/177827563

This certainly feels and looks hacky, but I am fairly certain that is the nature of this story...

"Follow-up" story is here: https://www.pivotaltracker.com/story/show/177974542


activestate.generators.yaml
```diff
@@ -56,9 +56,13 @@ scripts:
     description: Generate update files
     value: |
       set -e
+      export GOARCH=${1:-amd64}
       $constants.SET_ENV
 
-      export GOARCH=${1:-amd64}
+      # Generate legacy update
+      go build -o $BUILD_TARGET_DIR/$constants.BUILD_TRANSITION_TARGET ./cmd/state-transition-update
```
Naatan, line 63: I think we'll need to sign this on Windows, otherwise we might get the UAC prompt?

Naatan (changes requested):
Looks ok to me! Just need to ensure that the executable is signed on Windows.

activestate.generators.yaml
mdrohmann: Good point!  I think I'll cramp that in there too, as we are deleting that code soon anyways...

mdrohmann: Shoot. That won't work, because signing is only done on CI.  Okay, then I have to split it up.

Naatan (approved):

PR #1333, opened 2021-04-30
Fix update url for State Tool multi-file updates

mdrohmann:
https://www.pivotaltracker.com/story/show/177990143


test/integration/update_lock_int_test.go
```diff
@@ -38,6 +38,10 @@ func (suite *UpdateIntegrationTestSuite) TestLocked() {
 }
 
 func (suite *UpdateIntegrationTestSuite) TestLockedChannel() {
+	targetBranch := "release"
+	if constants.BranchName == "release" {
+		targetBranch = "master"
+	}
```
mdrohmann, line 44: This is just temporary, because forwarding to the latest master branch does not work anymore, as we have multi-file updates there now.  These tests will have to be adapted soon anyways...

internal/updater/checker.go
```diff
@@ -29,7 +30,7 @@ type Checker struct {
 var DefaultChecker = newDefaultChecker()
 
 func newDefaultChecker() *Checker {
-	updateURL := constants.APIUpdateURL
+	updateURL := constants.APIUpdateURL + url.QueryEscape(constants.CommandName)
```
Naatan, line 33: I think it would be cleaner simply to update `APIUpdateURL` itself to have the `/state` suffix.

Naatan: Also; no need to use `constants.CommandName`. The ship has sailed on keeping that name centralized.

Naatan (changes requested):

Naatan (approved):

PR #1334, opened 2021-05-03
Update Dialog installs update

Naatan:
https://www.pivotaltracker.com/story/show/177824506


cmd/state-installer/installer.go
```diff
@@ -111,11 +117,9 @@ func install(installPath string, cfg *config.Instance, out output.Outputer) erro
 	}
 
 	tmpDir := filepath.Dir(exe)
-	// clean-up temp directory when we are done.
-	defer os.RemoveAll(tmpDir)
-
 	inst := installer.New(filepath.Join(tmpDir, "bin"), installPath)
 	defer func() {
+		os.RemoveAll(tmpDir)
```
mdrohmann, line 122: Did you remove this for debugging purposes?

cmd/state-update-dialog/app.go
```diff
@@ -48,30 +49,19 @@ func (a *App) CurrentVersion() string {
...
-	if update == nil {
-		return errs.New("No updates available")
-	}
-	bindings.update = update
-
-	cfg, err := config.New()
+	bindings := &Bindings{cfg: a.cfg}
+	var err error
+	// bindings.update, err = updater.NewChecker(constants.APIUpdateURL, "master", "", httpreq.New()).Check()
+	bindings.update, err = updater.DefaultChecker.Check() // TODO: Replace with this!
```
mdrohmann, line 55:
```suggestion
	bindings.update, err = updater.DefaultChecker.Check()
```

scripts/s3-deployer/main.go
```diff
@@ -53,12 +53,25 @@ func run() {
...
+func (l *logger) Log(v ...interface{}) {
+	fmt.Printf("AWS Log: %v", v)
+}
+
 func createSession() {
 	// Specify profile to load for the session's config
 	var err error
+	var verboseErr = true
+	var logLevel = aws.LogDebug
+	_ = logLevel
```
mdrohmann, line 67: Do we still need the verbose error reporting?

mdrohmann (changes requested):

scripts/s3-deployer/main.go
Naatan: I left it in there so it's easy to toggle if we need it in the future.

cmd/state-installer/installer.go
Naatan: I didn't? It's just in a single defer now.

mdrohmann: Oh dear! How silly of me...

mdrohmann (changes requested):
I finally got to run it on Windows, tested it, and the Close button never gets enabled.  So, apparently, the `installDone` variable is never set to `true`...

mdrohmann (approved):

PR #1337, opened 2021-05-05
Rollbar is missing reports

mdrohmann:
https://www.pivotaltracker.com/story/show/178029680

What I did:
- I created a fake error message during `setup`, and forced it to report to the `master` branch on rollbar
- The error did not show up, but this falsely sent message appeared: https://rollbar.com/activestate/state-tool/items/22759/  (I fixed that in f19088d.
- Then I added a7be149, and ran it again.
- The error message appeared: https://rollbar.com/activestate/state-tool/items/22761/

It is possible that this also affected the analytics reporting, but I am not sure about this, of course...
I also do not know if current changes somehow made this more likely to happen.

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -202,8 +202,8 @@ func run(args []string, isInteractive bool, out output.Outputer) error {
 	if childCmd == nil || !childCmd.SkipChecks() {
 		// Auto update to latest state tool version, only runs once per day
 		// Todo: This is better done in the `state-svc` process  https://www.pivotaltracker.com/story/show/177730748
-		if updated, err := autoUpdate(args, cfg, pjPath); err != nil || updated {
-			return err
+		if _, err := autoUpdate(args, cfg, pjPath); err != nil {
+			logging.Error("Failed to initialize auto update: %v", err)
 		}
 	}
 
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/ActiveState/sysinfo"
 	"github.com/rollbar/rollbar-go"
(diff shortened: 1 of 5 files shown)


cmd/state/main.go
```diff
@@ -202,8 +209,8 @@ func run(args []string, isInteractive bool, out output.Outputer) error {
 	if childCmd == nil || !childCmd.SkipChecks() {
 		// Auto update to latest state tool version, only runs once per day
 		// Todo: This is better done in the `state-svc` process  https://www.pivotaltracker.com/story/show/177730748
-		if updated, err := autoUpdate(args, cfg, pjPath); err != nil || updated {
-			return err
+		if _, err := autoUpdate(args, cfg, pjPath); err != nil {
+			logging.Error("Failed to initialize auto update: %v", err)
 		}
```
mdrohmann, line 214: This is kind of unrelated.  But I ran into this reviewing the code: And I strongly believe (as we are now running this at the end of the State Tool), users might prefer this behavior.
Especially, when they are running automated commands, they probably do not want the State Tool to error out, because the auto-update could not be scheduled...

cmd/state/main.go
```diff
@@ -33,7 +33,7 @@ import (
 	"github.com/ActiveState/cli/pkg/projectfile"
 )
 
-func main() {
+func _main() (exitCode int) {
```
Naatan, line 36: This feels very proof of concept, let's clean this up.

I kinda like the pattern I landed on with the update dialog:

http://github.com/ActiveState/cli/blob/d00dfc6951aa008c2dd8761bf88c4cd4ea399277/cmd/state-update-dialog/main.go#L17-L17

ie. put the os.Exit() in a defer itself.

Naatan (changes requested):

Naatan (approved):

PR #1338, opened 2021-05-05
State branch panic fix

MDrakos:
https://www.pivotaltracker.com/story/show/177918239

### DIFF
--- a/test/integration/update_int_test.go
+++ b/test/integration/update_int_test.go
@@ -88,13 +88,13 @@ func (suite *UpdateIntegrationTestSuite) env(disableUpdates, testUpdate bool) []
 
 	if testUpdate {
 		env = append(env, fmt.Sprintf("_TEST_UPDATE_URL=http://localhost:%s/", testPort))
-
-		dir, err := ioutil.TempDir("", "system*")
-		suite.NoError(err)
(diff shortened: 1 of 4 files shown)


test/integration/update_int_test.go
```diff
@@ -88,14 +88,14 @@ func (suite *UpdateIntegrationTestSuite) env(disableUpdates, testUpdate bool) []
...
-		suite.NoError(err)
-		env = append(env, fmt.Sprintf("_TEST_SYSTEM_PATH=%s", dir))
 	} else {
 		env = append(env, fmt.Sprintf("%s=%s", constants.UpdateBranchEnvVarName, targetBranch))
 	}
 
+	dir, err := ioutil.TempDir("", "system*")
+	suite.NoError(err)
+	env = append(env, fmt.Sprintf("_TEST_SYSTEM_PATH=%s", dir))
+
```
Naatan, line 98: I'm not sure what moving this accomplishes, please flag martin if you want a second set of eyes on that.

Naatan (approved):

test/integration/update_int_test.go
MDrakos: Most of the update integration tests use this function to add additional environment variables to their commands. Moving this ensures that for those tests we always install the system files to a temporary directory rather than `~/Applications`

PR #1339, opened 2021-05-05
Add tray update notifcations

daved:
https://www.pivotaltracker.com/story/show/177825487


daved:
Noting specifically in case there is feedback:
1. I didn't want to mess with environment variables while developing this, so I temporarily added a simple "verbose" flag. This led to me moving the verbose envvar check to the main func rather than the run func. It seems worth leaving in the main now that it's there, but I would not have done so completely unprompted.
2. There is a super simple back-off setup if the update check fails. It was useful during testing. Again, I would not have added it unprompted, but it seems worth leaving now that it's there.

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -44,6 +44,19 @@ func (r *Resolver) Version(ctx context.Context) (*graph.Version, error) {
 	}, nil
 }
 
+func (r *Resolver) AvailableUpdate(ctx context.Context) (*graph.AvailableUpdate, error) {
```
Naatan, line 47: Why not have this return the actual update information? You don't have to use it besides seeing that it's provided, but then it's there for future use.

cmd/state-tray/main.go
```diff
@@ -44,18 +53,12 @@ func onReady() {
...
 
 	svcInfo := appinfo.SvcApp()
-	if !fileutils.FileExists(svcInfo.Exec()) {
-		return errs.New("Could not find: %s", svcInfo.Exec())
-	}
-
-	if _, err := exeutils.ExecuteAndForget(svcInfo.Exec(), "start"); err != nil {
-		return errs.Wrap(err, "Could not start %s", svcInfo.Exec())
+	if err := execute(svcInfo.Exec()); err != nil {
+		return errs.New("Could not execute: %s", svcInfo.Name())
```
Naatan, line 61: Nit: Please avoid doing this, it's completely unrelated to your PRs and this code is heavily in flux. so you're creating conflicts for no reason.

cmd/state-tray/main.go
```diff
@@ -68,10 +71,21 @@ func run() error {
 		return errs.Wrap(err, "Could not create new service model")
 	}
 
-	box := packr.NewBox("../../assets")
-	systray.SetIcon(box.Bytes("icon.ico"))
 	systray.SetTooltip(locale.Tl("tray_tooltip", "ActiveState State Tool"))
 
+	mUpdate := systray.AddMenuItem(
+		locale.Tl("tray_update_title", "Update Available"),
+		locale.Tl("tray_update_tooltip", "Update your State Tool installation"),
```
Naatan, line 78:
```suggestion
		locale.Tl("tray_update_tooltip", "Update your ActiveState Desktop installation"),
```

cmd/state-tray/update.go
```diff
@@ -0,0 +1,65 @@
...
+				return
+			}
+		}
+	}()
+
+	return func() { close(done) }
+}
+
+func needsUpdate(mdl *model.SvcModel) bool {
+	for i := 1; i <= 3; i++ {
```
Naatan, line 36: This isn't necessary, our graphql client has retrying built-in. YOU implemented that.. ;)

Naatan (changes requested):

cmd/state-svc/internal/resolver/resolver.go
daved: The additional information has been added.

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -50,11 +50,22 @@ func (r *Resolver) AvailableUpdate(ctx context.Context) (*graph.AvailableUpdate,
 		return nil, errs.Wrap(err, "Failed to check for update")
 	}
 
-	av := graph.AvailableUpdate{
-		Available: update != nil,
+	available := true
+	if update == nil {
+		available = false
+		update = &updater.AvailableUpdate{}
```
Naatan, line 56: If no update is available then the response should be empty/null.

Naatan (changes requested):

activestate.yaml
```diff
@@ -80,7 +80,7 @@ scripts:
           sudo apt-get update && sudo apt-get install ${pkgs} -y
         fi
       }
-      aptin libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev
+      aptin libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev sassc
```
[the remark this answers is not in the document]
daved: It didn't load up for me, then. Is there a way to ensure that it is available when it is called?

cmd/state-svc/internal/resolver/resolver.go
daved: This is done. I strongly recommend against returning `nil, nil`. It is not common, and can lead to mistakes. It is common for there to be explicit verification of data through errors or "ok" values (the "Available" field was the ok value in this case) with the most major exception being types which are commonly used as `nil` like slices.

activestate.yaml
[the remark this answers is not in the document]
Naatan: `state run` or `state activate` should do that for you, if that's not working please report a bug.

Let's not distract this PR with unrelated issues though.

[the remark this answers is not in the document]
daved: I did find libsass in our platform project, but I'm fairly sure that doesn't provide sassc itself.

Naatan (approved):

activestate.yaml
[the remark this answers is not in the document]
daved: This has been reported here: https://www.pivotaltracker.com/story/show/178152827

PR #1342, opened 2021-05-07
Fix text of log line which was the opposite of reality

nugget:
The text of this debug log line was actually the reverse of what it intended to say.

```
[DEBUG projectfile.go:471] Not merging activestate.windows.yaml as we're not on MacOS
```


Naatan (approved):
Thanks!

PR #1344, opened 2021-05-08
Add ProjectPath details to project output

swagile1:


cmd/state/internal/cmdtree/show.go
```diff
@@ -18,7 +18,14 @@ func newShowCommand(prime *primer.Values) *captain.Command {
...
 		prime.Config(),
-		nil,
+		[]*captain.Flag{
+			{
+				Name:        "detail",
+				Shorthand:   "d",
+				Description: locale.T("arg_state_show_remote_description"),
+				Value:       &params.Detail,
+			},
+		},
```
Naatan, line 28: I don't think this warrants a flag. Including the project path seems like something sensible that we should just always do.

Naatan (changes requested):
Looking good! Thanks for contributing :) Just one minor thing.

cmd/state/internal/cmdtree/show.go
swagile1: That makes it easier. I'll remove all the flag bits then.

Naatan (changes requested):
So I had not realized that you were exposing the namespace with the commit hash at the end. I read "ProjectPath" and assumed you were exposing the filepath of the project.

Now that I understand this, I'd say this would be a lot cleaner if we simply exposed the project url that you defined in the activestate.yaml. I understand you are only interested in the path bit, but that feels too specific to your niche use-case imo. 

Another thing we could do is expose these things as variables in your yaml, so you could use it like:

```
constants:
  - name: project-namespace
     value: ${project.Namespace()}
  - name: project-commit
     value: ${project.CommitID()}
```

Yet another solution could be to add it as a `state export` subcommand, which can be a little more specific.

Key point being that for the purposes of `state show` this feels too "advanced". I get why you added a flag now, but that feels more like a workaround than an appropriate solution.

PR #1345, opened 2021-05-10
Update error authentication error message on activation

MDrakos:
https://www.pivotaltracker.com/story/show/177956864

### DIFF
--- a/internal/runners/activate/activate.go
+++ b/internal/runners/activate/activate.go
@@ -189,6 +189,9 @@ func (r *Activate) run(params *ActivateParams) error {
 			return locale.WrapError(err, "err_activate_runtime", "Could not initialize a runtime for this project.")
 		}
 		if err = rt.Update(runbits.DefaultRuntimeEventHandler(r.out)); err != nil {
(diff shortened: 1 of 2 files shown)


internal/runners/activate/activate.go
```diff
@@ -189,6 +189,9 @@ func (r *Activate) run(params *ActivateParams) error {
 			return locale.WrapError(err, "err_activate_runtime", "Could not initialize a runtime for this project.")
 		}
 		if err = rt.Update(runbits.DefaultRuntimeEventHandler(r.out)); err != nil {
+			if errs.Matches(err, &model.ErrOrderAuth{}) {
+				return locale.WrapInputError(err, "err_update_auth", "Could not update runtime, if this is a private project you may need to authenticate with `state auth`")
```
Naatan, line 193:
```suggestion
				return locale.WrapInputError(err, "err_update_auth", "Could not update runtime, if this is a private project you may need to authenticate with `[ACTIONABLE]state auth[/RESET]`")
```

Naatan (changes requested):

PR #1348, opened 2021-05-10
Authenticate head-chef requests

mdrohmann:
https://www.pivotaltracker.com/story/show/178100022

### DIFF
--- a/pkg/platform/api/inventory/inventory.go
+++ b/pkg/platform/api/inventory/inventory.go
@@ -42,9 +42,9 @@ func New(serviceURL *url.URL, auth runtime.ClientAuthInfoWriter) (inventory_oper
 }
 
 // Get returns a cached version of the default api client
-func Get(auth *authentication.Auth) inventory_operations.ClientService {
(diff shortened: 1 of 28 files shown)


pkg/platform/api/inventory/inventory.go
```diff
@@ -22,24 +23,28 @@ var persist inventory_operations.ClientService
...
+	if auth != nil {
+		transportRuntime.DefaultAuthentication = auth
+	}
+
 	return inventory_operations.New(transportRuntime, strfmt.Default), transportRuntime
 }
 
 // Get returns a cached version of the default api client
-func Get() inventory_operations.ClientService {
+func Get(auth *authentication.Auth) inventory_operations.ClientService {
```
Naatan, line 45: This one doesn't really make sense, especially when all your use has you calling `authentication.Get()`. We know this is dated and needs to be removed, let's leave `Get()` calls alone for this PR.

Naatan (changes requested):

Naatan (approved):

PR #1349, opened 2021-05-11
State clean uninstall removes config dir and binary file

MDrakos:
https://www.pivotaltracker.com/story/show/177956680


internal/runners/clean/run_win.go
```diff
@@ -4,34 +4,60 @@ package clean
...
+
+	args := []string{"/C", sf.Filename(), fmt.Sprintf("%d", os.Getpid()), filepath.Base(exe)}
+	args = append(args, dirs...)
+	cmd := exec.Command("cmd.exe", args...)
+	cmd.SysProcAttr = osutils.SysProcAttrForBackgroundProcess()
 	err = cmd.Start()
 	if err != nil {
-		return err
+		return locale.WrapError(err, "err_clean_start", "Could not start remove direcotry script")
 	}
```
mdrohmann, line 61: nit: We could use the `exeutils.ExecuteAndForget` function here.

internal/runners/clean/run_win.go
```diff
@@ -4,34 +4,60 @@ package clean
...
+		return locale.WrapError(err, "err_clean_script", "Could not create new scriptfile")
 	}
 
-	cmd := exec.Command("cmd.exe", "/C", sf.Filename(), path)
+	exe, err := os.Executable()
+	if err != nil {
+		return locale.WrapError(err, "err_clean_executable", "Could not get executable name")
+	}
+
+	args := []string{"/C", sf.Filename(), fmt.Sprintf("%d", os.Getpid()), filepath.Base(exe)}
```
mdrohmann, line 54: I suppose this only ensures that the `state.exe` stopped running, but the `state-tray` and `state-svc` executables may be running and hence cannot be removed...  But I guess that is a problem that I have to deal with in my PR. :)

activestate.yaml
```diff
@@ -71,6 +71,7 @@ scripts:
     language: bash
     description: Install OS specific deps
     value: |
+      sudo apt-get update
```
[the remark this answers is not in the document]
MDrakos: I see that you've done that in your PR so I've used it here.

mdrohmann (dismissed):

internal/runners/clean/run_win.go
```diff
@@ -4,34 +4,57 @@ package clean
...
-	err = cmd.Start()
+	exe, err := os.Executable()
 	if err != nil {
-		return err
+		return locale.WrapError(err, "err_clean_executable", "Could not get executable name")
+	}
+
+	args := []string{"/C", sf.Filename(), fmt.Sprintf("%d", os.Getpid()), filepath.Base(exe)}
+	args = append(args, dirs...)
+	_, err = exeutils.ExecuteAndForget("cmd.exe", args...)
```
Naatan, line 55: This doesn't feel like it should be using `..AndForget()`. When `state clean` finishes I'd expect it to actually be done cleaning. If that's not possible for some reason we need to ensure that the UX properly conveys this.

Naatan (changes requested):

Naatan:
Oh, also; changelog entry please :)

internal/runners/clean/run_win.go
MDrakos: On Windows, we can't remove anything until the `state.exe` process has ended. I've updated the messaging for `state clean uninstall` on Windows though I'm not sure if that's what you had in mind.

mdrohmann: I am thinking about the optimal un-installation experience for Windows:

First of all I believe that we should have a roll-back mechanism in case anything during the un-installation fails.  That is true for all platforms and I created a ticket for that: https://www.pivotaltracker.com/story/show/178134918  But that is something different. :)

- On Windows we cannot remove the programme that we are currently running, so maybe we need to have some functionality, to know as well as possible that a script running in the background can actually remove the executable.  Maybe by checking that there are no other processes running the executable, and by moving it to a temporary directory (which should ensure that we have permissions to delete it)?
- Ideally, we would have an entry in the "Add/Remove application" in the Systems dialog...  Then, the OS would deal with all that crap.  But, of course, that means that we would have to publish entries to the registry with an installation database during the installation which is imitating MSI functionality...

internal/runners/clean/run_win.go
```diff
@@ -27,7 +27,7 @@ func (u *Uninstall) runUninstall() error {
 		return err
 	}
 
-	u.out.Print(locale.T("clean_success_message"))
+	u.out.Print(locale.Tl("clean_message", "Attempting to remove State Tool and related files"))
```
Naatan, line 30: This is still lacking UX. If we have to defer the deletion then we should at least tell the user how to find out if and why it failed.

Naatan (changes requested):

changelog.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.29.1
```
Naatan, line 9: This shouldn't be a new entry. The version in version.txt is 0.29.0, so that's where your change entry should go.

internal/runners/clean/run_win.go
MDrakos: I don't think we're able to tell them exactly why it failed as this script will execute after the state tool process has completed.  I've tried to pass off `stdout` to the script and release the process but it appears that that will keep that `state.exe` process alive. Added some more information to the uninstall messaging for likely reasons why uninstall may fail and also how to verify that it actually removed the install.

Naatan: Can we have it write to a log file, and tell the user where it's located?

MDrakos: 💡

internal/runners/clean/run_lin_mac.go
```diff
@@ -29,21 +30,27 @@ func (u *Uninstall) runUninstall() error {
...
+		return locale.WrapError(err, "err_clean_close", "Could not close logging file")
 	}
 
-	return os.RemoveAll(configPath)
+	err = os.RemoveAll(configPath)
+	if err != nil {
+		return locale.WrapError(err, "err_clean_config_remove", "Could not remove config directory")
+	}
+
+	out.Print(locale.Tl("clean_config_succes", "Succsessfully removed State Tool config directory"))
```
Naatan, line 52:
```suggestion
	out.Print(locale.Tl("clean_config_succes", "Successfully removed State Tool config directory"))
```

locale/en-us.yaml
```diff
@@ -1318,7 +1318,12 @@ clean_message_windows:
   other: |
     Attempting to remove State Tool and related files.
     To verify that the State Tool has been uninstalled please check the files at: [NOTICE]{{.V0}}[/RESET] and [NOTICE]{{.V1}}[/RESET].
-    If the files have not been removed please verify you have the correct access rights on the above directories and ensure no State Tool processes are running, then try [ACTIONABLE]state clean uninstall[/RESET] again
+    To access a log of the uninstallation please see: [NOTICE]{{.V2}}[/RESET]
+clean_config_message_windows:
+  other: |
+    Attempting to remove State Tool config directory.
+    To verify that the config directory has been removed please check [NOTICE]{{.V0}}[/RESET].
+    To access a log of the process please see: [NOTICE]{{.V1}}[/RESET]
```
Naatan, line 1326:
```suggestion
    Deletion of State Tool config directory has been scheduled.
    To access a log of the process please see: [NOTICE]{{.V0}}[/RESET]
```

internal/runners/clean/run_win.go
```diff
@@ -22,20 +24,31 @@ func (u *Uninstall) runUninstall() error {
...
 
-func removeConfig(configPath string) error {
-	return removePaths(configPath)
+func removeConfig(configPath string, out output.Outputer) error {
+	logFile, err := ioutil.TempFile("", "state-clean-config")
+	if err != nil {
+		return locale.WrapError(err, "err_clean_logfile", "Could not create temporary log file")
+	}
+
+	out.Print(locale.Tr("clean_config_message_windows", configPath, logFile.Name()))
```
Naatan, line 47:
```suggestion
	out.Print(locale.Tr("clean_config_message_windows", logFile.Name()))
```

Naatan: Related suggestion further down (I don't think we need to be telling them the configPath, that should be in the log).

internal/runners/clean/run_win.go
```diff
@@ -4,34 +4,70 @@ package clean
...
+	if err != nil {
+		return locale.WrapError(err, "err_clean_logfile", "Could not create temporary log file")
+	}
+
+	err = removePaths(logFile.Name(), u.installPath, u.cfg.ConfigPath())
+	if err != nil {
+		return err
+	}
+
+	u.out.Print(locale.Tr("clean_message_windows", u.installPath, u.cfg.ConfigPath(), logFile.Name()))
```
Naatan, line 37:
```suggestion
	u.out.Print(locale.Tr("clean_message_windows", logFile.Name()))
```

locale/en-us.yaml
```diff
@@ -1314,6 +1314,16 @@ clean_config_confirm:
   other: You are about the remove the State Tool config directory. Continue?
 clean_success_message:
   other: "Successfully removed State Tool and related files\n"
+clean_message_windows:
+  other: |
+    Attempting to remove State Tool and related files.
+    To verify that the State Tool has been uninstalled please check the files at: [NOTICE]{{.V0}}[/RESET] and [NOTICE]{{.V1}}[/RESET].
+    To access a log of the uninstallation please see: [NOTICE]{{.V2}}[/RESET]
```
Naatan, line 1321:
```suggestion
    Deletion of State Tool has been scheduled.
    To access the uninstall log please see: [NOTICE]{{.V0}}[/RESET]
```

Naatan (changes requested):

mdrohmann (approved):

Naatan (approved):

PR #1350, opened 2021-05-11
State clean uninstall uninstalls multi-file updates

mdrohmann:
https://www.pivotaltracker.com/story/show/177448021


internal/runners/clean/run.go
```diff
@@ -12,7 +15,7 @@ func (u *Uninstall) runUninstall() error {
 		return err
 	}
 
-	err = removeInstall(u.installPath)
+	err = removeInstall(u.installDir)
```
MDrakos, line 18: I think this function could fail on Windows as it would be unable to remove the installation directory if the tray or service is still running. We may need to kill the tray and stop the service before attempting to remove them.

MDrakos (changes requested):

internal/runners/clean/run.go
mdrohmann: Yes, I think I have to merge with your PR first for this to work... :)

internal/runners/clean/run.go
```diff
@@ -33,3 +38,16 @@ func removeCache(cachePath string) error {
...
 }
+
+func undoPrepare(out output.Outputer) {
+	toRemove := prepare.InstalledPreparedFiles()
+
+	for _, f := range toRemove {
+		if fileutils.TargetExists(f) {
+			err := os.Remove(f)
+			if err != nil {
+				out.Notice(locale.Tl("[ERROR]Warning: [/RESET] Could not remove file {{.V0}}.", f))
```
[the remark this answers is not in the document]
mdrohmann: I changed the entire behavior, because it was already inconsistent (sometimes we stop the un-installation half-way and sometimes we delete as much as possible).

Now: If an error occurs at any point, we are still trying to remove as much as possible.

We also report the files that could not be removed, such that the user can remove them manually.  IMO that appears to be a better approach than leaving the user with a half-way unusuble installation and no way to fix it or re-try it (in case the State Tool was already removed).

An even cleaner way would be functionality to roll back the un-installation if anything fails, and I filed a follow-up story for that: https://www.pivotaltracker.com/story/show/178134918

internal/runners/clean/run_lin_mac.go
```diff
@@ -25,6 +29,31 @@ func removeConfig(cfg configurable) error {
...
+				continue
+			}
+			aggErr = errs.Wrap(aggErr, "Could not remove %s: %v", info.Exec(), err)
+		}
+	}
+
+	appPath, err := installation.LauncherInstallPath()
+	if err != nil {
+		return errs.Wrap(aggErr, "Could not determine OS specific launcher install path")
+	}
```
MDrakos, line 52: nit: we could move this to the top of the function and if an error is encountered here fail early without having removed anything from the user's system.

internal/osutils/autostart/autostart_linux.go
```diff
@@ -55,21 +55,30 @@ func (a *App) Enable() error {
...
-func (a *App) Disable() error {
+func (a *App) Path() (string, error) {
 	enabled, err := a.IsEnabled()
 	if err != nil {
-		return errs.Wrap(err, "Could not check if app autostart is enabled")
+		return "", errs.Wrap(err, "Could not check if app autostart is enabled")
 	}
 	if !enabled {
-		return nil
+		return "", nil
```
Naatan, line 64: This feels inappropriate for `.Path()`, why should it care whether it is enabled? You could want to use this to set up the path.

internal/runners/clean/run.go
```diff
@@ -3,33 +3,63 @@ package clean
...
 	}
 
 	u.out.Print(locale.T("clean_success_message"))
 	return nil
 }
 
+func (u *Uninstall) reportError(msg string, err error) {
+	logging.Error("%s: %v", msg, errs.Join(err, ": "))
+	u.out.Notice(msg)
+}
```
MDrakos, line 41: If we encountered errors during the uninstall we may want some sort of messaging at the end telling the user what went wrong and how to fix it. Either removing it manually or fixing something and attempting the command again. I guess this could also be part of the individual error notices.

internal/runners/clean/run.go
```diff
@@ -3,33 +3,63 @@ package clean
...
 	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/runners/prepare"
 )
 
 func (u *Uninstall) runUninstall() error {
 	err := removeCache(u.cfg.CachePath())
 	if err != nil {
-		return err
+		u.reportError(locale.Tl("uninstall_remove_cache_err", "Failed to remove cache directory."), err)
```
Naatan, line 16: This breaks our regular error reporting, which has UX styling. It'd be more appropriate to combine these errors and return them at the end.

Naatan (changes requested):

MDrakos (approved):
Looks good to me. My comments are mainly nits/suggestions.

MDrakos (approved):

Naatan (approved):

PR #1352, opened 2021-05-12
Implemented `svcmanager`

Naatan:
https://www.pivotaltracker.com/story/show/178013203


cmd/state/main.go
```diff
@@ -127,6 +128,11 @@ func run(args []string, isInteractive bool, out output.Outputer) error {
 	machineid.SetErrorLogger(logging.Error)
 	logging.UpdateConfig(cfg)
 
+	svcm := svcmanager.New(cfg)
+	if err := svcm.Start(); err != nil {
```
mdrohmann, line 132: Why don't we have to wait here for it to become ready?

internal/svcmanager/svcmanager.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (m *Manager) Start() error {
+	svcInfo := appinfo.SvcApp()
+	if !fileutils.FileExists(svcInfo.Exec()) {
+		return errs.New("Could not find: %s", svcInfo.Exec())
+	}
+
+	if _, err := exeutils.ExecuteAndForget(svcInfo.Exec(), []string{"start"}); err != nil {
+		return errs.Wrap(err, "Could not start %s", svcInfo.Exec())
+	}
```
mdrohmann, line 37: I am realizing that we will need an IPC lock ie., a file locking mechanism for this block: https://github.com/ActiveState/cli/blob/9498ec5a8029d4b01c15514db21bcb84dad5d615/cmd/state-svc/manager.go#L31-L51

If two managers are started at the same time, they both might find out that the State Tool Service is not running yet, before they write their newly started process's PID in the config file.  Then, we end up with two state-svc services at the same time.

This will also create problem with updates, as one of the State Tool processes will be stray, as its PID has been overwritten in the configuration file.

mdrohmann (changes requested):

cmd/state/main.go
Naatan: Because that would slow down every single state tool invocation. If we have a command that depends on this we can pass the `svcm` down to it and then it cal `Wait()`.

internal/svcmanager/svcmanager.go
Naatan: I'll file a story, feel like that's separate from what this PR is trying to do.

Naatan: https://www.pivotaltracker.com/story/show/178150478

mdrohmann (approved):

PR #1354, opened 2021-05-13
Update silent error check

MDrakos:
https://www.pivotaltracker.com/story/show/177863704

Moved the check for a silent error as the `unwrapError` function returns a `nil` error if it encounters a silent error. https://github.com/ActiveState/cli/blob/master/cmd/state/errors.go#L90


changelog.md
```diff
@@ -8,6 +8,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## 0.29.0
 
+
+### Fixed
+
+- Removed unwanted output when error is silent
```
Naatan, line 14:
```suggestion
- Removed unwanted output (eg. `%!s(<nil>)`) when running scripts
```

cmd/state/main.go
```diff
@@ -84,8 +84,9 @@ func main() {
 	// Run our main command logic, which is logic that defers to the error handling logic below
 	err = run(os.Args, isInteractive, out)
 	if err != nil {
+		silent := isSilent(err)
```
Naatan, line 87:
```suggestion
```

cmd/state/main.go
```diff
@@ -84,8 +84,9 @@ func main() {
...
 		exitCode, err = unwrapError(err)
-		if !isSilent(err) {
+		if !silent {
```
Naatan, line 89:
```suggestion
		if err != nil {
```

Looking at the code of `unwrapError`, this feels more appropriate.

Naatan (changes requested):

Naatan (dismissed):

Naatan (approved):

PR #1356, opened 2021-05-13
Ensure project pkg mutations update asy commit id

daved:
https://www.pivotaltracker.com/story/show/177550503


daved:
I skipped resetting the commit id to that of the parent commit id if no changes were found based on my guess/assumption that that was being done as a sort of pseudo caching UX. Now that runtimes are cached effectively, I think this is no longer needed. Please correct me if I'm wrong.

Naatan (approved):

PR #1357, opened 2021-05-14
ensure that failure / success event is sent for deploy install

mdrohmann:
https://www.pivotaltracker.com/story/show/178134247

### DIFF
--- a/internal/runners/deploy/deploy.go
+++ b/internal/runners/deploy/deploy.go
@@ -157,6 +157,10 @@ func (d *Deploy) install(rtTarget setup.Targeter) error {
 		return locale.WrapError(err, "deploy_install_failed", "Installation failed.")
 	}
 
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
This looks like a workaround. I think if we wanna land the fix this way we should ensure we have a followup story scheduled for next sprint, as this feels way too easy to break.

mdrohmann:
How about this story: https://www.pivotaltracker.com/story/show/178161240

I am fairly confident that is the main culprit though.  I was looking through our GA data yesterday afternoon, searching for sessions that have a `runtime,start` but no finishing event and I found only cases of `state deploy install`.

Naatan (approved):

Naatan:
Awesome, thanks @mdrohmann! Fingers crossed this is it.

PR #1358, opened 2021-05-14
Use correct installation path for state update

mdrohmann:
First problem found and fixed during testing: https://www.pivotaltracker.com/story/show/178161668

### DIFF
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -8,11 +8,11 @@ import (
 	"os/exec"
 	"path/filepath"
 
+	"github.com/ActiveState/cli/internal/appinfo"
 	"github.com/ActiveState/cli/internal/constants"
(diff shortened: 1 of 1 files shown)


internal/updater/updater.go
```diff
@@ -50,7 +51,17 @@ func (u *AvailableUpdate) prepare() (string, string, error) {
 	if !fileutils.FileExists(installerPath) {
 		return "", "", errs.Wrap(err, "Downloaded update does not have installer")
 	}
-	installTargetPath := filepath.Dir(os.Args[0])
+
+	exec, err := os.Executable()
```
Naatan, line 55: This should probably be using `appinfo.<something>.Exec()`

Naatan (changes requested):

Naatan (approved):

PR #1360, opened 2021-05-14
Added updated notifications, removed auto-update

Naatan:


internal/svcmanager/svcmanager.go
```diff
@@ -16,6 +16,9 @@ import (
 	"golang.org/x/net/context"
 )
 
+// MinimumTimeout is the minimum timeout required for requests that are meant to be near-instant
+const MinimumTimeout = 500 * time.Millisecond
```
mdrohmann, line 20: nit: This name confused me at first: I though that it defines a minimum for our timeouts, and all others should be bigger than that.  Maybe we could name it `MinimalTimeout`, `TinyTimeout` or something like that...

mdrohmann (approved):

mdrohmann (dismissed):

mdrohmann (approved):

PR #1368, opened 2021-05-18
Add account preferences link to tray acct menu item

daved:
https://www.pivotaltracker.com/story/show/178161371


daved:
The primary url for account info/settings is "https://platform.activestate.com/preferences", and that is where the current link is directing. I was wondering if it would be preferable to, instead, land at "https://platform.activestate.com/preferences/account-settings".

Naatan (approved):

PR #1370, opened 2021-05-18
Fix WaitForLock() and config.Set() lock

mdrohmann:
https://www.pivotaltracker.com/story/show/177913373


MDrakos (changes requested):
Looks good to me just some minor cleanup

changelog.md
```diff
@@ -16,6 +16,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   will from now on be made available through the system tray application and you
   will have the choice to install it now or later.
 
+### Fixed
+
+- Fixed logic checking for deprecations.
```
Naatan, line 21: Remember to phrase this with end users in mind. What does this fix mean practically for end-users?

MDrakos (dismissed):

Naatan (changes requested):

Naatan (approved):

mdrohmann:
Okay, this works now, I had to re-iterate because:
-  there was another bug in the `pidlock` implementation that only happened on Windows:
When a process successfully locks the lockfile, but the file contents have a PID that is still running, `TryLock` returned `false`, but the file was locked (and never released).
- Next problem: I realized that I wanted to just remove the PID check. I remember that I added that in order to ensure that we can recognize lock files that try to lock processes that are not running anymore, but that is actually already handled by `flock` or `LockProcessEx`. So, it is unnecessary, and unnecessary complexity leads to errors, *BUT*:
   - in `process.NewActivation` we use the PID functionality (but not the actually locking)
   - the PID check actually did something: it ensured some INNER process synchronization. If `TryLock` was run in the same process, the `flock.TryLock` call always succeeds (that's only on Linux and Mac), because it allows several locks by the same process(!).  But as soon as the first `TryLock` call  writes its PID in there, the second one notices that there is already a running PID.  => So, our locking process needs an inner-process mutex to synchronize for these cases. 

The suggested solution implemented in this PR is as follows:
1. As the external package `gofrs/flock` already ensures inner process synchronization, I used that for the config synchronization
2. I fixed the `pidlock` implementation as much as I could
3. I filed a story to get rid of the pidlock implementation: https://www.pivotaltracker.com/story/show/178223007
4. Because `gofrs/flock` provides shared and exclusive file locking, I changed the lock for `Reload()` to a shared lock ie., it can be run in parallel by several processes.  It was easy and should speed things up, especially as we know have several processes contending for resources.

changelog.md
```diff
@@ -16,6 +16,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   will from now on be made available through the system tray application and you
   will have the choice to install it now or later.
 
+### Fixed
+
+- Fixed a rare case where deprecation information was not downloaded, when
+  requested by several State Tool instances in parallel. This also prevents the
+  configuration file from getting corrupted when two processes access it
+  simulataneously.  ([PR #1370] (https://github.com/ActiveState/cli/pull/1370))
```
Naatan, line 24:
```suggestion
- Fixed a rare case where the configuration file can get corrupted when two processes access it
  simultaneously.  ([PR #1370] (https://github.com/ActiveState/cli/pull/1370))
```

Naatan (changes requested):

Naatan (approved):

PR #1371, opened 2021-05-18
Add tip to check path if script language could not be found

mdrohmann:
https://www.pivotaltracker.com/story/show/178186713

I checked it on `beta` and `release` channel versions, and noticed that the error only shows for me, if `bash.exe` is not on the PATH.

@Naatan Can you please confirm that this is the problem, or specify which `bash.exe` we did not find?

### DIFF
--- a/internal/scriptrun/scriptrun.go
+++ b/internal/scriptrun/scriptrun.go
@@ -113,6 +113,7 @@ func (s *ScriptRun) Run(script *project.Script, args []string) error {
 	}
 
 	var attempted []string
+	var attempted3rdParty []string
 	for _, l := range script.Languages() {
 		execPath := l.Executable().Filename()
 		searchPath := s.venvExePath
@@ -127,6 +128,9 @@ func (s *ScriptRun) Run(script *project.Script, args []string) error {
 			break
 		}
 		attempted = append(attempted, l.String())
+		if l.Executable().CanUseThirdParty() {
+			attempted3rdParty = append(attempted3rdParty, l.Executable().Filename())
+		}
 	}
 
 	if script.Standalone() && !lang.Executable().CanUseThirdParty() {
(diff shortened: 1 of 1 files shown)


internal/scriptrun/scriptrun.go
```diff
@@ -135,12 +135,14 @@ func (s *ScriptRun) Run(script *project.Script, args []string) error {
...
-				strings.Join(language.RecognizedNames(), ", "),
-			)
+			return errs.AddTips(
+				locale.NewInputError(
+					"err_run_unknown_language_fallback",
+					"The language for this script is not supported or not available on your system. Attempted script execution with: {{.V0}}. Please configure the 'language' field with an available option (one, or more, of: {{.V1}})",
+					strings.Join(attempted, ", "),
+					strings.Join(language.RecognizedNames(), ", "),
+				),
+				locale.Tl("unknown_language_check_path", "If the language is not provided by the project runtime, please ensure that the required executable is discoverable via your PATH environment variable"))
```
Naatan, line 145: This is too long for a tip, keep it short and concise, eg. "Please ensure that bash is on your PATH".

internal/scriptrun/scriptrun.go
```diff
@@ -135,12 +135,14 @@ func (s *ScriptRun) Run(script *project.Script, args []string) error {
 
 	if lang == language.Unknown {
 		if len(attempted) > 0 {
-			return locale.NewInputError(
-				"err_run_unknown_language_fallback",
-				"The language for this script is not supported or not available on your system. Attempted script execution with: {{.V0}}. Please configure the 'language' field with an available option (one, or more, of: {{.V1}})",
-				strings.Join(attempted, ", "),
-				strings.Join(language.RecognizedNames(), ", "),
-			)
+			return errs.AddTips(
```
Naatan, line 138: We should only show this if `allowThirdParty=true`; http://github.com/ActiveState/cli/blob/9f570a0bdce116650606e8cfcc04cb9ecac3b64c/internal/language/language.go#L74-L74

There's no point in showing this tip if someone is using a platform provided language.

Naatan (changes requested):
Nice digging! I can confirm `bash` was not on my PATH.

Naatan (approved):
Nice work!

PR #1375, opened 2021-05-19
Installers only pass installation path if one is provided

MDrakos:
https://www.pivotaltracker.com/story/show/178165310


installers/install.sh
```diff
@@ -308,72 +308,21 @@ fetchArtifact () {
...
-      mkdir -p "$INSTALLDIR" || continue
-    fi
     fetchArtifact
-    info "Installing to $INSTALLDIR..."
-    $TMPDIR/$TMPEXE "$INSTALLDIR"
+    if [ ! -z "$TARGET" ]; then
+      INSTALL_OUTPUT=$($TMPDIR/$TMPEXE "$TARGET")
+    else
+      echo "Installing without target"
+      INSTALL_OUTPUT=$($TMPDIR/$TMPEXE)
```
mdrohmann, line 323: We should probably still print the `INSTALL_OUTPUT`. I also wonder if there is a way to print the installer output when it occurs and still capture it.

This might work: https://stackoverflow.com/a/12451419/4500584

And maybe there is a similar solution for powershell?

mdrohmann (changes requested):

installers/install.sh
MDrakos: A modification of your link worked for me in bash. Powershell should be working as well

installers/install.ps1
```diff
@@ -347,13 +347,13 @@ function install() {
 
     $InstallerPath = Join-Path -Path $tmpParentPath -ChildPath $installerexe
     if ($script:TARGET) {
-        $output = & "$InstallerPath" "$script:TARGET" 2>&1 | Out-String
+        & "$InstallerPath" "$script:TARGET" 2>&1 | Tee-Object -Variable output | Write-Host
```
mdrohmann, line 350: :+1:

installers/install.sh
```diff
@@ -308,72 +308,20 @@ fetchArtifact () {
...
   [Yy])
-    # Install.
-    if [ ! -e "$INSTALLDIR" ]; then
-      mkdir -p "$INSTALLDIR" || continue
-    fi
     fetchArtifact
-    info "Installing to $INSTALLDIR..."
-    $TMPDIR/$TMPEXE "$INSTALLDIR"
+    if [ ! -z "$TARGET" ]; then
+      INSTALL_OUTPUT=$($TMPDIR/$TMPEXE "$TARGET" 2>&1 | tee /dev/tty)
```
mdrohmann, line 320: I think /dev/tty does exist when run in a background process (that is not associated with any terminal).  I remember that we had some errors where `tput` did not work for the same reason.

I think it can be easily tested by running starting the install script with `setsid`: `setsid install.sh -f` would fail at this step with `/dev/tty: No such device or address`

mdrohmann (changes requested):

installers/install.sh
MDrakos: Good point. I've updated it again to use a temp file. I could not get the stack overflow example you linked to work as our script is using `sh` rather than `bash` so the part of the example where `cat` is started in a subshell kept giving me syntax errors. If you know of a way to make it work in `sh` I will update what I have currently.

installers/install.sh
```diff
@@ -308,72 +308,23 @@ fetchArtifact () {
...
 case "$RESPONSE" in
   [Yy])
-    # Install.
-    if [ ! -e "$INSTALLDIR" ]; then
-      mkdir -p "$INSTALLDIR" || continue
-    fi
     fetchArtifact
-    info "Installing to $INSTALLDIR..."
-    $TMPDIR/$TMPEXE "$INSTALLDIR"
+    exec 5>&1
```
mdrohmann, line 319:
```suggestion
```

installers/install.sh
```diff
@@ -308,72 +308,23 @@ fetchArtifact () {
...
+    OUTPUT_FILE=$TMPDIR/install_output.txt
+    if [ ! -z "$TARGET" ]; then
+      "$TMPDIR/$TMPEXE" "$TARGET" 2>&1 | tee $OUTPUT_FILE
+    else
+      "$TMPDIR/$TMPEXE" 2>&1 | tee $OUTPUT_FILE
+    fi
+    INSTALL_OUTPUT=$(cat $OUTPUT_FILE)
```
mdrohmann, line 326:
```suggestion
    INSTALL_OUTPUT=$(cat $OUTPUT_FILE)
    rm -f $OUTPUT_FILE
```

mdrohmann (changes requested):
:+1:

It probably makes more sense this way around: printing to stdout and capturing the output for later inspection.

mdrohmann (approved):

Naatan (approved):

PR #1376, opened 2021-05-19
Use functions for state script instead of aliases

shaunmlowry:
Using aliases in bash to add state tool scripts to the virtual
environments has a number of disadvantage.  One major drawback is that
these aliases are not available in subshells, or available to scripts.

That means doing things like:
```
$ myvar=$(mystatescript)
```
don't work.

Instead, create shell functions and export them with `export -f` so that
they can be used everywhere a regular command could be.

 https://www.pivotaltracker.com/story/show/178030472

### DIFF
--- a/assets/shells/bashrc.sh
+++ b/assets/shells/bashrc.sh
@@ -21,7 +21,10 @@ alias {{.ExecName}}='{{.ExecAlias}}'
 {{ end }}
 
 {{range $K, $CMD := .Scripts}}
-alias {{$K}}='{{$.ExecName}} run {{$CMD}}'
+function {{$K}} {
+    {{$.ExecName}} run {{$CMD}} "$@"
+}
+export -f {{$K}}
 {{end}}
 
 echo "{{.ActivateEventMessage}}"
--- a/test/integration/run_int_test.go
+++ b/test/integration/run_int_test.go
@@ -134,6 +134,33 @@ func (suite *RunIntegrationTestSuite) TestInActivatedEnv() {
 	)
 }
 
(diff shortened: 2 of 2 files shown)


assets/shells/bashrc.sh
```diff
@@ -21,7 +21,10 @@ alias {{.ExecName}}='{{.ExecAlias}}'
 {{ end }}
 
 {{range $K, $CMD := .Scripts}}
-alias {{$K}}='{{$.ExecName}} run {{$CMD}}'
+function {{$K}} {
+    {{$.ExecName}} run {{$CMD}} "$@"
+}
+export -f {{$K}}
```
Naatan, line 27: I don't think we need this `export`, given this code will effectively be injected into your bashrc

Naatan (changes requested):
Thanks! Do you know if this would also be relevant for zsh?

assets/shells/bashrc.sh
shaunmlowry: .bashrc isn't sourced for things like `bash -c` or scripts that start with `#!/bin/bash` so the export makes the commands available in those situations

shaunmlowry: the environment inherited is just the environment variables which *nix operating systems understand at a lower level (see `execle()`'s `char *const envp[]` argument for example).  It doesn't include shell-specific items like aliases and functions.  `export -f` is a special variant in bash which tells it to copy functions from the current shell into spawned _bash_ shells specifically.   See `man bash` for details on how `export -f` works

Naatan (approved):

PR #1382, opened 2021-05-20
Add update dialog to installed files

Naatan:
https://www.pivotaltracker.com/story/show/178235218


.github/workflows-src/steps.lib.yml
```diff
@@ -89,12 +68,7 @@ run: |
...
   signtool.exe sign -d "ActiveState State Tray" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/state-tray.exe 
   signtool.exe sign -d "ActiveState State Installer" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/state-installer.exe 
-  signtool.exe sign -d "ActiveState State Tool (32bit)" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/386/state.exe 
-  signtool.exe sign -d "ActiveState State Service (32bit)" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/386/state-svc.exe 
-  signtool.exe sign -d "ActiveState State Tray (32bit)" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/386/state-tray.exe 
-  signtool.exe sign -d "ActiveState State Installer (32bit)" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/386/state-installer.exe 
-  signtool.exe sign -d "ActiveState Transitional State Tool" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/state-transition-update.exe 
-  signtool.exe sign -d "ActiveState Transitional State Tool (32bit)" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/386/state-transition-update.exe 
+  signtool.exe sign -d "ActiveState Transitional State Tool" -f "Cert.p12" -p ${CODE_SIGNING_PASSWD} ./build/state-transition-update.exe
```
mdrohmann, line 71: Should we sign the update dialog too?

mdrohmann (changes requested):
I just think that we should probably also sign the update dialog executable.

mdrohmann (approved):
I think you mentioned that you wanted  to make a new beta release with this change.  I would suggest to wait for #1383 to merge, such that it can also go into that beta release.  Otherwise people's configuration files will break on a very regular basis.

PR #1383, opened 2021-05-20
Guard state-svc start by file lock

mdrohmann:
https://www.pivotaltracker.com/story/show/178150478

This also fixes:
https://www.pivotaltracker.com/story/show/178234870


internal/config/instance.go
```diff
@@ -169,7 +176,31 @@ func Get() (*Instance, error) {
...
+// The caller needs to explicitly guard this function by calling
+// GetLock()/ReleaseLock().
+// Example usage:
+//    cfg.GetLock()
+//    defer cfg.ReleaseLock()
+//    cfg.ReloadUnsafe()
+//    slice, _ := cfg.GetStringSlice("key")
+//    slice = append(slice, "value")
+//    cfg.SetUnsafe("key", slice)
+// Note, how in this example it is crucial to synchronize the appending to the slice variable.
```
MDrakos, line 190: Nit: In the docstring for `ReloadUnsafe` it says to see this docstring for when and how to use this function, however, this docstring is explaining how and not when. I can see from how you are using it in this PR that you want to hold the config lock until the service has started so no other services can start so we may want to add some more detail to this already long docstring. Or remove the reference in the `ReloadUnsafe` docstring

MDrakos (approved):

cmd/state-svc/manager.go
```diff
@@ -19,15 +18,25 @@ import (
...
 
 func (s *serviceManager) Start(args ...string) error {
+	err := s.cfg.GetLock()
+	if err != nil {
+		return errs.Wrap(err, "Failed to get configuration lock")
+	}
+	defer s.cfg.ReleaseLock()
+
+	// Reload the configuration to ensure that we have the latest pid information
+	err = s.cfg.ReloadUnsafe()
```
Naatan, line 36: This feels unnecessary. We should be able to record that we are already locked, and as such don't lock again.

Naatan (changes requested):

cmd/state-svc/manager.go
mdrohmann: I hear you, was thinking about that myself, but it is more difficult to do than it sounds, because the mutex also synchronizes behavior between threads inside the same process.  And when we have to check whether we are already locked, I don't know how to differentiate between the following two cases:

- The process holds the lock already and the call comes from the same thread => we can proceed
- The process holds the lock already, but the call comes from a different thread => we need to wait until the other thread releases the lock.

mdrohmann: I have a different idea to make this cleaner though.  Let me implement and push, so we can discuss it.

mdrohmann: This proposal would add a `SetSafe()` method to the config instance.  It is more explicitly targeted to the use case of manipulating values that depend on previous values.

cmd/state-svc/manager.go
```diff
@@ -19,44 +19,53 @@ import (
...
+	return &serviceManager{cfg}
 }
 
 func (s *serviceManager) Start(args ...string) error {
-	curPid, err := s.Pid()
-	if err == nil && curPid != nil {
-		return errs.New("Service is already running")
-	}
+	var proc *os.Process
+	err := s.cfg.SetSafe(constants.SvcConfigPid, func(oldPidI interface{}) interface{} {
```
Naatan, line 31:
```suggestion
	err := s.cfg.SetSafe(constants.SvcConfigPid, func(oldPidI interface{}) interface{} {
```

I like this approach, but I think putting this on `Set` is a bit confusing and limiting. What about a `cfg.WithLock(func() { ... })` ?

Also, I'd like to reserve the `unsafe` naming for things used in tests. `unsafe` should signal a big red flag and probably shouldn't be in production code.

Naatan: Another thought; maybe it's not appropriate to rely on the config locking mechanic here at all. The config locking should be an internal mechanic to locking. 

I think we should be trying to solve this in a way that doesn't require any special knowledge from code outside of the config package, though I recognize that's a bit out of scope for the current story. I'm ok landing this with a minimal changeset and filing a followup though.

Naatan (changes requested):

cmd/state-svc/manager.go
```diff
@@ -5,65 +5,84 @@ import (
...
 	if err != nil {
 		return errs.Wrap(err, "Could not get pid")
 	}
 	if pid == nil {
 		return nil
 	}
 
+	waitMgr := svcmanager.New(s.cfg)
+	err = waitMgr.Wait()
+	if err != nil {
```
Naatan, line 79: Why is this necessary? We already checked the PID.

Additionally this is assuming the meaning of `err`.

cmd/state-svc/manager.go
```diff
@@ -5,65 +5,84 @@ import (
...
+	return &serviceManager{cfg}
 }
 
 func (s *serviceManager) Start(args ...string) error {
-	curPid, err := s.Pid()
-	if err == nil && curPid != nil {
-		return errs.New("Service is already running")
-	}
+	var proc *os.Process
+	err := s.cfg.SetWithLock(constants.SvcConfigPid, func(oldPidI interface{}) (interface{}, error) {
```
Naatan, line 32: Please file a followup story to remove this. For the purpose of this story we can go ahead with this approach, but my ideal is still that consuming code has no awareness of locking.

cmd/state-svc/manager.go
mdrohmann: Yes, I am a little unsure about this:

We could certainly introduce a `state-svc.lock` lockfile for this, that just ensures that `state-svc` is never called twice, but IMO it is also okay to say that we need to ensure that we need a lock on the `constants.SvcConfigPid` key.  As this should not change between reading the old value and writing the new value.  Here, I opted for the latter and helped fix some other bugs :-).  But I could also be convinced that a second lock file is the way to go.

Naatan (changes requested):

cmd/state-svc/manager.go
mdrohmann: I am pretty sure that this happened in one of our integration test runs at some point.

The PID value had already been written, but the server has not been started yet ie., it has not written the PORT value to the configuration file. So, the `waitMgr` just waits for the process to start up such that it can talk to it via the port...

I am fine with removing it for now, and we should just be vigilant and see if it occurs again.  It may not be a real issue...

mdrohmann: I am pretty sure that this happened in one of our integration test runs at some point.

The PID value had already been written, but the server has not been started yet ie., it has not written the PORT value to the configuration file. So, the `waitMgr` just waits for the process to start up such that it can talk to it via the port...

I am fine with removing it for now, and we should just be vigilant and see if it occurs again.  It may not be a real issue...

Naatan: This is on "Stop" though?

mdrohmann: Yes, it is kind of silly to wait for the service to start in order to stop it.  But I saw at least one instance where it tried to do that on port `0`.  And that indicates that the port number was not written to the configuration yet...

cmd/state-svc/manager.go
mdrohmann: https://www.pivotaltracker.com/story/show/178253100

cmd/state-svc/manager.go
Naatan: Let's remove this. It sounds to me like this is here to work around an issue we don't fully understand, I'd rather we understand the issue and solve it properly. I can't imagine a legitimate use-case for this.

Naatan (changes requested):

mdrohmann:
Well, maybe we need the change: :-(

- https://github.com/ActiveState/cli/pull/1383/checks?check_run_id=2642804108#step:17:114
- https://github.com/ActiveState/cli/pull/1383/checks?check_run_id=2642804129#step:17:114

Naatan (approved):

PR #1386, opened 2021-05-24
Ensure runtime is updated if executors are missing despite payload being cached

daved:
https://www.pivotaltracker.com/story/show/178152827

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -7,6 +7,7 @@ import (
 	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
@@ -40,7 +41,9 @@ func newRuntime(target setup.Targeter) (*Runtime, error) {
 		store:  store.New(target.Dir()),
 	}
 
-	if setup.Required(target.Dir()) {
+	// added in 202105 for transition (invalidates setups produced by older
+	// tool versions): https://www.pivotaltracker.com/story/show/178152827
+	if !fileutils.DirExists(setup.ExecDir(target.Dir())) {
 		return rt, &NeedsUpdateError{errs.New("Runtime requires setup.")}
 	}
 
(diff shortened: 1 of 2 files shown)


mdrohmann (approved):

Naatan (changes requested):
This looks like a workaround to me. How did we end up in a state where `rt.store.MatchesCommit(target.CommitUUID())` returns true but the executor dir did not exist?

daved:
I took the commit sanity check to mean that the cache is loaded properly. The executors seem fundamentally different in that regard, so a separate check makes sense to me.

Naatan:
Executors are created here: http://github.com/ActiveState/cli/blob/0ec4186682f908080e65343ecf63f4e0c7a2c8b3/pkg/platform/runtime/setup/setup.go#L222-L222

The the commit is saved a few lines further down at: http://github.com/ActiveState/cli/blob/0ec4186682f908080e65343ecf63f4e0c7a2c8b3/pkg/platform/runtime/setup/setup.go#L253-L253

Regardless of whether the store and executors are fundamentally different, logically the executors should always exist with the current implementation. The fact that it doesn't is concerning and we should ensure that we understand why it's bugged.

mdrohmann:
@Naatan This only happens when you have a runtime activated by an old State Tool that hadn't had the executor logic in there yet.  With that old State Tool, the runtime directory would be set up with out the `exec` dir, but the `complete` marker.

Once, they update the State Tool, the runtime will stop working, because we only add the `exec` directory to the PATH which points to a void space...

mdrohmann:
So, it is transitional, and we should probably remove that check after a few months...

Naatan:
Ok, in that case can we drop the public functions and just have some really to-the-point code where it needs to be with a comment pointing to a PT ticket saying we should drop this after a while.

daved:
`filepath.Join(s.target.Dir(), "exec")` was like nails on a chalkboard for me, so I left in the setup.ExecDir func for now (will remove it if you'd like).

pkg/platform/runtime/runtime.go
```diff
@@ -36,9 +36,17 @@ func IsNeedsUpdateError(err error) bool {
...
 
 func newRuntime(target setup.Targeter) (*Runtime, error) {
-	rt := &Runtime{target: target}
+	rt := &Runtime{
+		target: target,
+		store:  store.New(target.Dir()),
+	}
+
+	// added in 202105 for transition (invalidates setups produced by older
+	// tool versions): https://www.pivotaltracker.com/story/show/178152827
```
Naatan, line 45: This should link to the followup story to remove this code in due time.

Naatan (changes requested):

Naatan (approved):

mdrohmann (approved):

PR #1387, opened 2021-05-25
Stop services when switching channel on update

MDrakos:
https://www.pivotaltracker.com/story/show/178207127

### DIFF
--- a/internal/runners/update/update.go
+++ b/internal/runners/update/update.go
@@ -3,10 +3,13 @@ package update
 import (
 	"context"
 	"os"
+	"path/filepath"
 
+	"github.com/ActiveState/cli/internal/appinfo"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/installation"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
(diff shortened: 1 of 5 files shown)


internal/runners/update/update.go
```diff
@@ -63,6 +66,13 @@ func (u *Update) Run(params *Params) error {
 		return nil
 	}
 
+	if up.Channel != constants.BranchName {
```
mdrohmann, line 69: So, we have to stop the running services from the calling State Tool, because the configuration directory changes, right?
Could you add a comment to explain this here?  I think that in a few months, we may have forgotten about those details and will wonder why this is here...

mdrohmann (approved):

internal/runners/update/update.go
MDrakos: Sure, I think that makes sense.

Initially, I tried to implement this as part of the deferred update but that leads to stopping the service before the request can be completed. The only other place I think this could live is in the installer itself but that would require some refactoring. I can file a follow-up story to discuss this.

mdrohmann (approved):

internal/runners/update/update.go
MDrakos: When we switch channels the config directory changes. The invoked installer sources its config from the new directory and will attempt to close any running applications from that channel, leaving the original ones running. This is what leads to the duplicate instances of the system tray and service when switching channels.

The installer could do this but we would have to make it aware of the original channel. I mentioned in the comment above that I can file a follow-up story for this if it's worth looking into.

Naatan: Hmm no I think this solution is actually more appropriate, since the shutdown logic will be owned by the version that introduced the code that's being shut down.

Thanks for explaining!

Naatan (approved):

PR #1390, opened 2021-05-26
State reset

MDrakos:
https://www.pivotaltracker.com/story/show/178222863

### DIFF
--- a/cmd/state/internal/cmdtree/reset.go
+++ b/cmd/state/internal/cmdtree/reset.go
@@ -12,8 +12,8 @@ func newResetCommand(prime *primer.Values) *captain.Command {
 
 	return captain.NewCommand(
 		"reset",
-		locale.Tl("reset_title", "Restting commit"),
-		locale.Tl("reset_description", "Reset to the most recent remote commit"),
+		locale.Tl("reset_title", "Reset to Same Commit as Platform Project"),
+		locale.Tl("reset_description", "Reset local checkout to be equal to the project on the platform."),
 		prime.Output(),
 		prime.Config(),
 		[]*captain.Flag{},
--- a/cmd/state/internal/cmdtree/reset.go
+++ b/cmd/state/internal/cmdtree/reset.go
@@ -1,0 +1,25 @@
+package cmdtree
+
+import (
+	"github.com/ActiveState/cli/internal/captain"
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/primer"
+	"github.com/ActiveState/cli/internal/runners/reset"
+)
+
+func newResetCommand(prime *primer.Values) *captain.Command {
+	runner := reset.New(prime)
+
+	return captain.NewCommand(
+		"reset",
+		locale.Tl("reset_title", "Restting commit"),
+		locale.Tl("reset_description", "Reset to the most recent remote commit"),
+		prime.Output(),
+		prime.Config(),
+		[]*captain.Flag{},
+		[]*captain.Argument{},
+		func(ccmd *captain.Command, args []string) error {
+			return runner.Run()
+		},
+	).SetGroup(VCSGroup)
+}
--- a/internal/runners/reset/reset.go
+++ b/internal/runners/reset/reset.go
@@ -6,7 +6,6 @@ import (
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/prompt"
 	"github.com/ActiveState/cli/internal/runbits"
-	"github.com/ActiveState/cli/pkg/cmdlets/commit"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/project"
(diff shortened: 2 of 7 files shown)


cmd/state/internal/cmdtree/reset.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/ActiveState/cli/internal/primer"
+	"github.com/ActiveState/cli/internal/runners/reset"
+)
+
+func newResetCommand(prime *primer.Values) *captain.Command {
+	runner := reset.New(prime)
+
+	return captain.NewCommand(
+		"reset",
+		locale.Tl("reset_title", "Restting commit"),
```
Naatan, line 15:
```suggestion
		locale.Tl("reset_title", "Reset to Same Commit as Platform Project"),
```

cmd/state/internal/cmdtree/reset.go
```diff
@@ -0,0 +1,25 @@
...
+		locale.Tl("reset_description", "Reset to the most recent remote commit"),
```
Naatan, line 16:
```suggestion
		locale.Tl("reset_description", "Reset local checkout to be equal to the project on the platform."),
```

internal/runners/reset/reset.go
```diff
@@ -0,0 +1,96 @@
...
+
+	latestCommit, err := model.BranchCommitID(r.project.Owner(), r.project.Name(), r.project.BranchName())
+	if err != nil {
+		return locale.WrapError(err, "err_reset_latest_commit", "Could not get latest commit ID")
+	}
+	if *latestCommit == r.project.CommitUUID() {
+		return locale.NewInputError("err_reset_latest", "You are already on the latest commit")
+	}
+
+	r.out.Print(locale.Tl("reset_commit", "Your project will be reset at [ACTIONABLE]{{.V0}}[/RESET]\n", latestCommit.String()))
```
Naatan, line 59:
```suggestion
	r.out.Print(locale.Tl("reset_commit", "Your project will be reset to [ACTIONABLE]{{.V0}}[/RESET]\n", latestCommit.String()))
```

Let's include the date and author of this commit, since the ID alone isn't very helpful.

internal/runners/reset/reset.go
```diff
@@ -0,0 +1,96 @@
...
+		return locale.WrapError(err, "err_revert_refresh")
+	}
+
+	changesMade := len(revertCommit.Changeset) > 0
+	if changesMade {
+		changes := commit.FormatChanges(revertCommit)
+		r.out.Print(locale.Tl("reset_changes", "Changes to local instance:"))
+		r.out.Print(changes)
+		r.out.Print("")
+	}
```
Naatan, line 73: This has the same issue as I ran into; it only compares two specific commits, not all commits leading up to the specified commits. So let's remove this as it will just produce confusing output.

internal/runners/reset/reset.go
```diff
@@ -0,0 +1,96 @@
...
+
+	confirm, err := r.prompt.Confirm("", locale.Tl("reset_confim", "Continue?"), new(bool))
```
Naatan, line 75: "Continue?" is not a very useful prompt. Explain what they are being prompted for, eg.

> Resetting is destructive, you will loose any changes that were not pushed. Are you sure you want to do this?

Naatan (changes requested):

internal/runners/reset/reset.go
MDrakos: I would argue that showing the output of all commits would be more confusing than the revert commit changes. If I were to revert to the most recent remove commit I would want to know what the changes will be, not all of the changes I've made. But maybe you have something else in mind. Either way I've removed this.

Naatan: I'm not saying it should show the output of all commits, I'm saying it should show the difference between what's on the remote project and what's local. Your current code would only show the difference between two specific commits, rather than the full range of local commits (great if that only happens to be a single commit, but it might not be).

internal/runners/reset/reset.go
```diff
@@ -56,36 +55,28 @@ func (r *Reset) Run() error {
...
 		return locale.WrapError(err, "err_reset_set_commit", "Could not update commit ID")
 	}
 
-	err = runbits.RefreshRuntime(r.auth, r.out, r.project, r.config.CachePath(), *latestCommit, changesMade)
+	revertCommit, err := model.GetRevertCommit(strfmt.UUID(originalCommitID), *latestCommit)
+	if err != nil {
+		return locale.WrapError(err, "err_revert_refresh")
+	}
+
+	err = runbits.RefreshRuntime(r.auth, r.out, r.project, r.config.CachePath(), *latestCommit, len(revertCommit.Changeset) > 0)
```
Naatan, line 79:
```suggestion
	revertCommit, err := model.GetRevertCommit(strfmt.UUID(originalCommitID), *latestCommit)
	if err != nil {
		return locale.WrapError(err, "err_revert_refresh")
	}

	err = runbits.RefreshRuntime(r.auth, r.out, r.project, r.config.CachePath(), *latestCommit, true)
```

We already know by the action that someone is taking that we'll want to pull in changes.

Naatan (approved):

PR #1394, opened 2021-05-27
Clean-up before updating from old State Tool

mdrohmann:
https://www.pivotaltracker.com/story/show/178254157

I finally realized that all the cleanup can be done in the transitional state tool, as it is called if and only if we are updating from a single-file State Tool to our new State Tool distribution.

### DIFF
--- a/assets/scripts/removePaths.bat
+++ b/assets/scripts/removePaths.bat
@@ -35,7 +35,7 @@ for /d %%i in (%paths%) do (
     if exist %%i\* (
         rmdir /s /q %%i >> %logfile%
     ) else (
-        del /s /q %%i >> %logfile%
+        del /q %%i >> %logfile%
     )
     if %ERRORLEVEL% NEQ 0 (
         echo "Could not remove directory: %%i" >> %logfile%
--- a/cmd/state-transition-update/main.go
+++ b/cmd/state-transition-update/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/ActiveState/cli/internal/appinfo"
@@ -11,6 +12,7 @@ import (
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/events"
+	"github.com/ActiveState/cli/internal/exeutils"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/runbits"
@@ -64,6 +66,10 @@ func removeOldStateToolEnvironmentSettings(cfg *config.Instance) error {
 }
 
 func run() error {
+	if len(os.Args) == 2 && os.Args[1] == "__is_transitional" {
+		fmt.Println("true")
+		return nil
+	}
 	// handle state export config --filter=dir (install scripts call this function to write the install-source file)
 	if len(os.Args) == 4 && os.Args[1] == "export" && os.Args[2] == "config" && os.Args[3] == "--filter=dir" {
 		cfg, err := config.Get()
@@ -91,10 +97,20 @@ func run() error {
 		return errs.Wrap(err, "failed to remove environment settings from old State Tool installation")
 	}
 
-	installTargetPath := filepath.Dir(appinfo.StateApp().Exec())
-	err = up.InstallBlocking(installTargetPath)
+	err = up.InstallBlocking("")
 	if err != nil {
 		return errs.Wrap(err, "Failed to install multi-file update.")
+	}
+
+	// if the transitional state tool has been replaced by the installer, we are done
+	stdout, _, err := exeutils.ExecSimple(appinfo.StateApp().Exec(), "__is_transitional")
+	if err == nil && strings.TrimSpace(stdout) == "true" {
+		return nil
+	}
+
+	// otherwise: remove the transitional State Tool
+	if err := removeSelf(); err != nil {
+		logging.Error("Failed to remove transitional State Tool: %s", errs.JoinMessage(err))
 	}
 
 	return nil
--- a/cmd/state-transition-update/main.go
+++ b/cmd/state-transition-update/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ActiveState/cli/internal/events"
 	"github.com/ActiveState/cli/internal/exeutils"
 	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/machineid"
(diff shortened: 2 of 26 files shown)


mdrohmann:
This does not set the installation path as requested by the ACs.  See https://www.pivotaltracker.com/story/show/178254157/comments/224523764

test/integration/install_scripts_int_test.go
```diff
@@ -198,22 +202,28 @@ func (suite *InstallScriptsIntegrationTestSuite) TestLegacyInstallShInstallMulti
 	}
 	suite.OnlyRunForTags(tagsuite.InstallScripts, tagsuite.Critical)
 
-	ts := e2e.New(suite.T(), false)
+	ts := e2e.New(suite.T(), true)
```
MDrakos, line 205: Was this for debugging?
```suggestion
	ts := e2e.New(suite.T(), false)
```

test/integration/install_scripts_int_test.go
```diff
@@ -198,22 +202,28 @@ func (suite *InstallScriptsIntegrationTestSuite) TestLegacyInstallShInstallMulti
...
+			fmt.Sprintf("_TEST_INSTALL_PATH=%s", filepath.Join(ts.Dirs.Work, "multi-file")),
+		))
 
 	expectLegacyStateToolInstallation(cp, "n")
 	cp.Expect("State Tool Installed")
 	cp.ExpectExitCode(0)
 
-	suite.FileExists(filepath.Join(ts.Dirs.Work, "state-svc"))
-	suite.FileExists(filepath.Join(ts.Dirs.Work, "state-tray"))
+	// Note: When updating from an old update, we always installing to the default installation path.
```
MDrakos, line 222:
```suggestion
	// Note: When updating from an old update, we always install to the default installation path.
```

test/integration/install_scripts_int_test.go
```diff
@@ -537,8 +555,12 @@ func (suite *InstallScriptsIntegrationTestSuite) TestLegacyInstallPs1MultiFileUp
 	paths := strings.Split(pathEnv, string(os.PathListSeparator))
 	suite.Assert().Contains(paths, ts.Dirs.Work, "Could not find installation path, output: %s", cp.TrimmedSnapshot())
 
-	suite.FileExists(filepath.Join(ts.Dirs.Work, "state-svc.exe"))
-	suite.FileExists(filepath.Join(ts.Dirs.Work, "state-tray.exe"))
+	// Note: When updating from an old update, we always installing to the default installation path.
```
MDrakos, line 558:
```suggestion
	// Note: When updating from an old update, we always install to the default installation path.
```

assets/scripts/removePaths.bat
```diff
@@ -35,7 +35,7 @@ for /d %%i in (%paths%) do (
     if exist %%i\* (
         rmdir /s /q %%i >> %logfile%
     ) else (
-        del /s /q %%i >> %logfile%
```
MDrakos, line 38: 🙈

MDrakos (dismissed):

cmd/state-transition-update/main.go
```diff
@@ -59,9 +91,35 @@ func run() error {
...
+		return errs.Wrap(err, "Failed to install multi-file update.")
+	}
+
+	logging.Debug("Multi-file State Tool is installed.")
+
+	// if the transitional state tool has been replaced by the installer, we are done
+	stdout, _, err := exeutils.ExecSimple(appinfo.StateApp().Exec(), "__is_transitional")
+	if err != nil || strings.TrimSpace(stdout) != "true" {
+		return nil
+	}
```
Naatan, line 117: I'm not understanding what's happening here. Is the intend to detect that the executable has been replaced? Wouldn't the new state tool be installed elsewhere, and so this would only ever return the old?

Additionally, if you're just trying to check if a file has changed I think looking at the timestamp of the file would be a lot cleaner.

internal/installation/paths.go
```diff
@@ -15,6 +15,9 @@ func InstallPath() (string, error) {
 	if !rtutils.BuiltViaCI && strings.Contains(appinfo.StateApp().Exec(), "/build/") {
 		return filepath.Dir(appinfo.StateApp().Exec()), nil
 	}
+	if path, ok := os.LookupEnv("_TEST_INSTALL_PATH"); ok {
+		return path, nil
+	}
```
Naatan, line 20: Please use constants

internal/subshell/cmd/cmd.go
```diff
@@ -90,6 +90,25 @@ func (v *SubShell) WriteUserEnv(cfg sscommon.Configurable, env map[string]string
...
+
+	// Clean up old entries
+	oldEnv := cfg.GetStringMap(envType.Key)
+	for k, v := range oldEnv {
+		if err := cmdEnv.unset(k, v.(string)); err != nil {
+			return err
+		}
+	}
+
+	osutils.PropagateEnv()
```
Naatan, line 104: This doesn't need an error check?

Naatan (changes requested):

cmd/state-transition-update/main.go
mdrohmann: Yes, if the State Tool was previously installed at the default installation path, this transitional tool is going to be replaced by the installer, and we do NOT want to remove it.

Checking the timestamp is an option, of course.  I will do that instead.

Naatan: But the default installation path is all new, there's no situation in which the old state tool would have been installed there unless someone manually set that path, which seems very unlikely.

internal/subshell/cmd/cmd.go
mdrohmann: It never had one, but I added it.  Because the Windows system function supports error handling.

assets/scripts/removePaths.bat
mdrohmann: :-)

internal/constants/constants.go
```diff
@@ -304,3 +304,9 @@ const ToplevelInstallArchiveDir = "state-install"
 
 // FirstMultiFileStateToolVersion is the State Tool version that introduced multi-file updates
 const FirstMultiFileStateToolVersion = "0.29.0"
+
+// OverwriteDefaultInstallationPath is the environment variable name to overwrite the default installation path FOR TESTING PURPOSES ONLY
+const OverwriteDefaultInstallationPathEnvVarName = "_TEST_INSTALL_PATH"
+
+// OverwriteDefaultSystemPathEnvVarName is the environment variable name to overwrite the system app installation directory updates FOR TESTING PURPOSES ONLY
+const OverwriteDefaultSystemPathEnvVarName = "_TEST_SYSTEM_PATH"
```
Naatan, line 312: All of our other env vars use the `ACTIVESTATE_` prefix.

Naatan (changes requested):

Naatan (approved):

PR #1399, opened 2021-06-01
State push creates a project without state init

MDrakos:
https://www.pivotaltracker.com/story/show/178251774

### DIFF
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -194,7 +194,7 @@ func (r *Push) getValidNamespace() (*project.Namespaced, error) {
 		}
 
 		name := lang.String()
-		name, err = r.prompt.Input("", locale.Tl("push_prompt_name", "Project name?"), &name)
+		name, err = r.prompt.Input("", locale.Tl("push_prompt_name", "What would you like the name of this project to be?"), &name)
 		if err != nil {
 			return nil, locale.WrapError(err, "err_push_get_name", "Could not determine project name")
 		}
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -203,7 +203,7 @@ func (r *Push) getValidNamespace() (*project.Namespaced, error) {
 
 	namespace, err := project.ParseNamespace(names)
 	if err != nil {
-		return nil, errs.Wrap(err, "Could not parse namespace %s to push headless commit to", names)
+		return nil, locale.WrapError(err, locale.Tl("err_push_parse_namespace", "Could not parse namespace"))
 	}
 
 	return namespace, nil
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -1,7 +1,7 @@
 package push
 
 import (
-	"path/filepath"
+	"fmt"
 
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/output"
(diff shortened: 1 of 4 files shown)


internal/runners/push/push.go
```diff
@@ -187,23 +179,41 @@ func (r *Push) Run(params PushParams) error {
...
 }
 
-func (r *Push) languageForProject(pj *project.Project) (*language.Supported, string, error) {
-	if pj.CommitID() != "" {
-		lang, err := model.FetchLanguageForCommit(pj.CommitUUID())
+func (r *Push) getNamespace() (*project.Namespaced, error) {
+	namespace := projectfile.GetProjectNameForPath(r.config, r.project.Source().Path())
+	if namespace == "" {
+		owner := authentication.Get().WhoAmI()
+		owner, err := r.prompt.Input("", locale.Tl("push_prompt_owner", "Who will be the owner of this project?"), &owner)
```
Naatan, line 186: Should give some context, eg.

```
Since this is your first time pushing we'll need to set up a project.
```

internal/runners/push/push.go
```diff
@@ -187,23 +179,41 @@ func (r *Push) Run(params PushParams) error {
...
-			return nil, "", errs.Wrap(err, "Failed to retrieve language information for headless commit.")
+			return nil, locale.WrapError(err, "err_push_get_owner", "Could not deterimine project owner")
 		}
 
-		l, err := language.MakeByNameAndVersion(lang.Name, lang.Version)
+		lang, _, err := fetchLanguage(r.project.CommitUUID())
 		if err != nil {
-			return nil, "", errs.Wrap(err, "Failed to convert commit language to supported language.")
+			return nil, locale.WrapError(err, "err_push_fetch_language", "Could not fetch project language")
 		}
```
Naatan, line 194: This shouldn't actually cause a failure if it fails, we should just fall back on an empty name.

Naatan (changes requested):

internal/runners/push/push.go
```diff
@@ -187,23 +180,43 @@ func (r *Push) Run(params PushParams) error {
...
-		ls := language.Supported{Language: l}
-		if !ls.Recognized() {
-			return nil, "", locale.NewError("err_push_invalid_language", lang.Name)
+
+		name, err = r.prompt.Input("", locale.Tl("push_prompt_name", "What would you like the name of this project to be?"), &name)
+		if err != nil {
+			return nil, locale.WrapError(err, "err_push_get_name", "Could not determine project name")
 		}
-		return &ls, lang.Version, nil
+		namespace = fmt.Sprintf("%s/%s", owner, name)
```
Naatan, line 204: use `project.NewNamespace().String()`

Naatan (changes requested):

internal/runners/push/push.go
```diff
@@ -91,7 +86,7 @@ func (r *Push) Run(params PushParams) error {
 	// Get the project remotely if it already exists
```
Naatan, line 86: As discussed on Slack I think we can just drop this altogether now.

Naatan (changes requested):

internal/runners/push/push.go
MDrakos: Looking at this more closely we need to keep the fetching of the remote project as we use that to determine if we're pushing to an existing project or creating a new project with `state init`. However, there is an error case that is no longer relevant and has been removed.

Naatan (approved):

PR #1400, opened 2021-06-01
Ensure tray is stopped during uninstall

daved:
https://www.pivotaltracker.com/story/show/178320487


daved:
The service is still not being halted during uninstall; A related story has been filed: https://www.pivotaltracker.com/story/show/178371438

Naatan (approved):

PR #1402, opened 2021-06-02
Add bash path function to script expander

mdrohmann:
https://www.pivotaltracker.com/story/show/178201884

### DIFF
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -138,17 +138,17 @@ scripts:
     value: |
       set -e
       echo "Building State Tool"
-      $scripts.build.path.posix()
+      $scripts.build.path._posix()
       echo "Building State Installer"
-      $scripts.build-installer.path.posix()
+      $scripts.build-installer.path._posix()
       echo "Building State Service"
-      $scripts.build-svc.path.posix()
+      $scripts.build-svc.path._posix()
       echo "Building State Tray App"
-      $scripts.build-tray.path.posix()
+      $scripts.build-tray.path._posix()
       echo "Building State Update Dialog Frontend"
-      $scripts.build-updlg-front.path.posix()
+      $scripts.build-updlg-front.path._posix()
       echo "Building State Update Dialog"
-      $scripts.build-updlg.path.posix()
+      $scripts.build-updlg.path._posix()
   - name: build-installer
     language: bash
     description: Builds the state-installer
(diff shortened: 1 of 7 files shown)


activestate.yaml
```diff
@@ -138,17 +138,17 @@ scripts:
     value: |
       set -e
       echo "Building State Tool"
-      $scripts.build.path()
+      $scripts.build.bash-path()
```
Naatan, line 141: Hmm what if we did `$scripts.build.path(posix)`, or `$scripts.build.path.posix()`.

I don't think we should use `bash`, that's too opinionated.

pkg/project/expander.go
```diff
@@ -134,8 +137,27 @@ func ScriptExpander(_ string, name string, meta string, isFunction bool, project
...
+				return "", err
+			}
+			return filepath.ToSlash(path), nil
+		case "bash-path":
+			path, err := expandPath(name, script)
+			if err != nil {
+				return "", err
+			}
+			if runtime.GOOS == "windows" {
+				return fmt.Sprintf("$((type cygpath > /dev/null && cygpath '%s') || (type wslpath > /dev/null && wslpath -u '%s') || echo '%s')", path, path, path), nil
```
Naatan, line 156: We need to run this ourselves in the background and pass back the result. `path()` returns a string that is the path of the file, not a string that needs to be evaled.

Naatan (approved):

Naatan (changes requested):
Woops, meant to request changes.

activestate.yaml
daved: This, in essence, is creating a portion of the API that we'll need to continually support in order to inject some context into the expander. Further, this is an easy-to-miss dependency. For example, if a bunch of Linux users don't use path.posix() and then a Windows user joins the team and unexpectedly gets the wrong value. It's not intuitive to need to repeat bash/posix while inside of a bash script.

I very much think that we, in the go code, should change the working directory to the one holding the script, then get the path to the script using the shell determined by the script language. This solution would require us to be able to send in contextual info (the shell type) to the expander logic.

Naatan: As I said yesterday this can be an iteration, I agree the ideal we're striving for is that `path()` just handles this intelligently, but there are always edge cases and having explicit behavior "as well" does not hurt.

daved: A "POSIX" version of a Windows path is not what is needed. The resolved path is environment-dependent. For example, `C:\Users\username` is `/mnt/c/Users/username` in WSL Bash on Windows. This would be `/c/Users/username` if "bash" is actually git-bash. The context that is needed in the expander is which shell is being used so that that shell's environment can be called to get the path. A simple "conversion to posix" is not sufficient.

activestate.yaml
```diff
@@ -138,17 +138,17 @@ scripts:
     value: |
       set -e
       echo "Building State Tool"
-      $scripts.build.path()
+      $scripts.build.path.posix()
```
Naatan, line 141: Please use the underscore variant we discussed, ie. `_posix()`. The underscore is important as it communicates "danger".

internal/osutils/osutils.go
```diff
@@ -56,10 +57,33 @@ func BashifyPath(absolutePath string) (string, error) {
...
+func winPathToLinPath(name string) (string, error) {
+	cmd := exec.Command("bash", "-c", "pwd")
+	cmd.Dir = filepath.Dir(name)
+
+	out, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+
+	path := strings.TrimSpace(string(out)) + "/" + filepath.Base(name)
```
Naatan, line 84: Wouldn't `out` already contain `filepath.Base(name)` ?

Naatan (changes requested):

internal/osutils/osutils.go
mdrohmann: No, in line 77, we remove the basename, so we can change into that directory.

Naatan (approved):

PR #1403, opened 2021-06-03
State clean cache panic fix

MDrakos:
https://www.pivotaltracker.com/story/show/178307440

### DIFF
--- a/internal/runners/clean/cache.go
+++ b/internal/runners/clean/cache.go
@@ -44,6 +44,10 @@ func (c *Cache) Run(params *CacheParams) error {
 	if params.Project != "" {
 		paths := projectfile.GetProjectPaths(c.config, params.Project)
 
+		if len(paths) == 0 {
+			return locale.NewInputError("err_cache_no_project", "Could not determine path to project {{.V0}}", params.Project)
+		}
+
 		for _, projectPath := range paths {
 			err := c.removeProjectCache(projectPath, params.Project, params.Force)
 			if err != nil {
@@ -82,13 +86,9 @@ func (c *Cache) removeProjectCache(projectDir, namespace string, force bool) err
 		}
 	}
 
-	projectInstallPath, err := runtime.ProjectDirToTargetDir(projectDir, c.config.CachePath())
-	if err != nil {
-		return locale.WrapError(err, "err_cache_target_dir", "Could not get target cache directory")
-	}
+	projectInstallPath := runtime.ProjectDirToTargetDir(projectDir, c.config.CachePath())
 	logging.Debug("Remove project path: %s", projectInstallPath)
-
-	err = os.RemoveAll(projectInstallPath)
+	err := os.RemoveAll(projectInstallPath)
 	if err != nil {
 		return locale.WrapError(err, "err_clean_remove_artifact", "Could not remove cached runtime environment for project: {{.V0}}", namespace)
 	}
--- a/internal/runners/clean/cache.go
+++ b/internal/runners/clean/cache.go
@@ -3,22 +3,19 @@ package clean
 import (
 	"os"
 
-	"github.com/ActiveState/cli/internal/config"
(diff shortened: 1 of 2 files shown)


pkg/platform/runtime/target.go
```diff
@@ -73,14 +80,12 @@ func (c *CustomTarget) OnlyUseCache() bool {
...
 
-func ProjectDirToTargetDir(projectDir, cacheDir string) string {
-	projectDirDirty := filepath.Dir(projectDir)
-	projectDir, err := fileutils.ResolveUniquePath(projectDirDirty)
+func ProjectDirToTargetDir(projectDir, cacheDir string) (string, error) {
+	resolvedDir, err := fileutils.ResolveUniquePath(projectDir)
 	if err != nil {
-		logging.Error("Could not resolve unique path for projectDir: %s, error: %s", projectDir, err.Error())
-		projectDir = projectDirDirty
+		return "", locale.WrapError(err, "err_target_resolve_dir", "Could not resolve project directory")
```
mdrohmann, line 86: hmm... unless, the `projectDir` is the empty string, it should be safe to proceed with that string, shouldn't it?

The function doesn't seem to [ever have returned an error](https://rollbar.com/activestate/all/items/?sort=%5Bobject%20Object%5D&status=active&enc_query=PaPqSdMv2PrGOy8peB3IzqJj5jOl8MdnmGGo0SqqdLBbsoFs7oZ8NzfUGZVDYZOSyhy%2F%2BdYFFLQhWPcq%2Bml8PA%3D%3D&date_from=&date_to=&environments=beta&environments=master&environments=release&environments=unstable&environments=unstable-install.ps1&environments=unstable-install.sh&environments=unstable-msi-silent&environments=unstable-msi-ui&environments=unstable-unknown&environments=unstable-%EF%BF%BD%EF%BF%BDi%00n%00s%00t%00a%00l%00l%00.%00p%00s%001%00%20%00%20%00&activated_to=&timezone=America%2FLos_Angeles&framework=&levels=10&levels=20&levels=30&levels=40&levels=50&activated_from=&offset=0&query=Could%20not%20resolve%20unique%20path%20for%20projectDir&assigned_user=&date_filtering=seen&projects=291966).  So, it shouldn't be necessary to change that behavior IMO...

mdrohmann (changes requested):

pkg/platform/runtime/target.go
MDrakos: Fair point. I've reverted the returning of an error but this function has still changed slightly

internal/runners/clean/cache.go
```diff
@@ -53,6 +50,7 @@ func (c *Cache) Run(params *CacheParams) error {
 				return err
 			}
 		}
+		return nil
```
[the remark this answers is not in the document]
MDrakos: This is a different issue from the original error. Without this return, after removing a project cache it will prompt the user to remove the entire cache.

mdrohmann (approved):

Naatan (changes requested):
These changes look sane but they don't appear to actually fix the panic, just work around it.

The issue being on this line:

http://github.com/ActiveState/cli/blob/0697c53dc3540e5776de3364faca0d7e7cb6bb33/internal/config/instance.go#L317-L317

It seems `i.configDir` is nil there, which looking at the code is a state we assume should never happen. I think we need to update the init code to ensure it errors out if it cannot set the configDir.

MDrakos:
I can certainly make that change however I do not think it was what caused the panic. The `i.configDir` was panicking because the actual instance was `nil`. In the `Cache` struct we had two members for the config, `config` which is an interface, and `cfg` which is a pointer to `config.Instance`. The latter was not set in the constructor.

Naatan:
@MDrakos thanks for clarifying, then if the root issue is addressed I'm thumbs up 👍 :)

Naatan (approved):

PR #1404, opened 2021-06-03
Update systray URLs

MDrakos:
https://www.pivotaltracker.com/story/show/178353848

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -164,6 +164,9 @@ const DateTimeFormatRecord = "Mon Jan 2 2006 15:04:05 -0700 MST"
 // PlatformSignupURL is the account creation url used by the platform
 const PlatformSignupURL = "https://platform.activestate.com" + "/create-account"
 
+// TrayDocumentationURL is the url for the state tool documentation to be used in the state tray application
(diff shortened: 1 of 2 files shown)


internal/constants/constants.go
```diff
@@ -164,6 +164,9 @@ const DateTimeFormatRecord = "Mon Jan 2 2006 15:04:05 -0700 MST"
 // PlatformSignupURL is the account creation url used by the platform
 const PlatformSignupURL = "https://platform.activestate.com" + "/create-account"
 
+// TrayDocumentationURL is the url for the state tool documentation to be used in the state tray application
+const TrayDocumentationURL = "http://docs.activestate.com/platform/state/?utm_source=platform-application-gui&utm_medium=activestate-desktop&utm_content=drop-down&utm_campaign=maru"
```
Naatan, line 168: This doesn't look right, it just points to the general state tool docs.

Naatan (changes requested):

internal/constants/constants.go
MDrakos: What should it be pointing to?

Naatan: Oh I see, this just points to the general documentation, but FROM the systray. That confused me, I thought this was the documentation FOR the systray.

PR #1407, opened 2021-06-07
Fix update dialog on Windows

mdrohmann:
https://www.pivotaltracker.com/story/show/178320910

I am still not clear why the CI seemingly links the State Tool Updater to an IE based JS interpreter, whereas on my machine it is Edge based.  But this change fixes the issue, as it makes it compatible with the IE engine.


Naatan (approved):
Nice digging!

PR #1396, opened 2021-05-28
State install does not update platform project

MDrakos:
~~Merging PR is blocked until we can update the error messaging around camel builds in progress.~~

https://www.pivotaltracker.com/story/show/178222594

PR #1408, opened 2021-06-07
Remove headless mode from installing while unauthenticated

Naatan:
https://www.pivotaltracker.com/story/show/178418483

Test failures seem to be coming from https://github.com/ActiveState/cli/pull/1396

### DIFF
--- a/test/integration/push_int_test.go
+++ b/test/integration/push_int_test.go
@@ -138,7 +138,7 @@ func (suite *PushIntegrationTestSuite) TestCarlisle() {
 	cp = ts.SpawnWithOpts(e2e.WithArgs(
 		"install", suite.extraPackage),
 		e2e.WithWorkDirectory(wd),
-		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"))
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false", "VERBOSE=true"))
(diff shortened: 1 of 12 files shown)


test/integration/push_int_test.go
```diff
@@ -161,12 +150,12 @@ func (suite *PushIntegrationTestSuite) TestCarlisle() {
...
 	prj, err := project.FromPath(filepath.Join(wd, constants.ConfigFileName))
 	suite.Require().NoError(err, "Could not parse project file")
-	suite.Assert().True(prj.IsHeadless(), "project should be headless: URL is %s", prj.URL())
+	suite.Assert().False(prj.IsHeadless(), "project should NOT be headless: URL is %s", prj.URL())
 
 	ts.LoginAsPersistentUser()
 
 	// convert to real project
-	cp = ts.SpawnWithOpts(e2e.WithArgs("init", namespace), e2e.WithWorkDirectory(wd))
+	cp = ts.SpawnWithOpts(e2e.WithArgs("init", namespace, "python3"), e2e.WithWorkDirectory(wd))
```
MDrakos, line 158: I think this _may_ cause a test failure. As the base project that is activated in this test can either by Perl-5.32 or Python. It is only a Python project on MacOS

MDrakos (changes requested):
Can you just check that the change to `TestCarlise` isn't causing a failure?

PR #1410, opened 2021-06-08
Config.lock is removed

mdrohmann:
- lock the configuration file directly (no need for a separate file)
- ensure that state-tray does not create defunct process

 https://www.pivotaltracker.com/story/show/178440954


internal/config/instance.go
```diff
@@ -437,19 +437,23 @@ func (i *Instance) ensureCacheExists() error {
 }
 
 func (i *Instance) getLockFile() string {
+	// On Linux and Darwin, we can use the config file itself for locking
+	if runtime.GOOS != "windows" {
+		return i.getConfigFile()
+	}
```
Naatan, line 443: Is this necessary though? I'd prefer we use the same mechanic everywhere so we get consistent behavior.

internal/exeutils/exeutils.go
```diff
@@ -157,5 +157,12 @@ func ExecuteAndForget(command string, args []string, opts ...func(cmd *exec.Cmd)
 		return nil, errs.Wrap(err, "Could not start %s %v", command, args)
 	}
 	cmd.Stdin = nil
+
+	// Wait for the command to finish in a go-routine.  If we do not do that, and the parent process keeps running,
+	// the launched process will keep around flagged <defunct> (at least on Linux)
```
Naatan, line 162: That's silly.. nice catch.

Naatan (changes requested):

internal/config/instance.go
mdrohmann: We could use the same mechanic, we use for Windows on all OSes.  IMO the Linux/Darwin way is the correct one though, and having to create a separate file on Windows, is just a hack...  But unfortunately, on Windows we would need to re-use the file handle we use for locking for our write operations, and the `gofrs/flock` package implementation does not support that functionality.

So, I see three options:
1. Use the separate lockfile on all OSes
2. Use the separate lockfile only on Windows (as in this PR)
3. modify the flock package such that the locked file handle and use that one to read and write the configuration file.

Naatan: I would go for option number 1. There's no technical issues with using a separate lockfile on linux/darwin right? Just that it creates a lockfile when technically it doesn't need a separate one.

mdrohmann: I think so.  My only concern was adding file-removal step to the equation here, as this could happen between a different process opening a file handle for the lock file and locking it:  Windows does not allow the removal of files after the file is opened, Linux/Darwin does, but also seems to be forgiving about locking a file that does not exist anymore...  So, yeah, seems to be fine.

Naatan (approved):

PR #1411, opened 2021-06-08
Default executors do not reference runtime executors

MDrakos:
https://www.pivotaltracker.com/story/show/178417791

In the case of a global default, we get a list of executables that we will create the executors from. Since the runtime has already been installed at this point the `exec` directory contains executable files. This means that we were creating executors from executors, leading to instances where an executor would essentially be self-referencing (ie. `python.bat` would end up invoking the `python.bat` on `PATH`).

This PR updates the `exeutils.Executables` function to take optional directories to skip. In the case of executors we want to skip the `exec` directory so we are only creating executors from the original executables.

### DIFF
--- a/internal/exeutils/exeutils.go
+++ b/internal/exeutils/exeutils.go
@@ -23,13 +23,17 @@ func Executables(bins []string, skip ...string) ([]string, error) {
 	for _, bin := range bins {
 		err := filepath.Walk(bin, func(fpath string, info os.FileInfo, err error) error {
 			// Filter out files that are not executable
-			if info == nil || info.IsDir() || !fileutils.IsExecutable(fpath) { // check if executable by anyone
+			if info == nil || !fileutils.IsExecutable(fpath) { // check if executable by anyone
+				return nil // not executable
+			}
+
+			if info.IsDir() {
 				for _, s := range skip {
 					if info.Name() == s {
 						return filepath.SkipDir
 					}
 				}
-				return nil // not executable
+				return nil
 			}
 
 			exes = append(exes, fpath)
--- a/internal/exeutils/exeutils.go
+++ b/internal/exeutils/exeutils.go
@@ -22,17 +22,23 @@ func Executables(bins []string) ([]string, error) {
 	exes := []string{}
 
 	for _, bin := range bins {
-		entries, err := ioutil.ReadDir(bin)
-		if err != nil {
-			return nil, errs.Wrap(err, "Could not read directory")
+		if fileutils.IsDir(bin) {
+			entries, err := ioutil.ReadDir(bin)
+			if err != nil {
+				return nil, errs.Wrap(err, "Could not read directory")
+			}
+
+			for _, entry := range entries {
+				fpath := filepath.Join(bin, entry.Name())
+				if fileutils.IsExecutable(fpath) {
+					exes = append(exes, fpath)
+				}
+			}
+			continue
 		}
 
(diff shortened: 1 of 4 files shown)


Naatan (changes requested):
I'd rather we treat the cause of the problem rather than the symptom. The problem is that it's detecting executables using `exec` in the PATH right? So we should in the relevant spot split `exec` from the PATH.

I think you can narrow the problem down to this spot:

http://github.com/ActiveState/cli/blob/0dbebe3514e8d135428d7d6af46bbb23b3a61081/pkg/platform/runtime/executor/executor.go#L56-L56

As in that's the most central common denominator. If you find where that function is used you'll find the spots where it's passing in an `envdef.ExecutablePaths` with executors on the PATH.

It looks like the source of the executablePaths is here: http://github.com/ActiveState/cli/blob/0dbebe3514e8d135428d7d6af46bbb23b3a61081/pkg/platform/runtime/envdef/environment.go#L364-L364

But that's suggesting that it's not inheriting any env info, so there might be a bug there.

Key point; let's address the root cause.

MDrakos:
The `exec` directory is not in `PATH` when `ExecutablePaths` is called; however, the runtime cache directory and some of its sub directories are. Here is an example of the directories that are on `PATH` when `ExecutablePaths` is called:

```
C:\Users\msft-admin\AppData\Local\activestate\3ad3401c 
C:\Users\msft-admin\AppData\Local\activestate\3ad3401c\usr\bin 
C:\Users\msft-admin\AppData\Local\activestate\3ad3401c\bin
```


The issue is that `exeutlils.Executables` walks the file paths it's given, essentially putting anything in the runtime directory and subdirectories on PATH. That is why this PR provides the option to skip a directory if it is encountered.

It might be more appropriate for `exeutils.Executables` to simply list the files in the directories it's given and find executables that way, rather than walking the paths. But this function is also used by deploy and I was not sure how that change might affect it.

Naatan:
> essentially putting anything in the runtime directory **and subdirectories** on PATH.
> ..  **and subdirectories**

If this is true then that's the bug. It should not be looking for executables in subdirs, as that's not what PATH does either.

internal/exeutils/exeutils.go
```diff
@@ -21,17 +22,23 @@ func Executables(bins []string) ([]string, error) {
...
 			}
+			continue
+		}
 
-			exes = append(exes, fpath)
-			return nil
-		})
-		if err != nil {
-			return exes, errs.Wrap(err, "Error while walking path")
+		if fileutils.IsExecutable(bin) {
```
Naatan, line 40: None of the bins themselves can be executables, the original code only looks for executables inside the bin paths.

internal/exeutils/exeutils.go
```diff
@@ -21,17 +22,23 @@ func Executables(bins []string) ([]string, error) {
 	exes := []string{}
 
 	for _, bin := range bins {
-		err := filepath.Walk(bin, func(fpath string, info os.FileInfo, err error) error {
-			// Filter out files that are not executable
-			if info == nil || info.IsDir() || !fileutils.IsExecutable(fpath) { // check if executable by anyone
-				return nil // not executable
+		if fileutils.IsDir(bin) {
```
Naatan, line 25: This should always be the case, if anything if this isn't a dir we should return an error.

Naatan (changes requested):

Naatan (approved):

PR #1412, opened 2021-06-09
Check error in logout runner

MDrakos:
While this doesn't address the issue from the ticket directly (as I was not able to reproduce it) this does catch an error that we have been ignoring for a while and could help debug if we run into this bug again.

 https://www.pivotaltracker.com/story/show/178440926


Naatan (approved):
Good catch. This does not resolve the story you're linking to though.

PR #1413, opened 2021-06-10
Invalidate executors on update

mdrohmann:
https://www.pivotaltracker.com/story/show/178453839


pkg/platform/runtime/setup/implementations/camel/runtime.go
```diff
@@ -18,12 +18,20 @@ func NewSetup(s *store.Store) *Setup {
...
+	}
 	return make(store.StoredArtifactMap)
 }
 
-func (s *Setup) DeleteOutdatedArtifacts(_ artifact.ArtifactChangeset, _, _ store.StoredArtifactMap) error {
+// DeleteOutdatedArtifacts deletes the entire installation directory, unless alreadyInstalled is not zero, which can happen when the executors directory needs to be re-generated.
+func (s *Setup) DeleteOutdatedArtifacts(_ artifact.ArtifactChangeset, _, alreadyInstalled store.StoredArtifactMap) error {
+	if len(alreadyInstalled) != 0 {
+		return nil
+	}
```
daved, line 34: This feels disjointed. Not a change for this PR, but it gives me a strong feeling that something is not modeled here. If my concern brings something to mind that isn't already in plans, please file a story.

internal/runners/prepare/prepare.go
```diff
@@ -39,6 +44,37 @@ func New(prime primeable) *Prepare {
...
 
+// resetExecutors removes the executor directories for all project installations, and rewrites the global default executors
+// This ensures that the installation is compatible with an updated State Tool installation
+func (r *Prepare) resetExecutors() error {
+	projects := projectfile.GetProjectMapping(r.cfg)
+	for _, projectDirs := range projects {
+		for _, projectDir := range projectDirs {
+			installDir := rt.ProjectDirToTargetDir(projectDir, r.cfg.CachePath())
+			logging.Debug("Reset executor for %s", projectDir)
+			_ = os.RemoveAll(setup.ExecDir(installDir))
```
daved, line 55: nit: It might be worth adding this err return to the logging.Error output.

daved (dismissed):

pkg/platform/runtime/setup/implementations/camel/runtime.go
mdrohmann: I agree: It is disjointed.  I think it is because this is (finally) designed for the "alt-builder"-use case.  In the "alt-case", the `alreadyInstalled` holds the *artifacts* that are already installed.  For the "camel"-use case, there are only two possibilities: 
1. None of the requested artifacts can be re-used
2. All of the requested artifacts can be re-used

I don't have an easy solution for it right now. Maybe we should file a ticket for it, discussing whether we have to clean up the "camel"-path?

Naatan (changes requested):
...

mdrohmann:
@Naatan It is ready for review now.

internal/runners/prepare/prepare.go
```diff
@@ -53,14 +77,13 @@ func (r *Prepare) Run(cmd *captain.Command) error {
 
 	if err := prepareCompletions(cmd, r.subshell); err != nil {
 		if !errs.Matches(err, &ErrorNotSupported{}) {
-			r.reportError(locale.Tr("err_prepare_completions", "Could not generate completions script, error received: {{.V0}}.", err.Error()), err)
+			r.reportError(locale.Tl("err_prepare_completions", "Could not generate completions script, error received: {{.V0}}.", err.Error()), err)
 		}
 	}
 
-	if err := prepareCompletions(cmd, r.subshell); err != nil {
```
Naatan, line 60: -.- good catch

pkg/platform/runtime/store/store.go
```diff
@@ -70,22 +72,41 @@ func (s *Store) HasMarker() bool {
...
+// MarkerIsValid checks if stored runtime is complete and can be loaded
+func (s *Store) MarkerIsValid(commitID strfmt.UUID) bool {
 	marker := s.markerFile()
 	if !fileutils.FileExists(marker) {
 		logging.Debug("Marker does not exist: %s", marker)
 		return false
 	}
 
-	contents, err := fileutils.ReadFile(marker)
+	f, err := os.OpenFile(marker, os.O_RDONLY, 0)
```
Naatan, line 83: I don't think we should be relying on the lower level implementations just so we can go line by line. Going line by line with just a string/bytearray as input is pretty simple, don't need to rely on IO libraries for that.

pkg/platform/runtime/store/store.go
```diff
@@ -70,22 +72,41 @@ func (s *Store) HasMarker() bool {
...
-	return strings.TrimSpace(string(contents)) == commitID.String()
+	sc := bufio.NewScanner(f)
+	ok := sc.Scan()
+	if !ok {
+		logging.Error("Expected commit ID in marker file: %v", sc.Err())
+		return false
+	}
+	parsedID := strings.TrimSpace(sc.Text())
+	logging.Debug("Matching commitID for %s, %s==%s", marker, parsedID, commitID.String())
+	if parsedID != commitID.String() {
```
Naatan, line 98: Might be good to debug log the expected vs actual here (same for state tool version further down).

pkg/platform/runtime/store/store.go
```diff
@@ -70,22 +72,41 @@ func (s *Store) HasMarker() bool {
...
+		return false
+	}
+
+	ok = sc.Scan()
+	if !ok {
+		logging.Debug("Expected State Tool version in marker file: %v", sc.Err())
+		return false
+	}
+	parsedStateToolVersion := strings.TrimSpace(sc.Text())
+	logging.Debug("Matching State Tool Version for %s, %s==%s@%s", marker, parsedStateToolVersion, constants.BranchName, constants.Version)
```
Naatan, line 108: Feels odd that we're debug logging what we expect and not what is actual.

Naatan (changes requested):

pkg/platform/runtime/store/store.go
mdrohmann: parsedStateToolVersion is the actual value.  But maybe I should just log this when the check fails with an "expect: %s, got: %s" message.

pkg/platform/runtime/store/store.go
```diff
@@ -80,33 +79,33 @@ func (s *Store) MarkerIsValid(commitID strfmt.UUID) bool {
...
+
+	if parsedCommitID != commitID.String() {
+		logging.Debug("Could not match commitID in %s, expected: %s, got: %s", marker, commitID.String(), parsedCommitID)
 		return false
 	}
 
-	ok = sc.Scan()
-	if !ok {
-		logging.Debug("Expected State Tool version in marker file: %v", sc.Err())
+	if parsedStateToolVersion != fmt.Sprintf("%s@%s", constants.BranchName, constants.Version) {
```
Naatan, line 103: Didn't notice this the first review; we should not include the branch name. Just the State Tool version. If we do our job right the version should be the same regardless of which branch it's on.

Naatan (changes requested):

pkg/platform/runtime/store/store.go
mdrohmann: the sha should always differ, that is true.

Naatan (approved):

PR #1415, opened 2021-06-14
Ensure project tracking in config works

daved:
https://www.pivotaltracker.com/story/show/178436609

### DIFF
--- a/internal/config/instance.go
+++ b/internal/config/instance.go
@@ -101,6 +101,9 @@ func (i *Instance) ReleaseLock() error {
 	if err := i.lock.Unlock(); err != nil {
 		return errs.Wrap(err, "Failed to release lock")
 	}
+
+	// Ignore the error, as there are legitimate cases where it will fail (when another processes has locked the file again)
+	_ = os.Remove(i.getLockFile())
 	return nil
 }
 
@@ -330,10 +333,7 @@ func (i *Instance) InstallSource() string {
 }
 
 func (i *Instance) ReadInConfig() error {
-	configFile, err := i.getConfigFile()
-	if err != nil {
-		return errs.Wrap(err, "Could not find config file")
-	}
+	configFile := i.getConfigFile()
 
 	configData, err := ioutil.ReadFile(configFile)
 	if err != nil {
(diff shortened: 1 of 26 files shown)


internal/runners/activate/activate.go
```diff
@@ -219,6 +219,9 @@ func (r *Activate) run(params *ActivateParams) error {
 		}
 	}
 
+	// update config with project
```
mdrohmann, line 222: Did you see that we are updating the project mapping whenever we are parsing the project file?
https://github.com/ActiveState/cli/blob/master/pkg/projectfile/projectfile.go#L494

Shouldn't that be enough, or do we still need this line?

mdrohmann (changes requested):

internal/runners/activate/activate.go
daved: No, I did not catch that. The depth at which this side-effect is buried is somewhat unnerving. I'll remove the explicit update and test locally to be sure it still has the desired behavior.

mdrohmann (dismissed):

internal/config/instance.go
```diff
@@ -177,11 +167,11 @@ func Get() (*Instance, error) {
 	return defaultConfig, nil
 }
 
-// SetWithLock updates a value at the given key. The valueF argument returns the
+// Update updates a value at the given key. The valueF argument returns the
 // new value based on the previous one.  If the function returns with an error, the
 // update is cancelled.  The function ensures that no-other process or thread can modify
 // the key between reading of the old value and setting the new value.
-func (i *Instance) SetWithLock(key string, valueF func(oldvalue interface{}) (interface{}, error)) error {
+func (i *Instance) Update(key string, valueF func(oldvalue interface{}) (interface{}, error)) error {
```
Naatan, line 174: I know I'm not flagged for review, but I want to ensure this isn't merged as is. 

I see two issues here.

  1. I expressed my issues with this function even existing. Why then rename it to make it even more prone to adoption?

  2. This struct already has a `Set()` method. Introducing a vague `Update()` analog is just confusing and inconsistent.

Primarily the first though. The fact that this was renamed makes me concerned that my concerns about this function were not heard or understood. To be clear: this function is (and SHOULD be) an eyesore. I want to get rid of it as soon as reasonably possible. That means minimizing its effect on our codebase and exploring better alternatives.

Naatan (changes requested):

mdrohmann (approved):

Naatan (approved):

PR #1417, opened 2021-06-14
Auth logout fix

MDrakos:
Through the imports of `state-svc` the `init` function from the `analytics` package is being called. This leads to an `authentication.Get` call however the configuration at that point is not up to date. For a more detailed explanation see here: https://www.pivotaltracker.com/story/show/178440926/comments/224797191. 

From tracing the import chain it was the `runbits` package that imports `analytics`. Moving the `HandlePanics` function to its own package fixed the log out bug in my testing.

Filed a follow-up story here: https://www.pivotaltracker.com/story/show/178529229 to remove the `init` function from the `analytics` package.

https://www.pivotaltracker.com/story/show/178440926

### DIFF
--- a/test/integration/auth_int_test.go
+++ b/test/integration/auth_int_test.go
@@ -59,7 +59,7 @@ func (suite *AuthIntegrationTestSuite) interactiveLogin(ts *e2e.Session, usernam
 func (suite *AuthIntegrationTestSuite) loginFlags(ts *e2e.Session, username string) {
 	cp := ts.Spawn("auth", "--username", username, "--password", "bad-password")
 	cp.Expect("Authentication failed")
-	cp.Expect("You are not authorized, did you provide valid login credentials?")
+	cp.ExpectLongString("You are not authorized, did you provide valid login credentials?")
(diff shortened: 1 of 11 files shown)


test/integration/auth_int_test.go
```diff
@@ -58,10 +59,16 @@ func (suite *AuthIntegrationTestSuite) interactiveLogin(ts *e2e.Session, usernam
 func (suite *AuthIntegrationTestSuite) loginFlags(ts *e2e.Session, username string) {
 	cp := ts.Spawn("auth", "--username", username, "--password", "bad-password")
 	cp.Expect("Authentication failed")
-	cp.Expect("You are not authorized, did you provide valid login credentials?")
+	cp.ExpectLongString("You are not authorized, did you provide valid login credentials?")
 	cp.ExpectExitCode(1)
 }
 
+func (suite *AuthIntegrationTestSuite) ensureLogout(ts *e2e.Session) {
```
Naatan, line 66: Nit: This doesn't need to be its own function.

Naatan (approved):

PR #1421, opened 2021-06-17
State install works without a project

MDrakos:
https://www.pivotaltracker.com/story/show/178223554


internal/runners/packages/install.go
```diff
@@ -38,16 +38,21 @@ func NewInstall(prime primeable) *Install {
...
-		return locale.WrapError(err, "err_fetch_languages")
+		language, err = model.LanguageForPackage(params.Package.Name())
+		if err != nil {
+			return locale.WrapError(err, "err_install_get_langauge", "Could not get language for package: {{.V0}}", params.Package.Name())
+		}
+	} else {
+		language, err = model.LanguageForCommit(a.proj.CommitUUID())
+		if err != nil {
+			return locale.WrapError(err, "err_fetch_languages")
+		}
```
Naatan, line 52: This is fundamentally flawed as we should not have any opinion on the language, that's what the solver is for.

Of course that does raise issues for certain State Tool use-cases, but we should be addressing those issues rather than working around them by spoofing the language.

internal/runners/packages/packages.go
```diff
@@ -42,7 +47,11 @@ type configurable interface {
 
 const latestVersion = "latest"
 
-func executePackageOperation(pj *project.Project, cfg configurable, out output.Outputer, authentication *authentication.Auth, prompt prompt.Prompter, name, version string, operation model.Operation, ns model.Namespace) error {
+func executePackageOperation(pj *project.Project, cfg configurable, out output.Outputer, authentication *authentication.Auth, prompt prompt.Prompter, packageName, packageVersion, languageName string, operation model.Operation, ns model.Namespace) error {
+	if pj == nil {
+		return installInitial(cfg, out, authentication, prompt, packageName, packageVersion, languageName, operation, ns)
+	}
```
Naatan, line 53: It doesn't make sense that we would just skip all the normal operations in favor of this special behavior. I think what we need to do at this point is create the project file, and then the rest of the logic can run as it normally would.

internal/runners/packages/packages.go
```diff
@@ -162,3 +171,57 @@ func getSuggestions(ns model.Namespace, name string) ([]string, error) {
 
 	return suggestions, nil
 }
+
+func installInitial(cfg configurable, out output.Outputer, authentication *authentication.Auth, prompt prompt.Prompter, packageName, packageVersion, languageName string, operation model.Operation, ns model.Namespace) error {
```
Naatan, line 175: fwiw not reviewing this function as I don't believe it should exist as per my comment above.

pkg/platform/model/projects.go
```diff
@@ -85,6 +86,35 @@ func LanguageForCommit(commitID strfmt.UUID) (string, error) {
 	return languages[0].Name, nil
 }
 
+// LanguageForPackage attempts to fetch the language belonging to the given package name
+func LanguageForPackage(name string) (string, error) {
```
Naatan, line 90: This shouldn't exist, that's what the solver is for.

Naatan (changes requested):
@MDrakos since you're going on vacation don't worry about addressing these comments, I'll get on this myself since it's unlikely we'd wrap this up today.

MDrakos:
The failing integration test is unrelated to this PR. Follow up story followed here: https://www.pivotaltracker.com/story/show/178788386

Naatan (dismissed):

PR #1424, opened 2021-06-18
Skip rollbar reporting for svc already running error

mdrohmann:
https://www.pivotaltracker.com/story/show/178563462


Naatan (changes requested):
We already have a pattern for these types of errors:

http://github.com/ActiveState/cli/blob/bdac00ecb11f14510820d58651e9bc85eb4d3741/cmd/state/errors.go#L94-L94

Naatan (approved):

daved (approved):

PR #1427, opened 2021-06-18
Unset PIDs on termination of state-svc and state-tray

mdrohmann:
https://www.pivotaltracker.com/story/show/178563481

### DIFF
--- a/cmd/state-svc/internal/server/router.go
+++ b/cmd/state-svc/internal/server/router.go
@@ -19,7 +19,7 @@ func (s *Server) setupRouting() {
 	})
 
 	s.httpServer.GET(QuitRoute, func(c echo.Context) error {
-		s.Close()
+		s.shutdown <- struct{}{}
 		return nil
 	})
 }
--- a/cmd/state-svc/internal/server/router.go
+++ b/cmd/state-svc/internal/server/router.go
@@ -3,8 +3,6 @@ package server
 import (
 	"github.com/99designs/gqlgen/graphql/playground"
 	"github.com/labstack/echo/v4"
-
-	"github.com/ActiveState/cli/internal/logging"
 )
 
 const QuitRoute = "/__quit"
@@ -21,12 +19,7 @@ func (s *Server) setupRouting() {
 	})
 
 	s.httpServer.GET(QuitRoute, func(c echo.Context) error {
-		go func() {
-			err := s.Shutdown()
-			if err != nil {
-				logging.Errorf("Shutting down server failed: %v", err)
-			}
-		}()
+		s.shutdown <- struct{}{}
 		return nil
 	})
 }
--- a/cmd/state-svc/internal/server/router.go
+++ b/cmd/state-svc/internal/server/router.go
@@ -19,7 +19,7 @@ func (s *Server) setupRouting() {
 	})
 
 	s.httpServer.GET(QuitRoute, func(c echo.Context) error {
-		s.shutdown <- struct{}{}
+		s.Close()
 		return nil
 	})
 }
--- a/cmd/state-svc/internal/server/router.go
+++ b/cmd/state-svc/internal/server/router.go
@@ -19,7 +19,7 @@ func (s *Server) setupRouting() {
 	})
 
 	s.httpServer.GET(QuitRoute, func(c echo.Context) error {
-		s.shutdown <- struct{}{}
+		s.shutdown()
 		return nil
 	})
 }
--- a/cmd/state-svc/internal/server/server.go
+++ b/cmd/state-svc/internal/server/server.go
@@ -25,23 +25,20 @@ import (
 
 type Server struct {
 	cfg         *config.Instance
-	ctx         context.Context
-	done        chan struct{}
-	cancel      context.CancelFunc
+	shutdown    chan<- struct{}
 	graphServer *handler.Server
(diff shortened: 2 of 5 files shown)


cmd/state-svc/main.go
```diff
@@ -96,21 +97,39 @@ func run() error {
...
+		defer wg.Done()
+		defer close(sig)
+		select {
+		case oscall, ok := <-sig:
+			if !ok {
+				return
+			}
+			logging.Debug("system call:%+v", oscall)
+		case <-shutdown:
+		}
```
daved, line 118: Based on the prior suggestion, this should also be simplified.

daved (changes requested):

cmd/state-svc/internal/server/server.go
```diff
@@ -12,27 +13,35 @@ import (
...
+	done        chan struct{}
+	cancel      context.CancelFunc
 	graphServer *handler.Server
 	listener    net.Listener
 	httpServer  *echo.Echo
 	port        int
 }
 
-func New(cfg *config.Instance) (*Server, error) {
+func New(cfg *config.Instance, parentCtx context.Context) (*Server, error) {
```
daved, line 37: nit: The context type should always be the first arg.

cmd/state-svc/internal/server/router.go
```diff
@@ -21,12 +19,7 @@ func (s *Server) setupRouting() {
 	})
 
 	s.httpServer.GET(QuitRoute, func(c echo.Context) error {
-		go func() {
-			err := s.Shutdown()
-			if err != nil {
-				logging.Errorf("Shutting down server failed: %v", err)
-			}
-		}()
+		s.Close()
```
daved, line 22: nit: Please consider keeping the term Shutdown. Generally, close implies cleanup of various allocations, and shutdown implies the same as well as some graceful management of things like client connections.

cmd/state-svc/internal/server/server.go
```diff
@@ -12,27 +13,35 @@ import (
...
 	"github.com/ActiveState/cli/internal/config"
+	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/logging"
 )
 
 type Server struct {
+	cfg         *config.Instance
+	ctx         context.Context
+	done        chan struct{}
```
daved, line 29: nit: It might be worth noting that this channel tracks the reporting of doneness as opposed to the context done channel which indicates a request for doneness. It threw me off a bit reading through. Maybe rename to `wait`.

cmd/state-svc/internal/server/server.go
```diff
@@ -55,18 +64,43 @@ func (s *Server) Port() int {
 	return s.port
 }
 
+func (s *Server) Wait() {
+	<-s.done
+}
+
```
daved, line 70: Note: This presents a sort of crossroads of error handling. The errors encountered in the lifetime of Server could relate to 1. startup errors 2. runtime errors 3. shutdown errors. Currently, the start method returns the start errors. The runtime errors are likely all coming from start as well. And the shutdown errors are logged in start.

I suggest that start should be focused on start errors, wait should be focused on runtime errors (though, we may not even be tracking/handling any "runtime errors"), and shutdown errors should be exposed by the shutdown method being exported. The behavior in close and shutdown should probably be in one function, but I may misunderstand a need to separate them. If this is not as simple as a bit of code shuffling, it might be worth ignoring this comment.

daved (changes requested):

cmd/state-svc/internal/server/server.go
mdrohmann: I hear you.  The reason I do it this way, is: When we issue the shutdown request on the `/_quit` route, we still want that route to return with a 200 response, indicating: "The server is shutting down now".  So, we have to run the `Shutdown` in a go routine, but also wait for that go routine to return.  If we don't it was never able to terminate, because the main function returned before that.

So, in my first approach, I put that go routine (the one shutting down the server) in the state-svc main function, as that is the place to wait for it.  In this iteration, that shutdown-go function moved to the `server`, which as you are now rightly pointing out, leads to a very convoluted interface for a server.

Thinking about it this way, I believe that I should merge the two approaches:
- do all shutdown handling in the main routine (as in the first commit), but
- still use a context for server cancellation
- and just forward the cancel function to the server route, such that we can cancel the *global* context in the `/_quit` route.  

Does that sound correct?

mdrohmann: Of course, we could also go a different route, and in the `/_quit` route resolver, do the actual server shutdown (not in a go-routine), but as a consequence that route would never return, and the client issuing that `/_quit` request would have to expect a `connection interrupted` error as the only successful outcome.  That does not feel like the the correct approach though...

daved: I agree that 200 responses make sense to indicate the shutdown request was successful. I stand by my original suggestion to achieve any improvement so long as it's just shuffling a little code from here. It's clear you understand where I'm coming from and while I haven't dug deep into your comment above, I'm happy to trust your suggestion.

daved (approved):

Naatan (approved):

PR #1428, opened 2021-06-21
Always stop services during state clean (config|cache|uninstall)

mdrohmann:
https://www.pivotaltracker.com/story/show/178564019

### DIFF
--- a/internal/runners/clean/run_win.go
+++ b/internal/runners/clean/run_win.go
@@ -14,12 +14,9 @@ import (
 	"github.com/ActiveState/cli/internal/appinfo"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/exeutils"
-	"github.com/ActiveState/cli/internal/fileutils"
-	"github.com/ActiveState/cli/internal/installation"
 	"github.com/ActiveState/cli/internal/language"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/scriptfile"
 )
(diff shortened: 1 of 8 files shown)


internal/runners/clean/stop.go
```diff
@@ -0,0 +1,47 @@
...
+	// Todo: https://www.pivotaltracker.com/story/show/177585085
+	// Yes this is awkward right now
+	if err := installation.StopTrayApp(cfg); err != nil {
+		if ignoreErrors {
+			return errs.AddTips(
+				locale.WrapError(err, "clean_stop_tray_failure", "Cleanup interrupted, because a running {{.V0}} process could not be stopped.", trayInfo.Name()),
+				cleanForceTip)
+		} else {
+			out.Print(locale.Tl("clean_stop_tray_warning", "[ERROR]Failed to stop running {{.V0}} process.[/RESET] Continuing anyways, because --force flag was provided.", trayInfo.Name()))
+		}
```
daved, line 29: I suggest making this !ignoreErrors and dropping the else. And the same below.

internal/runners/clean/run_win.go
```diff
@@ -66,28 +66,7 @@ func removeConfig(configPath string, out output.Outputer) error {
...
-		exitCode, _, err := exeutils.Execute(svcInfo.Exec(), []string{"stop"}, nil)
-		if err != nil {
-			return errs.Wrap(err, "Stopping %s returned error", svcInfo.Name())
-		}
-		if exitCode != 0 {
-			return errs.New("Stopping %s exited with code %d", svcInfo.Name(), exitCode)
-		}
-	}
-
+func removeInstall(cfg configurable, logFile, configPath string) error {
```
daved, line 66: configurable might be able to be dropped from the params.

daved (changes requested):

internal/runners/clean/stop.go
mdrohmann: :facepalm:

daved (approved):

internal/runners/clean/stop.go
```diff
@@ -0,0 +1,44 @@
...
+	// Stop state-svc before accessing its files
+	if fileutils.FileExists(svcInfo.Exec()) {
+		_, _, err := exeutils.Execute(svcInfo.Exec(), []string{"stop"}, nil)
+		if err != nil {
+			if !ignoreErrors {
+				return errs.AddTips(
+					locale.WrapError(err, "clean_stop_svc_failure", "Cleanup interrupted, because a running {{.V0}} process could not be stopped.", svcInfo.Name()),
+					cleanForceTip)
+			}
+			out.Print(locale.Tl("clean_stop_svc_warning", "[ERROR]Failed to stop running {{.V0}} process.[/RESET] Continuing anyways, because --force flag was provided.", svcInfo.Name()))
```
Naatan, line 40: I think this can be `out.Error()` and you can drop the `[ERROR]`

Naatan (changes requested):

Naatan (approved):

PR #1429, opened 2021-06-22
Add ask-once tracker as runbit

daved:
https://www.pivotaltracker.com/story/show/178438700


internal/globaldefault/default.go
```diff
@@ -80,3 +79,23 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 
 	return nil
 }
+
+// SetAskedOnce tracks whether it has been called before using the
+// configuration type. A response of true indicates that the config has been
+// updated. A response of false indicates that the config has not been updated.
+// If the configuration "set" call fails, this functions always returns true.
+func SetAskedOnce(cfg DefaultConfigurer) bool {
```
Naatan, line 87: It feels awkward to me that we're saving the state of this prompt in the `globaldefault` pkg whilst not actually doing anything around prompts in this pkg. I think this would be more at home inside the activate package. Leave the `globaldefault` package for setting up the global default, not for handling UX around it.

Naatan (changes requested):

internal/runbits/promptable/promptable.go
```diff
@@ -0,0 +1,51 @@
...
+// A response of false indicates that the config has not been updated. If the
+// configuration "set" call fails, this functions always returns true.
+func SetPrompted(cfg Configurer, key OnceKey) bool {
+	asked := cfg.GetBool(string(key))
+	if asked {
+		logging.Debug(fmt.Sprintf("%s: already asked", key))
+		return false
+	}
+
+	logging.Debug(fmt.Sprintf("%s: setting asked", key))
```
Naatan, line 46: fyi `logging.Debug` takes printf arguments.

internal/runbits/promptable/promptable.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/output"
+)
+
+// IsPromptable reports whether the output type permits prompts to be shown.
+func IsPromptable(out output.Outputer) bool {
+	return out.Type() == output.PlainFormatName
```
Naatan, line 12: Output has nothing to do with this. Look into how we're reading `--non-interactive`.

Naatan (changes requested):
Looking good except for the non-interactive check!

Naatan (approved):

PR #1430, opened 2021-06-22
Update hard coded os platform uuids

daved:
https://www.pivotaltracker.com/story/show/178437290


daved:
Looking over [this](https://github.com/ActiveState/TheHomeRepot/blob/801526838e62bf7e3fdd7409434e938d720974b6/service/dashboard/src/elm/PlatformMapping.elm), it seems that only one value needs to be updated.

Naatan (approved):

PR #1434, opened 2021-06-24
Update update notice

daved:
https://www.pivotaltracker.com/story/show/178578217

### DIFF
--- a/internal/runners/update/update.go
+++ b/internal/runners/update/update.go
@@ -15,7 +15,6 @@ import (
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/prompt"
 	"github.com/ActiveState/cli/internal/svcmanager"
-	"github.com/ActiveState/cli/internal/version"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/project"
 )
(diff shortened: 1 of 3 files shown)


daved:
@Naatan This will be a mess to test via E2Es. Would it be OK to make this its own function and add some unit tests?

internal/runners/update/update.go
```diff
@@ -86,6 +87,15 @@ func (u *Update) Run(params *Params) error {
 	}
 
 	u.out.Print(locale.Tl("version_updating_deferred", "Version update to {{.V0}}@{{.V1}} has started and should complete in seconds.\nRefer to log file [ACTIONABLE]{{.V2}}[/RESET] for progress.", up.Channel, up.Version, up.Logfile))
+
+	curVrsn, err := version.ParseStateToolVersion(constants.Version)
+	if err != nil {
+		return locale.WrapError(err, "err_update_parse_version", "Cannot parse current version")
+	}
+	if (channel == constants.ReleaseBranch || channel == constants.BetaBranch || channel == constants.ExperimentalBranch) && !version.IsMultiFileUpdate(curVrsn) {
```
Naatan, line 95:
```suggestion
	if (channel == constants.ReleaseBranch || channel == constants.BetaBranch || channel == constants.ExperimentalBranch) && !version.IsMultiFileUpdate(curVrsn) {
```

You're parsing the current version of the state tool to see if it's NOT a multi file update, which can't ever be true because by definition any update we release is a multi file update.

@mdrohmann can you advice? I think we'd need to put this in the transitional updater, or in the installer itself.

Naatan (changes requested):

internal/runners/update/update.go
daved: I see; Got overly focused on a particular solution without thinking thoroughly. @mdrohmann Would a message [here](https://github.com/ActiveState/cli/blob/78406b59921f97365c8fd4ec5ba5c1552f51ff0e/cmd/state-transition-update/main.go#L45) be the correct point to communicate the "restart shell" message?

mdrohmann (approved):

Naatan (dismissed):
Well that was easy.. !

Naatan (approved):

PR #1436, opened 2021-06-25
Fix config file corruption issue

mdrohmann:
https://www.pivotaltracker.com/story/show/178666781

It appears that with the gofrs/flock package implementation we still run into situations where a lock is acquired when it shouldn't (because a different process is already taking it).

As there are parts of the `flock` package code that I do not quite understand, I simplified the `internal/osutils/pidlock` package that we had previously, and made it work successfully.  I had to adapt it in a way such that the lockfile is not necessarily removed after we release the lock to it.   https://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition seems to be the explanation for this.

Tested like this:

- Activate a python 3.9 project with the `--default` flag

```
for ((i=0; i<100; i++)); do setsid python3 -c 'print("hi");'; done
```
fails after a while, because the configuration file gets corrupted.

After applying this patch, I can run
```
for ((i=0; i<250; i++)); do setsid python3 -c 'print("hi");'; done
```
and the configuration file gets *not* corrupted, **but** we get timeouts for 1-3 processes that have to wait for more than 5 seconds to acquire the lock.


internal/osutils/lockfile/lock_test.go
```diff
@@ -0,0 +1,198 @@
...
+				if sb == "DENIED" {
+					done <- "DENIED"
+					return
+				}
+				assert.Equal(tt, "LOCKED", string(buf[:6]))
+
+				done <- "LOCKED"
+
+				// wait for the signal to kill process and to release the lock
+				<-ctx.Done()
```
daved, line 128: I suggest placing the select/timeout here and using the waitgroup where the timeout is now. The deferred wait call significantly disconcerting.

daved: Though, if it's the simplest way for you to achieve your goal, no worries.

daved (changes requested):

internal/osutils/lockfile/lock_test.go
mdrohmann: I don't remember why I added the second go function.  I removed it...

internal/osutils/lockfile/lock_test.go
```diff
@@ -0,0 +1,193 @@
...
+	t.Run("stress-test", func(tt *testing.T) {
+		// stress tests runs numProcesses in parallel, and only one should get the lock
+		numProcesses := 10
+
+		lockFile := filepath.Join(tmpDir, "stress-test")
+
+		done := make(chan string, numProcesses+1)
+		defer close(done)
+		var wg sync.WaitGroup
+		defer wg.Wait()
```
daved, line 96: not nit, but nit for now: This is just so janky that I need to note it. Maybe we can talk about this sort of thing in architecture talks at some point.

daved (dismissed):

mdrohmann:
After running
```
for ((i=0; i<250; i++)); do setsid python3 -c 'print("hi");'; done
```
several more times, the configuration file got corrupted again. :-(  So, I had to do more digging, and found out two things:
1. Synchronizing the configuration file write, makes things more reliable, and it might be necessary even to ensure that the contents are actually written to the disk while the lock is held.  Otherwise the OS might delay the write.  I could not 100% verify that this is really necessary though
2.  And more important: Adding extensive logging to the code, revealed that some locks were never released.  I eventually released that this happens in the deferred analytics package: The write to the config file here is never guaranteed to complete before the main function just exits.  Such it can leave us with a half written configuration file. So, I temporarily de-activated this mechanic, and filed a follow-up story to address: https://www.pivotaltracker.com/story/show/178680862

It is very well possible that (2.) and removing the lockfile, were the only problems ever!  We might want to evaluate this later, but for now, I would like to merge it as is, because this is code that (so far) survived my stress-tests.

mdrohmann:
Okay.  The deferred analytics events are now working, and not corrupting our configuration file anymore.
I tested the stress-test command several times with 500 parallel executors.  It is very slow, but it works.

cmd/state-installer/installer.go
```diff
@@ -32,7 +32,9 @@ func main() {
 		if panics.HandlePanics() {
 			exitCode = 1
 		}
-		events.WaitForEvents(1*time.Second, rollbar.Close)
+		if err := events.WaitForEvents(1*time.Second, rollbar.Close); err != nil {
+			logging.Warning("Failed to wait for rollbar to close: %v", err)
```
mdrohmann, line 36: I added a warning message to the logs for cases where we run into these timeouts.

I couldn't use `logging.Error`, because that will inject random `Rollbar: unknown token` printouts, as rollbar likely is already in a half-closed state...

internal/config/instance.go
```diff
@@ -44,22 +44,18 @@ type Instance struct {
 	installSource string
 	lock          *flock.Flock
 	data          map[string]interface{}
-	// lockMutex ensures that file lock can be held only once per process.  Theoretically, this should be ensured by the `flock` package, but it isn't.  So, we need this hack.
```
mdrohmann, line 47: After realizing that the removal of the lock file caused problems, I removed this `lockMutex` and ensured that it still works.

Apparently, `gofrs/flock` **is** thread-safe. The actual issue that I observed when adding this comment, was due us removing the lockfile.  More details are here if interested: https://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition

internal/config/instance.go
```diff
@@ -364,6 +353,10 @@ func (i *Instance) save() error {
 		return errs.Wrap(err, "Could not write config file")
 	}
 
+	if err = f.Sync(); err != nil {
```
mdrohmann, line 356: I added a file sync, as the lockfile really just synchronizes the time slice in which processes run these commands.  
As file operations (without sync) are usually deferred by the OS, it is possible that a file operation is actually executed (written to the hard-disk) while the lock is released already.  Reasonable OS implementations should ensure that there is still some sensible order of these writes, but I thought it wouldn't hurt not to rely on the OS here.

daved (approved):

internal/analytics/deferred.go
```diff
@@ -98,69 +104,91 @@ type Configurable interface {
...
+		}
+		events = append(events, event)
+	}
+
+	return events, nil
+}
+
+func sendDeferred(cfg Configurable, sender func(string, string, string, map[string]string) error) error {
+	// move deferred data file, so it is not being appended anymore
+	outboxFile := filepath.Join(cfg.ConfigPath(), "deferred.outbox")
```
Naatan, line 148: This could still be a problem if two processes both try to flush the deferred analytics at the same time. I think we can just move it to a temp file that is uniquely named (ie. use a timestamp)

internal/analytics/deferred.go
```diff
@@ -22,13 +24,17 @@ type deferredData struct {
 	Dimensions map[string]string
 }
 
-const deferredCfgKey = "deferrer_analytics"
 const deferrerFileName = "deferrer"
+const deferredDataFileName = "deferred_data"
```
Naatan, line 28: I can't tell why we need both. The first one we just seem to use to look at the time it was last modified, which the second one should already provide.

Naatan (changes requested):

internal/analytics/deferred.go
mdrohmann: The first one is only created on the *first* deferral event, so basically the first time a deferred event is created (after the last "flush").  That way we can ensure that we flush at least once a day: https://github.com/ActiveState/cli/blob/master/internal/analytics/deferred.go#L43-L44

mdrohmann: With the patch that you introduced, this would never trigger, if a user runs only deferred commands, and if they run at least one of them per day.

PR #1440, opened 2021-07-05
`state push` creates fork if not authorized

MDrakos:
https://www.pivotaltracker.com/story/show/178506609


internal/runners/push/push.go
```diff
@@ -162,7 +162,28 @@ func (r *Push) Run(params PushParams) error {
 	// update the project at the given commit id.
 	err = model.UpdateProjectBranchCommitWithModel(pjm, branch.Label, commitID)
 	if err != nil {
-		return locale.WrapError(err, "push_project_branch_commit_err", "Failed to update new project {{.V0}} to current commitID.", pjm.Name)
+		if errs.Matches(err, &model.ErrUpdateBranchAuth{}) {
```
Naatan, line 165: Let's invert this check and return early, then this whole block doesn't need to be indented.

internal/runners/push/push.go
```diff
@@ -232,6 +239,29 @@ func (r *Push) getNamespace() (*project.Namespaced, error) {
 	return ns, nil
 }
 
+func (r *Push) promptNamespace() (*project.Namespaced, error) {
+	owner := authentication.Get().WhoAmI()
```
Naatan, line 248: Since you're calling this from inside the forking flow the namespace param seems to be ignored. Meaning that even when someone run `state push foo/bar` it will still prompt for an owner/name.

locale/en-us.yaml
```diff
@@ -1221,6 +1221,8 @@ push_description:
   other: Push your latest changes to the platform
 push_prompt_owner:
   other: "Since this is your first time running \"push\" you'll need to create a new project on the ActiveState Platform.\nWho would you like the owner of this project to be?"
+push_prompt_auth:
+  other: "You are not authorized to push to this project.\nEither get the project owner to invite you, or create a new project with these changes.\nWould you like to create a new project with your changes now?"
```
Naatan, line 1225:
```suggestion
push_prompt_auth:
  other: |
    You are not authorized to push to this project.
    Either get the project owner to invite you, or create a new project with these changes.
    Would you like to create a new project with your changes now?
```

Naatan (changes requested):

internal/runners/push/push.go
MDrakos: If a namespace parameter is passed then it is likely that the user will have permission to create a project under the given org with the given name. In that case, we will not hit the fork logic at all.

If the user does not have permission to create a project with the namespace they passed in then we will have to prompt them for a new one. So in this case I think we would _want_ to ignore their parameters. Or we could at least use the name as the default in the prompt as they are not authorized to create a project with the given org.

PR #1442, opened 2021-07-07
Config now uses sqlite

Naatan:


internal/analytics/deferred.go
```diff
@@ -44,7 +49,7 @@ func isDeferralDayAgo(cfg Configurable) bool {
 	return diff > 24*time.Hour
```
mdrohmann, line 49: This checks if the modification time of the deferrer log is older than 24 hours.  But as we are appending to that file, running a `state exec` command every day would ensure that this never happens.  

The simple solution I had was to have two different files: one that is only touched on the first deferral, and the log.  At least on linux, you could probably also compare against the "creation" time of that log file, but I think that is not cross-platform...

pkg/platform/authentication/auth.go
```diff
@@ -44,12 +44,13 @@ type Auth struct {
...
+	Close() error
 }
 
-// Get returns a cached version of Auth
-func Get() *Auth {
+// LegacyGet returns a cached version of Auth
+func LegacyGet() *Auth {
 	if persist == nil {
-		cfg, err := config.Get()
+		cfg, err := config.New()
```
mdrohmann, line 53: Hmm... Technically, creating a config instance with `New` means that we create a new `singlethread` instance too, no?  And that would mean that a process could have two `singlethread`s...

pkg/projectfile/projectfile.go
```diff
@@ -486,10 +487,11 @@ func Parse(configFilepath string) (*Project, error) {
 		return nil, errs.Wrap(err, "project.Init failed")
 	}
 
-	cfg, err := config.Get()
+	cfg, err := config.New()
```
mdrohmann, line 490: This may be the same problem as in the `auth.go` file.  This should create config instance with a second `singlethread` instance, next to the one we initialize in the `main.go` file...

internal/locale/locale.go
```diff
@@ -34,14 +35,20 @@ func init() {
...
 	locale := getLocaleFlag()
-	cfg, err := config.Get()
-	if err == nil {
-		setErr := cfg.Set("Locale", "en-US")
-		if err != nil {
-			logging.Error("Could not set locale entry in config, error: %v", setErr)
-		}
-		if locale == "" {
+	if locale == "" {
+		cfg, err := config.New()
```
mdrohmann, line 39: By creating a new instance here, we would also create a new `singlethread`.  I think that also happens in  `auth.go`.

internal/locale/locale.go
```diff
@@ -96,18 +103,20 @@ func getLocaleFlag() string {
...
+func Set(localeName string) (rerr error) {
 	if !funk.Contains(Supported, localeName) {
 		return errs.New("Locale does not exist: %s", localeName)
 	}
 
 	translateFunction, _ = i18n.Tfunc(localeName)
 	_ = translateFunction
 
-	cfg, err := config.Get()
+	cfg, err := config.New()
```
mdrohmann, line 114: same as above: New `singlethread` instance.

internal/svcmanager/svcmanager.go
```diff
@@ -64,11 +64,6 @@ func (m *Manager) Ready() bool {
 		return true
 	}
 
-	if err := m.cfg.Reload(); err != nil {
```
mdrohmann, line 67: :heart: that we don't need to reload anymore!

mdrohmann (changes requested):
nit: I wonder if `internal/config/test.db` is used somewhere, or can it be removed?

I am mostly worried that we are creating several `singlethread` instances everytime `config.New` is called which defeats the purpose of that package.  But maybe it doesn't matter, because `sqlite` is thread-safe?  In that case, we should probably not make the config instance have a `singlethread` at all, though...

internal/rtutils/singlethread/singlethread.go
```diff
@@ -0,0 +1,49 @@
...
+type Thread struct {
+	cbs    chan (callback)
+	close  chan (struct{})
+	closed bool
+}
+
+func New() *Thread {
+	t := &Thread{
+		make(chan (callback)),
+		make(chan (struct{}), 1),
```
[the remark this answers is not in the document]
Naatan: No reason, just my still coming to grips with channels. I'll refactor to use `close()`.

pkg/platform/authentication/auth.go
Naatan: I have a rationale for this;

Long term there should really only be one `config.New()` per process, though exceptions to that rule are of course possible. BUT when there are multiple `config.New()` calls it's unlikely that they'd be dealing with the same config keys.

We could also have a single singlethread instance, but I was trying to avoid getting THAT far into the weeds.

For now I consider this to be a non-issue, but it's something we should keep in mind when we design the reusable system to set up commands, which would include the primer stuff that would be necessary to properly address these types of use cases.

mdrohmann: Okay.  That is fine with me.

test/integration/update_int_test.go
```diff
@@ -175,13 +175,6 @@ func (suite *UpdateIntegrationTestSuite) pollForUpdateInBackground(output string
 
 	suite.Require().Equal(len(resultLogfile), 2, "expected to have logfile in output %s", output)
 	logpath := strings.TrimSpace(resultLogfile[1])
-
-	for x := 0; x < 5; x++ {
-		if fileutils.FileExists(logpath) {
-			break
-		}
-		time.Sleep(1 * time.Second)
-	}
```
mdrohmann: Are you sure that this is not necessary anymore?  I think the mocked updates may install fast enough, but we might need this again once we add "real" tests.

mdrohmann (dismissed):

test/integration/update_int_test.go
Naatan: This wasn't here prior to my adding it to try and fix the logFile issue. Since the issue got resolved differently I removed it. Are you saying it should've had this all along?

internal/logging/defaults.go
```diff
@@ -147,36 +162,40 @@ func init() {
...
-		return
-	}
+	path, isset := os.LookupEnv(constants.LogEnvVarName)
+	if isset {
+		// Use log file provided by env var
+		datadir = filepath.Dir(path)
+		datafile = filepath.Base(path)
+	} else {
+		// Clean up old log files
+		datadir, err := storage.AppDataPath()
```
mdrohmann, line 172:
```suggestion
                var err error
		datadir, err = storage.AppDataPath()
```

mdrohmann (approved):

PR #1443, opened 2021-07-08
Update project entries in systray

MDrakos:
https://www.pivotaltracker.com/story/show/178610570

### DIFF
--- a/cmd/state-tray/internal/menu/projects.go
+++ b/cmd/state-tray/internal/menu/projects.go
@@ -63,10 +63,8 @@ func trimEntry(entry string) string {
 		return entry
 	}
 
-	start := entryLength - maxEntryLength
-	end := start + maxEntryLength
 	e := []rune(entry)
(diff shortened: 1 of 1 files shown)


cmd/state-tray/internal/menu/projects.go
```diff
@@ -50,6 +57,18 @@ func (u *LocalProjectsUpdater) Update(projects []*graph.Project) {
...
 }
 
+func trimEntry(entry string) string {
+	entryLength := utf8.RuneCountInString(entry)
+	if entryLength <= maxEntryLength {
+		return entry
+	}
+
+	start := entryLength - maxEntryLength
+	end := start + maxEntryLength
```
Naatan, line 67: I don't think you need this. End would just be `entryLength`.

Naatan (changes requested):

Naatan (approved):

PR #1444, opened 2021-07-09
Replace cgo based sqlite with go based sqlite

Naatan:
https://www.pivotaltracker.com/story/show/178828288


mdrohmann (approved):
Nice, fast digging.
Let's just hope that modernc.go/sqlite is as stable as the c-library based one.

Naatan:
> Nice, fast digging.
> Let's just hope that modernc.go/sqlite is as stable as the c-library based one.

Our use of sqlite is so basic that hopefully this won't matter too much. Definitely making me realize how much I rely on Github's star/fork stats to gauge package stability though..

internal/config/instance.go
```diff
@@ -26,6 +26,7 @@ type Instance struct {
 	thread      *singlethread.Thread
 	closeThread bool
 	db          *sql.DB
+	closed      bool
```
mdrohmann, line 29: This should probably be guarded by a mutex.  Or maybe you can use a `sync.Once` instance in the `Close()` function to ensure that the thread is closed only once.

mdrohmann (changes requested):

mdrohmann (approved):

PR #1447, opened 2021-07-13
Add build log progress report

mdrohmann:
https://www.pivotaltracker.com/story/show/178590179

 https://www.pivotaltracker.com/story/show/178831673


pkg/platform/runtime/setup/buildlog/messages.go
```diff
@@ -0,0 +1,164 @@
+package buildlog
```
mdrohmann, line 1: I added this file comprising a serialization for all the messages that we can read on the build log streamer websocket connection.  This seemed necessary, as we are now distinguishing between more (and different) message types.

It also increases the legibility of the buildlog.go code IMO.

pkg/platform/api/headchef/headchef.go
```diff
@@ -187,11 +187,11 @@ func (r *Client) reqBuildSync(buildReq *headchef_models.V1BuildRequest) (BuildSt
 		payloadType := *created.Payload.Type
 
 		switch payloadType {
-		case headchef_models.BuildStatusResponseTypeBuildCompleted:
+		case headchef_models.V1BuildStatusResponseTypeBuildCompleted:
 			return Completed, created.Payload, nil
-		case headchef_models.BuildStatusResponseTypeBuildFailed:
-			return Failed, created.Payload, locale.WrapError(ErrBuildFailedResp, "headchef_build_failure", "Build Failed: {{.V0}}", created.Payload.Message)
```
mdrohmann, line 193: Note: I removed the `ErrBuildFailedResp`, because the information is already returned via the `Failed` flag.  In the `FetchBuildResult` function, we still want to process the returned information, and it didn't feel right to just remove the error there.

internal/language/language.go
```diff
@@ -129,7 +129,7 @@ func MakeByNameAndVersion(name, version string) (Language, error) {
 	if strings.ToLower(name) == Python2.Requirement() {
 		parts := strings.Split(version, ".")
 		if len(parts) == 0 || parts[0] == "" {
-			return Unknown, locale.NewError("err_invalid_version", "Invalid langauage version number: {{.V0}}", version)
+			return Unknown, locale.NewError("err_invalid_language_version", "Invalid language version number: {{.V0}}", version)
```
mdrohmann, line 132: This is just an error that I found randomly while trying to activate a project with an invalid language version.

pkg/platform/runtime/setup/buildlog/artifactlogs.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/gorilla/websocket"
+)
+
+type logConnection struct {
+	conn *websocket.Conn
+	log  *ArtifactLog
+}
+
+// ArtifactLogs manages websocket connections to the build-log-streamer for artifact specific logs
+// Unfortunately, we need to spawn a new connection for every artifactID that prints its loglines in real-time. If artifact logs for builds that happened in the past are requested, the information can be streamed in a single connection.
```
mdrohmann, line 19: During my tests this morning, I realized that I had to add a separate connection for each and every artifactID that I want to receive real-time build log information from.  Fortunately, this is only relevant in the `ACTIVESTATE_BUILD_VERBOSE=true` case right now.

pkg/platform/runtime/setup/setup.go
```diff
@@ -182,17 +199,12 @@ func (s *Setup) update() error {
...
-		return errs.Wrap(err, "Failed to select setup implementation")
-	}
-
 	storedArtifacts, err := s.store.Artifacts()
 	if err != nil {
 		return locale.WrapError(err, "err_stored_artifacts", "Could not unmarshal stored artifacts, your install may be corrupted.")
 	}
 
-	alreadyInstalled := setup.ReusableArtifacts(changedArtifacts, storedArtifacts)
+	alreadyInstalled := reusableArtifacts(buildResult.BuildStatusResponse.Artifacts, storedArtifacts)
```
mdrohmann, line 207: This addresses a problem where the `alreadyInstalled` slice included too many artifacts.  They sneaked in there after a failed build was incompletely installed and reverted.  The artifacts that had been installed as part of the failed commit were never removed and therefore appeared incorrectly in that list.

The solution proposed in this PR, is simpler as it is  the same for camel and alternative builds) by just comparing the `storedArtifacts` with the list of the artifacts that the head-chef returns in its build status response.

mdrohmann (commented):
I found a few issues during testing that I addressed.

pkg/platform/runtime/setup/setup.go
mdrohmann: Follow-up story is here: https://www.pivotaltracker.com/story/show/178901061

pkg/platform/runtime/setup/buildlog/messages.go
```diff
@@ -0,0 +1,164 @@
...
+const (
+	BuildStarted MessageEnum = iota
+	BuildSucceeded
+	BuildFailed
+	ArtifactStarted
+	ArtifactSucceeded
+	ArtifactFailed
+	ArtifactProgress
+	Heartbeat
+	UnknownMessage
```
daved, line 22: nit: Similar to this, I suggest using 0 for something like Unset. That way, if a struct has a MessageEnum field, the zero value makes it clear that it was not set.

pkg/platform/runtime/setup/buildlog/artifactlog.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// NewArtifactLog subscribes to events on the connection, and forwards build log events via the events handler
+func NewArtifactLog(artifactID artifact.ArtifactID, conn BuildLogConnector, events Events) (*ArtifactLog, error) {
+	errCh := make(chan error)
+
+	go func() {
+		defer close(errCh)
+
+		for {
```
daved, line 21: This seems like it may need something more to ensure it comes to an end. Context type, maybe. I also wonder if the error returned by conn.ReadJSON could be EOF and whether that can be handled here for callers such that the loop ends, but no error is returned.

pkg/platform/runtime/setup/buildlog/artifactlogs.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	err1 := lc.conn.Close()
+	err2 := lc.log.Wait()
+	delete(alm.logs, artifactID)
+	if err1 != nil {
+		return errs.Wrap(err1, "Failed to close websocket connection")
+	}
+	if err2 != nil {
+		// we just log this error, as it is probably just a "closed connection error"
```
daved, line 65: nit: Consider checking for that error and ignoring it if it should be.

daved (changes requested):

internal/runbits/buildlogfile/buildlogfile.go
```diff
@@ -0,0 +1,245 @@
+package buildlogfile
```
Naatan, line 1: This doesn't feel like it should be inside a runbit, or deal with output. Are we doing this due to the way event handling is set up? 

I would expect to see the majority of this logic somewhere inside the runtime package dir (some sub-package). The output part should be entirely separate from this.

internal/runbits/buildlogfile/buildlogfile.go
```diff
@@ -0,0 +1,245 @@
...
+	unsignedLogURI string
+}
+
+// verboseLogging returns true if the user provided an environment variable for it
+func verboseLogging() bool {
+	return os.Getenv(constants.LogBuildVerboseEnvVarName) == "true"
+}
+
+// downloadAndAddArtifactLog downloads an artifact build log and adds it to the build log file
+func (bl *BuildLogFile) downloadAndAddArtifactLog(ul artifactLog, ctx context.Context) error {
```
Naatan, line 49: Nit: this should be under the `New()` function.

internal/runbits/buildlogfile/buildlogfile.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		if err := bl.BuildArtifactProgress(ul.artifactID, ul.artifactName, am.Timestamp, am.Body.Message, am.Body.Facility, am.PipeName, am.Source); err != nil {
+			return errs.Wrap(err, "Failed to write build log line")
+		}
+	}
+	return nil
+}
+
+func New(out output.Outputer) (*BuildLogFile, error) {
+	logFile, err := os.CreateTemp("", fmt.Sprintf("build-log-%s", time.Now().Format("060102030405")))
```
Naatan, line 87: Can we use unix timestamp instead? This is needlessly vague.

internal/runbits/buildlogfile/buildlogfile.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return nil, errs.Wrap(err, "Failed to create temporary build log file")
+	}
+
+	bl := &BuildLogFile{out: out, logFile: logFile}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	var wg sync.WaitGroup
+	bl.logsCh = make(chan artifactLog)
+	for i := 0; i < maxConcurrency; i++ {
```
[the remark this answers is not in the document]
Naatan: Might be more appropriate/simpler to use workerpool here, eg:

http://github.com/ActiveState/cli/blob/3dc202147f99ba6f4a0711e9baf11961dc725f90/pkg/platform/runtime/setup/setup.go#L328-L328

internal/runbits/buildlogfile/buildlogfile.go
Naatan: Seeing the rest of the PR, it's clearer to me that in fact a lot of this logic DOES live in the runtime package. Which now makes me wonder; why is there so much logic in this runbit? It's not just passing data along, it's doing requests to get log files from the server. It feels a little unclear to me how the responsibilities of these packages is determined.

pkg/platform/runtime/setup/events/consumer_test.go
```diff
@@ -190,8 +203,11 @@ func TestRuntimeEventConsumer(t *testing.T) {
 					evCh <- ev
 				}
 			}()
-			err := consumer.Consume(evCh)
-			assert.NoError(t, err)
+
+			for ev := range evCh {
+				err := consumer.Consume(ev)
+				assert.NoError(t, err)
+			}
```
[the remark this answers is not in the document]
mdrohmann: I do not remember why I changed this, but I think you are right, and I'll change it back, such that the for loop is in the `Consume` function.

pkg/platform/runtime/setup/buildlog/artifactlog.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+)
+
+// ArtifactLog wraps a handler to listen for real-time build log messages for a specific artifact
+type ArtifactLog struct {
+	errCh chan error
+}
+
+// NewArtifactLog subscribes to events on the connection, and forwards build log events via the events handler
+func NewArtifactLog(artifactID artifact.ArtifactID, conn BuildLogConnector, events Events) (*ArtifactLog, error) {
```
Naatan, line 15: Can we make it a bit clearer from the structure of the code how this differs from `New()`? I'm grokking both as "a thing that logs artifact progress". Not clear to me what the distinction is without reading through all the code.

pkg/platform/runtime/setup/setup.go
```diff
@@ -351,7 +365,12 @@ func (s *Setup) installFromBuildLog(buildResult *model.BuildResult, artifacts ar
 	}
 	defer conn.Close()
 
-	buildLog, err := buildlog.New(artifacts, conn, s.events, *buildResult.Recipe.RecipeID)
+	alreadyBuilt := make(map[artifact.ArtifactID]struct{})
+	for _, d := range downloads {
+		alreadyBuilt[d.ArtifactID] = struct{}{}
+	}
+	artifactLogManager := buildlog.NewArtifactLogs(ctx, s.events)
+	buildLog, err := buildlog.New(artifacts, alreadyBuilt, conn, artifactLogManager, s.events, *buildResult.Recipe.RecipeID)
```
Naatan, line 373: buildlog taking artifactLogManager as an arg looks like a bit of a red flag to me. You seem to ONLY initialize it to then pass it into buildlog.New(), and they're both part of the same internal package. Doesn't seem we need to be exposing `buildlog.NewArtifactLogs` at all?

pkg/platform/runtime/setup/buildlog/buildlog.go
```diff
@@ -67,69 +65,93 @@ func New(artifactMap map[artifact.ArtifactID]artifact.ArtifactRecipe, conn Build
...
 					continue
 				}
+				// ignore already built artifacts (they have been counted already)
+				if _, ok := alreadyBuilt[m.ArtifactID]; ok {
+					continue
+				}
+				events.ArtifactBuildStarting(m.ArtifactID)
+
+				// if verbose build logging is requested: Also subscribe to log messages for this artifacts
+				if os.Getenv(constants.LogBuildVerboseEnvVarName) == "true" {
```
Naatan, line 96: Should just store the result of this somewhere, it's not going to change. No need to do a `os.Getenv()` call every time.

Naatan (changes requested):
This PR is HUGE.. I can't do a proper review without spending undue amounts of time. Overall the quality looks good to me, but this is probably a good case study for one of our architecture talks.

internal/runbits/buildlogfile/buildlogfile.go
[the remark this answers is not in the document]
mdrohmann: I see your point.  This can indeed block (in the consumer loop) in case we are downloading a lot of build logs, and if the download is hanging. I just wanted to limit the number of concurrent go functions here.

internal/runbits/buildlogfile/buildlogfile.go
```diff
@@ -0,0 +1,245 @@
...
+	bl.out.Print(locale.Tl("view_build_logfile_info", "View the Build Log to follow the build progress in detail: [ACTIONABLE]{{.V0}}[/RESET]", bl.logFile.Name()))
+	return bl.writeMessage("== Scheduled building of %d artifacts ==", totalArtifacts)
+}
+
+// BuildCompleted writes a message that the build has completed
+func (bl *BuildLogFile) BuildCompleted(withFailures bool) error {
+	outcome := "SUCCESSFULLY"
+	if withFailures {
+		outcome = "with FAILURES"
+	}
```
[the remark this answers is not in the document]
mdrohmann: I didn't do localization in the logfile. That seemed wrong, but I could be convinced otherwise. @Naatan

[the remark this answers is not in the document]
Naatan: Nah not worth it since 98% of it comes from headchef which isn't localized.

internal/runbits/buildlogfile/buildlogfile.go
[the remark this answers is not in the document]
mdrohmann: I think I know how to solve this: I will drop sending requests (in `addBuildLogs` if the request takes more than one second.)  We can log that to rollbar.  If we are getting to many rollbar errors, we can increase the `maxConcurrency` number.

internal/runbits/buildlogfile/buildlogfile.go
mdrohmann: I think that is a valuable criticism:  The download of the log files on failures, doesn't need to happen here.  That should probably move up the chain, and I think that the `consumer` would be the best place to handle that.

pkg/platform/runtime/setup/buildlog/artifactlog.go
mdrohmann: I'll add a comment here explaining it:  We should not need it at all (https://github.com/ActiveState/cli/pull/1447#discussion_r670620444), because ideally the `BuildLog` can handle these messages on the websocket-connection that is given to it.  My hope is that once https://www.pivotaltracker.com/story/show/178901762 is resolved, that we can get rid of this struct completely...

pkg/platform/runtime/setup/setup.go
mdrohmann: Agreed that it is awkward! See my comment above.  Adding the line 372 to the `New` function creates a problem during testing though, because the `ArtifactLogs` struct creates actual websocket connections.  And for testing purposes I tried to make the implementation of `BuildLog` independent of the connection implementation such that it can be faked easily...

pkg/platform/runtime/setup/buildlog/artifactlog.go
mdrohmann: Unfortunately, the `conn.ReadJSON` method does not support context timeouts.  So, I decided to just always exit without an error if `ReadJSON` returns.
It is also unfortunate, that the logs do not have a clear end-of-stream marker, because I would feel much more comfortable returning from the for-loop on an actual message that we receive.

pkg/platform/runtime/setup/buildlog/artifactlogs.go
mdrohmann: This has become obsolete due to the change above.  The `Wait()` method does not return an error anymore.

pkg/platform/runtime/setup/setup.go
Naatan: What I usually do in cases like this is use a separate `NewButMoreCustom()` function that is used from the regular `New()` call as well as from tests.

internal/runbits/buildlogfile/buildlogfile.go
mdrohmann: I actually put it in the `producer` because it is producing events after downloading the artifact log.  That seems much cleaner to me.

internal/runbits/buildlogfile/buildlogfile.go
mdrohmann: I added a human-readable timestamp.  The `CreateTemp` function adds a unique string (that is timestamp based) to the file anyways. If our only concern is uniqueness, it is probably better to *just* rely on that mechanism, as in the current commit.

pkg/platform/runtime/setup/setup.go
mdrohmann: Here is how that would look like: 5053602a4  I must admit that I am not a big fan of this, as then suddenly the `BuildLog` owns the connections (and has to close them too).  IMO it was cleaner before, even though it results in more code/effort for the caller of this constructor.  But it makes the boundaries of `BuildLog`'s implementation more confined (maybe even more "unixy").

Naatan (approved):

mdrohmann:
The issue with the build-log-streamer got resolved, such that I could remove the work-around that I initially created (ie., files `artifactlog.go` and `artifactlogs.go`).  Details are here: https://www.pivotaltracker.com/story/show/178901762/comments/225625041

Naatan (approved):

PR #1449, opened 2021-07-13
State Install without project - Review

Naatan:
https://www.pivotaltracker.com/story/show/178223554

### DIFF
--- a/cmd/state/internal/cmdtree/cmdtree.go
+++ b/cmd/state/internal/cmdtree/cmdtree.go
@@ -1,17 +1,13 @@
 package cmdtree
 
 import (
-	"github.com/ActiveState/cli/cmd/state/internal/cmdtree/intercepts/cmdcall"
 	"github.com/ActiveState/cli/internal/captain"
 	"github.com/ActiveState/cli/internal/condition"
-	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runners/state"
 	secretsapi "github.com/ActiveState/cli/pkg/platform/api/secrets"
-	"github.com/ActiveState/cli/pkg/project"
 )
 
 // CmdTree manages a tree of captain.Command instances.
@@ -270,10 +266,6 @@ func newStateCommand(globals *globalOptions, prime *primer.Values) *captain.Comm
 		},
 	)
 
-	cmdCall := cmdcall.New(prime)
-
-	cmd.SetInterceptChain(cmdCall.InterceptExec, interceptAddHeadlessNotify(prime.Output(), prime.Project()))
-
 	return cmd
 }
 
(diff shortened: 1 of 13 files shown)


cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -319,15 +311,3 @@ func (a *addCmdAs) deprecatedAlias(aliased *captain.Command, name string) {
 	a.parent.AddChildren(cmd)
 }
 
-func interceptAddHeadlessNotify(out output.Outputer, pj *project.Project) captain.InterceptFunc {
```
Naatan, line 322: Not needed anymore due to the local env management changes we're making.

internal/runners/packages/install.go
```diff
@@ -16,26 +12,16 @@ type InstallRunParams struct {
 
 // Install manages the installing execution context.
 type Install struct {
-	cfg  configurable
-	out  output.Outputer
-	proj *project.Project
-	prompt.Prompter
-	auth *authentication.Auth
+	prime primeable
```
Naatan, line 15: We should normally avoid doing this, but in this case passing `prime` along to the package operation saves us a lot of useless boilerplate.

internal/runners/packages/packages.go
```diff
@@ -43,23 +42,22 @@ type configurable interface {
...
-			return locale.WrapError(err, "err_install_get_langauge", "Could not get language for package: {{.V0}}", packageName)
+			return locale.WrapError(err, "err_package_get_project", "Could not get project from path")
 		}
+		defer func() {
+			if rerr != nil {
+				if err := os.Remove(pj.Source().Path()); err != nil {
+					logging.Error("could not remove temporary project file: %s", errs.JoinMessage(err))
+				}
+			}
+		}()
```
Naatan, line 60: This ensure we remove the as.yaml we created if the operation failed.

pkg/platform/model/inventory.go
```diff
@@ -83,32 +93,56 @@ func FetchAuthors(ingredID, ingredVersionID *strfmt.UUID) (Authors, error) {
 	return results.Payload.Authors, nil
 }
 
-func searchIngredientsNamespace(limit int, ns Namespace, name string) ([]*IngredientAndVersion, error) {
-	lim := int64(limit)
+func searchIngredientsNamespace(ns Namespace, name string, includeVersions bool) ([]*IngredientAndVersion, error) {
```
Naatan, line 96: I modified this function to make use of pagination as well as to add a param allowing you to include version variants of the ingredients or just the base ingredient itself.

pkg/platform/model/recipe.go
```diff
@@ -193,6 +174,24 @@ func FetchRecipe(commitID strfmt.UUID, owner, project string, hostPlatform *stri
 	return nil, locale.NewInputError("err_recipe_not_found")
 }
 
+func resolveSolverError(err error) error {
```
Naatan, line 177: Proper solver errors

internal/runners/packages/packages.go
```diff
@@ -154,22 +161,80 @@ func executePackageOperation(pj *project.Project, cfg configurable, out output.O
...
+		}
+		return "", ns, locale.WrapError(err, "err_pkgop_search",
+			"Could not match {{.V0}}. Did you mean:\n\n{{.V1}}", packageName, strings.Join(suggestions, "\n"))
+	}
+
+	//
+	choices := []string{}
+	values := map[string][]string{}
+	for _, i := range ingredients {
+		language := model.LanguageFromNamespace(*i.Ingredient.PrimaryNamespace)
```
MDrakos, line 190: While it's unlikely that this value is the empty string, not checking it here can lead to some odd display for the user and invalid namespaces. We could easily check if it's empty and then log and continue.

internal/runners/packages/packages.go
```diff
@@ -154,22 +161,80 @@ func executePackageOperation(pj *project.Project, cfg configurable, out output.O
...
+		// Generate ingredient choices to present to the user
+		name := fmt.Sprintf("%s (%s)", *i.Ingredient.Name, language)
+		choices = append(choices, name)
+		values[name] = []string{*i.Ingredient.Name, language}
+	}
+
+	// Prompt the user with the ingredient choices
+	choice, err := prompt.Select(
+		locale.Tl("prompt_pkgop_ingredient", "Multiple Matches"),
+		locale.Tl("prompt_pkgop_ingredient_msg", "Multiple ingredients matched your query, which one would you like to use?"),
```
MDrakos, line 207: Is `ingredient` a term we expect the user to understand? I'm not sure if we're exposing them to our kitchen metaphors anywhere else.

MDrakos (approved):

internal/runners/packages/packages.go
Naatan: This won't ever be nil, that's a swagger oddity. But it is also guarded against via the `SearchIngredientsStrict` function.

PR #1450, opened 2021-07-14
State revert does not modify remote project

MDrakos:
https://www.pivotaltracker.com/story/show/178776011


MDrakos:
When fixing the commit data not displaying I tweaked the UX as it wasn't very clear before

image

Naatan (approved):
Looking good!

PR #1451, opened 2021-07-15
Refactored and fixed state push and associated logic

Naatan:


internal/runners/push/push.go
```diff
@@ -42,146 +43,189 @@ type primeable interface {
...
+				locale.Tl("err_tip_push_outdated", "Run `[ACTIONABLE]state pull[/RESET]`"))
+		}
+	}
+
+	// Update the project at the given commit id.
+	err = model.UpdateProjectBranchCommitWithModel(targetPjm, branch.Label, commitID)
 	if err != nil {
-		return locale.WrapInputError(err, "push_remove_lang_err", "Failed to remove temporary language field from activestate.yaml.")
+		if errs.Matches(err, &model.ErrUpdateBranchAuth{}) {
+			return locale.WrapInputError(err, "push_project_branch_no_permission", "You do not have permission to push to {{.V0}}.", targetPjm.Name)
```
MDrakos, line 210: Does `targetPjm.Name` include the organization name? If not we may want to include it in the errors here.

internal/runners/push/push.go
```diff
@@ -197,31 +241,39 @@ func (r *Push) Run(params PushParams) error {
 	return nil
 }
 
-func (r *Push) getNamespace() (*project.Namespaced, error) {
+func (r *Push) verifyInput() error {
```
MDrakos, line 244: nit: this isn't really verifying any input as auth is a state and the project is sourced from the project file. That said I don't immediately have a better alternative name.

internal/runners/push/push.go
```diff
@@ -232,29 +284,25 @@ func (r *Push) getNamespace() (*project.Namespaced, error) {
 	return ns, nil
 }
 
-func (r *Push) languageForProject(pj *project.Project) (*language.Supported, string, error) {
-	if pj.CommitID() != "" {
-		return fetchLanguage(pj.CommitUUID())
+func (r *Push) promptNamespace() (*project.Namespaced, error) {
+	owner := authentication.LegacyGet().WhoAmI()
```
MDrakos, line 288: Can't we use `r.auth` here?

locale/en-us.yaml
```diff
@@ -1219,8 +1219,15 @@ init_description:
   other: Initialize a new project
 push_description:
   other: Push your latest changes to the platform
+push_first_new_project:
+  other: "Since this is your first time running \"push\" you'll need to create a new project on the ActiveState Platform."
```
MDrakos, line 1223:
```suggestion
  other: "Since this is your first time running \"push\" on this project you'll need to create a new project on the ActiveState Platform."
```

MDrakos (approved):
Looks good, consider my comments nits.

The push runner is easier to follow now though there is still quite a bit going on there.

internal/runners/push/push.go
```diff
@@ -42,146 +43,189 @@ type primeable interface {
...
+type intention uint16
+
+const (
+	pushCustomNamespace  intention = 0x0001 // User is pushing to a custom remote, ignoring the namespace in the current yaml
+	pushFromNoPermission           = 0x0002 // User made modifications to someone elses project, and it now trying to push them
+	pushFromHeadless               = 0x0004 // User is operating in headless mode and is now trying to push
+
+	// The rest is supplemental
+	intendCreateProject = 0x0008
+)
```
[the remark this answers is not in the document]
Naatan: You can, but this feels more effective. Originally I was expecting more values, but I don't think it hurts to commit it this way.

internal/runners/push/push.go
Naatan: Yeah it's a little bit of a stretch, but close enough. I'd still call the user being authenticated or not "input".

locale/en-us.yaml
Naatan: I think this is more confusing. We're talking about "This project" and then asking them to create a new one. I get what you mean, but I don't think this clarifies anything.

PR #1452, opened 2021-07-15
Ensure unique logfiles

MDrakos:
I was not able to consistently reproduce the bug on the linked story using the `beta` branch. I did not see the bug at all with this change. 

Regarding the PID reuse comment on the story, I think that it's very unlikely to happen on a Unix-based system but Windows might be a problem. The Unix-based systems do not reuse the PID until a wrap-around occurs after hitting `pid_max`. On Windows, a PID is available for reuse as soon as the last handle on that process has been closed however the timestamp should fix that.

 https://www.pivotaltracker.com/story/show/178851349


MDrakos:
Had to resort to using an exsiting global in order for the `FileName()` function to return the log file that the current process is using. 
The `FileName()` function is used on uninstall and when we capture panics.

We could file a follow-up story to replace the global with another mechanism or address it here.

internal/logging/defaults.go
```diff
@@ -24,7 +25,9 @@ import (
 // datadir is the base directory at which the log is saved
 var datadir string
 
-var datafile string
+var dataFileName string
+
+var dataFilePath string
```
Naatan, line 30: I'm not seeing the necessity for this. It just creates more logic for us maintained to have to grok. And since all of our Filename functions are just concatenating variables together caching doesn't to really serve much of a purpose.

test/integration/vscode_int_test.go
```diff
@@ -91,19 +91,29 @@ func (suite *PushIntegrationTestSuite) TestOrganizations_VSCode() {
...
+		},
+		{
+			"ActiveState-CLI",
+			"ActiveState-CLI",
+			"Team Tier",
+			true,
+		},
+	}
+
+	expected, err := json.Marshal(orgs)
```
Naatan, line 116: This is fixed on master. Please merge master and revert this change as the one on master is more forward-compatible.

Naatan (changes requested):

internal/logging/defaults.go
MDrakos: The issue is that one of those variables is now a timestamp. The change stems from our use of the `FilePath` function in two places one during [uninstall](https://github.com/ActiveState/cli/blob/master/internal/runners/clean/run_lin_mac.go#L58) and the other when we capture [panics](https://github.com/ActiveState/cli/blob/master/internal/runbits/panics/panics.go#L20). Without caching something (the filename or the timestamp) those functions return a file that does not exist, which was surfaced by integration test failures.

Naatan: Ok, then can we just capture the timestamp in `init()` and only use that as a global? Basically we just want the timestamp to be constant.

Naatan (approved):

PR #1454, opened 2021-07-16
Add uniqid to rollbar/analytics

daved:
https://www.pivotaltracker.com/story/show/178526418


daved:
@Naatan I'm wondering if I developed a blind spot for the solution of copying in fileutils logic. I just couldn't find a simpler way to avoid the import cycle.

Naatan:
> @Naatan I'm wondering if I developed a blind spot for the solution of copying in fileutils logic. I just couldn't find a simpler way to avoid the import cycle.

Please elaborate, how does this PR affect fileutils? I feel like I'm missing some context.

internal/uniqid/uniqid.go
```diff
@@ -0,0 +1,202 @@
...
+)
+
+// DirLocation represents tokens to indicate where uniqid file should be
+// located.
+type DirLocation int
+
+// DirLocation enums.
+const (
+	InHome DirLocation = iota
+	InTmp
```
Naatan, line 22: I THINK this means "based on home dir" and "based on tmp dir". Would be good to have some comments here.

internal/uniqid/uniqid.go
```diff
@@ -0,0 +1,202 @@
...
+)
+
+const (
+	fileName = "uniqid"
+	asDir    = "activestate_dat"
```
Naatan, line 27:
```suggestion
	asDir    = "activestate/persist"
```

Naatan: I think this is a bit friendlier to the user. Still a single main dir, and the subdir somewhat communicates its intend.

Naatan (changes requested):
Very nice! Just one nitpicky change. Also a note to self that we should probably look into a `fileutils` variant that's guaranteed not to create recursive dependencies.

Naatan (approved):

internal/analytics/analytics.go
```diff
@@ -117,6 +119,7 @@ func (d *customDimensions) toMap() map[string]string {
 		"8":  d.installSource,
 		"9":  d.machineID,
 		"10": d.projectName,
+		"11": d.uniqID,
```
Naatan, line 122: Please merge master and ensure you use `12`. I'll set up analytics accordingly.

Naatan (changes requested):

Naatan (approved):

internal/analytics/analytics.go
```diff
@@ -127,7 +128,11 @@ func (d *customDimensions) toMap() map[string]string {
 		"9":  d.machineID,
 		"10": d.projectName,
 		"11": d.sessionToken,
+		"12": d.uniqID,
 	}
+	logging.Debug("Custom dimensions analytics map formed as: %v", m)
```
Naatan, line 133: I was suggesting this for you to test things locally. Actually logging this is going to be too verbose as it would happen for every single analytics event.

Naatan (changes requested):

Naatan (approved):

PR #1458, opened 2021-07-20
Add solver error to logfile

mdrohmann:


pkg/platform/runtime/setup/setup.go
```diff
@@ -568,3 +574,25 @@ func reusableArtifacts(requestedArtifacts []*headchef_models.V1Artifact, storedA
...
+	errorLines := strings.Join(serr.ValidationErrors()[offset:], "\n")
+	isCropped := offset > 0
+	croppedMessage := ""
+	if isCropped {
+		croppedMessage = locale.Tl("solver_err_cropped_intro", "These are the last five lines of the error message:")
+	}
+
+	err = locale.WrapError(err, "solver_err", "The Platform failed to resolve the dependencies for this build. {{.V0}}\n{{.V1}}", croppedMessage, errorLines)
+	if serr.IsTransient() {
+		err = errs.AddTips(serr, locale.Tl("transient_solver_tip", "You may want to retry this command, as it can lead to a different result."))
```
MDrakos, line 595: Nit: this and the transient error message that we append to the build log could be a locale key that we reuse.

MDrakos (dismissed):

MDrakos (approved):

test/integration/push_int_test.go
```diff
@@ -116,6 +116,7 @@ func (suite *PushIntegrationTestSuite) TestInitAndPush() {
 // Test pushing to a new project from a headless commit
 func (suite *PushIntegrationTestSuite) TestPush_HeadlessConvert_NewProject() {
 	suite.OnlyRunForTags(tagsuite.Push)
+	suite.T().Skip("Skipping mac for now, as alternative Python builds are not available yet.")
```
mdrohmann, line 119: No idea why this has not been a problem in other PRs, or has it?

internal/runners/packages/packages.go
```diff
@@ -105,7 +104,7 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
 
 	// Verify that the provided package actually exists (the vcs API doesn't care)
 	_, err = model.FetchRecipe(commitID, pj.Owner(), pj.Name(), &model.HostPlatform)
-	if err != nil && !inventory.IsPlatformError(err) {
+	if err != nil && !model.IsPlatformError(err) {
```
mdrohmann, line 107: I am a little confused that we are ignoring platform errors here.  That basically means that we allow to add anything to the order, if the resulting solver error is that this package is not supported by the platform.

Naatan: It's ignoring platform errors as in the platform of the build (ie. the OS). Not platform as in "ActiveState Platform".

locale/en-us.yaml
```diff
@@ -1701,3 +1701,7 @@ operation_success_local:
 err_initialize_runtime_event_handler:
   other: |
     Could not determine how to display progress of runtime installation.
+solver_err:
+  other: The Platform failed to resolve the dependencies for this build. {{.V0}}\n{{.V1}}
+transient_solver_tip:
+  other: You may want to retry this command, as it can lead to a different result.
```
Naatan, line 1707:
```suggestion
  other: You may want to retry this command, as it can lead to a different result. 
```

This goes against the promise of the platform. Why would retrying it give a different result? Same request should give the same result, we should not have error messages that indicate otherwise.

pkg/platform/runtime/setup/setup.go
```diff
@@ -568,3 +574,25 @@ func reusableArtifacts(requestedArtifacts []*headchef_models.V1Artifact, storedA
 	}
 	return keep
 }
+
+func formatSolverError(serr *apimodel.SolverError) error {
```
Naatan, line 578: I think we should limit this based on length, not lines. 5 lines can still contain a TON of data (you've seen our solver errors).

Naatan (changes requested):

pkg/platform/runtime/setup/setup.go
mdrohmann: Yes, that makes sense.  There are some newline characters in the output, but the lines can get very large.  I think I'll trim it at 500 characters which woud be 5 lines on a 100 character wide display.

internal/runners/packages/packages.go
mdrohmann: > It's ignoring platform errors as in the platform of the build (ie. the OS). Not platform as in "ActiveState Platform".

I understand.  So, the idea is that we can add all kinds of ingredients to a MacOS platform, even if we do not know how to build it.  But I am still surprised, and wonder if this is really what we want to do here.

locale/en-us.yaml
mdrohmann: The solver has this error:  https://github.com/ActiveState/cli/blob/cd0ba72d833dcf6e3fafbb858c31144afc8b4fa8/pkg/platform/api/inventory/inventory_models/solver_error.go#L24-L25   I think it only happens when there is a resource problem that interrupted the solve.

Will check with PB.

mdrohmann: It's whenever a resource failure occurred during the solve: https://activestate.slack.com/archives/CCELJM8GL/p1627406491195600

Should hardly ever be the case...

Naatan:
```suggestion
  other: You may want to retry this command if the failure was related to a resourcing or networking issue.
```

I'm not a big fan of this still, but I'm especially not a fan of "it may lead to different results". Technically that is accurate, but I don't think we should be setting expectations where retrying magically fixes things.

Naatan (approved):

Naatan (changes requested):
Meant to request changes

MDrakos (dismissed):

PR #1459, opened 2021-07-20
Fix history UI wrapping too aggressively

Naatan:


internal/output/plain.go
```diff
@@ -100,8 +99,16 @@ func (f *Plain) writeNow(writer io.Writer, value string) {
...
+	var result []string
+	cropped := colorize.GetCroppedText(text, termutils.GetWidth())
+	for _, crop := range cropped {
+		result = append(result, crop.Line)
+	}
+	suffix := ""
+	if strings.HasSuffix(text, "\n") {
+		suffix = "\n"
+	}
+	return strings.Join(result, "\n") + suffix
```
daved, line 111: Just making sure there should be double newlines in some cases.

daved (approved):

internal/output/plain.go
```diff
@@ -105,8 +105,14 @@ func wordWrap(text string) string {
...
-	if strings.HasSuffix(text, "\n") {
-		suffix = "\n"
+
+	// Check for line endings at the end of the string
+	// We don't care about runes here since we are only looking at line endings
+	for x := len(text) - 1; x >= 0; x-- {
+		if string(text[x]) != "\n" {
+			break
+		}
+		suffix += "\n"
```
daved, line 115: Another verification; Should this treatment exist for each line?

daved (approved):

internal/output/plain.go
Naatan: I don't understand the question?

If the input had 5 line endings at the end of the string then the output should have 5 line endings at the end of the string.

PR #1462, opened 2021-07-21
Add session token

Naatan:
Interim release version: https://github.com/ActiveState/cli/pull/1463

### DIFF
--- a/cmd/state-installer/installer.go
+++ b/cmd/state-installer/installer.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -8,6 +9,7 @@ import (
 	"time"
 
 	"github.com/ActiveState/cli/cmd/state-installer/internal/installer"
+	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/appinfo"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
@@ -28,6 +30,10 @@ import (
 	"github.com/thoas/go-funk"
 )
 
+type Flags struct {
+	sessionToken string
+}
+
 func main() {
 	var exitCode int
 	defer func() {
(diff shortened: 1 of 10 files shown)


mdrohmann (changes requested):
It looks good, but the transitional State Tool should forward the session id to the installer.  I think that needs to happen here:
https://github.com/ActiveState/cli/blob/master/cmd/state-transition-update/main.go#L82

Naatan:
> It looks good, but the transitional State Tool should forward the session id to the installer. I think that needs to happen here:
> https://github.com/ActiveState/cli/blob/master/cmd/state-transition-update/main.go#L82

I didn't think that would matter since it would only ever get invoked through the auto updater, which won't ever be fed this token. Am I missing something?

mdrohmann:
> > It looks good, but the transitional State Tool should forward the session id to the installer. I think that needs to happen here:
> > https://github.com/ActiveState/cli/blob/master/cmd/state-transition-update/main.go#L82
> 
> I didn't think that would matter since it would only ever get invoked through the auto updater, which won't ever be fed this token. Am I missing something?

I think you are correct. The only exception would be that some-one is re-using the old install.sh / install.ps1 which (at some point) is targeting a transitional State Tool.  But who would do that?

mdrohmann (approved):

Naatan:
> > > It looks good, but the transitional State Tool should forward the session id to the installer. I think that needs to happen here:
> > > https://github.com/ActiveState/cli/blob/master/cmd/state-transition-update/main.go#L82
> > 
> > 
> > I didn't think that would matter since it would only ever get invoked through the auto updater, which won't ever be fed this token. Am I missing something?
> 
> I think you are correct. The only exception would be that some-one is re-using the old install.sh / install.ps1 which (at some point) is targeting a transitional State Tool. But who would do that?

Yeah, and even if it did happen it wouldn't happen in a way that we'd be trying to inject a session token.

mdrohmann:
> Yeah, and even if it did happen it wouldn't happen in a way that we'd be trying to inject a session token.

It could! After the proposed changes on the release channel merge via this line: https://github.com/ActiveState/cli/blob/4629c5300b7ebcbd32dfd70dbfb0bcbc8e0f0e93/installers/install.sh#L425

Naatan:
> > Yeah, and even if it did happen it wouldn't happen in a way that we'd be trying to inject a session token.
> 
> It could! After the proposed changes on the release channel merge via this line:
> 
> https://github.com/ActiveState/cli/blob/4629c5300b7ebcbd32dfd70dbfb0bcbc8e0f0e93/installers/install.sh#L425

D'oh, good eye! Will do a followup PR.

PR #1463, opened 2021-07-21
Add session token (current release channel version)

Naatan:
Beta version: https://github.com/ActiveState/cli/pull/1462


MDrakos (approved):
There might be a case to make the token an optional argument rather than a flag. While the `-v` flag isn't present on these install scripts it is on the `beta` channel. If a user tries to install an older version of the state tool that does not have the `--session-token` flag they will encounter an error when we get to the prepare step.

mdrohmann (approved):

daved (approved):

mdrohmann (approved):

PR #1464, opened 2021-07-21
Update cache path on Windows

MDrakos:
Changing the cache path will mean that current beta channel users on Windows will end up re-installing their local runtimes. Also, the list of runtimes in the system tray will be affected. We may want to warn these users and give them directions on how to remove the old cache directories that live at `C:/<user>/AppData/Local/activestate`

https://www.pivotaltracker.com/story/show/178591213


internal/installation/storage/storage.go
```diff
@@ -106,6 +106,9 @@ func CachePath() string {
 		}
 	} else if path := os.Getenv(constants.CacheEnvVarName); path != "" {
 		cachePath = path
+	} else if runtime.GOOS == "windows" {
+		// <UserDir>/AppData/Local/activestate/cache
+		cachePath = filepath.Join(configdir.New(constants.InternalConfigNamespace, "").QueryCacheFolder().Path, "cache")
 	} else {
 		cachePath = configdir.New(constants.InternalConfigNamespace, "").QueryCacheFolder().Path
```
Naatan, line 113:
```suggestion
	} else {
		cachePath = configdir.New(constants.InternalConfigNamespace, "").QueryCacheFolder().Path
		if runtime.GOOS == "windows" {
		    // Explicitly append "cache" dir as the cachedir on Windows is the same as the local appdata dir (conflicts with config)
		    cachePath = filepath.Join(cachePath, "cache")
		}
```

Naatan (changes requested):

Naatan (approved):

PR #1467, opened 2021-07-22
Improve performance of state tool

Naatan:
Each commit addresses a different issue, so it might be easier to review this commit by commit.


cmd/state-tray/main.go
```diff
@@ -162,7 +162,7 @@ func run() (rerr error) {
 	mReload := mProjects.AddSubMenuItem("Reload", "Reload the local projects listing")
 	localProjectsUpdater := menu.NewLocalProjectsUpdater(mProjects)
 
-	localProjects, err := model.LocalProjects()
+	localProjects, err := model.LocalProjects(context.Background())
```
mdrohmann, line 165: It looks like you use `context.Background()` everywhere.  Is that correct?  I assume that you have experimented with time-out contexts but determined that they are not necessary anymore.

I am fine if we leave this commit (814c6ed55ebe59d5) as-is anyways as it gives us the flexibility to introduce timeouts here.  But I would like to ensure that I understand it correctly that it does not change any behavior right now.

pkg/platform/api/svc/svc.go
```diff
@@ -25,23 +25,11 @@ func New(cfg configurable) (*Client, error) {
...
-func NewWithoutRetry(cfg configurable) (*Client, error) {
-	port := cfg.GetInt(constants.SvcConfigPort)
-	if port <= 0 {
-		return nil, locale.NewError("err_svc_no_port", "The State Tool service does not appear to be running (no local port was configured).")
-	}
 
 	baseUrl := fmt.Sprintf("http://127.0.0.1:%d", port)
 	return &Client{
-		Client:  gqlclient.NewWithOpts(fmt.Sprintf("%s/query", baseUrl), 0, graphql.WithHTTPClient(http.DefaultClient)),
+		// The custom client bypasses http-retry, which we don't need for doing local requests
```
mdrohmann, line 31: :+1:

test/integration/svc_int_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"golang.org/x/net/context"
+)
+
+func TestSvcManagerAndModel(t *testing.T) {
+	cfg, err := config.New()
+	require.NoError(t, err)
+	svcManager := svcmanager.New(cfg)
+	require.NoError(t, svcManager.Start())
+
+	ctx, _ := context.WithTimeout(context.Background(), 5*time.Second)
```
mdrohmann, line 20: Ah, I see.  It is used for testing. :)

internal/profile/profile.go
```diff
@@ -12,6 +12,11 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 )
 
+var profilingEnabled = false
+
+func init() {
+	profilingEnabled = os.Getenv(constants.ProfileEnvVarName) == "true"
```
mdrohmann, line 18: nit: We do not need an `init()` function for this initialization.

mdrohmann (dismissed):
LGTM

cmd/state-tray/main.go
Naatan: This is just preserving the behavior that was there prior to this PR. I just made it more explicit.

Some of these could certainly benefit from a timeout, but they were not the focus of this PR.

internal/profile/profile.go
Naatan: I could swear Go didn't let you call functions outside of a function.. not sure why this works.

test/integration/svc_int_test.go
Naatan: Yeah we're not really using this particular context atm, the refactor was mainly so that I could use the context for `NewSvcModel()` appropriately. In the previous version we were reusing it for all client requests which will almost certainly eventually trigger a timeout in a way that it shouldn't.

PR #1478, opened 2021-07-27
Commit language as part of the order

Naatan:


internal/constants/constants.go
```diff
@@ -84,8 +84,17 @@ const ProfileEnvVarName = "ACTIVESTATE_PROFILE"
...
-// NonInteractive is the name of the environment variable that specifies whether to run the State Tool without prompts
-const NonInteractive = "ACTIVESTATE_NONINTERACTIVE"
+// NonInteractiveEnvVarName is the name of the environment variable that specifies whether to run the State Tool without prompts
+const NonInteractiveEnvVarName = "ACTIVESTATE_NONINTERACTIVE"
+
+// OverwriteDefaultInstallationPathEnvVarName is the environment variable name to overwrite the default installation path FOR TESTING PURPOSES ONLY
+const OverwriteDefaultInstallationPathEnvVarName = "ACTIVESTATE_TEST_INSTALL_PATH"
+
+// OverwriteDefaultSystemPathEnvVarName is the environment variable name to overwrite the system app installation directory updates FOR TESTING PURPOSES ONLY
+const OverwriteDefaultSystemPathEnvVarName = "ACTIVESTATE_TEST_SYSTEM_PATH"
```
Naatan, line 94: Moved these and renamed for consistency.

cmd/state/main.go
```diff
@@ -85,7 +85,7 @@ func main() {
 	// Set up our legacy outputer
 	setPrinterColors(outFlags)
 
-	isInteractive := strings.ToLower(os.Getenv(constants.NonInteractive)) != "true" &&
+	isInteractive := strings.ToLower(os.Getenv(constants.NonInteractiveEnvVarName)) != "true" &&
```
Naatan, line 88: This is so I can test on macOS. Not intended for public use but don't care if someone figures out they can use this.

mdrohmann (approved):
Looks good.  But so, we are basically bound to have a single language per project still then...?

Naatan:
> Looks good. But so, we are basically bound to have a single language per project still then...?

This isn't itself a limit on number of languages. But it does basically mean you can't have a project -without- a language. Long term this will be addressed.

PR #1482, opened 2021-07-27
Use exact only flag in ingredient search

mdrohmann:


Naatan (approved):

test/integration/push_int_test.go
```diff
@@ -124,6 +124,12 @@ func (suite *PushIntegrationTestSuite) TestPush_HeadlessConvert_NewProject() {
 	namespace := fmt.Sprintf("%s/%s", username, pname)
 
 	cp := ts.SpawnWithOpts(e2e.WithArgs("install", suite.extraPackage))
+
+	cp.Expect("Multiple Matches")
+	cp.Expect(fmt.Sprintf("> %s", pname))
+	// pick the current default selection
+	cp.Send("")
```
mdrohmann, line 131: It turns out, that without the `exactTerm` search, the results were sorted in a strange way (probably by namespace first), such that in our paginated search duplicate entries for multiple languages were possibly missed. (Cf., https://github.com/ActiveState/cli/runs/3176554252#step:17:220)

So, now we actually find out that we have Perl and Ruby ingredients `JSON` and `json`, and Python and Ruby ingredients for `six` (on Mac).

I believe that this is now the correct behavior but I filed a small follow-up story so we can discuss the behavior in Stand: https://www.pivotaltracker.com/story/show/179043262

Naatan (approved):

PR #1483, opened 2021-07-27
Update MSI release version

MDrakos:
https://www.pivotaltracker.com/story/show/178965523

The version number is from the latest run on `release` here: https://github.com/ActiveState/cli/actions?query=branch%3Arelease


internal/runners/initialize/init.go
```diff
@@ -166,7 +166,7 @@ func run(params *RunParams, out output.Outputer) (string, error) {
 	if version == "" {
 		version = params.language.RecommendedVersion()
 	}
-	commitID, err := model.CommitInitial(model.HostPlatform, params.language.String(), version)
+	commitID, err := model.CommitInitial(model.HostPlatform, params.language.Requirement(), version)
```
Naatan, line 169: This doesn't seem related to your PR? Was something broken? `.String()` is what I wanted to use, as `.Requirement()` uses either `python2` or `python3` whereas we want `python`.

Naatan (changes requested):

internal/runners/initialize/init.go
MDrakos: It's the other way around, which is confusing. `.String()` gets you `python2` or `python3` and `.Requirement()` gets you `python`. 

It's not related to my PR, it was just an easy fix for an integration test. I can revert this and file a follow-up story if we want to address the naming of the methods.

 This was the error output from the test:

```
╔════════════════════╗                                                          
║ Installing Package ║                                                          
╚════════════════════╝                                                          
                                                                                        
Something Went Wrong                                                            
────────────────────                                                            
x Could not match six                                                          
x Your order refers to 1 non-existent namespace: language/python3    
```

Run here: https://github.com/ActiveState/cli/runs/3176631820

Naatan (approved):

PR #1487, opened 2021-07-30
`state install` filter packages for supported languages

mdrohmann:


pkg/platform/model/supportedlangs.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/pkg/platform/api/mediator"
+	"github.com/ActiveState/cli/pkg/platform/api/mediator/model"
+	"github.com/ActiveState/cli/pkg/platform/api/mediator/request"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+)
+
+// FetchSupportedLanguages returns the list of languages that the Platform supports ATM
+func FetchSupportedLanguages(auth *authentication.Auth) ([]string, error) {
```
Naatan, line 12: Is there a technical reason for requiring auth? I can't think of any reason why we should need auth for this.

Naatan (changes requested):
Looks good to me except for needlessly passing auth. Needlessly exposes us to bugs.

pkg/platform/model/supportedlangs.go
mdrohmann: Just future compatibility.  I removed it for now.

Naatan (approved):

PR #1489, opened 2021-08-03
Enable history output without requiring push

daved:
https://www.pivotaltracker.com/story/show/179074445

### DIFF
--- a/internal/runners/history/history.go
+++ b/internal/runners/history/history.go
@@ -59,17 +59,19 @@ func (h *History) Run(params *HistoryParams) error {
 			return locale.NewInputError("err_history_no_project", "A namespace was not provided and a project could not be found. Please use a project namespace or run this command in a project directory")
 		}
 
-		remoteBranch, err := model.BranchForProjectNameByName(h.project.Owner(), h.project.Name(), h.project.BranchName())
-		if err != nil {
-			return locale.WrapError(err, "err_history_remote_branch", "Could not get branch by local branch name")
-		}
-
-		remoteCommitID := remoteBranch.CommitID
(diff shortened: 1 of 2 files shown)


Naatan (changes requested):
This seems a much larger logical change than necessary to address the bug. Remember we're at the polish stage; our changes should be as small and non-disruptive as possible.

daved:
@Naatan I've swapped the meaning of "nil latest remote" to mean "all local", then explicitly set the "latest remote id" to the initial commit id so that all show as remote. This looks like a bigger change than it is because of indenting.

internal/runners/history/history.go
```diff
@@ -54,22 +54,27 @@ func (h *History) Run(params *HistoryParams) error {
...
-			return locale.WrapError(err, "err_history_remote_branch", "Could not get branch by local branch name")
-		}
-
-		remoteCommitID := remoteBranch.CommitID
 		localCommitID := h.project.CommitUUID()
 
-		lastRemoteID, err = model.CommonParent(remoteCommitID, &localCommitID)
-		if err != nil {
-			return locale.WrapError(err, "err_history_common_parent", "Could not determine common parent commit")
+		if !h.project.IsHeadless() && h.project.Name() != "" {
```
Naatan, line 68: Why have `h.project.Name() != ""`? The `IsHeadless()` should already cover that.

Naatan (changes requested):

internal/runners/history/history.go
daved: From when I misunderstood how specifically "headless" was defined... fixed

Naatan (approved):

PR #1490, opened 2021-08-03
Stop prompting for name when signing up, as it can conflict and isn't actually required

Naatan:
https://www.pivotaltracker.com/story/show/179075171


daved (approved):
What about just not making it required?

Naatan:
> What about just not making it required?

That's the API's job. I've reported this issue to dashboard.

PR #1497, opened 2021-08-04
Uninstall removes start menu item

MDrakos:
https://www.pivotaltracker.com/story/show/179075797

### DIFF
--- a/internal/runners/prepare/prepare_windows.go
+++ b/internal/runners/prepare/prepare_windows.go
@@ -17,6 +17,8 @@ import (
 	"github.com/ActiveState/cli/internal/osutils/shortcut"
 )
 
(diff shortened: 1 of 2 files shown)


internal/runners/prepare/prepare_windows.go
```diff
@@ -51,6 +48,10 @@ func (r *Prepare) prepareStartShortcut() error {
 	return nil
 }
 
+func shortcutDir() string {
+	return filepath.Join(os.Getenv("USERPROFILE"), "AppData", "Roaming", "Microsoft", "Windows", "Start Menu", "Programs", "ActiveState")
+}
+
```
Naatan, line 54: This should just be a constant or a variable.

Naatan (changes requested):

Naatan (approved):

PR #1500, opened 2021-08-04
Add instruction on how to use runtime

Naatan:


MDrakos (approved):
I wonder if we want to make this even more prominent, rather than just tacking it on to the end of all of the `state install` output. But I imagine that could be determined with another pass over the UX

PR #1502, opened 2021-08-05
Remove environment PATH entries

mdrohmann:


internal/runners/clean/run_win.go
```diff
@@ -22,24 +22,36 @@ import (
 )
 
 func (u *Uninstall) runUninstall() error {
+	// we aggregate installation errors, such that we can display all installation problems in the end
```
mdrohmann, line 25: I know this change is not in scope for this PR. I realized that the behavior was inconsistent between Mac/Linux and Windows.  I am pretty sure that we want it this way, but I am happy to take it out and create a separate story for this.

internal/runners/clean/run.go
```diff
@@ -31,3 +35,29 @@ func undoPrepare(cfg configurable) error {
...
+	if err != nil {
+		return errs.Wrap(err, "Could not determine if running as Windows administrator")
+	}
+
+	// remove shell file additions
+	s := subshell.New(cfg)
+	if err := s.CleanUserEnv(cfg, sscommon.InstallID, isAdmin); err != nil {
+		return errs.Wrap(err, "Failed to State Tool installation PATH")
+	}
+	// Default and deployed projects will stop working, so we return them from the PATH as well
```
mdrohmann, line 50: I was not 100% sure whether we should also remove default and deploy directory PATHs here, but as the executors they contain will not work without the State Tool, IMO this should be the correct behavior.

Without removing the entries, the user may get errors like "Could not find executable `state`" when calling `python` or `perl` which I believe is very confusing to the average user.

MDrakos (approved):

Naatan (approved):

PR #1503, opened 2021-08-05
Improve CTA message in installer scripts

daved:
https://www.pivotaltracker.com/story/show/179072804

### DIFF
--- a/installers/install.sh
+++ b/installers/install.sh
@@ -352,7 +352,8 @@ if [ "`dirname \`which $STATEEXE\` 2>/dev/null`" = "$INSTALLDIR" ]; then
 fi
 
 
-info "State Tool installation complete."
+info "State Tool successfully installed."
+info "Reminder: Start a new shell in order to start using the State Tool."
 
 # Keep --activate and --activate-default flags for backwards compatibility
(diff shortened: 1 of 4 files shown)


installers/install.sh
```diff
@@ -342,17 +342,11 @@ esac
...
-# Check if the installation is in $PATH, if so we also check if the activate
-# flag was passed and attempt to activate the project
-if [ "`dirname \`which $STATEEXE\` 2>/dev/null`" = "$INSTALLDIR" ]; then
-  info "State Tool installation complete."
-fi
-
-
-info "State Tool installation complete."
+info "State Tool successfully installed."
+info "Reminder: Start a new shell in order to start using the State Tool."
```
Naatan, line 349: Can we highlight both of these like we do on Windows (we're doing this elsewhere in the sh)

Naatan (changes requested):

installers/install.sh
daved: The `info` function does make it bold already. Would you like a different style like `warn` or a yellow coloring like the Windows script?

Naatan (approved):

PR #1504, opened 2021-08-05
Remove old config entry when converting a headless project

MDrakos:
https://www.pivotaltracker.com/story/show/179092113

### DIFF
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -236,7 +236,7 @@ func (r *Push) Run(params PushParams) error {
 	// If this project was initially headless and we have sucessfully converted
 	// to a named project, we update the configuration project mapping
 	if intend&pushFromHeadless > 0 {
-		projectfile.CleanProjectPathMapping(r.config, filepath.Dir(r.project.Source().Path()))
+		projectfile.CleanProjectMapping(r.config, filepath.Dir(r.project.Source().Path()))
 		projectfile.StoreProjectMapping(r.config, targetNamespace.String(), filepath.Dir(r.project.Source().Path()))
 	}
 
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -235,7 +235,6 @@ func (r *Push) Run(params PushParams) error {
 
 	// We have sucessfully pushed, if the project was initially headless we update the config
 	if intend&pushFromHeadless > 0 {
-		projectfile.CleanProjectMapping(r.config, filepath.Dir(r.project.Source().Path()))
 		projectfile.StoreProjectMapping(r.config, targetNamespace.String(), filepath.Dir(r.project.Source().Path()))
 	}
 
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -2,6 +2,7 @@ package push
 
 import (
 	"errors"
+	"path/filepath"
 
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/logging"
@@ -230,6 +231,12 @@ func (r *Push) Run(params PushParams) error {
 		if err := r.project.Source().SetBranch(branch.Label); err != nil {
 			return errs.Wrap(err, "Could not set branch")
 		}
+	}
+
+	// If this project was initially headless and we have sucessfully converted
+	// to a named project, we update the configuration project mapping
+	if intend&pushFromHeadless > 0 {
+		projectfile.UpdateHeadlessConfigMapping(r.config, targetNamespace.String(), filepath.Dir(r.project.Source().Path()))
 	}
 
 	if projectCreated {
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -236,7 +236,8 @@ func (r *Push) Run(params PushParams) error {
 	// If this project was initially headless and we have sucessfully converted
 	// to a named project, we update the configuration project mapping
 	if intend&pushFromHeadless > 0 {
-		projectfile.UpdateHeadlessConfigMapping(r.config, targetNamespace.String(), filepath.Dir(r.project.Source().Path()))
+		projectfile.CleanProjectPathMapping(r.config, filepath.Dir(r.project.Source().Path()))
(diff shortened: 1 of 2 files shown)


pkg/projectfile/projectfile.go
```diff
@@ -1307,7 +1307,7 @@ func storeProjectMapping(cfg ConfigGetter, namespace, projectPath string) {
 
 // CleanProjectMapping removes projects that no longer exist
 // on a user's filesystem from the projects config entry
-func CleanProjectMapping(cfg ConfigGetter) {
+func CleanProjectMapping(cfg ConfigGetter, projectPaths ...string) {
```
Naatan, line 1310: This shouldn't need projectPaths as input. We're removing duplicate entries that already exist. So just iterate through the list and when encountering a duplicate entry only keep the latest one.

pkg/projectfile/projectfile.go
```diff
@@ -1273,9 +1273,9 @@ func GetProjectPaths(cfg ConfigGetter, namespace string) []string {
 	return paths
 }
 
-// storeProjectMapping associates the namespace with the project
+// StoreProjectMapping associates the namespace with the project
 // path in the config
-func storeProjectMapping(cfg ConfigGetter, namespace, projectPath string) {
+func StoreProjectMapping(cfg ConfigGetter, namespace, projectPath string) {
```
Naatan, line 1278: This doesn't seem to be necessary for this PR.

Naatan (changes requested):
Please clean up this PR so that it only updates what's really needed. We're at the feature freeze stage so these changesets should be as small as possible.

internal/runners/push/push.go
```diff
@@ -232,6 +233,12 @@ func (r *Push) Run(params PushParams) error {
 		}
 	}
 
+	// We have sucessfully pushed, if the project was initially headless we update the config
+	if intend&pushFromHeadless > 0 {
```
Naatan, line 237: This conditional isn't necessary. It doesn't hurt to clean the project mapping even when it's redundant.

pkg/projectfile/projectfile.go
MDrakos: There are no duplicate entries at this point so I'm passing the path to the headless project so we can remove it. The next line adds a project entry for the newly created project.

pkg/projectfile/projectfile.go
MDrakos: Push isn't storing a project entry for a new project. At least it wasn't happening during my testing. Only when the project was activated did the config change. So I exposed this function so push could update the config when converting a headless project. If we don't want that behaviour in push I can revert this.

pkg/projectfile/projectfile.go
MDrakos: The duplicate entries won't be in the same list, one is in the entry for the headless project and one is in the entry for the newly created project on `push`. That said there is a bug with this approach. I think we need a clean config function that takes a `namespace` and `path` argument so we can remove the path entry from the headless project specifically.

internal/runners/push/push.go
Naatan: This should still be addressed.

pkg/projectfile/projectfile.go
```diff
@@ -1286,6 +1286,17 @@ func storeProjectMapping(cfg ConfigGetter, namespace, projectPath string) {
 
 			projectPath = filepath.Clean(projectPath)
 
+			for name, paths := range projects {
+				if name == namespace {
+					continue
+				}
```
Naatan, line 1292: This shouldn't be necessary. Duplicate path entries are duplicate path entries.

pkg/projectfile/projectfile.go
```diff
@@ -1286,6 +1286,17 @@ func storeProjectMapping(cfg ConfigGetter, namespace, projectPath string) {
...
+				for i, path := range paths {
+					if path == projectPath {
```
Naatan, line 1294: We should pass these through `fileutils.ResolveUniquePath` (`path` here, `projectPath` we can replace the `filepath.Clean` on line 1287.

Naatan (changes requested):

pkg/projectfile/projectfile.go
MDrakos: Without this check, if a `namespace` and `projectPath` are already stored in the config and this function is called, it will first remove that path and then later store it again.

pkg/projectfile/projectfile.go
```diff
@@ -1284,16 +1284,29 @@ func StoreProjectMapping(cfg ConfigGetter, namespace, projectPath string) {
 				logging.Errorf("Projects data in config is abnormal (type: %T)", v)
 			}
 
-			projectPath = filepath.Clean(projectPath)
+			projectPath, err = fileutils.ResolveUniquePath(projectPath)
+			if err != nil {
+				logging.Errorf("Could not resolve uniqe project path, %v", err)
```
Naatan, line 1289: We should fall back on `filepath.Clean()` here.

pkg/projectfile/projectfile.go
```diff
@@ -1284,16 +1284,29 @@ func StoreProjectMapping(cfg ConfigGetter, namespace, projectPath string) {
...
 
 			for name, paths := range projects {
 				if name == namespace {
 					continue
 				}
+
 				for i, path := range paths {
+					path, err = fileutils.ResolveUniquePath(path)
+					if err != nil {
+						logging.Errorf("Could not resolve unique path, :%v", err)
```
Naatan, line 1300: Fall back here as well.

Naatan (changes requested):

pkg/projectfile/projectfile.go
Naatan: I don't see what's the problem with that. Either way you'd end up with the same result, the only difference is that removing this check ensures there's no duplicate path entries.

Naatan (changes requested):
Looks good aside from the namespace check part.

pkg/projectfile/projectfile.go
MDrakos: That's fair

Naatan (approved):

PR #1505, opened 2021-08-09
submit default language version on initial commit

mdrohmann:
https://www.pivotaltracker.com/story/show/179137942

### DIFF
--- a/internal/runners/packages/packages.go
+++ b/internal/runners/packages/packages.go
@@ -43,20 +43,11 @@ type configurable interface {
 
 const latestVersion = "latest"
 
-func supportedLanguageByName(supported []medmodel.SupportedLanguage, langName string) medmodel.SupportedLanguage {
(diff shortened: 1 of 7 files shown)


Naatan (changes requested):
This is making more logical changes than I think is necessary for the phase we're in. Especially considering the logic changes are in a crucial place (package operations).

Could you please move the logic of verifying the language here:

http://github.com/ActiveState/cli/blob/fb38d53559fb12416aefbaf185363d31caf4d166/internal/runners/packages/packages.go#L179-L179

The over-arching logic in the `executePackageOperation` function shouldn't need to be changed far as I can tell.

Naatan (approved):
Agree this isn't an ideal solution, but this perfectly addresses the story for the phase that we're at. Also totally agree a followup is warranted. Nice work!

Naatan (approved):

PR #1506, opened 2021-08-09
Changes defaults in state tool installers

rawktron:
Makes modifications to the install scripts to change the default behaviour to default to YES rather than NO and adds additional contextual information in the prompt.

### DIFF
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -63,7 +63,7 @@ function isAdmin
 
 function promptYN([string]$msg)
 {
-    $response = Read-Host -Prompt $msg" [Y/n]`y"
+    $response = Read-Host -Prompt $msg" [Y/n]`n"
 
     if ($response.ToLower() -eq "n")
     {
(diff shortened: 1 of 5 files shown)


installers/install.ps1
```diff
@@ -63,9 +63,9 @@ function isAdmin
 
 function promptYN([string]$msg)
 {
-    $response = Read-Host -Prompt $msg" [y/N]`n"
+    $response = Read-Host -Prompt $msg" [Y/n]`y"
```
Naatan, line 66:
```suggestion
    $response = Read-Host -Prompt $msg" [Y/n]`n"
```

I think that was a line ending

Naatan (changes requested):

installers/install.ps1
rawktron: Fixed.

rawktron: Hold on, that commit isn't showing up here.

rawktron: Ok fixed -- weird GitHub issue. I had to push a new commit to get the other one to show up here.

Naatan (approved):

PR #1507, opened 2021-08-11
Wrap git.PlainClone call for possible panic (wsarecv)

daved:
https://www.pivotaltracker.com/story/show/178726147


daved:
Follow-up to remove the logging to rollbar filed here: https://www.pivotaltracker.com/story/show/179187192

The more I think about this, I wonder if this can affect any http request made in Windows. If so, I think this logic is better placed in a/the http client. Alternately, it may simply be rare enough that handling this at all is not worthwhile.

Naatan (approved):

PR #1510, opened 2021-08-13
add tagging support to updater

mdrohmann:
https://www.pivotaltracker.com/story/show/179199122


installers/legacy-install.sh
```diff
@@ -3,6 +3,8 @@
 #
 # Usage: ./install.sh [-b branch]
 
+set -x
```
Naatan, line 6: I guess you were testing something, but we shouldn't land this.

installers/legacy-install.sh
```diff
@@ -125,10 +129,20 @@ x86_64)
...
+    TMPDIR=`mktemp -d`
+  else
+    TMPDIR="${TMPDIR:-/tmp}/state-installer.$$"
+    # clean-up previous temp dir
+    rm -rf $tdir
+    mkdir -p $TMPDIR
+  fi
 
+}
+set_tempdir
```
Naatan, line 143: This doesn't seem relevant to this PR. Prefer we don't touch anything we don't need to.

internal/updater/checker.go
```diff
@@ -29,28 +34,33 @@ type Checker struct {
 var DefaultChecker = newDefaultChecker()
 
 func newDefaultChecker() *Checker {
+	infoURL := constants.APIUpdateInfoURL
+	if url, ok := os.LookupEnv("_TEST_UPDATE_INFO_URL"); ok {
+		infoURL = url
+	}
```
Naatan, line 40: For integration tests we should just use the actual endpoint. For unit tests we shouldn't need env vars, the code structure should facilitate testing. eg. just invoke `NewChecker()` from your tests instead of using `newDefaultChecker()`.

internal/updater/checker.go
```diff
@@ -62,43 +72,57 @@ func (u *Checker) CheckFor(desiredChannel, desiredVersion string) (*AvailableUpd
...
+	v.Set("source", "update")
+
+	if desiredVersion != "" {
+		v.Set("target-version", desiredVersion)
+	}
+
+	tag := cfg.GetString(CfgUpdateTag)
+	if tag != "" {
+		v.Set("tag", tag)
+	}
```
Naatan, line 88: Tag should be passed in directly, rather than passing cfg. 

If it really needs to use cfg in the update checker then I'd rather we pass it to the struct rather than this specific call.

Naatan (changes requested):

cmd/state-installer/installer.go
```diff
@@ -61,7 +62,7 @@ func main() {
 	if len(os.Args) > 1 {
 		installPath = os.Args[1]
 	}
-	if err := run(out, installPath, os.Getenv(constants.SessionTokenEnvVarName)); err != nil {
+	if err := run(out, installPath, os.Getenv(constants.SessionTokenEnvVarName), os.Getenv(constants.UpdateTagEnvVarName)); err != nil {
```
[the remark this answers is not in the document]
mdrohmann: The only code path, where the installer is called without the Update tag env var set, would be when a user invokes the installer by themselves...  I guess in that case, we could recycle the tag from the configuration file.  Should we support that?
If it is set and empty, we want to overwrite the tag though.

installers/legacy-install.ps1
```diff
@@ -30,6 +30,8 @@ Set-StrictMode -Off
 
 # ignore project file if we are already in an activated environment
 $Env:ACTIVESTATE_PROJECT=""
+# disable auto-updates to prevent that the State Tool automatically updates to the latest version
+$Env:ACTIVESTATE_CLI_DISABLE_UPDATES = "true"
```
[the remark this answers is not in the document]
mdrohmann: Indeed weird.  But, in my local tests, it got updated to the latest version...  That might have been because I didn't run `state clean uninstall`, such that the config with the timestamp still existed from an earlier installation of that channel.

installers/legacy-install.sh
mdrohmann: Yes, I am still trying to get an understanding of why the mac integration test is failing.

installers/legacy-install.sh
mdrohmann: Okay, I can remove that.  I noticed that this was a block that wasn't unified between the installers.  But I can remove it.

internal/updater/checker.go
mdrohmann: Oh, now I remember why I did that: The `DefaultChecker` is initialized as a package variable.  So, we cannot easily pass arguments to it: https://github.com/ActiveState/cli/blob/1dcea3dcee8042cfbb9bf0ee22812038a5370fa6/internal/updater/checker.go#L34

mdrohmann: I should probably get rid of that DefaultChecker as a variable completely.  It is not saving us anything actually, as it is just returning the struct...

installers/legacy-install.ps1
```diff
@@ -184,7 +186,8 @@ function warningIfadmin() {
 }
 
 function runPreparationStep($installDirectory) {
-    &$installDirectory\$script:STATEEXE _prepare | Write-Host
+    $env:ACTIVESTATE_UPDATE_TAG = $script:UPDATE_TAG
```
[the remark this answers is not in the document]
mdrohmann: I hear you.  Unfortunately, it is very difficult to return values from a powershell function, hence we are manipulating a variable in the global (script) scope.  We do a similar trick in the `install.ps1` already with the `SESSION_TOKEN`.
This probably can be made nicer, but it didn't feel worth it, considering that we are retiring the install scripts anyways...

installers/legacy-install.sh
mdrohmann: It is removed now.  I figured out what it was, the updated `sed` regexp had a non-posix term `\s` which is not supported on the MacOS installation of `sed`.

internal/updater/checker.go
mdrohmann: Yeah.  We already have a story to remove all of these mocked integration tests: https://www.pivotaltracker.com/story/show/177957411
The idea was to do this story, once we have a multi-file update on the release branch, as we cannot really test that we can do all the installs and updates there, before we have an update deployed on s3.  I think that is still a good idea to keep it around until then...  But you can convince me otherwise.  After all, we have tested our update mechanism on the beta channel quite a bit already.

cmd/state-installer/installer.go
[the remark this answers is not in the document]
Naatan: The way I see it we may end up using this through different code paths in the future. And I think it's safe for us to check the config here as well, if the env var is not set.

installers/legacy-install.sh
Naatan: I've got nothing against it other than it could create some new bugs and I'd rather minimize the odds of that happening. This'll move into the go installer soon anyway.

internal/updater/checker.go
Naatan: Yeah makes sense, let's just drop it.

internal/updater/checker.go
```diff
@@ -19,26 +19,36 @@ type httpGetter interface {
...
+	fileURL        string
 	currentChannel string
 	currentVersion string
 	httpreq        httpGetter
 }
 
-var DefaultChecker = newDefaultChecker()
-
-func newDefaultChecker() *Checker {
+func DefaultChecker(cfg Configurable) *Checker {
```
Naatan, line 35: Suggest naming this `NewDefaultChecker`, or something else that clearly conveys we're creating a new checker rather than using some existing default.

Naatan (changes requested):

Naatan (approved):

PR #1511, opened 2021-08-16
Make captain package reusable

daved:
https://www.pivotaltracker.com/story/show/179151696

### DIFF
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/ActiveState/cli/internal/analytics"
+	"github.com/ActiveState/cli/internal/captain"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
@@ -17,6 +18,7 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/machineid"
+	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/rtutils"
 	"github.com/ActiveState/cli/internal/runbits/panics"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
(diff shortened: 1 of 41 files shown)


cmd/state-svc/main.go
```diff
@@ -6,39 +6,37 @@ import (
...
 	"github.com/ActiveState/cli/internal/rtutils"
 	"github.com/ActiveState/cli/internal/runbits/panics"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/rollbar/rollbar-go"
 )
 
-type command string
+var (
+	cmdTop = path.Base(os.Args[0])
+)
```
Naatan, line 31: This doesn't seem like it needs to be in the global namespace.

internal/analytics/analytics.go
```diff
@@ -25,6 +28,22 @@ import (
...
+		return "unknown"
+	}
+	cmdName = path.Base(cmdName)
+	cmdName = strings.TrimSuffix(cmdName, path.Ext(cmdName))
+
+	if cmdName == constants.CommandName {
+		return ""
+	}
+
+	return cmdName + "_"
```
Naatan, line 44: The commands we log to analytics are space separated.

internal/analytics/analytics.go
```diff
@@ -207,6 +226,8 @@ func Configure(cfg configurable) {
 
 // Event logs an event to google analytics
 func Event(category string, action string) {
+	category = appPrefix + category
```
Naatan, line 229: This is prefixing every single category with the appPrefix. What we want is to prefix the event (not category) with the prefix ONLY for the following events:

- run-command
- command-exit

This logic should not live in the analytics package. It should live where those events are being fired.

Naatan (changes requested):

internal/captain/command.go
```diff
@@ -27,6 +29,22 @@ import (
 	"github.com/ActiveState/cli/internal/table"
 )
 
+// appEventPrefix is used for all executables except for the State Tool itself.
+var appEventPrefix string = func() string {
+	cmdName, err := os.Executable()
+	if err != nil {
+		return "unknown"
+	}
```
Naatan, line 37: Let's fall back on `os.Args[0]` instead.

Naatan (changes requested):

internal/captain/command.go
daved: This is done, and I also moved the logic to osutils. It seems like it would be useful in other places. Please let me know if you'd like it back in the captain package.

Naatan (approved):

PR #1512, opened 2021-08-17
Handle tagged update info on release branch

mdrohmann:

### DIFF
--- a/cmd/state/forward.go
+++ b/cmd/state/forward.go
@@ -29,7 +29,7 @@ const LatestVersion = "latest"
 
 type forwardFunc func() error
 
-func forwardFn(cfg updater.Configurable, bindir string, args []string, out output.Outputer, pj *project.Project) (forwardFunc, error) {
+func forwardFn(bindir string, args []string, out output.Outputer, pj *project.Project) (forwardFunc, error) {
 	if pj == nil {
 		return nil, nil
 	}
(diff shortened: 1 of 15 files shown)


cmd/state/forward.go
```diff
@@ -29,7 +29,7 @@ const LatestVersion = "latest"
 
 type forwardFunc func() error
 
-func forwardFn(bindir string, args []string, out output.Outputer, pj *project.Project) (forwardFunc, error) {
+func forwardFn(cfg updater.Configurable, bindir string, args []string, out output.Outputer, pj *project.Project) (forwardFunc, error) {
```
Naatan, line 32: IIRC the forwarder isn't used by the auto updater, it's only used for locked versions. In which case we don't want it to care about tags.

cmd/state/main.go
```diff
@@ -182,7 +182,7 @@ func run(args []string, isInteractive bool, out output.Outputer) error {
 
 	if childCmd != nil && !childCmd.SkipChecks() {
 		// Auto update to latest state tool version, only runs once per day
-		if updated, err := autoUpdate(args, out, pjPath); err != nil || updated {
+		if updated, err := autoUpdate(args, cfg, out, pjPath); err != nil || updated {
```
Naatan, line 185: If this is cheaper to do for the release branch then I'm on board but I'd prefer we don't pass the cfg around so much and instead pass around the thing we actually want, ie. the tag.

installers/install.ps1
```diff
@@ -185,10 +185,14 @@ function warningIfadmin() {
```
Naatan, line 185: We don't need to modify these installers on the release branch, as by the time this lands it'd be the ones on the current beta branch that'd end up being used.

That said, we need to revisit our legacy-installers to fully address this. I'll file a story for that.

Naatan: https://www.pivotaltracker.com/story/show/179305221

Naatan (changes requested):

installers/install.ps1
mdrohmann: Yes, you are right.  The installers are by now almost the same though, only that the `legacy-install` scripts always require a version.  So, it shouldn't be too complicated...

Naatan (approved):

PR #1513, opened 2021-08-17
Initial tracker implementation

MDrakos:
https://www.pivotaltracker.com/story/show/179186755

I left out the synchronization code (ie. `singlethread`, mutexes and a `SetWithLock` function) for this first iteration. Since this is largely based on our config library it would be very easy for me to add if we feel this library should need it.

As per our discussion earlier I can also remove the `Key`s for now and move to labels if we want to go that route. The label's story is here: https://www.pivotaltracker.com/story/show/179259881


internal/tracker/tracker.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return nil, errs.Wrap(err, "Get environment variables query failed")
+	}
+
+	env := make(map[string]string)
+	for rows.Next() {
+		var key, value string
+		err := rows.Scan(&key, &value)
+		if err != nil {
+			logging.Error("Failed to scan path value: %v", err)
```
Naatan, line 145: Should return this error?

Naatan (changes requested):

PR #1515, opened 2021-08-18
Disable systray autostart

MDrakos:
https://www.pivotaltracker.com/story/show/179272109


cmd/state-installer/installer.go
```diff
@@ -175,11 +174,6 @@ func install(installPath string, cfg *config.Instance, out output.Outputer) erro
 		logging.Error("_prepare failed after update: %v\n\nstdout: %s\n\nstderr: %s", err, stdout, stderr)
 	}
 
-	out.Print("Starting ActiveState Desktop")
```
mdrohmann, line 178: I wonder if we should still start the tray app, in case the user had enabled it before.  If that is easy to do, I would add that capability here...

mdrohmann (approved):

mdrohmann (approved):

Naatan (approved):
Nice work!

PR #1516, opened 2021-08-19
Add stdout/stderr info to gio cmd error logging

daved:
https://www.pivotaltracker.com/story/show/179196120


Naatan (changes requested):
I think you can just use this instead:

http://github.com/ActiveState/cli/blob/05262723a028fb2b4a319e982ae7c55b42118bf0/internal/exeutils/exeutils.go#L89-L89

Naatan (approved):

PR #1517, opened 2021-08-23
Disable updates in systray for now

Naatan:
https://www.pivotaltracker.com/story/show/179289140


MDrakos (approved):
We may want to add a comment here as to why this code is commented out and on what conditions we might re-enable it.

PR #1518, opened 2021-08-23
Make legacy-installers the default for now

Naatan:
https://www.pivotaltracker.com/story/show/179305221

@mdrohmann since the legacy-installer scripts are not equal to the release version of the installer scripts, I decided it would be simpler to add a third variant. Of course this does add a lot of overhead, I've recorded this in our release checklist to ensure that these files are up to date before we release.

Ignore the diffs on the installer files. I've made zero changes to them in this PR, I just moved them around and apparently github does not pick up git renames between commits.


mdrohmann (changes requested):
I think the integration tests will need to be updated by pointing them to the correct installers again.

The third AC "Legacy installers should be runnable without a target version" is not implemented.  IIC, because that functionality is brought in by copying the install scripts from the release channel now, but then we would also have to make these scripts use the new info service and tagging, no?

Basically, these changes: I think, the `install.sh` and `install.ps1` should be updated to use

Naatan:
> The third AC "Legacy installers should be runnable without a target version" is not implemented

I addressed it by brining back the release channel installers, which work without the target version.

We need both the release variant and the legacy variant because our locking mechanic directs people to the legacy variant. Technically we could combine the two but it's not worth the extra work for the limited time that they'll be around.

mdrohmann:
> We need both the release variant and the legacy variant because our locking mechanic directs people to the legacy variant. Technically we could combine the two but it's not worth the extra work for the limited time that they'll be around.

That I understand, but note that the `install.sh` and `install.ps1` scripts (as they are), do not use service and therefore not save the update-tag.  And I believe they need to.  Because that will be the scripts that we want new user to install the latest release(single-file) State Tool with, and not update them until we release the tag. No?

mdrohmann (changes requested):
> We need both the release variant and the legacy variant because our locking mechanic directs people to the legacy variant. Technically we could combine the two but it's not worth the extra work for the limited time that they'll be around.

That I understand, but note that the `install.sh` and `install.ps1` scripts (as they are), do not use the service and therefore not save the update-tag.  And I believe they need to.  Because that will be the scripts that we want new user to install the latest release(single-file) State Tool with, and not update them until we release the tag. No?

Naatan:
> > We need both the release variant and the legacy variant because our locking mechanic directs people to the legacy variant. Technically we could combine the two but it's not worth the extra work for the limited time that they'll be around.
> 
> That I understand, but note that the `install.sh` and `install.ps1` scripts (as they are), do not use the service and therefore not save the update-tag. And I believe they need to. Because that will be the scripts that we want new user to install the latest release(single-file) State Tool with, and not update them until we release the tag. No?

Ahh gotcha. I should've copied it from the prerelease channel rather than the release channel. Will fix.

mdrohmann (approved):

PR #1522, opened 2021-08-26
re-implement auto-update

mdrohmann:


Naatan:
Closing this PR in favour of https://github.com/ActiveState/cli/pull/1527

This PR is doing too much at the moment. We discussed that the original AC's were in fact wrong; we don't want to use the state-svc.

cmd/state-installer/internal/installer/installation_darwin.go
```diff
@@ -53,3 +53,8 @@ func createNewSymlink(target, filename string) error {
 	}
 	return nil
 }
+
+// BackupFiles on Darwin does nothing.  The backup/restore mechanism in combination with auto-updates, results in weird behavior, where the forward to an updated State Tool is interrupted by a KILL signal (this possibly is done by a malware protection algorithm: https://developer.apple.com/forums/thread/663456
+func (i *Installation) BackupFiles() error {
+	return nil
+}
```
mdrohmann, line 60: De-activating the backup of existing files on darwin, results in just overwriting them.  Which is allowed in the Linux world, even when the overwritten executable is running.

PR #1523, opened 2021-08-26
Quit event

MDrakos:
https://www.pivotaltracker.com/story/show/179151639

### DIFF
--- a/cmd/state-svc/internal/resolver/resolver.go
+++ b/cmd/state-svc/internal/resolver/resolver.go
@@ -138,6 +138,9 @@ func (r *Resolver) Quit(ctx context.Context) (<-chan bool, error) {
 	logging.Debug("Quit resolver")
 	done := make(chan bool)
 
+	// Each subscriber gets its own copy of a done channel. We wait on the main
+	// done channel and send a response once that channel is closed or a response
+	// is sent
 	go func() {
 		<-r.done
 		done <- true


cmd/state-svc/internal/server/server.go
```diff
@@ -61,6 +65,12 @@ func (s *Server) Start() error {
 	return s.httpServer.Start(s.listener.Addr().String())
 }
 
+func (s *Server) quit() {
+	s.done <- true
+	close(s.done)
```
mdrohmann, line 70: This doesn't look quite right.  As the done channel is non-blocking, it wouldn't wait for the reader to consume the true event, and just immediately close the channel.  That of course, indicates the quite event, by waking up the channel, but most likely the client will not retrieve the `true`.

I think that is okay though.  And we should just remove line 69 (and the buffer size of 1 on the channel).

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -129,3 +133,8 @@ func (r *Resolver) Projects(ctx context.Context) ([]*graph.Project, error) {
 
 	return projects, nil
 }
+
+func (r *Resolver) Quit(ctx context.Context) (<-chan bool, error) {
+	logging.Debug("Quit resolver")
+	return r.done, nil
```
mdrohmann, line 139: As you are returning the `r.done` channel which is the same for every subscriber to this quit event, at most one of the subscribers would receive the `true` value.  But probably none of them (see other comment).  All of them would wake up, which we could argue is enough.  In that case I would change the return channel to `<-chan struct{}` though.

If you actually want to return a `true` event on the channel, I think something like this would work here:
```suggestion
	res := make (chan bool)
	go func() {
		<-r.done
		res <- true
	}()
	return res, nil
```

mdrohmann (changes requested):

PR #1527, opened 2021-08-30
Implemented auto updating

Naatan:

This is largely forked off of https://github.com/ActiveState/cli/pull/1522

Reverting the individual commits to state-svc related code was no viable, so I started from scratch and copied over the changes that were needed, and then modified those further.


MDrakos (dismissed):

cmd/state/autoupdate.go
```diff
@@ -0,0 +1,164 @@
...
+		return false
+
+	// Updates are disabled
+	case strings.ToLower(os.Getenv(constants.DisableUpdates)) == "true":
+		logging.Debug("Not running auto updates because updates are disabled by env var")
+		return false
+
+	// We're on CI
+	case (condition.OnCI()) && strings.ToLower(os.Getenv(constants.DisableUpdates)) != "false":
+		logging.Debug("Not running auto updates because we're on CI")
```
mdrohmann, line 123: Do we need the second part of the boolean expression (the one that checks for `DisableUpdates != "false"`.  We already know that it is not `true`, due to the check above...

cmd/state/autoupdate.go
```diff
@@ -0,0 +1,164 @@
...
+	out.Notice("") // Ensure output doesn't stick to our messaging
+	code, err := relaunch(args)
+	if err != nil {
+		return true, errs.WrapExitCode(err, code)
+	}
+
+	return true, nil
+}
+
+func shouldRunAutoUpdate(args []string, cfg *config.Instance) bool {
```
mdrohmann, line 94: I like that it is split up like this, with log messages indicating the reason for why it was skipped. :+1:

cmd/state-installer/internal/installer/installation.go
```diff
@@ -1,125 +1,69 @@
...
-	if err := fileutils.CopyAndRenameFiles(filepath.Join(i.fromDir, "bin"), i.binaryDir); err != nil {
-		return errs.Wrap(err, "Failed to copy installation files to dir %s", i.binaryDir)
-	}
-	if err := InstallSystemFiles(filepath.Join(i.fromDir, "system"), i.binaryDir, i.appDir); err != nil {
-		return errs.Wrap(err, "Installation of system files failed.")
+		targetFile := filepath.Join(i.binaryDir, file.Name())
+		if fileutils.TargetExists(targetFile) {
+			if err := os.Rename(targetFile, filepath.Join(temp, file.Name())); err != nil {
+				return errs.Wrap(err, "Could not move executable aside prior to install: %s", targetFile)
+			}
```
mdrohmann, line 64: Unfortunately, this might be a problem.  Because `os.Rename` does not move across disks.  So, if the State Tool installation is on a different disk than the `/tmp` directory, it will fail with a weird error: https://github.com/golang/go/issues/13766#issuecomment-167941305

mdrohmann (changes requested):

cmd/state/autoupdate.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		log := logging.Error
+		innerErr := errs.InnerError(err)
+		if os.IsPermission(innerErr) {
+			return false, locale.WrapInputError(err, "auto_update_permission_err", "", constants.DocumentationURL, errs.JoinMessage(err))
+		}
+		if errors.As(err, new(*lockfile.AlreadyLockedError)) {
+			log("Auto update failed because the update lock file is already in use")
+			return false, nil
+		}
```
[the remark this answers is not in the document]
Naatan: That's good! Though you would know better than I, and you don't seem certain. Seems harmless to leave this check in considering?

cmd/state/autoupdate.go
Naatan: I'm pretty sure this is so that for certain CI tests we can explicitly say "do NOT disable updates". Since they are otherwise disabled by default.

I could be wrong though. This isn't new code, I just moved it into a switch.

cmd/state/autoupdate.go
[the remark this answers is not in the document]
mdrohmann: I just confirmed that we do not need the check anymore. But I leave it up to you, whether you want to remove it.  It's certainly harmless to check for it.

cmd/state/autoupdate.go
Naatan: Yeah the conditional was getting a little unwieldy :p

cmd/state/autoupdate.go
mdrohmann: Ah! That makes sense.

cmd/state/autoupdate.go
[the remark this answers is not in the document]
Naatan: If you're certain then I'll just remove it :) Thanks!

cmd/state-installer/internal/installer/installation.go
Naatan: Thanks for spotting that. I've moved it to be a Windows only mechanic and targeted the same folder as the original (like you had originally done). I also added some cleanup code but it's only triggered on a new install. Meaning on Windows you'll always have the old variant around as well, but since it doesn't have an executable extension name it's no runnable so it's only using up disk space.

I figure we can improve this once we can depend more on the state-svc. For now I think it makes sense.

mdrohmann (dismissed):

PR #1531, opened 2021-09-03
Handle panics in analytics and logging/rollbar

mdrohmann:
https://www.pivotaltracker.com/story/show/179457181

Note, that the story actually shows the regular panic message, so it was not even handled by our `main()` function panic handler.
I figured out that this panic handlers only handle panics in the go-routine that their running in: 
https://play.golang.org/p/iXAnMonq4tE

Which means, that other functions that are running go-routines are also not handled by our panic-handlers!
Follow-up story is here: https://www.pivotaltracker.com/story/show/179501830

### DIFF
--- a/cmd/state-installer/installer.go
+++ b/cmd/state-installer/installer.go
@@ -37,7 +37,7 @@ func main() {
 		if panics.HandlePanics(recover(), debug.Stack()) {
 			exitCode = 1
 		}
-		if err := events.WaitForEvents(1*time.Second, rollbar.Close, authentication.LegacyClose); err != nil {
+		if err := events.WaitForEvents(1*time.Second, analytics.Wait, rollbar.Close, authentication.LegacyClose); err != nil {
 			logging.Warning("Failed to wait for rollbar to close: %v", err)
 		}
 		os.Exit(exitCode)
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -357,5 +357,5 @@ func handlePanics(err interface{}) {
 	if err == nil {
 		return
 	}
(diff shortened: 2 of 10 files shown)


cmd/state-installer/installer.go
```diff
@@ -37,7 +37,7 @@ func main() {
 		if panics.HandlePanics(recover(), debug.Stack()) {
 			exitCode = 1
 		}
-		if err := events.WaitForEvents(1*time.Second, rollbar.Close, authentication.LegacyClose); err != nil {
+		if err := events.WaitForEvents(1*time.Second, analytics.Wait, rollbar.Close, authentication.LegacyClose); err != nil {
```
mdrohmann, line 40: I am fairly sure that adding `analytics.Wait` here, fixes the root problem ie., the bug observed by Jeremy.

As the analytics functions may trigger a logging error event that creates a rollbar message, we should wait for the analytics (background) processes to finish before we are closing the rollbar.  That will avoid this issue in the future.

internal/logging/logging.go
```diff
@@ -351,5 +351,11 @@ func BridgeStdLog(level int) {
 			log.SetOutput(b)
 		}
 	}
+}
 
+func handlePanics(err interface{}) {
+	if err == nil {
+		return
+	}
+	fmt.Fprintf(os.Stderr, "Panic occurred in logging module: %v\n", err)
```
Naatan, line 360:
```suggestion
	fmt.Fprintf(os.Stderr, "Failed to log error. Please report this on the forums if it keeps happening. Error: %v\n", err)
```

Naatan (changes requested):

Naatan (approved):

Naatan (approved):

PR #1537, opened 2021-09-08
End-2-end test for install of release candidate

mdrohmann:


test/integration/v29_int_test.go
```diff
@@ -0,0 +1,174 @@
...
+				fmt.Sprintf("%s=%s", constants.OverwriteDefaultInstallationPathEnvVarName, filepath.Join(ts.Dirs.Work, "multi-file")),
+			),
+		)
+		cp.Expect("You are already using the latest State Tool version")
+		cp.ExpectExitCode(0)
+	})
+
+	suite.Run("Update without tag", func() {
+		// remove update-tag
+		err := os.Remove(filepath.Join(ts.Dirs.Config, "config.yaml"))
```
MDrakos, line 103: Rather than removing the config file entirely, if we just removed the tag entry we could also capture the `importLegacyConfig` part of the flow.  Though that is admittedly more work.

installers/install.ps1
```diff
@@ -208,7 +208,7 @@ By running the State Tool installer you consent to the Privacy Policy. This is r
 function fetchArtifacts($downloadDir, $statepkg) {
     # State Tool binary base dir
     $FILE_URL= "https://state-tool.s3.amazonaws.com/update/state"
-    $jsonURL= "https://https://platform.activestate.com/sv/state-update/api/v1/info/legacy?channel=$script:BRANCH&platform=windows&source=install"
+    $jsonURL= "https://platform.activestate.com/sv/state-update/api/v1/info/legacy?channel=$script:BRANCH&platform=windows&source=install"
```
MDrakos, line 211: I feel like this was addressed in a different PR? May want to check that this PR is up to date.

test/integration/v29_int_test.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		suite.Assert().FileExists(filepath.Join(ts.Dirs.Work, "state"+exeExt), "Transitional state tool script does not exist.")
+		suite.FileExists(filepath.Join(ts.Dirs.Work, "multi-file", "state-svc"+exeExt))
+		suite.FileExists(filepath.Join(ts.Dirs.Work, "multi-file", "state-tray"+exeExt))
+	})
+
+}
+
+// TestAutoUpdateFlow meant to test whether updating from an untagged release-candidate will auto-update to a v29 update forwarding to the new State Tool.
+func (suite *V29TestSuite) TestAutoUpdateFlow() {
```
MDrakos, line 129: This test seems to be repeating the `Update without tag` test from above. Are they both needed?

MDrakos (approved):
Looks good to me, just some minor comments

test/integration/v29_int_test.go
mdrohmann: Yeah.  This one is testing the Auto update flow.  `Update without tag` is updating via `state update`.

installers/install.ps1
mdrohmann: will merge with master

test/integration/v29_int_test.go
mdrohmann: :-) That's what I figured.

PR #1538, opened 2021-09-08
Auto Update Fix

MDrakos:
https://www.pivotaltracker.com/story/show/179513231


mdrohmann (approved):
LGTM

cmd/state-installer/installer.go
```diff
@@ -116,6 +116,11 @@ func run(out output.Outputer, installPath, sessionToken string, updateTag *strin
 		}
 	}
 
+	err = cfg.Set(installation.ConfigKeyInstallPath, installPath)
+	if err != nil {
+		logging.Error("Could not set install path in config: %v", errs.JoinMessage(err))
+	}
```
Naatan, line 122: This isn't necessary. By definition people hitting our use-case will use the default install path.

cmd/state-transition-update/main.go
```diff
@@ -157,3 +170,31 @@ func runDefault() (rerr error) {
...
+		"path": filepath.Join(newInstallpath, filepath.Base(exec)),
+	}
+
+	fileBytes := box.Bytes(boxFile)
+	fileStr, err := strutils.ParseTemplate(string(fileBytes), tplParams)
+	if err != nil {
+		return errs.Wrap(err, "Could not parse %s template", boxFile)
+	}
+
+	if err = ioutil.WriteFile(exec, []byte(fileStr), 0755); err != nil {
```
Naatan, line 195: I think for Linux/Mac we'd also have to mark the file executable.

Naatan (changes requested):

cmd/state-transition-update/main.go
MDrakos: Does the `0755` not cover this? The code here is very similar to the executor code [here](https://github.com/ActiveState/cli/blob/master/pkg/platform/runtime/executor/executor.go#L163)

Naatan: I suppose it should. If you tested it and it worked then I'd say we're good to go.

assets/state/state.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+{{.path}} "$@"
```
Naatan, line 3:
```suggestion
"{{.path}}" "$@"
```

I think we'd need this on macOS/Linux as well, for paths with spaces (however unlikely that might be on those platforms).

Naatan (changes requested):

cmd/state-transition-update/main.go
MDrakos: I tested this code when it was in the installer and it worked on macOS. Once this is in place we should put it through and end to end test.

Naatan (approved):

PR #1542, opened 2021-09-09
Transitional state update does not bail when env settings cannot be removed

mdrohmann:
https://www.pivotaltracker.com/story/show/179531537


Naatan (changes requested):
We should create the .profile if it doesn't exist.

mdrohmann:
> We should create the .profile if it doesn't exist.

Even when we just are trying to remove a line from the file?  I feel like that in that case it would be okay to return an error...

Naatan (approved):

PR #1544, opened 2021-09-09
Place additional script file on windows to address niche use-case

Naatan:


mdrohmann (changes requested):
How about this instead? https://github.com/ActiveState/cli/pull/1545

PR #1545, opened 2021-09-10
Transitional State Tool acts as command forwarder

mdrohmann:
https://www.pivotaltracker.com/story/show/179532858

The transitional State Tool updater overwrites itself with a script that calls the State Tool at the new install location.

That works nicely on Linux/Darwin, but on Windows two problem arise:
1. If the updating State tool is called with extension `.exe`, the `.bat` script will not be found
2. If the transitional State Tool has not been removed by the time the relaunch is attempted (this can happen, because the removal happens in the background), it will be called again.

Suggested solution:
- we do not remove the transitional state tool
- we write a configuration entry "new state tool location" into the config file instead.
- When the transitional State tool is called and the "new state tool location" is found in the config file, it is fowarding the command to the new state tool ie., the transitional state tool also acts as the transitional script.

That is the only solution I can come up with that addresses BOTH problems.


cmd/state-transition-update/main.go
```diff
@@ -17,22 +14,23 @@ import (
...
-	"github.com/ActiveState/cli/internal/strutils"
 	"github.com/ActiveState/cli/internal/subshell"
 	"github.com/ActiveState/cli/internal/subshell/sscommon"
 	"github.com/ActiveState/cli/internal/updater"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
-	"github.com/gobuffalo/packr"
 	"github.com/rollbar/rollbar-go"
 )
 
+const cfgNewInstallPath = "new_installation_path"
```
Naatan, line 32: Let's make this a constant in the installation package so we can use it elsewhere. This is really just "installation path".

cmd/state-transition-update/main.go
```diff
@@ -78,18 +76,32 @@ func removeOldStateToolEnvironmentSettings(cfg *config.Instance) error {
...
+	cfg, err := config.New()
+	if err != nil {
+		return errs.Wrap(err, "Could not initialize config")
+	}
+	defer rtutils.Closer(cfg.Close, &rerr)
+
+	// The transitional State Tool will forward all commands to the installed multi-file State Tool, once it has been successfully installed
+	newInstallPath := cfg.GetString(cfgNewInstallPath)
+	newStatePath := appinfo.StateApp(newInstallPath).Exec()
+	if newInstallPath != "" && fileutils.TargetExists(newStatePath) {
```
Naatan, line 89: Let's fall back on the default installPath

Naatan (changes requested):

Naatan (approved):

PR #1549, opened 2021-09-10
Remove transitional state tool

mdrohmann:
https://www.pivotaltracker.com/story/show/179543041

### DIFF
--- a/cmd/state-transition-update/main.go
+++ b/cmd/state-transition-update/main.go
@@ -87,6 +87,10 @@ func run() (rerr error) {
 		logging.Error("Could not determine installation path to check for installation: %s", errs.JoinMessage(err))
 	} else {
 		newStatePath := appinfo.StateApp(newInstallPath).Exec()
+
+		// ensure that path to transitional state tool is stored in the config file
+		transitionalStatePath := appinfo.StateApp().Exec()
+		cfg.Set(installation.CfgTransitionalStateToolPath, transitionalStatePath)
 		if fileutils.TargetExists(newStatePath) && newStatePath != appinfo.StateApp().Exec() {
 			code, _, _ := exeutils.ExecuteAndPipeStd(newStatePath, os.Args[1:], []string{})
 			os.Exit(code)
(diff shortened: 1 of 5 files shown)


cmd/state-transition-update/main.go
```diff
@@ -87,6 +87,10 @@ func run() (rerr error) {
 		logging.Error("Could not determine installation path to check for installation: %s", errs.JoinMessage(err))
 	} else {
 		newStatePath := appinfo.StateApp(newInstallPath).Exec()
+
+		// ensure that path to transitional state tool is stored in the config file
+		transitionalStatePath := appinfo.StateApp().Exec()
+		cfg.Set(installation.CfgTransitionalStateToolPath, transitionalStatePath)
```
Naatan, line 93: This is a non-option. It's already out there. We need to assert it's transitional using the data that exists at the current time.

internal/runners/clean/run_lin_mac.go
```diff
@@ -105,6 +105,12 @@ func removeInstall(cfg configurable) error {
 		}
 	}
 
+	if transitionalStatePath := cfg.GetString(installation.CfgTransitionalStateToolPath); transitionalStatePath != "" {
+		if err := os.Remove(transitionalStatePath); err != nil && !errors.Is(err, os.ErrNotExist) {
+			aggErr = errs.Wrap(aggErr, "Could not remove %s: %v", transitionalStatePath, err)
+		}
+	}
```
Naatan, line 112: This is not what this story is about. We will do this, but only after enough time has passed. It'd be too soon to remove it.

This story is about removing it when you run `state clean uninstall`.

Naatan (changes requested):

internal/runners/clean/run_lin_mac.go
mdrohmann: Yes, this is called during `state clean uninstall`.

cmd/state-transition-update/main.go
mdrohmann: I don't quite understand.  Note that the same block is also on line 165, where the transitional State Tool notifies us about its location, when it is running as `state _prepare`, so when it is being installed.

This block is just a fallback, in case the configuration got deleted inbetween, or the user has already done the v28->v29 transition before...

mdrohmann: Of course, it still leaves room for cases where we do not know about the transitional state tool location, but it would be very difficult to always find it.  We probably do not want to search the entire hard-drive...

Naatan: But it would only do this for people who update after we land this. People who have already updated are out of luck. Seeing as we've had VERY few updates so far and we're aiming to get this out today I guess we can work with that though.. as I don't know how else we would assert this either.

internal/runners/clean/run_lin_mac.go
Naatan: Oh sorry, I missed that this was in a different file.

Naatan (approved):

cmd/state-transition-update/main.go
mdrohmann: Yes, exactly.  With this block, if they still have the PATH to the old (transitional) tool set, there is a high likelihood that they run this code block at least once.  It should even work for `state clean uninstall` actually.

If it is not on their PATH anymore, the stray executable will probably not bother them so much...

PR #1552, opened 2021-09-13
Update activated success message

MDrakos:
https://www.pivotaltracker.com/story/show/179323339

### DIFF
--- a/internal/subshell/sscommon/rcfile.go
+++ b/internal/subshell/sscommon/rcfile.go
@@ -274,7 +274,10 @@ func SetupProjectRcFile(prj *project.Project, templateName, ext string, env map[
 
 	isConsole := ext == ".bat" // yeah this is a dirty cheat, should find something more deterministic
 
-	activatedMessage := txtstyle.NewTitle(locale.Tl("language_installed", "{{.V0}} has been sucessfully activated", prj.Name()))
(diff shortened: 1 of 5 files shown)


internal/subshell/sscommon/rcfile.go
```diff
@@ -274,7 +274,7 @@ func SetupProjectRcFile(prj *project.Project, templateName, ext string, env map[
 
 	isConsole := ext == ".bat" // yeah this is a dirty cheat, should find something more deterministic
 
-	activatedMessage := txtstyle.NewTitle(locale.Tl("youre_activated", "You're Activated!"))
+	activatedMessage := txtstyle.NewTitle(locale.Tl("language_installed", "{{.V0}} has been sucessfully activated", prj.Name()))
```
Naatan, line 277: This needs to account for the project been headless. Maybe for that we can just say "Your virtual environment has been successfully activated".

Naatan (changes requested):

internal/subshell/sscommon/rcfile.go
MDrakos: I think that makes sense. I don't think there is a way for a headless project to be activated as default so `virtual environment` in the messaging should work. Do we want to get product sign-off on this?

Naatan (approved):

PR #1555, opened 2021-09-15
Ensure only one event of the same name is returned

MDrakos:
https://www.pivotaltracker.com/story/show/179541727

This logic could be in a few places. Specifically where it is in this PR or in the `sscommon` package when we are building the project RC file. If we don't want to support multiple events with the same name we could also error when parsing the config, but that will be more involved when taking the constraints into account. Also the AC states it should use the _first_ event it encounters with a name.

### DIFF
--- a/cmd/state/internal/cmdtree/cmdtree.go
+++ b/cmd/state/internal/cmdtree/cmdtree.go
@@ -3,6 +3,7 @@ package cmdtree
 import (
 	"time"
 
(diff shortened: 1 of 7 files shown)


Naatan (changes requested):
This is addressing the issue in the wrong place. It's the firing of the event that cares about the first entry. The project.Events() method should have no opinion on this matter.

Naatan (dismissed):
Very nice.

Naatan (approved):

PR #1556, opened 2021-09-16
Start arch doc

daved:
https://www.pivotaltracker.com/story/show/179459065

### DIFF
--- a/architecture.md
+++ b/architecture.md
@@ -90,8 +90,8 @@ Localization keys and associated values.
 ### pkg/
 
 Packages that are made available for use by any application, including external
-code. The packages that are currently located here are not intended for use by
-external code, and will eventually be moved into the `internal/` directory.
+code. Note we are changing our strategy with regards to these packages, 
+and these will eventually be moved into the `internal/` directory.
 
 #### pkg/cmdlets/
 
@@ -100,7 +100,7 @@ Packages that are made available for use by "runner" packages. Synonymous with
 
 #### pkg/platform/
 
-Packages focused on interacting with the platform. Much of the behavior is
+Packages focused on interacting with platform API's. Much of the behavior is
 generated, but there are also critical components providing platform-dependent
 client-side logic (e.g. `pkg/platform/runtime`).
 
@@ -110,7 +110,7 @@ Packages that provide setup and interaction with the activestate.yaml files.
 
 ### scripts/
 
-Helper scripts.
+Helper scripts for development and deployment processes.
 
 ### test/integration/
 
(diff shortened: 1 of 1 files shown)


architecture.md
```diff
@@ -0,0 +1,143 @@
...
+
+### locale/
+
+Localization keys and associated values.
+
+### pkg/
+
+Packages that are made available for use by any application, including external
+code. The packages that are currently located here are not intended for use by
+external code, and will eventually be moved into the `internal/` directory.
```
Naatan, line 94:
```suggestion
code. Note we are changing our strategy with regards to these packages, 
and these will eventually be moved into the `internal/` directory.
```

architecture.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### pkg/cmdlets/
+
+Packages that are made available for use by "runner" packages. Synonymous with
+`internal/runbits/`, and all new runner-common packages should be placed there.
```
Naatan, line 99: Other way around; runbits is new, cmdlets are old.

architecture.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### pkg/platform/
+
+Packages focused on interacting with the platform. Much of the behavior is
```
Naatan, line 103:
```suggestion
Packages focused on interacting with platform API's. Much of the behavior is
```

architecture.md
```diff
@@ -0,0 +1,143 @@
...
+generated, but there are also critical components providing platform-dependent
+client-side logic (e.g. `pkg/platform/runtime`).
+
+#### pkg/{project,projectfile}/
+
+Packages that provide setup and interaction with the activestate.yaml files.
+
+### scripts/
+
+Helper scripts.
```
Naatan, line 113:
```suggestion
Helper scripts for development and deployment processes.
```

Naatan (changes requested):

architecture.md
daved: That is what I tried to convey, that runbits is the location of all new common packages. Should I word it differently? Maybe:
```suggestion
#### pkg/cmdlets/

Packages that are made available for use by "runner" packages. Synonymous with
`internal/runbits/` with all new runner-common packages being placed there.
```

Naatan: I think you should make the title `pkg/runbits` and instead hint at cmdlets. runbits are the way forward, so it should take center stage.

daved: That directory is covered: https://github.com/ActiveState/cli/pull/1556/files#diff-4e03c430fc7f575c72fd989625acdc7d8d70fd20d3693c893d67d9dbd461ccdbR77-R81

daved: Just pushed a change to remove the pkg/cmdlets entry and update the runbits entry.

Naatan (approved):

Naatan (approved):

PR #1558, opened 2021-09-17
Analytics are sent through state-svc

mdrohmann:
https://www.pivotaltracker.com/story/show/179516742

The project_id is added to the custom dimensions in PR #1559


cmd/state-installer/installer.go
```diff
@@ -37,7 +37,7 @@ func main() {
 		if panics.HandlePanics(recover(), debug.Stack()) {
 			exitCode = 1
 		}
-		if err := events.WaitForEvents(1*time.Second, analytics.Wait, rollbar.Close, authentication.LegacyClose); err != nil {
```
mdrohmann, line 40: This PR switched to a non-global analytics client.  As it is not used anywhere in the `state-installer`, we can be certain that it will not be called anywhere.  So, we do not need to wait for it finish either.

Naatan (changes requested):
Looks great overall!

cmd/state-installer/installer.go
mdrohmann: I think the use-case is still there, yes.  It is just not used right now...  Should I then initialize an `analytics.DefaultClient` anyways.  So, the installer is at least set up for it?

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -0,0 +1,264 @@
...
+func (r *Resolver) Close() {
+	r.closer()
+}
+
+// Events returns a channel to feed eventData directly to the event loop
+func (r *Resolver) Events() chan<- deferred.EventData {
+	return r.events
+}
+
+func (r *Resolver) event(ev deferred.EventData) {
```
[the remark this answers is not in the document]
mdrohmann: I think we should have an `analytics.EventData` struct that can be shared by the deferred package and the resolver...

cmd/state-svc/internal/resolver/analytics/client.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	logging.Error("Error during analytics.sendEvent: %v", errs.Join(err, ":"))
+}
+
+// Configure ties this client to the events loop running as part of the Resolver, un-configured clients defer events to the hard-drive
+func (c *Client) Configure(events chan<- deferred.EventData) {
+	c.events = events
+}
+
+func (c *Client) SetDeferred(da bool) {
```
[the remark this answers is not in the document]
mdrohmann: The only place where we force "Deferrals" is in the `state exec`-runner. I assumed that deferred event processing would be significantly faster, But after doing some basic benchmarking I am not sure anymore:

Deferred timings:
```
» hyperfine --warmup 2 'state exec -- echo deferred'
Benchmark #1: state exec -- echo deferred
  Time (mean ± σ):     342.9 ms ±  72.6 ms    [User: 193.8 ms, System: 65.8 ms]
  Range (min … max):   268.3 ms … 534.4 ms    10 runs
```

Using analytics in state-svc:
```
» hyperfine --warmup 2 'state exec -- echo undeferred'
Benchmark #1: state exec -- echo undeferred
  Time (mean ± σ):     322.7 ms ±   9.0 ms    [User: 197.8 ms, System: 55.6 ms]
  Range (min … max):   308.2 ms … 335.6 ms    10 runs
```

Could just be that the bottleneck for `state exec` is still somewhere else. From the tests above though, I would say that we can get rid of `SetDeferred` yes.

cmd/state-svc/internal/resolver/analytics/analytics.go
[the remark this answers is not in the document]
mdrohmann: I moved the `EventData` to a more prominent place.  It is really not tight to the `deferred` use-case.

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -119,41 +119,15 @@ func NewResolver(cfg *config.Instance) *Resolver {
...
-			Label:       lbl,
-			ProjectName: pn,
-			Output:      o,
-			UserID:      uid,
-		}:
+		case r.events <- ev:
 		case <-r.ctx.Done():
 			// try to defer event if it cannot be scheduled in this session
-			_ = deferred.DeferEvent(category, action, lbl, pn, o, uid)
+			_ = deferred.DeferEvent(ev)
```
Naatan: I don't think the state-svc should "defer" anything. We need to get rid of the concept of deferring altogether. State-svc is the solution. If the state-svc can't send an event then we need log an error so we can address it.

ie. the situation in which it cannot send an event is an error and should not be worked around.

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -0,0 +1,234 @@
...
+	r.closer = func() {
+		cancel()
+		wg.Wait()
+	}
+
+	return r
+}
+
+// AnalyticsEvent schedules the sending of the analytics event
+func (r *Resolver) AnalyticsEvent(_ context.Context, category, action string, label, projectName, out, userID *string) (*graph.AnalyticsEventResponse, error) {
```
Naatan, line 120: Let's move this into the resolver package. 

I think most of the implementation detail you have in this package has nothing to do with "resolving". The resolving layer should be lightweight and simple. It should direct the request to the lower level implementation, which would be this package (though it probably shouldn't live under the `resolver` package at all).

cmd/state-svc/internal/resolver/analytics/client.go
```diff
@@ -0,0 +1,75 @@
...
+	// For now analytics events triggered by the state-svc are NEVER bound to a project or an output-type
+	ev := event.New(category, action, &label, nil, nil, &userID)
+
+	// if events channel is not set yet, we will defer the events to the file system
+	if c.events == nil {
+		if err := deferred.DeferEvent(ev); err != nil {
+			return locale.WrapError(err, "err_analytics_defer", "Could not defer event")
+		}
+		return nil
+	}
```
Naatan, line 62: Please get rid of any deferring altogether. Deferring was a legacy workaround for not having the state-svc. If you still see use-cases for it please talk to me on Slack so we can come to a better solution.

internal/analytics/event/event.go
```diff
@@ -0,0 +1,44 @@
...
+type EventData struct {
+	Category    string `json:"category"`
+	Action      string `json:"action"`
+	Label       string `json:"label"`
+	ProjectName string `json:"project"`
+	Output      string `json:"output_type"`
+	UserID      string `json:"user_id"`
+}
+
+func New(category, action string, label, projectName, out, userID *string) EventData {
```
Naatan, line 15: Can we make the dimensions a separate struct? This all being flat feels messy. Too many arguments.

eg. `events.New(cat, act, lbl, events.Dimensions{})`

Outside of the graphql layers I'd like to see an evident distinction between event data and dimensions.

Naatan (changes requested):

cmd/state-svc/internal/resolver/analytics/client.go
mdrohmann: If we get rid of the deferring, everything will be much simpler...  I can get rid of the event loop too then.  I'll refactor it.

internal/analytics/client.go
```diff
@@ -0,0 +1,103 @@
...
+	a.EventWithLabel(category, action, "")
+}
+
+// EventWithLabel logs an event with a label to google analytics
+func (a *DefaultClient) EventWithLabel(category string, action string, label string) {
+	err := a.sendEvent(category, action, label)
+	if err == nil {
+		return
+	}
+	logging.Error("Error during analytics.sendEvent: %v", errs.Join(err, ":"))
```
Naatan, line 44:
```suggestion
	if err != nil {
		logging.Error("Error during analytics.sendEvent: %v", errs.Join(err, ":"))
	}
```

internal/analytics/mock/analytics.go
```diff
@@ -0,0 +1,46 @@
+package mock
```
Naatan, line 1: Please don't use mocks. Either the functions we're testing need to be written in such a way that they don't rely on functionality we're not interested in testing, or that functionality should just have sensible defaults that don't throw a wrench into the mix.

IIRC our analytics code had a check in there to see if we were in tests, and if so it didn't fire events. I'd still favour that over us writing and maintaining mocks.

Consider mocks banned from our codebase, unless we agree as a team that they are feasible for a particular use-case.

internal/analytics/svc/analytics.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (a *Analytics) Configure(cfg *config.Instance, auth *authentication.Auth) {
+	installSource, err := storage.InstallSource()
+	if err != nil {
+		logging.Error("Could not detect installSource: %s", errs.Join(err, " :: ").Error())
+	}
+
+	id := machineid.UniqID()
+	if id == "unknown" {
```
Naatan, line 49: Where do we return "unknown"? Seems the value used for this case is "11111111-1111-1111-1111-111111111111".

pkg/platform/api/svc/request/analytics.go
```diff
@@ -0,0 +1,39 @@
...
+	}`
+}
+
+func (e *AnalyticsEvent) Vars() map[string]interface{} {
+	return map[string]interface{}{
+		"c":   e.category,
+		"a":   e.action,
+		"l":   e.label,
+		"pn":  e.projectName,
+		"out": e.outputType,
```
Naatan, line 37: Please use full words. I foresee we'll probably want to debug this stuff at some point, so being able to grok this quickly is still important.

Naatan (changes requested):

cmd/state-svc/main.go
```diff
@@ -75,17 +78,19 @@ func run() (rerr error) {
 	}
 	defer rtutils.Closer(cfg.Close, &rerr)
 
-	analytics.Configure(cfg)
 	machineid.Configure(cfg)
 	machineid.SetErrorLogger(logging.Error)
+	an.Configure(cfg, authentication.LegacyGet())
```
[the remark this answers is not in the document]
mdrohmann: But this is inside the `run()` scope...

cmd/state-transition-update/main.go
```diff
@@ -7,7 +7,7 @@ import (
 	"runtime/debug"
 	"time"
 
-	"github.com/ActiveState/cli/internal/analytics"
+	anaConsts "github.com/ActiveState/cli/internal/analytics/constants"
```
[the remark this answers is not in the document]
mdrohmann: You're right.  But I suggest that we do that in a different PR: https://www.pivotaltracker.com/story/show/177974542

internal/analytics/svc/analytics.go
mdrohmann: Ha.  I just copied that from the original file.  But that was outdated, I guess...  Will fix.

cmd/state-svc/main.go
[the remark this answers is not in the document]
Naatan: Sorry I should've been more descriptive. I meant that creating and closing the analytics struct in the `main()` function and then configuring it in the `run()` function feels disjointed.

cmd/state-transition-update/main.go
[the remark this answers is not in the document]
Naatan: Sure. But let's revert these changes in this PR. I don't want to risk breaking something that's currently working and doesn't require any changes.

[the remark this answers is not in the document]
Naatan: Actually, looking again this seems necessary and benign. Let's ship it!

Naatan (dismissed):
Looking good!

PR #1559, opened 2021-09-20
Add projectID to custom dimensions (analytics)

mdrohmann:
https://www.pivotaltracker.com/story/show/179516465


cmd/state-svc/internal/resolver/analytics/customdimensions.go
```diff
@@ -29,5 +29,6 @@ func (d *customDimensions) toMap(projectName, output, userID string) map[string]
 		"11": d.sessionToken,
 		"12": d.uniqID,
 		"13": d.updateTag,
+		"14": projectID,
```
mdrohmann, line 32: @Naatan I think this needs to be added to GA

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -225,36 +228,64 @@ func (r *Resolver) eventLoop() {
 	tick := time.NewTicker(time.Minute * 5)
 	defer tick.Stop()
 
+	projectIDMap := make(map[string]string)
+
 	// flush the deferred data initially
-	if err := r.flushDeferred(); err != nil {
+	if err := r.flushDeferred(projectIDMap); err != nil {
```
Naatan, line 234: This would give these events the project ID at the time that we flush the events. It should be set at the time that the event was triggered. Otherwise you might end up with the wrong project.

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -169,8 +171,9 @@ func (r *Resolver) Events() chan<- deferred.EventData {
 	return r.events
 }
 
-func (r *Resolver) event(ev deferred.EventData) {
-	dimensions := r.customDimensions.toMap(ev.ProjectName, ev.Output, ev.UserID)
+func (r *Resolver) event(ev deferred.EventData, projectIDMap map[string]string) {
```
Naatan, line 174: It does not feel right that this function has such explicit awareness of the project ID. I would expect to see this logic live in the same place where we set the project owner/name dimension.

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -225,36 +228,64 @@ func (r *Resolver) eventLoop() {
...
 	for _, event := range events {
-		r.event(event)
+		r.event(event, projectIDMap)
 	}
 
 	return nil
 }
 
+// projectID resolves the projectID from projectName and caches the result in the provided projectIDMap
+func projectID(projectIDMap map[string]string, projectName string) string {
```
Naatan, line 264: I suggest instead relying on `github.com/patrickmn/go-cache`, which we're already using elsewhere on the state-svc.

cmd/state-svc/internal/resolver/analytics/analytics.go
mdrohmann: If we set the project id _before_ writing them to the "deferred" file, we'll have to resolve the project id in the client (State Tool) process though.  That is we would have to send the API request with every new State Tool invocation.  That might not be the biggest issue if we indeed get rid of "forced" deferred events though:  Then, we'll only do that in occasions where the State Tool does not have a connection to the `state-svc` yet which should almost never happen.  But by adding such an analytics call, we would involuntarily pull in an API call then...

cmd/state-svc/internal/resolver/analytics/analytics.go
mdrohmann: I think you are correct, and the project ID should be resolved outside of this function.

cmd/state-svc/internal/resolver/analytics/analytics.go
mdrohmann: In the latest iteration, I am setting the projectID [as early as possible in the state-svc](https://github.com/ActiveState/cli/blob/603be56e975c4964104344d7bc3053f8f324b1cb/cmd/state-svc/internal/resolver/analytics/analytics.go#L133), but not in the `DefaultClient` yet for the reasons explained above.

Note, however, that this takes the `projectID` resolution out of the event loop, such that I had to [introduce a mutex](https://github.com/ActiveState/cli/blob/603be56e975c4964104344d7bc3053f8f324b1cb/cmd/state-svc/internal/resolver/analytics/analytics.go#L252-L254)

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -219,12 +227,52 @@ func (r *Resolver) flush() error {
...
+func (r *Resolver) addProjectIDToEvent(ev *event.EventData) {
+	if ev.ProjectName == "" || ev.ProjectID != "" {
+		return
+	}
+
+	ev.ProjectID = r.projectID(ev.ProjectName)
+}
+
+// projectID resolves the projectID from projectName and caches the result in the provided projectIDMap
+func (r *Resolver) projectID(projectName string) string {
```
Naatan, line 247: Nit: `projectName` is the namespace I assume? Would prefer we name this something that conveys that (goes for other places where this is used also).

cmd/state-svc/internal/resolver/analytics/analytics.go
```diff
@@ -219,12 +227,52 @@ func (r *Resolver) flush() error {
...
+		logging.Error("Failed to parse project namespace %s: %s", projectName, errs.JoinMessage(err))
+	}
+
+	pj, err := model.FetchProjectByName(pn.Owner, pn.Project)
+	if err != nil {
+		logging.Error("Failed get project by name: %s", errs.JoinMessage(err))
+	}
+
+	pi := string(pj.ProjectID)
+	r.projectIDCache.Set(projectName, pi, cache.DefaultExpiration)
```
Naatan, line 271: `cache.DefaultExpiration` appears to be `0`:

https://github.com/ActiveState/cli/blob/55c0ada4317b700374fff8c780d56b657c226f03/vendor/github.com/patrickmn/go-cache/cache.go#L32

Either way I'd prefer we don't rely on a default that's defined by a third party for something like this.

cmd/state-svc/internal/resolver/analytics/customdimensions.go
Naatan: Done

Naatan (changes requested):

cmd/state-svc/internal/resolver/analytics/analytics.go
mdrohmann: The `DefaultExpiration` is the the first argument to the `cache.New()` function.  That way you have to set it in only one place.  It is not determined by the package.

Naatan (approved):

PR #1562, opened 2021-09-22
Minimize failed file removal logging in moveallfilesrecursively

daved:
https://www.pivotaltracker.com/story/show/179628286

### DIFF
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -472,6 +472,7 @@ func MoveAllFilesRecursively(fromPath, toPath string, cb MoveAllFilesCallback) e
 	}
 
 	// any found files and dirs
+	var loggedFailedRemoval bool
 	for _, fileInfo := range fileInfos {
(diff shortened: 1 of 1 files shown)


daved:
"only reported once per runtime invocation"
This solution reports only once per MoveAllFilesRecursively call. Something more directly tied to runtime invocations will require more complexity. Please let me know if I should take that approach.

mdrohmann (commented):
I realized that we might not need this even.  On the release branch, the `if err != nil` check isn't there yet.  So, we get the message even if the `os.Remove` block succeeds! 

https://github.com/ActiveState/cli/blob/release/internal/fileutils/fileutils.go#L493-L494

All the rollbar errors that I looked at actually report the error as `<nil>`.

mdrohmann:
So, I think we should make sure that this PR gets into release: https://github.com/ActiveState/cli/pull/1532

daved:
Closing in favor of #1564

PR #1563, opened 2021-09-22
Verify state service version

MDrakos:
https://www.pivotaltracker.com/story/show/179542964


cmd/state-svc/manager.go
```diff
@@ -75,6 +75,7 @@ func (s *serviceManager) Stop() error {
 
 	ctx, cancel := context.WithTimeout(context.Background(), svcmanager.MinimalTimeout)
 	defer cancel()
+	svcmgr.SetCheckVersion(false)
```
Naatan, line 78: I'd rather see this as internal logic of the svcmgr. ie. the manager should know not to care about version checking for the "stop" event.

internal/runners/clean/stop.go
```diff
@@ -55,7 +56,11 @@ func stopServices(cfg configurable, out output.Outputer, ignoreErrors bool) erro
 		var isStopped bool
 		m := svcmanager.New(cfg)
 		for x := 0; x < 30; x++ {
-			isStopped = !m.Ready()
+			err = m.Ready()
+			if err != nil {
+				logging.Debug("Ready error: %v", err)
+				isStopped = true
+			}
```
Naatan, line 63: Please use a separate bool for the ready check itself. The Ready() function can then decide that error equals "not ready".

internal/svcmanager/svcmanager.go
```diff
@@ -58,9 +65,14 @@ func (m *Manager) WaitWithContext(ctx context.Context) error {
 		case <-ctx.Done():
 			return nil
 		default:
-			if m.Ready() {
+			err := m.Ready()
+			if err == nil {
 				return nil
 			}
+			if m.checkVersion && errs.Matches(err, errVersionMismatch{}) {
```
Naatan, line 72: Checking `m.checkVersion` seems unnecessary. Or said differently; if `m.checkVersion=false` then `errVersionMismatch` should never happen in the first place.

internal/svcmanager/svcmanager.go
```diff
@@ -58,9 +65,14 @@ func (m *Manager) WaitWithContext(ctx context.Context) error {
...
+				return errs.Wrap(err, "Incorrect State Service version")
+			}
+			logging.Debug("Ready failed, assuming we're not ready: %v", errs.JoinMessage(err))
```
Naatan, line 75: There's no need for assumptions.

internal/svcmanager/svcmanager.go
```diff
@@ -72,23 +84,22 @@ func (m *Manager) Wait() error {
...
-func (m *Manager) Ready() bool {
+func (m *Manager) Ready() error {
 	if m.ready {
-		return true
+		return nil
 	}
 
 	if m.cfg.GetInt(constants.SvcConfigPort) == 0 {
-		return false
+		return errs.New("Service port not set in config")
```
Naatan, line 93: This isn't an error. It not being said simply means it's not ready because it hasn't started.

internal/svcmanager/svcmanager.go
```diff
@@ -72,23 +84,22 @@ func (m *Manager) Wait() error {
...
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), MinimalTimeout)
 	defer cancel()
 	if err := m.ping(ctx); err != nil {
-		logging.Debug("Ping failed, assuming we're not ready: %v", errs.JoinMessage(err))
-		return false
+		return errs.Wrap(err, "Ping failed, service may not yet be ready")
```
Naatan, line 99: Also not an error.

internal/svcmanager/svcmanager.go
```diff
@@ -101,5 +112,10 @@ func (m *Manager) ping(ctx context.Context) error {
 	if err := client.RunWithContext(ctx, r, &resp); err != nil {
 		return err
 	}
+
+	if resp.Version.State.Version != constants.Version && resp.Version.State.Branch != constants.BranchName {
+		return errVersionMismatch{locale.NewError("err_ping_version_mismatch")}
```
Naatan, line 117: So far our typed errors have always been pointers IIRC. Prefer we stick with that convention.

locale/en-us.yaml
```diff
@@ -1728,4 +1728,8 @@ lock_update_legacy_version:
 lock_version_mismatch:
   other: This project is locked at State Tool version [NOTICE]{{.V0}}[/RESET]. Your current State Tool version is [NOTICE]{{.V1}}@{{.V2}}[/RESET].
 lock_update_lock:
-  other: You can lock the project to the running State Tool version with `[ACTIONABLE]state update lock[/RESET]`.
\ No newline at end of file
+  other: You can lock the project to the running State Tool version with `[ACTIONABLE]state update lock[/RESET]`.
+err_ping_version_mismatch:
+  other: |
+    The installed version of the State Service does not match that of the client.
```
Naatan, line 1734:
```suggestion
    The installed version of the State Tool Service does not match that of the client.
```

locale/en-us.yaml
```diff
@@ -1728,4 +1728,8 @@ lock_update_legacy_version:
...
+    Please stop the service with [ACTIONABLE]state-svc stop[/RESET] and try your command again.
```
Naatan, line 1735:
```suggestion
    Please stop the service with [ACTIONABLE]state-svc stop[/RESET] and try your command again.
    If this issue persists please reinstall the State Tool.
```

Naatan (changes requested):

internal/svcmanager/svcmanager.go
```diff
@@ -101,5 +115,10 @@ func (m *Manager) ping(ctx context.Context) error {
 	if err := client.RunWithContext(ctx, r, &resp); err != nil {
 		return err
 	}
+
+	if m.checkVersion && resp.Version.State.Version != constants.Version && resp.Version.State.Branch != constants.BranchName {
+		return &errVersionMismatch{locale.NewError("err_ping_version_mismatch")}
```
Naatan, line 120: Nit: Just occurred to me, you don't really need the error type. Since you're not wrapping another error or need any custom methods could use: 

```
var MyError = locale.NewError("err_myerror")
errors.Is(err, MyError)
```

That should work even if `MyError` ends up getting wrapped itself.

pkg/platform/model/svc.go
```diff
@@ -95,6 +96,7 @@ func (m *SvcModel) StopServer() error {
 		return errs.Wrap(err, "Could not create request to quit svc")
 	}
 
+	m.svcm.SetCheckVersion(false)
```
Naatan, line 99: My feedback was that this should be internal to the svcm, this is still external.

Naatan (changes requested):

pkg/platform/model/svc.go
MDrakos: Right, still working on it. Wasn't ready for review. The issue with making this internal the the service manager is that it has no knowledge of a `stop` event. This service manager just starts and waits for the service. I'm looking for a way to give the service manager that knowledge without refactoring the code too much.

Naatan: Oh I'm sorry. I thought it was ready for review.

internal/svcmanager/svcmanager.go
```diff
@@ -20,20 +21,27 @@ import (
...
 
 func New(cfg configurable) *Manager {
-	mgr := &Manager{false, cfg}
+	mgr := &Manager{false, true, cfg}
 	return mgr
 }
 
+func (m *Manager) SetCheckVersion(check bool) {
+	m.checkVersion = check
+}
```
Naatan, line 43: This isn't used (nor should it be).

internal/svcmanager/svcmanager.go
```diff
@@ -20,20 +21,27 @@ import (
 // MinimalTimeout is the minimum timeout required for requests that are meant to be near-instant
 const MinimalTimeout = 500 * time.Millisecond
 
+var errVersionMismatch = locale.NewError("err_ping_version_mismatch")
+
 type Manager struct {
-	ready bool
-	cfg   configurable
+	ready        bool
+	checkVersion bool
```
Naatan, line 28: This isn't doing anything, it's always true as per the logic. Should just drop this.

pkg/platform/model/svc.go
```diff
@@ -85,10 +85,15 @@ func (m *SvcModel) CheckUpdate(ctx context.Context) (*graph.AvailableUpdate, err
...
+func StopServer(cfg *config.Instance) error {
 	htClient := retryhttp.DefaultClient.StandardClient()
 
-	quitAddress := fmt.Sprintf("%s/__quit", m.client.BaseUrl())
+	client, err := svc.New(cfg)
+	if err != nil {
+		return errs.Wrap(err, "Could not initialize svc client")
+	}
+
+	quitAddress := fmt.Sprintf("%s/__quit", client.BaseUrl())
```
Naatan, line 96: This doesn't seem to be an improvement? What's wrong with `m.client`?

pkg/platform/model/svc.go
```diff
@@ -85,10 +85,15 @@ func (m *SvcModel) CheckUpdate(ctx context.Context) (*graph.AvailableUpdate, err
 	return &u.AvailableUpdate, nil
 }
 
-func (m *SvcModel) StopServer() error {
+func StopServer(cfg *config.Instance) error {
```
Naatan, line 88: I don't think this should be on the model at all. This was only called from one place right? Just implement the logic there. Or if it needs to stay in the model; name it something more explicit. `model.StopServer()`; what server? Probably want `model.StopStateService()` or something. But I'd rather not see it on the model at all.

Naatan (changes requested):

internal/svcmanager/svcmanager.go
MDrakos: Whoops, meant to remove this and the `bool`

pkg/platform/model/svc.go
MDrakos: I've moved the function to the service manager that called it originally. Though if we ever do have a need to stop the server from the model we will have to move the code back.

Naatan (dismissed):

PR #1566, opened 2021-09-23
Use mediator to get default lang version for init

daved:
https://www.pivotaltracker.com/story/show/179642618


daved:
Closing in favor of #1580

PR #1570, opened 2021-09-24
State learn

MDrakos:
https://www.pivotaltracker.com/story/show/179323455

Not entirely sure about the messaging here as the comments on the ticket don't exactly account for Go's error handling.

### DIFF
--- a/internal/runners/learn/learn.go
+++ b/internal/runners/learn/learn.go
@@ -1,13 +1,12 @@
 package learn
 
 import (
+	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/skratchdot/open-golang/open"
 )
-
-const cheetSheetURL = "https://platform.activestate.com/state-tool-cheat-sheet"
 
 type Learn struct {
 	out output.Outputer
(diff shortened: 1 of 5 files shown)


internal/runners/learn/learn.go
```diff
@@ -0,0 +1,32 @@
+package learn
+
+import (
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/internal/primer"
+	"github.com/skratchdot/open-golang/open"
+)
+
+const cheetSheetURL = "https://platform.activestate.com/state-tool-cheat-sheet"
```
Naatan, line 10: Please use the constants package.

internal/runners/learn/learn.go
```diff
@@ -0,0 +1,32 @@
...
+	return &Learn{prime.Output()}
+}
+
+func (l *Learn) Run() error {
+	err := open.Run(cheetSheetURL)
+	if err != nil {
+		return locale.WrapError(err, "err_learn_open", cheetSheetURL)
+	}
+
+	l.out.Print(locale.Tr("learn_info", "Opening [ACTIONABLE]{{.V0}}[/RESET] in browser", cheetSheetURL))
```
Naatan, line 30: Should probably print this before and not after, in case it errors. But also to satisfy my OCD :p

locale/en-us.yaml
```diff
@@ -1729,9 +1729,11 @@ lock_version_mismatch:
   other: This project is locked at State Tool version [NOTICE]{{.V0}}[/RESET]. Your current State Tool version is [NOTICE]{{.V1}}@{{.V2}}[/RESET].
 lock_update_lock:
   other: You can lock the project to the running State Tool version with `[ACTIONABLE]state update lock[/RESET]`.
+err_learn_open:
+  other: "Could not open cheat sheet in browser, visit: [ACTIONABLE]{{.V0}}[/RESET] to learn about common commands"
```
Naatan, line 1733: Since we'll already print the "Opening .." message we can probably drop the URL here so we don't stutter.

Naatan (changes requested):

Naatan (approved):

PR #1571, opened 2021-09-27
Improve executor recursion detection

mdrohmann:
https://www.pivotaltracker.com/story/show/179277337

Apparently, the python debug launcher in VSCode is smart enough to call the script that it is trying to debug with the exact same python executable that the launcher was started with.  Ie., it is calling the script with the full path to the python executor (`<HOME>/.cache/activestate/bin/python3`).  So, the executor gets called in an environment that has already been set up by a previous executor.  That is unfortunate, but actually completely ok.

So, I added some integration tests and made the loop detection more sophisticated: It is now checking if the subprocess called by the exec function would actually call an executor script...  If not, then we should be fine. :pray:


test/integration/activate_int_test.go
```diff
@@ -217,14 +218,36 @@ func (suite *ActivateIntegrationTestSuite) activatePython(version string, extraE
...
 	)
 	cp.Expect("ActiveState Software Inc.")
 	cp.ExpectExitCode(0)
+
+	// check that default activation does not recurse
+	cp = ts.SpawnCmdWithOpts(
+		executor,
+		e2e.WithArgs("-c", fmt.Sprintf(`import subprocess; subprocess.run(["%s", "-c", "print('hello')"])`, executor)),
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+	)
```
daved, line 236: I surely may be testing incorrectly; I don't think this represents the recursion issue when VSCode is involved. I've tested this locally without success.

daved (changes requested):

test/integration/activate_int_test.go
daved: Scratch this. Somehow I was on the wrong commit. It does seem to work.

daved (approved):

internal/exeutils/exeutils_windows.go
```diff
@@ -1,3 +1,25 @@
...
+
+	"github.com/ActiveState/cli/internal/fileutils"
+)
+
 const Extension = ".exe"
+
+// PathForExecutable returns the first path from the PATH env var for which the executable exists
+func PathForExecutable(executable string) string {
+	candidates := strings.Split(os.Getenv("PATH"), string(os.PathListSeparator))
+	exts := append([]string{""}, strings.Split(os.Getenv("PATHEXT"), string(os.PathListSeparator))...)
```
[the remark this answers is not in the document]
mdrohmann: In case the executable is eg., `python.exe` we do not want to append the PATHEXT extensions to it, as that would look for `python.exe.exe`.  I'll leave a comment.

internal/runners/exec/exec.go
```diff
@@ -104,7 +109,12 @@ func (s *Exec) Run(params *Params, args ...string) error {
 	logging.Debug("Trying to exec %s on PATH=%s", args[0], env["PATH"])
 	// Ensure that we are not calling the exec recursively
 	if _, isBeingShimmed := env[constants.ExecEnvVarName]; isBeingShimmed {
```
Naatan, line 111: We should just get rid of the env var, since the new detection you implemented doesn't depend on it.

internal/exeutils/exeutils_windows.go
```diff
@@ -1,3 +1,32 @@
...
+
 const Extension = ".exe"
+
+// PathForExecutable returns the first path from the PATH env var for which the executable exists
+func PathForExecutable(executable string) string {
+	candidates := strings.Split(os.Getenv("PATH"), string(os.PathListSeparator))
+	var exts []string // list of extensions to look for
+	pathexts := strings.Split(os.Getenv("PATHEXT"), string(os.PathListSeparator))
+	// if executable has valid extension for an executable file, we have to check for its existence without appending more extensions
+	if funk.ContainsString(pathexts, filepath.Ext(executable)) {
```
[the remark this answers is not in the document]
mdrohmann: If not provided with an exe-extension, it still tries to find the file after appending any of the exts in `pathexts`.  I think I should add a unit test for it though.

internal/exeutils/exeutils_windows.go
```diff
@@ -1,3 +1,32 @@
...
+		exts = append(exts, "")
+	}
+	exts = append(exts, pathexts...)
+	for _, p := range candidates {
+		for _, ext := range exts {
+			if fileutils.TargetExists(filepath.Join(p, executable+ext)) {
```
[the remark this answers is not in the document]
mdrohmann: That's why (in that case), exts = `[]string{"", ".com", ".exe", ...}` So, it also checks it with an empty `ext`.

internal/runners/exec/exec.go
```diff
@@ -104,7 +109,12 @@ func (s *Exec) Run(params *Params, args ...string) error {
...
-		return locale.NewError("err_exec_recursive_loop", "Detected recursive loop while calling {{.V0}}", args[0])
+		// Ensure that we would not call the executor recursively: The path for the executable should be different from the default bin dir
+		p := exeutils.PathForExecutable(filepath.Base(args[0]))
```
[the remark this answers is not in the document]
mdrohmann: I checked, it has the expected string in it.

daved (dismissed):
This is still working for me under the same conditions in VSCode which produced the original error.

internal/exeutils/exeutils_windows.go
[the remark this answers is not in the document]
Naatan: Yeah unit test would be good for this

internal/exeutils/exeutils_unix_test.go
```diff
@@ -0,0 +1,25 @@
...
+func Test_PathForExecutables(t *testing.T) {
+	tmpdir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpdir)
+
+	fileutils.Touch(filepath.Join(tmpdir, "state"))
+	assert.Equal(t, filepath.Join(tmpdir, "state"), FindExecutableOnPath("state", "/other_path:"+tmpdir))
+	assert.Equal(t, "", FindExecutableOnPath("non-existent", "/other_path:"+tmpdir))
+	assert.Equal(t, "", FindExecutableOnPath("state", "/other_path"))
+}
```
Naatan, line 25: I think we can implement this code so that the unit tests aren't dependent on the filesystem or environment. Implement a low level function that has the actual logic we want to test and takes the files on PATH as a string slice.

internal/exeutils/exeutils_windows.go
```diff
@@ -11,20 +11,29 @@ import (
...
+
+func FindExecutableOnPath(executable, PATH string) string {
+	return findExecutable(executable, PATH, os.Getenv("PATHEXT"))
+}
+
+func findExecutable(executable, PATH, PATHEXT string) string {
+	candidates := strings.Split(PATH, string(os.PathListSeparator))
 	var exts []string // list of extensions to look for
-	pathexts := strings.Split(os.Getenv("PATHEXT"), string(os.PathListSeparator))
+	pathexts := funk.Map(strings.Split(PATHEXT, string(os.PathListSeparator)), strings.ToLower).([]string)
```
Naatan, line 26: Does windows really ignore case here?

internal/exeutils/exeutils_windows_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_PathForExecutables(t *testing.T) {
```
Naatan, line 16: Same as the other test; I think we can implement this without relying on filesystem or env.

internal/runners/exec/exec.go
```diff
@@ -108,16 +113,20 @@ func (s *Exec) Run(params *Params, args ...string) error {
...
+	lastLvl, err := strconv.ParseInt(os.Getenv(constants.ExecRecursionLevelEnvVarName), 10, 32)
+	if err != nil {
+		recursionLvl = int(lastLvl) + 1
+	}
+
+	// Report recursive execution of executor: The path for the executable should be different from the default bin dir
+	p := exeutils.FindExecutableOnOSPath(filepath.Base(args[0]))
+	binDir := filepath.Clean(globaldefault.BinDir(s.cfg))
+	if p == binDir && (recursionLvl == 1 || recursionLvl == 10 || recursionLvl == 50) {
+		return logging.Criticalf("executor recursion detected: %s (lvl=%d)", strings.Join(args, " "), recursionLvl)
```
Naatan, line 127: Why is this returning? Also as noted we just want to log this without affecting the user. So that makes it not critical.

Naatan: Additionally this should include the process args of the invoking process, as noted in my comment on the previous review. Logging this is pointless if we can't identify what led to the recursion.

Naatan (changes requested):

internal/exeutils/exeutils_windows.go
mdrohmann: Windows file systems are case insensitive, yes.  Usually `PATHEXT` has all upper case endings like `.COM;.EXE;.BAT`, but of course, lower case endings are accepeted as well.

internal/exeutils/exeutils_windows_test.go
mdrohmann: I can do that.  Two concerns though:
1. In this case, especially when we deal with Windows, I prefer to also test how the file system functions operate.  When abstracting this, I might miss some important behavior...
2. Performance: Gathering all the potential paths, looks at more files and directories than necessary.  Usually not a big problem, but in the executors, performance is a priority of course, even if it is just a few milliseconds...

internal/runners/exec/exec.go
mdrohmann: I found a package that (platform-independently) resolves the executable name of the calling process.  That might be enough to give us hints.  Getting all the arguments for a process seems to be more difficult to do (on all platforms!).

internal/exeutils/exeutils_windows_test.go
Naatan: Those are good points. But I'd argue if we really want automated tests to cover those things they should be integration tests.

internal/runners/exec/exec.go
Naatan: Ok, let's start with the executable name.

Note I think we have a psutil lib in use already. Might want to see if that one can already offer this feature so we don't bring in a new package for nothing.

internal/runners/exec/exec.go
```diff
@@ -146,12 +146,18 @@ func (s *Exec) Run(params *Params, args ...string) error {
...
+	p, err := process.NewProcess(int32(os.Getppid()))
 	if err != nil {
 		logging.Debug("Could not find parent process of executor: %v", err)
 		return "unknown"
 	}
 
-	return p.Executable()
+	args, err := p.CmdlineSlice()
+	if err != nil {
+		logging.Debug("")
```
Naatan, line 158: Forgot?

test/integration/activate_int_test.go
```diff
@@ -215,25 +215,28 @@ func (suite *ActivateIntegrationTestSuite) activatePython(version string, extraE
 	cp.Expect("ActiveState Software Inc.")
 	cp.ExpectExitCode(0)
 
-	// check that default activation does not recurse
+	// check that default activation takes count of recursion level
```
Naatan, line 218: Please implement a standalone test. We're already overloading a lot of these integration tests. Makes it hard to maintain.

Naatan (changes requested):

Naatan (approved):

PR #1575, opened 2021-10-01
Don't install again if state tool is already installed

Naatan:
https://activestatef.atlassian.net/browse/DX-413

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -197,18 +197,18 @@ func main() {
 }
 
 func execute(out output.Outputer, cfg *config.Instance, args []string, params *Params) error {
-	installPath, _ := installation.InstallPath()
-	// Check if state tool already installed
-	if !params.force && fileutils.TargetExists(appinfo.StateApp(installPath).Exec()) || fileutils.TargetExists(appinfo.StateApp(params.path).Exec()) {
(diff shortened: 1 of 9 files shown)


cmd/state-installer/cmd.go
```diff
@@ -192,7 +199,17 @@ func main() {
...
-		return installOrUpdateFromLocalSource(out, cfg, args, params)
+		if err := installOrUpdateFromLocalSource(out, cfg, args, params); err != nil {
+			return err
+		}
+		return postInstallEvents(out, params)
+	}
+
+	installPath, _ := installation.InstallPath()
+	// Check if state tool already installed
+	if !params.force && fileutils.TargetExists(appinfo.StateApp(installPath).Exec()) || fileutils.TargetExists(appinfo.StateApp(params.path).Exec()) {
```
mdrohmann, line 210: I think you need parentheses here:

```suggestion
	if !params.force && (fileutils.TargetExists(appinfo.StateApp(installPath).Exec()) || fileutils.TargetExists(appinfo.StateApp(params.path).Exec())) {
```

mdrohmann (changes requested):

cmd/state-installer/cmd.go
Naatan: Good eye!

mdrohmann (dismissed):

PR #1580, opened 2021-10-05
Use mediator to get default lang version for init

daved:
https://activestatef.atlassian.net/browse/DX-242

### DIFF
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -4,9 +4,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/ActiveState/cli/pkg/platform/model"
-	"github.com/gobuffalo/packr"
-
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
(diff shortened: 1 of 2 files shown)


internal/runners/initialize/init.go
```diff
@@ -184,3 +181,18 @@ func run(params *RunParams, out output.Outputer) (string, error) {
 
 	return params.Path, nil
 }
+
+func deriveVersion(lang language.Language, version string) string {
+	if version == "" {
```
Naatan, line 186: Just return early if `!= ""`

internal/runners/initialize/init.go
```diff
@@ -184,3 +181,18 @@ func run(params *RunParams, out output.Outputer) (string, error) {
...
+		version = lang.RecommendedVersion()
+		langs, err := model.FetchSupportedLanguages(model.HostPlatform)
+		if err == nil {
```
Naatan, line 189: Should at least log the error here. And imo too much logic inside this conditional, rather invert it as well.

Naatan (changes requested):

Naatan (changes requested):

PR #1582, opened 2021-10-05
State Update is now blocking

Naatan:
https://activestatef.atlassian.net/browse/DX-418


mdrohmann (approved):
This all looks good to me.

PR #1585, opened 2021-10-07
Fix channel parsing in install.sh

MDrakos:


installers/install.sh
```diff
@@ -6,7 +6,7 @@ BASE_FILE_URL="https://state-tool.s3.amazonaws.com/update/state"
 # Name of the installer executable and archive.
 INSTALLERNAME="state-installer"
 # Channel the installer will target
-CHANNEL='release'
+CHANNEL=""
```
Naatan, line 9: This will break https://github.com/ActiveState/cli/blob/23e6e69f9e15104508ecb2fe1ce8e58357bd5e78/activestate.yaml#L250

installers/install.sh
```diff
@@ -21,12 +21,23 @@ if [ "$SESSION_TOKEN" != "$SESSION_TOKEN_VERIFY" ]; then
 fi
 
 parseChannel() {
-  i=0
-  for arg in ${@}; do
-    i=$((i + 1)) && [ "${arg}" != "-b" ] && continue
-    echo "${@}" | cut -d' ' -f$((${i} + 1)) && return
+  cap="false"
+  channel="release"
```
Naatan, line 25: Can just use the CHANNEL var here and revert the above change emptying it.

installers/install.sh
```diff
@@ -20,15 +20,20 @@ if [ "$SESSION_TOKEN" != "$SESSION_TOKEN_VERIFY" ]; then
...
-parseChannel() {
-  i=0
-  for arg in ${@}; do
-    i=$((i + 1)) && [ "${arg}" != "-b" ] && continue
-    echo "${@}" | cut -d' ' -f$((${i} + 1)) && return
-  done
-  echo $CHANNEL
-}
-CHANNEL=$(parseChannel "$@")
+cap="false"
```
Naatan, line 23:
```suggestion
# Set the channel according to the `-b` flag, if provided
cap="false"
```

Naatan (dismissed):

PR #1589, opened 2021-10-08
Also fire pixel events using more reliable means

Naatan:


cmd/state-svc/main.go
```diff
@@ -227,3 +235,18 @@ func runStatus(cfg *config.Instance) error {
...
 	return nil
 }
+
+func fireEvent(query string) {
+	pixelURL, err := url.Parse("https://state-tool.s3.amazonaws.com/pixel")
+	if err != nil {
+		logging.Error("Invalid URL for analytics S3 pixel")
+		return
+	}
+	pixelURL.RawQuery = query
```
mdrohmann, line 245: I am trusting that you tested this actually works. :)

mdrohmann (approved):
This should do the trick...

PR #1592, opened 2021-10-12
Use `.bashrc` on MacOS

MDrakos:
We were using `.profile` which appears to only be sourced for login shells. Some SO answers claim that the Terminal app creates a login shell when it is started but this did not appear to be the case for me. On my setup the installer was updating `~/.profile` but that was not being picked up by new iTerm or Terminal windows. Changing to `~/.bashrc` for the bash shell fixes it.

It also appears that if the user has a `~/.bash_profile` the `~/.profile` may not be sourced even for a login shell. The subshell refactor could take this into account by attempting to check which configuration files exist on the user's system for the current shell and using the most appropriate one.

### DIFF
--- a/internal/subshell/bash/bash.go
+++ b/internal/subshell/bash/bash.go
@@ -107,8 +107,8 @@ func (v *SubShell) RcFile() (string, error) {
 	}
 
 	rcFilePath := filepath.Join(homeDir, rcFileName)
-	// On MacOS the .bashrc file is not created by default
-	if runtime.GOOS == "darwin" && !fileutils.FileExists(rcFilePath) {
+	// Ensure the .bashrc file exists. On some platforms it is not created by default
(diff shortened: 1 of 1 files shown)


internal/subshell/bash/bash.go
```diff
@@ -112,7 +106,16 @@ func (v *SubShell) RcFile() (string, error) {
 		return "", errs.Wrap(err, "IO failure")
 	}
 
-	return filepath.Join(homeDir, rcFileName), nil
+	rcFilePath := filepath.Join(homeDir, rcFileName)
+	// On MacOS the .bashrc file is not created by default
+	if runtime.GOOS == "darwin" && !fileutils.FileExists(rcFilePath) {
```
Naatan, line 111: I don't think this needs to be macOS only. Also use `TargetExists` instead; since their rc file could be a symlink.

PR #1605, opened 2021-10-22
Start subshell for new installs

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-151" title="DX-151" target="_blank">StoryDX-151</a>  Always start subshell after installation
  </td></table>
  <br />
 

https://activestatef.atlassian.net/browse/DX-151


cmd/state-installer/cmd.go
```diff
@@ -190,13 +192,27 @@ func main() {
 	)
 
 	if err := cmd.Execute(processedArgs[1:]); err != nil {
+		// if error carries an exit code, set that and skip error logging
+		if errExitCode := extractExitCode(err); errExitCode > 0 {
+			exitCode = errExitCode
+			return
+		}
```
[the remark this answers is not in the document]
daved: The exit code errors will only come from subshell executions that contain an exit code and it is the expected behavior of subshells.

cmd/state-installer/cmd.go
```diff
@@ -190,13 +192,27 @@ func main() {
...
+
 		logging.Error(errs.JoinMessage(err))
 		out.Error(err.Error())
 		exitCode = 1
 		return
 	}
 }
 
+func extractExitCode(err error) int {
```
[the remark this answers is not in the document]
daved: UnwrapExitCode does not allow a distinction to be determined between a non-exitcode error (code: 1) and an exitcode error (code: non zero). I've updated the extractExitCode function, and I suggest that we modify UnwrapExitCode and use that instead.

cmd/state-installer/cmd.go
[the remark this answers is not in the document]
Naatan: It's still an assumption. I understand that's what it does now, but the underlying code might change in the future.

cmd/state-installer/cmd.go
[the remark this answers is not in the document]
Naatan: I don't follow your meaning. There is exit code zero (success) and anything else (failure). Are you suggesting a third error type?

cmd/state-installer/cmd.go
```diff
@@ -210,14 +212,34 @@ func main() {
 			an.EventWithLabel(AnalyticsCat, "error", errs.JoinMessage(err))
 			logging.Critical("Installer error: " + errs.JoinMessage(err))
 		}
+
 		an.EventWithLabel(AnalyticsFunnelCat, "fail", err.Error())
+		// if error carries an exit code, set that and skip error logging
+		if exCode, ok := extractExitCode(err); ok {
+			exitCode = exCode
+			return
+		}
```
Naatan, line 221: You didn't achieve anything by creating this function, you're still exiting 1 even when the error didn't satisfy the interface. Please just use `errs.UnwrapExitCode`.

I think what you want is some way to communicate "Exit code came from subprocess". There's merit there but let's not get distracted in this story. Feel free to file a followup.

cmd/state-installer/cmd.go
```diff
@@ -354,20 +375,43 @@ func postInstallEvents(out output.Outputer, an analytics.Dispatcher, params *Par
 		an.Event(AnalyticsFunnelCat, "forward-activate-default")
 
 		out.Print(fmt.Sprintf("\nRunning `[ACTIONABLE]state activate --default %s[/RESET]`\n", params.activateDefault.String()))
-		if _, _, err := exeutils.ExecuteAndPipeStd(stateExe, []string{"activate", params.activateDefault.String(), "--default"}, env); err != nil {
+		if _, _, err := exeutils.ExecuteAndPipeStd(stateExe, []string{"activate", params.activateDefault.String(), "--default"}, envSlice(binPath)); err != nil {
 			an.EventWithLabel(AnalyticsFunnelCat, "forward-activate-default-err", err.Error())
 			return errs.Wrap(err, "Could not activate %s, error returned: %s", params.activateDefault.String(), errs.JoinMessage(err))
 		}
+	default:
+		if !isUpdate {
```
Naatan, line 383: I think this would be cleaner as `case !isUpdate`

cmd/state-installer/cmd.go
```diff
@@ -354,20 +375,43 @@ func postInstallEvents(out output.Outputer, an analytics.Dispatcher, params *Par
...
+			ss := subshell.New(cfg)
+			ss.SetEnv(envMap(binPath))
+			if err := ss.Activate(nil, cfg, out); err != nil {
+				return errs.Wrap(err, "Subshell setup; error returned: %s", errs.JoinMessage(err))
+			}
+			ers := ss.Errors()
+			for err = range ers {
+				return errs.Wrap(err, "Subshell execution; error returned: %s", errs.JoinMessage(err))
```
Naatan, line 391: Return in a for loop?

cmd/state-installer/cmd.go
```diff
@@ -354,20 +375,43 @@ func postInstallEvents(out output.Outputer, an analytics.Dispatcher, params *Par
...
 }
 
+func envSlice(binPath string) []string {
+	return []string{"PATH=" + binPath + string(os.PathListSeparator) + os.Getenv("PATH")}
+}
+
+func envMap(binPath string) map[string]string {
+	return map[string]string{
+		"PATH":   binPath + string(os.PathListSeparator) + os.Getenv("PATH"),
+		"TESTER": "TEST",
```
Naatan, line 406: Leftover debugging code?

internal/subshell/bash/bash.go
```diff
@@ -141,18 +141,26 @@ func (v *SubShell) Quote(value string) string {
...
+	if proj != nil {
+		env := sscommon.EscapeEnv(v.env)
+		var err error
+		if v.rcFile, err = sscommon.SetupProjectRcFile(proj, "bashrc.sh", "", env, out, cfg); err != nil {
+			return err
+		}
+
+		shellArgs = append(shellArgs, "--rcfile", v.rcFile.Name())
+	} else {
+		directEnv = sscommon.EnvSlice(v.env)
```
Naatan, line 156: This needs a comment explaining the difference in env handling

test/integration/install_scripts_int_test.go
```diff
@@ -136,12 +124,28 @@ func scriptPath(t *testing.T, targetDir string) string {
...
-
 func expectStateToolInstallation(cp *termtest.ConsoleProcess) {
 	cp.Expect("Preparing Installer for State Tool Package Manager")
 	cp.Expect("Installation Complete", time.Second*40)
 }
 
+func expectStateToolWorks(cp *termtest.ConsoleProcess, activation bool) {
+	if activation {
+		cp.Expect("Creating a Virtual Environment")
+		cp.Expect("Quick Start", time.Second*60)
```
Naatan, line 135: Nit: These don't belong with "expect state tool works". Since you're only using this function in one place I'd just leave the code where it was.

Naatan (changes requested):

cmd/state-installer/cmd.go
daved: I've updated this to handle the channel as it is in the activate runner. I expect subshell modifications will result in a cleaner API so that a channel is not directly exposed. Otherwise, if that is preferred, I'd like to file a ticket to create a helper function in the errs package.

cmd/state-installer/cmd.go
daved: Added as a comment to https://activestatef.atlassian.net/browse/DX-163

Naatan (approved):

PR #1620, opened 2021-11-01
Add headless dimension

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-470" title="DX-470" target="_blank">Sub-taskDX-470</a>  Add a "Headless" dimension (which is true or false) to our events
  </td></table>
  <br />
 

https://activestatef.atlassian.net/browse/DX-470


internal/runners/exec/exec.go
```diff
@@ -77,7 +85,14 @@ func (s *Exec) Run(params *Params, args ...string) error {
 	// Detect target and project dir
 	// If the path passed resolves to a runtime dir (ie. has a runtime marker) then the project is not used
 	if params.Path != "" && runtime.IsRuntimeDir(params.Path) {
-		rtTarget = runtime.NewCustomTarget("", "", "", params.Path, trigger)
+		projectDir = projectFromRuntimeDir(s.cfg, params.Path, storage.CachePath())
```
Naatan, line 88: Nit: `storage.CachePath()` probably doesn't need to be passed as a var.

internal/runners/exec/exec.go
```diff
@@ -173,15 +188,29 @@ func (s *Exec) Run(params *Params, args ...string) error {
 	return s.subshell.Run(sf.Filename(), args[1:]...)
 }
 
-func handleRecursion(env map[string]string, args []string) (error) {
+func projectFromRuntimeDir(cfg Configurable, runtimeDir, cacheDir string) string {
+	projects := cfg.GetStringMapStringSlice(projectfile.LocalProjectsConfigKey)
```
Naatan, line 192: https://github.com/ActiveState/cli/blob/dfa978d6babd5727da8ecc93a37825be15eee536/pkg/projectfile/projectfile.go#L1171

pkg/platform/runtime/target.go
```diff
@@ -70,22 +71,27 @@ func (p *ProjectTarget) Trigger() string {
 	return p.trigger.String()
 }
 
+func (p *ProjectTarget) Headless() string {
```
Naatan, line 74: I don't think the runtime target should have an opinion on how this gets used. So this should just be a bool. We can then convert it to a string at the time that we know what we're using the value for.

Naatan (changes requested):

pkg/platform/runtime/target.go
MDrakos: The only issue with this is the use case here: https://github.com/ActiveState/cli/pull/1620/files#diff-a1ef26e8e5f00016c2a2c76985086840e00bf6a87901d4f77b3d89f9cf3d05d9R89 where we are passing in the empty string as the headless value as we do not actually know if the project is headless or not. We could default to false here as it's very unlikely that the project is headless.

Naatan: Yeah let's just default if we can't know. And by definition if we don't know then it would be headless. ie. not knowing = there's no "head".

Naatan (approved):
Nice work!

PR #1622, opened 2021-11-02
Don't include system path when filtering executors

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-517" title="DX-517" target="_blank">BugDX-517</a>  Executor using system python
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/envdef/environment.go
+++ b/pkg/platform/runtime/envdef/environment.go
@@ -410,10 +410,15 @@ func (ed *EnvironmentDefinition) ExecutableDirs() (ExecutablePaths, error) {
 	if err != nil {
 		return nil, errs.Wrap(err, "Could not get executable paths")
 	}
+	fmt.Println("Exepaths:", exes)
 
 	var dirs ExecutablePaths
 	for _, p := range exes {
-		dirs = append(dirs, filepath.Dir(p))
+		dir := filepath.Dir(p)
+		if funk.Contains(dirs, dir) {
+			continue
+		}
+		dirs = append(dirs, dir)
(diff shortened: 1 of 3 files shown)


pkg/platform/runtime/envdef/environment.go
```diff
@@ -405,6 +405,25 @@ func (ed *EnvironmentDefinition) ExecutablePaths() (ExecutablePaths, error) {
 	return exes, nil
 }
 
+func (ed *EnvironmentDefinition) ExecutableDirs() (ExecutablePaths, error) {
+	exes, err := ed.ExecutablePaths()
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not get executable paths")
+	}
+	fmt.Println("Exepaths:", exes)
```
Naatan, line 413:
```suggestion
```

pkg/platform/runtime/envdef/environment.go
```diff
@@ -405,6 +405,25 @@ func (ed *EnvironmentDefinition) ExecutablePaths() (ExecutablePaths, error) {
...
+
+	var dirs ExecutablePaths
+	for _, p := range exes {
+		dir := filepath.Dir(p)
+		if funk.Contains(dirs, dir) {
+			continue
+		}
```
Naatan, line 420: Let's instead do a `funk.Uniq()` at the end. That'd make only one additional loop rather than having it loop the same entries inside another loop.

Naatan (changes requested):

Naatan (approved):

PR #1623, opened 2021-11-03
Fix environment inheritance not working or containing malformed env values

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-523" title="DX-523" target="_blank">BugDX-523</a>  missing $GOPATH when we run `state build`
  </td></table>
  <br />


MDrakos (approved):
👍

PR #1624, opened 2021-11-03
Fix installer not reporting errors

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-526" title="DX-526" target="_blank">BugDX-526</a>  Installer still not reporting errors
  </td></table>
  <br />
 

The root issue is this:

https://github.com/ActiveState/cli/commit/28a42a28c7fd3835e637a2d6fa43ef77f1829409

But this PR is addressing it in a way where it's less likely to happen in the future.

### DIFF
--- a/internal/constants/preprocess/preprocess.go
+++ b/internal/constants/preprocess/preprocess.go
@@ -16,7 +16,7 @@ import (
 
 // Constants holds constants that will be preprocessed, meaning the key value parts here will be built into the constants
 // package as actual constants, using the build-time interpretations
-var Constants = map[string]func() string{}
+var Constants = map[string]func() interface{}{}
 
 func init() {
 	branchName, commitRef := branchName()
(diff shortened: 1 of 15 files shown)


internal/constants/preprocess/preprocess.go
```diff
@@ -16,7 +16,7 @@ import (
 
 // Constants holds constants that will be preprocessed, meaning the key value parts here will be built into the constants
 // package as actual constants, using the build-time interpretations
-var Constants = map[string]func() string{}
+var Constants = map[string]func() interface{}{}
```
MDrakos, line 19: nit: I'm not sure making the function return type an `interface{}`  is necessary here as the `OnCI` value you added is still a string

MDrakos (approved):

internal/constants/preprocess/preprocess.go
Naatan: Yeah I couldn't get `jen` to actually write it as a boolean :\ But I figure this still gets it a bit closer to accepting mixed types. Can cross the other bridge at a later time if we need to.

PR #1625, opened 2021-11-03
Update uniqid file location on Windows

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-460" title="DX-460" target="_blank">BugDX-460</a>  uniqid file is written to AppData/activestate on Windows
  </td></table>
  <br />
 

On Windows this updates the location of the persist directory from `<AppDataDir>/activestate/persist` to `<AppDataDir>/Local/activestate_persist`

### DIFF
--- a/internal/uniqid/legacy.go
+++ b/internal/uniqid/legacy.go
@@ -1,0 +1,50 @@
+package uniqid
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/ActiveState/cli/internal/errs"
+)
+
+const legacyPersistDir = "activestate/persist"
+
+func moveLegacyFile(destination string) error {
+	legacyDir, err := legacyStorageDir()
+	if err != nil {
+		return errs.Wrap(err, "Could not get legacy storage directory")
+	}
+
+	// If the legacy file does not not exist there is nothing to move
+	if !fileExists(filepath.Join(legacyDir, fileName)) {
+		return nil
+	}
+
+	destinationDir := filepath.Dir(destination)
+	err = mkdir(destinationDir)
+	if err != nil {
+		return errs.Wrap(err, "Could not create new persist directory")
+	}
+
+	err = moveAllFiles(legacyDir, destinationDir)
+	if err != nil {
+		return errs.Wrap(err, "Could not move legacy uniqid file")
+	}
+
+	// The legacy directory is a sub directory, we want to remove the parent
+	err = os.RemoveAll(filepath.Dir(legacyDir))
+	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/uniqid/legacy.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil
+	}
+
+	destinationDir := filepath.Dir(destination)
+	err = mkdir(destinationDir)
+	if err != nil {
+		return errs.Wrap(err, "Could not create new persist directory")
+	}
+
+	err = moveAllFiles(legacyDir, destinationDir)
```
Naatan, line 29: This seems like overkill since there's only one file to be moved.

internal/uniqid/legacy.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return errs.Wrap(err, "Could not move legacy uniqid file")
+	}
+
+	// The legacy directory is a sub directory, we want to remove the parent
+	err = os.RemoveAll(filepath.Dir(legacyDir))
```
Naatan, line 35: This should also just remove the one file, and the folder itself. If there are other files there then we don't "own" those so we should probably error out anyway.

Naatan (changes requested):

internal/uniqid/legacy.go
MDrakos: If we remove just the directory containing the uniqID file then the `activestate` directory will still exist at the root of `AppData`. I can add code to check that the `activestate` directory contains only one subdirectory and that directory only contains one file, but that feels like overkill considering we're the ones who own this directory and this is the only package that's writing to it.

I could also be misunderstanding what you're asking for 😛

Naatan: I would just do a remove of the uniqid file itself, followed by a removal of the dir. If either failed; ignore it. I see this as us doing minimal effort to retain the ID. If it failed because of permissions or something then we'll just end up with a new ID.

internal/uniqid/legacy.go
```diff
@@ -0,0 +1,51 @@
...
+	if !fileExists(legacyUniqIDFile) {
+		return nil
+	}
+
+	err = mkdir(filepath.Dir(destination))
+	if err != nil {
+		return errs.Wrap(err, "Could not create new persist directory")
+	}
+
+	err = os.Rename(legacyUniqIDFile, destination)
```
Naatan, line 30: Might be safer to do a copy, in case we don't have write permission to the file for some reason.

Naatan (changes requested):
Just wanna get rid of the RemoveAll. See my latest response.

Naatan (approved):

PR #1629, opened 2021-11-08
Ensure project dimension is accurate

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-468" title="DX-468" target="_blank">Sub-taskDX-468</a>  Ensure project is always accurate with runtime events
  </td></table>
  <br />
 

I _think_ this covers the AC in the ticket. There is one instance where we create a custom target where we do not have the project information. Other than that, including the project namespace in the `runtime-start`  event should be accurate.


Naatan (approved):
Clean and simple :)

PR #1632, opened 2021-11-09
Runtime heartbeat event

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-488" title="DX-488" target="_blank">StoryDX-488</a>  Runtime usage is tracked with heartbeats
  </td></table>
  <br />
 

https://activestatef.atlassian.net/browse/DX-488


pkg/platform/runtime/target/target.go
```diff
@@ -33,6 +34,13 @@ const (
 	triggerUnknown  Trigger = "unknown"
 )
 
+// usageTriggers are triggers that indicate actual usage of the runtime (as oppose to simply making changes to the runtime)
+var usageTriggers = []Trigger{TriggerActivate, TriggerScript, TriggerDeploy, TriggerExec}
```
MDrakos, line 38: If `TriggerDeploy` is a usage trigger should `TriggerDefault` be as well? I could also argue to include default as it's similar to activate

pkg/platform/runtime/runtime.go
```diff
@@ -12,23 +12,28 @@ import (
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/installation/storage"
+	"github.com/ActiveState/cli/internal/instanceid"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/rtutils/p"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
+	model2 "github.com/ActiveState/cli/pkg/platform/model"
```
MDrakos, line 21: nit: rename this import to `platformModel` or something similar

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -0,0 +1,116 @@
...
+	an       *sync.Client
+	cfg      *config.Instance
+	watching []entry
+	stop     chan struct{}
+}
+
+func New(cfg *config.Instance, an *sync.Client) *Watcher {
+	w := &Watcher{an: an, stop: make(chan struct{}), cfg: cfg}
+
+	if watchersJson := w.cfg.GetString(CfgKey); watchersJson != "" {
```
MDrakos, line 33: I'm not sure if we care enough but if a savvy user cleaned their config and restarted the state service we would not be able to track the heartbeat anymore. This could also potentially happen if something went wrong with an update. While it wouldn't provide us _much_ more protection we could use the `persist` directory that contains the `uniqid` file. At least this way we aren't providing users with a path to undermine the heartbeat tracking.

Might be better for a follow-up story though.

test/integration/analytics_int_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/ActiveState/cli/internal/analytics/client/sync/reporters"
+	"github.com/ActiveState/cli/internal/analytics/constants"
+	constants2 "github.com/ActiveState/cli/internal/constants"
```
MDrakos, line 13: nit: naming

MDrakos (changes requested):
Looks good to me! Just found a few small things and some nits

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -141,3 +148,14 @@ func (r *Resolver) AnalyticsEvent(_ context.Context, category, action string, _l
 
 	return &graph.AnalyticsEventResponse{Sent: true}, nil
 }
+
+func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimensionsJSON string) (*graph.RuntimeUsageResponse, error) {
+	var dims *dimensions.Values
+	if err := json.Unmarshal([]byte(dimensionsJSON), &dims); err != nil {
+		return &graph.RuntimeUsageResponse{Received: false}, errs.Wrap(err, "Could not unmarshal")
```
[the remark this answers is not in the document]
Naatan: This is wrapping the error. The context of what it couldn't unmarshal would come from the unmarshal function itself, and the context of where it happened would come from the place that handles this error.

internal/testhelpers/e2e/dirs.go
```diff
@@ -56,6 +60,15 @@ func NewDirs(base string) (*Dirs, error) {
 	return &dirs, nil
 }
 
+func (d *Dirs) InheritGlobalCache() {
+	globalCachePath := storage.CachePath()
+	if fileutils.FileExists(globalCachePath) {
+		fmt.Println("Warming cache")
+		fileutils.CopyFiles(globalCachePath, d.Cache)
+		fmt.Println("Cache warmed")
```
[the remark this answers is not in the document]
Naatan: No, this is test-only code. And since this could be slow I figured I'd add some prints so you'd know why it's hanging.

pkg/platform/runtime/target/target.go
Naatan: I didn't use it because looking at the code it doesn't seem to be about default at all. Though I should've renamed it while I was at it. Will do that now.

cmd/state-svc/internal/rtwatcher/watcher.go
Naatan: Not worth it. The runtime tracking is a best effort kind of thing. There's lots of things people could do to circumvent it or limit its use, but we care about the 99% of the users who don't mess about rather than the 1% that do.

MDrakos: Fair enough. Though if I recall this data is going to be used to help determine billing? Or what tier a user/org should be on. If that's the case this may be something we have to revisit.

MDrakos (approved):

cmd/state-svc/internal/rtwatcher/watcher.go
Naatan: Yeah but the data is only used to manually determine the bill. I gave the same concerns you did when we discussed this, the above take was effectively the answer I received.

PR #1633, opened 2021-11-10
Search for requested pkg before running pkg op

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-500" title="DX-500" target="_blank">StoryDX-500</a>  State Tool uses search endpoint to validate names before solving in state install
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/packages.go
+++ b/internal/runners/packages/packages.go
@@ -15,7 +15,6 @@ import (
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/prompt"
 	"github.com/ActiveState/cli/internal/runbits"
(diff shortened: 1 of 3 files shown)


Naatan (changes requested):
We should've been more specific in the AC's, your story is to replace the code here:

https://github.com/ActiveState/cli/blob/7f5510fc8dd6271636adfd0a5bb7bfa5bf2dad21/internal/runners/packages/packages.go#L133

Not to introduce new functionality. The functionality already exists, it just needs to use a different method for resolving the ingredient.

Naatan:
Closing this PR in favour of https://github.com/ActiveState/cli/pull/1638

PR #1634, opened 2021-11-10
Pack binaries with UPX

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-453" title="DX-453" target="_blank">StoryDX-453</a>  Shrink our binaries with UPX
  </td></table>
  <br />


activestate.ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - name: ci-build
+    language: bash
+    description: Builds the project with the host OS as the target OS.
+    value: |
+      set -e
+      $constants.SET_ENV
+      export CGO_ENABLED=0
+
+      go build -tags "$GO_BUILD_TAGS" -o $BUILD_TARGET_DIR/$constants.BUILD_TARGET $constants.CLI_BUILDFLAGS $constants.CLI_PKGS
+      $UPX_TARGET $BUILD_TARGET_DIR/$constants.BUILD_TARGET
```
Naatan, line 21: Rather than introducing what amounts to duplicates of the existing scripts I think we can reduce the maintainability headache by modifying the existing scripts with something along the lines of:

```
if [ $CI != "" ]; then
  $UPX_TARGET $BUILD_TARGET_DIR/$constants.BUILD_TARGET
fi
```

Naatan (changes requested):

activestate.yaml
```diff
@@ -51,6 +51,15 @@ constants:
...
     value: .sh
+  - name: UPX_TARGET
+    if: eq .OS.Name "Windows"
+    value: ./.github/deps/Windows/bin/upx.exe
+  - name: UPX_TARGET
+    if: eq .OS.Name "MacOS"
+    value: ./.github/deps/macOS/bin/upx
+  - name: UPX_TARGET
+    if: eq .OS.Name "Linux"
+    value: ./.github/deps/Linux/bin/upx
```
Naatan, line 62: This should not be necessary due to http://github.com/ActiveState/cli/blob/8de7f9a720b1db6ded36740da29e39e8ccaa4380/.github/workflows/build.yml#L83-L83

Naatan (changes requested):

Naatan (approved):

PR #1636, opened 2021-11-16
Update push branch check

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-260" title="DX-260" target="_blank">BugDX-260</a>  Creating fork with `state push` fails when source is custom branch
  </td></table>
  <br />
 

Judging from the ticket I think this is the behaviour we want.

1. User is on a non-default branch of a project
2. `state push <new/namespace>`
3. state push creates a new project. The new project's default branch has the latest commit from (1)

### DIFF
--- a/test/integration/pull_int_test.go
+++ b/test/integration/pull_int_test.go
@@ -81,7 +81,7 @@ func (suite *PullIntegrationTestSuite) TestPull_Merge() {
 	ts.LoginAsPersistentUser()
 
 	cp := ts.SpawnWithOpts(e2e.WithArgs("push"), e2e.WithWorkDirectory(wd))
-	cp.ExpectLongString("Your project has new commits available")
+	cp.ExpectLongString("Your project has new changes available")
 	cp.ExpectExitCode(1)
 
 	cp = ts.SpawnWithOpts(e2e.WithArgs("pull"), e2e.WithWorkDirectory(wd))
--- a/test/integration/pull_int_test.go
+++ b/test/integration/pull_int_test.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"fmt"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


Naatan (commented):
-removed-

Naatan (changes requested):
This feels a little error prone. Basically we have two cases state push addresses:

- I'm pushing a change to an existing project
- I'm pushing to an all-new project

The changes here are addressing the second use-case, but implicitly. Meaning it's not clear from the code that this is why it's what it is doing. I also feel like assuming "branch doesn't exist means use the default" is error-prone for the first use-case.

I think the real issue here might be that when we create the project we create it with the default branch, rather than the one that was requested.

test/integration/pull_int_test.go
```diff
@@ -81,20 +85,16 @@ func (suite *PullIntegrationTestSuite) TestPull_Merge() {
...
 	cp.ExpectExitCode(0)
 
 	cp = ts.SpawnCmd("bash", "-c", fmt.Sprintf("cd %s && %s history | head -n 10", wd, ts.ExecutablePath()))
-	cp.ExpectLongString(locale.T("pull_merge_commit"))
+	cp.ExpectLongString("Merged")
 	cp.ExpectExitCode(0)
-
-	cp = ts.SpawnWithOpts(e2e.WithArgs("push"), e2e.WithWorkDirectory(wd))
-	cp.ExpectLongString("You do not have permission") // This tells us the history is valid, we don't need an actual successful push
-	cp.ExpectExitCode(1)
```
MDrakos, line 97: I'm not sure how this worked before... we've just pushed and it worked, we push again and expect it to fail?

Naatan (approved):

PR #1639, opened 2021-11-16
input errors are more friendly

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-502" title="DX-502" target="_blank">StoryDX-502</a>  State install error for incorrect package names are more friendly
  </td></table>
  <br />

### DIFF
--- a/internal/locale/errors.go
+++ b/internal/locale/errors.go
@@ -135,6 +135,18 @@ func IsInputError(err error) bool {
 	return false
 }
 
+// IsInputError checks if the given error contains a InputError anywhere in the unwrap stack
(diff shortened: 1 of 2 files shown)


internal/locale/errors.go
```diff
@@ -135,6 +135,18 @@ func IsInputError(err error) bool {
 	return false
 }
 
+// IsInputError checks if the given error contains a InputError anywhere in the unwrap stack
```
MDrakos, line 138:
```suggestion
// IsInputErrorNonRecursive checks if the given error is an InputError
```

MDrakos (approved):

PR #1640, opened 2021-11-16
Regenerate mono and inventory API

antoine-activestate:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-511" title="DX-511" target="_blank">StoryDX-511</a>  Generate API client to make Org.Name optional
  </td></table>
  <br />


antoine-activestate:
Need to reopen with simpler name

PR #1642, opened 2021-11-17
Handle spaces in cache dir (executors)

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-438" title="DX-438" target="_blank">BugDX-438</a>  Executors fail when paths have spaces in them on Windows
  </td></table>
  <br />

### DIFF
--- a/test/integration/exec_int_test.go
+++ b/test/integration/exec_int_test.go
@@ -185,7 +185,8 @@ func (suite *ExecIntegrationTestSuite) TestExec_SpaceInCacheDir() {
 	cp.SendLine("python3 --version")
 	cp.Expect("Python 3.")
 	if runtime.GOOS == "linux" {
-		cp.SendLine("env | grep ACTIVE")
+		cp.SendLine(fmt.Sprintf(`ls -l "${%s}"`, constants.CacheEnvVarName))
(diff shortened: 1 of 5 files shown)


test/integration/exec_int_test.go
```diff
@@ -165,6 +166,28 @@ echo "Hello $name!"
...
+		e2e.AppendEnv(fmt.Sprintf(`%s=""`, constants.DisableRuntime)),
+		e2e.WithArgs("activate", "ActiveState-CLI/Python3"),
+	)
+
+	cp.SendLine("python3 --version")
+	cp.Expect("Python 3.")
+
+	cp.SendLine("exit")
+	cp.ExpectExitCode(0)
+}
```
Naatan, line 189: This feels like it belongs in activate_int_test.

Naatan (changes requested):

test/integration/exec_int_test.go
daved: I thought about that, but the focus is particularly on the effectiveness of the executor .sh and .bat files that are used within an activated state.

Naatan: `exec_int_test` focusses on `state exec`. Not to be confused with executors that use `state exec` under the hood. The latter is already covered in activate_int_test.

Naatan (approved):

PR #1643, opened 2021-11-17
Show install and activate progress around recipe solves

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-501" title="DX-501" target="_blank">StoryDX-501</a>  State install has progress animations and status messages during long non-interactive phases (solving, name resolution, etc.)
  </td></table>
  <br />


cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -29,7 +29,7 @@ type Watcher struct {
 }
 
 func New(cfg *config.Instance, an *sync.Client) *Watcher {
-	w := &Watcher{an: an, stop: make(chan struct{}), cfg: cfg, interval: defaultInterval}
+	w := &Watcher{an: an, stop: make(chan struct{}, 1), cfg: cfg, interval: defaultInterval}
```
MDrakos, line 32: nit: Is this buffer on the channel necessary? We could potentially remove the buffer and simply close the channel in the `Close()` function. Same for the `stop` channel in the `DotProgress`

internal/runbits/progressbar/decor.go
```diff
@@ -18,6 +18,20 @@ func (pb *RuntimeProgress) addTotalBar(name string, total int64) *mpb.Bar {
 	return pb.addBar(name, total, false, mpb.BarFillerClearOnComplete())
 }
 
+// addTotalBar adds a bar counting a number of sub-events adding up to total
```
MDrakos, line 21:
```suggestion
// addSpinnerBar adds a bar with a spinning progress indicator
```

internal/runners/packages/packages.go
```diff
@@ -80,15 +95,40 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
...
+		if len(packages) == 0 {
+			suggestions, err := getSuggestions(ns, packageName)
+			if err != nil {
+				logging.Error("Failed to retrieve suggestions: %v", err)
+			}
+			if len(suggestions) == 0 {
+				return locale.WrapInputError(err, "package_ingredient_alternatives_nosuggest", "", packageName)
+			}
+			return locale.WrapInputError(err, "package_ingredient_alternatives", "", packageName, strings.Join(suggestions, "\n"))
+		}
```
MDrakos, line 122: Isn't this https://activestatef.atlassian.net/browse/DX-500?

MDrakos (approved):
Looks good to me. Just nits and docstring updates

cmd/state-svc/internal/rtwatcher/watcher.go
Naatan: It's effectively a failsafe. I don't think it's necessary with how we're using this channel, but it's better to be explicit.

internal/runners/packages/packages.go
Naatan: It is. I didn't introduce this code in this PR, I just moved it.

PR #1655, opened 2021-11-24
Implemented pip and PPM shim as activestate.yaml script

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-416" title="DX-416" target="_blank">StoryDX-416</a>  ppm shim is updated
  </td></table>
  <br />


MDrakos (approved):
Looks good 👍 I like this approach over what we had previously.

rawktron (approved):
It's fine. I think we really need to see some usage data before knowing what to tweak. It's a reasonable place to start.

PR #1667, opened 2021-12-10
Add category and action info to GA reporter error logging

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-588" title="DX-588" target="_blank">BugDX-588</a>  Could not send GA Event: unexpected EOF: unexpected EOF
  </td></table>
  <br />


Naatan (changes requested):
This doesn't address the following AC:

* Underlying GA error (eg. request body too big)

You were talking about identifying things like the above (request body too big). Can we get that into the code so we can tell it from the rollbar report?

daved:
The underlying GA error already comes through. The problem is that it's not always a GA error that is the cause. Should I compile those which do provide a clear GA error?

Naatan (approved):

PR #1670, opened 2021-12-13
Trial go111module envvar fix for 'go install' use in asy

daved:
hotfix for failing deploy job on master
@Naatan I'm not sure this will work, but it's my best guess at the moment.


Naatan (approved):

daved:
DX-596 has been added to fix CI and remove this change.

PR #1671, opened 2021-12-13
Added Ruby support

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-592" title="DX-592" target="_blank">StoryDX-592</a>  Ruby supports activate and install
  </td></table>
  <br />

### DIFF
--- a/internal/language/language.go
+++ b/internal/language/language.go
@@ -23,6 +23,7 @@ const (
 	Perl
 	Python3
 	Python2
+	Ruby
 )
 
 // UnrecognizedLanguageError contains info related to the usage of an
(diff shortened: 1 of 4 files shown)


internal/language/language.go
```diff
@@ -97,6 +98,10 @@ var lookup = [...]languageData{
 		"python2", "Python 2", ".py", true, "python", "2.7.14",
 		Executable{constants.ActivePython2Executable, false},
 	},
+	{
+		"ruby", "Ruby", ".rb", true, "ruby", "3.0.1",
+		Executable{constants.RubyExecutable, false},
```
daved, line 103: It might be worth checking if the default languages info endpoint in the mediator returns this info as well - this data is currently mostly a fallback.

daved (approved):

internal/language/language.go
Naatan: That's only used when initializing a project. In my tests it wasn't an issue, and this story only covers activate/install.

PR #1672, opened 2021-12-13
Always use go get in install-deps script

daved:
@Naatan The temporary directory juggling here is due to how `go get` works with modules and alters the go.mod and go.sum files.

As of go1.16, `go install` is the recommended way to build and install commands. That would remove the need for this change. See: https://activestatef.atlassian.net/browse/DX-596


daved:
Resolved by commit 3a0af5c.

PR #1677, opened 2021-12-24
Activating when already in an activated state does not error

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-414" title="DX-414" target="_blank">StoryDX-414</a>  State Activate when already activated (same project) doesn't error
  </td></table>
  <br />


internal/runners/activate/activate.go
```diff
@@ -336,3 +347,12 @@ func warningForAdministrator(out output.Outputer) {
 		))
 	}
 }
+
+func activatedProjectName() (string, error) {
```
Naatan, line 351: Nit: I think it would be more intuitive if this was called something like `parentNamespace()` and returned the namespace type rather than a string.

Naatan (dismissed):
Looks good! This is making me realize we have to get rid of the project env var we are currently using though, as it feels very error-prone. Maybe in the future we can have the state-svc provide a mapping of PIDs and their associated projects. That's just thoughts for the future though. Great work!

MDrakos:
> Maybe in the future we can have the state-svc provide a mapping of PIDs and their associated projects. That's just thoughts for the future though.

I like this idea. Filed a story so we don't lose it https://activestatef.atlassian.net/browse/DX-621

PR #1682, opened 2022-01-06
Reduce GA and rollbar events

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-615" title="DX-615" target="_blank">StoryDX-615</a>  Audit rollbar error reporting
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -118,8 +118,6 @@ func main() {
 	an = sync.New(cfg, nil)
 	an.Event(AnalyticsFunnelCat, "start")
 
-	logging.SetupRollbarReporter(func(msg string) { an.Event("rollbar", msg) })
-
 	params := newParams()
 	cmd := captain.NewCommand(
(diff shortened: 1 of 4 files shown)


cmd/state-installer/cmd.go
```diff
@@ -118,8 +118,6 @@ func main() {
 	an = sync.New(cfg, nil)
 	an.Event(AnalyticsFunnelCat, "start")
 
-	logging.SetupRollbarReporter(func(msg string) { an.Event("rollbar", msg) })
```
MDrakos, line 121: This is only in the installer. Do we also want to disable the reporting in the state tool and state service?

MDrakos (approved):

cmd/state-installer/cmd.go
Naatan: You just saved me a huge headache 👍

MDrakos (approved):

PR #1685, opened 2022-01-07
Logging calls do not block

MDrakos:


internal/logging/file.go
```diff
@@ -0,0 +1,210 @@
...
+	handler := fileHandler{
+		DefaultFormatter,
+		nil,
+		sync.Mutex{},
+		safeBool{},
+		&sync.WaitGroup{},
+		make(chan entry, defaultMaxEntries),
+		make(chan struct{}),
+	}
+	go handler.start()
```
daved, line 45:
```suggestion
	handler.wg.Add(1)
	go func() {
		defer handler.wg.Done()
		handler.start()
	}()
```

internal/logging/file.go
```diff
@@ -0,0 +1,210 @@
...
+	return &handler
+}
+
+func (l *fileHandler) start() {
+	for {
+		select {
+		case entry := <-l.queue:
+			l.emit(entry.ctx, entry.message, entry.args)
+			l.wg.Done()
```
daved, line 54:
```suggestion
```

internal/logging/file.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return fmt.Errorf("Could not open log file for writing: %s: %w", filename, err)
+	}
+	l.file = f
+	return nil
+}
+
+func (l *fileHandler) Close() {
+	defer close(l.quit)
+	defer close(l.queue)
```
daved, line 208:
```suggestion
	close(l.quit)
```

internal/logging/file.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (l *fileHandler) start() {
+	for {
+		select {
+		case entry := <-l.queue:
+			l.emit(entry.ctx, entry.message, entry.args)
+			l.wg.Done()
+		case <-l.quit:
+			return
```
daved, line 56:
```suggestion
			close(l.queue)
			for range l.queue {
			}
			return
```

daved (changes requested):
The suggested changes will ensure that the channel is emptied and that any in-progress logging is waited for. The current design looks like it will not necessarily handle buffered data. Apologies if I misunderstood the intent here. I do wonder if a different approach could be taken, but I wanted to keep the changes as close to the original as possible, so I didn't get creative.

internal/logging/file.go
daved: I'm sending mixed messages here, sorry. If you want to send all the messages before collapsing, the for loop needs to call `l.emit`. The above will just drain cleanly and then collapse.

internal/logging/file.go
```diff
@@ -0,0 +1,210 @@
...
+func (l *fileHandler) Emit(ctx *MessageContext, message string, args ...interface{}) error {
+	var a []interface{}
+	a = append(a, args)
+	e := entry{
+		ctx:     ctx,
+		message: message,
+		args:    a,
+	}
+	l.wg.Add(1)
+	l.queue <- e
```
daved, line 82:
```suggestion
	select {
	case <-l.quit:
		return nil
	default: 
		l.queue <- e
	}
```

internal/logging/file.go
MDrakos: Yeah, I was missing this part in my original changeset. I've added an `l.emit` call to the for loop. Thanks for the suggestion!

daved (approved):

internal/logging/file.go
```diff
@@ -0,0 +1,220 @@
...
+	return nil
+}
+
+func (l *fileHandler) emit(ctx *MessageContext, message string, args ...interface{}) {
+	defer handlePanics(recover())
+	// In this function we close and open the file handle to the log file. In
+	// order to ensure this is safe to be called across threads, we just
+	// synchronize the entire function
+	l.mu.Lock()
+	defer l.mu.Unlock()
```
Naatan, line 103: This doesn't seem necessary anymore since we're processing them on a queue. @daved can you confirm my assumption?

internal/logging/file.go
```diff
@@ -0,0 +1,220 @@
...
+	case <-l.quit:
+		return nil
+	default:
+		l.queue <- e
+	}
+	return nil
+}
+
+func (l *fileHandler) emit(ctx *MessageContext, message string, args ...interface{}) {
+	defer handlePanics(recover())
```
Naatan, line 98: This should probably move to the `start()` function.

Naatan (changes requested):

cmd/state/main.go
```diff
@@ -60,7 +60,7 @@ func main() {
 		}
 
 		// ensure rollbar messages are called
-		if err := events.WaitForEvents(5*time.Second, rollbar.Wait, rollbar.Close, authentication.LegacyClose); err != nil {
+		if err := events.WaitForEvents(5*time.Second, rollbar.Wait, rollbar.Close, authentication.LegacyClose, logging.Close); err != nil {
```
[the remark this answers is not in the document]
MDrakos: If we didn't close logging I think we would just potentially lose log messages as the goroutine that processes entries would be killed. I don't think there would be any locking up of the state tool as the `Wait` call is in the `Close` function

internal/logging/file.go
daved: I think it's not needed for the current design, but I take the file sync to be distinct from the log emission synchronization. Also, I'd like to see a test to be sure if race detection or output verification (less likely) show that the mutex is needed or not. If so, the mutex hold-time should be decreased (likely by wrapping the file in a struct to wrap the Write/Read methods so that the mutex is held as little as possible). Maybe it should be left for now and a follow-up story filed.

MDrakos: Added a unit test and ran it with the `-race` flag successfully.

internal/logging/file_test.go
```diff
@@ -0,0 +1,31 @@
...
+	loggingFile, err := ioutil.TempFile("", "")
+	assert.NoError(t, err)
+	fh.file = loggingFile
+
+	for i, message := range messages {
+		t.Run(fmt.Sprintf("Parallel emit %d", i), func(t *testing.T) {
+			t.Parallel()
+			emitErr := fh.Emit(&MessageContext{}, message)
+			assert.NoError(t, emitErr)
+		})
```
daved, line 29: nit: It might be worth putting this in another loop (maybe 10 count) to ensure that the window of time that the detector is sensitive to is landed on.

daved (approved):

Naatan (approved):

daved (approved):

PR #1686, opened 2022-01-10
Dx 479

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-479" title="DX-479" target="_blank">StoryDX-479</a>  sysinfo employs caching
  </td></table>
  <br />
 

Use caching for sysinfo package.

### DIFF
--- a/pkg/sysinfo/sysinfo_darwin.go
+++ b/pkg/sysinfo/sysinfo_darwin.go
@@ -6,7 +6,11 @@ import (
 	"os/exec"
 	"regexp"
 	"strconv"
+
+	"github.com/patrickmn/go-cache"
 )
+
+var sysinfoCache *cache.Cache = cache.New(cache.NoExpiration, cache.NoExpiration)
 
 // OS returns the system's OS
 func OS() OsInfo {
@@ -19,6 +23,11 @@ var (
 
 // OSVersion returns the system's OS version.
 func OSVersion() (*OSVersionInfo, error) {
+	cached, found := sysinfoCache.Get(osVersionInfoCacheKey)
+	if found {
+		return cached.(*OSVersionInfo), nil
+	}
+
 	// Fetch OS version.
 	version, err := getDarwinProductVersion()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/sysinfo/sysinfo_darwin.go
```diff
@@ -19,6 +23,11 @@ var (
 
 // OSVersion returns the system's OS version.
 func OSVersion() (*OSVersionInfo, error) {
+	cached, found := sysinfoCache.Get(osVersionInfoCacheKey)
+	if found {
```
Naatan, line 27:
```suggestion
	if cached, found := sysinfoCache.Get(osVersionInfoCacheKey); found {
```

Fwiw; this is the shorthand version. Not asking for change so much as I'm sharing for future reference, should you want to use it.

pkg/sysinfo/sysinfo_darwin.go
```diff
@@ -6,8 +6,12 @@ import (
 	"os/exec"
 	"regexp"
 	"strconv"
+
+	"github.com/patrickmn/go-cache"
 )
 
+var sysinfoCache *cache.Cache = cache.New(cache.NoExpiration, cache.NoExpiration)
```
Naatan, line 13: If you set this in sysinfo.go then you won't have to define it 3 times.

Naatan (approved):
Looking good! Only had two potential tips to share.

pkg/sysinfo/sysinfo_darwin.go
mitchell-as: As I was writing, I was wondering if Go had this feature. I'm glad it does :) I'll make the change because it'll get me in the mindset of doing this.

pkg/sysinfo/sysinfo_darwin.go
mitchell-as: Good point. Will do.

Naatan (approved):

PR #1691, opened 2022-01-11
Log label if shimming

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-628" title="DX-628" target="_blank">StoryDX-628</a>  We log the label for `state run ppm` and `state run pip`
  </td></table>
  <br />

### DIFF
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -40,6 +40,11 @@ var appEventPrefix string = func() string {
 
 var cobraMapping map[*cobra.Command]*Command = make(map[*cobra.Command]*Command)
 
+const (
+	ppmShim = "ppm"
+	pipShim = "pip"
+)
+
 type cobraCommander interface {
 	GetCobraCmd() *cobra.Command
 }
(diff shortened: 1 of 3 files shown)


internal/captain/command.go
```diff
@@ -502,11 +502,16 @@ func (c *Command) runner(cobraCmd *cobra.Command, args []string) error {
...
 	subCommandString := c.UseFull()
-	logging.CurrentCmd = appEventPrefix+subCommandString
+	logging.CurrentCmd = appEventPrefix + subCommandString
 
-	// Send  GA events unless they are handled in the runners...
+	// Send GA events unless they are handled in the runners...
 	if c.analytics != nil {
-		c.analytics.Event(anaConsts.CatRunCmd, appEventPrefix+subCommandString)
+		var label string
+		if len(args) > 0 && (args[0] == constants.PpmShim || args[0] == constants.PipShim) {
```
Naatan, line 510: This doesn't seem right, it would run as `state run ppm` I believe, making it `args[2]`. Could you manually test and ensure this is correct?

Naatan (changes requested):

internal/captain/command.go
MDrakos: Yep, I manually tested this before opening the PR. I think the first argument is being consumed to determine what the command is. Added an integration test so we can verify.

Naatan (approved):

PR #1692, opened 2022-01-12
Update projectfile commit after refresh

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-589" title="DX-589" target="_blank">StoryDX-589</a>  `state install` does not commit until it solves
  </td></table>
  <br />


internal/runners/packages/packages.go
```diff
@@ -292,6 +292,7 @@ func initializeProject() (*project.Project, error) {
 	}
 
 	_, err = projectfile.Create(createParams)
+	fmt.Println("Create error:", errs.JoinMessage(err))
```
Naatan, line 295: ^

pkg/projectfile/projectfile.go
```diff
@@ -1005,7 +1005,7 @@ func createCustom(params *CreateParams, lang language.Language) (*Project, error
 	}
 
 	content := params.Content
-	if content == "" {
+	if content == "" && lang.String() != "" {
```
Naatan, line 1008: This check didn't exist before I think? Fact is the asset change broke this, I'd rather understand why it broke than put a bandaid on the symptom.

Totally fair to descope that from this PR though.

Naatan (changes requested):

pkg/projectfile/projectfile.go
MDrakos: Filed story here: https://activestatef.atlassian.net/browse/DX-654. 

Do we want to keep this band-aid in the code for now or remove it and properly address it with the bug? From my short investigation into the issue this band-aid may end up being the solution. `packr` was likely also not finding the asset file, it just wasn't erroring either.

mitchell-as: A better check would be to see if lang is `Unknown`. The `Create()` function calls `language.MakeByName`, which can return `Unknown`.

MDrakos: I tried that earlier and it did not work. However, I just realized that it didn't work because the language is not `Unknown` but `Unset`. Updated the check. Thanks!

pkg/projectfile/projectfile.go
```diff
@@ -1005,7 +1005,7 @@ func createCustom(params *CreateParams, lang language.Language) (*Project, error
 	}
 
 	content := params.Content
-	if content == "" && lang.String() != "" {
+	if content == "" && lang != language.Unset {
```
Naatan, line 1008: The old version seems less error-prone to me. My main concern with raising this code as is that we understand why it broke in the first place. I have no issues with the original code in and of itself.

Case in point; what if in the future we introduce a regression of some sort where it might send a `language.Unknown`.

Naatan (changes requested):

pkg/projectfile/projectfile.go
mitchell-as: Can we create a unit test that would catch a regression? (I don't know if this is possible/feasible/reasonable. It's simply a question.)

MDrakos: I'm not sure the unit test would capture a regression here as the language would be coming from outside of the package.

I've reverted the change to the original. There is a story filed to deal with this bug so if there is a better way we can handle it there. This should at least ensure the affected workflow isn't "broken"

Naatan (approved):

PR #1693, opened 2022-01-12
Initial attempt to replace packr with go:embed.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1" title="DX-1" target="_blank">StoryDX-1</a>  Replace packr with go embed
  </td></table>
  <br />
 

Moved shared assets into common package and moved locale files into locale package.


internal/locale/locale.go
```diff
@@ -73,10 +79,10 @@ func init() {
 }
 
 // getLocalePath exists to facilitate running Go test scripts from their sub-directories, if no tests are being ran
-// this just returns `locale/`
+// this just returns `internal/locale/`
 func getLocalePath() string {
 	pathsep := string(os.PathSeparator)
-	path := "locale" + pathsep
+	path := "internal/locale" + pathsep
```
daved, line 85: Please consider
```suggestion
	path := filepath.Join("internal","locale") + pathsep
```
or something like
```suggestion
	path := fmt.Sprintf("internal%[1]slocale%[1]s", pathsep)
```
Don't apply these direclty. You'll need to ensure that the imports are included and the code works correctly.

An alternative is to make the value work without the trailing slash. It's somewhat odd as-is, IMO.

daved (changes requested):

daved (approved):

PR #1696, opened 2022-01-15
Ensure service is stopped

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-604" title="DX-604" target="_blank">StoryDX-604</a>  We have a better method of stopping state-svc on update
  </td></table>
  <br />


cmd/state-svc/manager.go
```diff
@@ -130,11 +133,13 @@ func (s *serviceManager) CheckPid(pid int) (*int, error) {
...
-	exe, err := p.Exe()
-	if err != nil {
-		logging.Error("Could not detect executable for pid, error: %s", errs.JoinMessage(err))
-	} else if filepath.Clean(exe) != filepath.Clean(osutils.Executable()) {
-		return nil, nil
+	if runtime.GOOS == "windows" {
+		exe, err := p.Exe()
+		if err != nil {
+			logging.Error("Could not detect executable for pid, error: %s", errs.JoinMessage(err))
+		} else if filepath.Clean(exe) != filepath.Clean(osutils.Executable()) {
```
Naatan, line 140: Let's check the basename rather than the full path. Since we will also have people stranded on the old version where state-svc was at a different location.

Naatan:
Tagging @daved for review as well, as this ones a big deal and I want to make sure we give this our best effort in terms of avoiding new issues.

internal/installation/stop.go
```diff
@@ -35,5 +60,53 @@ func StopRunning(installPath string) (rerr error) {
...
+	// no running state-svc processes which could lead to
+	// errors when updating
+	for _, p := range procs {
+		n, err := p.Name()
+		if err != nil {
+			logging.Error("Could not get process name: %v", err)
+			continue
+		}
+
+		if n == constants.ServiceCommandName {
```
Naatan, line 78: Can we get the full path here? I know it's very unlikely that there are other processes named `state-svc`, but it would be nice if we could have an extra safety check here. eg. check that `activestate` is in the path.

internal/installation/stop.go
```diff
@@ -15,15 +22,33 @@ func StopRunning(installPath string) (rerr error) {
...
-	svcInfo := appinfo.SvcApp(installPath)
+
+	err = stopTray(installPath, cfg)
+	if err != nil {
+		return errs.Wrap(err, "Could not stop tray")
+	}
+
+	err = stopSvc(installPath)
+	if err != nil {
+		return errs.Wrap(err, "Could not stop service")
```
Naatan, line 33: Please make a localized error that explains to the user that they should manually kill the state-svc process and try again.

Naatan (changes requested):

internal/installation/stop.go
```diff
@@ -30,7 +32,7 @@ func StopRunning(installPath string) (rerr error) {
 
 	err = stopSvc(installPath)
 	if err != nil {
-		return errs.Wrap(err, "Could not stop service")
+		return locale.NewError("err_stop_svc", "Unable to stop state-svc process. Please manually kill any running processes with name [NOTICE]state-svc[/RESET] and try again")
```
Naatan, line 35: Let's log this as a critical error as well.

internal/installation/stop.go
```diff
@@ -75,7 +77,13 @@ func stopSvc(installPath string) error {
 			continue
 		}
 
-		if n == constants.ServiceCommandName {
+		exe, err := p.Exe()
+		if err != nil {
+			logging.Error("Could not get process exectuable path: %v", err)
```
Naatan, line 82: We should log the process name as well, since that might not be in the error.

cmd/state-svc/manager.go
```diff
@@ -137,7 +137,7 @@ func (s *serviceManager) CheckPid(pid int) (*int, error) {
 		exe, err := p.Exe()
 		if err != nil {
 			logging.Error("Could not detect executable for pid, error: %s", errs.JoinMessage(err))
-		} else if filepath.Clean(exe) != filepath.Clean(osutils.Executable()) {
+		} else if filepath.Base(exe) != filepath.Base(osutils.Executable()) {
```
Naatan, line 140: We should use `constants.ServiceCommandName` here since we're now only looking at the basename. Also let's check the path for `activestate` like in the other section.

Naatan (changes requested):

internal/installation/stop.go
```diff
@@ -79,7 +80,7 @@ func stopSvc(installPath string) error {
 
 		exe, err := p.Exe()
 		if err != nil {
-			logging.Error("Could not get process exectuable path: %v", err)
```
Naatan: This is missing the value for `%s`.

Naatan (changes requested):

Naatan (dismissed):

internal/installation/stop.go
```diff
@@ -78,14 +78,18 @@ func stopSvc(installPath string) error {
...
-			continue
-		}
+		if n == constants.ServiceCommandName {
+			exe, err := p.Exe()
+			if err != nil {
+				logging.Error("Could not get executable path for state-svc process, error: %v", err)
+				continue
+			}
+
+			if !strings.Contains(strings.ToLower(exe), "activestate") {
```
Naatan, line 88: Come to think of it, let's log this as an error. I'd like to know if this happens.

Naatan (changes requested):

Naatan (dismissed):

internal/installation/stop.go
```diff
@@ -35,5 +63,66 @@ func StopRunning(installPath string) (rerr error) {
...
+	}()
+
+	select {
+	case err := <-signalErrs:
+		if err != nil {
+			return errs.Wrap(err, "Could not send SIGTERM to service process")
+		}
+		logging.Debug("Stopped state-svc process with SIGTERM")
+		return nil
+	case <-ctx.Done():
```
daved, line 118: nit: If this is just acting as a timeout within this function and the context type isn't really needed to be passed around, I suggest:
```suggestion
	signalErrs := make(chan error)
	go func() {
		signalErrs <- proc.SendSignal(syscall.SIGTERM)
	}()

	select {
	case err := <-signalErrs:
		if err != nil {
			return errs.Wrap(err, "Could not send SIGTERM to service process")
		}
		logging.Debug("Stopped state-svc process with SIGTERM")
		return nil
	case <-time.After(time.Second):
```

daved (approved):

PR #1698, opened 2022-01-21
Renamed config.SetWithLock() to config.GetThenSet()

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-48" title="DX-48" target="_blank">StoryDX-48</a>  Get rid of `config.SetWithLock()`
  </td></table>
  <br />
 

This better captures what is going on.

I created a separate [JIRA ticket](https://activestatef.atlassian.net/jira/software/c/projects/DX/issues/DX-651) on improving concurrency so we can use something more generic like `config.WithLock()` and pass a function that calls `config.Get()` and then `config.Set()`. Until then, this at least limits confusion with `config.Set()`.

 Note: `config.Set()` is already thread-safe.

### DIFF
--- a/cmd/state-svc/manager.go
+++ b/cmd/state-svc/manager.go
@@ -35,8 +35,8 @@ func NewServiceManager(cfg *config.Instance) *serviceManager {
 
 func (s *serviceManager) Start(args ...string) error {
 	var proc *os.Process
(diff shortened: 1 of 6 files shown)


Naatan (approved):
These code changes totally hide the amount of work that went into it :p Great work!

PR #1705, opened 2022-01-24
Ensure InstanceID is included in analytics reports

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-655" title="DX-655" target="_blank">BugDX-655</a>  Analytics InstanceID should be send with ALL events
  </td></table>
  <br />

### DIFF
--- a/internal/analytics/client/sync/client.go
+++ b/internal/analytics/client/sync/client.go
@@ -14,7 +14,6 @@ import (
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/installation/storage"
-	"github.com/ActiveState/cli/internal/instanceid"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/machineid"
 	"github.com/ActiveState/cli/internal/rtutils/p"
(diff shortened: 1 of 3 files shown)


internal/analytics/client/sync/client.go
```diff
@@ -94,6 +95,7 @@ func New(cfg *config.Instance, auth *authentication.Auth) *Client {
 		UpdateTag:     p.StrP(tag),
 		UserID:        p.StrP(userID),
 		Flags:         p.StrP(dimensions.CalculateFlags()),
+		InstanceID:    p.StrP(instanceid.ID()),
```
Naatan, line 98: Since the sync client is used in `state-svc` wouldn't this override the InstanceID given to it from the async client?

Naatan (changes requested):

Naatan (changes requested):
Looks good! But I think we should cover it with a test, should be able to update this one:

http://github.com/ActiveState/cli/blob/9efa0b52c01d9ab1f344c5e3cf2b0997ec742160/test/integration/analytics_int_test.go#L26-L26

Alternatively manually testing this and filing a followup for the integration test would be fine as well I think. Just given the severity I think we need to make sure this works as intended.

Naatan (approved):

PR #1706, opened 2022-02-01
State config

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-629" title="DX-629" target="_blank">StoryDX-629</a>  We have a new `state config` command
  </td></table>
  <br />


internal/runners/config/meta.go
```diff
@@ -0,0 +1,23 @@
...
+type event func() error
+
+type configType int
+
+const (
+	Int configType = iota
+	Bool
+)
+
+type configMeta struct {
```
MDrakos, line 14: I struggled with the naming in this file. Open to suggestions.

internal/runners/config/config.go
```diff
@@ -0,0 +1,39 @@
...
+	return usageFunc()
+}
+
+func validateKey(key string) error {
+	regex := regexp.MustCompile(`^[A-Za-z0-9\.]+$`)
+	if !regex.MatchString(key) {
+		return locale.NewInputError("err_config_invalid_key", "The config key can only consist of alphanumeric characters and a '.'")
+	}
+	return nil
+}
```
daved, line 39: I suggest that this behavior should be part of the parameter type so that it trips early and provides feedback to the user before it even reaches the runner. That is, we should use a typed string that implements the captain.ArgMarshaler interface so that it validates earlier in the control flow (when processed by Captain). If this sounds too involved for this PR, please file a story.

internal/runners/config/meta.go
daved: Maybe configRule, configSetup, or something similar.

internal/runners/config/set.go
```diff
@@ -0,0 +1,70 @@
...
+	if !ok {
+		return nil
+	}
+
+	if value.setEvent == nil {
+		return nil
+	}
+
+	return value.setEvent()
+}
```
daved, line 70: nit: The key text casing is not handled when setEvent is called or within the function itself, but it is in the switch above. If that is a fundamental part of the meta map interaction, that map might be better made a custom map type that has methods to get values. Likewise for setting values. So, `meta.Set(key, getFn, nil)` where the nil is automatically set as `func() error { return nil} }` so that callers do not have to check nil, and `val, ok := meta.Get(key)` - both Get and Set methods would process the key so callers do not need to.

internal/runners/config/meta.go
```diff
@@ -0,0 +1,23 @@
...
+	allowedType configType
+	getEvent    event
+	setEvent    event
+}
+
+var meta = map[string]configMeta{
+	constants.SvcConfigPid:  {Int, nil, nil},
+	constants.SvcConfigPort: {Int, nil, nil},
+}
```
daved, line 23: nit: Ultimately, IIUC, this logic makes most sense in the internal/config package. It would need to be configured either kinda-globally in that config package, or structurally in the main package to be passed into the internal/config construction func. For this PR, it's not being used, and it looks like it can be moved easily enough, so no worries for now.

internal/runners/config/set.go
daved: If this logic does end up in the internal/config package, the runner will be much simpler. It would just be a set for set and a get for get. The verification of type and the callbacks would be handled in the lower-level package.

daved (changes requested):
Probably not much actionable for this PR. If there is anything you don't action that seems worthwhile, and you think the knowledge could be lost, please file stories.

internal/runners/config/meta.go
MDrakos: If either you or @Naatan feels strongly about this being in the `internal/config` package I can move it. I didn't want to introduce the config meta to that package as I didn't want the `internal/config` package to have knowledge of specific config keys. If anything this could be its own package outside of both? But it felt better in the runner as that's where we need to apply rules to setting/getting config values.

internal/runners/config/set.go
MDrakos: This function isn't necessary anymore with some of the changes I made to the config rules

internal/runners/config/meta.go
daved: For now, I think it's best to land this as it is. If we do end up moving it into internal/config, we can work through how to ensure there's separation of concerns at that time.

daved (dismissed):

cmd/state/internal/cmdtree/config.go
```diff
@@ -0,0 +1,75 @@
...
+			}
+			return runner.Run(ccmd.Usage)
+		}).SetGroup(UtilsGroup)
+}
+
+func newConfigGetCommand(prime *primer.Values) *captain.Command {
+	params := config.GetParams{}
+	return captain.NewCommand(
+		"get",
+		locale.Tl("config_get_title", "Getting config value"),
```
Naatan, line 31:
```suggestion
		locale.Tl("config_get_title", "Get config value"),
```

cmd/state/internal/cmdtree/config.go
```diff
@@ -0,0 +1,75 @@
...
+			runner := config.NewGet(prime)
+			return runner.Run(params)
+		})
+}
+
+func newConfigSetCommand(prime *primer.Values) *captain.Command {
+	params := config.SetParams{}
+	return captain.NewCommand(
+		"set",
+		locale.Tl("config_set_title", "Setting config value"),
```
Naatan, line 53:
```suggestion
		locale.Tl("config_set_title", "Set config value"),
```

cmd/state/internal/cmdtree/config.go
Naatan: This is also used in `--help`, where 'Getting' would sound a bit weird.

cmd/state/internal/cmdtree/config.go
```diff
@@ -0,0 +1,75 @@
...
+		locale.Tl("config_set_description", "Set config values"),
```
Naatan, line 54:
```suggestion
		"",
```

internal/runners/config/get.go
```diff
@@ -0,0 +1,36 @@
...
+
+func (g *Get) Run(params GetParams) error {
+	value := g.cfg.Get(params.Key.String())
+	if value == nil {
+		return locale.NewInputError("err_config_not_found", "No config value for key: {{.V0}}", params.Key.String())
+	}
+
+	err := rules.Get(params.Key).getEvent()
+	if err != nil {
+		logging.Error("Could not execute additional logic on config set")
```
Naatan, line 31:
```suggestion
		logging.Error("Could not execute additional logic on config get")
```

internal/runners/config/key.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/ActiveState/cli/internal/locale"
+)
+
+type Key string
+
+// Set implements the captain ArgMarshaler interface
+func (k *Key) Set(v string) error {
+	regex := regexp.MustCompile(`^[A-Za-z0-9\.]+$`)
+	if !regex.MatchString(v) {
+		return locale.NewInputError("err_set_invalid_key", "Invalid config key. The config key can only consist of alphanumeric characters and a '.'")
```
Naatan, line 15:
```suggestion
		return locale.NewInputError("err_set_invalid_key", "Invalid config key. The config key can only consist of alphanumeric and dot characters")
```

internal/runners/config/rules.go
```diff
@@ -0,0 +1,36 @@
...
+type configType int
+
+const (
+	String configType = iota
+	Int
+	Bool
+)
+
+type configRule struct {
+	allowedType configType
```
Naatan, line 18: Nit: I would just drop "allowed" in the name here, it doesn't seem to add anything but it creates overhead in my parsing of the code.

internal/runners/config/get.go
```diff
@@ -0,0 +1,36 @@
...
+	return &Get{prime.Output(), prime.Config()}
+}
+
+func (g *Get) Run(params GetParams) error {
+	value := g.cfg.Get(params.Key.String())
+	if value == nil {
+		return locale.NewInputError("err_config_not_found", "No config value for key: {{.V0}}", params.Key.String())
+	}
+
+	err := rules.Get(params.Key).getEvent()
```
Naatan, line 29: This doesn't appear to affect the `value`. The point of the event is to be able to interpret and modify the value.

internal/runners/config/set.go
```diff
@@ -0,0 +1,50 @@
...
+	default:
+		value = params.Value
+	}
+
+	err := s.cfg.Set(params.Key.String(), value)
+	if err != nil {
+		return locale.WrapError(err, "err_cofing_set", fmt.Sprintf("Could not set value %s for key %s", value, params.Key))
+	}
+
+	err = rules.Get(params.Key).setEvent()
```
Naatan, line 43: Like with `Get`, the point of the event is to modify the value. So this should h appen before it is stored.

internal/runners/config/rules.go
```diff
@@ -0,0 +1,36 @@
...
+	rule, ok := c[key]
+	if !ok {
+		return configRule{String, emptyEvent, emptyEvent}
+	}
+	return rule
+}
+
+var rules = configRules{
+	constants.SvcConfigPid:  {Int, emptyEvent, emptyEvent},
+	constants.SvcConfigPort: {Int, emptyEvent, emptyEvent},
```
Naatan, line 35: This would require the events to be registered from inside this package. The idea is that whatever package "owns" the config key can register its usage from its own package.

Naatan: The svc one is kind of interesting because it is only used in the state-svc internal packages, which obviously wouldn't be able to register stuff that gets included in the `state` runtime. Maybe for those we could use the `svcmanager`.

This also raises an interesting point in that no one should ever be manually setting these config values. Maybe we need a way to say "set is not allowed for these keys". If that's cheaply added then I'd say go for it, but otherwise we can do a followup. Pretty unlikely someone would be running this in the interim.

internal/runners/config/key.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"regexp"
+
+	"github.com/ActiveState/cli/internal/locale"
+)
+
+type Key string
+
+// Set implements the captain ArgMarshaler interface
+func (k *Key) Set(v string) error {
```
Naatan, line 12: This seems like a good candidate for a unit test.

internal/runners/config/rules.go
MDrakos: I started moving things around and ended up putting the `rules` in the `internal/config` package like @daved originally suggested. Not sure if this is what you intended from your comment, but take a look and let me know.

> This also raises an interesting point in that no one should ever be manually setting these config values. Maybe we need a way to say "set is not allowed for these keys". If that's cheaply added then I'd say go for it, but otherwise we can do a followup. Pretty unlikely someone would be running this in the interim.

I'll see what if I can come up with something that might work but if that ends up taking a lot of time I will file a follow up.

daved: Just quickly (sorry if it's not the clearest info) - We'd likely want to inject this from the main package. That is, configure the configuration. Alternately, if that information that affects the config config is not needed elsewhere, it can be set as internal/config package global. The disallowed keys as well as callbacks can be configured in this way.

Naatan: To be clear; I want to configure things like `constants.SvcConfigPort: {Int, emptyEvent, emptyEvent}` from the package that "owns" this aspect of the configuration. Eg. `SvcManager`.

Neither `main` nor `config` should have opinions on `SvcConfigPort` or other similar config keys.

cmd/state-svc/manager.go
```diff
@@ -31,6 +31,8 @@ type serviceManager struct {
 }
 
 func NewServiceManager(cfg *config.Instance) *serviceManager {
+	config.NewRule(constants.SvcConfigPid, config.Int, config.EmptyEvent, config.EmptyEvent)
+	config.NewRule(constants.SvcConfigPort, config.Int, config.EmptyEvent, config.EmptyEvent)
```
Naatan, line 35: This looks like the right package, but the wrong location. I think this should be inside an `init()` function as it's effectively a global setting. It doesn't seem worth passing this around in the config instance itself, though I'm open to being wrong about that. @daved thoughts?

internal/config/instance.go
```diff
@@ -119,6 +119,22 @@ func (i *Instance) setWithCallback(key string, valueF func(currentValue interfac
...
+	case Int:
+		v = cast.ToInt(v)
+	case String:
+		v = cast.ToString(v)
+	}
+
+	err = rule.SetEvent(v)
+	if err != nil {
+		logging.Error("Could not execute additional logic on config set, err: %w", err)
+	}
```
Naatan, line 136: This logic should be unique to the `config` runner, and not part of our internal logic.

internal/config/rules.go
```diff
@@ -0,0 +1,44 @@
+package config
```
Naatan, line 1: Let's put this in `internal/mediators/config`, the idea being that `mediators` is what will eventually be the `internal` directory for all our commands (whereas currently it is obfuscated with a lot of `cmd/state` specific stuff.

internal/config/instance.go
```diff
@@ -119,6 +119,22 @@ func (i *Instance) setWithCallback(key string, valueF func(currentValue interfac
...
+	switch rule.Type {
+	case Bool:
+		v = cast.ToBool(v)
+	case Int:
+		v = cast.ToInt(v)
+	case String:
+		v = cast.ToString(v)
+	}
+
+	err = rule.SetEvent(v)
```
Naatan, line 133: Can we make this more explicit by returning the intended new value of `v` rather than modifying the pointer?

Naatan (changes requested):

internal/runners/config/rules.go
daved: In this case, I suggest that the need is disparate from the config enough to warrant a different type. I thought it easiest to express this in a play: https://go.dev/play/p/sqFallRBDOR

This would mean that the underlying config is still passed around and whatnot, but logic/runners further down in the stack can wrap that config and have their own sets of rules. The keys can be codified in a central package and whatnot for ease.

internal/runners/config/get.go
```diff
@@ -25,6 +28,11 @@ func (g *Get) Run(params GetParams) error {
 		return locale.NewInputError("err_config_not_found", "No config value for key: {{.V0}}", params.Key.String())
 	}
 
+	value, err := configMediator.GetRule(params.Key.String()).GetEvent(value)
+	if err != nil {
+		logging.Error("Could not execute additional logic on config get, err: %w", err)
```
Naatan, line 33: This should return as an error.

internal/runners/config/set.go
```diff
@@ -23,7 +27,24 @@ func NewSet(prime primeable) *Set {
...
+		value = cast.ToBool(params.Value)
+	case configMediator.Int:
+		value = cast.ToInt(params.Value)
+	default:
+		value = params.Value
+	}
+
+	value, err := rule.SetEvent(value)
+	if err != nil {
+		logging.Error("Could not execute additional logic on config set, err: %w", err)
```
Naatan, line 44: This should return as an error.

internal/mediators/config/rules.go
```diff
@@ -42,3 +41,7 @@ func NewRule(key string, t Type, get Event, set Event) {
 func SetRule(key string, rule Rule) {
 	rules[key] = rule
 }
+
+func init() {
+	rules = make(Rules)
+}
```
Naatan, line 47: I think you can call `make()` in the global context (ie. on line 27).

Naatan (changes requested):

internal/runners/config/get.go
```diff
@@ -30,7 +29,7 @@ func (g *Get) Run(params GetParams) error {
 
 	value, err := configMediator.GetRule(params.Key.String()).GetEvent(value)
 	if err != nil {
-		logging.Error("Could not execute additional logic on config get, err: %w", err)
+		return locale.WrapError(err, "err_config_get_event", "Could not execute config get event")
```
Naatan, line 32:
```suggestion
		return locale.WrapError(err, "err_config_get_event", "Could not retrieve config value, if this continues to happen please contact support.")
```

Always write these with the end-user in mind.

internal/runners/config/set.go
```diff
@@ -41,7 +40,7 @@ func (s *Set) Run(params SetParams) error {
 
 	value, err := rule.SetEvent(value)
 	if err != nil {
-		logging.Error("Could not execute additional logic on config set, err: %w", err)
+		return locale.WrapError(err, "err_config_set_event", "Could not execute config set event")
```
Naatan, line 43:
```suggestion
		return locale.WrapError(err, "err_config_set_event", "Could not store config value, if this continues to happen please contact support.")
```

Naatan (changes requested):

Naatan (approved):

PR #1707, opened 2022-02-01
Check config before sending rollbar reports

MDrakos:


cmd/state-svc/main.go
```diff
@@ -48,13 +48,21 @@ func main() {
 		os.Exit(exitCode)
 	}()
 
+	cfg, err := config.New()
+	if err != nil {
+		// We do not want to log an error here as we want to avoid potential rollbar reports until we load the config
+		logging.Debug("Failed to load configuration: %v", err)
+	} else {
+		logging.CurrentHandler().SetConfig(cfg)
+	}
```
MDrakos, line 57: This is a little bit awkward. Most of the run functions also initialize at config instance and handle the error returned from `config.New()`. We need the config to determine if we should send rollbar reports so we initialize it here before the run function and only debug log the error, leaving the run function to also initialize a config instance and handle the error.

Naatan: I think this means we need to initialize cfg earlier in the process rather than in the `run()` function.

cmd/state-svc/main.go
```diff
@@ -48,13 +48,21 @@ func main() {
...
+
 	logging.SetupRollbar(constants.StateServiceRollbarToken)
 
 	if os.Getenv("VERBOSE") == "true" {
 		logging.CurrentHandler().SetVerbose(true)
 	}
 
-	err := run()
+	err = run()
```
Naatan, line 65: Better not to reuse `err` as you'll run into false negatives for `!= nil` if the original type doesn't match. Yeah that's a thing... -.-

Naatan (changes requested):

cmd/state-svc/main.go
MDrakos: There is a bit of a chicken and egg problem here. Being unable to load the config would be an error we would want to get via rollbar but we need to load the config in order to determine if we want to send rollbar reports in the first place. I've left the config initialization in the `run()` function as we're only debug logging the error here.

Naatan (changes requested):
My other comment isn't addressed:

https://github.com/ActiveState/cli/pull/1707#discussion_r800976338

Edit: what the hell github, it wasn't showing your comment.. Revising..

cmd/state-svc/main.go
Naatan: I don't think that matters, unless I'm missing something. If it can't initialize config then it can report it to rollbar, and then error out. My comment was more about only initializing config once, and then passing it around. If it failed then it wouldn't pass it around cause it would've errored out anyway.

cmd/state-svc/main.go
```diff
@@ -50,19 +50,24 @@ func main() {
...
-		logging.CurrentHandler().SetConfig(cfg)
+		logging.Critical("Could not initialize config: %v", errs.JoinMessage(err))
+		exitCode = 1
+		return
 	}
+	defer func() {
+		if err := cfg.Close(); err != nil {
+			logging.Error("Failed to close config after exiting systray: %w", err)
+		}
+	}()
```
Naatan: We already have a defer above. Prefer to keep it all centralized there. I think we can probably use a function that does what is currently happening on line 71.

cmd/state-svc/main.go
```diff
@@ -50,19 +50,24 @@ func main() {
 
 	cfg, err := config.New()
 	if err != nil {
-		// We do not want to log an error here as we want to avoid potential rollbar reports until we load the config
-		logging.Debug("Failed to load configuration: %v", err)
-	} else {
-		logging.CurrentHandler().SetConfig(cfg)
+		logging.Critical("Could not initialize config: %v", errs.JoinMessage(err))
```
Naatan, line 43: We should print something to the user also, otherwise it just exits with code `1` and no explanation.

Naatan (changes requested):

cmd/state-svc/main.go
```diff
@@ -38,28 +38,29 @@ const (
...
 	cfg, err := config.New()
 	if err != nil {
 		logging.Critical("Could not initialize config: %v", errs.JoinMessage(err))
+		os.Stderr.WriteString(locale.Tr("err_main_config", errs.JoinMessage(err)))
 		exitCode = 1
 		return
 	}
+	logging.CurrentHandler().SetConfig(cfg)
+
 	defer func() {
```
Naatan, line 50: This needs to stay on line 41, it can't defer unless the defer was registered first. You could instead register the cfg variable before you define it.

cmd/state-svc/main.go
```diff
@@ -38,28 +38,29 @@ const (
...
-		if err := events.WaitForEvents(5*time.Second, rollbar.Wait, rollbar.Close, authentication.LegacyClose, logging.Close); err != nil {
-			logging.Warning("Failing to wait for rollbar to close")
-		}
-		os.Exit(exitCode)
-	}()
-
 	cfg, err := config.New()
 	if err != nil {
 		logging.Critical("Could not initialize config: %v", errs.JoinMessage(err))
+		os.Stderr.WriteString(locale.Tr("err_main_config", errs.JoinMessage(err)))
```
Naatan, line 44: I wouldn't use locale here, this early on in the execution that could open us up to some brand new type of bugs.

Naatan (changes requested):

cmd/state-svc/main.go
MDrakos: Ah right, that makes sense

cmd/state-svc/main.go
MDrakos: Fair enough. Note that we are using `locale` pretty early in the state tool here: https://github.com/ActiveState/cli/blob/master/cmd/state/main.go#L76. I can sneak in a fix for this and other instances of locale so early in this PR if that's something we want.

cmd/state-svc/main.go
```diff
@@ -38,26 +38,40 @@ const (
...
 
+	var cfg *config.Instance
 	defer func() {
 		if panics.HandlePanics(recover(), debug.Stack()) {
 			exitCode = 1
 		}
+
+		if err := cfg.Close(); err != nil {
+			logging.Error("Failed to close config after exiting systray: %w", err)
+		}
```
Naatan, line 49: Can't we put this in the `WaitForEvents` call below?

cmd/state-svc/main.go
```diff
@@ -38,26 +38,40 @@ const (
...
+
 		if err := events.WaitForEvents(5*time.Second, rollbar.Wait, rollbar.Close, authentication.LegacyClose, logging.Close); err != nil {
 			logging.Warning("Failing to wait for rollbar to close")
```
Naatan, line 52: Just noticed this says "rollbar" rather than "events". Simple tweak we can probably include in this PR.

Naatan (changes requested):

cmd/state-svc/main.go
MDrakos: Unfortunately no, the `WaitForEvents` only takes functions that do not return errors

Naatan (approved):

PR #1708, opened 2022-02-01
Handle non-bin execs on uninstall

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-648" title="DX-648" target="_blank">BugDX-648</a>  Uninstall doesn't uninstall
  </td></table>
  <br />
 

Putting up as a start for reworking.

### DIFF
--- a/internal/appinfo/appinfo.go
+++ b/internal/appinfo/appinfo.go
@@ -3,6 +3,7 @@ package appinfo
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/ActiveState/cli/internal/condition"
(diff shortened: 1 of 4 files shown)


Naatan (changes requested):
Nice, I like your solution :)

Given this is a regression and an important feature I think we should also make this integration test critical:

http://github.com/ActiveState/cli/blob/3edcf652ab7b7ed8ab5c7e563580b18cc579c2c3/test/integration/uninstall_int_test.go#L18-L18

Naatan (approved):

PR #1709, opened 2022-02-01
Check config before sending analytics reports

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-630" title="DX-630" target="_blank">StoryDX-630</a>  Analytics are not sent if config: `report.analytics=false`
  </td></table>
  <br />


internal/analytics/client/async/client.go
```diff
@@ -30,9 +30,9 @@ type Client struct {
 	output           string
 	projectNameSpace string
 	eventWaitGroup   *sync.WaitGroup
-
-	sessionToken string
-	updateTag    string
+	cfg              *config.Instance
```
Naatan, line 33: We don't need this on the async client, since the async client on its own doesn't really do anything but report it through the state-svc to the sync client. Adding it on the async client would only introduce unnecessary slowdown in the code.

internal/analytics/client/sync/client.go
```diff
@@ -125,6 +127,14 @@ func (a *Client) Wait() {
 
 // Events returns a channel to feed eventData directly to the report loop
 func (a *Client) report(category, action, label string, dimensions *dimensions.Values) {
+	send := true
+	if a.cfg != nil && a.cfg.IsSet(constants.ReportAnalayticsConfig) {
+		send = a.cfg.GetBool(constants.ReportAnalayticsConfig)
+	}
+	if !send {
+		return
+	}
```
Naatan, line 136:
```suggestion
	if a.cfg != nil && a.cfg.IsSet(constants.ReportAnalayticsConfig) && a.cfg.GetBool(constants.ReportAnalayticsConfig) {
		return
	}
```

Naatan: Also; is there a use-case for `a.cfg` being nil? I think that constitutes an error in our code and I'd almost rather want it to bubble up to the panic/recover so that we'd hear about it rather than leave use flying blind.

Naatan (changes requested):

Naatan:
Oh also; can we add an integration test for this?

Naatan (approved):
Nice one!

PR #1713, opened 2022-02-10
Remove version handling from installer

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-671" title="DX-671" target="_blank">BugDX-671</a>  Installer seems to install executables to both legacy and new (bin) locations
  </td></table>
  <br />


cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -29,16 +29,18 @@ func (suite *InstallerIntegrationTestSuite) TestInstallFromLocalSource() {
 	ts.UseDistinctStateExes()
 	target := filepath.Join(ts.Dirs.Work, "installation")
 
+	sourceDir := suite.populateSourceDir(ts)
+
 	// Run installer with source-path flag (ie. install from this local path)
 	cp := ts.SpawnCmdWithOpts(
 		ts.InstallerExe,
-		e2e.WithArgs(target, "--source-path", filepath.Dir(ts.InstallerExe)),
+		e2e.WithArgs(target, "--source-path", sourceDir),
```
Naatan, line 37: Can you summarize why this was necessary again? I wonder if the problem here was that `ts.InstallerExe` was pointing to the wrong file to begin with.

Naatan (changes requested):

cmd/state-installer/test/integration/installer_int_test.go
MDrakos: When I initially ran these tests I kept getting `Installing Update` not `Installing State Tool`. This was because the `state-installer` binary was being run with the state tool next to it. When we made those updates we may not have updated this test. The `sourceDir` contains only our 3 executables.

Naatan: I see. I think maybe a cleaner solution then is to put the installer in a subdir:

http://github.com/ActiveState/cli/blob/950bf1083327954d4a1e807c14de37fd26d771f7/internal/testhelpers/e2e/session.go#L138-L138

Naatan (changes requested):

cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -82,27 +83,15 @@ func (suite *InstallerIntegrationTestSuite) AssertConfig(ts *e2e.Session) {
...
-	err = fileutils.CopyFile(ts.Exe, filepath.Join(sourceDir, filepath.Base(ts.Exe)))
-	suite.NoError(err)
-	err = os.Remove(ts.Exe)
-	suite.NoError(err)
-
-	err = fileutils.CopyFile(ts.SvcExe, filepath.Join(sourceDir, filepath.Base(ts.SvcExe)))
-	suite.NoError(err)
-	err = os.Remove(ts.SvcExe)
+func (suite *InstallerIntegrationTestSuite) prepareInstaller(ts *e2e.Session) {
+	err := fileutils.CopyFile(ts.InstallerExe, filepath.Join(ts.Dirs.Work, "installer", filepath.Base(ts.InstallerExe)))
```
Naatan, line 87: I meant we should update this exact code here:

https://github.com/ActiveState/cli/blob/950bf1083327954d4a1e807c14de37fd26d771f7/internal/testhelpers/e2e/session.go#L138-L138

Because the behaviour is part of the installer, if we don't address it there it'll keep biting us.

Naatan (changes requested):

Naatan (approved):

PR #1714, opened 2022-02-10
Ensure heartbeats are thumping

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-672" title="DX-672" target="_blank">BugDX-672</a>  Heartbeat event not triggering on package install success or runtime execution
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/target/target.go
+++ b/pkg/platform/runtime/target/target.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/go-openapi/strfmt"
+	"github.com/thoas/go-funk"
 
 	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/hash"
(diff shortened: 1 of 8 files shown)


pkg/platform/runtime/target/target.go
```diff
@@ -35,16 +36,34 @@ const (
...
+	TriggerScript,
+	TriggerDeploy,
+	TriggerExec,
+	TriggerBranch,
+	TriggerImport,
+	TriggerPackage,
+	TriggerPull,
+	TriggerReset,
+	TriggerRevert,
+}
```
Naatan, line 50: I think at this point we should probably invert this to a blacklist rather than a whitelist :p Will file a followup.

Naatan (dismissed):
Beautifully simple fix! I can see why you'd be tempted to touch more code.. ;)

pkg/platform/runtime/target/target_test.go
```diff
@@ -15,7 +15,7 @@ func TestTrigger_IndicatesUsage(t *testing.T) {
 		},
 		{
 			"Import does not count as usage",
-			TriggerImport,
+			triggerUnknown,
 			false,
```
Naatan, line 19: This feels incorrect, the `TriggerImport` was right but given the changes we want to change the `false` to `true`.

Naatan:
Closing for https://github.com/ActiveState/cli/pull/1721

PR #1715, opened 2022-02-11
Unstable commands

MDrakos:


internal/captain/command.go
```diff
@@ -336,6 +342,15 @@ func (c *Command) interceptFunc() InterceptFunc {
 	}
 }
 
+// SetUnstable denotes if the command as a beta feature. This will remove the command
+// from state help, disable the commmand for those who haven't opted in to beta features,
+// and add a warning banner for those who have.
+func (c *Command) SetUnstable(unstable bool) *Command {
+	c.unstable = unstable
+	c.cobra.Hidden = unstable
```
Naatan, line 350: This wasn't in the AC's (my bad), but I think if you opted in it should show the command help, along with a warning saying it is unstable.

I've updated the AC's now to reflect this.

internal/captain/command.go
```diff
@@ -499,7 +514,15 @@ func (c *Command) subCommandNames() []string {
...
 func (c *Command) runner(cobraCmd *cobra.Command, args []string) error {
-	defer profile.Measure(fmt.Sprintf("captain:runner"), time.Now())
+	defer profile.Measure("captain:runner", time.Now())
+
+	if c.unstable {
+		if !c.cfg.GetBool(constants.UnstableConfig) {
+			c.out.Print(locale.Tr("unstable_command_warning", c.Name()))
+			return nil
+		}
+		c.out.Print(locale.T("beta_feature_banner"))
```
Naatan, line 524: For our internal logic please use "unstable" terminology rather than "beta". So we're consistent and it's easier to refactor if we need to.

internal/locale/locales/en-us.yaml
```diff
@@ -1786,4 +1786,11 @@ progress_fail:
...
-  other: " • Searching for [ACTIONABLE]{{.V0}}[/RESET] across all languages in the ActiveState Catalog"
\ No newline at end of file
+  other: " • Searching for [ACTIONABLE]{{.V0}}[/RESET] across all languages in the ActiveState Catalog"
+beta_feature_banner:
+  other: "[HEADING]█ Beta Feature: This feature is still in beta and may be unstable.[/RESET]"
+unstable_command_warning:
+  other: |
+    This feature [ACTIONABLE]state {{.V0}}[/RESET] is still in beta. If you want to opt-in to unstable features, run the following command:
+
+    state config set optin.unstable true
```
Naatan, line 1796:
```suggestion
    [ACTIONABLE]state config set optin.unstable true[/RESET]
```

Naatan (changes requested):
Nice and simple! Unfortunately I missed an AC myself.. Sorry about that.

Naatan (changes requested):

internal/constants/constants.go
```diff
@@ -352,6 +352,9 @@ const SvcConfigPort = "svc-port"
 // SvcConfigPid is the config key used for storing the svc pid
 const SvcConfigPid = "svc-pid"
 
+// UnstableConfig is the config key used to determine if a command is unstable
+const UnstableConfig = "optin.unstable"
```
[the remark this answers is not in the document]
MDrakos: Yes, we definitely do. Thanks!

internal/captain/command.go
MDrakos: Updated this to add `(Unstable)` to the end of any unstable commands if the user has opted in. Pinged Luke and Pete incase they want to change how this looks

internal/captain/command.go
```diff
@@ -347,7 +355,10 @@ func (c *Command) interceptFunc() InterceptFunc {
 // and add a warning banner for those who have.
 func (c *Command) SetUnstable(unstable bool) *Command {
 	c.unstable = unstable
-	c.cobra.Hidden = unstable
+	if !c.cfg.GetBool(constants.UnstableConfig) {
+		c.cobra.Hidden = unstable
+	}
+	c.name = fmt.Sprintf("%s (Unstable)", c.cobra.Use)
```
Naatan, line 361: This feels error prone. The `Command.name` isn't exclusively meant for the help output. I'd rather we address this in the help template.

Naatan (changes requested):

internal/captain/command.go
MDrakos: That's fair. We have a helper function that populates the `state` command's children. Moved this logic there.

Naatan (approved):

PR #1716, opened 2022-02-14
state auth authenticates via dashboard

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-657" title="DX-657" target="_blank">StoryDX-657</a>  As a user, I can sign into the state tool using the platform SSO service so that I can use GitHub to authenticate
  </td></table>
  <br />
 

* `state auth` and `state auth signup` now use the Platform API by default.
* `state auth --interactive` and `state auth signup --interactive` use the CLI.

https://activestatef.atlassian.net/browse/DX-657

Note: there are no integration tests for this feature because the Platform API is rate-limited. The integration tests would either fail or take forever to run.


pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
...
+		if response != nil {
+			s.storeAuthenticatedDevice(deviceCode, response)
+			break
+		} else if errs.Matches(err, &oauth.AuthDeviceGetBadRequest{}) {
+			badRequest := err.(*oauth.AuthDeviceGetBadRequest)
+			errorString := *badRequest.Payload.Error
+			const expiredToken = "expired_token"
+			const invalidClient = "invalid_client"
+			const slowDown = "slow_down"
+			if errorString == expiredToken || time.Since(startTime) >= timeout {
```
Naatan, line 286: This isn't addressing http timeouts or anything else happening further down the chain.

Consider using a context with timeouts, eg. https://golang.cafe/blog/golang-context-with-timeout-example.html

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
...
+				return locale.NewInputError("auth_device_timeout")
+			} else if errorString == invalidClient {
```
Naatan, line 288: We want to avoid string matching like this, unless it's done via swagger generated types. Is this errorString being defined in the swagger code? If so re-generating the mono-api should give you these strings: http://github.com/ActiveState/cli/blob/b94e2af417de13db488fc241fcfa86c24f8ede6e/activestate.generators.yaml#L2-L2

cmd/state/internal/cmdtree/auth.go
```diff
@@ -52,15 +58,23 @@ func newAuthCommand(prime *primer.Values) *captain.Command {
 
 func newSignupCommand(prime *primer.Values) *captain.Command {
 	signupRunner := auth.NewSignup(prime)
+	interactive := false
```
Naatan, line 61: Please use the same `params` pattern as used with the `newAuthCommand` implementation.

internal/locale/locales/en-us.yaml
```diff
@@ -874,10 +878,12 @@ auth_required_activate:
   other: You need to be authenticated to activate an ActiveState Platform project
 prompt_login_or_signup:
   other: Would you like to login to an existing ActiveState Platform account, or create a new account?
+prompt_login_browser_action:
+  other: Login with my existing account using the ActiveState Platform website
```
Naatan, line 882:
```suggestion
  other: Login with my existing account using your browser
```

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +181,19 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
 
 	return nil
 }
+
+// AuthenticateWithDevice attempts to authenticate this device with the Platform.
+func AuthenticateWithDevice(out output.Outputer) error {
+	err := authentication.LegacyGet().AuthenticateWithDevice(func(userCode, uri string) {
```
Naatan, line 187: The callback feels unnecessary. Just have `AuthenticateWithDevice` communicate back its result and use that.

Naatan (changes requested):

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,40 @@ func (s *Auth) Close() error {
...
 func (s *Auth) Authenticated() bool {
-	return s.clientAuth != nil
+	if s.clientAuth != nil {
+		return true
+	} else if existingDeviceCode := s.cfg.GetString(deviceCodeConfigKey); existingDeviceCode != "" && s.bearerToken == "" {
+		// Check if the device is still authenticated with the Platform and if so, get a token.
+		deviceCode := strfmt.UUID(existingDeviceCode)
+		params := oauth.NewAuthDeviceGetParams()
+		params.SetDeviceCode(deviceCode)
+		if response, err := mono.Get().Oauth.AuthDeviceGet(params); response != nil {
```
MDrakos, line 141: The `err` here is not being checked here, just that the `response` is not `nil`

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,40 @@ func (s *Auth) Close() error {
...
+			s.storeAuthenticatedDevice(deviceCode, response)
```
MDrakos, line 142: This function returns an error that is not being checked as well

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,40 @@ func (s *Auth) Close() error {
...
+		} else {
+			// Either the token for deviceCode has expired, we are rate-limited by the Platform and
+			// have to try again later, or the Platform is unreachable.
+			// Rate limiting can happen during testing.
+			if badRequest, ok := err.(*oauth.AuthDeviceGetBadRequest); ok && *badRequest.Payload.Error == "slow_down" {
+				logging.Warning("Attempting to query the Platform for device authentication status too frequently.")
+			}
```
MDrakos, line 149: Do we want to give any indication to the user of what's happening in this case?

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,40 @@ func (s *Auth) Close() error {
...
+		deviceCode := strfmt.UUID(existingDeviceCode)
+		params := oauth.NewAuthDeviceGetParams()
+		params.SetDeviceCode(deviceCode)
+		if response, err := mono.Get().Oauth.AuthDeviceGet(params); response != nil {
+			s.storeAuthenticatedDevice(deviceCode, response)
+		} else {
+			// Either the token for deviceCode has expired, we are rate-limited by the Platform and
+			// have to try again later, or the Platform is unreachable.
+			// Rate limiting can happen during testing.
+			if badRequest, ok := err.(*oauth.AuthDeviceGetBadRequest); ok && *badRequest.Payload.Error == "slow_down" {
```
MDrakos, line 147: You can use the generated swagger-generated constant rather than hard-coding the string. See: https://github.com/ActiveState/cli/pull/1716/files#diff-f57554ac911491c800c556763e8e626f0cdc6f7fc54d7757ef21e19000634fa3R222

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
...
+	// Wait for the authorization.
+	deviceCode := strfmt.UUID(*response.Payload.DeviceCode)
+	getParams := oauth.NewAuthDeviceGetParams()
+	getParams.SetDeviceCode(deviceCode)
+	startTime := time.Now()
+	const timeout = 6 * 60 * time.Second
+	for {
+		response, err := mono.Get().Oauth.AuthDeviceGet(getParams)
+		if response != nil {
+			s.storeAuthenticatedDevice(deviceCode, response)
```
MDrakos, line 278: Missing error handling

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
...
+			break
+		} else if errs.Matches(err, &oauth.AuthDeviceGetBadRequest{}) {
+			badRequest := err.(*oauth.AuthDeviceGetBadRequest)
+			errorString := *badRequest.Payload.Error
+			const expiredToken = "expired_token"
+			const invalidClient = "invalid_client"
+			const slowDown = "slow_down"
```
MDrakos, line 285: There should be swagger generated constants for all of these that you can use

MDrakos (changes requested):
I didn't delve too deeply into the `AuthenticateWithDevice` function as Nathan has already requested changes on it. My comments are mostly small things that may have been missed.

internal/runners/auth/auth.go
```diff
@@ -61,7 +62,12 @@ func (a *Auth) Run(params *AuthParams) error {
 
 func (a *Auth) authenticate(params *AuthParams) error {
 	if params.Token == "" {
-		err := authlet.AuthenticateWithInput(params.Username, params.Password, params.Totp, a.Cfg, a.Outputer, a.Prompter)
+		var err error
+		if !params.Interactive {
```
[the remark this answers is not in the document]
mitchell-as: So if the user runs `state auth --username foo`, that should silently be interpreted as `state auth --interactive --username foo`?

pkg/cmdlets/auth/login.go
mitchell-as: The POST in `AuthenticateWithDevice` produces a user code and a URI that need to be communicated to the user before they can click the "Authorize" button. `AuthenticateWithDevice` is blocking on receiving the click, so the callback is used to facilitate this sequence of events.

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,40 @@ func (s *Auth) Close() error {
 	return nil
 }
 
+func (s *Auth) storeAuthenticatedDevice(deviceCode strfmt.UUID, response *oauth.AuthDeviceGetOK) error {
+	defer s.updateRollbarPerson()
+	s.user = response.Payload.AccessToken.User
+	s.bearerToken = response.Payload.AccessToken.Token
+	clientAuth := httptransport.BearerToken(s.bearerToken)
+	s.clientAuth = &clientAuth
+	err := s.cfg.Set(deviceCodeConfigKey, deviceCode)
```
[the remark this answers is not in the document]
mitchell-as: The deviceCode *is* the API token as far as I can tell (that is the thing used for `state push` and other commands that require authentication). I'll probably have to ask Carey about this in further detail.

pkg/platform/authentication/auth.go
mitchell-as: I've looked at the generated swagger code and if response is `nil`, then `err` is not nil and vice-versa. I could change it to be an `} else if err != nil {`. Is that preferrable?

pkg/platform/authentication/auth.go
mitchell-as: No, this is something internal that I don't think we should expose to the user. The worst thing that can happen at this point is the user will have to re-authenticate.

pkg/platform/authentication/auth.go
mitchell-as: Great, that's just what I was looking for.

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
...
+	}
+	// Post the authentication request to the Platform.
+	postParams := oauth.NewAuthDevicePostParams()
+	response, err := mono.Get().Oauth.AuthDevicePost(postParams)
+	if err != nil {
+		logging.Error("Error requesting device authentication: %v", err)
+		return locale.NewError("err_auth_device")
+	}
+	// Prompt the user to authorize the request.
+	promptCallback(*response.Payload.UserCode, *response.Payload.VerificationURIComplete)
```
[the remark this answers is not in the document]
mitchell-as: Sounds like we need two functions here: `RequestDeviceAuthentication()`, which returns a userCode and URI that the cmdlet can show. Then `WaitForAuthorization()`, which waits for the user to click the "Authorize" button. Does that sound better than what you see?

I honestly prefer the callback mechanism so there's less for the cmdlet (or developer) to do and less moving parts to keep track of.

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
...
+			const slowDown = "slow_down"
+			if errorString == expiredToken || time.Since(startTime) >= timeout {
+				return locale.NewInputError("auth_device_timeout")
+			} else if errorString == invalidClient {
+				logging.Error("Error requesting device authentication: invalid client") // IP address mismatch
+				return locale.NewError("err_auth_device")
+			} else if errorString == slowDown {
+				logging.Warning("Attempting to check for authorization status too frequently.")
+			}
+			time.Sleep(5 * time.Second) // then try again
```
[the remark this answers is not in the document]
mitchell-as: Yes, this is the polling. Carey told me that we need to stop polling on "expired_token" and "invalid_client". Other than that, poll away.

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,40 @@ func (s *Auth) Close() error {
...
+	defer s.updateRollbarPerson()
+	s.user = response.Payload.AccessToken.User
+	s.bearerToken = response.Payload.AccessToken.Token
+	clientAuth := httptransport.BearerToken(s.bearerToken)
+	s.clientAuth = &clientAuth
+	err := s.cfg.Set(deviceCodeConfigKey, deviceCode)
+	if err != nil {
+		return errs.Wrap(err, "Could not set deviceCode credentials in config")
+	}
+	return err
```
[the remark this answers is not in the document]
mitchell-as: Done

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,40 @@ func (s *Auth) Close() error {
...
+}
+
 // Authenticated checks whether we are currently authenticated
 func (s *Auth) Authenticated() bool {
-	return s.clientAuth != nil
+	if s.clientAuth != nil {
+		return true
+	} else if existingDeviceCode := s.cfg.GetString(deviceCodeConfigKey); existingDeviceCode != "" && s.bearerToken == "" {
```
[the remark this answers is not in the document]
mitchell-as: Okay.

internal/locale/locales/en-us.yaml
```diff
@@ -114,11 +114,15 @@ auth_description:
...
+  other: The username to authenticate with
 flag_state_auth_password_description:
-  other: The password to authenticate with, will be prompted for otherwise
+  other: The password to authenticate with
 flag_state_auth_totp_description:
   other: The TOTP code generated via two-factor authentication
+flag_state_auth_interactive_description:
+  other: Prompt for username and password. Otherwise the user is prompted to log into the ActiveState Platform using a web browser.
+flag_state_auth_signup_interactive_description:
+  other: Prompt for username and password. Otherwise the user is prompted to sign up for the ActiveState Platform using a web browser.
```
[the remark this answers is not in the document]
mitchell-as: Done

cmd/state/internal/cmdtree/auth.go
mitchell-as: Done.

internal/locale/locales/en-us.yaml
mitchell-as: Done.

internal/locale/locales/en-us.yaml
```diff
@@ -114,11 +114,15 @@ auth_description:
...
 flag_state_auth_username_description:
-  other: The username to authenticate with, will be prompted for otherwise
+  other: The username to authenticate with
 flag_state_auth_password_description:
-  other: The password to authenticate with, will be prompted for otherwise
+  other: The password to authenticate with
 flag_state_auth_totp_description:
   other: The TOTP code generated via two-factor authentication
+flag_state_auth_interactive_description:
+  other: Prompt for username and password. Otherwise the user is prompted to log into the ActiveState Platform using a web browser.
```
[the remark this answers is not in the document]
mitchell-as: Done.

internal/runners/auth/signup.go
```diff
@@ -17,11 +17,6 @@ func NewSignup(prime primeable) *Signup {
...
 
-func (s *Signup) Run() error {
-	err := authlet.Signup(s.Configurable, s.Outputer, s.Prompter)
-	if err != nil {
-		return err
-	}
-
-	return nil
+func (s *Signup) Run(interactive bool) error {
+	return authlet.Signup(s.Configurable, s.Outputer, s.Prompter, interactive)
```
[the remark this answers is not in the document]
mitchell-as: Done

pkg/platform/authentication/auth.go
mitchell-as: Done.

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
 	})
 }
 
+// AuthenticateWithDeviceCode posts a request to authenticate this device on the Platform and waits
+// for the user to authorize the request.
+// The given callback function is called when the user should be prompted to authorize the request.
+// Returns an error if authentication cannot be performed (e.g. timeout or Platform is unreachable).
+func (s *Auth) AuthenticateWithDevice(promptCallback func(userCode, uri string)) error {
```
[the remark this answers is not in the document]
mitchell-as: Okay, I've split it up into two parts and put those parts in the model package. Part 1 posts the device authorization and part 2 waits for a user action.

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +249,57 @@ func (s *Auth) AuthenticateWithToken(token string) error {
...
+			} else if errorString == slowDown {
+				logging.Warning("Attempting to check for authorization status too frequently.")
+			}
+			time.Sleep(5 * time.Second) // then try again
+		} else {
+			logging.Error("Error requesting device authentication status: %v", err)
+			return locale.NewError("err_auth_device")
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
mitchell-as: Done.

pkg/cmdlets/auth/login.go
mitchell-as: I have broken up the authentication process into two parts: request authorization, and wait for user action. This routine now makes use of those without a callback.

pkg/platform/authentication/auth.go
mitchell-as: The underlying `mono.Oauth.AuthDevicePost()` function appears to have built-in timeout handling. I've turned off my computer's networking before, and in the middle of requests, and each time the state tool reports an error authenticating. I think it's working as expected.

pkg/platform/authentication/auth.go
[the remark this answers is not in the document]
Naatan: That's missing one of my AC's though, which Carey hasn't been exposed to:

```
- We poll for the login success (ie. we do not show the ENTER prompt like the script Carey wrote)

  - BUT; when we first request the authorization via https://platform.activestate.com/api/v1/oauth/authorize/device this should optionally return a token which can tell the State tool NOT to pull and instead show the ENTER prompt

  - This is in case the polling turns out to create too high a load for our API/DB

  - This token is not implemented on the server at this time, nor is it meant to be. It is something we should be able to change API side after the fact if the need arises, but the State Tool already needs to be aware of this.
```

So the idea here is that we can at a later date update the API to give some sort of signal that it should not poll automatically and instead the user should press "Enter" to continue (ie. poll on Enter).

It doesn't feel like `expired_token` or `invalid_client` addresses this use-case, nor that we have the "Enter" prompt.

pkg/platform/authentication/auth.go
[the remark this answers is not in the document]
Naatan: The method that we use in State Tool to preserve the authenticated state is the API key, ie this thing:

https://platform.activestate.com/swagger-ui/#/Authentication/listTokens

We don't want this PR to change that. What it should do is once it receives authorization we should use that to generate an API token and store that just like we do for the interactive mode of authentication.

internal/runners/auth/auth.go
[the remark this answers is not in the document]
Naatan: Yes, I think that makes sense. It doesn't make sense that you'd have to specify `--interactive` when you already supplied the login information imo.

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +182,27 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
 
 	return nil
 }
+
+// AuthenticateWithDevice attempts to authenticate this device with the Platform.
+func AuthenticateWithDevice(out output.Outputer) error {
+	deviceCode, err := model.RequestDeviceAuthorization()
```
Naatan, line 188: Nit: Bit miss-leading calling this `deviceCode` when it returns a struct with things like the verification URL.

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +182,27 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
+	if err != nil {
+		return err
+	}
+	out.Notice(locale.Tr("auth_device_verify_security_code", *deviceCode.UserCode))
+	err = OpenURI(*deviceCode.VerificationURIComplete)
```
Naatan, line 193: You should only ever do this in Go if you first nil check the reference. Otherwise you can get a panic if `deviceCode.VerificationURIComplete=nil`.

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +182,27 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
 
 	return nil
 }
+
+// AuthenticateWithDevice attempts to authenticate this device with the Platform.
+func AuthenticateWithDevice(out output.Outputer) error {
+	deviceCode, err := model.RequestDeviceAuthorization()
+	if err != nil {
+		return err
```
Naatan, line 190: Please wrap your errors: https://github.com/ActiveState/cli/blob/master/docs/errors.md

(goes for all such cases in this PR)

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +182,27 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
+	}
+	out.Notice(locale.Tr("auth_device_verify_security_code", *deviceCode.UserCode))
+	err = OpenURI(*deviceCode.VerificationURIComplete)
+	if err != nil {
+		logging.Error("Could not open browser: %v", err)
+		out.Notice(locale.Tr("err_browser_open", *deviceCode.VerificationURIComplete))
```
Naatan, line 196: Is there a use-case for there being an error but us still wanting to continue? ie. why not return this error?

pkg/platform/authentication/auth.go
```diff
@@ -113,9 +116,51 @@ func (s *Auth) Close() error {
...
+		}
+	} else {
+		// Either the token for deviceCode has expired, we are rate-limited by the Platform and
+		// have to try again later, or the Platform is unreachable.
+		// Rate limiting can happen during testing.
+		if badRequest, ok := err.(*oauth.AuthDeviceGetBadRequest); ok && *badRequest.Payload.Error == oauth.AuthDeviceGetBadRequestBodyErrorSlowDown {
+			logging.Warning("Attempting to query the Platform for device authentication status too frequently.")
+		}
+	}
+	return false
```
Naatan, line 163: None of this should be required if we use the API key to store the authenticated state.

pkg/platform/model/auth.go
```diff
@@ -0,0 +1,60 @@
...
+// RequestDeviceAuthorization posts a request to authorize this device on the Platform and
+// returns the device code needed for authorization.
+// The user is subsequently required to visit the device code's URI and click the "Authorize"
+// button.
+func RequestDeviceAuthorization() (*mono_models.DeviceCode, error) {
+	postParams := oauth.NewAuthDevicePostParams()
+	response, err := mono.Get().Oauth.AuthDevicePost(postParams)
+	if err != nil {
+		logging.Error("Error requesting device authorization: %v", err)
+		return nil, locale.NewError("err_auth_device")
```
Naatan, line 25: Use error wrapping instead. No need to log anything here, that'll happen down the chain when the error is "recognized".

pkg/platform/model/auth.go
```diff
@@ -0,0 +1,60 @@
...
+
+// WaitForAuthorization waits for the user to authorize a previously posted device authorization
+// request and returns the completed authorization.
+// Returns an error if authorization cannot be performed (e.g. timeout or Platform is unreachable).
+func WaitForAuthorization(deviceCodePayload *mono_models.DeviceCode) (*mono_models.DeviceCodeComplete, error) {
+	deviceCode := strfmt.UUID(*deviceCodePayload.DeviceCode)
+	getParams := oauth.NewAuthDeviceGetParams()
+	getParams.SetDeviceCode(deviceCode)
+	startTime := time.Now()
+	const timeout = 5 * 60 * time.Second
```
Naatan, line 38: Go has a cleaner pattern for this: https://gobyexample.com/timeouts

pkg/platform/model/auth.go
```diff
@@ -0,0 +1,60 @@
...
+				logging.Error("Error requesting device authentication: invalid client") // IP address mismatch
+				return nil, locale.NewError("err_auth_device")
+			} else if errorString == oauth.AuthDeviceGetBadRequestBodyErrorSlowDown {
+				logging.Warning("Attempting to check for authorization status too frequently.")
+			}
+			time.Sleep(6 * time.Second) // then try again (6 seconds is the Platform rate limit)
+		} else {
+			logging.Error("Error requesting device authentication status: %v", err)
+			return nil, locale.NewError("err_auth_device")
+		}
```
Naatan, line 58: All these if's an elses make this difficult to grok, consider maybe using a `switch` here to increase readability.

https://go.dev/doc/effective_go#switch

Naatan (changes requested):
Shaping up nicely! Please also note my responses to earlier comments.

internal/runners/auth/signup.go
```diff
@@ -17,11 +19,14 @@ func NewSignup(prime primeable) *Signup {
...
+	if authentication.LegacyGet().Authenticated() {
+		return locale.NewInputError("err_auth_authenticated", "You are already authenticated as: {{.V0}}. You can log out by running `state auth logout`.", authentication.LegacyGet().WhoAmI())
 	}
 
-	return nil
+	if !interactive {
+		return authlet.AuthenticateWithDevice(s.Outputer) // user can sign up from this page too
+	} else {
+		return authlet.Signup(s.Configurable, s.Outputer, s.Prompter)
+	}
```
MDrakos, line 31: nit: since you're returning in the `if` case the else is unecessary
```suggestion
	if !interactive {
		return authlet.AuthenticateWithDevice(s.Outputer) // user can sign up from this page too
	}
	return authlet.Signup(s.Configurable, s.Outputer, s.Prompter)
```

pkg/platform/authentication/auth.go
```diff
@@ -49,6 +50,8 @@ type Configurable interface {
 	Close() error
 }
 
+const deviceCodeConfigKey = "deviceCode"
```
MDrakos, line 53: I think we wanted to standardize our config keys so this should be `device.code` @Naatan can you confirm?

pkg/platform/model/auth.go
MDrakos: I'm not sure if it will work for your use case but we have used a retry HTTP client before. It may require some special configuration and even then still might not work for what you want but could be worth a shot. It's called `retryhttp` and it's already in our vendor directory.

Naatan: I think retrying would only further complicate the timeout in this case though. And given the use-case retrying seems unnecessary.

pkg/platform/authentication/auth.go
Naatan: Good point! But in this case this shouldn't be stored in the config at all.

pkg/platform/authentication/auth.go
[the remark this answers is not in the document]
th3coop: I'm not sure what problem you're trying to solve @Naatan that the API doesn't solve already.  If clients poll faster than every 5 second they get a slow down warning which is why Mitchell has the 5 second wait here.  Why is a prompt necessary from the API?  You suggest deviate from the IETF spec which I'm going to push back on unless there is a really good reason to do so.

pkg/platform/authentication/auth.go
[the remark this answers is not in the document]
mitchell-as: Okay.

pkg/platform/authentication/auth.go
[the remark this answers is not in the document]
Naatan: This is a failsafe to save us from problems like the one we had recently where we were sending billions of events, if that ever happened on our own API's we'd be swamped. I discussed this with Shane and we felt this would be a cheap solution. To be clear no action is required on your end @th3coop other than to advice if there is a sensible place in the payload that could carry this signal to the client.

So to be clear; if all goes well then we would never use this and the API would never have to support it. But IF we end up having to need it because the API is being overwhelmed this is what would happen:

- The client calls the API as normal to start the auth flow, but it has a simple check client-side that validates that if the API responds with some additional parameter then we should not poll and instead prompt the user to hit ENTER.
- We deploy a change to the API that will now have it provide this parameter
- The client sees this and prompts the user to press ENTER and saves us from having our API hammered
- Crisis averted.

This ought to be a really simple tweak we can make client-side that will allow us to avoid a big crisis. Low cost, high value. We can come up with a parameter of our own, but I figure it would be better for you to provide one so we're not making any false assumptions.

internal/runners/auth/auth.go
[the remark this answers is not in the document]
mitchell-as: Okay.

pkg/platform/authentication/auth.go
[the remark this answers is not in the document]
th3coop: I don't understand how this does anything better than what the `slow_down` warning does.  The `slow_down` warning will prevent exactly the scenario you describe but your suggest will require not only the API to change but all 3rd party clients to account for and use the new API data.  Why deviate from how the DCAG spec defined to handle this scenario?  Just seems like you're adding complication with no benegit.

cmd/state/internal/cmdtree/auth.go
```diff
@@ -52,15 +58,23 @@ func newAuthCommand(prime *primer.Values) *captain.Command {
 
 func newSignupCommand(prime *primer.Values) *captain.Command {
 	signupRunner := auth.NewSignup(prime)
+	params := auth.AuthParams{}
```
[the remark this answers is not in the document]
mitchell-as: You're right.

internal/runners/auth/signup.go
```diff
@@ -17,11 +19,14 @@ func NewSignup(prime primeable) *Signup {
 	return &Signup{prime.Output(), prime.Prompt(), prime.Config()}
 }
 
-func (s *Signup) Run() error {
-	err := authlet.Signup(s.Configurable, s.Outputer, s.Prompter)
-	if err != nil {
-		return err
+func (s *Signup) Run(interactive bool) error {
+	if authentication.LegacyGet().Authenticated() {
```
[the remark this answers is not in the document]
mitchell-as: prime.Config() is a `keypairs.Configurable`, nothing like a `config.Instance` or `auth.Configurable` which is needed for `authentication.New()`. I tried changing it to be compatible and promptly walked into a minefield. The entire CLI API appears to not know when it wants to use a `keypairs.Configurable` or `config.Instance` and also appears to use them interchangeably. The auth legacy and config stuff clearly needs an overhaul and now is probably not the time to do it.

pkg/cmdlets/auth/login.go
mitchell-as: Okay.

pkg/cmdlets/auth/login.go
mitchell-as: I was looking at the swagger-generated code and it would never do this. A check seemed redundant, but I'll do it anyway. It just likes to pass string pointers in order to reduce allocations.

pkg/cmdlets/auth/login.go
mitchell-as: This is not an error that would prevent authentication. The user is prompted here to manually copy-paste the URL into their browser and authorize the request. In the meantime, we can still poll for authorization.

pkg/cmdlets/auth/login.go
mitchell-as: Okay.

pkg/platform/authentication/auth.go
mitchell-as: Agreed.

pkg/platform/model/auth.go
mitchell-as: Okay

pkg/platform/model/auth.go
mitchell-as: Pete suggested a timeout of 5 minutes in order to allow the user to complete the Platform sign up process if they have not already. The device authorization also has an internal timeout. The State Tool will poll for an affirmative authorization request until we either hit Pete's suggested timeout, or the Platform says we timed out, whichever comes first.

This is not an HTTP network timeout. Sorry if I'm stating the obvious.

pkg/platform/model/auth.go
mitchell-as: I did originally, but it interfered with breaking out of the loop in my first iteration. This current iteration simply uses return, so the switch can come back. Thanks for noticing.

internal/runners/auth/signup.go
mitchell-as: Okay

pkg/cmdlets/auth/login.go
```diff
@@ -185,23 +185,27 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
 // AuthenticateWithDevice attempts to authenticate this device with the Platform.
 func AuthenticateWithDevice(out output.Outputer) error {
-	deviceCode, err := model.RequestDeviceAuthorization()
+	response, err := model.RequestDeviceAuthorization()
 	if err != nil {
-		return err
+		return locale.WrapError(err, "err_auth_device")
+	}
+	if response.UserCode == nil || response.VerificationURIComplete == nil {
+		logging.Error("Platform API error: device authorization request's response has nil UserCode and/or VerificationURIComplete")
```
Naatan, line 193: Anywhere where you return the error you do not need to log it; logging is handled where the error is handled, not where it originates.

pkg/cmdlets/auth/login.go
```diff
@@ -185,23 +185,27 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
+		return locale.NewError("err_auth_device")
 	}
-	out.Notice(locale.Tr("auth_device_verify_security_code", *deviceCode.UserCode))
-	err = OpenURI(*deviceCode.VerificationURIComplete)
+	out.Notice(locale.Tr("auth_device_verify_security_code", *response.UserCode))
+	err = OpenURI(*response.VerificationURIComplete)
 	if err != nil {
 		logging.Error("Could not open browser: %v", err)
-		out.Notice(locale.Tr("err_browser_open", *deviceCode.VerificationURIComplete))
+		out.Notice(locale.Tr("err_browser_open", *response.VerificationURIComplete))
```
Naatan, line 200: Shouldn't this be returning the error?

pkg/platform/model/auth.go
Naatan: I don't think either points raised here have anything to do with my original comment, which is that Go has a pattern for handling timeouts that I feel we should adopt here. See the link in my original comment.

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +182,31 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
+	if err != nil {
+		return locale.WrapError(err, "err_auth_device")
+	}
+	err = authentication.LegacyGet().AuthenticateWithDevice(authorization.AccessToken)
+	if err != nil {
+		return locale.WrapError(err, "err_auth_device")
+	}
+	out.Notice(locale.T("auth_device_success"))
+	return nil
+}
```
Naatan, line 212:
```suggestion
func AuthenticateWithDevice(out output.Outputer) error {
	response, err := model.RequestDeviceAuthorization()
	if err != nil {
		return locale.WrapError(err, "err_auth_device")
	}
	
	if response.UserCode == nil || response.VerificationURIComplete == nil {
		logging.Error("Platform API error: device authorization request's response has nil UserCode and/or VerificationURIComplete")
		return locale.NewError("err_auth_device")
	}
	
	out.Notice(locale.Tr("auth_device_verify_security_code", *response.UserCode))
	
	err = OpenURI(*response.VerificationURIComplete)
	if err != nil {
		logging.Error("Could not open browser: %v", err)
		out.Notice(locale.Tr("err_browser_open", *response.VerificationURIComplete))
	}
	
	authorization, err := model.WaitForAuthorization(response)
	if err != nil {
		return locale.WrapError(err, "err_auth_device")
	}
	
	err = authentication.LegacyGet().AuthenticateWithDevice(authorization.AccessToken)
	if err != nil {
		return locale.WrapError(err, "err_auth_device")
	}
	
	out.Notice(locale.T("auth_device_success"))
	
	return nil
}
```

Nit: Whitespace.

pkg/platform/authentication/auth.go
```diff
@@ -49,6 +49,8 @@ type Configurable interface {
 	Close() error
 }
 
+const ApiTokenConfigKey = "apiToken"
+
```
Naatan, line 53: Please drop any code change regarding saving the authenticated state. This is already implemented, please use:

http://github.com/ActiveState/cli/blob/950bf1083327954d4a1e807c14de37fd26d771f7/pkg/platform/authentication/auth.go#L293-L293

For example:

http://github.com/ActiveState/cli/blob/950bf1083327954d4a1e807c14de37fd26d771f7/pkg/platform/authentication/auth.go#L194-L194

pkg/platform/model/auth.go
```diff
@@ -0,0 +1,60 @@
...
+			badRequest := err.(*oauth.AuthDeviceGetBadRequest)
+			errorString := *badRequest.Payload.Error
+			if errorString == oauth.AuthDeviceGetBadRequestBodyErrorExpiredToken || time.Since(startTime) >= timeout {
+				return nil, locale.NewInputError("auth_device_timeout")
+			} else if errorString == oauth.AuthDeviceGetBadRequestBodyErrorInvalidClient {
+				logging.Error("Error requesting device authentication: invalid client") // IP address mismatch
+				return nil, locale.NewError("err_auth_device")
+			} else if errorString == oauth.AuthDeviceGetBadRequestBodyErrorSlowDown {
+				logging.Warning("Attempting to check for authorization status too frequently.")
+			}
```
Naatan, line 54:
```suggestion
			errorToken := *badRequest.Payload.Error
			if errorToken == oauth.AuthDeviceGetBadRequestBodyErrorExpiredToken || time.Since(startTime) >= timeout {
				return nil, locale.NewInputError("auth_device_timeout")
			} else if errorToken == oauth.AuthDeviceGetBadRequestBodyErrorInvalidClient {
				logging.Error("Error requesting device authentication: invalid client") // IP address mismatch
				return nil, locale.NewError("err_auth_device")
			} else if errorToken == oauth.AuthDeviceGetBadRequestBodyErrorSlowDown {
				logging.Warning("Attempting to check for authorization status too frequently.")
			}
```

Total nitpick but my OCD is triggering.. :p

Naatan (changes requested):

pkg/cmdlets/auth/login.go
Naatan: Ahh I see, makes sense! In that case please change `logging.Error` to `logging.Warning`.

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +182,31 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
+		return locale.WrapError(err, "err_auth_device")
+	}
+	if response.UserCode == nil || response.VerificationURIComplete == nil {
+		logging.Error("Platform API error: device authorization request's response has nil UserCode and/or VerificationURIComplete")
+		return locale.NewError("err_auth_device")
+	}
+	out.Notice(locale.Tr("auth_device_verify_security_code", *response.UserCode))
+	err = OpenURI(*response.VerificationURIComplete)
+	if err != nil {
+		logging.Error("Could not open browser: %v", err)
```
Naatan, line 199:
```suggestion
		logging.Warning("Could not open browser: %v", err)
```

internal/runners/auth/signup.go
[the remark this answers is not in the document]
mitchell-as: Yes, that one is okay, but there's one in *pkg/cmdlets/auth/login.go* that has already converted the `config.Instance` to a `keypairs.Configurable`. I haven't found a way to reconcile that one without it quickly blowing up in my face.

pkg/cmdlets/auth/login.go
mitchell-as: Okay.

internal/runners/auth/auth.go
```diff
@@ -63,7 +67,7 @@ func (a *Auth) Run(params *AuthParams) error {
 func (a *Auth) authenticate(params *AuthParams) error {
 	if params.Token == "" {
 		var err error
-		if !params.Interactive {
+		if !params.Interactive && params.Username == "" {
```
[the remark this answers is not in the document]
mitchell-as: Okay.

pkg/cmdlets/auth/login.go
mitchell-as: Okay.

pkg/cmdlets/auth/login.go
mitchell-as: Okay.

pkg/cmdlets/auth/login.go
mitchell-as: Okay.

pkg/platform/authentication/auth.go
mitchell-as: Good catch.

pkg/platform/model/auth.go
mitchell-as: Okay.

pkg/platform/model/auth.go
mitchell-as: I've just removed an explicit timeout from the client. The Platform has its own timeout and returns an "expired" status that is currently handled. There's no need to have another timeout mechanic.

pkg/platform/authentication/auth.go
Naatan: Response being nil seems like an error to me, so checking the error feels like the correct thing to do here unless there's a legitimate reason for there not being a response?

pkg/platform/model/auth.go
```diff
@@ -34,25 +34,24 @@ func WaitForAuthorization(deviceCodePayload *mono_models.DeviceCode) (*mono_mode
 	deviceCode := strfmt.UUID(*deviceCodePayload.DeviceCode)
 	getParams := oauth.NewAuthDeviceGetParams()
 	getParams.SetDeviceCode(deviceCode)
-	startTime := time.Now()
-	const timeout = 5 * 60 * time.Second
```
Naatan: Did you remove the timeout altogether?

pkg/cmdlets/auth/login.go
```diff
@@ -178,3 +182,37 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
+		logging.Warning("Could not open browser: %v", err)
+		out.Notice(locale.Tr("err_browser_open", *response.VerificationURIComplete))
+	}
+
+	authorization, err := model.WaitForAuthorization(response)
+	if err != nil {
+		return locale.WrapError(err, "err_auth_device")
+	}
+
+	err = authentication.LegacyGet().AuthenticateWithDevice(authorization.AccessToken)
```
Naatan, line 210: Still using legacy here.

pkg/platform/authentication/auth.go
```diff
@@ -215,6 +217,24 @@ func (s *Auth) AuthenticateWithToken(token string) error {
 	})
 }
 
+// AuthenticateWithDeviceCode authenticates with the given access token obtained via a Platform
+// device authentication request and response.
+func (s *Auth) AuthenticateWithDevice(accessToken *mono_models.JWT) error {
```
Naatan, line 222:
```suggestion
// AuthenticateWithToken authenticates with the given access token 
func (s *Auth) AuthenticateWithToken(accessToken *mono_models.JWT) error {
```

It doesn't seem like there's any need for this method to have an opinion on where it was obtained.

As part of this could you also update the code where `s.CreateToken()` is currently called? (ie. AuthenticateWithModel). Seems like a sensible change.

pkg/platform/model/auth.go
Naatan: This still isn't wrapping the error nor dropping the log.

pkg/platform/model/auth.go
```diff
@@ -0,0 +1,59 @@
...
+		case response != nil:
+			return response.Payload, nil
+		case errs.Matches(err, &oauth.AuthDeviceGetBadRequest{}):
+			badRequest := err.(*oauth.AuthDeviceGetBadRequest)
+			errorToken := *badRequest.Payload.Error
+			switch {
+			case errorToken == oauth.AuthDeviceGetBadRequestBodyErrorExpiredToken:
+				return nil, locale.NewInputError("auth_device_timeout")
+			case errorToken == oauth.AuthDeviceGetBadRequestBodyErrorInvalidClient:
+				logging.Error("Error requesting device authentication: invalid client") // IP address mismatch, but do not inform the user of this detail
```
Naatan, line 49: No need to log this, it'll happen down the chain as the error is returned.

pkg/platform/model/auth.go
```diff
@@ -0,0 +1,59 @@
...
+		response, err := mono.Get().Oauth.AuthDeviceGet(getParams)
+		switch {
+		case response != nil:
+			return response.Payload, nil
+		case errs.Matches(err, &oauth.AuthDeviceGetBadRequest{}):
+			badRequest := err.(*oauth.AuthDeviceGetBadRequest)
+			errorToken := *badRequest.Payload.Error
+			switch {
+			case errorToken == oauth.AuthDeviceGetBadRequestBodyErrorExpiredToken:
+				return nil, locale.NewInputError("auth_device_timeout")
```
Naatan, line 47: This should wrap the error, ie. `locale.WrapInputError()`

pkg/platform/model/auth.go
```diff
@@ -0,0 +1,59 @@
...
+			case errorToken == oauth.AuthDeviceGetBadRequestBodyErrorInvalidClient:
+				logging.Error("Error requesting device authentication: invalid client") // IP address mismatch, but do not inform the user of this detail
+				return nil, locale.NewError("err_auth_device")
```
Naatan, line 50: This should wrap the error.

Naatan (changes requested):

pkg/platform/authentication/auth.go
mitchell-as: This code has been changed. I cannot find this snippet anywhere else, so I think this has been addressed.

The reason I wasn't worried about errors is that response and err are mutually exclusive according to swagger. If response is not nil, there won't be an error, and vice-versa.

internal/runners/auth/signup.go
```diff
@@ -12,20 +12,22 @@ import (
...
 
 func NewSignup(prime primeable) *Signup {
 	return &Signup{prime.Output(), prime.Prompt(), prime.Config()}
 }
 
 func (s *Signup) Run(params *SignupParams) error {
-	if authentication.LegacyGet().Authenticated() {
-		return locale.NewInputError("err_auth_authenticated", "You are already authenticated as: {{.V0}}. You can log out by running `state auth logout`.", authentication.LegacyGet().WhoAmI())
+	auth := authentication.New(s.cfg)
+	defer auth.Close()
```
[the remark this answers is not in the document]
mitchell-as: Thanks for letting me know. I didn't know this was available!

pkg/platform/model/auth.go
mitchell-as: Yep: https://github.com/ActiveState/cli/pull/1716/commits/25d226c4a638aa4a2e4e314ce933830852af46d4

pkg/cmdlets/auth/login.go
mitchell-as: Right. This is where I was saying in an earlier comment that the code doesn't know whether it wants a `config.Instance` or a `keypair.Configurable`. With the help of your previous `primer.Author` comment I was able to fix this.

pkg/platform/authentication/auth.go
mitchell-as: Okay. I could use some input here though: https://github.com/ActiveState/cli/pull/1716/commits/8fdc40f972c91e70a0f42fa2205718457f62cb4d#diff-e253b887a875b9e415b41ca40c7a25272a7aa49379f529c57ebce4f871fb01d3R185

pkg/platform/model/auth.go
Naatan: Won't this lead to polling indefinitely if the user doesn't complete the auth?

mitchell-as: Gah, my bad. I linked the wrong commit: https://github.com/ActiveState/cli/pull/1716/commits/2204dbc4e5f197762ee6fa73d52f9755cf48e8e3

The Platform has its own timeout and we stop polling when that timeout is reached.

pkg/platform/model/auth/auth.go
```diff
@@ -0,0 +1,48 @@
...
+	getParams.SetDeviceCode(deviceCode)
+
+	response, err := mono.Get().Oauth.AuthDeviceGet(getParams)
+	if err != nil {
+		// Identify input or benign errors
+		if errs.Matches(err, &oauth.AuthDeviceGetBadRequest{}) {
+			errorToken := err.(*oauth.AuthDeviceGetBadRequest).Payload.Error
+			switch *errorToken {
+			case oauth.AuthDeviceGetBadRequestBodyErrorAuthorizationPending, oauth.AuthDeviceGetBadRequestBodyErrorSlowDown:
+				return nil, nil
```
MDrakos, line 38: Might be helpful to add a `logging.Debug` here

pkg/platform/authentication/auth.go
```diff
@@ -34,6 +35,8 @@ type ErrUnauthorized struct{ *locale.LocalizedError }
 
 type ErrTokenRequired struct{ *locale.LocalizedError }
 
+var ErrNotGranted = locale.NewInputError("err_auth_device_noauth")
```
MDrakos, line 38: Nit: This error is only used in this file. Do we need to export it?

MDrakos (approved):
Looks good to me 👍  My comments are just minor nits

MDrakos (approved):

Naatan (approved):

PR #1718, opened 2022-02-23
Update watcher executable verification

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-684" title="DX-684" target="_blank">BugDX-684</a>  Watcher does not match process
  </td></table>
  <br />
 

While this is niche the executable path from `args` and `e.Exec` can be different. This updates the check to just check the base.

### DIFF
--- a/cmd/state-svc/internal/rtwatcher/entry.go
+++ b/cmd/state-svc/internal/rtwatcher/entry.go
@@ -37,7 +37,7 @@ func (e entry) IsRunning() (bool, error) {
 		return false, errs.New("Process args are empty: %d", e.PID)
 	}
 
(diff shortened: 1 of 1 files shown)


mitchell-as (approved):

Naatan (changes requested):
This would not allow us to differentiate between a Python3 coming from us versus one coming from another install.

Any idea why the two paths aren't matching? That in itself seems like an issue.

MDrakos:
Filed a follow-up story [here](https://activestatef.atlassian.net/browse/DX-711). Closing this in favor of that,

PR #1719, opened 2022-02-25
Remove language MSI building from github actions

Naatan:


.github/workflows/build.yml
```diff
@@ -349,7 +338,6 @@ jobs:
         run: |
           state run deploy-updates
           state run deploy-installers
-          state run deploy-msi
```
MDrakos, line 352: Any reason to keep this script in the as.yaml?

MDrakos (approved):

.github/workflows/build.yml
Naatan: I didn't want to make this PR about removing the whole mechanic altogether, just wanted to remove it from our CI builds.

PR #1721, opened 2022-02-25
Fix missing heartbeats and success/failure events

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-672" title="DX-672" target="_blank">BugDX-672</a>  Heartbeat event not triggering on package install success or runtime execution
  </td></table>
  <br />


pkg/platform/runtime/runtime.go
```diff
@@ -127,14 +131,7 @@ func (r *Runtime) Env(inherit bool, useExecutors bool) (map[string]string, error
...
-			r.analytics.EventWithLabel(anaConsts.CatRuntime, anaConsts.ActRuntimeFailure, anaConsts.LblRtFailEnv)
-		} else {
-			r.analytics.Event(anaConsts.CatRuntime, anaConsts.ActRuntimeSuccess)
-			if r.target.Trigger().IndicatesUsage() {
-				r.recordUsage()
-			}
-		}
+		r.recordCompletion(err)
 		r.envAccessed = true
 	}
```
daved, line 136: Since the only thing interesting being protected here is to call r.recordCompletion, the r.completed guard should be enough and r.envAccessed can be dropped.

daved (approved):

daved (approved):

PR #1723, opened 2022-02-25
Notify user an existing API token is preventing a complete logout.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-697" title="DX-697" target="_blank">BugDX-697</a>  User can still state activate a new private project after they have state auth logout
  </td></table>
  <br />
 

I was not able to activate a private repository after logging out from the following cases:
* Log in using web authentication
* Log in using interactive authentication

I was able to activate a private repository after logging out from the following case:
* Log in using an API token (passwordless login)

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -219,6 +219,8 @@ logout_description:
   other: Logout
 logged_out:
   other: You have been logged out
+logout_still_have_api_token:
+  other: You still have an API token configured. You will not be completely logged out until you remove it.
 signup_success:
   other: Your account has been registered and a confirmation email has been sent to [NOTICE]{{.Email}}[/RESET], your account will have limited permissions until you confirm it.
 signup_failure:
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -220,7 +220,7 @@ logout_description:
 logged_out:
   other: You have been logged out
 logout_still_have_api_token:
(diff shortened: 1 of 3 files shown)


Naatan (changes requested):
This seems like a bandaid for a bug on our end. There should be no API token remaining locally after a user has logged out.

mitchell-as-test4:
We do not have the ability to modify our process' environment variables such that they persist for the next invocation of `state`. Or do we? I'm not aware of this capability.

mitchell-as:
We do not have the ability to modify our process' environment variables such that they persist for the next invocation of state. Or do we? I'm not aware of this capability.

Naatan:
Oh I see. The language of the error made me interpret this as meaning the API token "we" store. In that case I think it makes sense, though we should probably explicitly say "environment variable" in the error.

internal/locale/locales/en-us.yaml
```diff
@@ -219,6 +219,8 @@ logout_description:
   other: Logout
 logged_out:
   other: You have been logged out
+logout_still_have_api_token:
+  other: You still have an API token configured (environment variable or Google Cloud). You will not be completely logged out until you remove it.
```
Naatan, line 223:
```suggestion
  other: Logout is incomplete because you have authenticated with an environment variable. To complete logout please unset the '[ACTIONABLE]ACTIVESTATE_API_KEY[/RESET]' environment variable.
```

Naatan (changes requested):

Naatan (approved):

PR #1731, opened 2022-03-02
Address potential race condition when closing the config

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-723" title="DX-723" target="_blank">BugDX-723</a>  `state auth logout` isn't working for some
  </td></table>
  <br />

### DIFF
--- a/internal/config/instance.go
+++ b/internal/config/instance.go
@@ -113,13 +113,20 @@ func (i *Instance) GetThenSet(key string, valueF func(currentValue interface{}) 
 
 const CancelSet = "__CANCEL__"
 
-func (i *Instance) setWithCallback(key string, valueF func(currentValue interface{}) (interface{}, error)) error {
+func (i *Instance) setWithCallback(key string, valueF func(currentValue interface{}) (interface{}, error)) (rerr error) {
+	func() {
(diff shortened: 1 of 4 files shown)


internal/config/instance.go
```diff
@@ -113,13 +113,20 @@ func (i *Instance) GetThenSet(key string, valueF func(currentValue interface{})
 
 const CancelSet = "__CANCEL__"
 
-func (i *Instance) setWithCallback(key string, valueF func(currentValue interface{}) (interface{}, error)) error {
+func (i *Instance) setWithCallback(key string, valueF func(currentValue interface{}) (interface{}, error)) (rerr error) {
+	func() {
+		if rerr != nil {
```
mitchell-as, line 118: Isn't this always going to be `false` at this point in the function? Do you want to `defer` the anonymous function instead of calling it here?

mitchell-as (changes requested):

internal/config/instance.go
Naatan: Woops, that was indeed meant to be a defer. Good eye!

mitchell-as (approved):

PR #1732, opened 2022-03-03
Include commit ID in GA reporting for all runtime events.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-692" title="DX-692" target="_blank">StoryDX-692</a>  We sent the commit_id with ALL runtime events
  </td></table>
  <br />
 

Is this what you had in mind? I wasn't sure how to test it either.

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -181,15 +181,10 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 		}
 	}
 
-	pjNamespace := ""
-	if pj != nil {
-		pjNamespace = pj.Namespace().String()
-	}
-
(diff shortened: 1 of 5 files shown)


Naatan (changes requested):
I think you miss-interpreted the story. It says to include it with all runtime events, not to NOT include it.

Naatan (changes requested):
This is assuming we're always working with the current project's commit ID.

I think for this we need to do something like this, but for all runtime events:

http://github.com/ActiveState/cli/blob/039e7ea9239023a93b0b196d6fd59ca48ab85625/pkg/platform/runtime/runtime.go#L77-L77

Naatan (approved):

PR #1734, opened 2022-03-03
Use IPv4 in communications with the Platform.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-726" title="DX-726" target="_blank">BugDX-726</a>  Device authorization does not appear to support ipv6
  </td></table>
  <br />
 

By default the underlying system chooses whether to connect over IPv4 or IPv6. The Platform device authorization system requires device IP address to match web request IP address, so prefer IPv4 like web browsers currently do. There is no way to obtain IPv4 address from an IPv6 connection, so the only way to identify a match is if both connections use the same IP protocol.


pkg/platform/api/mono/mono.go
```diff
@@ -34,7 +37,32 @@ func Init(serviceURL *url.URL, auth *runtime.ClientAuthInfoWriter) *mono_client.
...
+
+	// For the Oauth client, prefer use of IPv4. This is needed particularly for device
+	// authorization, which compares the IP address of the State Tool request and the IP address of
+	// the web client request. If there's a mismatch, authorization fails. When this happens, it's
+	// often because the State Tool connects to the Platform via IPv6, but the browser does via IPv4
+	// (browsers apparently prefer IPv4 for now).
+
+	httpTransport, ok := http.DefaultTransport.(*http.Transport)
+	if !ok {
+		return client // for some reason the assertion can fail...
```
Naatan, line 50: I think this is a good example of a rare case in which we want a panic. This shouldn't ever happen, and if we put a panic here we're sure to know that it is happening long before we put out a release with this issue.

pkg/platform/api/mono/mono.go
```diff
@@ -34,7 +37,32 @@ func Init(serviceURL *url.URL, auth *runtime.ClientAuthInfoWriter) *mono_client.
...
+	}
+
+	ipv4TransportRuntime := httptransport.New(serviceURL.Host, serviceURL.Path, []string{serviceURL.Scheme})
+	var dialer net.Dialer
+	ipv4Transport := httpTransport.Clone()
+	ipv4Transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		if conn, err := dialer.DialContext(ctx, "tcp4", addr); conn != nil {
```
Naatan, line 57:
```suggestion	
	ipv4Transport := httpTransport.Clone()
	ipv4Transport.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
		dialer := &net.Dialer{}
		if conn, err := dialer.DialContext(ctx, "tcp4", addr); conn != nil {
```

Naatan: What you have might work but it's not a pattern I've seen before and I'm inclined to stick with what I know. From a quick reference this also seems to be how net.Dialer is normally used. Happy to be educated though.

pkg/platform/api/mono/mono.go
```diff
@@ -34,7 +37,32 @@ func Init(serviceURL *url.URL, auth *runtime.ClientAuthInfoWriter) *mono_client.
...
+			return conn, err
+		}
+		return dialer.DialContext(ctx, "tcp", addr)
```
Naatan, line 60: I guess this is the fallback to ipv6? Should add some comments pointing this out.

pkg/platform/api/mono/mono.go
```diff
@@ -34,7 +37,32 @@ func Init(serviceURL *url.URL, auth *runtime.ClientAuthInfoWriter) *mono_client.
...
+	// the web client request. If there's a mismatch, authorization fails. When this happens, it's
+	// often because the State Tool connects to the Platform via IPv6, but the browser does via IPv4
+	// (browsers apparently prefer IPv4 for now).
+
+	httpTransport, ok := http.DefaultTransport.(*http.Transport)
+	if !ok {
+		return client // for some reason the assertion can fail...
+	}
+
+	ipv4TransportRuntime := httptransport.New(serviceURL.Host, serviceURL.Path, []string{serviceURL.Scheme})
```
Naatan, line 53: Since this can also return an ipv6 transport I don't think we should call it `ipv4`, unless we call it something like `ipv4Preferred`.

Naatan (changes requested):

pkg/platform/api/mono/mono.go
mitchell-as: It does panic during some unit tests in the project package. I don't understand why, as this is part of Go's stdlib. I can try to spend more time looking into it, but I'm worried about rabbit holes.

pkg/platform/api/mono/mono.go
mitchell-as: I thought it was expensive to create a `net.Dialer` each time a DialContext is used, but perhaps this doesn't happen very often.

pkg/platform/api/mono/mono.go
mitchell-as: Yes it is. I'll add a comment.

pkg/platform/api/mono/mono.go
mitchell-as: Fair point.

pkg/platform/api/mono/mono.go
mitchell-as: Here's one of the panics: https://github.com/ActiveState/cli/runs/5414128742?check_suite_focus=true#step:9:79

mitchell-as: Oh... it's in a mock test. I think I found something to delete... Okay, we can panic then.

pkg/platform/api/mono/mono.go
Naatan: You can't call `DialContext` without creating it anyway. I think your `var declaration previously on line 51 was effectively doing this for you.

Naatan (approved):

PR #1735, opened 2022-03-03
Fix logout issue

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-723" title="DX-723" target="_blank">BugDX-723</a>  `state auth logout` isn't working for some
  </td></table>
  <br />
 

The main fix here is moving the CreateToken() logic out of the implicit behaviours of the auth package and into the cmdlets (ie. base it on the action the user was taking). Without it there can be race conditions when you assert vs write the apiToken to the config.

The rest of the changes are all related to getting the auth tests working again.


pkg/platform/authentication/auth.go
```diff
@@ -89,27 +90,27 @@ func ClientAuth() runtime.ClientAuthInfoWriter {
...
 	defer profile.Measure("auth:New", time.Now())
 	auth := &Auth{
 		cfg: cfg,
 	}
 
-	if auth.AvailableAPIToken() != "" {
+	return auth
+}
+
+func (s *Auth) Sync() error {
```
mitchell-as, line 106: Please document this. Why is it here? When does it have to be put after a call to `New()`? When does it not have to be called after `New()`?

mitchell-as (changes requested):

pkg/platform/authentication/auth.go
MDrakos: This is something that was done implicitly with `New`. This change makes me think that we want to keep the `LegacyGet` around. If so then maybe we should rename it. We may also want to update the docstrings to illustrate that `Sync` could be necessary. Otherwise, we should get rid of `LegacyGet` and rely on `New` so users of this API don't have to think about calling `Sync` at all.

MDrakos (approved):
Approved as I'm not sure what our intentions are around `LegacyGet` and I didn't want to hold up the PR.

Naatan:
I think the point is moot, because we should not be calling `New()` at all outside of the main package, and we indeed should get rid of `LegacyGet`. The fact that it's called `Legacy` should hopefully convey that already.

Combine that with the fact that this lib is ripe for a rewrite; it doesn't seem worth doing much else with it at the moment. A polished turd is still a turd :p

PR #1736, opened 2022-03-04
Auth requires both --username and --password flags

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-735" title="DX-735" target="_blank">BugDX-735</a>  CLI - AUTH: Using only the "--password" flag does not generate an error message and starts auth by authorization via the web browser.
  </td></table>
  <br />

### DIFF
--- a/internal/runners/auth/auth.go
+++ b/internal/runners/auth/auth.go
@@ -49,7 +49,7 @@ func (p AuthParams) verify() error {
 	}
 
 	if p.Totp != "" && (p.Username == "" || p.Password == "") {
-		return locale.WrapInputError("err_auth_invalid_totp_param", "[ACTIONABLE]--totp[/RESET] flag requires both [ACTIONABLE]--username[/RESET] and [ACTIONABLE]--password[/RESET] flags", args ...string)
+		return locale.NewInputError("err_auth_invalid_totp_param", "[ACTIONABLE]--totp[/RESET] flag requires both [ACTIONABLE]--username[/RESET] and [ACTIONABLE]--password[/RESET] flags")
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


internal/runners/auth/auth.go
```diff
@@ -39,6 +39,18 @@ type AuthParams struct {
...
 
+func (p AuthParams) verify() error {
+	if p.Username != "" && p.Password == "" {
+		return locale.NewInputError("err_auth_invalid_username_flag", "[ACTIONABLE]--username[/RESET] flag requires [ACTIONABLE]--password[/RESET] flag")
+	}
+
+	if p.Username == "" && p.Password != "" {
+		return locale.NewInputError("err_auth_invalid_password_param", "[ACTIONABLLE]--password[/RESET] flag requires [ACTIONABLE]--username[/RESET] flag")
+	}
+
```
Naatan, line 50: Should check totp here too?

internal/runners/auth/auth.go
```diff
@@ -68,27 +88,24 @@ func (a *Auth) Run(params *AuthParams) error {
...
-			err = authlet.AuthenticateWithInput(params.Username, params.Password, params.Totp, a.Cfg, a.Outputer, a.Prompter, a.Auth)
-		} else {
-			err = authlet.AuthenticateWithBrowser(a.Outputer, a.Auth, a.Prompter)
-		}
-		if err != nil {
-			return locale.WrapError(err, "login_err_auth")
-		}
-	} else {
-		err := a.Auth.AuthenticateWithModel(&mono_models.Credentials{
+	if params.Interactive || params.Username != "" || params.Totp != "" {
```
Naatan, line 91: Don't think we need totp here, the `verify` step should catch that.

Totp refers to "two factor auth". It is useless if you're not also supplying the username and password.

Naatan (changes requested):

internal/runners/auth/auth.go
MDrakos: Thanks! I was uncertain about that flag.

Naatan (approved):

PR #1738, opened 2022-03-07
Fix keypair not saving because of stale auth

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-741" title="DX-741" target="_blank">BugDX-741</a>  Can't auth with `--interactive` due to keypair failure
  </td></table>
  <br />

### DIFF
--- a/internal/keypairs/keypair.go
+++ b/internal/keypairs/keypair.go
@@ -103,7 +103,7 @@ func SaveEncodedKeypair(cfg Configurable, secretsClient *secretsapi.Client, encK
 		PublicKey:           &encKeypair.EncodedPublicKey,
 	})
 
-	if _, err := secretsClient.Keys.SaveKeypair(params, authentication.New(cfg).ClientAuth()); err != nil {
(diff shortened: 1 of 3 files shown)


internal/keypairs/keypair.go
```diff
@@ -103,7 +103,7 @@ func SaveEncodedKeypair(cfg Configurable, secretsClient *secretsapi.Client, encK
 		PublicKey:           &encKeypair.EncodedPublicKey,
 	})
 
-	if _, err := secretsClient.Keys.SaveKeypair(params, authentication.New(cfg).ClientAuth()); err != nil {
+	if _, err := secretsClient.Keys.SaveKeypair(params, authentication.LegacyGet().ClientAuth()); err != nil {
```
mitchell-as, line 106: I presume you have a good reason to use LegacyGet(), so I'll let it slide this time.

mitchell-as (approved):

internal/keypairs/keypair.go
Naatan: Yeah; secrets code is itself legacy and refactoring this out for this bug would be far too costly.

PR #1741, opened 2022-03-07
DX-745 Adds a fallback mechanism for opening URIs on linux

shaunmlowry:
On linux distributions with no display or otherwise running without a
native display manager (e.g. if they're running under WSL) there is no
`xdg-open` executable for generally opening files in a graphical
context.  They may however, still be able to open URIs in an
alternative way.  The open-golang package we're currently using only
knows about using `xdg-open` so this commit adds a fallback to try
using `x-www-browser` as provided by the alternatives mechanism on
linux systems where `xdg-open` fails

### DIFF
--- a/internal/osutils/openuri.go
+++ b/internal/osutils/openuri.go
@@ -1,0 +1,18 @@
+// +build !windows,!darwin
+
+package osutils
+
+import (
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/skratchdot/open-golang/open"
+)
+
+func OpenURI(input string) error {
+	err := open.Run(input)
+
+	if err != nil {
+		logging.Warning("open.Run failed, attempting alternative method to open browser. Error received: %v", err)
+		err = open.RunWith(input, "x-www-browser")
+	}
+	return err
+}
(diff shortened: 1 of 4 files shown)


internal/exeutils/openuri.go
```diff
@@ -0,0 +1,15 @@
...
+package exeutils
+
+import (
+	"github.com/skratchdot/open-golang/open"
+)
+
+func OpenURI(input string) error {
+	err := open.Run(input)
+	if err != nil {
+		err = open.RunWith(input, "x-www-browser")
```
Naatan, line 12:
```suggestion
        logging.Warning("open.Run failed, attempting alternative method to open browser. Error received: %v", err)
		err = open.RunWith(input, "x-www-browser")
```

internal/exeutils/openuri.go
```diff
@@ -0,0 +1,15 @@
+// +build !windows,!darwin
```
Naatan, line 1: The `exeutils` is for dealing with executables. Opening browser URI's doesn't feel like it belongs here. Perhaps `osutils` would be more appropriate, seeing as it's a very OS specific thing.

Naatan (changes requested):

internal/exeutils/openuri.go
shaunmlowry: moved

Naatan (approved):

PR #1742, opened 2022-03-08
Set channel changes installation directory

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-575" title="DX-575" target="_blank">BugDX-575</a>  `state update --set-channel` works
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -243,6 +243,7 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		}
 	}
 
+	// If we are installing a different branch, update install path
 	var switchChannel bool
 	if filepath.Base(params.path) != constants.BranchName {
 		params.path = filepath.Join(filepath.Dir(params.path), constants.BranchName)
@@ -341,26 +342,21 @@ func installOrUpdateFromLocalSource(out output.Outputer, cfg *config.Instance, a
 func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher, params *Params, isUpdate, switchChannel bool) error {
 	an.Event(AnalyticsFunnelCat, "post-install-events")
 
-	logging.Debug("postInstallEvents params.path: %s", params.path)
 	installPath, err := resolveInstallPath(params.path)
 	if err != nil {
 		return errs.Wrap(err, "Could not resolve installation path")
 	}
-	logging.Debug("postInstallEvents installPath: %s", installPath)
 
 	stateExe := appinfo.StateApp(installPath).Exec()
 	binPath, err := installation.BinPathFromInstallPath(installPath)
 	if err != nil {
 		return errs.Wrap(err, "Could not detect installation bin path")
 	}
-	logging.Debug("postInstallEvents binPath: %s", binPath)
-	logging.Debug("postInstallEvents isUpdate: %b", isUpdate)
 
 	// Execute requested command, these are mutually exclusive
 	switch {
 	// Execute provided --command
 	case params.command != "":
-		logging.Debug("postInstallEvents running command")
 		an.Event(AnalyticsFunnelCat, "forward-command")
 
 		out.Print(fmt.Sprintf("\nRunning `[ACTIONABLE]%s[/RESET]`\n", params.command))
@@ -371,7 +367,6 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		}
 	// Activate provided --activate Namespace
 	case params.activate.IsValid():
-		logging.Debug("postInstallEvents activating")
 		an.Event(AnalyticsFunnelCat, "forward-activate")
 
 		out.Print(fmt.Sprintf("\nRunning `[ACTIONABLE]state activate %s[/RESET]`\n", params.activate.String()))
@@ -381,7 +376,6 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		}
 	// Activate provided --activate-default Namespace
 	case params.activateDefault.IsValid():
-		logging.Debug("postInstallEvents default")
 		an.Event(AnalyticsFunnelCat, "forward-activate-default")
 
 		out.Print(fmt.Sprintf("\nRunning `[ACTIONABLE]state activate --default %s[/RESET]`\n", params.activateDefault.String()))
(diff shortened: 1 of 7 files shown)


cmd/state-installer/cmd.go
```diff
@@ -243,6 +243,13 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		}
 	}
 
+	// If we are installing a different branch, update install path
+	var switchChannel bool
+	if filepath.Base(params.path) != constants.BranchName {
+		params.path = filepath.Join(filepath.Dir(params.path), constants.BranchName)
+		switchChannel = true
+	}
```
Naatan, line 251: This feels too implicit. This is assuming that the path param is always implicitly given and it assumes that the installation path always ends in the branch name.

Both assumptions might currently be correct, but they may not always be. I'm not against doing this so long as we implement this in a way that clearly communicates the behavior. 

I suggest:

- Can we get rid of the path param so that we don't need this implicit logic in the first place? If not; why is it needed and how does that impact our use-case here?
- Implement a `installation.BranchForInstallPath()` function that returns which branch the given path is for. It can use the `filepath.Base()` approach; point is that it's centralized and clearly conveys the intend.

Naatan: More importantly though; I feel like this logic being in the installer at all is an issue. The `--set-channel` flag handling logic should instrument the installer in a way that provides us with the desired behavior without the installer needing to have any special knowledge over that behavior.

internal/installation/paths.go
```diff
@@ -29,23 +29,38 @@ func InstallPath() (string, error) {
...
 
+	activeStateOwnedPath := strings.Contains(strings.ToLower(stateInfo.Exec()), "activestate")
+	if filepath.Base(filepath.Dir(stateInfo.Exec())) == BinDirName && activeStateOwnedPath {
+		return filepath.Dir(filepath.Dir(filepath.Dir(stateInfo.Exec()))), nil // <return this>/<branch>/bin/state.exe
+	}
+
 	return DefaultInstallPath()
 }
 
+func BranchPathFromInstallPath(branch string) (string, error) {
```
Naatan, line 47: This seems inverted; you're returning the path for a branch, not a branch for a path. AAlso this feels like it overlaps with `installation.InstallPathForBranch`

Naatan (changes requested):

cmd/state-installer/cmd.go
MDrakos: > More importantly though; I feel like this logic being in the installer at all is an issue. The --set-channel flag handling logic should instrument the installer in a way that provides us with the desired behavior without the installer needing to have any special knowledge over that behavior.

This is what I initially went with but changed my approach as it would not affect previous versions of the state tool. So if you ran `state-update` with a version of the state tool that didn't have this fix it wouldn't change the installation directory. When the state tool updates itself it downloads the new installer so we can apply this fix to all of the older versions as well. If that's not an issue then we can go with that approach, however, we will have to disable the `update` integration tests for now as they are testing specific older versions of the state tool.

I like all the other suggestions in your first comment and will start looking into them 👍

Naatan: That's fine. Again this is mainly for internal use, and we wouldn't be using this with older State Tool versions once the fix lands.

The installer using the latest to perform the installation is to work around issues where the old installer is outright failing. It is not a mechanism intended to fix bugs in older State Tool versions unless it's something that's outright breaking updates.

internal/installation/paths.go
```diff
@@ -35,14 +35,15 @@ func InstallPath() (string, error) {
 
 	// If State Tool is already exists then we should detect the install path from there
 	stateInfo := appinfo.StateApp()
-	activeStateOwnedPath := strings.Contains(strings.ToLower(stateInfo.Exec()), "activestate")
-	if fileutils.TargetExists(stateInfo.Exec()) {
-		if filepath.Base(filepath.Dir(stateInfo.Exec())) == BinDirName && activeStateOwnedPath {
-			return filepath.Dir(filepath.Dir(stateInfo.Exec())), nil // <return this>/bin/state.exe
-		}
+	if !fileutils.TargetExists(stateInfo.Exec()) {
```
Naatan, line 38: This seems wrong, you return the dir if it doesn't exist?

internal/installation/paths.go
```diff
@@ -35,14 +35,15 @@ func InstallPath() (string, error) {
...
 		return filepath.Dir(stateInfo.Exec()), nil // <return this>/state.exe
 	}
 
+	activeStateOwnedPath := strings.Contains(strings.ToLower(stateInfo.Exec()), "activestate")
+	if filepath.Base(filepath.Dir(stateInfo.Exec())) == BinDirName && activeStateOwnedPath {
+		return filepath.Dir(filepath.Dir(filepath.Dir(stateInfo.Exec()))), nil // <return this>/<branch>/bin/state.exe
+	}
```
Naatan, line 45: What is this change fixing?

internal/updater/updater.go
```diff
@@ -84,13 +83,6 @@ func (u *AvailableUpdate) prepareInstall(installTargetPath string, args []string
 		return "", nil, errs.Wrap(err, "Downloaded update does not have installer")
 	}
 
-	if installTargetPath == "" {
-		installTargetPath, err = installation.InstallPath()
-		if err != nil {
-			return "", nil, errs.Wrap(err, "Could not detect install path")
-		}
-	}
```
Naatan, line 92: I could see why this would be problematic for `--set-channel`, but won't this cause issues for the default flow?

Naatan (changes requested):

internal/installation/paths.go
MDrakos: This was a leftover from my initial attempts. Reverted now

internal/updater/updater.go
MDrakos: Yep, you're right. I've put it back.

internal/installation/paths.go
Naatan: It doesn't look like it's reverted?

MDrakos: Should be now

internal/runners/update/update.go
```diff
@@ -66,7 +67,15 @@ func (u *Update) Run(params *Params) error {
 		return nil
 	}
 
-	err = up.InstallBlocking("")
+	var installPath string
+	if params.Channel != "" {
```
Naatan, line 71:
```suggestion
	if params.Channel != constants.BranchName { {
```

Naatan (changes requested):

PR #1746, opened 2022-03-10
Fix token authentication not saving

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-768" title="DX-768" target="_blank">BugDX-768</a>  `state auth --token` isn't preserved
  </td></table>
  <br />


internal/runners/auth/auth.go
```diff
@@ -31,11 +30,12 @@ func NewAuth(prime primeable) *Auth {
...
-	Username    string
-	Password    string
-	Totp        string
-	Interactive bool
+	Token          string
+	Username       string
+	Password       string
+	Totp           string
+	Interactive    bool
+	NonInteractive bool
```
mitchell-as, line 38: What is the relationship between `Interactive` and `NonInteractive`? Are they mutually exclusive? If so, perhaps these two fields can be replaced by an `Interactivity` enum.

internal/runners/auth/auth.go
```diff
@@ -61,6 +61,10 @@ type SignupParams struct {
 // Run runs our command
 func (a *Auth) Run(params *AuthParams) error {
 	if !a.Authenticated() {
+		if params.NonInteractive {
+			return locale.NewInputError("err_auth_loggedout", "You are not logged out.")
```
mitchell-as, line 65: Did you mean "You are not logged in."? Or perhaps "You are logged out."

pkg/cmdlets/auth/login.go
```diff
@@ -68,6 +68,23 @@ func AuthenticateWithInput(username, password, totp string, cfg keypairs.Configu
 	return nil
 }
 
+// AuthenticateWithInput will prompt the user for authentication if the input doesn't already provide it
```
mitchell-as, line 71: Should be AuthenticateWithToken. Comment body should probably also be updated to mention token instead of input.

pkg/platform/authentication/auth.go
```diff
@@ -366,7 +366,17 @@ func (s *Auth) CreateToken() error {
...
+	err = s.SaveToken(token)
+	if err != nil {
+		return errs.Wrap(err, "SaveToken failed")
+	}
+
+	return nil
+}
+
+// SaveToken will save an API token
+func (s *Auth) SaveToken(token string) error {
```
mitchell-as, line 378: Did you intentionally want to expose "SaveToken" outside the package? Or should it be private?

mitchell-as (changes requested):

internal/runners/auth/auth.go
Naatan: They are not mutually exclusive. Interactive in this case basically means "interact via my terminal" and non-interactive means "do not interact with me in ANY way; browser or otherwise".

pkg/platform/authentication/auth.go
Naatan: Yes, it's used in the authlet.

mitchell-as (changes requested):

mitchell-as (approved):

PR #1747, opened 2022-03-10
Fix SQL error on update

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-746" title="DX-746" target="_blank">BugDX-746</a>  CLI- UPG: Getting install log with errors after execution "state update"
  </td></table>
  <br />
 

On update, there is an event that fires after the update is completed. This has been causing an SQL error as the database is closed. This updates the analytics client to not check the config database for every event.


Naatan (approved):
Approving this ahead of @mitchell-as as I'd like to get this into our next beta that I'm cutting pretty much "now".

PR #1750, opened 2022-03-11
Remove implicit rollbar logging

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-717" title="DX-717" target="_blank">StoryDX-717</a>  Remove implicit rollbar reporting from logging lib
  </td></table>
  <br />
 

* Move logging of errors into a new `multilog` package that calls the corresponding `logging.*` and `rollbar.*` functions.
* Added a new `multilog.Log()` function that allows invocation of logging functions manually for special use cases (e.g. log an error without a stack trace).
* Removed `logging.*f()` functions because their return values were not being used at all.
* Fixed numerous instances of incorrect logging usage uncovered by new `multilog` package.


internal/rollbar/rollbar.go
```diff
@@ -61,3 +63,19 @@ func UpdateRollbarPerson(userID, username, email string) {
...
+// Wait is a wrapper around rollbar.Wait().
+func Wait() { rollbar.Wait() }
+
+// Critical is a wrapper around rollbar.Critical().
+func Critical(interfaces ...interface{}) { rollbar.Critical(interfaces...) }
+
+// Error is a wrapper around rollbar.Error().
+func Error(interfaces ...interface{}) { rollbar.Error(interfaces...) }
+
+func handlePanics(err interface{}) {
```
Naatan, line 76: This doesn't seem to be doing anything.

internal/logging/file.go
```diff
@@ -112,64 +108,6 @@ func (l *fileHandler) Emit(ctx *MessageContext, message string, args ...interfac
 
 func (l *fileHandler) emit(ctx *MessageContext, message string, args ...interface{}) {
 	filename := FilePath()
-	originalMessage := fmt.Sprintf(message, args...)
```
Naatan, line 115: There is important logic happening here that doesn't seem replicated in the new implementation, namely:

* Only reporting on "public" channels
* Including the most recent logging messages
* Including the command that was ran along with its flags

Naatan (changes requested):
This isn't convering the second AC:

> We call out to this package explicitly from the main code (rather than implicitly from a sub-package) whenever we want to log something to rollbar

Granted that leaves a lot open to the interpretation. But the intend is that we still report the same stuff to rollbar as before, just not through the logging lib. Normally and `logging.Error` and `logging.Critical` calls would report to rollbar, I think you should review all those instances and add rollbar calls where relevant (likely all instances, but maybe not.. ?).

mitchell-as:
Okay, there are almost 200 instances of `logging.Error()`. It will take time to review and add rollbar calls.

mitchell-as:
Okay, addressed your commits. This still doesn't feel right, but is it along the lines of what you were expecting?

mitchell-as:
This is ready to be looked over again. I wonder if I should rename `logging.Error()` and `logging.Critical()` to something else so that anyone who tries to use it will ultimately use `multilog.Error()` and `multilog.Critical()` instead. There are legitimate uses of those functions by themselves, but as noted in the new doc strings, they are not common.

Naatan (changes requested):
Looking good! I like the `multilog` name; clearly conveys its purpose and why you may want to think twice of using it.

internal/logging/ring_buffer.go
```diff
@@ -0,0 +1,52 @@
+package logging
+
+// Ring buffer structure, but without a start position for reading. All reads are tail reads with
+// no concept for what has already been read. The use case here is for logging to keep track of the
+// last X bytes written.
+type ringBuffer struct {
+	buffer []byte
+	size   int
+	end    int
+	full   bool
```
Naatan, line 10: Can't all of these be extrapolated with `len(buffer)`?

internal/logging/ring_buffer.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func newRingBuffer(size int) *ringBuffer {
+	return &ringBuffer{make([]byte, size), size, 0, false}
+}
+
+// Read reads the last len(p) bytes from the buffer.
+// Typically len(p) should be the size of the buffer.
+// This method satisfies the io.Reader interface.
+func (b *ringBuffer) Read(p []byte) (int, error) {
```
Naatan, line 20: This feels like it's overkill for what we need. Is this just so that we can satisfy io.Reader? What is satisfying that interface giving us?

internal/rollbar/rollbar.go
```diff
@@ -0,0 +1,140 @@
...
+
+func logToRollbar(critical bool, message string, args ...interface{}) {
+	// only log to rollbar when on release, beta or unstable branch and when built via CI (ie., non-local build)
+	isPublicChannel := (constants.BranchName == constants.ReleaseBranch || constants.BranchName == constants.BetaBranch || constants.BranchName == constants.ExperimentalBranch)
+	if !isPublicChannel || !condition.BuiltViaCI() {
+		return
+	}
+
+	data := map[string]interface{}{}
+	logDatab, err := ioutil.ReadFile(logging.FilePath())
```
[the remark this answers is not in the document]
mitchell-as: Okay, done.

Naatan (changes requested):

internal/logging/ring_buffer.go
mitchell-as: No, `len(buffer)` will always return the allocated size, not the number of bytes written to it. `end` revolves around the buffer as writes happen, and `full` indicates whether or not `end` has wrapped once.

internal/logging/ring_buffer.go
mitchell-as: What do you mean by "this" in "Is this just so ..."?

The contents of this method are required for reading correctly from the ring buffer. The signature is simply to satisfy the io.Reader interface because I thought that would be a good idea.

mitchell-as: Fair enough.

Naatan (approved):

PR #1752, opened 2022-03-15
Fix `$INSTALLDIR` casing

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-608" title="DX-608" target="_blank">StoryDX-608</a>  $INSTALLDIR casing should match actual directory path
  </td></table>
  <br />
 

Currently a WIP until I can test it properly

### DIFF
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -1021,9 +1021,14 @@ func ModTime(path string) (time.Time, error) {
 
 func CaseSensitivePath(path string) (string, error) {
 	var searchPath string
+	var err error
 	if runtime.GOOS != "windows" {
 		searchPath = caseSensitiveSearchPath(path)
 	} else {
+		path, err = GetLongPathName(path)
+		if err != nil {
+			return "", errs.Wrap(err, "Could not evaluate long pathname")
+		}
 		volume := filepath.VolumeName(path)
 		remainder := strings.TrimLeft(path, volume)
 		searchPath = filepath.Join(volume, caseSensitiveSearchPath(remainder))
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -1025,10 +1025,6 @@ func CaseSensitivePath(path string) (string, error) {
 	if runtime.GOOS != "windows" {
 		searchPath = caseSensitiveSearchPath(path)
 	} else {
-		path, err = GetLongPathName(path)
-		if err != nil {
-			return "", errs.Wrap(err, "Could not evaluate long pathname")
-		}
 		volume := filepath.VolumeName(path)
 		remainder := strings.TrimLeft(path, volume)
 		searchPath = filepath.Join(volume, caseSensitiveSearchPath(remainder))
--- a/internal/fileutils/fileutils.go
+++ b/internal/fileutils/fileutils.go
@@ -1023,11 +1023,11 @@ func CaseSensitivePath(path string) (string, error) {
 	var searchPath string
 	var err error
 	if runtime.GOOS != "windows" {
-		searchPath = caseSensitiveSearchPath(path)
+		searchPath = globPath(path)
 	} else {
 		volume := filepath.VolumeName(path)
 		remainder := strings.TrimLeft(path, volume)
-		searchPath = filepath.Join(volume, caseSensitiveSearchPath(remainder))
+		searchPath = filepath.Join(volume, globPath(remainder))
 	}
 
 	matches, err := filepath.Glob(searchPath)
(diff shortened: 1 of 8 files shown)


internal/installation/storage/storage.go
```diff
@@ -108,12 +108,24 @@ func CachePath() string {
...
 	} else {
 		cachePath = configdir.New(constants.InternalConfigNamespace, "").QueryCacheFolder().Path
-		if runtime.GOOS == "windows" {
-			// Explicitly append "cache" dir as the cachedir on Windows is the same as the local appdata dir (conflicts with config)
-			cachePath = filepath.Join(cachePath, "cache")
+	}
+
+	configRoot := filepath.Dir(cachePath)
+	entries, err := os.ReadDir(configRoot)
+	if err == nil {
```
MDrakos, line 115: This is to avoid larger refactoring. I would prefer to log this error but importing the logging package here leads to an import cycle. I spent some time trying to fix it but I think that is better left for a different story. If we are going to rely on this `storage` package in many places outside of installation then we should reorganize the layout

internal/installation/storage/storage.go
```diff
@@ -108,12 +108,24 @@ func CachePath() string {
...
+		for _, e := range entries {
+			if strings.EqualFold(e.Name(), constants.InternalConfigNamespace) {
+				cachePath = filepath.Join(configRoot, e.Name())
+				break
+			}
```
Naatan, line 120: This is all to receive the case sensitive variant? It feels a bit expensive and error-prone having to list the parent dir and then retroactively get the right entry based on a string match.

MDrakos: See [this](https://github.com/ActiveState/cli/pull/1752/commits/2df01e5ef43280ec0ef5885884d5a432489c14aa) commit for a rough alternate approach using `filepath.Glob`.

My guess is that attempting to determine the case of a directory on a case-insensitive filesystem will likely amount to searching.

We could dumb this down and change `InternalConfigNamespace` to `"ActiveState"` or I could track down what variable is being used to create the config directory in the first place though that would only take for new installs of the state tool.

If you have any ideas for alternatives let me know.

Naatan: Have you tried to see if using things like `filepath.Abs()` and `filepath.Clean()` might give us the case sensitive variant? And also the `path.*` variants.

MDrakos: Yes, unfortunately, no easy wins there. I can certainly spend some more time digging to see if there is a better way. This approach was just one of the easier ones that came to mind.

Naatan: Ok. Can we make a helper function (eg. `fileutils.CaseSensivePath()`) and only utilize it for the specific INSTALLLDIR use-case?

MDrakos:
Tested this fix locally with a ruby + rails build and I was able to successfully start a rails server

Naatan (changes requested):
I think you missed this part of my review comment:

> and only utilize it for the specific INSTALLLDIR use-case?

So the idea is that we ONLY utilize this for unpacking the INSTALLDIR var, since that's where the issue is centralized.

I'm concerned of the performance impact in putting it on CachePath, as well as any edge case bugs it might create.

MDrakos:
> I think you missed this part of my review comment:
> 
> > and only utilize it for the specific INSTALLLDIR use-case?
> 
> So the idea is that we ONLY utilize this for unpacking the INSTALLDIR var, since that's where the issue is centralized.
> 
> I'm concerned of the performance impact in putting it on CachePath, as well as any edge case bugs it might create.

You're right I missed that. We can now put it in `fileutils` like you originally suggested. I've also changed the strategy to using `glob`. This way we can search the entire path in a case insensitive way but still get the listing that's on disk.

internal/fileutils/fileutils.go
```diff
@@ -1017,3 +1018,38 @@ func ModTime(path string) (time.Time, error) {
...
+	}
+
+	if len(matches) == 0 {
+		return "", errs.New("Could not find path: %s", path)
+	}
+
+	return matches[0], nil
+}
+
+func caseSensitiveSearchPath(path string) string {
```
Naatan, line 1045: Based on your explanation on Slack maybe we should call this `GlobPath()` or `UnixPath()` (latter not being technically correct, but much easier to interpret)?

Naatan (changes requested):

internal/fileutils/fileutils.go
MDrakos: Thank you! I was having trouble thinking of names for this.

Naatan (approved):

PR #1754, opened 2022-03-15
State use

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-680" title="DX-680" target="_blank">StoryDX-680</a>  As a user I can change or unset my default environment so that I have control over which language runtime I'm using
  </td></table>
  <br />


cmd/state/internal/cmdtree/use.go
```diff
@@ -0,0 +1,37 @@
...
+	runner := use.NewUse(prime)
+
+	params := &use.Params{
+		Namespace: &project.Namespaced{},
+	}
+
+	cmd := captain.NewCommand(
+		"use",
+		"",
+		"Switch to using the given project",
```
Naatan, line 21:
```suggestion
		"Use the given project runtime as the default for your system",
```

internal/runners/use/use.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime"
+	"github.com/ActiveState/cli/pkg/platform/runtime/target"
+	"github.com/ActiveState/cli/pkg/project"
+	"github.com/ActiveState/cli/pkg/projectfile"
+)
+
+type Params struct {
+	// TODO: Make owner optional?
```
Naatan, line 33:
```suggestion
```

(Potential) Todo's should be on Jira, not in our code.

internal/runners/use/use.go
```diff
@@ -0,0 +1,170 @@
...
+
+	dir, err := osutils.Getwd()
+	if err != nil {
+		return "", locale.WrapError(err, "err_use_get_wd", "Could not get current working directory")
+	}
+
+	return filepath.Join(dir, namespace.Project), nil
+}
+
+func initProject(ns *project.Namespaced, dir string) (*project.Project, error) {
```
Naatan, line 136: I think this is the same as we have in our activate runner, we should centralize this logic somewhere.

internal/runners/use/use.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if err := u.config.Set("projects.active", proj.Namespace().String()); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (u *Use) getProjectPath(namespace *project.Namespaced) (string, error) {
```
Naatan, line 122: I think this is the same as we have in our activate runner, we should centralize this logic somewhere.

internal/runners/use/use.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := globaldefault.SetupDefaultActivation(u.subshell, u.config, rti, filepath.Dir(proj.Source().Path())); err != nil {
+		return locale.WrapError(err, "err_use_default", "Could not configure your project as the global default.")
+	}
+
+	u.out.Print(fmt.Sprintf(`[NOTICE]Switched to[/RESET] [ACTIONABLE]%s[/RESET]`, params.Namespace.Project))
+
+	if rt.GOOS == "windows" {
+		u.out.Print(locale.Tl("use_reset_notice_windows", "Note you may need to start a new command prompt to reset your environment"))
+	} else {
+		u.out.Print(locale.Tl("use_reset_notice", "Note you may need to run '[ACTIONABLE]hash -r[/RESET]' or start a new shell to reset your environment."))
```
Naatan, line 112:
```suggestion
		u.out.Notice(locale.Tl("use_reset_notice_windows", "Note you may need to start a new command prompt to reset your environment"))
	} else {
		u.out.Notice(locale.Tl("use_reset_notice", "Note you may need to run '[ACTIONABLE]hash -r[/RESET]' or start a new shell to reset your environment."))
```

internal/runners/use/use.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if err := u.config.Set("projects.active", proj.Namespace().String()); err != nil {
```
Naatan, line 115: This should probably happen as part of `globaldefault.SetupDefaultActivation()`

internal/runners/use/use.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if err := globaldefault.SetupDefaultActivation(u.subshell, u.config, rti, filepath.Dir(proj.Source().Path())); err != nil {
+		return locale.WrapError(err, "err_use_default", "Could not configure your project as the global default.")
+	}
+
+	u.out.Print(fmt.Sprintf(`[NOTICE]Switched to[/RESET] [ACTIONABLE]%s[/RESET]`, params.Namespace.Project))
+
+	if rt.GOOS == "windows" {
+		u.out.Print(locale.Tl("use_reset_notice_windows", "Note you may need to start a new command prompt to reset your environment"))
```
Naatan, line 110:
```suggestion
		u.out.Print(locale.Tl("use_reset_notice_windows", "Note you may need to start a new command prompt to fully update your environment."))
```

internal/runners/use/use.go
Naatan, line 112:
```suggestion
		u.out.Print(locale.Tl("use_reset_notice", "Note you may need to run '[ACTIONABLE]hash -r[/RESET]' or start a new shell to fully update your environment."))
```

Naatan (changes requested):

internal/runners/use/use.go
MDrakos: Do we also want to set this config key on regular activation? Or is this just for defaults?

Naatan: Just for defaults.

internal/globaldefault/default.go
```diff
@@ -67,14 +69,19 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
...
 	if err := fw.Update(exes); err != nil {
 		return locale.WrapError(err, "err_globaldefault_fw", "Could not set up forwarders")
 	}
 
-	if err := cfg.Set(constants.GlobalDefaultPrefname, projectPath); err != nil {
+	if err := cfg.Set(constants.GlobalDefaultPrefname, projectDir); err != nil {
 		return locale.WrapError(err, "err_set_default_config", "Could not set default project in config file")
 	}
 
+	if err := cfg.Set(constants.ActiveProjectConfig, proj.Namespace().String()); err != nil {
```
Naatan, line 82: hmm, do we need both? Technically the project dir could be used to receive the namespace.

We should maybe at least ensure that they match in terms of their pattern, eg. `projects.active.path` and `projects.active.namespace`.

internal/runners/activate/activate.go
```diff
@@ -72,8 +72,7 @@ type primeable interface {
 
 func NewActivate(prime primeable) *Activate {
 	return &Activate{
-		NewNamespaceSelect(prime.Config()),
-		NewCheckout(git.NewRepo(), prime),
```
Naatan, line 76: I think we can drop `NewCheckout`?

Naatan (changes requested):

internal/globaldefault/default.go
MDrakos: What do we intend to use `projects.active.namespace` for? There isn't currently one in the code.

The project path only has the project name. We would have to parse the projectfile in order to get the namespace, but it's certainly feasible.

internal/runners/activate/activate.go
MDrakos: Yes, `NewCheckout` is gone in favor of `checkout.New()`

internal/globaldefault/default.go
Naatan: I meant we should replace the value used under `constants.GlobalDefaultPrefname` with `projects.active.namespace`. But only if we actually need it. I still think we might be able to just get rid of it and rely solely on the path.

Naatan (changes requested):
.

internal/runners/activate/activate.go
Naatan: Its use is, but so far as I can tell from the PR diff the function definition wasn't deleted.

internal/globaldefault/default.go
MDrakos: I don't see a clear use case for having the namespace in the config at the moment so I've removed it in favor of the project path which we are already using.

internal/runners/activate/activate.go
MDrakos: The file `internal/runners/activate/checkout.go` was deleted. I think the GitHub diff defaults to not actually showing that file, just that it was deleted.

Naatan (approved):
Nice work!

PR #1756, opened 2022-03-18
Fixed display of commit history despite committers having been deleted by their orgs

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-713" title="DX-713" target="_blank">BugDX-713</a>  `state history` breaks if org members are deleted
  </td></table>
  <br />
 

Show a placeholder username instead of erroring out.

For example:
```
  Commit     218dd5ee-cba7-43ca-a926-536610daf0dc                       
  Author     <deleted>                                                  
  Date       25 Apr 20 00:29 UTC                                        
  Message    Manual update by ActiveState support.
```

I also forked ActiveState/cli and reverted to Aleks' commit, which did not error:

```
█ Reverting Commit

You are about to revert to the following commit:
  Commit     218dd5ee-cba7-43ca-a926-536610daf0dc   
  Author     <deleted>                              
  Date       25 Apr 20 00:29 UTC                    
  Message    Manual update by ActiveState support.  

[...]

Sucessfully reverted to commit: 218dd5ee-cba7-43ca-a926-536610daf0dc
```

### DIFF
--- a/pkg/cmdlets/commit/commit.go
+++ b/pkg/cmdlets/commit/commit.go
@@ -16,10 +16,6 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/model"
 )
 
-// AuthorDeleted is the placeholder username to use if an author has since been deleted from an
-// organization and their username can no longer be obtained.
-const AuthorDeleted = "<deleted>"
-
 type commitData struct {
 	Hash    string   `locale:"hash,[HEADING]Commit[/RESET]"`
 	Author  string   `locale:"author,[HEADING]Author[/RESET]"`
@@ -78,8 +74,7 @@ func commitDataFromCommit(commit *mono_models.Commit, orgs []gmodel.Organization
 	if commit.Author != nil && orgs != nil {
 		username, err = usernameForID(*commit.Author, orgs)
 		if err != nil {
-			logging.Debug("Could not determine username for commit author '%s'. Using placeholder value '%s'.", *commit.Author, AuthorDeleted)
-			username = AuthorDeleted
+			return commitData{}, locale.WrapError(err, "err_commit_print_username", "Could not determine username for commit author")
 		}
 	}
 
@@ -173,5 +168,7 @@ func usernameForID(id strfmt.UUID, orgs []gmodel.Organization) (string, error) {
 		}
 	}
 
-	return "", locale.NewError("err_user_not_found", id.String())
+	placeholder := locale.Tl("deleted_username", "<deleted>")
+	logging.Debug("Could not determine username for commit author '%s'. Using placeholder value '%s'.", id, placeholder)
+	return placeholder, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/cmdlets/commit/commit.go
```diff
@@ -8,13 +8,18 @@ import (
...
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/output"
 	gmodel "github.com/ActiveState/cli/pkg/platform/api/graphql/model"
 	"github.com/ActiveState/cli/pkg/platform/api/mono/mono_models"
 	"github.com/ActiveState/cli/pkg/platform/model"
 )
 
+// AuthorDeleted is the placeholder username to use if an author has since been deleted from an
+// organization and their username can no longer be obtained.
+const AuthorDeleted = "<deleted>"
```
Naatan, line 21:
```suggestion
const AuthorDeleted = locale.Tl("deleted_username", "<deleted>")
```

pkg/cmdlets/commit/commit.go
```diff
@@ -73,7 +78,8 @@ func commitDataFromCommit(commit *mono_models.Commit, orgs []gmodel.Organization
 	if commit.Author != nil && orgs != nil {
 		username, err = usernameForID(*commit.Author, orgs)
 		if err != nil {
-			return commitData{}, locale.WrapError(err, "err_commit_print_username", "Could not determine username for commit author")
+			logging.Debug("Could not determine username for commit author '%s'. Using placeholder value '%s'.", *commit.Author, AuthorDeleted)
+			username = AuthorDeleted
```
Naatan, line 82: This logic should be contained within `userrnameForID`, or alternatively it should return an identifiable error so that we can fall back. `err != nil` could have any number of causes as far as this code is concerned.

Naatan (changes requested):

pkg/cmdlets/commit/commit.go
mitchell-as: Okay, but this will have to be inside a function since it's not available at compile time. As a result, this constant will be removed and any outside packages that care about this value will need to use the locale module to get it.

pkg/cmdlets/commit/commit.go
mitchell-as: Fair enough.

mitchell-as:
<ping>

pkg/cmdlets/commit/commit.go
```diff
@@ -167,5 +168,7 @@ func usernameForID(id strfmt.UUID, orgs []gmodel.Organization) (string, error) {
 		}
 	}
 
-	return "", locale.NewError("err_user_not_found", id.String())
+	placeholder := locale.Tl("deleted_username", "<deleted>")
+	logging.Debug("Could not determine username for commit author '%s'. Using placeholder value '%s'.", id, placeholder)
+	return placeholder, nil
```
Naatan, line 173: Doesn't look like this function ever returns an `error` anymore, so might as well drop that.

mitchell-as: Thanks, good catch.

Naatan (approved):

PR #1758, opened 2022-03-21
Recognize '-h' and '--help' for `state exec`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-197" title="DX-197" target="_blank">BugDX-197</a>  state exec help text is wrong
  </td></table>
  <br />
 

Similar to `state run`, which has the same argument structure.

### DIFF
--- a/cmd/state/internal/cmdtree/exec.go
+++ b/cmd/state/internal/cmdtree/exec.go
@@ -28,6 +28,11 @@ func newExecCommand(prime *primer.Values, args ...string) *captain.Command {
 		},
 		[]*captain.Argument{},
 		func(ccmd *captain.Command, args []string) error {
+			if len(args) > 0 && (args[0] == "-h" || args[0] == "--help") {
+				prime.Output().Print(ccmd.UsageText())
+				return nil
+			}
+
 			return runner.Run(params, args...)
 		},
 	)


MDrakos (changes requested):
This is different from the AC on the ticket:

> Change "need more help" prompt to use the `state help exec` variant (instead of using the flag).

I think the reason for this is that the executors use `state exec`. So we don't want someone running `python3 --help` to get the help output from the state tool. The `--help` flag needs to be forwarded to the underlying command.

mitchell-as:
I thought it about it and making `state exec` a special case here: https://github.com/ActiveState/cli/blob/d175aa23dee18158523757bb27b84c808143e74c/cmd/state/main.go#L261 did not feel like a good solution. With this PR, `state exec` is only checking the first argument for `-h` or `--help`, just like `state run` (https://github.com/ActiveState/cli/blob/d175aa23dee18158523757bb27b84c808143e74c/cmd/state/internal/cmdtree/run.go#L29), so `state exec python3 --help` would pass `--help` to `python3` as expected.

MDrakos (approved):
Nice! Makes sense to me and this also makes it that `state exec` is in alignment with our other commands. It still might be worth checking that this interpretation of the AC is okay before you merge.

PR #1764, opened 2022-03-21
Fixed spelling errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-802" title="DX-802" target="_blank">BugDX-802</a>  CLI - HELP: Typo error for "state deploy --help" - instead of "run" it is "ran".
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1436,11 +1436,11 @@ deploy_cmd_description:
 deploy_install_cmd_description:
   other: Only install the runtime environment files, but do not configure it in any way (effectively caches the runtime environment)
 deploy_configure_cmd_description:
-  other: Configure the runtime environment for your current shell (must have run install first or this will err)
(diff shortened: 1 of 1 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1436,11 +1436,11 @@ deploy_cmd_description:
 deploy_install_cmd_description:
   other: Only install the runtime environment files, but do not configure it in any way (effectively caches the runtime environment)
 deploy_configure_cmd_description:
-  other: Configure the runtime environment for your current shell (must have ran install first or this will err)
+  other: Configure the runtime environment for your current shell (must have run install first or this will err)
```
MDrakos, line 1439: While we're here can we change this as well?

```suggestion
  other: Configure the runtime environment for your current shell (must have run install first or this will error)
```

MDrakos (changes requested):

MDrakos (approved):

PR #1766, opened 2022-03-22
Fixed multi-add of the same bin path in ENV["PATH"] when more than one exclusion path is present.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-798" title="DX-798" target="_blank">BugDX-798</a>  CLI - Deploy: Binary directories provided same path twice and added twice to the PATH env variable.
  </td></table>
  <br />
 

.

### DIFF
--- a/pkg/platform/runtime/envdef/environment.go
+++ b/pkg/platform/runtime/envdef/environment.go
@@ -369,10 +369,14 @@ func FilterPATH(env map[string]string, excludes ...string) {
 	paths := strings.Split(PATH, string(os.PathListSeparator))
 	for _, p := range paths {
 		pc := filepath.Clean(p)
+		include_path := true
 		for _, exclude := range excludes {
 			if pc == filepath.Clean(exclude) {
-				continue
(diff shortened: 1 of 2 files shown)


pkg/platform/runtime/envdef/environment.go
```diff
@@ -369,10 +369,14 @@ func FilterPATH(env map[string]string, excludes ...string) {
 	paths := strings.Split(PATH, string(os.PathListSeparator))
 	for _, p := range paths {
 		pc := filepath.Clean(p)
+		include_path := true
```
MDrakos, line 372: The Go convention is camel case.

MDrakos (changes requested):

pkg/platform/runtime/envdef/environment.go
mitchell-as: Whoops, sorry about that.

MDrakos (approved):

PR #1769, opened 2022-03-23
Fix uninstall issues

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-799" title="DX-799" target="_blank">BugDX-799</a>  CLI - Uninstall: "state clean uninstall" not cleaning all files and not deleting paths from PATH env variable.
  </td></table>
  <br />


cmd/state-installer/installer.go
```diff
@@ -97,6 +97,11 @@ func (i *Installer) Install() (rerr error) {
 		return errs.Wrap(err, "Could not update PATH")
 	}
 
+	err = i.cfg.Set(constants.InstalledAdminConfig, isAdmin)
+	if err != nil {
+		return errs.Wrap(err, "Failed to set current privilege level in config")
+	}
```
Naatan, line 103:
```suggestion
	err = i.cfg.Set(constants.InstalledAdminConfig, isAdmin)
	if err != nil {
		return errs.Wrap(err, "Failed to set current privilege level in config")
	}
```

This doesn't seem viable since we expose `state clean config` and `state clean uninstall` as separate steps. Meaning you can clean your config, which would purge this value, and the run uninstall which would no longer have knowledge over how it was installed.

Since this is Windows only I think either we save it to the registry, or potentially we could just check the ownership on the state.exe?

internal/analytics/client/async/client.go
```diff
@@ -91,6 +92,10 @@ func (a *Client) Wait() {
 }
 
 func (a *Client) sendEvent(category, action, label string, dims ...*dimensions.Values) error {
+	if a.closed {
+		return nil
```
Naatan, line 96: Should debug log this.

internal/runners/clean/run_lin_mac.go
```diff
@@ -127,5 +127,15 @@ func removeInstall(cfg configurable) error {
 		aggErr = errs.Wrap(aggErr, "Failed to remove system files at %s: %v", appPath, err)
 	}
 
+	installPath, err := installation.InstallPath()
+	if err != nil {
+		aggErr = errs.Wrap(aggErr, "Could not get installation path")
+	}
+
+	err = os.RemoveAll(installPath)
```
Naatan, line 135: Can we make this only remove all if the path is empty? Otherwise pring a warning (don't error).

I'm really scared otherwise of how this would combine with people installing to custom dirs, or legacy installations that installed to eg. `/usr/local/bin`. I'd like to stay away from removing dirs recursively as much as possible.

internal/subshell/cmd/cmd.go
```diff
@@ -96,7 +97,12 @@ func (v *SubShell) CleanUserEnv(cfg sscommon.Configurable, envType sscommon.RcId
 	// Clean up old entries
 	oldEnv := cfg.GetStringMap(envType.Key)
 	for k, v := range oldEnv {
-		if err := cmdEnv.unset(k, v.(string)); err != nil {
+		val, ok := v.(string)
+		if !ok {
+			logging.Debug("Invalid non-string value in environment mapping")
+			continue
+		}
```
Naatan, line 104: You ran into this? What was the value? I would think there's no version of this where the values aren't strings.

internal/subshell/cmd/env.go
```diff
@@ -36,28 +38,44 @@ func getEnvironmentPath(userScope bool) string {
...
 
-	// Delete value
-	return key.DeleteValue(name)
+	return nil
+}
+
+func cleanPath(keyValue, oldEntry string) string {
+	var newValue []string
+	for _, entry := range strings.Split(keyValue, string(os.PathListSeparator)) {
+		if entry == oldEntry {
```
Naatan, line 73: We should wrap these in `filepath.Clean()`.

Naatan (changes requested):

internal/runners/clean/run_lin_mac.go
MDrakos: Good point!

cmd/state-installer/installer.go
MDrakos: Checking the ownership of the file requires us to interact with the Windows ACL so I opted to store this in the registry.

internal/subshell/cmd/cmd.go
MDrakos: I didn't run into this but it is a potential panic that we can handle here. I think this is better than relying on the `map[string]interface{}` being what we expect.

internal/installation/paths.go
```diff
@@ -20,6 +20,9 @@ const CfgTransitionalStateToolPath = "transitional_installation_path"
 
 const BinDirName = "bin"
 
+// AdminInstallRegistry is the registry key name used to determine if the State Tool was installed as administrator
+const AdminInstallRegistry = "Installed As Admin"
```
Naatan, line 24: These can have spaces?

internal/installation/paths_windows.go
```diff
@@ -13,3 +13,7 @@ func defaultSystemInstallPath() (string, error) {
 	// There is no system install path for Windows
 	return "", nil
 }
+
+func InstallRegistryKeyPath() string {
+	return `SOFTWARE\ActiveState\install`
+}
```
Naatan, line 19: Can we make this a constant? Note it's only used on Windows so doesn't need to exist on Linux/Mac.

internal/runners/clean/run_lin_mac.go
```diff
@@ -32,6 +34,7 @@ func (u *Uninstall) runUninstall() error {
 	}
 
 	err = removeInstall(u.cfg)
+	fmt.Println("Remove err:", errs.JoinMessage(err))
```
Naatan, line 37: Debugging code?

internal/runners/clean/run_win.go
```diff
@@ -129,3 +132,44 @@ func removePaths(logFile string, paths ...string) error {
 
 	return nil
 }
+
+func checkAdmin() error {
+	doAdminCheck := true
+	installedAsAdmin, err := getAdminInstall()
+	if err != nil {
+		doAdminCheck = false
+		multilog.Error("Could not check if initial installation was run as admin, error: %v", err)
```
Naatan, line 141: This seems like an actual error, let's return it.

internal/runners/clean/run_win.go
```diff
@@ -129,3 +132,44 @@ func removePaths(logFile string, paths ...string) error {
...
+	}
+
+	isAdmin, err := osutils.IsAdmin()
+	if err != nil {
+		doAdminCheck = false
+		multilog.Error("Could not check if current user is an administrator, error: %v", err)
```
Naatan, line 147: Again, seems like a legitimate error.

internal/runners/clean/run_win.go
```diff
@@ -129,3 +132,44 @@ func removePaths(logFile string, paths ...string) error {
 
 	return nil
 }
+
+func checkAdmin() error {
```
Naatan, line 136: This either needs a comment or a more descriptive name (.. or both).

internal/runners/clean/run_win.go
```diff
@@ -129,3 +132,44 @@ func removePaths(logFile string, paths ...string) error {
...
+	}
+
+	if doAdminCheck && installedAsAdmin && !isAdmin {
+		return locale.NewInputError("err_uninstall_privlege_mismatch")
+	}
+
+	return nil
+}
+
+func getAdminInstall() (bool, error) {
```
Naatan, line 157: I think we should centralize the behavior around storing and retrieving this state in the same package. Maybe `installation`?

Naatan (changes requested):

internal/runners/clean/run_win.go
MDrakos: That's a really good idea!

internal/runners/clean/run_win.go
MDrakos: Agreed. Naming is hard 😩

internal/installation/paths.go
MDrakos: Yep!

internal/runners/clean/run_win.go
MDrakos: I guess auto-update will create the registry key even for old installations of the state tool which is the use case I was trying to be mindful of.

MDrakos: Though this will mean that if the key isn't there for whatever reason then the user will not be able to uninstall

Naatan: The error handling for auto updating is centralized in the auto updating mechanic though, you don't have to worry about that here.

The key not being there should probably be a use-case you handle explicitly. But `err != nil` is too vague in that regard.

internal/installation/context.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/ActiveState/cli/internal/errs"
+)
+
+type Context struct {
+	InstalledAsAdmin bool
+}
+
+func GetContext() (*Context, error) {
+	installedAsAdmin, err := getAdminInstallInformation()
```
Naatan, line 12: Nit: Why not implement this logic here? You already have platform specific files, just implement `GetContext()` specifically per platform. You're effecively already doing this through `getAdminInstallInformation()`.

internal/installation/context_windows.go
```diff
@@ -0,0 +1,57 @@
...
+
+	err = key.SetStringValue(adminInstallRegistry, strconv.FormatBool(context.InstalledAsAdmin))
+	if err != nil {
+		return errs.Wrap(err, "Could not set registry key value")
+	}
+
+	return nil
+}
+
+func getAdminInstallInformation() (bool, error) {
```
Naatan, line 39: As you noted this will error if the key doesn't exist. I think we should return an identifiable error for this and have the consumer check for this, and if the key was not found - assume we installed as user.

Naatan (approved):

PR #1771, opened 2022-03-24
Only allow setting config values for known/registered keys.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-809" title="DX-809" target="_blank">StoryDX-809</a>  `state config` only acts on recognized keys
  </td></table>
  <br />
 

Eligible keys should be registered using the config mediator, as is already done in *internal/captain/command.go*, and *cmd/state-svc/manager.go*.


internal/captain/command.go
```diff
@@ -33,6 +33,7 @@ import (
 
 func init() {
 	configMediator.NewRule(constants.UnstableConfig, configMediator.Bool, configMediator.EmptyEvent, configMediator.EmptyEvent)
+	configMediator.NewRule(constants.AutoUpdateConfigKey, configMediator.Bool, configMediator.EmptyEvent, configMediator.EmptyEvent)
```
Naatan, line 36: This should go in `cmd/state/autoupdate.go`. The reason we architected it this way is so that the package that utilizes the key is also the one registering it.

internal/mediators/config/rules.go
```diff
@@ -26,12 +26,14 @@ type Rules map[string]Rule
...
+// GetRule returns the rule for the given key, or the default rule if none exists.
+// Also returns whether or not a rule was found.
+func GetRule(key string) (Rule, bool) {
 	rule, ok := rules[key]
 	if !ok {
-		return defaultRule
+		return defaultRule, false
 	}
-	return rule
+	return rule, true
```
Naatan, line 36: I don't think this is adding anything useful. Now we need to check the bool and have specialized knowledge in the consumer. The previous logic already covered this use case by not actually doing anything with the rule other than satisfying the expectations of the returned type.

internal/runners/config/set.go
```diff
@@ -28,7 +27,10 @@ func NewSet(prime primeable) *Set {
 func (s *Set) Run(params SetParams) error {
 	// Cast to rule type if applicable
 	var value interface{}
-	rule := configMediator.GetRule(params.Key.String())
+	rule, found := configMediator.GetRule(params.Key.String())
+	if !found {
+		return locale.NewInputError("unknown_config_key", "Unknown config key: {{.V0}}", params.Key.String())
+	}
```
Naatan, line 33: I think what we want here is a `configMediator.KnownRule` or something like that. This is overloading the use-case for `GetRule()`.

Naatan (changes requested):
In addition to my comments, I think this story calls for a rename of the "Rule" type into something slightly less opinionated like "Registry" or something like that. The previous use-case was strictly for enforcing behaviors on config keys, whereas the new use-case expands this to also cover whether certain config keys exist or not.

internal/captain/command.go
mitchell-as: What is this magic? I can just add an `init()` function to *autoupdate.go* and it magically gets called? I mean, I just tried it and it appears to work, but ??!!

internal/mediators/config/rules.go
mitchell-as: Okay.

Naatan (changes requested):
Beautiful.

Naatan (approved):
Meant to approve..

PR #1772, opened 2022-03-25
Fixed help description for "state pull --set-project"

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-813" title="DX-813" target="_blank">BugDX-813</a>  CLI - PULL: Help output has descriptions starting with NON capitalized letters and part of the descriptions looked off.
  </td></table>
  <br />
 

Also replace `state pull --force` with `state pull --non-interactive`. Force was changed 9 months ago to be simply a non-interactive type of option. A merge is still required, and if the target project is completely unrelated, the pull can still fail.

This PR will also address https://activestatef.atlassian.net/browse/DX-817

I also went through all of our "*_description" localization strings and they all appear to be capitalized properly. This looks like the only one.

### DIFF
--- a/cmd/state/internal/cmdtree/pull.go
+++ b/cmd/state/internal/cmdtree/pull.go
@@ -11,7 +11,6 @@ func newPullCommand(prime *primer.Values, globals *globalOptions) *captain.Comma
 	runner := pull.New(prime)
 
 	params := &pull.PullParams{}
-	params.Force = globals.NonInteractive
 
 	return captain.NewCommand(
 		"pull",
@@ -28,6 +27,7 @@ func newPullCommand(prime *primer.Values, globals *globalOptions) *captain.Comma
 		},
 		[]*captain.Argument{},
 		func(cmd *captain.Command, args []string) error {
+			params.Force = globals.NonInteractive
 			return runner.Run(params)
 		}).SetGroup(VCSGroup)
 }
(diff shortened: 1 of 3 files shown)


daved (dismissed):

cmd/state/internal/cmdtree/pull.go
```diff
@@ -18,21 +18,16 @@ func newPullCommand(prime *primer.Values) *captain.Command {
...
 				Name:        "set-project",
 				Shorthand:   "",
-				Description: locale.Tl("flag_state_pull_set_project_description", "project even if it is unrelated to checked out one"),
+				Description: locale.Tl("flag_state_pull_set_project_description", "Pull from the specified project instead of the checked out one"),
 				Value:       &params.SetProject,
 			},
 		},
 		[]*captain.Argument{},
 		func(cmd *captain.Command, args []string) error {
+			params.Force = globals.NonInteractive
```
daved, line 30: nit: If this doesn't need to happen within the handler itself, it is probably more appropriate to set this in the declaration at line 13.

daved: oof, nm, I didn't think I was late enough on this review to check if it was already completed. Sorry about that.

mitchell-as: I had it there, but it failed the unit tests because I think the globals contents are edited after declaration.

PR #1783, opened 2022-03-31
Clarified how to activate another branch.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-795" title="DX-795" target="_blank">BugDX-795</a>  CLI - Activate: Any attempts to use "--branch" flag are fail with the  same error message
  </td></table>
  <br />

### DIFF
--- a/internal/runners/activate/activate.go
+++ b/internal/runners/activate/activate.go
@@ -147,12 +147,7 @@ func (r *Activate) run(params *ActivateParams) error {
 		}
 
 		if params.Branch != proj.BranchName() {
-			return locale.NewInputError(
-				"err_conflicting_branch_while_checkedout",
-				"Cannot activate branch [NOTICE]{{.V0}}[/RESET]. Branch [NOTICE]{{.V1}}[/RESET] is already checked out.\n"+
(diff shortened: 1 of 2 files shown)


internal/runners/activate/activate.go
```diff
@@ -149,7 +149,8 @@ func (r *Activate) run(params *ActivateParams) error {
 		if params.Branch != proj.BranchName() {
 			return locale.NewInputError(
 				"err_conflicting_branch_while_checkedout",
-				"Cannot activate branch [NOTICE]{{.V0}}[/RESET]; Branch [NOTICE]{{.V1}}[/RESET] is already checked out.",
+				"Cannot activate branch [NOTICE]{{.V0}}[/RESET]. Branch [NOTICE]{{.V1}}[/RESET] is already checked out.\n"+
+					"Please use [ACTIONABLE]state branch switch <name>[/RESET] followed by [ACTIONABLE]state activate[/RESET] instead.",
```
MDrakos, line 153: The comment makes sense but if it's going to be multiple lines or just longer than others we should move it to the localization file.

MDrakos (changes requested):

PR #1784, opened 2022-03-31
Install logic ONLY supports our new installation format

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-790" title="DX-790" target="_blank">StoryDX-790</a>  Install logic ONLY supports our new installation format
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -245,7 +245,7 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 	}
 
 	// Detect installed state tool
-	stateToolInstalled, stateToolPath, err := installation.InstalledOnPath(params.path)
+	stateToolInstalled, installPath, err := installation.InstalledOnPath(params.path)
 	if err != nil {
 		return errs.Wrap(err, "Could not detect if State Tool is already installed.")
 	}
@@ -300,7 +300,7 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 	// Check if state tool already installed
 	if !params.force && stateToolInstalled {
 		logging.Debug("Cancelling out because State Tool is already installed")
-		out.Print(fmt.Sprintf("State Tool Package Manager is already installed at [NOTICE]%s[/RESET]. To reinstall use the [ACTIONABLE]--force[/RESET] flag.", stateToolPath))
+		out.Print(fmt.Sprintf("State Tool Package Manager is already installed at [NOTICE]%s[/RESET]. To reinstall use the [ACTIONABLE]--force[/RESET] flag.", installPath))
 		an.Event(AnalyticsFunnelCat, "already-installed")
 		return postInstallEvents(out, cfg, an, params, true)
 	}
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -242,23 +242,22 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		if err != nil {
 			return errs.Wrap(err, "Could not detect installation path.")
 		}
-	} else {
-		if fileutils.DirExists(params.path) {
-			empty, err := fileutils.IsEmptyDir(params.path)
-			if err != nil {
-				return errs.Wrap(err, "Could not check if install path is empty")
-			}
-			if !empty {
-				return locale.NewInputError("err_install_nonempty_dir", "Installation path must be an empty directory")
-			}
-		}
 	}
 
 	// Detect installed state tool
 	stateToolInstalled, stateToolPath, err := installation.InstalledOnPath(params.path)
-	fmt.Println("state tool installed:", stateToolInstalled)
 	if err != nil {
 		return errs.Wrap(err, "Could not detect if State Tool is already installed.")
+	}
+
+	if !stateToolInstalled && fileutils.DirExists(params.path) {
+		empty, err := fileutils.IsEmptyDir(params.path)
+		if err != nil {
+			return errs.Wrap(err, "Could not check if install path is empty")
+		}
+		if !empty {
+			return locale.NewInputError("err_install_nonempty_dir", "Installation path must be an empty directory")
+		}
 	}
 
 	// Detect state tool alongside installer executable
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -236,7 +236,6 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		targetBranch = constants.BranchName
 	}
 
-	fmt.Println("Initial path:", params.path)
 	if params.path == "" {
 		var err error
 		params.path, err = installation.InstallPathForBranch(targetBranch)
@@ -289,7 +288,6 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 	}
 	an.Event(AnalyticsFunnelCat, route)
 
-	fmt.Printf("Parmas: %+v\n", *params)
 	// if sourcePath was provided we're already using the right installer, so proceed with installation
 	if params.sourcePath != "" {
 		if err := installOrUpdateFromLocalSource(out, cfg, an, params, isUpdate); err != nil {
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -253,13 +253,25 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 	// If this is a fresh installation we ensure that the target directory is empty
 	if !stateToolInstalled && fileutils.DirExists(params.path) {
 		fmt.Println("Checking dir:", params.path)
-		empty, err := fileutils.IsEmptyDir(params.path)
+		// empty, err := fileutils.IsEmptyDir(params.path)
+		// if err != nil {
+		// 	return errs.Wrap(err, "Could not check if install path is empty")
+		// }
+		files, err := os.ReadDir(params.path)
 		if err != nil {
-			return errs.Wrap(err, "Could not check if install path is empty")
+			return errs.Wrap(err, "Could not read install path")
 		}
-		if !empty {
-			return locale.NewInputError("err_install_nonempty_dir", "Installation path must be an empty directory")
+
+		for _, file := range files {
+			if isStateExecutable(strings.ToLower(file.Name())) || strings.ToLower(file.Name()) == installation.InstallDirMarker {
+				continue
+			} else {
+				return errs.Wrap(err, "Installation directory contains unknown file")
+			}
 		}
+		// if !empty {
+		// 	return locale.NewInputError("err_install_nonempty_dir", "Installation path must be an empty directory")
+		// }
 	}
 
 	// Detect state tool alongside installer executable
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -250,9 +250,9 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		return errs.Wrap(err, "Could not detect if State Tool is already installed.")
 	}
 
+	fmt.Println("installDir:", params.path)
 	// If this is a fresh installation we ensure that the target directory is empty
 	if !stateToolInstalled && fileutils.DirExists(params.path) {
-		fmt.Println("Checking dir:", params.path)
 		// empty, err := fileutils.IsEmptyDir(params.path)
 		// if err != nil {
 		// 	return errs.Wrap(err, "Could not check if install path is empty")
@@ -266,7 +266,7 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 			if isStateExecutable(strings.ToLower(file.Name())) || strings.ToLower(file.Name()) == installation.InstallDirMarker {
 				continue
 			} else {
-				return errs.Wrap(err, "Installation directory contains unknown file")
+				return errs.Wrap(err, "Installation path must be an empty directory")
 			}
 		}
 		// if !empty {
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -236,7 +236,6 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		targetBranch = constants.BranchName
 	}
 
-	fmt.Println("path param:", params.path)
 	if params.path == "" {
 		var err error
 		params.path, err = installation.InstallPathForBranch(targetBranch)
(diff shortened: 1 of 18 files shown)


cmd/state-installer/cmd.go
```diff
@@ -250,6 +250,16 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		return errs.Wrap(err, "Could not detect if State Tool is already installed.")
 	}
 
+	if !stateToolInstalled && fileutils.DirExists(params.path) {
```
Naatan, line 253: Please add a comment above, the installer in particular is very much geared towards outsiders reading this code.

cmd/state-installer/installer_lin_mac.go
```diff
@@ -0,0 +1,34 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/fileutils"
+)
+
+func (i *Installer) cleanInstallPath() error {
```
Naatan, line 16: We should make sure here that we don't need to worry about the additional application files that are created on macOS.

cmd/state-installer/installer_windows.go
```diff
@@ -84,3 +72,67 @@ func (i *Installer) PrepareBinTargets(useBinDir bool) error {
 
 	return nil
 }
+
+func (i *Installer) cleanInstallPath() error {
+	if fileutils.DirExists(i.path) {
```
Naatan, line 77: Return early

cmd/state-installer/installer_windows.go
```diff
@@ -84,3 +72,67 @@ func (i *Installer) PrepareBinTargets(useBinDir bool) error {
...
+	if err != nil {
+		return errs.Wrap(err, "Could not determine if running as Windows administrator")
+	}
+
+	// Since we are repairing a corrupted install we need to also remove the old
+	// PATH entry. The new PATH entry will be added later in the install/update process.
+	// This is only an issue on Windows as on other platforms we can simply rewrite
+	// the PATH entry.
+	s := subshell.New(i.cfg)
+	if err := s.CleanUserEnv(i.cfg, sscommon.InstallID, !isAdmin); err != nil {
```
Naatan, line 107: This should use the `isAdmin` value that we record at install time (ie. if the user installed State Tool as admin then we should also try to remove env vars here as admin, and if they aren't running as admin then we error out).

We should probably do this check WAY at the start of the updater. Meaning if you installed as admin then you HAVE to update as admin.

cmd/state-installer/installer_windows.go
```diff
@@ -84,3 +72,67 @@ func (i *Installer) PrepareBinTargets(useBinDir bool) error {
...
+		return errs.Wrap(err, "Failed to State Tool installation PATH")
+	}
+
+	return nil
+}
+
+func removeOldExecutables(dir string) error {
+	if fileutils.TargetExists(dir) {
```
Naatan, line 115: Return early

cmd/state-installer/installer_windows.go
```diff
@@ -84,3 +72,67 @@ func (i *Installer) PrepareBinTargets(useBinDir bool) error {
...
+				continue
+			}
+
+			if strings.HasSuffix(file.Name(), ".old") {
+				logging.Debug("Deleting old file: %s", file.Name())
+				oldFile := filepath.Join(dir, file.Name())
+				if err := os.Remove(oldFile); err != nil {
+					multilog.Error("Failed to remove old executable: %s. Error: %s", oldFile, errs.JoinMessage(err))
+				}
+				continue
```
Naatan, line 132: Doesn't serve any purpose.

internal/appinfo/appinfo.go
```diff
@@ -22,6 +22,10 @@ type AppInfo struct {
 func execDir(baseDir ...string) (resultPath string) {
 	defer func() {
 		// Account for legacy use-case that wasn't using the correct bin dir
```
Naatan, line 24: One of the purposes of this story is that we can get rid of this legacy handling logic. So I would expect to see this defer gone as part of this PR, rather than added to.

internal/installation/paths.go
```diff
@@ -9,16 +10,23 @@ import (
...
 
-// CfgTransitionalStateToolPath is the configuration key for the path where a transitional State Tool might still be stored
-const CfgTransitionalStateToolPath = "transitional_installation_path"
+	// CfgTransitionalStateToolPath is the configuration key for the path where a transitional State Tool might still be stored
+	CfgTransitionalStateToolPath = "transitional_installation_path"
 
-const BinDirName = "bin"
+	BinDirName = "bin"
+
+	InstallDirMarker = ".install_root"
```
Naatan, line 26: We should probably name this something more activestate specific. eg. `.state_install_root`.

internal/installation/paths.go
```diff
@@ -63,11 +67,18 @@ func BinPathFromInstallPath(installPath string) (string, error) {
 }
 
 func InstalledOnPath(installPath string) (bool, string, error) {
```
Naatan, line 69: This seems to be containing legacy logic again. As per our new logic the path would ALWAYS be `installPath/bin`, so we don't need to be looking at the StateApp exec path.

If we rely on this function for detecting legacy executables then we should split that off into separate logic that is ONLY used for the update logic. We don't want to sort of "infect" our main application logic with this legacy handling.

internal/installation/paths.go
```diff
@@ -77,3 +88,65 @@ func LauncherInstallPath() (string, error) {
...
+	}
+
+	// Detect if the install dir has files in it
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return errs.Wrap(err, "Could not read directory: %s", path)
+	}
+
+	for _, file := range files {
+		if !file.IsDir() || strings.ToLower(file.Name()) != InstallDirMarker {
```
Naatan, line 128: This is probably too aggressive. I think we can get by just checking if it has state executables in the wrong dir.

internal/locale/locales/en-us.yaml
```diff
@@ -1835,6 +1835,13 @@ unstable_command_warning:
     This feature [ACTIONABLE]state {{.V0}}[/RESET] is still in beta. If you want to opt-in to unstable features, run the following command:
 
     [ACTIONABLE]state config set optin.unstable true[/RESET]
+err_update_corrupt_install:
```
Naatan, line 1838: I don't think we actually use this, since the PR moved from erroring out to actually addressing the error. Though maybe we should print this as a fallback when the automatic cleanup fails.

test/integration/update_int_test.go
```diff
@@ -143,6 +144,39 @@ func (suite *UpdateIntegrationTestSuite) testUpdate(ts *e2e.Session, baseDir str
...
+	ts.UseDistinctStateExes()
+
+	cfg, err := config.NewCustom(ts.Dirs.Config, singlethread.New(), true)
+	suite.Require().NoError(err)
+	defer cfg.Close()
+
+	subBinDir := filepath.Join(ts.Dirs.Bin, "bin")
+	ts.CopyExeToDir(ts.Exe, filepath.Join(subBinDir, filepath.Base(ts.Exe)))
+	ts.CopyExeToDir(ts.SvcExe, filepath.Join(subBinDir, filepath.Base(ts.SvcExe)))
+	ts.CopyExeToDir(ts.TrayExe, filepath.Join(subBinDir, filepath.Base(ts.TrayExe)))
```
Naatan, line 161: Let's copy recursively from the dir that `ts.Exe` is in, that might help surface future edge-cases.

test/integration/install_scripts_int_test.go
```diff
@@ -126,6 +127,33 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
...
+		)
+	} else {
+		cp = ts.SpawnCmdWithOpts("powershell.exe", e2e.WithArgs(argsWithActive...),
+			e2e.AppendEnv("SHELL="),
+			e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+		)
+	}
+	cp.ExpectLongString("Installation path must be an empty directory")
+	if runtime.GOOS != "windows" {
+		cp.ExpectExitCode(1)
```
Naatan, line 153: Should file a bug; Windows should still exit non-zero.

Naatan (changes requested):

cmd/state-installer/installer_lin_mac.go
MDrakos: The system (launcher) files are reinstalled with every update. So we rewrite the install files including any symlinks so this shouldn't be an issue on macOS.

test/integration/update_int_test.go
MDrakos: Could not use `fileutils.CopyFiles` as it lead to recursion issues. Instead, I'm just reading the files in the `bin` directory and copying those.

test/integration/install_scripts_int_test.go
MDrakos: Created here: https://activestatef.atlassian.net/browse/DX-826

cmd/state-installer/installer.go
```diff
@@ -160,6 +162,66 @@ func (i *Installer) sanitizeInput() error {
...
+		return errs.Wrap(errCorruptedInstall, "Bin path does not exist: %s", binPath)
+	}
+
+	// Detect if the install dir has files in it
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return errs.Wrap(err, "Could not read directory: %s", path)
+	}
+
+	for _, file := range files {
```
Naatan, line 198:
```suggestion
	// Executable files should be in bin dir, not root dir
	for _, file := range files {
```

cmd/state-installer/installer.go
```diff
@@ -160,6 +162,66 @@ func (i *Installer) sanitizeInput() error {
...
+		return nil
+	}
+
+	// Detect if bin dir exists
+	binPath, err := installation.BinPathFromInstallPath(path)
+	if err != nil {
+		return errs.Wrap(err, "Could not detect bin path")
+	}
+	if !fileutils.DirExists(binPath) {
+		return errs.Wrap(errCorruptedInstall, "Bin path does not exist: %s", binPath)
```
Naatan, line 189: Going by the rest of the logic I don't think this actually matters.

internal/installation/paths.go
```diff
@@ -66,20 +60,20 @@ func BinPathFromInstallPath(installPath string) (string, error) {
...
-	}
-
-	binPath, err := BinPathFromInstallPath(installPath)
-	if err != nil {
-		return false, "", errs.Wrap(err, "Could not detect binPath from BinPathFromInstallPath")
+func InstalledOnPath(installRoot string) (bool, string, error) {
+	var binPath string
+	var err error
+	if filepath.Base(installRoot) == "bin" {
+		binPath = installRoot
```
Naatan, line 67: I think this is where we can use the InstallDirMarker. Rather than make assumptions about file paths we can just scan up the tree until we find InstallDirMarker.

internal/updater/updater.go
```diff
@@ -126,6 +136,8 @@ func (u *AvailableUpdate) InstallBlocking(installTargetPath string, args ...stri
 	if u.Tag != nil {
 		envs = append(envs, fmt.Sprintf("%s=%s", constants.UpdateTagEnvVarName, *u.Tag))
 	}
+
+	logging.Debug("Executing: %s %s env: %s\n", installerPath, args, envs)
```
Naatan, line 140: We're already debug logging this: https://github.com/ActiveState/cli/blob/3bf100013b38a2752b32c98192dde0dc40953860/internal/exeutils/exeutils.go#L120

Naatan (changes requested):

internal/installation/paths.go
```diff
@@ -62,13 +61,31 @@ func BinPathFromInstallPath(installPath string) (string, error) {
...
-func InstalledOnPath(installPath string) (bool, string, error) {
-	binPath, err := BinPathFromInstallPath(installPath)
+func InstalledOnPath(installRoot string) (bool, string, error) {
+	if !fileutils.DirExists(installRoot) {
+		return false, "", nil
+	}
+
+	path := appinfo.StateApp(installRoot).Exec()
+	if fileutils.TargetExists(path) {
+		return true, filepath.Dir(path), nil
```
MDrakos, line 71: Right now I'm using the `InstallDirMarker` file as a fallback because older versions of the state tool will not have that file when updating.

cmd/state-installer/installer.go
```diff
@@ -185,16 +185,14 @@ func detectCorruptedInstallDir(path string) error {
 	if err != nil {
 		return errs.Wrap(err, "Could not detect bin path")
 	}
-	if !fileutils.DirExists(binPath) {
```
Naatan: You'll have to account for this on line 203 where you try to read the `binPath`.

internal/installation/paths.go
Naatan: Please move any sort of logic dealing with old installation formats to the updater. The point of this story is that we can get rid of all this type of fallback / legacy logic and assume a pristine installation structure.

This could very well mean moving the entire `InstalledOnPath()` function into the updater/installer because that's the only place it's used. Point here is that the installer is the ONLY part of the code vulnerable to legacy or corrupted install dir formats.

Naatan (changes requested):

cmd/state-installer/installer.go
MDrakos: Don't we want to fail if the `bin` path doesn't exist? With the check removed this will fail if the `binPath` does not exist

MDrakos: ReadDir doesn't seems to return any identifiable error so I've added this check to the error case

cmd/state-installer/installer.go
```diff
@@ -226,3 +229,30 @@ func isStateExecutable(name string) bool {
...
+	if !fileutils.DirExists(installRoot) {
+		return false, "", nil
+	}
+
+	path := appinfo.StateApp(installRoot).Exec()
+	if fileutils.TargetExists(path) {
+		return true, filepath.Dir(path), nil
+	}
+
+	found, err := fileutils.FindFileInPath(installRoot, installation.InstallDirMarker)
```
Naatan, line 243: I don't think this is where we should be using the InstallDirMarker. I would see it used in places like `appinfo.newAppInfo()` and `installation.InstallPath()`. This function is just checking whether the resulting install path has state tool on it.

Naatan (changes requested):

cmd/state-installer/installer.go
```diff
@@ -140,3 +161,98 @@ func (i *Installer) sanitize() error {
...
+	}
+
+	// Executable files should be in bin dir, not root dir
+	for _, file := range files {
+		if isStateExecutable(strings.ToLower(file.Name())) {
+			return errs.Wrap(errCorruptedInstall, "Install directory should only contain dirs: %s", path)
+		}
+	}
+
+	// Ensure that bin dir has at least the state and state-svc executables
```
Naatan, line 202: As discussed I think we can drop this as it's of no consequence.

Naatan (changes requested):

cmd/state-installer/installer.go
MDrakos: That's fair. I added it because of this comment: https://github.com/ActiveState/cli/pull/1784#discussion_r842183809 but the logic was different at the time.

Naatan (approved):

PR #1787, opened 2022-04-05
Query and directly fetch latest version to install.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-806" title="DX-806" target="_blank">StoryDX-806</a>  Our installers don't download the interim installer
  </td></table>
  <br />
 

Do not download the latest and then have it fetch the latest version to install...


cmd/state-installer/cmd.go
```diff
@@ -300,14 +312,6 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
...
-	if !params.force && stateToolInstalled {
-		logging.Debug("Cancelling out because State Tool is already installed")
-		out.Print(fmt.Sprintf("State Tool Package Manager is already installed at [NOTICE]%s[/RESET]. To reinstall use the [ACTIONABLE]--force[/RESET] flag.", installPath))
-		an.Event(AnalyticsFunnelCat, "already-installed")
-		return postInstallEvents(out, cfg, an, params, true)
-	}
-
 	// If no sourcePath was provided then we still need to download the source files, and defer the actual
 	// installation to the installer contained within the source file
 	return installFromRemoteSource(out, cfg, an, args, params)
```
Naatan, line 317: We should get rid of this flow, since it's now replaced by the install.sh and install.ps1 and so there's no use-case for this anymore. Keeping it around probably makes our code more complex and error-prone than it needs to be.

installers/install.ps1
```diff
@@ -6,12 +6,14 @@ install.ps1 -b branchToInstall
...
 
+# URL to fetch update infos from.
+$script:BASEINFOURL = "https://platform.activestate.com/sv/state-update/api/v1/info"
 # URL to fetch installer archive from
 $script:BASEFILEURL = "https://state-tool.s3.amazonaws.com/update/state"
 # The name of the remove archive to download
 $script:ARCHIVENAME = "state-installer.zip"
 # Name of the installer executable to ultimately use
-$script:INSTALLERNAME = "state-installer.exe"
+$script:INSTALLERNAME = "state-install\\state-installer.exe"
```
Naatan, line 16: I'm trying to understand why this changed but I can't figure it out, can you explain? Is the install archive for "latest" structured differently than that of a specific version?

installers/install.ps1
```diff
@@ -126,6 +135,17 @@ catch [System.Exception]
     return 1
 }
 
+# Verify checksum.
+$hash = (Get-FileHash -Path $zipPath -Algorithm SHA256).Hash
+  if ($hash -ne $checksum){
```
Naatan, line 140:
```suggestion
if ($hash -ne $checksum){
```

Naatan (changes requested):

installers/install.ps1
mitchell-as: The archive pointed to by platform.activestate.com has a slightly different directory structure than the archive we were hardcoded to look for.

cmd/state-installer/cmd.go
mitchell-as: Okay.

installers/install.ps1
mitchell-as: Thanks for catching this.

Naatan (changes requested):
This comment doesn't appear to be addressed: https://github.com/ActiveState/cli/pull/1787#discussion_r849852153

cmd/state-installer/cmd.go
mitchell-as: Oh sorry, GitHub confused me with both its e-mail and review interface. I thought something completely different :(

cmd/state-installer/cmd.go
```diff
@@ -300,23 +305,23 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
...
 	}
 
-	// Check if state tool already installed
-	if !params.force && stateToolInstalled {
-		logging.Debug("Cancelling out because State Tool is already installed")
-		out.Print(fmt.Sprintf("State Tool Package Manager is already installed at [NOTICE]%s[/RESET]. To reinstall use the [ACTIONABLE]--force[/RESET] flag.", installPath))
-		an.Event(AnalyticsFunnelCat, "already-installed")
-		return postInstallEvents(out, cfg, an, params, true)
-	}
+	storeInstallSource(params.sourceInstaller)
```
Naatan, line 308: This doesn't seem to get called under successful conditions.

cmd/state-installer/cmd.go
```diff
@@ -300,23 +305,23 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
...
 
-	// If no sourcePath was provided then we still need to download the source files, and defer the actual
-	// installation to the installer contained within the source file
-	return installFromRemoteSource(out, cfg, an, args, params)
+	return locale.NewError("err_install_source_path_not_provided", "sourcePath was not provided by install.sh or install.ps1")
```
Naatan, line 310:
```suggestion
	return locale.NewError("err_install_source_path_not_provided", "Installer was called without an installation payload. Please make sure you're using the install.sh or install.ps1 scripts.")
```

Localized errors are user facing.

cmd/state-installer/cmd.go
```diff
@@ -272,10 +265,14 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 	// Detect whether this is a fresh install or an update
 	isUpdate := false
 	switch {
+	case (params.sourceInstaller == "install.sh" || params.sourceInstaller == "install.ps1") && fileutils.FileExists(packagedStateExe):
+		logging.Debug("Not using update flow as installing via " + params.sourceInstaller)
+		params.sourcePath = installerPath
```
Naatan, line 270: I like this change, as it makes the behavior more explicit and less error prone.

That said, this has surfaced another bug in our code. When I run:

```
./installers/install.sh -b dx-806
```

I get:

```
Installation path must be an empty directory
```

It seems we might be targeting the wrong directory because I've never installed this branch before. Previously this probably didn't trigger any errors because it considered it an "update".

I just checked, and it does indeed seem we might be targeting the wrong install dir:

```
❯ which state
/Users/nathanrijksen/.local/ActiveState/StateTool/bin/state
```

That should have been:

```
/Users/nathanrijksen/.local/ActiveState/StateTool/beta/bin/state
```

Note the first variant is missing the branch.

Naatan: I realize this is becoming a bit of a scope creep for your story. If you prefer we can file a separate ticket for this issue.

Naatan (changes requested):

cmd/state-installer/cmd.go
mitchell-as: Strange, I'm not seeing that. `-b dx-806` works just fine for me:

```
mitchellb@ManBearPig cli % ls ~/.local/ActiveState/StateTool
release
mitchellb@ManBearPig cli % which state
/Users/mitchellb/.local/ActiveState/StateTool/release/bin/state
mitchellb@ManBearPig cli % state --version
ActiveState CLI by ActiveState Software Inc.
License BSD 3
Version 0.33.0-SHA7b42690
Revision 7b4269018813827216b40f975aca606e02e1b4de
Branch release
Built Tue Mar 22 2022 20:18:03 +0000 UTC

mitchellb@ManBearPig cli % ./installers/install.sh -b dx-806
• Preparing Installer for State Tool Package Manager version 0.34.0-SHA6f6f995... ✔ Done

█ Installing State Tool Package Manager

The State Tool lets you install and manage your language runtimes.

ActiveState collects usage statistics and diagnostic data about failures. 
By using the State Tool Package Manager you agree to the terms of ActiveState’s Privacy Policy, 
available at: https://www.activestate.com/company/privacy-policy

• Installing State Tool to /Users/mitchellb/.local/ActiveState/StateTool/dx-806... ✔ Done

█ State Tool Package Manager Installation Complete

State Tool Package Manager has been successfully installed.
mitchellb@ManBearPig cli % which state
/Users/mitchellb/.local/ActiveState/StateTool/dx-806/bin/state
mitchellb@ManBearPig cli % state --version
ActiveState CLI by ActiveState Software Inc.
License BSD 3
Version 0.34.0-SHA6f6f995
Revision 6f6f99565390592ab07fb8abb9ea11a470b90521
Branch dx-806
Built Wed Apr 13 2022 21:43:28 +0000 UTC
```

cmd/state-installer/cmd.go
mitchell-as: You're correct. I misunderstood the meaning/purpose of this function.

cmd/state-installer/cmd.go
mitchell-as: I was under the impression that this could only happen if we (the developers) miscoded something, so this error message was primarily for us. I can see that it's possible a user might attempt to invoke `state-installer` directly, and would thus run into this. Thanks for pointing it out.

cmd/state-installer/cmd.go
Naatan: I think I had a corrupted install somehow. I reinstalled fresh and then tried again, this time I did not get a repro.

Naatan (approved):

PR #1790, opened 2022-04-06
Fix inconsistent install path

MDrakos:
Also fixes potential error when killing child processes

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -249,6 +249,9 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 	if err != nil {
 		return errs.Wrap(err, "Could not detect if State Tool is already installed.")
 	}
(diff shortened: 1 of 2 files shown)


cmd/state-installer/cmd.go
```diff
@@ -249,6 +249,9 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 	if err != nil {
 		return errs.Wrap(err, "Could not detect if State Tool is already installed.")
 	}
+	if installPath != params.path {
+		params.path = installPath
+	}
```
Naatan, line 254: Please add the logic for adding the branch earlier on, and modify `installPath`. By modifying `params.path` you're creating the potential for new logic paths that we aren't interested in for this fix.

Naatan (changes requested):

Naatan (approved):

PR #1792, opened 2022-04-06
Retire legacy installation code that's causing grief

Naatan:


internal/installation/paths.go
```diff
@@ -28,7 +25,20 @@ func DefaultInstallPath() (string, error) {
...
-func InstallPath() (string, error) {
+func InstallRoot(path string) (string, error) {
+	currentPath := path
+	for filepath.Dir(currentPath) != currentPath {
+		candidate := filepath.Join(currentPath, InstallDirMarker)
+		if fileutils.TargetExists(candidate) {
+			return filepath.Dir(candidate), nil
+		}
+		currentPath = filepath.Dir(currentPath)
+	}
```
MDrakos, line 36: Can we use [this](https://github.com/ActiveState/cli/blob/master/internal/fileutils/fileutils.go#L389) instead?

internal/updater/updater.go
```diff
@@ -88,7 +88,7 @@ func (u *AvailableUpdate) prepareInstall(installTargetPath string, args []string
 	}
 
 	if installTargetPath == "" {
-		installTargetPath, err = installation.InstallPath()
+		installTargetPath, err = installation.InstallPathFromArg0()
```
MDrakos, line 91: How will this get the correct install path when a user is just running `state update`? I do not think `installTargetPath` is normally set at this point. It is in the bugged version but other versions I've tested it is not.

internal/runners/clean/run_lin_mac.go
```diff
@@ -124,11 +119,11 @@ func removeInstall(cfg configurable) error {
 		return errs.Wrap(aggErr, "Could not determine OS specific launcher install path")
 	}
 
-	if err := installation.RemoveSystemFiles(appPath); err != nil {
+	if err := installmgr.RemoveSystemFiles(appPath); err != nil {
 		aggErr = errs.Wrap(aggErr, "Failed to remove system files at %s: %v", appPath, err)
 	}
 
-	installPath, err := installation.InstallPath()
+	installPath, err := installation.InstallPathFromArg0()
```
MDrakos, line 126: Similar to my other comment, if a user runs `state clean uninstall` how will this get the install path when `os.Args[0]` is an empty string?

MDrakos (changes requested):

internal/installation/paths.go
Naatan: I'd forgotten about that, good thinking!

internal/updater/updater.go
Naatan: It would get it from the `state` executable that it's calling from: http://github.com/ActiveState/cli/blob/remove-legacy-installer/internal/installation/paths.go#L46-L46

MDrakos (approved):

PR #1794, opened 2022-04-06
Ensure service shuts down properly

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-840" title="DX-840" target="_blank">BugDX-840</a>  Some integration tests fail to stop/start the service
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/service.go
+++ b/cmd/state-svc/service.go
@@ -75,3 +75,7 @@ func (s *service) Stop() error {
 
 	return nil
 }
+
+func (s *service) Wait() {
+	_ = s.ipcSrv.Wait()
+}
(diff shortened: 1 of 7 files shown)


cmd/state-svc/service.go
```diff
@@ -75,3 +75,7 @@ func (s *service) Stop() error {
 
 	return nil
 }
+
+func (s *service) Wait() {
+	_ = s.ipcSrv.Wait()
```
Naatan, line 80: No chance we'd want to log this?

internal/ipc/server.go
```diff
@@ -88,71 +88,93 @@ func (ipc *Server) Start() error {
...
-			return errs.Wrap(err, "Unexpected accept error")
+		}()
+
+		wg.Wait()
+	}()
+
+	var retErr error
+	for err := range errsc {
+		if err != nil && retErr == nil {
+			retErr = err
```
Naatan, line 139: No chance there'd be multiple errors?

Naatan (approved):
Looks tentatively ok, depending on your answer to my comments.

PR #1796, opened 2022-04-07
Deprecation check in state service

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-803" title="DX-803" target="_blank">StoryDX-803</a>  Our deprecation mechanic uses the state-svc
  </td></table>
  <br />


cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -94,17 +97,9 @@ func (checker *Checker) check(versionNumber string) (*Info, error) {
...
-		infos, err = checker.fetchDeprecationInfo()
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		infos, err = checker.cachedDeprecationInfo()
-		if err != nil {
-			return nil, errs.Wrap(err, "cachedDeprecationInfo failed")
-		}
+	infos, err := checker.cachedDeprecationInfo()
```
Naatan, line 100: I would just use the cache lib: http://github.com/ActiveState/cli/blob/80228c0a9d9c0c76daabe6d756eb027b1428b910/cmd/state-svc/internal/resolver/resolver.go#L40-L40

Much of the logic here can then probably be dumped, as it should just be a simple matter of "if cache is empty; refresh".

testing/main.go
```diff
@@ -0,0 +1,27 @@
+package main
```
Naatan, line 1: Debugging code?

Naatan (changes requested):

testing/main.go
MDrakos: Whoops

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,187 @@
...
+// This uses a fairly short timeout to Check against our deprecation url, so this should not be considered conclusive.
+func (checker *Checker) Check() (*graph.DeprecationInfo, error) {
+	if !constvers.NumberIsProduction(constants.Version) {
+		return nil, nil
+	}
+
+	versionInfo, err := version.NewVersion(constants.Version)
+	if err != nil {
+		return nil, errs.Wrap(err, "Invalid version number: %s", constants.Version)
+	}
```
Naatan, line 77: It doesn't seem like we need to run this logic if the cache hit below succeeds.

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,187 @@
...
+	if !exists {
+		go func() {
+			err := checker.updateDeprecationInfo()
+			if err != nil {
+				multilog.Critical("Could not update deprecation information %s", errs.JoinMessage(err))
+			}
+		}()
+
+		return nil, nil
+	}
```
Naatan, line 89: This is basically ensuring that some calls won't get deprecation info simply because we haven't warmed the cache. I think we should warm the cache in the background, on a timer. If there are no network issues clients should get deprecation info 100% of the time.

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/multilog"
+	"github.com/hashicorp/go-version"
+	"github.com/patrickmn/go-cache"
+)
+
+const (
+	// DefaultTimeout defines how long we should wait for a response from constants.DeprecationInfoURL
+	DefaultTimeout = time.Second
```
Naatan, line 25: This timeout should be WAY higher now that we don't care about how long it takes. In fact I think we should be using the retry http client.

Naatan (changes requested):

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,200 @@
...
+				Reason:      info.Reason,
+			}, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func (checker *Checker) pollDeprecationInfo() {
+	timer := time.NewTicker(1 * time.Minute)
```
Naatan, line 105: Once an hour is fine. Don't want someone to get offended cause we're polling from their machine every minute.

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,200 @@
...
+
+	return checker
+}
+
+// Check will Check if the current version of the tool is deprecated and returns deprecation info if it is.
+// This uses a fairly short timeout to Check against our deprecation url, so this should not be considered conclusive.
+func (checker *Checker) Check() (*graph.DeprecationInfo, error) {
+	data, exists := checker.cache.Get(cacheKey)
+	if !exists {
+		return nil, nil
```
Naatan, line 73: Since in THEORY this should never happen we should error log this.

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	if !constvers.NumberIsProduction(constants.Version) {
+		return nil, nil
```
Naatan, line 77: Don't think we need this anymore.

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,200 @@
...
+	for _, info := range infos {
+		if versionInfo.LessThan(info.versionInfo) || versionInfo.Equal(info.versionInfo) {
+			return &graph.DeprecationInfo{
+				Version:     info.Version,
+				Date:        info.Date.Format(constants.DateFormatUser),
+				DateReached: info.DateReached,
+				Reason:      info.Reason,
+			}, nil
+		}
+	}
```
Naatan, line 99: I think you can cache the result of all of this. Don't need to assert it based on the cached info.

cmd/state-svc/internal/deprecation/deprecation.go
```diff
@@ -0,0 +1,200 @@
...
+func (checker *Checker) pollDeprecationInfo() {
+	timer := time.NewTicker(1 * time.Minute)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-timer.C:
+			err := checker.Refresh()
+			if err != nil {
+				multilog.Critical("Could not update deprecation information %s", errs.JoinMessage(err))
```
Naatan, line 113:
```suggestion
				multilog.Error("Could not update deprecation information %s", errs.JoinMessage(err))
```

Critical is for user facing non-input errors.

cmd/state-svc/internal/server/server.go
```diff
@@ -32,13 +33,13 @@ type Server struct {
 	analytics   *sync.Client
 }
 
-func New(cfg *config.Instance, an *sync.Client, shutdown context.CancelFunc) (*Server, error) {
+func New(cfg *config.Instance, an *sync.Client, checker *deprecation.Checker, shutdown context.CancelFunc) (*Server, error) {
```
Naatan, line 36: I don't think we need to pass checker into the server. The checker is pretty much confined to the resolver package, so constructing it can happen there.

Naatan (changes requested):

cmd/state-svc/internal/deprecation/deprecation.go
MDrakos: When the change that is suggested [here](https://github.com/ActiveState/cli/pull/1796#discussion_r854577055) this will happen often so I've just made it a debug log.

cmd/state-svc/internal/deprecation/deprecation.go
MDrakos: `Critical` doesn't really communicate "user facing non-input errors" I would think that not being able to refresh the deprecation info would be fairly critical.

Naatan: Critical tells us on rollbar that the end-user was directly affected by the error. It's not meant to convey "this error is more serious than that error", it's just meant to convey "this error directly affected the user".

cmd/state-svc/main.go
```diff
@@ -167,20 +159,25 @@ func run(cfg *config.Instance) (rerr error) {
...
-	ctx, shutdown := context.WithCancel(context.Background())
-	p := NewService(cfg, an, deprecation, shutdown)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	p := NewService(ctx, cfg, an, auth)
+
+	if err := p.Start(); err != nil {
+		return errs.Wrap(err, "Could not start service")
+	}
```
Naatan, line 180: Trying to understand why this is being called now but wasn't before?

Naatan (changes requested):

cmd/state-svc/internal/deprecation/deprecation.go
MDrakos: What does "directly affect the user" mean here? The vast majority of our errors bubble up to the user but those aren't considered critical. By using critical for user-facing non-input errors we are saying that they are more serious than other errors. I'm being nit-picky here but I still think using critical logging for critical functionality makes sense. If deprecation is breaking it is something that should be conveyed to us in a more serious way. My misunderstanding could just be from my lack of familiarity with logging levels and what they are traditionally used for.

cmd/state-svc/main.go
MDrakos: Not sure how this showed up in the diff. It's already on master [here](https://github.com/ActiveState/cli/blob/master/cmd/state-svc/main.go#L175). The only thing I changed on the state-svc main file is to handle an error that we were ignoring.

Naatan (approved):

Naatan (approved):

PR #1800, opened 2022-04-08
Windows install script should indicate installer failure.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-826" title="DX-826" target="_blank">BugDX-826</a>  State Tool not exiting with code 1 on error in integration tests
  </td></table>
  <br />

### DIFF
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -147,10 +147,10 @@ $PSDefaultParameterValues['*:Encoding'] = 'utf8'
 
 $env:ACTIVESTATE_SESSION_TOKEN = $script:SESSION_TOKEN_VALUE
 & $exePath $args --source-installer="install.ps1"
-if ( !$? ) {
-  exit 1
-}
 if (Test-Path env:ACTIVESTATE_SESSION_TOKEN)
 {
     Remove-Item Env:\ACTIVESTATE_SESSION_TOKEN
+}
+if ( !$? ) {
+  exit 1
 }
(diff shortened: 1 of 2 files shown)


installers/install.ps1
```diff
@@ -147,7 +147,11 @@ $PSDefaultParameterValues['*:Encoding'] = 'utf8'
 
 $env:ACTIVESTATE_SESSION_TOKEN = $script:SESSION_TOKEN_VALUE
 & $exePath $args --source-installer="install.ps1"
+$success = $?
```
MDrakos, line 150: Cool. I didn't realize this would work in PowerShell

installers/install.ps1
```diff
@@ -147,7 +147,11 @@ $PSDefaultParameterValues['*:Encoding'] = 'utf8'
...
 if (Test-Path env:ACTIVESTATE_SESSION_TOKEN)
 {
     Remove-Item Env:\ACTIVESTATE_SESSION_TOKEN
-}
\ No newline at end of file
+}
+if ( !$success ) {
+  exit 1
```
MDrakos, line 156: Is there a way for us to exit with the error code from the state tool? This is currently in line with what we were doing before. Just curious.

MDrakos (changes requested):

installers/install.ps1
mitchell-as: Nope, doesn't look like it :( https://github.com/ActiveState/cli/runs/5891595613?check_suite_focus=true

MDrakos (approved):

PR #1803, opened 2022-04-11
Add shortcut to uninstall State Tool on Windows.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-827" title="DX-827" target="_blank">StoryDX-827</a>  During installation of state tool, a start menu item for Uninstall is created
  </td></table>
  <br />

### DIFF
--- a/internal/osutils/shortcut/shortcut_windows.go
+++ b/internal/osutils/shortcut/shortcut_windows.go
@@ -56,10 +56,6 @@ func (s *Shortcut) Enable() error {
 	if err != nil {
 		return errs.Wrap(err, "Could not set Shortcut target")
 	}
-	err = s.setArguments(s.args)
-	if err != nil {
-		return errs.Wrap(err, "Could not set Shortcut arguments")
-	}
 
 	return nil
 }
--- a/internal/osutils/shortcut/shortcut_windows.go
+++ b/internal/osutils/shortcut/shortcut_windows.go
@@ -17,12 +17,13 @@ type Shortcut struct {
 	dir      string
 	name     string
 	target   string
+	args     string
 	dispatch *ole.IDispatch
 }
 
-func New(dir, name, target string) *Shortcut {
+func New(dir, name, target, args string) *Shortcut {
 	return &Shortcut{
-		dir, name, target, nil,
+		dir, name, target, args, nil,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


internal/osutils/shortcut/shortcut_windows.go
```diff
@@ -17,12 +17,13 @@ type Shortcut struct {
 	dir      string
 	name     string
 	target   string
+	args     string
 	dispatch *ole.IDispatch
 }
 
-func New(dir, name, target string) *Shortcut {
+func New(dir, name, target, args string) *Shortcut {
```
MDrakos, line 24: functions with an `args` argument are typically variadic. That will make the last argument optional and you won't have to update existing calls of this function.

internal/runners/prepare/prepare_windows.go
```diff
@@ -50,6 +50,12 @@ func (r *Prepare) prepareStartShortcut() error {
 		return locale.WrapError(err, "err_preparestart_icon", "Could not set icon for shortcut file")
 	}
 
+	sc = shortcut.New(shortcutDir, "Uninstall State Tool", filepath.Join(os.Getenv("windir"), "system32", "cmd.exe"), "/C \"state clean uninstall\"")
+	err = sc.Enable()
+	if err != nil {
+		return locale.WrapError(err, "err_prepare_shortcut", "Could not create shortcut")
```
MDrakos, line 56: Nit: This is the exact same error message as above on line 41. Let's move the actual text to our localization file.

internal/runners/prepare/prepare_windows.go
```diff
@@ -50,6 +50,12 @@ func (r *Prepare) prepareStartShortcut() error {
 		return locale.WrapError(err, "err_preparestart_icon", "Could not set icon for shortcut file")
 	}
 
+	sc = shortcut.New(shortcutDir, "Uninstall State Tool", filepath.Join(os.Getenv("windir"), "system32", "cmd.exe"), "/C \"state clean uninstall\"")
```
MDrakos, line 53: The subshell package has a function for getting the shell binary. See [here](https://github.com/ActiveState/cli/blob/master/internal/subshell/subshell.go#L143)

MDrakos (changes requested):

internal/osutils/shortcut/shortcut_windows.go
mitchell-as: I thought about this, but I didn't want args to be more than one in length (so truly an optional argument, not variadic). Nevertheless, I've implemented your suggestion.

internal/runners/prepare/prepare_windows.go
mitchell-as: Thanks for the tip. Very useful.

internal/runners/prepare/prepare_windows.go
```diff
@@ -50,6 +50,12 @@ func (r *Prepare) prepareStartShortcut() error {
...
+	err = sc.Enable()
```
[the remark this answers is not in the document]
mitchell-as: Since this is a shortcut to cmd.exe, it uses cmd.exe's icon. Attempting to set it results in a permission error.

MDrakos (approved):

PR #1806, opened 2022-04-13
Fix svc system sig handling

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-861" title="DX-861" target="_blank">BugDX-861</a>  Service does not shutdown by system signal properly
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/service.go
+++ b/cmd/state-svc/service.go
@@ -17,15 +17,14 @@ import (
 
 type service struct {
 	ctx    context.Context
-	cancel context.CancelFunc
 	cfg    *config.Instance
 	an     *anaSvc.Client
 	server *server.Server
 	ipcSrv *ipc.Server
 }
 
-func NewService(ctx context.Context, cancel context.CancelFunc, cfg *config.Instance, an *anaSvc.Client) *service {
-	return &service{ctx: ctx, cancel: cancel, cfg: cfg, an: an}
+func NewService(ctx context.Context, cfg *config.Instance, an *anaSvc.Client) *service {
+	return &service{ctx: ctx, cfg: cfg, an: an}
 }
 
 func (s *service) Start() error {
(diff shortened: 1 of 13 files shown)


daved:
Resolving this was significantly frustrated by the signal handling found in the captain package. Bug filed: https://activestatef.atlassian.net/browse/DX-886

cmd/state-svc/service.go
```diff
@@ -69,13 +68,16 @@ func (s *service) Stop() error {
...
 	}
 
 	return nil
 }
 
-func (s *service) Wait() {
-	_ = s.ipcSrv.Wait()
+func (s *service) Wait() error {
+	if err := s.ipcSrv.Wait(); err != nil {
+		return errs.Wrap(err, "IPC server operating failure")
```
Naatan, line 80:
```suggestion
		return errs.Wrap(err, "Failed to wait for ipc service")
```

Error was a bit generic.

internal/ipc/comm.go
```diff
@@ -40,12 +39,10 @@ func getPing(ctx context.Context, c *Client) (string, error) {
...
 }
 
-func stopHandler(c io.Closer) RequestHandler {
+func stopHandler(stop func() error) RequestHandler {
 	return func(key string) (string, bool) {
 		if key == keyStop {
-			defer func() {
-				go c.Close()
-			}()
+			_ = stop()
```
Naatan, line 45: If the error is of no consequence why is it returning an error?

internal/ipc/server.go
```diff
@@ -133,35 +144,37 @@ func (ipc *Server) Start() error {
...
-	return nil
 }
 
 func (ipc *Server) Wait() error {
-	ipc.wg.Wait()
-	return nil
+	var retErr error
+	for err := range ipc.errsc {
+		if err != nil && retErr == nil {
+			retErr = err
```
Naatan, line 165: We could at least wrap the error messages. Might loose the error type, but you're already at risk of that anyway.

Naatan (changes requested):

cmd/state-svc/service.go
daved: The Start function returns errors only related to starting, Stop only related to stopping, and Wait returns errors related to normal operation. So it's not failing to wait, it's relaying a malfunction after the wait unblocks. I'm happy to include the suggestion if you still think the change is useful.

internal/ipc/server.go
daved: This is just a funneling of returned errors, so the wrapping was already performed.

Naatan (dismissed):

daved:
The uninstall failures are reported here: https://activestatef.atlassian.net/browse/DX-895

Naatan (approved):

PR #1810, opened 2022-04-19
Added environment variable to disable analytics.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-872" title="DX-872" target="_blank">StoryDX-872</a>  Expose an ENV Var to disable analytics
  </td></table>
  <br />
 

In conjunction with https://github.com/ActiveState/documentation/pull/56

### DIFF
--- a/internal/analytics/client/async/client.go
+++ b/internal/analytics/client/async/client.go
@@ -34,7 +34,6 @@ type Client struct {
 	sessionToken     string
 	updateTag        string
 	closed           bool
-	cfg              *config.Instance
 }
 
 var _ analytics.Dispatcher = &Client{}
@@ -51,7 +50,6 @@ func New(svcModel *model.SvcModel, cfg *config.Instance, auth *authentication.Au
 	a.output = o
 	a.projectNameSpace = projectNameSpace
 	a.auth = auth
-	a.cfg = cfg
 
 	if condition.InUnitTest() {
 		return a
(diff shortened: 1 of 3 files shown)


MDrakos (changes requested):
Disabling analytics here introduces two places in our code where they can be disabled. The service uses the sync client so that means that this environment variable will not have an effect on the events the service sends. I think we may need to come up with a way that the presence or absence of the env var will enable/disable the analytics in the sync client.

mitchell-as:
Do the new changes address the issue you're bringing up?

internal/analytics/client/async/client.go
```diff
@@ -91,9 +94,14 @@ func (a *Client) Wait() {
...
 func (a *Client) sendEvent(category, action, label string, dims ...*dimensions.Values) error {
-	if a.closed {
+	switch {
+	case a.closed:
 		logging.Debug("Client is closed, not sending event")
 		return nil
+	case (a.cfg.IsSet(constants.ReportAnalayticsConfig) && !a.cfg.GetBool(constants.ReportAnalayticsConfig)) ||
+		strings.ToLower(os.Getenv(constants.DisableAnalyticsEnvVarName)) == "true":
+		logging.Debug("Analytics are disabled; not sending event")
+		return nil
```
Naatan, line 104: This is introducing new behavior, and is causing the config to be evaluated on every analytical event which slows down execution. I don't think we need this either because the async client reports into the sync client, which already has this logic.

Naatan (changes requested):

internal/analytics/client/async/client.go
mitchell-as: Thanks for letting me know that async reports to sync. That's good to know.

Naatan (approved):

Naatan (approved):

PR #1811, opened 2022-04-19
Removed logging of error wrapping and creation.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-878" title="DX-878" target="_blank">StoryDX-878</a>  Stop reporting "error created" and "wrapped error" logs
  </td></table>
  <br />
 

It's cluttering up the log file. These errors will ultimately be reported anyway.

### DIFF
--- a/cmd/state/errors.go
+++ b/cmd/state/errors.go
@@ -101,13 +101,6 @@ func unwrapError(err error) (int, error) {
 		return code, nil
 	}
 
-	// Log error if this isn't a user input error
-	if !locale.IsInputError(err) {
-		multilog.Critical("Returning error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
-	} else {
-		logging.Debug("Returning input error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
-	}
-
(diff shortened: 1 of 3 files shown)


cmd/state/errors.go
```diff
@@ -101,13 +101,6 @@ func unwrapError(err error) (int, error) {
 		return code, nil
 	}
 
-	// Log error if this isn't a user input error
-	if !locale.IsInputError(err) {
-		multilog.Critical("Returning error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
-	} else {
-		logging.Debug("Returning input error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
-	}
```
Naatan, line 109: These are actual errors being logged. It should not be removed.

Naatan (changes requested):

cmd/state/errors.go
mitchell-as: Got it. I was honestly unsure about those lines. I thought perhaps they would be ultimately logged. Thanks for pointing out my error though.

Naatan (approved):

PR #1813, opened 2022-04-19
Fix windows PATH on uninstall

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-882" title="DX-882" target="_blank">BugDX-882</a>  Uninstall on Windows corrupts PATH
  </td></table>
  <br />


MDrakos:
The failing integration tests need #1814 to be pass

internal/runners/clean/run_win.go
```diff
@@ -117,7 +119,8 @@ func removePaths(logFile string, paths ...string) error {
 
 	args := []string{"/C", sf.Filename(), logFile, fmt.Sprintf("%d", os.Getpid()), filepath.Base(exe)}
 	args = append(args, paths...)
-	_, err = exeutils.ExecuteAndForget("cmd.exe", args)
+
+	_, err = exeutils.ExecuteAndForget(subshell.DetectShellBinary(cfg), args)
```
Naatan, line 123: Should revert this. The flags are specifically for cmd.exe. And the issue of not finding cmd.exe on path was a symptom, not the problem we're fixing.

Naatan (changes requested):

Naatan (dismissed):

Naatan (approved):

PR #1814, opened 2022-04-19
Ensure uninstall runner is removing bin directory

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-862" title="DX-862" target="_blank">BugDX-862</a>  CLI _ UNINSTALL: Uninstall failed for Version 0.34.0-SHA570ccb1 on Windows 11 
  </td></table>
  <br />

### DIFF
--- a/internal/runners/clean/run.go
+++ b/internal/runners/clean/run.go
@@ -2,14 +2,9 @@ package clean
 
 import (
 	"os"
-	"path/filepath"
 
-	"github.com/ActiveState/cli/internal/appinfo"
-	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
-	"github.com/ActiveState/cli/internal/exeutils"
 	"github.com/ActiveState/cli/internal/fileutils"
-	"github.com/ActiveState/cli/internal/installation"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/runners/prepare"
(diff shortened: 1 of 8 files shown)


internal/runners/clean/run_lin_mac.go
```diff
@@ -123,15 +136,11 @@ func removeInstall(cfg configurable) error {
 		aggErr = errs.Wrap(aggErr, "Failed to remove system files at %s: %v", appPath, err)
 	}
 
-	installPath, err := installation.InstallPathFromExecPath()
-	if err != nil {
-		aggErr = errs.Wrap(aggErr, "Could not get installation path")
-	}
-
-	if fileutils.DirExists(installPath) {
```
Naatan, line 131: Isn't your change leaving things like the .App around if you're only deleting the bin dir?

Naatan (changes requested):

internal/runners/clean/run_lin_mac.go
MDrakos: This is restoring the uninstall functionality to what it was before. The branch folder will still exist with the `.state_install_root` file and the `state-installer` executable but the `bin` directory will be removed.

Naatan: Ok, that doesn't seem like a desirable outcome. We want the entire folder to be uninstalled, unless I am missing something?

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/1814#discussion_r854561819

internal/runners/clean/run_lin_mac.go
```diff
@@ -34,7 +34,9 @@ func (u *Uninstall) runUninstall() error {
 	}
 
 	err = removeInstall(u.cfg)
-	if err != nil {
+	if errors.Is(err, errDirNotEmpty) {
+		u.out.Notice(locale.Tl("uninstall_warn_not_empty", "[WARNING]Unable to remove all files in the directory {{.V0}}[/RESET]"))
```
Naatan, line 38: Can't we give a reason here? It sounds like this would happen if the install dir had files that we don't recognize.

internal/runners/clean/run.go
```diff
@@ -58,3 +62,50 @@ func removeEnvPaths(cfg configurable) error {
...
+		var err error
+		if fileutils.IsDir(f) && fileutils.DirExists(f) {
+			err = os.RemoveAll(f)
+		} else if fileutils.FileExists(f) {
+			err = os.Remove(f)
+		}
+		if err != nil {
+			return errs.Wrap(err, "Could not clean install directory")
+		}
+	}
```
Naatan, line 108: I'd add the bin dir executables here as well, keep it all nice and central.

Also I think Windows and Linux might have some additional files? Can you make sure this is covered by an integration test which is ran for this PR?

Naatan (changes requested):

internal/runners/clean/run.go
MDrakos: Linux has no additional files. Windows has an icon file which I added to the integration test but this function does not run on Windows.

internal/runners/clean/run_lin_mac.go
```diff
@@ -146,3 +136,58 @@ func removeInstall(cfg configurable) error {
...
+		// application files. It is safe for us to remove this directory
+		// without first inspecting the contents.
+		"system",
+	}
+
+	for _, file := range asFiles {
+		f := filepath.Join(dir, file)
+
+		var err error
+		if fileutils.IsDir(f) && fileutils.DirExists(f) {
```
Naatan, line 182: Nit: It wouldn't be able to assert `IsDir()` unless it exists.

Naatan (approved):

internal/runners/clean/run_lin_mac.go
```diff
@@ -146,3 +136,58 @@ func removeInstall(cfg configurable) error {
...
+		// without first inspecting the contents.
+		"system",
+	}
+
+	for _, file := range asFiles {
+		f := filepath.Join(dir, file)
+
+		var err error
+		if fileutils.DirExists(f) && fileutils.IsDir(f) {
+			err = os.RemoveAll(f)
```
Naatan, line 183: Hmm we should probably do the empty check here as well.

Naatan (changes requested):

internal/runners/clean/run_lin_mac.go
MDrakos: We're relying on the order of the slice above rather than using the empty check. The `system` directory is not empty. For that, we would have to walk the directory and delete the application files or have a representation of the directory structure somewhere to refer to.

Naatan (approved):

PR #1815, opened 2022-04-20
Fix configurability of "report.analytics" and "report.errors".

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-881" title="DX-881" target="_blank">BugDX-881</a>  `state config set` can't set report.*
  </td></table>
  <br />
 

Also move configurability of logging to rollbar since its functionality moved there.

The failing integration test may be fixed in a separate PR by @Naatan .


Naatan (changes requested):
This PR is doing too much. The only change I'd expect to see is the `configMediator.RegisterOption` additions. The behavior of these config vars is already implemented: 

https://github.com/ActiveState/cli/pull/1709
https://github.com/ActiveState/cli/pull/1707

mitchell-as:
`report.errors` was broken when rollbar logging was moved into its own package. This PR fixes that, in addition to failing integration tests. If you prefer I break those into separate tickets, let me know.

cmd/state-update-dialog/main.go
```diff
@@ -46,8 +46,8 @@ func main() {
 		exitCode = 1
 		return
 	}
-	logging.CurrentHandler().SetConfig(cfg)
 	rollbar.SetupRollbar(constants.StateTrayRollbarToken) // We're using the state tray project cause it's closely related
+	rollbar.SetConfig(cfg)
```
Naatan, line 50: Can we amend `SetupRollbar` instead?

internal/analytics/client/async/client.go
```diff
@@ -91,9 +93,13 @@ func (a *Client) Wait() {
...
 
 func (a *Client) sendEvent(category, action, label string, dims ...*dimensions.Values) error {
-	if a.closed {
+	switch {
+	case a.closed:
 		logging.Debug("Client is closed, not sending event")
 		return nil
+	case a.cfg.IsSet(constants.ReportAnalyticsConfig) && !a.cfg.GetBool(constants.ReportAnalyticsConfig):
+		logging.Debug("Analytics are disabled; not sending event")
+		return nil
```
Naatan, line 102: This should happen in the constructor instead, otherwise we're checking the config on each event as well as opening ourselves up to race conditions with events that send at the end of the process.

See the sync client for example:

http://github.com/ActiveState/cli/blob/5e633be5fc4afb884136a2fd871abc75622e2cd7/internal/analytics/client/sync/client.go#L91-L91

internal/analytics/client/sync/client.go
```diff
@@ -83,10 +81,6 @@ func New(cfg *config.Instance, auth *authentication.Auth) *Client {
 		a.cfg = cfg
 	}
 
-	if a.cfg.IsSet(constants.ReportAnalayticsConfig) && !a.cfg.GetBool(constants.ReportAnalayticsConfig) {
-		a.sendReports = false
-	}
-
```
Naatan, line 89: Please revert, this was intentioanlly done on construction to avoid needless processing and race conditions.

internal/rollbar/rollbar.go
```diff
@@ -77,8 +94,9 @@ func Wait() { rollbar.Wait() }
 
 func logToRollbar(critical bool, message string, args ...interface{}) {
 	// only log to rollbar when on release, beta or unstable branch and when built via CI (ie., non-local build)
-	isPublicChannel := (constants.BranchName == constants.ReleaseBranch || constants.BranchName == constants.BetaBranch || constants.BranchName == constants.ExperimentalBranch)
-	if !isPublicChannel || !condition.BuiltViaCI() {
+	isPublicChannel := constants.BranchName == constants.ReleaseBranch || constants.BranchName == constants.BetaBranch || constants.BranchName == constants.ExperimentalBranch
+	reportingDisabled := currentCfg != nil && !currentCfg.Closed() && currentCfg.IsSet(constants.ReportErrorsConfig) && !currentCfg.GetBool(constants.ReportErrorsConfig)
+	if !isPublicChannel || !condition.BuiltViaCI() || !reportingDisabled {
```
Naatan, line 99: Please evaluate this during construction (eg. when the config is provided).

test/integration/rollbar_int_test.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func TestRollbarReportConfig(t *testing.T) {
+	ts := e2e.New(t, true)
+	defer ts.Close()
+
+	cp := ts.Spawn("config", "set", constants.ReportErrorsConfig, "false")
+	cp.Expect("Successfully")
+	cp.ExpectExitCode(0)
+}
```
Naatan, line 17: I think having a test for this is only really worth it if we also test that the config change had an effect. Otherise you're basically just testing whether `config set` works.

I'd consider this out of scope for the story, so feel free to disregard the test instead. Don't think we have any existing code that tests rollbar events.

Naatan (changes requested):

cmd/state-update-dialog/main.go
mitchell-as: I tried this initially, but there are places where we want to setup rollbar first, then load the config. If there's an error setting up the config, we want to report that to rollbar. If we wait to setup rollbar until after loading config, and that load fails, we'll never know about it.

internal/analytics/client/async/client.go
mitchell-as: Yes, you're right.

internal/analytics/client/sync/client.go
mitchell-as: Okay.

internal/rollbar/rollbar.go
mitchell-as: Okay.

test/integration/rollbar_int_test.go
mitchell-as: Well considering this ticket is all about the fact that `config set` didn't work for `report.errors`, I thought this was a worthwhile test. If a regression happens, it'll get caught. I agree that ideally it would test functionality, but I don't see an easy way to do that.

mitchell-as: Per our zoom call, I have removed this.

Naatan (approved):

Naatan (approved):

PR #1818, opened 2022-04-21
Added more helpful message on what to do if update fails in original subshell from installation.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-879" title="DX-879" target="_blank">BugDX-879</a>  CLI - UPDATE: Update prevented with `A State Tool update appears to already be running.`
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1763,7 +1763,7 @@ update_permission_err:
     Error received: {{.V1}}
 err_update_in_progress:
   other: |
-    A State Tool update appears to already be running.
+    A State Tool update already appears to be running.
     Please close any open shells, open a new one, and then try this command again.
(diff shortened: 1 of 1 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1764,6 +1764,7 @@ update_permission_err:
 err_update_in_progress:
   other: |
     A State Tool update appears to already be running.
```
rawktron, line 1766: Can we fix the split infinitive?

"A State Tool update appears to be already running." Otherwise good!

rawktron (changes requested):

internal/locale/locales/en-us.yaml
mitchell-as: Done.

rawktron (approved):

PR #1819, opened 2022-04-21
Tool halts when svc fails to start (and fix start wait timeout)

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-884" title="DX-884" target="_blank">BugDX-884</a>  state-svc failures should error out state tool
  </td></table>
  <br />

### DIFF
--- a/internal/svcctl/svcctl.go
+++ b/internal/svcctl/svcctl.go
@@ -1,5 +1,7 @@
 // Package svcctl provides functions that make use of an IPC device, as well as
-// common IPC handlers and requesters.
+// common IPC handlers and requesters. The intent is to guard the authority and
+// uniqueness of the state service, so localized error messages refer to the
+// "service" rather than just the IPC server.
 package svcctl
 
 import (
@@ -58,19 +60,19 @@ func EnsureExecStartedAndLocateHTTP(ipComm IPCommunicator, exec string) (addr st
 	addr, err = LocateHTTP(ipComm)
 	if err != nil {
 		if !errs.Matches(err, &ipc.ServerDownError{}) {
-			return "", err
+			return "", errs.Wrap(err, "Cannot locate HTTP port of service")
 		}
 
 		ctx, cancel := context.WithTimeout(context.Background(), commonTimeout)
 		defer cancel()
 
 		if err := startAndWait(ctx, ipComm, exec); err != nil {
-			return "", err
+			return "", errs.Wrap(err, "Cannot start service at %q", exec)
 		}
 
 		addr, err = LocateHTTP(ipComm)
 		if err != nil {
-			return "", errs.Wrap(err, "After start succeeded...")
+			return "", errs.Wrap(err, "Cannot locate HTTP port of service after start succeeded")
 		}
 	}
 
@@ -101,7 +103,7 @@ func StopServer(ipComm IPCommunicator) error {
 
 	err := stopAndWait(ctx, ipComm)
 	if err != nil && !errs.Matches(err, &ipc.ServerDownError{}) {
-		return err
+		return errs.Wrap(err, "Cannot stop service")
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


Naatan (changes requested):
This PR is doing too much. The story calls for returning an error. That's it.

Also note: The localized errors don't seem very well localized. Localized errors should be written with the end user in mind. And end users won't know what ipc means, for example.

daved:
@Naatan This PR returns the error, and for the error to have more meaning, it also adds some more info. The fix to the context timeout duration is critical for the error being returned to be accurate, and I did ask if I could include it. Should I remove the extra info and timeout increase? If kept, I will surely improve the user-facing info.

Naatan:
I can be convinced to leave most of the stuff in, but the localization changes don't make sense. Either revert them or truly localize them. Otherwise we're just kidding ourselves; these are not localized as is.

daved:
Here's an example of the current output:
```bash
$ state run build

Something Went Wrong
────────────────────
 x Failed to start state-svc at state tool invocation
 x Waiting for service startup confirmation failed
 x Timed out waiting for service to start (11.458s). Are you running software that could prevent State Tool from running local processes/servers?

Need More Help?
───────────────
```
or
```bash
daved@gm:cli{/svcstart_fail_halts_to…}$ state run build


Something Went Wrong
────────────────────
 x Failed to start state-svc at state tool invocation
 x Waiting for service startup confirmation failed
 x Ping encountered unexpected failure: some error message

Need More Help?
───────────────
```

daved:
Here's an example of the current output:
```bash
$ state run build

Something Went Wrong
────────────────────
 x Failed to start state-svc at state tool invocation
 x Waiting for service startup confirmation failed
 x Timed out waiting for service to start (11.458s). Are you running software that could prevent State Tool from running local processes/servers?

Need More Help?
───────────────
```
or
```bash
daved@gm:cli{/svcstart_fail_halts_to…}$ state run build


Something Went Wrong
────────────────────
 x Failed to start state-svc at state tool invocation
 x Waiting for service startup confirmation failed
 x Ping encountered unexpected failure: some error message

Need More Help?
───────────────
```

internal/svcctl/svcctl.go
```diff
@@ -50,19 +58,19 @@ func EnsureExecStartedAndLocateHTTP(ipComm IPCommunicator, exec string) (addr st
...
 
 		if err := startAndWait(ctx, ipComm, exec); err != nil {
-			return "", errs.Wrap(err, "Cannot start ipc server at %q", exec)
+			return "", err
 		}
 
 		addr, err = LocateHTTP(ipComm)
 		if err != nil {
-			return "", errs.Wrap(err, "Cannot locate HTTP port of ipc server after start succeeded")
+			return "", err
```
Naatan, line 73: This doesn't seem like an improvement?

internal/svcctl/svcctl.go
```diff
@@ -93,7 +101,7 @@ func StopServer(ipComm IPCommunicator) error {
 
 	err := stopAndWait(ctx, ipComm)
 	if err != nil && !errs.Matches(err, &ipc.ServerDownError{}) {
-		return errs.Wrap(err, "Cannot stop ipc server")
+		return err
```
Naatan, line 104: Not an improvement.

Naatan (changes requested):

internal/svcctl/svcctl.go
daved: The caller would report "Cannot stop server", then this function would report "Cannot stop and wait to confirm stop", then stopAndWait would report "Because {reasons}". So I found the message in this function redundant.

internal/svcctl/svcctl.go
daved: Same as the message in StopServer, these messages are mostly redundant, but the "after start" is very useful, so I've included that again.

Naatan: It's not just about the errors being useful, it's also about tracing where the error came from. You advocated we removed stack traces from errors, which we did because we should effectively get these from the error wrapping anyway. This is undermining that decision.

Naatan (changes requested):
Please wrap remaining unwrapped errors (see my comment).

Naatan (approved):

PR #1821, opened 2022-04-22
Added support for listening for config key changes, notifying state-svc, and reacting to them.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-893" title="DX-893" target="_blank">StoryDX-893</a>  state-svc observes report.* changes
  </td></table>
  <br />
 

The idea is to have subscribers listen for config keys they are interested in and execute a callback function when those keys change. Our pure Go SQLite driver does not have the ability to notify of database changes, so we have to do this manually.

Note that a C-Go SQLite driver (https://pkg.go.dev/github.com/mattn/go-sqlite3#SQLiteConn.RegisterUpdateHook) offers an update hook callback function that would probably be good enough, but we are not using that driver, presumably for cross-platform reasons.

### DIFF
--- a/cmd/state-svc/internal/resolver/resolver.go
+++ b/cmd/state-svc/internal/resolver/resolver.go
@@ -166,6 +166,6 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
 }
 
 func (r *Resolver) ConfigChanged(ctx context.Context, key string) (*graph.ConfigChangedResponse, error) {
-	configMediator.NotifyListeners(key)
+	go configMediator.NotifyListeners(key)
 	return &graph.ConfigChangedResponse{Received: true}, nil
 }
--- a/cmd/state-svc/internal/resolver/resolver.go
+++ b/cmd/state-svc/internal/resolver/resolver.go
@@ -20,6 +20,7 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/graph"
 	"github.com/ActiveState/cli/internal/logging"
+	configMediator "github.com/ActiveState/cli/internal/mediators/config"
 	"github.com/ActiveState/cli/internal/updater"
 	"github.com/ActiveState/cli/pkg/projectfile"
 	"github.com/patrickmn/go-cache"
@@ -162,4 +163,9 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
 	r.rtwatch.Watch(pid, exec, dims)
 
 	return &graph.RuntimeUsageResponse{Received: true}, nil
+}
+
+func (r *Resolver) ConfigChanged(ctx context.Context, key string) (*graph.ConfigChangedResponse, error) {
+	configMediator.NotifyListeners(key)
+	return &graph.ConfigChangedResponse{Received: true}, nil
 }
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -71,9 +71,6 @@ func main() {
 	}
 
 	runErr := run(cfg)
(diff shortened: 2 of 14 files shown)


cmd/state-svc/main.go
```diff
@@ -85,15 +87,9 @@ func main() {
 	}
 }
 
-func run(cfg *config.Instance) (rerr error) {
+func run(cfg *config.Instance) error {
 	args := os.Args
 
-	cfg, err := config.New()
```
mitchell-as, line 91: This is overwriting the `cfg` passed in, and apparently unnecessary.

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -163,3 +164,8 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
 
 	return &graph.RuntimeUsageResponse{Received: true}, nil
 }
+
+func (r *Resolver) ConfigChanged(ctx context.Context, key string) (*graph.ConfigChangedResponse, error) {
+	configMediator.NotifyListeners(key)
```
Naatan, line 169: This should happen as a subroutine, so as not block the request.

cmd/state-svc/main.go
```diff
@@ -72,6 +71,9 @@ func main() {
 	}
 
 	runErr := run(cfg)
+	if runErr == nil {
+		runErr = cfg.Close()
```
Naatan, line 75: This is already being handled around line 50

internal/runners/config/set.go
```diff
@@ -45,11 +50,23 @@ func (s *Set) Run(params SetParams) error {
...
 	if err != nil {
 		return locale.WrapError(err, "err_config_set", fmt.Sprintf("Could not set value %s for key %s", params.Value, params.Key))
 	}
 
+	// Notify listeners that this key has changed.
+	configMediator.NotifyListeners(key)
+
+	// Notify state-svc that this key has changed.
+	if s.svcModel != nil {
+		if err := s.svcModel.ConfigChanged(context.Background(), key); err != nil {
```
Naatan, line 65: Please double check this wouldn't create a recursion issue if we set config values in the state-svc.

internal/runners/config/set.go
```diff
@@ -45,11 +50,23 @@ func (s *Set) Run(params SetParams) error {
...
+			logging.Debug("Failed to report config change via state-svc: %s", errs.JoinMessage(err))
```
Naatan, line 66: This can be an error rather than a debug.

Naatan (changes requested):
See comments

internal/runners/config/set.go
mitchell-as: I added the conditional, but this wouldn't recurse because it happens in `state config set`, which would not be called by `state-svc`. I suppose even if it could, it would result in two notifications, but nothing more.

Naatan: Ahh I missed that in the review, in my head this was in the config package.

internal/condition/condition.go
```diff
@@ -26,4 +26,8 @@ func OnCI() bool {
 
 func BuiltViaCI() bool {
 	return constants.OnCI == "true"
-}
\ No newline at end of file
+}
+
+func InStateSvc() bool {
+	return strings.HasSuffix(os.Args[0], constants.StateSvcCmd)
```
Naatan, line 32: Should use `osutils.ExecutableName()`. But given my missconception about where the notification was sent I'm ok with just dropping this altogether. The state-svc will never run any runner code.

Naatan (changes requested):

Naatan (approved):

PR #1824, opened 2022-04-26
Fix heartbeat test

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-902" title="DX-902" target="_blank">BugDX-902</a>  Analytics integration test failure with heartbeats
  </td></table>
  <br />


internal/fileutils/fileutils_test.go
```diff
@@ -626,3 +627,20 @@ func testCaseSensitivePath(t *testing.T, dirName, variant string) {
...
+	v, err := PathsMatch(p1, p2)
+	require.NoError(t, err, errs.JoinMessage(err))
+
+	v1, err := ResolvePath(p1)
+	require.NoError(t, err, errs.JoinMessage(err))
+	v2, err := ResolvePath(p2)
+	require.NoError(t, err, errs.JoinMessage(err))
+
+	require.True(t, v, "PathsMatch should return true, path1: %s, path2: %s", v1, v2)
+}
```
MDrakos, line 646: nit: This test should run on all platforms and be expanded a bit. Right now it's covering a very narrow use-case for what could become a widely used function.

MDrakos (approved):

internal/fileutils/fileutils_test.go
Naatan: I've only experienced this issue on Mac so that's what I targeted. Can definitely expand if we find this issue on other platforms.

PR #1825, opened 2022-04-27
Use PID for logfile in foreground

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-830" title="DX-830" target="_blank">StoryDX-830</a>  Drop the PID mechanic from the state-svc
  </td></table>
  <br />
 

It doesn't appear that the PID is used for the state service. This just adds a line to the `status` command for the log file and removes some constants.

### DIFF
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -236,7 +236,7 @@ func runStatus(out output.Outputer) error {
 
 	out.Print(fmt.Sprintf("Port: %s", port))
 	out.Print(fmt.Sprintf("Dashboard: http://127.0.0.1%s", port))
-	out.Print(fmt.Sprintf("Log: %s\n", logging.FilePathFor(logfile)))
+	out.Print(fmt.Sprintf("Log: %s", logging.FilePathFor(logfile)))
 
 	return nil
 }
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -229,7 +229,7 @@ func runStatus(out output.Outputer) error {
 		return errs.Wrap(err, "Service cannot be reached")
 	}
 
-	logfile, err := svcctl.LocateLogFile(ipcClient)
+	logfile, err := svcctl.LogFileName(ipcClient)
 	if err != nil {
 		return errs.Wrap(err, "Service could not locate log file")
 	}
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -139,7 +139,7 @@ func run(cfg *config.Instance) error {
 			p, nil, nil,
 			func(ccmd *captain.Command, args []string) error {
 				logging.Debug("Running CmdStatus")
-				return runStatus(out, cfg)
+				return runStatus(out)
 			},
 		),
 		captain.NewCommand(
@@ -221,7 +221,7 @@ func runStop() error {
 	return nil
 }
 
-func runStatus(out output.Outputer, cfg *config.Instance) error {
+func runStatus(out output.Outputer) error {
 	ipcClient := svcctl.NewDefaultIPCClient()
 	// Don't run in background if we're already running
 	port, err := svcctl.LocateHTTP(ipcClient)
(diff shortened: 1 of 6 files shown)


daved (changes requested):

cmd/state-svc/main.go
```diff
@@ -231,7 +231,7 @@ func runStatus(out output.Outputer) error {
 
 	out.Print(fmt.Sprintf("Port: %s", port))
 	out.Print(fmt.Sprintf("Dashboard: http://127.0.0.1%s", port))
-	//fmt.Printf("Log: %s\n", logging.FilePathFor(logging.FileNameFor(*pid)))
+	out.Print(fmt.Sprintf("Log: %s\n", logging.FilePathFor(logging.FileNameFor(os.Getpid()))))
```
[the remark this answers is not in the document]
MDrakos: Oh, good point. This also surfaced an issue with `logging.FileNameFor`. It was giving me a file with an incorrect timestamp as that is generated at the time of the `state-svc status` invocation. So I've resorted to storing the file path when the service starts up.

cmd/state-svc/main.go
```diff
@@ -231,7 +231,7 @@ func runStatus(out output.Outputer) error {
 
 	out.Print(fmt.Sprintf("Port: %s", port))
 	out.Print(fmt.Sprintf("Dashboard: http://127.0.0.1%s", port))
-	out.Print(fmt.Sprintf("Log: %s\n", logging.FilePathFor(logging.FileNameFor(os.Getpid()))))
+	out.Print(fmt.Sprintf("Log: %s\n", cfg.GetString(constants.SvcLogConfig)))
```
daved, line 234: I feel fairly strongly that we should not be using the config db for IPC, and that this info should be passed through the authoritative instance of the service. In other words, I still think the service will likely benefit from having a PID tcp-endpoint, or we should just not report this info. @Naatan did indicate we can drop it and wait to see if we reach more of a need for that endpoint, but pinging him to be sure.

daved (changes requested):

cmd/state-svc/main.go
Naatan: Actually I just ran into a use-case for this myself: integration tests.

How trivial would it be to employ IPC for this? I agree it sounds like a better solution, just wondering about the cost vs value picture.

daved: Quite trivial. I can make a story with the AC being clear steps to take and it should take very little time.

Naatan: Ok, then let's go the IPC route.

Naatan: Just update the current story though, don't need a new one.

internal/svcctl/comm.go
```diff
@@ -34,6 +36,19 @@ func HTTPAddrHandler(addr string) ipc.RequestHandler {
 	}
 }
 
+func LogFileHandler(logFile string) ipc.RequestHandler {
+	return func(input string) (string, bool) {
+		if input == KeyLogFile {
+			return logging.FileName(), true
```
daved, line 42: Please return logFile instead of calling the logging package.

cmd/state-svc/service.go
```diff
@@ -51,6 +51,7 @@ func (s *service) Start() error {
 	spath := svcctl.NewIPCSockPathFromGlobals()
 	reqHandlers := []ipc.RequestHandler{ // caller-defined handlers to expand ipc capabilities
 		svcctl.HTTPAddrHandler(":" + strconv.Itoa(s.server.Port())),
+		svcctl.LogFileHandler(":" + strconv.Itoa(s.server.Port())),
```
daved, line 54: The value returned by logging.FileName() is needed here.

daved (changes requested):

daved (dismissed):

cmd/state-svc/main.go
```diff
@@ -229,9 +229,14 @@ func runStatus(out output.Outputer) error {
...
 
+	logfile, err := svcctl.LogFileName(ipcClient)
+	if err != nil {
+		return errs.Wrap(err, "Service could not locate log file")
+	}
+
 	out.Print(fmt.Sprintf("Port: %s", port))
 	out.Print(fmt.Sprintf("Dashboard: http://127.0.0.1%s", port))
-	//fmt.Printf("Log: %s\n", logging.FilePathFor(logging.FileNameFor(*pid)))
+	out.Print(fmt.Sprintf("Log: %s\n", logging.FilePathFor(logfile)))
```
Naatan, line 239:
```suggestion
	out.Print(fmt.Sprintf("Log: %s", logging.FilePathFor(logfile)))
```

`\n` appears redundant

internal/svcctl/svcctl.go
```diff
@@ -90,6 +90,22 @@ func LocateHTTP(ipComm IPCommunicator) (addr string, err error) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), commonTimeout)
+	defer cancel()
+
+	logfile, err := comm.GetLogFileName(ctx)
+	if err != nil {
+		return "", errs.Wrap(err, "Log file request failed")
+	}
+
+	logging.Debug("Log file name %s", logfile)
```
Naatan, line 104:
```suggestion
	logging.Debug("Service returned log file: %s", logfile)
```

Naatan (changes requested):

Naatan (approved):

PR #1827, opened 2022-04-27
Extend update qry timeout, comment with reasoning for timeout

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-860" title="DX-860" target="_blank">BugDX-860</a>  Mac e2e update test failing by timeout
  </td></table>
  <br />

### DIFF
--- a/pkg/cmdlets/checker/checker.go
+++ b/pkg/cmdlets/checker/checker.go
@@ -58,8 +58,13 @@ func CommitsBehind(p *project.Project) (int, error) {
 
 func RunUpdateNotifier(svc *model.SvcModel, out output.Outputer) {
 	defer profile.Measure("RunUpdateNotifier", time.Now())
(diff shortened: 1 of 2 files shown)


daved:
@Naatan The error that brought up this issue, if I recall and understand correctly, is fine. We wanted to cap the delay caused by an update lookup, so some requests that lag due to the service making a backend request will be clipped. I extended the timeout since it shouldn't happen often (unless unsuccessful repeatedly, but I suggest a new story to mitigate that if desirable). The main concern was whether the service startup itself was causing this, and it is not. If it were, it would be returning a "connection refused" error rather than "context deadline" error.

Naatan (changes requested):
Thanks for digging! Good that this is just a timeout working as intended. I've filed the following story to follow up:

https://activestatef.atlassian.net/browse/DX-907

Please close this PR, I don't think increasing the timeout is desirable as an interim solution.

daved:
Closing in favor of DX-907

PR #1828, opened 2022-04-28
Include device ID when adding a token

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-739" title="DX-739" target="_blank">StoryDX-739</a>  We record the device_id alongside the API token on the platform
  </td></table>
  <br />


pkg/cmdlets/auth/login.go
```diff
@@ -241,25 +240,9 @@ func AuthenticateWithBrowser(out output.Outputer, auth *authentication.Auth, pro
 		out.Notice(locale.Tr("err_browser_open", *response.VerificationURIComplete))
 	}
 
-	if !response.Nopoll {
```
MDrakos, line 244: `NoPoll` was removed from the response type. No new fields were added. Not sure if this is how we want to handle this.

Naatan (approved):

PR #1831, opened 2022-04-28
Prevent access to secrets unless `optin.unstable` is true.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-909" title="DX-909" target="_blank">StoryDX-909</a>  Secrets are unstable
  </td></table>
  <br />

### DIFF
--- a/internal/keypairs/mock/mock.go
+++ b/internal/keypairs/mock/mock.go
@@ -1,0 +1,41 @@
+package mock
+
+import (
+	"github.com/ActiveState/cli/internal/keypairs"
+)
+
+type Mock struct {
+	keypairs.Configurable
+	cfg map[string]interface{}
+}
+
+func (m *Mock) ConfigPath() string {
+	return ""
+}
+
+func (m *Mock) Close() error {
+	return nil
+}
+
+func (m *Mock) Set(key string, value interface{}) error {
+	if m.cfg == nil {
+		m.cfg = make(map[string]interface{})
+	}
+
+	m.cfg[key] = value
+
+	return nil
+}
+
+func (m *Mock) GetString(key string) string {
+	if value, found := m.cfg[key]; found {
+		return value.(interface{}).(string) // assume we stored the correct type
+	} else {
+		return ""
+	}
+}
+
+func (m *Mock) GetBool(key string) bool {
+	value, found := m.cfg[key]
+	return found && value.(interface{}).(bool) // assume we stored the correct type
+}
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1848,6 +1848,11 @@ unstable_command_warning:
     This feature [ACTIONABLE]state {{.V0}}[/RESET] is still in beta. If you want to opt-in to unstable features, run the following command:
 
     [ACTIONABLE]state config set optin.unstable true[/RESET]
+secrets_unstable_warning:
+  other: |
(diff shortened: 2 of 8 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1847,6 +1847,11 @@ unstable_command_warning:
   other: |
     This feature [ACTIONABLE]state {{.V0}}[/RESET] is still in beta. If you want to opt-in to unstable features, run the following command:
 
+    [ACTIONABLE]state config set optin.unstable true[/RESET]
+secrets_unstable_warning:
+  other: |
+    The use of secrets is still in beta. If you want to opt-in to unstable features, run the following command:
```
Naatan, line 1853:
```suggestion
    The secrets feature is currently unstable and not ready for production use. If you want to opt-in to unstable features, run the following command:
```

pkg/project/secrets.go
```diff
@@ -244,6 +246,10 @@ var ErrSecretNotFound = errors.New("secret not found")
 
 // Expand will expand a variable to a secret value, if no secret exists it will return an empty string
 func (e *SecretExpander) Expand(_ string, category string, name string, isFunction bool, project *Project) (string, error) {
+	if !e.cfg.GetBool(constants.UnstableConfig) && !condition.InUnitTest() {
```
Naatan, line 249: I'd prefer we avoid using `condition.InUnitTest()` if we can. I'd rather have the tests contain logic to satisfy our code instead of the other way around.

We could address this by moving the code after the config check into a separate function and testing that instead. ie. only test the "unit" we're interested in.

Naatan (changes requested):

pkg/project/secrets.go
mitchell-as: I agree. I originally thought I'd have to spawn a process to run `state config set optin.unstable true`, but I found an easier way after your comment.

pkg/project/secrets_test.go
```diff
@@ -77,6 +77,7 @@ func (suite *SecretsExpanderTestSuite) BeforeTest(suiteName, testName string) {
 	suite.graphMock.ProjectByOrgAndName(mock.NoOptions)
 
 	suite.cfg, err = config.New()
+	suite.cfg.Set(constants.UnstableConfig, true)
```
Naatan, line 80: Since this isn't mocking the config running this test would mess with your own local config. A simple config mock might be better, eg. http://github.com/ActiveState/cli/blob/e8c10335bc8ee70a5e7f01d4f91b55b7c472cef1/internal/updater/updater_test.go#L34-L34

Naatan (changes requested):

pkg/project/secrets_test.go
mitchell-as: Thanks for pointing this out. I had no idea.

internal/keypairs/mock/mock.go
```diff
@@ -0,0 +1,41 @@
+package mock
```
Naatan, line 1: Please rename this file to `mock_test.go`

Naatan (changes requested):

internal/keypairs/mock/mock.go
mitchell-as: I couldn't do that because then the structs would not be found. I moved this package into the `internal/testhelpers/` area, as this was done with other mocks like secretsapi. Let me know if you prefer a different package name.

Naatan (approved):

PR #1836, opened 2022-05-03
Implemented LTS mechanic

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-764" title="DX-764" target="_blank">StoryDX-764</a>  We have an LTS version of the State Tool
  </td></table>
  <br />


.github/workflows/build.yml
```diff
@@ -183,7 +184,7 @@ jobs:
       - # === Sign Install Scripts (Windows only) ===
         name: Sign Install Scripts (Windows only)
         shell: powershell
-        if: runner.os == 'Windows' && contains(fromJSON('["refs/heads/master", "refs/heads/beta", "refs/heads/release"]'), github.ref)
+        if: runner.os == 'Windows' && contains(fromJSON('["refs/heads/master", "refs/heads/beta", "refs/heads/release", "refs/heads/LTS2022"]'), github.ref)
```
mitchell-as, line 187: This is going to get long as we have more and more LTS heads. How about instead of using `contains()`, using `github.event_name == 'push'`? Right now actions are run on `push` or `pull`, and all of the heads mentioned here are in the `push` event.

.github/workflows/build.yml
```diff
@@ -212,7 +213,7 @@ jobs:
           echo "Labels: $LABELS"
           echo "Target Branch: $TARGET_BRANCH"
 
-          if [ "$LABELCHECK" == "" ] && [ "$TARGET_BRANCH" != "beta" ] && [ "$TARGET_BRANCH" != "release" ]; then
+          if [ "$LABELCHECK" == "" ] && [ "$TARGET_BRANCH" != "beta" ] && [ "$TARGET_BRANCH" != "release" ] && [ "$TARGET_BRANCH" != "LTS2022" ]; then
```
mitchell-as, line 216: It would be nice to clean this up in a similar way. Either replace the multiple `&&` branches with `${{ github.* }}` or check for an LTS prefix like ``[[ ! -z "`echo $TARGET_BRANCH | grep -o '^LTS'`" ]]``

.github/workflows/build.yml
```diff
@@ -283,7 +284,7 @@ jobs:
       SHELL: bash
       GITHUB_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     timeout-minutes: 10
-    if: contains(fromJSON('["refs/heads/master", "refs/heads/beta", "refs/heads/release"]'), github.ref)
+    if: contains(fromJSON('["refs/heads/master", "refs/heads/beta", "refs/heads/release", "refs/heads/LTS2022"]'), github.ref)
```
mitchell-as, line 287: Same concerns about growth over time.

mitchell-as (changes requested):

.github/workflows/build.yml
```diff
@@ -212,7 +213,7 @@ jobs:
           echo "Labels: $LABELS"
           echo "Target Branch: $TARGET_BRANCH"
 
-          if [ "$LABELCHECK" == "" ] && [ "$TARGET_BRANCH" != "beta" ] && [ "$TARGET_BRANCH" != "release" ]; then
+          if [ "$LABELCHECK" == "" ] && [ "$TARGET_BRANCH" != "beta" ] && [ "$TARGET_BRANCH" != "release" ] && [ "$TARGET_BRANCH" != "LTS"* ]; then
```
mitchell-as, line 216: That doesn't seem right. I ran `if [ "LTS2022" != "LTS"* ]; then echo 'hi'; fi` in my shell and it printed 'hi', which it probably shouldn't for your purposes.

mitchell-as (changes requested):

.github/workflows/build.yml
Naatan: Bah, seems this is a zsh feature.

mitchell-as (approved):

PR #1837, opened 2022-05-03
Initial nightly integration test workflow.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-894" title="DX-894" target="_blank">StoryDX-894</a>  We run our full integration test suite nightly and we have a process for following up on them
  </td></table>
  <br />
 

Procedure documentation is here: https://github.com/ActiveState/TheHomeRepot/pull/8312

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -315,7 +315,7 @@ jobs:
                     "type": "plain_text",
                     "text": "Nightly integration test failure(s) on ${{ runner.os }}"
                   }
-                }
+                },
                 {
                   "type": "section",
                   "text": {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -202,20 +202,22 @@ jobs:
 
       - # === Deploy for Integration Tests # NEVER run this against production branches. This is meant for PR deployments. ===
         name: Deploy for Integration Tests # NEVER run this against production branches. This is meant for PR deployments.
-        if: github.ref != 'refs/heads/release' && github.ref != 'refs/heads/master'
+        if: github.ref != 'refs/heads/release'
         shell: bash
         run: |
-          set +e
-          LABELS="${{ join(github.event.pull_request.labels.*.name, ',') }}"
-          LABELCHECK="$(echo $LABELS | grep "Test:")" # This line is causing a non-zero exit if `set -e` is set, and I can't figure out why
-          TARGET_BRANCH="${{ github.event.pull_request.base.ref }}"
-          set -e
-          echo "Labels: $LABELS"
-          echo "Target Branch: $TARGET_BRANCH"
+          if [ "$GITHUB_EVENT_NAME" != "schedule" ]; then
+            set +e
+            LABELS="${{ join(github.event.pull_request.labels.*.name, ',') }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -35,7 +37,6 @@ jobs:
       GOFLAGS: -mod=vendor
       SHELL: bash
       GITHUB_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    timeout-minutes: 30
```
mitchell-as, line 38: Instead of specifying timeout here, do so in the integration test runners that would be responsible for such a timeout. Manual integration test runs have a 10m timeout, and automated nightly integration test runs have a 90m timeout.

.github/workflows/build.yml
```diff
@@ -201,7 +202,7 @@ jobs:
 
       - # === Deploy for Integration Tests # NEVER run this against production branches. This is meant for PR deployments. ===
         name: Deploy for Integration Tests # NEVER run this against production branches. This is meant for PR deployments.
-        if: github.ref != 'refs/heads/release'
+        if: github.ref != 'refs/heads/release' && github.ref != 'refs/heads/master'
```
Naatan, line 205: This feels wrong, now this job has implicit logic around nightly integration tests that might end up biting us later. You can use the conditional you used for the job on line 229 inside the run logic instead.

.github/workflows/build.yml
```diff
@@ -13,6 +13,8 @@ name: Build-Test-Deploy
       - opened
       - synchronize
       - reopened
+  schedule:
+    - cron: 0 0 * * *
```
Naatan, line 17: Shouldn't this also target a branch? Otherwise this would end up running for all branches with this workflow on it.

.github/workflows/build.yml
```diff
@@ -248,14 +260,40 @@ jobs:
...
+          INTEGRATION_TEST_PASSWORD: ${{ secrets.INTEGRATION_TEST_PASSWORD }}
+          INTEGRATION_TEST_TOKEN: ${{ secrets.INTEGRATION_TEST_TOKEN }}
+          PLATFORM_API_TOKEN: ${{ secrets.PLATFORM_API_TOKEN }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      - # === Nightly Integration Tests ===
+        name: Nightly Integration Tests
+        id: nightly_integration_tests
+        if: github.event_name == 'schedule'
```
Naatan, line 286: Don't need a separate job, just use this conditional inside the script logic for the job above.

Naatan (changes requested):

.github/workflows/build.yml
mitchell-as: No, schedule only works on the "default" branch, which happens to be "master" for us.

.github/workflows/build.yml
mitchell-as: Okay, I'll merge the two separate integration test steps into one.

.github/workflows/build.yml
mitchell-as: Ok.

Naatan (approved):

PR #1839, opened 2022-05-04
Only deploy as latest when label is set

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-102" title="DX-102" target="_blank">StoryDX-102</a>  We can deploy specific versions without making them "latest"
  </td></table>
  <br />

### DIFF
--- a/scripts/update-generator/main.go
+++ b/scripts/update-generator/main.go
@@ -27,7 +27,7 @@ var latestFlag *bool
 
 func printUsage() {
 	fmt.Println("")
-	fmt.Println("[-o outputDir] [-b branchOverride] [-v versionOverride] [--platform platformOverride] --latest <directory>")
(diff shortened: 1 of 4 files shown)


scripts/update-generator/main.go
```diff
@@ -15,20 +15,19 @@ import (
...
 
 var exit = os.Exit
 
 var outputDirFlag, platformFlag, branchFlag, versionFlag *string
+var latestFlag *bool
 
 func printUsage() {
 	fmt.Println("")
-	fmt.Println("[-o outputDir] [-b branchOverride] [-v versionOverride] [--platform platformOverride] <directory>")
+	fmt.Println("[-o outputDir] [-b branchOverride] [-v versionOverride] [--platform platformOverride] --latest <directory>")
```
mitchell-as, line 30: Shouldn't this be `[--latest]` becase it's optional?

mitchell-as (changes requested):

PR #1843, opened 2022-05-04
Implement deprecation test and fix deprecation mechanic

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-877" title="DX-877" target="_blank">StoryDX-877</a>  We have an integration test for the deprecation mechanic
  </td></table>
  <br />


MDrakos (commented):
Look good but can we run the integration tests you added on this PR just to be certain?

PR #1845, opened 2022-05-04
Move Appinfo

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-843" title="DX-843" target="_blank">StoryDX-843</a>  Get rid of appinfo and centralize on installation package
  </td></table>
  <br />


cmd/state-installer/cmd.go
```diff
@@ -355,12 +354,18 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		return errs.Wrap(err, "Could not resolve installation path")
 	}
 
-	stateExe := appinfo.StateApp(installPath).Exec()
 	binPath, err := installation.BinPathFromInstallPath(installPath)
 	if err != nil {
 		return errs.Wrap(err, "Could not detect installation bin path")
 	}
 
+	stateInfo, err := installation.NewAppInfoInDir(installPath, installation.StateApp)
```
Naatan, line 362: This is slightly nitpicky, but I feel like this is still more complex than it needs to be. The appinfo architecture evolved organically and was never designed from the ground up. I feel like this PR is just perpetuating the organic design rather than revisiting it for the use-cases we have.

Looking at the data that AppInfo is returning (exec and name), I'm not seeing any reason for us to keep using this central struct based approach. It feels like we're trying to be too clever.

`installation.StateExec()` and `installation.StateName()` or something along those lines would service the exact same use-cases, without being needlessly clever. Dumb code that just works feels more desirable here because frankly there isn't much here for us to be smart with.

All of this I'm caveating as a nitpick though. Because technically all of this works fine, and I see no holes in the high level logic aside from that it feels like a roundabout way of facilitating the use-cases we have.

cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -41,8 +42,8 @@ func (suite *InstallerIntegrationTestSuite) TestInstallFromLocalSource() {
 	suite.NotContains(cp.TrimmedSnapshot(), "Downloading State Tool")
 
 	// Assert expected files were installed (note this didn't use an update payload, so there's no bin directory)
-	suite.FileExists(appinfo.StateApp(target).Exec())
-	suite.FileExists(appinfo.SvcApp(target).Exec())
+	suite.FileExists(filepath.Join(target, installation.BinDirName, constants.StateCmd+osutils.ExeExt))
+	suite.FileExists(filepath.Join(target, installation.BinDirName, constants.ServiceCommandName+osutils.ExeExt))
```
Naatan, line 46: Why not use appinfo here?

cmd/state-tray/internal/menu/projects.go
```diff
@@ -54,7 +55,14 @@ func (u *LocalProjectsUpdater) Update(projects []*graph.Project) {
 		u.items = append(u.items, &localProjectsMenuItem{mitem, "", "", cb, make(chan struct{})})
 	}
 
-	u.startEventLoops()
+	stateApp, err := installation.NewAppInfo(installation.StateApp)
+	if err != nil {
+		return locale.WrapError(err, "err_state_info")
+	}
```
Naatan, line 61: Nit: Feels more like something to set during construction rather than fed into the Update call.

cmd/state/autoupdate.go
```diff
@@ -84,12 +84,19 @@ func autoUpdate(args []string, cfg *config.Instance, out output.Outputer) (bool,
 			logging.Debug("Update already in progress")
 			return false, nil
 		}
+		fmt.Println("err:", err)
```
Naatan, line 87: Debugging?

cmd/state/autoupdate.go
```diff
@@ -84,12 +84,19 @@ func autoUpdate(args []string, cfg *config.Instance, out output.Outputer) (bool,
...
 	out.Notice(locale.Tr("auto_update_relaunch"))
 	out.Notice("") // Ensure output doesn't stick to our messaging
-	code := relaunch(args)
+
+	stateInfo, err := installation.NewAppInfo(installation.StateApp)
+	if err != nil {
+		return false, locale.WrapError(err, "err_state_info")
+	}
+
+	code := relaunch(stateInfo.Exec(), args)
```
Naatan, line 99: This feels like it's logic for inside of `relaunch`, not something we pass to it.

internal/installation/appinfo.go
```diff
@@ -0,0 +1,86 @@
...
+
+	pathEvaled, err := filepath.EvalSymlinks(path)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not eval symlinks")
+	}
+	path = pathEvaled
+
+	info := appdata[exec]
+	info.executable = filepath.Join(filepath.Dir(path), info.cmd+osutils.ExeExt)
+	return info, nil
```
Naatan, line 57: I think most of this belongs in `Exec()`.

internal/installation/appinfo.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func NewAppInfoInDir(baseDir string, exec executable) (*AppInfo, error) {
```
Naatan, line 60: To reduce the number of code paths and keep things consistent it might be good if `NewAppInfo` invoked `NewAppInfoInDir()`. Ie. `NewAppInfo` is basically just `NewAppInfoInDir` except it's opinionated about the dir.

test/integration/install_scripts_int_test.go
```diff
@@ -102,8 +102,7 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
 
 			cp.ExpectExitCode(0)
 
-			state := appinfo.StateApp(installDir)
-			suite.FileExists(state.Exec())
+			suite.FileExists(filepath.Join(installDir, installation.BinDirName, constants.StateCmd+osutils.ExeExt))
```
Naatan, line 105: Use AppInfo?

test/integration/install_scripts_int_test.go
```diff
@@ -203,8 +202,7 @@ func (suite *InstallScriptsIntegrationTestSuite) assertCorrectVersion(ts *e2e.Se
 		Branch  string `json:"branch"`
 	}
 
-	state := appinfo.StateApp(installDir)
-	cp := ts.SpawnCmd(state.Exec(), "--version", "--output=json")
+	cp := ts.SpawnCmd(filepath.Join(installDir, installation.BinDirName, constants.StateCmd+osutils.ExeExt), "--version", "--output=json")
```
Naatan, line 205: Use AppInfo?

test/integration/update_int_test.go
```diff
@@ -135,7 +134,7 @@ func (suite *UpdateIntegrationTestSuite) testUpdate(ts *e2e.Session, baseDir str
 		spawnOpts = append(spawnOpts, opts...)
 	}
 
-	cp := ts.SpawnCmdWithOpts(stateExe.Exec(), spawnOpts...)
+	cp := ts.SpawnCmdWithOpts(filepath.Join(baseDir, installation.BinDirName, constants.StateCmd+osutils.ExeExt), spawnOpts...)
```
Naatan, line 137: Use AppInfo?

test/integration/update_int_test.go
```diff
@@ -274,10 +271,10 @@ func (suite *UpdateIntegrationTestSuite) testAutoUpdate(ts *e2e.Session, baseDir
 		spawnOpts = append(spawnOpts, opts...)
 	}
 
-	cp := ts.SpawnCmdWithOpts(stateExe.Exec(), spawnOpts...)
+	cp := ts.SpawnCmdWithOpts(filepath.Join(baseDir, "bin", constants.StateCmd+osutils.ExeExt), spawnOpts...)
```
Naatan, line 274: Use AppInfo?

test/integration/update_int_test.go
```diff
@@ -298,9 +295,9 @@ func (suite *UpdateIntegrationTestSuite) installLatestReleaseVersion(ts *e2e.Ses
 			e2e.AppendEnv("SHELL="),
 		)
 	}
-	cp.Expect("Installation Complete", time.Second*30)
+	cp.Expect("Installation Complete", 5*time.Minute)
 
-	suite.FileExists(appinfo.StateApp(dir).Exec())
+	suite.FileExists(filepath.Join(dir, installation.BinDirName, constants.StateCmd+osutils.ExeExt))
```
Naatan, line 300: Use AppInfo?

test/integration/upgen_int_test.go
```diff
@@ -52,8 +52,7 @@ func (suite *UpdateGenIntegrationTestSuite) TestUpdateBits() {
 
 	cp.ExpectExitCode(0)
 
-	state := appinfo.StateApp(filepath.Join(tempPath, constants.ToplevelInstallArchiveDir, "bin"))
-	cp = ts.SpawnCmd(state.Exec(), "--version")
+	cp = ts.SpawnCmd(filepath.Join(tempPath, constants.ToplevelInstallArchiveDir, "bin", constants.StateCmd+osutils.ExeExt), "--version")
```
Naatan, line 55: Use AppInfo?

Naatan (changes requested):

cmd/state-installer/test/integration/installer_int_test.go
MDrakos: Because of the test conditional, we end up with an `appinfo` from the build dir and not the one that was just installed

MDrakos: Yeah, I'm working on a fix for it.

internal/installation/appinfo.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	} else {
+		path, err = os.Executable()
+		if err != nil {
+			multilog.Error("Could not determine executable: %v", err)
+			path, err = filepath.Abs(os.Args[0])
+			if err != nil {
+				return "", errs.Wrap(err, "Could not get absolute directory of os.Args[0]")
+			}
+		}
```
Naatan, line 51: You can use `osutils.Executable()` instead.

internal/runners/clean/run_lin_mac.go
```diff
@@ -158,15 +157,30 @@ func removeEmptyDir(dir string) error {
...
+
+	serviceExec, err := installation.NewExec(installation.ServiceExec)
+	if err != nil {
+		return locale.WrapError(err, "err_service_exec")
+	}
+
+	trayExec, err := installation.NewExec(installation.TrayExec)
+	if err != nil {
+		return locale.WrapError(err, "err_tray_exec")
+	}
```
Naatan, line 173: Perhaps we could have a single `installation.Executables()` function which enumerates ALL executables. The we only need a single error assertion.

Naatan (changes requested):

internal/runners/clean/run_lin_mac.go
MDrakos: Good idea!

Naatan (approved):

PR #1846, opened 2022-05-04
Remove svc stop from clean config runner

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-865" title="DX-865" target="_blank">BugDX-865</a>  state clean config does not work if state-svc could not be stopped
  </td></table>
  <br />


daved:
@Naatan I am a bit concerned about removing the .db file while the service is running, so I did test and found the service to seemingly work fine. It doesn't really settle my concerns, though.

PR #1847, opened 2022-05-05
Updated installation scripts to allow installing specific versions via '-v' option.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-916" title="DX-916" target="_blank">BugDX-916</a>  CLI - install.sh: `-v` flag is not working.
  </td></table>
  <br />

### DIFF
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -139,7 +139,7 @@ catch [System.Exception]
     progress_fail
     Write-Error "Could not download $zipURL to $zipPath."
     Write-Error $_.Exception.Message
-    return 1
+    exit 1
 }
 
 # Verify checksum if possible.
@@ -150,7 +150,7 @@ if ($checksum -and $hash -ne $checksum)
     Write-Warning "Expected: $checksum"
     Write-Warning "Received: $hash"
     Write-Warning "Aborting installation"
-    return 1
+    exit 1
 }
 
 # Extract it.
@@ -162,7 +162,7 @@ catch
 {
     progress_fail
     Write-Error $_.Exception.Message
-    return 1
+    exit 1
 }
 progress_done
 
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -66,12 +66,12 @@ function download([string] $url, [string] $out)
         {
             if ($Retrycount -gt 5)
             {
-                Write-Error "Could not Download after 5 retries."
+                Write-Error "Could not download after 5 retries."
                 throw $_
             }
             else
             {
-                Write-Host "Could not Download, retrying..."
+                Write-Host "Could not download, retrying..."
                 Write-Host $_
                 $Retrycount = $Retrycount + 1
             }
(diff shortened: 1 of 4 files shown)


installers/install.ps1
```diff
@@ -155,7 +162,7 @@ catch
 {
     progress_fail
     Write-Error $_.Exception.Message
-    return 1
+    exit 1
```
Naatan, line 165: Was this necessary? Little concerned of whatever the result of this might be. Were we just exiting zero before?

installers/install.sh
```diff
@@ -99,32 +107,39 @@ if [ -z "$TMPDIR" ]; then
...
-rm $TMPDIR/info.json
 
-# Fetch the latest version.
+# Fetch the requested or latest version.
 progress "Preparing Installer for State Tool Package Manager version $VERSION"
 STATEURL="$BASE_FILE_URL/$RELURL"
 ARCHIVE="$OS-amd64$DOWNLOADEXT"
-if ! $FETCH $TMPDIR/$ARCHIVE $STATEURL ; then
+$FETCH $TMPDIR/$ARCHIVE $STATEURL
+if [ $? -ne 0 -o \( "`echo $FETCH | grep -o 'curl'`" == "curl" -a -z "`file -b $TMPDIR/$ARCHIVE | grep -o 'data'`" \) ]; then
```
Naatan, line 134: I'm not sure I follow what is happening here. Could you add a comment explaining?

test/integration/install_scripts_int_test.go
```diff
@@ -152,6 +152,31 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall_NonEmptyTarget() {
 	cp.ExpectExitCode(1)
 }
 
+func (suite *InstallScriptsIntegrationTestSuite) TestInstall_VersionDoesNotExist() {
```
Naatan, line 155: Kudos!

Naatan (changes requested):

installers/install.ps1
mitchell-as: Yes, we were exiting zero before.

installers/install.sh
mitchell-as: Thanks for the reminder. I totally meant to do that.

installers/install.sh
```diff
@@ -131,6 +131,11 @@ progress "Preparing Installer for State Tool Package Manager version $VERSION"
 STATEURL="$BASE_FILE_URL/$RELURL"
 ARCHIVE="$OS-amd64$DOWNLOADEXT"
 $FETCH $TMPDIR/$ARCHIVE $STATEURL
+# wget and curl differ on how to handle AWS' "Forbidden" result for unknown versions.
+# wget will exit with nonzero status. curl simply creates an XML file with the forbidden error.
+# If curl was used, make sure the file downloaded is of type 'data', according to the UNIX `file`
+# command. (The XML error will be reported as a 'text' type.)
+# If wget returned an error or curl fetched a "forbidden" response, raise an error and exit.
```
Naatan, line 138: Nice find!

Naatan (approved):

PR #1848, opened 2022-05-05
Specify channel to installer.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-923" title="DX-923" target="_blank">BugDX-923</a>  Our install.ps1 does not have the branch injected on deploy
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -40,7 +40,6 @@ type Params struct {
 	sourceInstaller string
 	path            string
 	updateTag       string
-	branch          string
 	command         string
 	force           bool
 	activate        *project.Namespaced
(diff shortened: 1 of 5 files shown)


cmd/state-installer/cmd.go
```diff
@@ -132,11 +132,6 @@ func main() {
 				Description: "Defaults to 'release'.  Specify an alternative channel to install from (eg. beta)",
 				Value:       &params.branch,
 			},
-			{
-				Shorthand: "b", // backwards compatibility
-				Hidden:    true,
-				Value:     &params.branch,
-			},
```
Naatan, line 139: Good catch! There's also the channel flag though. And the branch definition itself in the types.

installers/install.ps1
```diff
@@ -175,7 +175,7 @@ $PSDefaultParameterValues['*:Encoding'] = 'utf8'
 
 # Run the installer.
 $env:ACTIVESTATE_SESSION_TOKEN = $script:SESSION_TOKEN_VALUE
-& $exePath $args --source-installer="install.ps1"
+& $exePath $args --channel="$script:CHANNEL" --source-installer="install.ps1"
```
Naatan, line 178: I don't think there's any reason for us to pass this at all anymore. The script is what handles the channel flag. The installer just installs the payload it has alongside it, and has the channel baked into it.

Naatan (changes requested):

Naatan (approved):

PR #1853, opened 2022-05-06
Fix integration tests

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-925" title="DX-925" target="_blank">BugDX-925</a>  Fix failing integration tests
  </td></table>
  <br />


internal/locale/locale.go
```diff
@@ -45,6 +45,7 @@ func init() {
 	if locale == "" {
 		cfg, err := config.New()
 		if err == nil {
+			defer cfg.Close()
```
MDrakos, line 48: Should we use the closer here so any close errors are logged?

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -178,9 +173,9 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
 func (r *Resolver) CheckDeprecation(ctx context.Context) (*graph.DeprecationInfo, error) {
 	logging.Debug("Check deprecation resolver")
 
-	deprecated, err := r.deprecation.Check()
-	if err != nil {
-		return nil, errs.Wrap(err, "Could not fetch deprecation information")
+	deprecated, ok := r.depPoller.ValueFromCache().(*graph.DeprecationInfo)
```
MDrakos, line 176: This won't panic if the value from `ValueFromCache()` is nil right? I'm pretty sure it won't. Might be something nice to confirm in the unit test for the `poller`

test/integration/upgen_int_test.go
```diff
@@ -52,8 +57,8 @@ func (suite *UpdateGenIntegrationTestSuite) TestUpdateBits() {
 
 	cp.ExpectExitCode(0)
 
-	stateExec, err := installation.StateExecFromDir(filepath.Join(tempPath, constants.ToplevelInstallArchiveDir))
-	suite.NoError(err)
+	baseDir := filepath.Join(tempPath, constants.ToplevelInstallArchiveDir)
+	stateExec := filepath.Join(baseDir, installation.BinDirName, constants.StateCmd+osutils.ExeExt)
```
MDrakos, line 61: This should work with `installation.StateExecFromDir` and I'm pretty sure it was in the PR that moved appinfo. What changed?

MDrakos (dismissed):
Looks good to me overall. Lots of convenient changes here. 

My comments are mostly just questions and nits.

internal/testhelpers/e2e/session.go
```diff
@@ -180,6 +185,7 @@ func new(t *testing.T, retainDirs, updatePath bool, extraEnv ...string) *Session
 		constants.ProjectEnvVarName + "=",
 		constants.E2ETestEnvVarName + "=true",
 		constants.DisableUpdates + "=true",
+		constants.OptinUnstableEnvVarName + "=true",
```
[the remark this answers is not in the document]
Naatan: Yep, to overwrite you just specify the env var with an empty value. Or really any value other than true.

internal/locale/locale.go
Naatan: This is an init function, so logging may not be available. I'd rather not risk making this a bigger issue than it needs to be, especially considering long term we'd be getting rid of this init function anyway.

cmd/state-svc/internal/resolver/resolver.go
Naatan: I can't confirm it in the poller, because the value type is an interface and we can't reliably assert nil unless we know the type.

We're not doing anything with the returned value here, other than also return it. And as per the contract of the function it CAN return a nil, which isn't a new change.

test/integration/upgen_int_test.go
Naatan: That function is for using information from the installed state tool. This test is using a generated payload. While those are similar things they are not the same.

I believe in this case it would fail because the payload has no install dir marker.

PR #1854, opened 2022-05-09
Enforce separation of installation payload and source.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-869" title="DX-869" target="_blank">StoryDX-869</a>  Separate "installed state tool" from "install payload" in our integration tests
  </td></table>
  <br />
 

The test failures are coming from the inability to run `state-svc stop`. This PR's tests are passing on CI as well as on my local machine.

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -234,6 +234,12 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		}
 	}
 
+	if condition.OnCI() {
+		if _, err := storage.InstallSource(); err == nil {
+			return errs.New("Cannot run state-installer from an installation directory. Run it from an installer instead.")
+		}
+	}
+
 	// Detect state tool alongside installer executable
 	installerPath := filepath.Dir(osutils.Executable())
 	packagedStateExe := filepath.Join(installerPath, installation.BinDirName, constants.StateCmd+exeutils.Extension)
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ActiveState/cli/internal/analytics/client/sync"
 	"github.com/ActiveState/cli/internal/appinfo"
 	"github.com/ActiveState/cli/internal/captain"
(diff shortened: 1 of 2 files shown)


cmd/state-installer/cmd.go
```diff
@@ -234,6 +235,12 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		}
 	}
 
+	if condition.OnCI() {
+		if _, err := storage.InstallSource(); err == nil {
+			return errs.New("Cannot run state-installer from an installation directory. Run it from an installer instead.")
+		}
+	}
```
Naatan, line 242: This has multiple issues:

- You're ignoring the actual error returned and making assumptions about what it means.
- InstallSource's job isn't to assert whether we are running from an install directory, that's a connotation you're only creating here, making this very error prone wrt future iterations of InstallSource.
- This type of assertion should happen in tests, not in the code we ship.

cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -21,24 +25,64 @@ type InstallerIntegrationTestSuite struct {
...
+	b, err := download.GetDirect(archiveUrl)
+	suite.Require().NoError(err)
+
+	// Extract it.
+	installerDir := filepath.Join(ts.Dirs.Work, "installer")
+	if runtime.GOOS != "windows" {
+		suite.Require().NoError(unarchiver.NewTarGzBlob(b).Unarchive(installerDir))
+	} else {
+		suite.Require().NoError(unarchiver.NewZipBlob(b).Unarchive(installerDir))
+	}
```
Naatan, line 72: This test is specifically testing the local source (note the suffix: `FromLocalSource()`).

So this is breaking the test.

Naatan (changes requested):
This doesn't seem to be addressing the story.

The issue is that ts.InstallerExe is using the executables in ts.Dirs.Bin as the installation payload. While those are also the executables used to actually run tests. The story here is to ensure those are separate files and directories.

I think when we set up the test suite InstallerExe should be copied to a distinct directory with a distinct payload.

PR #1856, opened 2022-05-10
 Make activation project not found error message more visible.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-931" title="DX-931" target="_blank">StoryDX-931</a>  Our installer conveys errors from forwarded state tool activation more prominently
  </td></table>
  <br />
 

Error tips can be disabled using an environment variable, and the installer sets this environment variable when invoking the `state` tool (i.e. activating an environment upon install or running a command).

Also refactored and centralized silent error and exit code handling.

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -389,7 +389,10 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 }
 
 func envSlice(binPath string) []string {
-	return []string{"PATH=" + binPath + string(os.PathListSeparator) + os.Getenv("PATH")}
+	return []string{
+		"PATH=" + binPath + string(os.PathListSeparator) + os.Getenv("PATH"),
+		constants.DisableErrorTipsEnvVarName + "=true",
+	}
 }
 
 // storeInstallSource writes the name of the install client (eg. install.sh) to the appdata dir
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -355,7 +355,7 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		cmd, args := exeutils.DecodeCmd(params.command)
 		if _, _, err := exeutils.ExecuteAndPipeStd(cmd, args, envSlice(binPath)); err != nil {
 			an.EventWithLabel(AnalyticsFunnelCat, "forward-command-err", err.Error())
-			return errs.Wrap(err, "Running provided command failed, error returned: %s", errs.JoinMessage(err))
+			return errs.Silence(err)
 		}
 	// Activate provided --activate Namespace
 	case params.activate.IsValid():
@@ -364,7 +364,7 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		out.Print(fmt.Sprintf("\nRunning `[ACTIONABLE]state activate %s[/RESET]`\n", params.activate.String()))
 		if _, _, err := exeutils.ExecuteAndPipeStd(stateExe, []string{"activate", params.activate.String()}, envSlice(binPath)); err != nil {
 			an.EventWithLabel(AnalyticsFunnelCat, "forward-activate-err", err.Error())
-			return errs.Wrap(err, "Could not activate %s, error returned: %s", params.activate.String(), errs.JoinMessage(err))
+			return errs.Silence(err)
 		}
 	// Activate provided --activate-default Namespace
 	case params.activateDefault.IsValid():
@@ -373,7 +373,7 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		out.Print(fmt.Sprintf("\nRunning `[ACTIONABLE]state activate --default %s[/RESET]`\n", params.activateDefault.String()))
 		if _, _, err := exeutils.ExecuteAndPipeStd(stateExe, []string{"activate", params.activateDefault.String(), "--default"}, envSlice(binPath)); err != nil {
 			an.EventWithLabel(AnalyticsFunnelCat, "forward-activate-default-err", err.Error())
-			return errs.Wrap(err, "Could not activate %s, error returned: %s", params.activateDefault.String(), errs.JoinMessage(err))
+			return errs.Silence(err)
 		}
 	case !isUpdate:
 		ss := subshell.New(cfg)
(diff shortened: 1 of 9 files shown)


MDrakos (commented):
During stand it was mentioned that the installer should not print the `Need More Help?` section when forwarding commands to the state tool. @Naatan can you confirm if that's something we want?

Naatan:
> During stand it was mentioned that the installer should not print the `Need More Help?` section when forwarding commands to the state tool. @Naatan can you confirm if that's something we want?

Right. I think the issue here is that we need to slim down the output. Not make some individual line stand out more.

See places like this:

http://github.com/ActiveState/cli/blob/90c8423f510bd7c73c475fbe8b21c17ff08b6ce7/cmd/state-installer/cmd.go#L356-L356

The command it executed already handled the UI for error reporting. So any errors that come from running this command should merely forward the exit code, but not provide any additional UI.

An example of such logic can be found here:

http://github.com/ActiveState/cli/blob/90c8423f510bd7c73c475fbe8b21c17ff08b6ce7/internal/subshell/sscommon/sscommon.go#L191-L191

mitchell-as:
Error tips can be disabled using an environment variable, and the installer sets this environment variable when invoking the state tool (i.e. activating an environment upon install or running a command).

Also refactored and centralized silent error and exit code handling.

internal/subshell/sscommon/sscommon.go
```diff
@@ -188,7 +175,7 @@ func runDirect(env []string, name string, args ...string) error {
 	err := runCmd.Run()
 	// silence exit code errors
 	if eerr, ok := err.(*exec.ExitError); ok {
-		return silentExitCodeError{eerr}
+		return errs.WrapExitCode(eerr, eerr.ExitCode())
```
Naatan, line 178: This is changing subshell behaviour to now no longer silence the actual error. That's going to have implications on many things, not just the installer. I don't feel comfortable making this change at this stage.

cmd/state/errors.go
```diff
@@ -65,9 +66,11 @@ func (o *OutputError) MarshalOutput(f output.Format) interface{} {
 	errorTips = append(errorTips, locale.Tl("err_help_forum", "[NOTICE]Ask For Help →[/RESET] [ACTIONABLE]{{.V0}}[/RESET]", constants.ForumsURL))
 
 	// Print tips
-	outLines = append(outLines, output.Heading(locale.Tl("err_more_help", "Need More Help?")).String())
-	for _, tip := range errorTips {
-		outLines = append(outLines, fmt.Sprintf(" [DISABLED]•[/RESET] %s", trimError(tip)))
+	if _, disableTips := os.LookupEnv(constants.DisableErrorTipsEnvVarName); !disableTips {
```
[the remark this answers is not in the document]
Naatan: I'd rather we still check for "true", as that's what we do everywhere else.

Naatan (changes requested):

internal/subshell/sscommon/sscommon.go
mitchell-as: Thanks for catching this. I was under the impression ExitCode was silent, but it was not.

cmd/state-installer/cmd.go
```diff
@@ -355,7 +355,7 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		cmd, args := exeutils.DecodeCmd(params.command)
 		if _, _, err := exeutils.ExecuteAndPipeStd(cmd, args, envSlice(binPath)); err != nil {
 			an.EventWithLabel(AnalyticsFunnelCat, "forward-command-err", err.Error())
-			return errs.Wrap(err, "Running provided command failed, error returned: %s", errs.JoinMessage(err))
+			return errs.Silence(err)
```
Naatan, line 358: This is dropping the error message it was adding, which is valuable as Go errors don't have stack traces.

I won't comment on the other instances, but it goes for all of them.

cmd/state/autoupdate.go
```diff
@@ -92,7 +92,7 @@ func autoUpdate(args []string, cfg *config.Instance, out output.Outputer) (bool,
 
 	code, err := relaunch(args)
 	if err != nil {
-		return true, &forwardExitError{code}
+		return true, errs.WrapExitCode(err, code, false)
```
Naatan, line 95:
```suggestion
		return true, errs.WrapExitCode(err, code, true)
```

It looks like `forwardExitError` is silenced.

Also this change means we can drop `forwardExitError`.

internal/errs/exitcode.go
```diff
@@ -7,10 +7,11 @@ type ExitCodeable interface {
 type ExitCode struct {
 	code       int
 	wrappedErr error
+	silent     bool
 }
 
-func WrapExitCode(err error, code int) error {
-	return &ExitCode{code, err}
+func WrapExitCode(err error, code int, silent bool) error {
```
Naatan, line 13: This is overloading the meaning of WrapExitCode. Its responsibility is wrapping exit codes, full stop. I your use-cases for this should instead also invoke `Silence()`.

internal/errs/silencederror.go
```diff
@@ -0,0 +1,24 @@
+package errs
+
+import (
+	"errors"
+)
+
+type silencedError struct {
+	error
```
Naatan, line 8: I know you just copied this, but please make this a named property instead, and satisfy the error interface manually.

Without this Unwrap would effectively return the same error, which could result in funky behaviour.

internal/errs/silencederror.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+func Silence(err error) silencedError {
+	return silencedError{err}
```
Naatan, line 12:
```suggestion
func Silence(err error) *silencedError {
	return &silencedError{err}
```

 Your methods are using pointers.

Naatan (changes requested):

internal/errs/silencederror.go
```diff
@@ -5,14 +5,16 @@ import (
...
 }
 
-func Silence(err error) silencedError {
-	return silencedError{err}
+func Silence(err error) *silencedError {
+	return &silencedError{err}
 }
 
-func (s *silencedError) Unwrap() error { return s.error }
+func (s *silencedError) Error() string { return s.err.Error() }
```
Naatan, line 15:
```suggestion
func (s *silencedError) Error() string { return "silenced error" }
```

This will tell us that it was funnelled through silencedError when we receive the error. Otherwise we just get the same error twice.

Naatan (changes requested):

internal/errs/silencederror.go
mitchell-as: Good call

Naatan (approved):

PR #1858, opened 2022-05-12
Added some unit tests for state-svc.

mitchell-as:
TestSingleSvc currently fails.

### DIFF
--- a/cmd/state-svc/test/integration/svc_int_test.go
+++ b/cmd/state-svc/test/integration/svc_int_test.go
@@ -1,0 +1,124 @@
+package integration
+
+import (
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/exeutils"
+	"github.com/ActiveState/cli/internal/testhelpers/e2e"
+	"github.com/ActiveState/cli/internal/testhelpers/tagsuite"
+	"github.com/shirou/gopsutil/process"
+	"github.com/stretchr/testify/suite"
+)
+
+type SvcIntegrationTestSuite struct {
+	tagsuite.Suite
+}
+
+func (suite *SvcIntegrationTestSuite) TestStartStop() {
+	suite.OnlyRunForTags(tagsuite.Service)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("start"))
+	cp.Expect("Starting")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Checking")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("stop"))
+	cp.Expect("Stopping")
+	cp.ExpectExitCode(0)
+}
+
+func (suite *SvcIntegrationTestSuite) TestSignals() {
+	suite.OnlyRunForTags(tagsuite.Service)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	// SIGINT (^C)
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("foreground"))
+	cp.Expect("Starting")
+	cp.Signal(syscall.SIGINT)
+	cp.Expect("caught a signal: interrupt")
+	cp.ExpectNotExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+
+	// SIGTERM
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("foreground"))
+	cp.Expect("Starting")
+	cp.Signal(syscall.SIGTERM)
+	suite.NotContains(cp.TrimmedSnapshot(), "caught a signal")
+	cp.ExpectNotExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+}
+
+func (suite *SvcIntegrationTestSuite) TestSingleSvc() {
+	suite.OnlyRunForTags(tagsuite.Service)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	oldCount := suite.GetNumStateSvcProcesses() // may be non-zero due to non-test state-svc processes
+	for i := 1; i <= 10; i++ {
+		go func() {
+			ts.SpawnCmdWithOpts(ts.Exe, e2e.WithArgs("--version"))
+		}()
+	}
+	time.Sleep(1 * time.Second) // allow for some time to spawn the processes
+	suite.Equal(oldCount+1, suite.GetNumStateSvcProcesses())
+
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("stop"))
+	cp.ExpectExitCode(0)
+
+	suite.Equal(oldCount, suite.GetNumStateSvcProcesses())
+}
+
+func (suite *SvcIntegrationTestSuite) GetNumStateSvcProcesses() int {
+	procs, err := process.Processes()
+	suite.NoError(err)
+
+	count := 0
+	for _, p := range procs {
+		name, err := p.Name()
+		suite.NoError(err)
+
+		if svcName := constants.ServiceCommandName + exeutils.Extension; name == svcName {
+			count++
+		}
+	}
+
+	return count
+}
+
+func (suite *SvcIntegrationTestSuite) TestResolveRequestsBeforeStop() {
+	suite.OnlyRunForTags(tagsuite.Service)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("foreground"))
+	cp.Expect("Starting")
+
+	cp2 := ts.SpawnCmdWithOpts(ts.Exe, e2e.WithArgs("update"))
+	cp.Signal(syscall.SIGINT)
+	cp2.Expect("Updating")
+	cp2.Expect("Done")
+	cp2.ExpectExitCode(0)
+}
+
+func TestSvcIntegrationTestSuite(t *testing.T) {
+	suite.Run(t, new(SvcIntegrationTestSuite))
+}
--- a/cmd/state-installer/test/integration/installer_int_test.go
+++ b/cmd/state-installer/test/integration/installer_int_test.go
@@ -11,10 +11,9 @@ import (
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/installation"
-	"github.com/stretchr/testify/suite"
-
 	"github.com/ActiveState/cli/internal/testhelpers/e2e"
 	"github.com/ActiveState/cli/internal/testhelpers/tagsuite"
+	"github.com/stretchr/testify/suite"
 )
 
 type InstallerIntegrationTestSuite struct {
--- a/internal/installmgr/stop.go
+++ b/internal/installmgr/stop.go
@@ -80,10 +80,7 @@ func stopSvc(installPath string) error {
 			continue
 		}
 
-		svcName := constants.ServiceCommandName
-		if runtime.GOOS == "windows" {
-			svcName = svcName + ".exe"
-		}
+		svcName := constants.ServiceCommandName + exeutils.Extension
 		if n == svcName {
 			exe, err := p.Exe()
 			if err != nil {
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -6,8 +6,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/shirou/gopsutil/process"
-
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/fileutils"
@@ -15,6 +13,7 @@ import (
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/osutils/lockfile"
 	"github.com/ActiveState/cli/internal/rollbar"
+	"github.com/shirou/gopsutil/process"
 )
 
 type Configurable interface {
--- a/internal/testhelpers/e2e/session.go
+++ b/internal/testhelpers/e2e/session.go
@@ -13,6 +13,12 @@ import (
 	"time"
 
 	"github.com/ActiveState/cli/internal/condition"
+	"github.com/ActiveState/cli/internal/config"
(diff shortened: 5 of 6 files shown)


cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,119 @@
...
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("start"))
+	cp.Expect("Starting")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Checking")
```
daved, line 34: It might be worth using a RegEx check here to ensure that the port and/or log file are included.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+
+	// SIGTERM
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("foreground"))
+	cp.Expect("Starting")
+	cp.Signal(syscall.SIGTERM)
+	suite.NotContains(cp.TrimmedSnapshot(), "caught a signal")
```
daved, line 62: Does this work? There is a story to fix this, but I didn't think it landed yet.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,119 @@
...
+	cp.Expect("Starting")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Checking")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("stop"))
+	cp.Expect("Stopping")
+	cp.ExpectExitCode(0)
```
daved, line 39: Please consider checking if the sock file dir is empty before and after the service lifetime.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,119 @@
...
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+}
+
+func (suite *SvcIntegrationTestSuite) TestSingleSvc() {
+	suite.OnlyRunForTags(tagsuite.Service)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	oldCount := suite.GetNumStateSvcProcesses() // may be non-zero due to non-test state-svc processes
```
daved, line 75: I think we can make this more certain. Maybe add a function to svcctl that checks the sock file directory and sends stop to each socket before any tests run. To segregate the service instances running in tests versus the system running the tests, the temp directory envvar can be set (I think that should work) so that the sock file directory is set/known.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,119 @@
...
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("foreground"))
+	cp.Expect("Starting")
+
+	cp2 := ts.SpawnCmdWithOpts(ts.Exe, e2e.WithArgs("update"))
+	cp.Signal(syscall.SIGINT)
+	cp2.Expect("Updating")
+	cp2.ExpectExitCode(0)
```
daved, line 114: This feels like a race. Also, there is a good chance certain requests might get cut off. Our current client/server relationship is not robust enough to have a thoroughly graceful shutdown. Adding this now may be premature.

daved (changes requested):

cmd/state-svc/test/integration/svc_int_test.go
mitchell-as: Okay

cmd/state-svc/test/integration/svc_int_test.go
mitchell-as: Is that really necessary in the context of running unit tests on GitHub Actions? I only added this for running tests on my local machine where I already have state-svc processes running. On GitHub Actions I don't think there should really be any.

cmd/state-svc/test/integration/svc_int_test.go
mitchell-as: It is a race. I commented on this in the JIRA ticket, but I've removed the test for now.

cmd/state-svc/test/integration/svc_int_test.go
daved: Then I suggest splitting the non-ci code out in block using `e2e.RunningOnCI()` so that it's more clear that it is not meant to be helpful there.

mitchell-as: Hmm, I was wrong. It looks like there is a running, non-test state-svc process on CI.

mitchell-as: I think it's the installer subshell.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,114 @@
...
+	cp.ExpectExitCode(1)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("start"))
+	cp.Expect("Starting")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Checking")
+	cp.ExpectRe("Port:\\s+:\\d+")
+	cp.ExpectRe("Log:\\s+.+?\\.log")
```
Naatan, line 38: Would be good if we can also assert here that the log file given exists.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("start"))
+	cp.Expect("Starting")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
```
Naatan, line 35: Seeing as `start` isn't blocking we probably want some form of delay before calling status. Otherwise this is likely to fail.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/ActiveState/cli/internal/testhelpers/tagsuite"
+	"github.com/shirou/gopsutil/process"
+	"github.com/stretchr/testify/suite"
+)
+
+type SvcIntegrationTestSuite struct {
+	tagsuite.Suite
+}
+
+func (suite *SvcIntegrationTestSuite) TestStartStop() {
```
Naatan, line 22: This function seems to only be asserting that the commands give the expected output and exit codes. I think we should actually assert that a process is actually running on the given port, and that the log file is being populated. Likewise we should assert that it stopped at the end.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,114 @@
...
+	// SIGINT (^C)
+	cp := ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("foreground"))
+	cp.Expect("Starting")
+	cp.Signal(syscall.SIGINT)
+	cp.Expect("caught a signal: interrupt")
+	cp.ExpectNotExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
```
Naatan, line 65: We should also check that the socket file has been cleaned up for the SIGINT case.

cmd/state-svc/test/integration/svc_int_test.go
Naatan: We run tests with the state tool, so yeah we can expect one state-svc process to be running that's unrelated to our tests.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,114 @@
...
+	defer ts.Close()
+
+	oldCount := suite.GetNumStateSvcProcesses() // may be non-zero due to non-test state-svc processes
+	for i := 1; i <= 10; i++ {
+		go func() {
+			ts.SpawnCmdWithOpts(ts.Exe, e2e.WithArgs("--version"))
+		}()
+		time.Sleep(50 * time.Millisecond)
+	}
+	time.Sleep(1 * time.Second) // allow for some time to spawn the processes
```
Naatan, line 91: I think we'll want to give this some more time for the state-svc processes to complete their checks.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,114 @@
...
+	suite.Equal(oldCount+1, suite.GetNumStateSvcProcesses())
+}
+
+func (suite *SvcIntegrationTestSuite) GetNumStateSvcProcesses() int {
+	procs, err := process.Processes()
+	suite.NoError(err)
+
+	count := 0
+	for _, p := range procs {
+		name, err := p.Name()
```
Naatan, line 101: IIRC this can be an absolute path depending on how they were launched. Should `filepath.Base()` this.

Naatan (changes requested):

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -31,20 +34,37 @@ func (suite *SvcIntegrationTestSuite) TestStartStop() {
...
+	suite.True(len(fileutils.ReadFileUnsafe(logFile)) > 0)
+
+	cp.ExpectExitCode(0)
+
 	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("stop"))
 	cp.Expect("Stopping")
 	cp.ExpectExitCode(0)
+	time.Sleep(500 * time.Millisecond) // wait for service to stop
 
+	// Verify it deleted its socket file.
```
Naatan, line 67: Let's also verify here that the port is now available. That should help surface any sort of edge cases where listening on the port was failing due to other reasons (eg. firewall).

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,140 @@
...
+func (suite *SvcIntegrationTestSuite) TestSingleSvc() {
+	suite.OnlyRunForTags(tagsuite.Service)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	oldCount := suite.GetNumStateSvcProcesses() // may be non-zero due to non-test state-svc processes
+	for i := 1; i <= 10; i++ {
+		go func() {
+			ts.SpawnCmdWithOpts(ts.Exe, e2e.WithArgs("--version"))
+		}()
```
Naatan, line 113:
```suggestion
		go ts.SpawnCmdWithOpts(ts.Exe, e2e.WithArgs("--version"))
```

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,140 @@
...
+		time.Sleep(50 * time.Millisecond)
```
Naatan, line 114:
```suggestion
```

This shouldn't be necessary. If not having this breaks the test then they are correctly breaking and we need to address the underlying issue.

Naatan (changes requested):

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,148 @@
...
+	portRe := regexp.MustCompile("Port:\\s+:(\\d+)")
+	port := portRe.FindStringSubmatch(cp.TrimmedSnapshot())[1]
+	_, err := net.Listen("tcp", "localhost:"+port)
+	suite.Error(err)
+
+	// Verify it created and wrote to its reported log file.
+	cp.ExpectRe("Log:\\s+.+?\\.log")
+	logRe := regexp.MustCompile("Log:\\s+(.+?\\.log)")
+	logFile := logRe.FindStringSubmatch(cp.TrimmedSnapshot())[1]
+	suite.True(fileutils.FileExists(logFile))
```
Naatan, line 58: These kind of assertions show up as "Should be true" when they fail. Which gives zero insight into what actually failed. You then have to look at the stack and actual code to find out what went wrong.

Please at minimum at a description that explains what it was asserting here. Though I believe there is an actual `FileExists` assertion.

This goes for any true/false assertions.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -64,18 +65,26 @@ func (suite *SvcIntegrationTestSuite) TestStartStop() {
...
 	suite.False(fileutils.TargetExists(sockFile))
+	server, err := net.Listen("tcp", "localhost:"+port)
+	suite.NoError(err)
+	server.Close()
 }
 
 func (suite *SvcIntegrationTestSuite) TestSignals() {
+	if runtime.GOOS == "windows" {
+		suite.T().Skip("Windows does not support signal sending.")
+	}
```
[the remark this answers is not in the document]
mitchell-as: https://pkg.go.dev/os#Process.Signal

daved (approved):

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -94,7 +94,10 @@ func (suite *SvcIntegrationTestSuite) TestSignals() {
 	cp.ExpectExitCode(1)
 
 	sockFile := svcctl.NewIPCSockPathFromGlobals().String()
-	suite.False(fileutils.TargetExists(sockFile))
+	if runtime.GOOS != "linux" {
+		// TODO: this fails on only Linux for some reason, but it is not reproducable outside of CI
```
Naatan, line 98:
```suggestion
		// TODO: this fails on only Linux for some reason, but it is not reproducable outside of CI
		// https://activestatef.atlassian.net/browse/DX-964
```

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -0,0 +1,153 @@
...
+	cp.Expect("caught a signal: interrupt")
+	cp.ExpectNotExitCode(0)
+
+	cp = ts.SpawnCmdWithOpts(ts.SvcExe, e2e.WithArgs("status"))
+	cp.Expect("Service cannot be reached")
+	cp.ExpectExitCode(1)
+
+	sockFile := svcctl.NewIPCSockPathFromGlobals().String()
+	if runtime.GOOS != "linux" {
+		// TODO: this fails on only Linux for some reason, but it is not reproducable outside of CI
```
daved, line 98: Please file a story, or convey it in an existing story, then leave a link in the comment instead of a TODO.

daved (changes requested):

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -108,7 +111,10 @@ func (suite *SvcIntegrationTestSuite) TestSignals() {
 	cp.Expect("Service cannot be reached")
 	cp.ExpectExitCode(1)
 
-	suite.False(fileutils.TargetExists(sockFile))
+	if runtime.GOOS != "linux" {
+		// TODO: this fails on only Linux for some reason, but it is not reproducable outside of CI
```
Naatan, line 115:
```suggestion
		// TODO: this fails on only Linux for some reason, but it is not reproducable outside of CI
		// https://activestatef.atlassian.net/browse/DX-964
```

Naatan (approved):

Naatan (approved):

PR #1859, opened 2022-05-16
Fixed potential segfault in secrets get with no project file present.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-921" title="DX-921" target="_blank">BugDX-921</a>  CLI - SECRETS: Panic runtime error when executing `state secrets get project.FOO` not from the project folder.
  </td></table>
  <br />

### DIFF
--- a/internal/runners/secrets/secrets.go
+++ b/internal/runners/secrets/secrets.go
@@ -60,7 +60,7 @@ func NewList(client *secretsapi.Client, p listPrimeable) *List {
 
 // Run executes the list behavior.
 func (l *List) Run(params ListRunParams) error {
-	if l.proj == nil || l.proj.Source() == nil {
+	if l.proj == nil {
 		return locale.NewInputError("err_no_project")
 	}
 	if err := checkSecretsAccess(l.proj); err != nil {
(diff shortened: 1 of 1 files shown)


MDrakos (changes requested):
The `project` struct is just a `projectfile` with an `outputter`. If we are creating a `project` without a `projectfile` somewhere we should find out where and fix that... unless there is a legitimate case for it, but I can't think of one at the moment.

Naatan:
@MDrakos @mitchell-as We just pass the project in that was detected at runtime. It being nil is a legitimate use-case we should be addressing. The current fix seems appropriate to me.

MDrakos:
> @MDrakos @mitchell-as We just pass the project in that was detected at runtime. It being nil is a legitimate use-case we should be addressing. The current fix seems appropriate to me.

@Naatan If no project was detected then shouldn't that project struct be nil? Why is it not nil but the underlying projectfile is?

Naatan:
> > @MDrakos @mitchell-as We just pass the project in that was detected at runtime. It being nil is a legitimate use-case we should be addressing. The current fix seems appropriate to me.
> 
> @Naatan If no project was detected then shouldn't that project struct be nil? Why is it not nil but the underlying projectfile is?

You're totally right. Ignore everything I said.

MDrakos (approved):

PR #1860, opened 2022-05-16
For `state run` or `state exec`, only interpret "-v" as verbose if it's the first arg.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-928" title="DX-928" target="_blank">BugDX-928</a>  argument to scripts -v is interpreted by the state tool
  </td></table>
  <br />

### DIFF
--- a/cmd/state/internal/cmdtree/exec.go
+++ b/cmd/state/internal/cmdtree/exec.go
@@ -31,6 +31,12 @@ func newExecCommand(prime *primer.Values, args ...string) *captain.Command {
 			if len(args) > 0 && (args[0] == "-h" || args[0] == "--help") {
 				prime.Output().Print(ccmd.UsageText())
 				return nil
+			} else if len(args) > 0 && (args[0] == "-v" || args[0] == "--verbose") {
+				if len(args) > 1 {
+					args = args[1:]
+				} else {
+					args = []string{}
+				}
 			}
 
 			return runner.Run(params, args...)
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -139,8 +139,7 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 		defer cleanup()
 	}
 
-	args, hasVerboseArg := removeVerboseArg(args)
-	logging.CurrentHandler().SetVerbose(os.Getenv("VERBOSE") != "" || hasVerboseArg)
+	logging.CurrentHandler().SetVerbose(os.Getenv("VERBOSE") != "" || argsHaveVerbose(args))
 
 	logging.Debug("ConfigPath: %s", cfg.ConfigPath())
 	logging.Debug("CachePath: %s", storage.CachePath())
@@ -260,18 +259,24 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	return err
 }
 
-// Returns copy of args with any "-v" or "--verbose" flag removed, and true if such a flag was removed.
-func removeVerboseArg(args []string) ([]string, bool) {
-	isRunOrExec := len(args) > 1 && (args[1] == "run" || args[1] == "exec")
+func argsHaveVerbose(args []string) bool {
+	var commandFound, isRunOrExec bool
+	nextArg := 0
+
 	for i, arg := range args {
+		if i > 0 && !commandFound && !strings.HasPrefix(arg, "-") {
+			commandFound = true
+			isRunOrExec = arg == "run" || arg == "exec"
+			nextArg = i + 1
+		}
+
 		// Skip looking for verbose args after --, eg. for `state shim -- perl -v`
 		if arg == "--" {
-			return args, false
+			return false
 		}
-		if (arg == "--verbose" || arg == "-v") && (!isRunOrExec || i <= 2) {
-			ret := append(args[:i], args[i+1:]...)
-			return ret, true
+		if (arg == "--verbose" || arg == "-v") && (!isRunOrExec || i == nextArg) {
+			return true
 		}
 	}
-	return args, false
+	return false
 }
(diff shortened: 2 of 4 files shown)


daved:
@mitchell-as Please consider not setting the "Persistent" field to true and dropping as much related logic as possible. If this is still the best approach after that consideration, I'll approve this PR as soon as I read your reply.

daved:
Ah, sorry. I'm suggesting that, instead of manipulating the args themselves, we change the rules of the flags. This has to do with how our captain/cobra library works. The relevant "Persistent" field is set here: https://github.com/ActiveState/cli/blob/1c4b8865440a3f146c90fb13f920991f15fbcf17/cmd/state/internal/cmdtree/cmdtree.go#L225-L236

If "verbose" is no longer set, it should only be valid for the top-level command and no longer honored for subcommands. But I leave that to you to decide, but @Naatan may want to weigh in on this too.

Naatan:
This is purposefully not using the captain flag registration as that comes into play too late in our runtime. We would be missing out on verbose logs if we waited for the proper flag registration.

daved:
Understood. That makes me think changes need to happen both places. IOW: Detect early and only for top-level, don't manipulate the args, turn off Persist.

mitchell-as:
So can this PR proceed as-is? Or do I need to do more?

Naatan:
I don't understand why you want to change Persist? `-v` is still a global flag. That's not changing.

daved:
It makes more sense to me that we alter the nature of the verbose flag rather than post process the arguments.

cmd/state/main.go
```diff
@@ -260,15 +260,18 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	return err
 }
 
-func argsHaveVerbose(args []string) bool {
-	for _, arg := range args {
+// Returns copy of args with any "-v" or "--verbose" flag removed, and true if such a flag was removed.
+func removeVerboseArg(args []string) ([]string, bool) {
+	isRunOrExec := len(args) > 1 && (args[1] == "run" || args[1] == "exec")
```
daved, line 265: This will detect the run or exec commands only when they are the first argument. If another flag is set as the first argument, then the command will not be detected. This increases my thought that this issue should be resolved by altering the flag handling rather than post-processing the args.

daved (changes requested):

mitchell-as:
@daved ping
As noted, we have to look for verbose flags ASAP, so modifying the existing cmdtree handlers isn't early enough. Also, they'll find the verbose flag anyway for `state run` and `state exec` commands, no matter where they are. We only want to run in verbose mode if the flag comes immediately after the command, rather than anywhere in that command's argument list.

mitchell-as:
Okay, thanks for the clarification. I've turned off "Persist" for the verbose flag as you requested.

cmd/state/main.go
```diff
@@ -261,12 +260,21 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
...
+	var commandFound, isRunOrExec bool
+	nextArg := 0
+
+	for i, arg := range args {
+		if i > 0 && !commandFound && !strings.HasPrefix(arg, "-") {
+			commandFound = true
+			isRunOrExec = arg == "run" || arg == "exec"
+			nextArg = i + 1
+		}
+
```
daved, line 272: I've not reasoned this concern fully, but... I think this will trip up on `state --output json`. I wonder if the spf13/cobra or spf13/pflag packages have helper functions we can leverage.

daved (changes requested):

cmd/state/internal/cmdtree/exec.go
```diff
@@ -31,6 +31,12 @@ func newExecCommand(prime *primer.Values, args ...string) *captain.Command {
 			if len(args) > 0 && (args[0] == "-h" || args[0] == "--help") {
 				prime.Output().Print(ccmd.UsageText())
 				return nil
+			} else if len(args) > 0 && (args[0] == "-v" || args[0] == "--verbose") {
+				if len(args) > 1 {
+					args = args[1:]
+				} else {
+					args = []string{}
+				}
```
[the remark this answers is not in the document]
mitchell-as: Looks like it's necessary. If I don't skip over '-v', I get an error `The script '-v' is not defined in activestate.yaml.`

cmd/state/main.go
mitchell-as: You are right. I couldn't find any helper functions. If there were, there'd be a lot of repetition thanks to our `cmdtree.New()` structure. Perhaps this is one of those cases where "good enough" applies, rather than 100% correct.

mitchell-as:
I put the `Persist` field back in because commands like `state config set -v foo bar` no longer turn on verbose logging, and I would expect them to do so. Otherwise we require `state -v config set foo bar` or `state config -v set foo bar`, which seems a bit unreasonable to me.

cmd/state/internal/cmdtree/exec.go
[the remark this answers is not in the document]
daved: Let's go with it for now, with a bug I've filed: https://activestatef.atlassian.net/browse/DX-960

cmd/state/main.go
daved: I hear you, but I disagree that this is good enough. Unless we document that double dash flags cannot use spaces and must use the equals sign. But that breaks POSIX, I think (I've not looked it up, but it seems like it would).

daved: As for taking a direction, please go wit this for now and file a bug for us to resolve handling this issue. It may be related to the other bug I've filed, but I suggest we decide that during triage.

mitchell-as: Sorry, what I mean by "good enough" was my most recent commit: https://github.com/ActiveState/cli/pull/1860/commits/6f9b08ce0f041aa3b31f5e4957f84cf7c469a76b

It does not trip up on `--option value` flags.

mitchell-as: I've filed a bug: https://activestatef.atlassian.net/browse/DX-961

daved (approved):

PR #1861, opened 2022-05-17
Enforce minimum windows version

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-953" title="DX-953" target="_blank">StoryDX-953</a>  We error out with a friendly message if trying to run State Tool on older Windows Versions
  </td></table>
  <br />


cmd/state-installer/cmd.go
```diff
@@ -419,3 +426,16 @@ func resolveInstallPath(path string) (string, error) {
 		return installation.DefaultInstallPath()
 	}
 }
+
+func assertCompatibility() error {
+	if sysinfo.OS() == sysinfo.Windows {
+		osv, err := sysinfo.OSVersion()
+		if err != nil {
+			return locale.WrapError(err, "windows_compatibility_warning", "", err.Error())
+		} else if osv.Major < 10 || osv.Micro < 17134 {
```
mitchell-as, line 435: Is there a risk that Windows 11.0.0.0 will break this? (I just made up that micro, but you get the idea)

mitchell-as (changes requested):

mitchell-as (approved):

PR #1865, opened 2022-05-18
Escape '$' in arguments passed to executors for bash-like shells.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-914" title="DX-914" target="_blank">StoryDX-914</a>  Command-line arguments are not double-interpolated in activated projects
  </td></table>
  <br />
 

We double-quote arguments passed to executors before passing to the shell, which interprets special characters like '$'. Those should be escaped.

### DIFF
--- a/test/integration/run_int_test.go
+++ b/test/integration/run_int_test.go
@@ -296,7 +296,10 @@ func (suite *RunIntegrationTestSuite) TestRun_Perl_Variable() {
 
 	cp := ts.SpawnWithOpts(
 		e2e.WithArgs("activate"),
-		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+		e2e.AppendEnv(
+			"ACTIVESTATE_CLI_DISABLE_RUNTIME=false",
+			"PERL_VERSION=does_not_exist",
+		),
 	)
 	cp.Expect("Activated")
 	cp.WaitForInput(10 * time.Second)
--- a/test/integration/run_int_test.go
+++ b/test/integration/run_int_test.go
@@ -294,7 +294,10 @@ func (suite *RunIntegrationTestSuite) TestRun_Perl_Variable() {
     project: https://platform.activestate.com/ActiveState-CLI/Perl-5.32?commitID=a4762408-def6-41e4-b709-4cb548765005
 	`))
 
-	cp := ts.SpawnWithOpts(e2e.WithArgs("activate"))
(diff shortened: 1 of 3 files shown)


test/integration/run_int_test.go
```diff
@@ -281,6 +281,32 @@ func (suite *RunIntegrationTestSuite) TestRun_BadLanguage() {
 	cp.Expect("The language for this script is not supported", 5*time.Second)
 }
 
+func (suite *RunIntegrationTestSuite) TestRun_Perl_Variable() {
+	if runtime.GOOS == "windows" {
+		suite.T().Skip("Testing exec of Perl with variables is not applicable on Windows")
```
Naatan, line 286: Why not?

Naatan: Oh cause it's running CMD. IIRC we do allow running bash on Windows. But probably not worth it for this test.

mitchell-as: Windows doesn't interpret '$' specially, so this isn't an issue. I'll go ahead and enable it anyway.

test/integration/run_int_test.go
```diff
@@ -281,6 +281,28 @@ func (suite *RunIntegrationTestSuite) TestRun_BadLanguage() {
...
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	ts.PrepareActiveStateYAML(strings.TrimSpace(`
+    project: https://platform.activestate.com/ActiveState-CLI/Perl-5.32?commitID=a4762408-def6-41e4-b709-4cb548765005
+	`))
+
+	cp := ts.SpawnWithOpts(
+		e2e.WithArgs("activate"),
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
```
swagile1, line 295: One other option here: `e2e.AppendEnv("PERL_VERSION=not_a_version")`; that way if this test fails, we'll get a very specific error from the `cp.Expect(...)` call later.

test/integration/run_int_test.go
mitchell-as: Scratch that, Windows just doesn't like the command syntax and this really isn't applicable.

test/integration/run_int_test.go
mitchell-as: Okay.

daved (approved):

PR #1866, opened 2022-05-23
Added `state use reset` subcommand.

mitchell-as:
Also localized some `state use` messages.

### DIFF
--- a/internal/globaldefault/default.go
+++ b/internal/globaldefault/default.go
@@ -52,7 +52,7 @@ func Prepare(cfg DefaultConfigurer, subshell subshell.SubShell) error {
 	}
 
 	if err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true); err != nil {
-		return locale.WrapError(err, "err_globaldefault_update_env", "Could not write to user environment.")
+		return locale.WrapError(err, "err_globaldefault_update_env")
 	}
 
 	return nil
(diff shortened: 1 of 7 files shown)


test/integration/use_int_test.go
```diff
@@ -53,6 +55,36 @@ func (suite *UseIntegrationTestSuite) TestUse() {
...
+		python3Exe = python3Exe + ".bat"
+	}
+	suite.True(fileutils.TargetExists(python3Exe), python3Exe+" not found")
+
+	cp = ts.SpawnWithOpts(e2e.WithArgs("use", "reset"))
+	cp.Expect("Reset default project runtime")
+	cp.Expect("Note you may need to")
+	cp.ExpectExitCode(0)
+
+	suite.False(fileutils.TargetExists(python3Exe), python3Exe+" still exists")
```
MDrakos, line 81: Can this test also cover the environment cleanup?

internal/runners/use/reset.go
```diff
@@ -0,0 +1,48 @@
...
+		prime.Output(),
+		prime.Config(),
+		prime.Subshell(),
+	}
+}
+
+func (u *Reset) Run() error {
+	logging.Debug("Resetting default project runtime")
+
+	reset, err := globaldefault.ResetDefaultActivation(u.subshell, u.config)
```
MDrakos, line 31: Do we want a prompt here to confirm the reset? @Naatan @rawktron

MDrakos, line 31: Do we want a prompt here to confirm the reset? @Naatan @rawktron

internal/globaldefault/default.go
```diff
@@ -82,3 +82,31 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 
 	return nil
 }
+
+func ResetDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer) (bool, error) {
+	logging.Debug("Resetting globaldefault")
+
+	projectDir := cfg.GetString(constants.GlobalDefaultPrefname)
+	if projectDir == "" {
+		return false, nil // nothing to reset
```
MDrakos, line 91: We may want to log this. Since we're relying on the config there is a way for users to have a default installation and be unable to remove it because of `state clean config`

MDrakos, line 91: We may want to log this. Since we're relying on the config there is a way for users to have a default installation and be unable to remove it because of `state clean config`

MDrakos (changes requested):
Just some minor comments

test/integration/use_int_test.go
mitchell-as: Yep, at least on Linux and macOS where we have RC files that change the environment.

internal/globaldefault/default.go
mitchell-as: Fair enough.

internal/runners/use/reset.go
mitchell-as: I went ahead and added the prompt, like we have for similar reset-type of commands.

PR #1869, opened 2022-05-27
Added missing dimensions to legacy dimension map

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-965" title="DX-965" target="_blank">BugDX-965</a>  SequenceID not reported to Pixel
  </td></table>
  <br />

### DIFF
--- a/internal/analytics/client/sync/reporters/ga-state.go
+++ b/internal/analytics/client/sync/reporters/ga-state.go
@@ -1,6 +1,7 @@
 package reporters
 
 import (
+	"strconv"
 	"strings"
 
 	anaConsts "github.com/ActiveState/cli/internal/analytics/constants"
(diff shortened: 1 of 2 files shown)


internal/analytics/client/sync/reporters/ga-state.go
```diff
@@ -91,5 +92,7 @@ func legacyDimensionMap(d *dimensions.Values) map[string]string {
 		"17": p.PStr(d.InstanceID),
 		"18": p.PStr(d.Headless),
 		"19": p.PStr(d.CommitID),
+		"20": p.PStr(d.Command),
+		"21": strconv.Itoa(p.PInt(d.Sequence)),
```
MDrakos, line 96: Not sure how much it matters for GA reporting but isn't there a limit to the number of dimensions we can have? Not sure what the limit is either I just seem to recall this being brought up at some point.

MDrakos (approved):

internal/analytics/client/sync/reporters/ga-state.go
Naatan: We don't report to GA anymore :) We'll probably refactor this out in the future.

PR #1874, opened 2022-05-31
Default to AMD64 on macOS

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-911" title="DX-911" target="_blank">StoryDX-911</a>  We should default to amd64 when running under arm64 for mac
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/checkpoints.go
+++ b/pkg/platform/model/checkpoints.go
@@ -268,7 +268,7 @@ func platformArchToHostArch(arch, bits string) string {
 	return "unrecognized"
 }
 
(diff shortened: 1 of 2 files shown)


Naatan (changes requested):
I'd rather we implement this in a forward compatible way. Meaning that we only fall back to amd64 if no arm64 builds are available.

Of course there won't be any arm64 builds at this stage, but they will eventually be available. We shouldn't need a State Tool update to then support them.

Naatan (approved):

PR #1875, opened 2022-06-02
Cache artifact downloads.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-941" title="DX-941" target="_blank">StoryDX-941</a>  We cache artifact downloads
  </td></table>
  <br />


internal/constants/constants.go
```diff
@@ -161,6 +160,10 @@ const ArtifactCacheFileName = "artifact_cache.json"
 // ArtifactMetaDir is the directory in which we store meta information about artifacts
 const ArtifactMetaDir = "artifacts"
 
+// ArtifactCacheSizeEnvVarName is the maximum size in MB of the artifact cache.
+// The default is 500MB.
+const ArtifactCacheSizeEnvVarName = "ACTIVESTATE_ARTIFACT_CACHE_SIZE_MB"
```
Naatan, line 165: Can we move or duplicate this to the `state config` command?

pkg/platform/runtime/artifactcache/artifactcache.go
```diff
@@ -0,0 +1,172 @@
...
+	artifacts   map[artifact.ArtifactID]*cachedArtifact
+	mutex       sync.Mutex
+}
+
+const MB int64 = 1024 * 1024
+
+// New returns a new artifact cache in the State Tool's cache directory with the default maximum size of 500MB.
+func New() (*ArtifactCache, error) {
+	var maxSize int64 = 500 * MB
+	if sizeOverride, err := strconv.Atoi(os.Getenv(constants.ArtifactCacheSizeEnvVarName)); err != nil && sizeOverride > 0 {
```
Naatan, line 42: We should at least print a warning to the user if they passed an invalid value.

pkg/platform/runtime/artifactcache/artifactcache.go
```diff
@@ -0,0 +1,172 @@
...
+		logging.Debug("Storing artifact in cache would exceed cache size; finding least-recently accessed artifact")
+		var lastAccessed *cachedArtifact
+		for _, artifact := range cache.artifacts {
+			if lastAccessed == nil || artifact.LastAccessTime < lastAccessed.LastAccessTime {
+				lastAccessed = artifact
+			}
+		}
+
+		if lastAccessed == nil {
+			logging.Debug("Cannot avoid exceeding cache size; not storing artifact")
```
Naatan, line 124: Since this is a failsafe that should never happen we should log this to rollbar

pkg/platform/runtime/artifactcache/artifactcache.go
```diff
@@ -0,0 +1,172 @@
...
+		err := os.Remove(lastAccessed.ArchivePath)
+		if err != nil {
+			return errs.Wrap(err, "Unable to remove cached artifact '%s'", lastAccessed.ArchivePath)
+		}
+		delete(cache.artifacts, lastAccessed.Id)
+		cache.currentSize -= lastAccessed.Size
+	}
+
+	targetPath := filepath.Join(cache.dir, string(a))
+	err = fileutils.CopyFile(archivePath, targetPath)
```
Naatan, line 138: Let's time our CopyFile operations since they could add significant time to the total setup process. Then when we reach Save() we add up the times and log to rollbar if the total time is over 5 seconds.

pkg/platform/runtime/artifactcache/artifactcache_test.go
```diff
@@ -0,0 +1,114 @@
+package artifactcache
```
Naatan, line 1: Since this is reaching out to the filesystem we should mark this as an integration test (ie. rename it to *_int_test.go).

Alternatively we could segment out the parts where we do IO such that the unit tests can run without needing IO.

pkg/platform/runtime/artifactcache/artifactcache_test.go
```diff
@@ -0,0 +1,114 @@
...
+		"2C36A315-59ED-471B-8629-2663ECC95476",
+		"57E8EAF4-F7EE-4BEF-B437-D9F0A967BA52",
+		"E299F10C-7B5D-4B25-B821-90E30193A916",
+		"F95C0ECE-9F69-4998-B83F-CE530BACD468",
+		"CAC9708D-FAA6-4295-B640-B8AA41A8AABC",
+		"009D20C9-0E38-44E8-A095-7B6FEF01D7DA",
+	}
+
+	dir, err := os.MkdirTemp("", "")
+	assert.NoError(t, err)
```
Naatan, line 32: Suggest changing assert to require, so that the test stops when the assertion fails.

test/integration/activate_int_test.go
```diff
@@ -551,3 +551,30 @@ func (suite *ActivateIntegrationTestSuite) TestActivateBranchNonExistant() {
 
 	cp.Expect("has no branch")
 }
+
+func (suite *ActivateIntegrationTestSuite) TestActivateArtifactsCached() {
```
Naatan, line 555: Suggest also testing that the cache is being used. You can do this by running with VERBOSE=true and asserting that the right debug logs happen (try not to be too specific in the assertions as debug messages tend to change over time).

Naatan (changes requested):

internal/constants/constants.go
mitchell-as: As discussed in Slack, this will be in a separate ticket.

pkg/platform/runtime/artifactcache/artifactcache.go
mitchell-as: Leaving this for implementation with config option as it's rather involved with our event producer -> consumer framework.

internal/constants/constants.go
```diff
@@ -160,10 +160,6 @@ const ArtifactCacheFileName = "artifact_cache.json"
 // ArtifactMetaDir is the directory in which we store meta information about artifacts
 const ArtifactMetaDir = "artifacts"
 
-// ArtifactCacheSizeEnvVarName is the maximum size in MB of the artifact cache.
-// The default is 500MB.
-const ArtifactCacheSizeEnvVarName = "ACTIVESTATE_ARTIFACT_CACHE_SIZE_MB"
```
Naatan: Seeing as we deferred handling this via config I think it makes sense to keep this in.

pkg/platform/runtime/artifactcache/artifactcache.go
```diff
@@ -135,7 +137,11 @@ func (cache *ArtifactCache) Store(a artifact.ArtifactID, archivePath string) err
 	}
 
 	targetPath := filepath.Join(cache.dir, string(a))
+	startTime := time.Now()
 	err = fileutils.CopyFile(archivePath, targetPath)
+	endTime := time.Now()
+	cache.timeSpentCopying += endTime.Sub(startTime)
```
Naatan, line 143: Suggest using `time.Since(startTime)`.

pkg/platform/runtime/artifactcache/artifactcache.go
```diff
@@ -168,5 +174,11 @@ func (cache *ArtifactCache) Save() error {
 		return errs.Wrap(err, "Unable to write artifact cache's "+cache.infoJson)
 	}
 
+	if cache.timeSpentCopying > 5*time.Second {
+		multilog.Log(logging.Debug, rollbar.Error)("Spent %.1f seconds copying %.1fMB of artifacts from cache", float64(cache.timeSpentCopying)/float64(time.Second), float64(cache.sizeCopied)/float64(MB))
```
Naatan, line 178:
```suggestion
		multilog.Log(logging.Debug, rollbar.Error)("Spent %.1f seconds copying %.1fMB of artifacts to cache", cache.timeSpentCopying.Seconds(), float64(cache.sizeCopied)/float64(MB))
```

Naatan (changes requested):

PR #1882, opened 2022-06-03
Add executors

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-131" title="DX-131" target="_blank">StoryDX-131</a>  We have a lightweight executable specifically for shimming executors
  </td></table>
  <br />


cmd/state-svc/service.go
```diff
@@ -82,3 +92,32 @@ func (s *service) Wait() error {
...
+
+type resolveExec struct {
+	resolver *resolver.Resolver
+}
+
+func (r *resolveExec) ReportRuntimeUsage(ctx context.Context, pid, exec string) {
+	dims := &dimensions.Values{
+		Trigger:          p.StrP(target.TriggerExec.String()),
+		Headless:         p.StrP("true"),
+		CommitID:         new(string),
```
Naatan, line 104: Needs actual commit ID.

cmd/state-svc/service.go
```diff
@@ -82,3 +92,32 @@ func (s *service) Wait() error {
...
+		ProjectNameSpace: p.StrP(project.NewNamespace("", "", "").String()),
```
Naatan, line 105: Needs namespace.

internal/globaldefault/default.go
```diff
@@ -70,9 +71,16 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 		return locale.WrapError(err, "err_globaldefault_rtexes", "Could not retrieve runtime executables")
 	}
 
+	env, err := runtime.Env(false, false)
+	if err != nil {
+		return locale.WrapError(err, "err_globaldefault_rtenv", "Could not construct runtime environment variables")
+	}
+
+	sockPath := svcctl.NewIPCSockPathFromGlobals().String()
```
Naatan, line 79: Doesn't seem there's any value in passing this as an arg.

activestate.yaml
```diff
@@ -137,6 +137,33 @@ scripts:
...
+      zig build-exe --strip -O ReleaseFast "./cmd/${appName}/${appName}.zig"
+      mv "${appName}" "${BUILD_TARGET_PREFIX_DIR}"
+
+  - name: build-exec
+    if: eq .OS.Name "Windows"
+    description: Builds the State Executor application
+    language: bash
+    value: |
+      appName="state-exec"
+      zig build-exe --strip -static -O ReleaseFast -target "x86_64-windows.win10_rs4" "./cmd/${appName}/${appName}.zig"
```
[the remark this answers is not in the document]
daved: Yes. This is the same limitation we have regarding socket usage on Windows in Go.

internal/globaldefault/default.go
daved: "globaldefault" makes use of the more primitive packages like runtime. It seems awkward for the runtime package to have a dependency on the svcctl package when it can, instead, be structured here as a parameter. Also, it makes future refactoring or updating easier. For example, in the case that the NewIPCSockPathFromGlobals function ends up needing some argument.

scripts/benchmark
```diff
@@ -0,0 +1,7 @@
+#! /usr/bin/env bash
```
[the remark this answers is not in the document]
daved: I left this where I worked on it to see if you even wanted it in the PR at all. It has a dependency named "perf" that is specific to Linux. It surely can come in handy. Making it cross-platform is likely costly.

internal/ipc/server.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	msgWidth = 64
+	msgWidth = 512
```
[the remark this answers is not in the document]
daved: The messages coming in to the service via raw TCP are potentially larger now that they can contain multiple Windows file paths. I've now bumped this up to 640 to better ensure data is not dropped.

cmd/state-svc/service.go
```diff
@@ -82,3 +92,32 @@ func (s *service) Wait() error {
 	}
 	return nil
 }
+
+type resolveExec struct {
+	resolver *resolver.Resolver
+}
+
+func (r *resolveExec) ReportRuntimeUsage(ctx context.Context, pid, exec string) {
```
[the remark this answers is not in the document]
daved: I can see why it looks like corner cutting. The resolver has the method RuntimeUsage which is fairly generic in that it can accommodate both `state exec` and `state-exec` cases. The calling logic for `state exec` is in the relevant runner, but the case for `state-exec` doesn't have a nice home, so I placed it here to wait for your direction. In essence, this is a new scope (that correlates to a "runner" package) that isn't yet organized. Until there's more of a need, I am content with it being here.

cmd/state-exec/state-exec.zig
```diff
@@ -0,0 +1,112 @@
...
+
+    _ = argIt.skip();
+    const path = (argIt.next(a) orelse return Error.ArgMissingOne) catch return Error.ArgInvalidOne;
+    const runt = (argIt.next(a) orelse return Error.ArgMissingTwo) catch return Error.ArgInvalidTwo;
+
+    var pid: i32 = @truncate(i32, @bitCast(i64, Thread.getCurrentId()));
+
+    const exec = fs.selfExePathAlloc(a) catch return Error.InspectSelfPath;
+
+    const clientThread = Thread.spawn(.{}, sendMsgToServer, .{ stderr, a, path, pid, exec }) catch {
```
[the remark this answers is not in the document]
daved: I'll make a quick example to prove this in the morning, but, in theory, the heartbeat signal is being sent concurrently while the runtime is launched, so the user should not notice delay in their runtime usage start. The duration of that thread should be very short since it's using a very fast form of communication with the service, so there should be little to no delay in runtime duration.

.github/workflows/build.yml
```diff
@@ -358,6 +373,12 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
 
+      - # === Install Zig ===
```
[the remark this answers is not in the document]
daved: Oops, probably not. I was just throwing things in when I added that, just to get to something that was working. This has been removed.

cmd/state-svc/service.go
```diff
@@ -82,3 +92,32 @@ func (s *service) Wait() error {
...
+	dims := &dimensions.Values{
+		Trigger:          p.StrP(target.TriggerExec.String()),
+		Headless:         p.StrP("true"),
```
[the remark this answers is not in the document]
daved: This is expressly for the `state-exec` use case, so I think Headless will always be true. Right? Requiring the commit id and namespace will induce more work, that, I think, will necessitate the executors having that data printed into them when they are generated. Sorry if I'm misunderstanding. I'm not sure how else the commit id and namespace would be acquired otherwise, but I'm open to options.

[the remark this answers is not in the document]
daved: Oh, hmm, is that headless commit? I'm not even sure what I had it in mind as. If so, that data will need to be included in the executor templates as well.

cmd/state-exec/state-exec.zig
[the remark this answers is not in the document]
Naatan: This is assuming that our socket performs faster than the actual executable that's being ran underneath. Seems plausible but yes; please verify (maybe make a very lightweight hello world executable to test with).

cmd/state-svc/service.go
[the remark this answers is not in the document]
Naatan: I don't see the current organization of the svc that way. Resolver is anything that handles requests. Whether those requests come in through the socket or through http ought to be irrelevant to the resolver package.

So I would still advocate we put it in the resolver package.

scripts/benchmark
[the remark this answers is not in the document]
Naatan: I have no idea how to use this, but if you get use out of it I'm all for it. Just put it in the yaml though.

cmd/state-svc/service.go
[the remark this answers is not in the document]
Naatan: Headless has no bearing on executors. Headless just means what state your local project is in, and executors follow the state of your local project. It can be true or false.

cmd/state-exec/state-exec.zig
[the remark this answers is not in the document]
daved: On my machine...
Calling the thread that calls the service and not waiting for it takes .08ms.
Calling the thread that calls the service and waiting for it takes .10ms.
Calling the thread that calls the service and putting it to the side while python3 is called as a child process takes around 12.80 - 13.00ms.
Calling python3 directly takes around 12.06-12.28ms.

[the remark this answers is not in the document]
Naatan: If you were testing with Python3 then that invalidates the test. Python3 is just ONE potential executable that the executor could invoke. A proper test should use a very lightweight executable such that the overhead of the executable itself is as minimal as possible and we can properly assert that the threading approach doesn't cause any significant delays.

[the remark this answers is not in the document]
daved: I might be misunderstanding... The cost of the concurrent call  seems to be .08ms. The cost of the heartbeat messaging seems to be .02ms. So, .10ms total. It's probably safe to assume all runtimes will take longer than that. Where this helps us is if the service has some delay responding and that .10ms becomes 20ms... the user will see their output immediately, but the process won't complete until the end of 20ms. If we run this serially, then that 20ms will need to pass before the user sees any output from their runtime.

pkg/platform/runtime/setup/setup.go
```diff
@@ -163,15 +162,15 @@ func NewWithModel(target Targeter, msgHandler Events, model ModelProvider, an an
 }
 
 // Update installs the runtime locally (or updates it if it's already partially installed)
-func (s *Setup) Update() error {
+func (s *Setup) Update(projPath, sockPath string) error {
```
Naatan, line 165: We record the project path on the runtime target, eg.:

https://github.com/ActiveState/cli/blob/ceeec3c0bbc10747cb307a60f6ffc6e487d2753a/pkg/platform/runtime/target/target.go#L79

Though this is raising an issue, which you would have faced as well; not all runtimes have activestate.yaml's. Eg offline installations and deployments.

I think maybe instead we need to record the relevant runtime information to a meta file in the runtime directory, eg.

https://github.com/ActiveState/cli/blob/b930714d64c0b3cd3485a2d012867fb2d94f6512/pkg/platform/runtime/store/store.go#L26

Then that would always be installed regardless of activestate.yaml existing.

Sorry about this, I should have foreseen this.

Naatan (changes requested):

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -170,6 +177,49 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
 	return &graph.RuntimeUsageResponse{Received: true}, nil
 }
 
+// ReportRuntimeUsage is an alternate version of RuntimeUsage which meets the
+// needs of the ipc package.
+func (r *Resolver) ReportRuntimeUsage(ctx context.Context, pid, exec, projDir string) {
```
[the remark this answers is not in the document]
daved: Done, but awkward IMO. This brings back my initial feeling that there should be an adapter involved. I can explain further if you'd like.

daved:
@Naatan The two files with changes that I'm concerned about most are pkg/platform/runtime/setup/setup.go and internal/globaldefault/default.go. Please be extra skeptical/careful there.

Naatan:
@daved did you mean to tag for re-review?

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -179,45 +176,31 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
...
-	logging.Debug("Packing state-exec data for process %s running project %s from %s", pid, projDir, exec)
+func (r *Resolver) ReportRuntimeUsage(ctx context.Context, pid, exec, namespace, commit, headless string) error {
 	dims := &dimensions.Values{
 		Trigger:          p.StrP(target.TriggerExec.String()),
 		Headless:         &headless,
-		CommitID:         &commitID,
-		ProjectNameSpace: &nameSpace,
+		CommitID:         &commit,
+		ProjectNameSpace: &namespace,
 		InstanceID:       p.StrP(instanceid.ID()),
```
Naatan, line 185: We'll need the instanceID as well, otherwise it's reporting the svc instance ID.

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -179,45 +176,31 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
...
-	if err != nil {
-		multilog.Critical("Could not convert pid string to int in proxied runtime-usage report: %s", errs.JoinMessage(err))
-	} else {
-		headless = fmt.Sprintf("%t", proj.IsHeadless())
-		commitID = proj.CommitID()
-		nameSpace = proj.Namespace().String()
-	}
-
-	logging.Debug("Packing state-exec data for process %s running project %s from %s", pid, projDir, exec)
+func (r *Resolver) ReportRuntimeUsage(ctx context.Context, pid, exec, namespace, commit, headless string) error {
```
Naatan, line 179:
```suggestion
func (r *Resolver) RuntimeUsageFromExec(ctx context.Context, pid, exec, namespace, commit, headless string) error {
```

Since the trigger is hard coded it makes sense to make the function specific to that. This function itself doesn't care about IPC, that just happens to be where you're using it from.

Does this still make sense as a separate function though? Why not call `RuntimeUsage()` directly?

Note you can refactor `RuntimeUsage()` to address both use-cases if that makes sense.

internal/assets/contents/executors/executor.bat
```diff
@@ -12,4 +12,4 @@ set {{$K}}={{$V}}
 {{- end}}
 {{- end}}
 
-"{{.stateExec}}" "{{.stateSock}}" "{{.target}}" "{{.projectDir}}" %*
+"{{.stateExec}}" "{{.stateSock}}" "{{.targetFile}}" "{{.nameSpace}}" "{{.commitID}}" "{{.headless}}" %*
```
Naatan, line 15: Note you could just pass the dimensions json blob here instead. That's one argument, and can be passed directly to `resolver.RuntimeUsage()`.

internal/runners/exec/exec.go
```diff
@@ -123,7 +124,8 @@ func (s *Exec) Run(params *Params, args ...string) error {
 		if err != nil {
 			return locale.WrapError(err, "err_initialize_runtime_event_handler")
 		}
-		if err := rt.Update(s.auth, eh); err != nil {
+		sockPath := svcctl.NewIPCSockPathFromGlobals().String()
+		if err := rt.Update(s.auth, eh, sockPath); err != nil {
```
Naatan, line 128: Is there a use-case for passing this as an argument rather than having the runtime call `svcctl.NewIPCSockPathFromGlobals().String()` itself?

Naatan (changes requested):

internal/runners/exec/exec.go
daved: This is to avoid having the svcctl dependency expressed so deep in the stack.

internal/assets/contents/executors/executor.bat
daved: I'll surely keep an open mind to that if more data needs to pass, but, for now, I'd like to keep it just space separated instead of having the executor marshal anything.

internal/runners/exec/exec.go
Naatan: That's a cost without a value in my book. We can always go that route if we have a use-case for it, it wouldn't require much refactoring. Now we're having to do this every time we init a runtime, which adds specialized knowledge. Take BE for example developing the offline installer, they'd need to understand the basics of our IPC socket just to use runtimes.

Please do this in the runtime itself.

internal/svcctl/comm.go
```diff
@@ -57,7 +63,7 @@ func (c *Comm) GetLogFileName(ctx context.Context) (string, error) {
 }
 
 type RuntimeUsageReporter interface {
-	ReportRuntimeUsage(ctx context.Context, pid, exec, namespace, commit, headless string) error
+	RuntimeUsage(ctx context.Context, pid int, exec, dimensionsJSON string) (*graph.RuntimeUsageResponse, error)
```
Naatan, line 66: Huh, it looks like this was always passed as an int:

https://github.com/ActiveState/cli/blob/22be2a51eb0bca4c8c80e4a103a0b72ee1dad738/cmd/state-svc/internal/server/generated/generated.go#L1009

How was this working before?

internal/instanceid/instanceid.go
```diff
@@ -6,17 +6,21 @@ import (
 	"github.com/google/uuid"
 )
 
+func Make() string {
```
Naatan, line 9:
```suggestion
func New() string {
```

This feels more in line with how we normally do stuff, though happy to be educated if this is part of some convention.

internal/instanceid/instanceid.go
```diff
@@ -6,17 +6,21 @@ import (
...
+
 var (
-	id string
-	mu sync.Mutex
+	appID string
+	mu    sync.Mutex
 )
 
-func ID() string {
+func AppID() string {
```
Naatan, line 18: Nit: this feels unnecessary

Naatan (changes requested):
See my comment here in addition to the new comments: 

https://github.com/ActiveState/cli/pull/1882#discussion_r930216021

internal/svcctl/comm.go
daved: I'm not following why it wouldn't work before.

internal/instanceid/instanceid.go
daved: New relates to the builtin new() func that returns a pointer to a type. Make relates to the make() builtin which returns an instance of a complex type.

internal/instanceid/instanceid.go
daved: Since I added a construction func, I wanted to ensure the meaning of this function was more clear since it is acting as an app global singleton.

internal/svcctl/comm.go
Naatan: Cause pid on `RuntimeUsage` was a string, but it was passing an int.

internal/instanceid/instanceid.go
Naatan: I'd say if anything this is more confusing though. It's not an "App ID", that sounds like it would be constant for any state tool invocation.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/1882#discussion_r930216021

Naatan (approved):

Naatan (approved):

PR #1886, opened 2022-06-06
Installer has update flag

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-970" title="DX-970" target="_blank">StoryDX-970</a>  We explicitly invoke the installer as either "install" or "update"
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -262,26 +262,26 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		logging.Debug("Cancelling out because State Tool is already installed")
 		out.Print(fmt.Sprintf("State Tool Package Manager is already installed at [NOTICE]%s[/RESET]. To reinstall use the [ACTIONABLE]--force[/RESET] flag.", installPath))
 		an.Event(AnalyticsFunnelCat, "already-installed")
-		return postInstallEvents(out, cfg, an, params, true)
+		return postInstallEvents(out, cfg, an, params)
 	}
 
 	// if sourcePath was provided we're already using the right installer, so proceed with installation
 	if params.sourcePath != "" {
-		if err := installOrUpdateFromLocalSource(out, cfg, an, params, params.isUpdate); err != nil {
+		if err := installOrUpdateFromLocalSource(out, cfg, an, params); err != nil {
 			return err
 		}
 		storeInstallSource(params.sourceInstaller)
-		return postInstallEvents(out, cfg, an, params, params.isUpdate)
+		return postInstallEvents(out, cfg, an, params)
 	}
 
 	return locale.NewError("err_install_source_path_not_provided", "Installer was called without an installation payload. Please make sure you're using the install.sh or install.ps1 scripts.")
 }
 
 // installOrUpdateFromLocalSource is invoked when we're performing an installation where the payload is already provided
-func installOrUpdateFromLocalSource(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher, params *Params, isUpdate bool) error {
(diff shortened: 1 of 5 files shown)


cmd/state-installer/cmd.go
```diff
@@ -237,39 +244,21 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
...
-		break // When ran with `--force` we always use the install UX
-	case params.sourcePath == "" && fileutils.FileExists(packagedStateExe):
-		// Facilitate older versions of state tool which do not invoke the installer with `--source-path`
-		logging.Debug("Using update flow as installer is alongside payload")
-		isUpdate = true
-		params.sourcePath = installerPath
-	case stateToolInstalled:
-		// This should trigger AFTER the check above where sourcePath is defined
-		logging.Debug("Using update flow as state tool is already installed")
-		isUpdate = true
```
Naatan, line 262: This is also dropping cases of `params.sourcePath = installerPath`, have you verified that this is no longer needed?

cmd/state-installer/cmd.go
```diff
@@ -278,11 +267,11 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 
 	// if sourcePath was provided we're already using the right installer, so proceed with installation
 	if params.sourcePath != "" {
-		if err := installOrUpdateFromLocalSource(out, cfg, an, params, isUpdate); err != nil {
+		if err := installOrUpdateFromLocalSource(out, cfg, an, params, params.isUpdate); err != nil {
 			return err
 		}
 		storeInstallSource(params.sourceInstaller)
-		return postInstallEvents(out, cfg, an, params, isUpdate)
+		return postInstallEvents(out, cfg, an, params, params.isUpdate)
```
Naatan, line 274: These are already passing `params`.

cmd/state-installer/cmd.go
```diff
@@ -442,3 +431,31 @@ func assertCompatibility() error {
...
+		isUpdate = true
+		params.sourcePath = installerPath
+	case stateToolInstalled:
+		// This should trigger AFTER the check above where sourcePath is defined
+		logging.Debug("Using update flow as state tool is already installed")
+		isUpdate = true
+	}
+
+	return isUpdate
+}
```
Naatan, line 461: If we've determined setting sourcePath is no longer need then we should be able to drop it here as well.

Naatan (changes requested):

cmd/state-installer/cmd.go
MDrakos: Good point. There is no case where we _don't_ want to have `params.sourcePath` set so I've updated the code to reflect that.

cmd/state-installer/cmd.go
MDrakos: 🤦‍♂️

Naatan (changes requested):
As discussed in stand yesterday; I think we can drop the source-path param except for not erroring out on it for backwards compatibility.

cmd/state-installer/cmd.go
```diff
@@ -244,16 +240,14 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		}
 	}
 
-	installerPath := filepath.Dir(osutils.Executable())
-	if params.sourcePath == "" {
-		params.sourcePath = installerPath
-	}
+	// We expect the installer payload to be in the same directory as the installer itself
+	params.payloadPath = filepath.Dir(osutils.Executable())
```
Naatan, line 244: `params` is for params the user can pass via flags. Since the user isn't passing this we shouldn't be using it. Just use a regular var that you pass around.

Naatan (changes requested):

Naatan (approved):

PR #1889, opened 2022-06-09
Allow installers to set up a runtime offline

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-958" title="DX-958" target="_blank">StoryDX-958</a>  Our runtime package can perform offline installations
  </td></table>
  <br />
 

There is a new target that can be passed to runtime.New(). The Update() method should install all artifacts in the directory specified by the target.

For example:
```
import (
	"github.com/ActiveState/cli/internal/analytics/client/blackhole"
	"github.com/ActiveState/cli/internal/runbits/buildlogfile"
	"github.com/ActiveState/cli/internal/testhelpers/outputhelper"
	"github.com/ActiveState/cli/pkg/platform/runtime"
	"github.com/ActiveState/cli/pkg/platform/runtime/setup/events"
	"github.com/ActiveState/cli/pkg/platform/runtime/target"
	"github.com/ActiveState/cli/pkg/platform/runtime/testhelper"
	"github.com/go-openapi/strfmt"
)

func createRuntimeOffline() {
	// ...
	owner := "testOwner"
	name := "testName"
	commitID := strfmt.UUID("000000000-0000-0000-0000-000000000000")
	artifactsDir := "path/to/directory/of/artifacts/to/install"
	offlineTarget := target.NewInstallationTarget(owner, name, commitID, dir, artifactsDir)

	analytics := blackhole.New()
	mockProgress := &testhelper.MockProgressOutput{}
	logfile, _ := buildlogfile.New(outputhelper.NewCatcher())
	eventHandler := events.NewRuntimeEventHandler(mockProgress, nil, logfile)

	rt, err := runtime.New(offlineTarget, analytics, nil)
	if err != nil && runtime.IsNeedsUpdateError(err) {
		err = rt.Update(nil, eventHandler)
		if err != nil {
			return error
		}
	}
	// the runtime has been installed offline
	// ...
}
```


pkg/platform/runtime/setup/setup.go
```diff
@@ -110,6 +111,8 @@ type Targeter interface {
 
 	// OnlyUseCache communicates that this target should only use cached runtime information (ie. don't check for updates)
 	OnlyUseCache() bool
+	// OnlyInstallFromDir communicates that this target should only install artifacts from the given directory (i.e. offline installer)
+	OnlyInstallFromDir() *string
```
Naatan, line 115: Can we name this `CacheDir()` instead? That's effectively what it is. The `OnlyInstallFrom` naming just throws my brain for a loop.

pkg/platform/runtime/runtime_int_test.go
```diff
@@ -0,0 +1,91 @@
...
+	assert.Equal(t, len(testArtifacts)*artifactsPerArtifact, mockProgress.ArtifactStartedCalled)
+	assert.Equal(t, 2*len(testArtifacts)*artifactsPerArtifact, mockProgress.ArtifactIncrementCalled) // start and stop each have one count
+	assert.Equal(t, len(testArtifacts)*artifactsPerArtifact, mockProgress.ArtifactCompletedCalled)
+	assert.Equal(t, 0, mockProgress.ArtifactFailureCalled)
+
+	for name, contents := range testFileContents {
+		filename := filepath.Join(dir, "tmp", name) // each file is in a "tmp" dir in the archive
+		assert.True(t, fileutils.FileExists(filename), "file '%s' was not extracted from its artifact", filename)
+		assert.Equal(t, contents, string(fileutils.ReadFileUnsafe(filename)))
+	}
```
Naatan, line 90: This feels out of scope for what you want to assert in this test. This test should be about ensuring that it can source an install from local files. Whether it installs those files correctly should be irrelevant as it should follow the same procedures as for an online install, the only difference should be where it gets the files.

pkg/platform/runtime/setup/implementations/installer/runtime.go
```diff
@@ -0,0 +1,32 @@
+package installer
```
Naatan, line 1: I thought we discussed and agreed that we do not want a separate setup implementation to facilitate offline installs. It should use the exact same mechanic as online builds, and we want a new layer that communicates the information that can be either offline or online.

This will require refactoring of our setup implementations to split out any sort of resourcing of bits to this layer. But this should not translate into us having online and offline setup implementations.

pkg/platform/runtime/setup/setup.go
```diff
@@ -160,6 +165,13 @@ func NewWithModel(target Targeter, msgHandler Events, model ModelProvider, an an
 
 // Update installs the runtime locally (or updates it if it's already partially installed)
 func (s *Setup) Update() error {
+	if s.target.OnlyInstallFromDir() != nil {
+		return s.installFromDir()
+	}
+	return s.updateFromRecipe()
```
Naatan, line 171: I feel like we were saying we want to avoid this type of logic. Basically I don't want any logic of the kind "if this is an offline install then do this, otherwise do that".

Instead sourcing of online/offline materials should be hidden behind a new interface which can be satisfied by an online and an offline implementation. Basically any sort of offline and online type logic should be centralized on this layer.

I though this is where we landed after we discussed it last time on zoom. Happy to discuss more if you don't agree with this approach or want more details.

Naatan (changes requested):
Didn't finish the review as the approach of this PR seems significantly different from what my understanding was of what we had discussed.

pkg/platform/runtime/runtime.go
```diff
@@ -57,7 +57,7 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
 	}
 
 	if !rt.store.MarkerIsValid(target.CommitUUID()) {
-		if target.OnlyUseCache() {
+		if target.OnlyUseCache() && target.OnlyInstallFromDir() == nil {
```
mitchell-as, line 60: Rename OnlyUseCache to be more accurate.

pkg/platform/runtime/setup/implementations/installer/runtime.go
mitchell-as: As discussed on Zoom, we'll leave the larger refactor for later. For now, the offline installer capability is more important.

pkg/platform/runtime/runtime.go
```diff
@@ -57,7 +57,7 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
 	}
 
 	if !rt.store.MarkerIsValid(target.CommitUUID()) {
-		if target.OnlyUseCache() {
+		if target.ReadOnly() && target.InstallFromDir() == nil {
```
Naatan, line 60: Why does we need to assert InstallFromDir here? The behaviour of "this runtime is not set up, please update it" should still work the same way for offline installs too.

pkg/platform/runtime/setup/implementations/installer/runtime.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"github.com/ActiveState/cli/pkg/platform/api/headchef/headchef_models"
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+	"github.com/ActiveState/cli/pkg/platform/runtime/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/store"
+)
+
+type Setup struct {
```
Naatan, line 10: Can we add some comments here explaining the nature of this implementation and the fact that it is temporary. Add a link to https://activestatef.atlassian.net/browse/DX-846

pkg/platform/runtime/setup/implementations/installer/runtime.go
Naatan, line 1: Can we name this "offline" or something. "installer" doesn't convey anything useful that differentiates it from the other implementations.

pkg/platform/runtime/setup/setup.go
```diff
@@ -160,6 +165,13 @@ func NewWithModel(target Targeter, msgHandler Events, model ModelProvider, an an
 
 // Update installs the runtime locally (or updates it if it's already partially installed)
 func (s *Setup) Update() error {
+	if s.target.InstallFromDir() != nil {
```
Naatan, line 168: Please add comment referencing what the distinction is, what the temporary nature of this is, and a link to https://activestatef.atlassian.net/browse/DX-846

pkg/platform/runtime/setup/setup.go
```diff
@@ -290,11 +281,7 @@ func (s *Setup) Update() error {
 		return errs.Wrap(err, "Could not save recipe file.")
 	}
 
-	if err := s.store.MarkInstallationComplete(s.target.CommitUUID()); err != nil {
-		return errs.Wrap(err, "Could not mark install as complete.")
-	}
-
-	return nil
+	return s.completeSetup(buildResult.OrderedArtifacts())
```
Naatan, line 284: Rather than move this logic into a new `completeSetup()` function couldn't this instead be placed in the `Update()` function after it calls the `if/else` for offline vs online installs. Reduces the number of layers to jump through when grokking code and more clearly conveys the desired behaviour of the code.

pkg/platform/runtime/setup/setup.go
```diff
@@ -341,16 +328,21 @@ func (s *Setup) setupArtifactSubmitFunction(a artifact.ArtifactDownload, buildRe
...
-		if err := s.setupArtifact(buildResult.BuildEngine, a.ArtifactID, a.UnsignedURI); err != nil {
-			if err != nil {
-				name := setup.ResolveArtifactName(a.ArtifactID)
-				errors <- locale.WrapError(err, "artifact_setup_failed", "", name, a.ArtifactID.String())
-			}
+		as, err := s.selectArtifactSetupImplementation(setup.BuildEngine(), a.ArtifactID)
+		if err != nil {
+			errors <- errs.Wrap(err, "Failed to select artifact setup implementation")
+			return
+		}
```
Naatan, line 335: Reading through the changes I'm not seeing why it was necessary to extract this behaviour from `setupArtifact`. Can we revert this?

pkg/platform/runtime/setup/setup.go
```diff
@@ -341,16 +328,21 @@ func (s *Setup) setupArtifactSubmitFunction(a artifact.ArtifactDownload, buildRe
...
+		err = s.setupArtifact(a.ArtifactID, a.UnsignedURI, as)
+		if err != nil {
+			name := setup.ResolveArtifactName(a.ArtifactID)
+			errors <- locale.WrapError(err, "artifact_setup_failed", "", name, a.ArtifactID.String())
 		}
 	}
 }
 
-func (s *Setup) installFromBuildResult(buildResult *model.BuildResult, downloads []artifact.ArtifactDownload, alreadyInstalled store.StoredArtifactMap, setup Setuper) error {
+func (s *Setup) installFromBuildResult(downloads []artifact.ArtifactDownload, alreadyInstalled store.StoredArtifactMap, setup Setuper) error {
```
Naatan, line 345: Nit: Should rename this if it's no longer "installing from build result".

pkg/platform/runtime/setup/setup.go
```diff
@@ -420,20 +412,17 @@ func (s *Setup) installFromBuildLog(buildResult *model.BuildResult, artifacts ar
...
 	targetDir := filepath.Join(s.store.InstallPath(), constants.LocalRuntimeTempDirectory)
 	if err := fileutils.MkdirUnlessExists(targetDir); err != nil {
 		return errs.Wrap(err, "Could not create temp runtime dir")
 	}
 
 	unarchiver := as.Unarchiver()
 	var archivePath string
-	if cachedPath, found := s.artifactCache.Get(a); found {
+	if s.target.InstallFromDir() != nil {
+		archivePath = unsignedURI
```
Naatan, line 424: So `unsignedURI` can be a path or an URI? That feels a little dirty and error prone. Please either refactor the function or use the arguments as intended. This might pass the type checking but it's not sane.

pkg/platform/runtime/setup/setup.go
```diff
@@ -573,6 +562,10 @@ func (s *Setup) selectSetupImplementation(buildEngine model.BuildEngine, artifac
 }
 
 func (s *Setup) selectArtifactSetupImplementation(buildEngine model.BuildEngine, a artifact.ArtifactID) (ArtifactSetuper, error) {
+	if s.target.InstallFromDir() != nil {
+		return alternative.NewArtifactSetup(a, s.store), nil // offline installer artifacts are in this format
```
Naatan, line 566:
```suggestion
        // Temporary workaround for offline installer artifacts
        // https://activestatef.atlassian.net/browse/DX-846
		return alternative.NewArtifactSetup(a, s.store), nil 
```

pkg/platform/runtime/setup/setup.go
```diff
@@ -625,3 +618,70 @@ func formatSolverError(serr *apimodel.SolverError) error {
...
+		}
+		wp.StopWait()
+	})
+
+	if err = <-aggregatedErr; err != nil {
+		return err
+	}
+
+	return s.completeSetup(artifactIDs)
+}
```
Naatan, line 687: This function is a bit difficult to grok. Please add comments and employ whitespace.

pkg/platform/runtime/setup/setup.go
```diff
@@ -625,3 +618,70 @@ func formatSolverError(serr *apimodel.SolverError) error {
...
+
+	setup := installer.NewSetup(s.store)
+
+	bgErrs, aggregatedErr := aggregateErrors()
+	mainthread.Run(func() {
+		defer close(bgErrs)
+		wp := workerpool.New(MaxConcurrency)
+		for i, a := range artifacts {
+			filename := a.Path()
+			filenameNoExt := filepath.Base(filename[0:strings.Index(filename, ".")])
```
Naatan, line 674: This is going to panic if we encounter a filename without an extension. Granted that shouldn't happen unless someone gives us a bad payload, but I'd rather we handle such situations somewhat gracefully.

pkg/platform/runtime/target/target.go
```diff
@@ -152,13 +162,60 @@ func (c *CustomTarget) Headless() bool {
...
+func (i *InstallationTarget) CommitUUID() strfmt.UUID {
+	return i.commitUUID
+}
+
+func (i *InstallationTarget) Dir() string {
+	return i.dir
+}
+
+func (i *InstallationTarget) Trigger() Trigger {
+	return triggerUnknown
```
Naatan, line 208: `triggerUnknown` is an error case, effectively. It's saying "we tried but we can't figure it out". I don't think that's warranted here. We should introduce a new trigger maybe, eg. "offline".

Naatan (changes requested):

pkg/platform/runtime/runtime.go
mitchell-as: Oops, this was leftover from a previous attempt.

pkg/platform/runtime/setup/setup.go
mitchell-as: Probably leftover from a previous attempt. I'll revert it.

pkg/platform/runtime/setup/setup.go
mitchell-as: Oh, I thought `:-1` was a valid slice. Oops.

pkg/platform/runtime/setup/setup.go
mitchell-as: Are you okay with `file://` in order to make it a URI instead of a path?

pkg/platform/runtime/setup/setup.go
```diff
@@ -625,3 +618,70 @@ func formatSolverError(serr *apimodel.SolverError) error {
...
+			artifactID := artifact.ArtifactID(filenameNoExt)
+			wp.Submit(s.setupArtifactSubmitFunction(artifact.ArtifactDownload{artifactID, filename, "", ""}, setup, bgErrs))
```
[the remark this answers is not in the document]
mitchell-as: If you're okay with `file://`, I'll use that and keep `ArtifactDownload`, since it's not much of a stretch to think that offline installer artifacts are downloads, but they've been "pre-downloaded" and their URI is just the local filesystem.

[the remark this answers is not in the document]
mitchell-as: Also, I'm hesitant to refactor with separate types or interfaces because that's going to complicate things and may be a potential rabbit hole.

pkg/platform/runtime/setup/setup.go
Naatan: I think that still has the same problem where the treatment of this variable is based on special hidden logic rather than logic that's conveyed through the function arguments themselves.

Can we create a new `setupArtifactFromURI()` function or something, which then calls `setupArtifact()` once it has downloaded the artifact and has a usable path?

Basically I don't want this argument to hold a value that depending on the way it is invoked is used in two completely distinctly different ways. That's hamfisting behaviour into the pre-existing function arguments.

pkg/platform/runtime/setup/setup.go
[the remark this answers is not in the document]
Naatan: I'm not on board with `file://`, but see my suggestion in the other comment.

> Also, I'm hesitant to refactor with separate types or interfaces because that's going to complicate things and may be a potential rabbit hole.

Absolutely. Please remain sensitive of rabbit holes, if anything I'm asking adds significant scope then we should probably find ways to iterate.

pkg/platform/runtime/setup/setup.go
```diff
@@ -421,7 +463,10 @@ func (s *Setup) setupArtifact(a artifact.ArtifactID, unsignedURI string, as Arti
 	unarchiver := as.Unarchiver()
 	var archivePath string
 	if s.target.InstallFromDir() != nil {
-		archivePath = unsignedURI
+		if !strings.HasPrefix(unsignedURI, "file://") {
+			return errs.New("Local artifacts must be in URI form (missing 'file://' prefix for %s)", unsignedURI)
+		}
+		archivePath = unsignedURI[len("file://"):]
```
Naatan, line 469: Please see https://github.com/ActiveState/cli/pull/1889#discussion_r897365543 and https://github.com/ActiveState/cli/pull/1889#discussion_r897366664

Naatan (changes requested):
Looking good, just one last iteration I think.

mitchell-as:
I was able to move all artifact installing to the `Update()` method, regardless of how those artifacts are sourced. There is another layer that sources the artifacts (either from the Platform or from the offline installer directory).

I expect another iteration or two on this, so don't feel bad about requesting more changes.

pkg/platform/runtime/setup/setup.go
```diff
@@ -212,19 +247,24 @@ func (s *Setup) Update() error {
 	return nil
 }
 
-func (s *Setup) updateFromRecipe() ([]artifact.ArtifactID, error) {
-	var noArtifacts []artifact.ArtifactID // in case of errors
+func (s *Setup) fetchArtifacts(installFunc artifactInstaller) error {
```
Naatan, line 250: Nit: It seems unnecessary that this function has an opinion on what the callback is used for (talking about the naming of the arg and type).

pkg/platform/runtime/setup/setup.go
```diff
@@ -160,6 +163,98 @@ func NewWithModel(target Targeter, msgHandler Events, model ModelProvider, an an
 
 // Update installs the runtime locally (or updates it if it's already partially installed)
 func (s *Setup) Update() error {
+	var artifacts []artifact.ArtifactID
+	mutex := sync.Mutex{}
```
Naatan, line 167:
```suggestion
	mutex := &sync.Mutex{}
```

Looking at the source of sync.Mutex it looks like its methods are expecting a pointer.

pkg/platform/runtime/setup/setup.go
```diff
@@ -160,6 +163,98 @@ func NewWithModel(target Targeter, msgHandler Events, model ModelProvider, an an
...
+
+	// Fetch and install each runtime artifact.
+	err := s.fetchArtifacts(func(a artifact.ArtifactID, archivePath string, as ArtifactSetuper) error {
```
Naatan, line 170: Can we move the callback into its own function? In terms of reading the diff it'd help if this was close to the original location. I can't easily tell how much of the underlying code here was changed, if at all.

pkg/platform/runtime/setup/setup.go
```diff
@@ -544,6 +554,36 @@ func (s *Setup) downloadArtifact(unsignedURI string, targetFile string, progress
 	return nil
 }
 
+//downloadArtifact downloads an artifact and returns the local path to that artifact's archive.
```
Naatan, line 557:
```suggestion
// downloadArtifact downloads an artifact and returns the local path to that artifact's archive.
```

pkg/platform/runtime/setup/setup.go
```diff
@@ -544,6 +554,36 @@ func (s *Setup) downloadArtifact(unsignedURI string, targetFile string, progress
...
+func (s *Setup) downloadArtifact(a artifact.ArtifactDownload, as ArtifactSetuper) (string, error) {
```
Naatan, line 558: This function seems to be doing too much. It says "download" in the name of the function but it's also performing setup. Please separate it out such that setup is performed with the result of this function, not inside this function. Keep it unixy.

pkg/platform/runtime/setup/setup.go
```diff
@@ -625,3 +665,48 @@ func formatSolverError(serr *apimodel.SolverError) error {
...
+			if extIndex == -1 {
+				extIndex = len(filename)
+			}
+			filenameNoExt := filepath.Base(filename[0:extIndex])
+			artifactID := artifact.ArtifactID(filenameNoExt)
+
+			// Submit the artifact for setup and install.
+			wp.Submit(func() {
+				as := alternative.NewArtifactSetup(artifactID, s.store) // offline installer artifacts are in this format
+				err = installFunc(artifactID, filename, as)
```
Naatan, line 701: I feel like installFunc should take s.store instead of an NewArtifactSetup instance. Because now your function that is responsible for fetching artifacts is also mixing in setup logic.

Naatan (changes requested):
This looks like a good direction! But I feel like we're suffering more from functions having mixed responsibilities that aren't always clearly conveyed through the code. I'd love to get this in a place where there is clear separation of concerns between the sourcing of artifacts and the setting up of artifacts. I think this can be done with minimal changes to the current state of things, but please let me know if that's not the case.

pkg/platform/runtime/setup/setup.go
mitchell-as: It looks like it's doing more, but it's just downloading. I've changed the `as` parameter to just be the file extension, because that's all that argument was needed for.

pkg/platform/runtime/setup/setup.go
mitchell-as: I'd prefer not because it's a closure (uses `mutex` and `artifacts`). I could make them belong to a `Setup` object, but that doesn't feel right. I'm open to suggestions.

pkg/platform/runtime/setup/setup.go
mitchell-as: Unfortunately I cannot figure out how to cleanly separate this. `installFunc` needs to know how to unpack the filename and potentially what BuildEngine produced it. I think it will get messy otherwise.

pkg/platform/runtime/setup/setup.go
mitchell-as: I could go with `callback FetchCallback`, but it's not as readable to me when used in-line. `callback(...)` vs. `installFunc(...)`. If you have a strong preference, I can do the former.

Naatan (approved):
Only comments I had left were easier to push as a commit rather than go back and forth in the PR. Those changes are purely organizational, far as I can tell no logic was impacted.

PR #1894, opened 2022-06-13
Installer is clickable on Windows

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-971" title="DX-971" target="_blank">StoryDX-971</a>  You can double click the state-installer in an install payload
  </td></table>
  <br />
 

This does work on macOS but is currently giving the update flow. I will test it again once https://github.com/ActiveState/cli/pull/1886 is merged as this should fix the issue. I don't currently have a Linux VM set up so I will do that and update this comment once I've tested there.


mitchell-as (approved):

Naatan (changes requested):

cmd/state-installer/cmd.go
```diff
@@ -285,6 +299,9 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 		return postInstallEvents(out, cfg, an, params, isUpdate)
 	}
 
+	if params.startedByExplorer {
+		return locale.NewError("err_install_explorer", "Installer must be started in the same directory as the install playload")
+	}
```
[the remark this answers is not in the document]
MDrakos: This error is no longer printed after the changes from #1886 go through

Naatan (changes requested):
I don't feel like this addressed my comments here: https://github.com/ActiveState/cli/pull/1894#discussion_r904026188

I would not expect to see any code that handles the concept of "explorer", only code that handles the concept of "no arguments were passed".

Whether we run by double clicking or by manually running via the CLI without any arguments the resulting behaviour would be the same.

cmd/state-installer/cmd.go
```diff
@@ -201,6 +204,13 @@ func main() {
 		if !errs.IsSilent(err) {
 			out.Error(err.Error())
 		}
+
+		// Installer was likely started via a double click so we keep the terminal window open
+		if noArgs() {
+			out.Print(locale.Tl("installer_pause", "Press return to close the console window..."))
```
Naatan, line 210:
```suggestion
			out.Print(locale.Tl("installer_pause", "Press ENTER to exit..."))
```

cmd/state-installer/cmd.go
```diff
@@ -439,7 +449,7 @@ func determineLegacyUpdate(stateToolInstalled bool, packagedStateExe, payloadPat
 	// Detect whether this is a fresh install or an update
 	var isUpdate bool
 	switch {
-	case (params.sourceInstaller == "install.sh" || params.sourceInstaller == "install.ps1") && fileutils.FileExists(packagedStateExe):
+	case (params.sourceInstaller == "install.sh" || params.sourceInstaller == "install.ps1" || noArgs()) && fileutils.FileExists(packagedStateExe):
```
Naatan, line 452: This is gonna conflict with your PR about handling updates with the update flag.

Naatan (changes requested):

cmd/state-installer/cmd.go
MDrakos: That PR is already in master and this branch is update to date with it.

Naatan: ooh this is in the legacy code. I don't think double clicking would ever reach this considering it'd only get here if you pass legacy flags?

Naatan (approved):
Looks good, though maybe double check my response to this comment: https://github.com/ActiveState/cli/pull/1894/files#r905497758

cmd/state-installer/cmd.go
MDrakos: It will still hit this as the update flag isn't set. Adding the `noArgs` call here just ensures we don't get the update UX.

Naatan (approved):
Nice catch!

PR #1898, opened 2022-06-15
Fix TestSingleSvc

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-996" title="DX-996" target="_blank">BugDX-996</a>  TestSingleSvc still fails sometimes
  </td></table>
  <br />


mitchell-as (approved):
Let's give it a try.

PR #1899, opened 2022-06-15
Added first Automation test

marikmAS:
Search test as per TP added

### DIFF
--- a/test/automation/search_automation_test.go
+++ b/test/automation/search_automation_test.go
@@ -1,0 +1,128 @@
+package automation
+
+import (
+	"fmt"
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/ActiveState/cli/internal/testhelpers/e2e"
+	"github.com/ActiveState/cli/internal/testhelpers/tagsuite"
+	"github.com/stretchr/testify/suite"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+type SearchAutomationTestSuite struct {
+	tagsuite.Suite
+}
+
+func (suite *SearchAutomationTestSuite) TestSearch_NoArg() {
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.Spawn("search")
+	cp.ExpectLongString("The following argument is required:")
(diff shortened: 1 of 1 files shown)


test/automation/search_automation_test.go
```diff
@@ -0,0 +1,128 @@
...
+	cp.ExpectExitCode(1)
+
+}
+
+func (suite *SearchAutomationTestSuite) TestSearch_OutProject() {
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.Spawn("search", "flask")
+	cp.ExpectLongString("Language must be provided by flag or by running this command within a project.")
```
Naatan, line 47: I would recommend shrinking this assertion to something like:

```
cp.ExpectLongString("Language must be provided by flag")
```

Two reasons for this:

1. The longer the assertion the more likely we'll run into text wrapping issues in the assertion
2. The longer the assertion the more likely it is to break if we make copy changes

test/automation/search_automation_test.go
```diff
@@ -0,0 +1,128 @@
...
+	cp.ExpectExitCode(1)
+
+}
+
+func (suite *SearchAutomationTestSuite) TestSearch_Flask() {
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	url := "https://platform.activestate.com/qamainorg/public?branch=main&commitID=32e543ee-b6ab-4f59-9b28-ad830ec6980e"
+	suite.Require().NoError(fileutils.WriteFile(filepath.Join(ts.Dirs.Work, "activestate.yaml"), []byte("project: "+url)))
```
Naatan, line 57: Was this necessary? IIRC you can use `state search` without a project.

Naatan (approved):
Looking good! Nothing alarming on my end, just flagged some things that you might want to have a closer look at.

test/automation/search_automation_test.go
marikmAS: Using search without `--language` should generate an error in some cases including this one.
Another use is verification which in all situations it provides required info.

PR #1901, opened 2022-06-15
Only care about whether we end up with too many services

Naatan:


mitchell-as (commented):
How will we know that a state service was properly spawned and persists? The whole point of this test is to assert that after 10 attempts to spawn the process, only one persists.

Naatan:
> How will we know that a state service was properly spawned and persists? The whole point of this test is to assert that after 10 attempts to spawn the process, only one persists.

`state --version` relies on the svc to give it version info, without the svc it would error out. So that in itself is an assertion that the service was started.

To me the value of this test is ensuring that we don't end up with rogue state-svc processes. There's value in asserting whether a service remains running and is servicing more requests but I never interpreted this test as being meant to assert that condition. Technically the count of svc processes doesn't assert that either, just hints at it.

mitchell-as (approved):

PR #1903, opened 2022-06-16
Separate installer payload in integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-869" title="DX-869" target="_blank">StoryDX-869</a>  Separate "installed state tool" from "install payload" in our integration tests
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/test/integration/installer_int_test.go
+++ b/cmd/state-installer/test/integration/installer_int_test.go
@@ -20,7 +20,6 @@ import (
 
 type InstallerIntegrationTestSuite struct {
 	tagsuite.Suite
-	payloadDir string
 }
 
 func (suite *InstallerIntegrationTestSuite) TestInstallFromLocalSource() {
@@ -35,7 +34,7 @@ func (suite *InstallerIntegrationTestSuite) TestInstallFromLocalSource() {
 	// Run installer with source-path flag (ie. install from this local path)
 	cp := ts.SpawnCmdWithOpts(
 		ts.InstallerExe,
-		e2e.WithArgs(target, "--source-path", suite.payloadDir),
+		e2e.WithArgs(target),
 		e2e.AppendEnv(constants.DisableUpdates+"=false"),
 	)
 
@@ -102,7 +101,7 @@ func (suite *InstallerIntegrationTestSuite) TestInstallIncompatible() {
 	// Run installer with source-path flag (ie. install from this local path)
 	cp := ts.SpawnCmdWithOpts(
 		ts.InstallerExe,
-		e2e.WithArgs(target, "--source-path", suite.payloadDir),
+		e2e.WithArgs(target),
 		e2e.AppendEnv(constants.DisableUpdates+"=false", sysinfo.VersionOverrideEnvVar+"=10.0.0"),
 	)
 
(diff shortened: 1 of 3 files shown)


internal/testhelpers/e2e/session.go
```diff
@@ -207,7 +207,10 @@ func new(t *testing.T, retainDirs, updatePath bool, extraEnv ...string) *Session
 	session.Exe = session.copyExeToBinDir(exe)
 	session.SvcExe = session.copyExeToBinDir(svcExe)
 	session.TrayExe = session.copyExeToBinDir(trayExe)
-	session.InstallerExe = session.CopyExeToDir(installExe, dirs.Base)
+	session.InstallerExe = session.CopyExeToDir(installExe, dirs.InstallerBin)
+	session.CopyExeToDir(exe, filepath.Join(dirs.InstallerPayload, installation.BinDirName))
+	session.CopyExeToDir(svcExe, filepath.Join(dirs.InstallerPayload, installation.BinDirName))
+	session.CopyExeToDir(trayExe, filepath.Join(dirs.InstallerPayload, installation.BinDirName))
```
Naatan, line 213: This creates extra IO for every single integration test. Considering we only need it in two tests can we move this logic into the test itself?

Naatan (changes requested):

cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -137,6 +141,12 @@ func (suite *InstallerIntegrationTestSuite) AssertConfig(ts *e2e.Session) {
 	}
 }
 
+func (s *InstallerIntegrationTestSuite) setupPayload(ts *e2e.Session) {
```
Naatan, line 144: Now `ts.Dirs.InstallerPayload` is part of the testsuite, but is only used here. I think we can move this method to the test suite (call it `SetupInstallerPayload()` or something), or alternatively move `ts.Dirs.InstallerPayload` outside of the test suite if it has no ownership of it.

Naatan (changes requested):

Naatan (changes requested):
Is this up to date with master? It's using `--source-path` which shouldn't be doing anything. And ts.InstallerExe is in a dir completely distinct from the payload, so this seems like it should fail?

cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -35,7 +34,7 @@ func (suite *InstallerIntegrationTestSuite) TestInstallFromLocalSource() {
 	// Run installer with source-path flag (ie. install from this local path)
 	cp := ts.SpawnCmdWithOpts(
 		ts.InstallerExe,
-		e2e.WithArgs(target, "--source-path", suite.payloadDir),
+		e2e.WithArgs(target),
```
Naatan, line 37: `target` is pointing at an installer in `workdir/installation`, but the payload you're setting up below uses the dir the InstallerExe is in which is `basedir`. These two do not seem to match. If the test still passes it's likely because it's using the files meant to run the state tool as oppose to the payload you've set up.

Naatan (changes requested):

cmd/state-installer/test/integration/installer_int_test.go
MDrakos: The dir for `InstallerExe` is a separate directory now at `base/installer`. That's the change from [here](https://github.com/ActiveState/cli/pull/1903/files#diff-3d51f7fa49a5898d7c83fcf6ff24388448fd98b3c41e0925b3c0ac90c9fe84f2L210)

PR #1904, opened 2022-06-17
Fix potential project panic

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-999" title="DX-999" target="_blank">BugDX-999</a>  Panic when trying to receive parent namespace
  </td></table>
  <br />

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -278,9 +278,6 @@ func (p *Project) Environments() string { return p.projectfile.Environments }
 
 // New creates a new Project struct
 func New(p *projectfile.Project, out output.Outputer) (*Project, error) {
-	if p == nil {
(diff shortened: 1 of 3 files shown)


pkg/project/project.go
```diff
@@ -278,6 +278,9 @@ func (p *Project) Environments() string { return p.projectfile.Environments }
 
 // New creates a new Project struct
 func New(p *projectfile.Project, out output.Outputer) (*Project, error) {
+	if p == nil {
+		return nil, locale.NewError("err_project_projectfile", "Cannot create new project without a projectfile")
+	}
```
MDrakos, line 283: I'm not certain that we should be policing the input of this function but this change will ensure that we do not have projects without projectfiles

PR #1906, opened 2022-06-22
Service instances that are not authoritative should die proactively

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1003" title="DX-1003" target="_blank">BugDX-1003</a>  Non-authoritative service instances stay alive.
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -197,7 +197,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 
 		cancel()
 		if err := p.Stop(); err != nil {
-			logging.Debug("Service stop failed: %v", err)
+			multilog.Critical("Service stop failed: %v", errs.JoinMessage(err))
 		}
 	})
 
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -192,6 +192,15 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 	signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
 	defer signal.Stop(sig)
 
+	p.RunIfNotAuthority(time.Second*3, svcctl.NewDefaultIPCClient(), func(err error) {
+		fmt.Fprintln(os.Stderr, err)
+
+		cancel()
+		if err := p.Stop(); err != nil {
+			logging.Debug("Service stop failed: %v", err)
+		}
+	})
+
 	if err := p.Wait(); err != nil {
 		return errs.Wrap(err, "Failure while waiting for server stop")
 	}
--- a/cmd/state-svc/service.go
+++ b/cmd/state-svc/service.go
@@ -104,5 +104,9 @@ func (s *service) RunIfNotAuthority(checkWait time.Duration, ipComm svcctl.IPCom
 }
 
 func portText(srv *server.Server) string {
+	if srv == nil || srv.Port() <= 0 {
+		return ""
+	}
+
 	return ":" + strconv.Itoa(srv.Port())
 }
--- a/cmd/state-svc/service.go
+++ b/cmd/state-svc/service.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/ActiveState/cli/cmd/state-svc/internal/server"
 	anaSvc "github.com/ActiveState/cli/internal/analytics/client/sync"
@@ -50,7 +51,7 @@ func (s *service) Start() error {
 
 	spath := svcctl.NewIPCSockPathFromGlobals()
 	reqHandlers := []ipc.RequestHandler{ // caller-defined handlers to expand ipc capabilities
-		svcctl.HTTPAddrHandler(":" + strconv.Itoa(s.server.Port())),
+		svcctl.HTTPAddrHandler(portText(s.server)),
 		svcctl.LogFileHandler(logging.FileName()),
 	}
 	s.ipcSrv = ipc.NewServer(s.ctx, spath, reqHandlers...)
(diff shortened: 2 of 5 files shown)


cmd/state-svc/main.go
```diff
@@ -192,6 +192,15 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 	signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
 	defer signal.Stop(sig)
 
+	p.RunIfNotAuthority(time.Second*3, svcctl.NewDefaultIPCClient(), func(err error) {
+		fmt.Fprintln(os.Stderr, err)
```
Naatan, line 196: Debugging code? Not sure I understand why this one is printed but the Stop error is logged.

cmd/state-svc/main.go
```diff
@@ -192,6 +192,15 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
...
+
+		cancel()
+		if err := p.Stop(); err != nil {
+			logging.Debug("Service stop failed: %v", err)
```
Naatan, line 200: This seems like something we'd want to tell rollbar about.

cmd/state-svc/main.go
```diff
@@ -192,6 +192,15 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 	signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
 	defer signal.Stop(sig)
 
+	p.RunIfNotAuthority(time.Second*3, svcctl.NewDefaultIPCClient(), func(err error) {
```
Naatan, line 195: Given this issue seems to stem from processors being overloaded; is 3 seconds enough? What's the harm in giving this like 30 seconds?

End of the day we want to make sure redundant service processes are shut down. I don't think we care as much that it happens within a certain limited timeframe, other than reproducibility of tests which we should be careful about having that influence our shipped logic.

I guess this begs the question; is this PR effectively only reducing the time it would take for the service to shut down? ie. even without these changes the service would inevitably shut itself down anyway?

cmd/state-svc/service.go
```diff
@@ -82,3 +83,26 @@ func (s *service) Wait() error {
...
+func (s *service) RunIfNotAuthority(checkWait time.Duration, ipComm svcctl.IPCommunicator, fn func(err error)) {
+	addr := portText(s.server)
+
+	go func() {
+		select {
+		case <-s.ctx.Done():
+			return
+		case <-time.After(checkWait):
+			checkedAddr, err := svcctl.LocateHTTP(ipComm)
+			if err == nil && (addr == "" || checkedAddr != addr) {
```
Naatan, line 96:
```suggestion
			if err == nil && checkedAddr != addr {
```

`addr == ""` is pointless noise considering `portText()` would at minimum return `:`.

Naatan (changes requested):

cmd/state-svc/main.go
daved: This error is printed because it isn't exactly expected by the user, but I didn't want to return non-zero. It is being printed just as the main function prints errors there: 4f636e9969b885b7241d139d9ffe38b9cc7c983da7882b66dcee6211b9ba88b0L82

cmd/state-svc/service.go
daved: This concern is a great catch, I hope you don't mind that I solved it slightly differently by permitting the portText function to return empty string in some reasonable conditions.

internal/ipc/server.go
```diff
@@ -137,6 +138,10 @@ func (ipc *Server) Shutdown() {
 }
 
 func (ipc *Server) Wait() error {
+	if ipc.errsc == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
Naatan: Why would them terminating themselves before they reach this mechanism matter? That's exactly what we want. The mechanism is meant to cover processes that didn't self-terminate.

cmd/state-svc/service.go
Naatan: Long as it's solving a real problem it sounds good to me 👍

cmd/state-svc/main.go
daved: Once the processes lock up the CPU, it seems that the wait time is mostly irrelevant. So, 3 seconds or 30 won't really improve or harm anything. This redundant mechanism isn't for that use case; It's pretty much just for unforeseen cases where service instances do exist while not established as the authoritative instance (i.e. they, somehow, make it past the sock guard).

internal/ipc/server.go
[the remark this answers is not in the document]
daved: Due to this being a redundant mechanism, we have to somewhat defeat the original mechanism in order to test this  one.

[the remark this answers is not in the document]
Naatan: Can we make this an explicit property on the struct or something? I prefer we avoid code that's specifically meant to satisfy some test case, but if we do it's better to be explicit about it so we don't accidentally break either the test or code being tested.

cmd/state-svc/main.go
Naatan: Ok, but why not return it? `runForeground()` returns errors, and its caller will print them. Printing them here doesn't seem desirable.

Naatan (changes requested):
See my comment responses on previous review items as well.

cmd/state-svc/main.go
daved: Returning an error would mean that we have to wait for it, and if the application dies before the check start or completes, I think we should be ok with not reaching the error. More moving parts can be added to make RunIfNotAuthority more mechanically correct, but I didn't think it was worthwhile.

internal/ipc/server.go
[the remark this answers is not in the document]
daved: This design isn't strongly test-specific. It does overlap with a condition we need for testing in this change, but it's fully useful on its own.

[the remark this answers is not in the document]
Naatan: Admittedly this is pretty nitpicky, but the part that concerns me is that we are overloading the error channel as the closer. Seems easy for someone to break the closing mechanic in the future cause they only write to the error channel when there's an actual error.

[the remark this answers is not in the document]
daved: It's not being overloaded. The wait function is meant to be dependent on the errors channel. The shutdown function is, basically, the "close" func, and it uses the done channel for signalling.

cmd/state-svc/main.go
daved: I think I worded my previous message poorly. RunIfNotAuthority is running concurrently. If we block on it, then the error returned from Wait may never be reached. Specifically, if the server has a problem starting up, we want to response to that immediately and not have to wait for a later scheduled authority check. We could wait, and collapse the scheduled authority check, but that would require a more complex mechanism. Printing the stop failure is the simplest solution that permits us to know of a Stop failure and still reports a Wait failure.

Naatan: I see. That makes sense, thanks for sticking with it.

Naatan (approved):

PR #1911, opened 2022-06-23
Do not attempt to setup environment for DisabledRuntime.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1008" title="DX-1008" target="_blank">BugDX-1008</a>  runtime panic "getenvbasedon"
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -92,6 +92,10 @@ func New(target setup.Targeter, an analytics.Dispatcher, svcm *model.SvcModel) (
 // Update updates the runtime by downloading all necessary artifacts from the Platform and installing them locally.
 // This function is usually called, after New() returned with a NeedsUpdateError
 func (r *Runtime) Update(auth *authentication.Auth, msgHandler *events.RuntimeEventHandler) error {
+	if r == DisabledRuntime {
+		return nil // nothing to do
+	}
+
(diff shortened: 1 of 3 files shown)


Naatan (changes requested):
This looks good, however I'd like to take the opportunity to make this code a little more robust and obvious.

I noticed when we "disable" the runtime we just pass around an essentially empty Runtime struct:

https://github.com/ActiveState/cli/blob/6aa8d11a7d27c8fca58c61950ebb1ae6fcb771c0/pkg/platform/runtime/runtime.go#L73

I think instead we should have something explicit (eg. a property `disabled: true`), which we then check for explicitly when doing eg. an update, and cancel out early.

This should hopefully prevent issues like this in the future. The issue being that the disabled behaviour is really implicit at the moment, and easy to miss while making changes.

mitchell-as:
The integration test actually uncovered a bug. `updateArtifacts()` needs to return a list of all *required* artifacts for the runtime, not just the artifacts that *were installed*.

Naatan (changes requested):
Nice work finding that other bug! But could you please separate this into its own PR? Reason being; when I check release notes or cherry pick PRs for a new release I won't be able to tell that this one contains a fix unrelated to it. I've filed a story for this here: https://activestatef.atlassian.net/browse/DX-1019

mitchell-as:
Okay, that makes sense. I've reverted the most recent commits and left the stuff we talked about on Zoom.

Naatan (approved):

PR #1913, opened 2022-06-23
Fixed bug updating a runtime where artifacts are already installed.

mitchell-as:


pkg/platform/runtime/setup/setup.go
```diff
@@ -272,24 +269,28 @@ func (s *Setup) updateExecutors(artifacts []artifact.ArtifactID) error {
 	return nil
 }
 
-func (s *Setup) fetchAndInstallArtifacts(installFunc artifactInstaller) error {
+// fetchAndInstallArtifacts needs to return all artifacts needed by the runtime, even if some or
+// all of them were already installed.
```
MDrakos, line 273: Nit:
```suggestion
// fetchAndInstallArtifacts returns all artifacts needed by the runtime, even if some or
// all of them were already installed.
```

pkg/platform/runtime/setup/setup.go
```diff
@@ -272,24 +269,28 @@ func (s *Setup) updateExecutors(artifacts []artifact.ArtifactID) error {
...
+func (s *Setup) fetchAndInstallArtifacts(installFunc artifactInstaller) ([]artifact.ArtifactID, error) {
 	if s.target.InstallFromDir() != nil {
 		return s.fetchAndInstallArtifactsFromDir(installFunc)
 	}
 	return s.fetchAndInstallArtifactsFromRecipe(installFunc)
 }
 
-func (s *Setup) fetchAndInstallArtifactsFromRecipe(installFunc artifactInstaller) error {
+func (s *Setup) fetchAndInstallArtifactsFromRecipe(installFunc artifactInstaller) ([]artifact.ArtifactID, error) {
+	var noArtifacts []artifact.ArtifactID
```
MDrakos, line 282: Couldn't you just return `nil` rather than having this slice? Same goes for other uses of the `noArtifacts` slice.

MDrakos (approved):
I'm not as familiar with this code but the changes make sense. Just a few nits.

pkg/platform/runtime/setup/setup.go
mitchell-as: I didn't know that. Thanks for the suggestion!

MDrakos (approved):

PR #1914, opened 2022-06-23
Initial stab at creating a Platform stub file for the State Tool.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-993" title="DX-993" target="_blank">StoryDX-993</a>  We generate platform stub files when we create a State Tool release
  </td></table>
  <br />
 

The goal here is to see if the stub file is generated by CI. Uploading the bits to the Platform is another matter.


.github/workflows/build.yml
```diff
@@ -419,3 +419,63 @@ jobs:
...
+        with:
+          repository: ActiveState/TheHomeRepot
+          path: TheHomeRepot
+
+      - # === Upload Stub ===
+        name: Upload Stub
+        if: false
+        shell: bash
+        run: |
+          apt-get install -y pipenv
```
Naatan, line 468: Note we should be dogfooding the platform, meaning pip is banned. Is this what LE/BE does?

Naatan: @icanhasmath Are those docs up to date?

icanhasmath: > @icanhasmath Are those docs up to date?

All the pipenv stuff is out of date.

icanhasmath: The adding ingredient documentation here should be mostly up to date:
https://handbook.activestate.build/teams/build-engineering/ingredients/add-new-ingredient/#import-ingredients-manually-using-be-tooling
It still references 
```
                 sortable_version:
                  - '0'
```
Which is not needed anymore.

PR #1918, opened 2022-06-29
Show local commits with `state show`.

mitchell-as:


internal/runners/show/show.go
```diff
@@ -348,10 +348,11 @@ func commitsData(owner, project, branchName string, commitID strfmt.UUID, localP
 		}
 		behind, err := model.CommitsBehind(latestCommitID, commitID)
 		if err != nil {
-			return "", locale.WrapError(err, "err_show_commits_behind", "Could not determine number of commits behind latest")
+			logging.Debug("Could not determine number of commits behind latest; assuming there are local commits")
+			return fmt.Sprintf("%s %s", localProject.CommitID(), locale.T("commit_display_local")), nil
```
MDrakos, line 352: Couldn't we verify that this commit id is not a part of the remote commits and then we _know_ it's local rather than assuming based on an error?

I realize that this would be more work than a simple bug fix so we could file a follow-up if we don't want to address this here.

MDrakos (changes requested):

internal/runners/show/show.go
mitchell-as: Okay, I found a way to compute commits ahead so now `state show` will annotate a local commit with "(X ahead of latest)".

MDrakos (commented):
~~Makes sense to me~~

~~Are there any integration tests related to this that we could run or is it possible to add one?~~

MDrakos (approved):
Never mind, I'm blind.

PR #1926, opened 2022-07-01
Moved integration tests into test/integration if they weren't already there

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-997" title="DX-997" target="_blank">StoryDX-997</a>  Our test runners properly differentiate between unit tests and integration tests
  </td></table>
  <br />
 

Go's test facilities use any file that ends in "_test.go". It doesn't matter if they end with "_int_test.go"; there is no way to filter by filename. The `-run` option allows regex to run tests filtered by name (e.g. "Test*Integration"). So we can do one of the following:

1. Filter by package (as we have been doing), and ensure all integration tests are in a package like tests/integration.
2. Filter by name regex, and ensure all integration tests follow a particular naming convention (e.g. have "Integration" in the name).

For the sake of discovery, we should keep the "_int_test.go" suffix for integration tests, and their location makes sense as they often test many working parts of the codebase and not just a single package. Requiring there to be a naming convention for integration tests is an additional burden that we probably don't want to bare, so I think we should keep going with option 1 and not do option to.


pkg/platform/runtime/artifactcache/testhelper.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+)
+
+type testArtifactCache struct {
+	cache *ArtifactCache
+}
+
+// TestNew is only meant to be called from tests. Use New() instead.
+func TestNew(dir string, maxSize int64) (*testArtifactCache, error) {
```
daved, line 15: Please consider following convention here, which would lead to this being named NewTestArtifactCache.

daved (approved):

daved (approved):

mitchell-as:
@Naatan would you please merge this? The only test failures are state-svc StartStop, which is to be addressed in a separate ticket I filed.

mitchell-as:
@Naatan ping. Please merge this. The integration test failures are in the state-svc and will be addressed in another ticket. I don't want to keep running the tests hoping for good RNG so that they'll all pass I can merge.

PR #1930, opened 2022-07-11
Report build failures as build failures instead of artifact download failures.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1036" title="DX-1036" target="_blank">BugDX-1036</a>  Certain build errors are not handled properly
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/artifact/download.go
+++ b/pkg/platform/runtime/artifact/download.go
@@ -45,13 +45,17 @@ func NewDownloadsFromCamelBuild(buildStatus *headchef_models.V1BuildStatusRespon
 			if strings.HasSuffix(a.URI.String(), ".tar.gz") || strings.HasSuffix(a.URI.String(), ".zip") {
 				return []ArtifactDownload{{ArtifactID: *a.ArtifactID, UnsignedURI: a.URI.String(), UnsignedLogURI: a.LogURI.String(), Checksum: a.Checksum}}, nil
 			}
-
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/platform/runtime/artifact/download.go
```diff
@@ -45,13 +45,17 @@ func NewDownloadsFromCamelBuild(buildStatus *headchef_models.V1BuildStatusRespon
...
 
-	if buildStatus.Type != nil && *buildStatus.Type == headchef_models.V1BuildStatusResponseTypeBuildStarted {
+	if buildStatusType := buildStatus.Type; buildStatusType != nil {
 		logging.Debug("buildStatus=%v", buildStatus)
-		return nil, CamelRuntimeBuilding
+		switch {
+		case *buildStatusType == headchef_models.V1BuildStatusResponseTypeBuildStarted:
+			return nil, CamelRuntimeBuilding
+		case *buildStatusType == headchef_models.V1BuildStatusResponseTypeBuildFailed:
+			return nil, errs.New("Build error: %+v", buildStatus)
```
Naatan, line 57: I think this should be localized, the error might be helpful for the user.

MDrakos (changes requested):
Make sense, just requesting changes for Nathan's comment.

pkg/platform/runtime/artifact/download.go
mitchell-as: I went ahead and localized this message, but there is a note about errors from headchef: "Note that these errors may not be suitable for presenting to users and should simply be logged for further investigation."

MDrakos (approved):

PR #1933, opened 2022-07-12
Localize another pflag error.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1028" title="DX-1028" target="_blank">BugDX-1028</a>  state [--language]: MUST ADDRESS: Error does not have localization: flag needs an argument: --language (wrapped) flag needs an argument: --language
  </td></table>
  <br />

### DIFF
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -711,6 +711,12 @@ func setupSensibleErrors(err error) error {
 		)
 	}
 
+	if strings.Contains(errMsg, "flag needs an argument: ") {
+		flag := strings.SplitN(errMsg, ": ", 2)[1]
+		return locale.NewInputError(
+			locale.Tl("command_flag_needs_argument", "Flag needs an argument: [NOTICE]{{.V0}}[/RESET]", flag))
+	}
+
 	return err
 }
 
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -697,6 +697,13 @@ func setupSensibleErrors(err error) error {
 		return locale.NewInputError("command_flag_invalid_value", "", flagText, msg)
 	}
 
+	// pflag error of the form "flag needs an argument: <flag>, called at: "
+	if strings.Contains(errMsg, "flag needs an argument: ") {
+		flag := strings.SplitN(errMsg, ": ", 2)[1]
+		return locale.NewInputError(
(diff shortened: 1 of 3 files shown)


internal/captain/command.go
```diff
@@ -711,6 +711,12 @@ func setupSensibleErrors(err error) error {
 		)
 	}
 
+	if strings.Contains(errMsg, "flag needs an argument: ") {
+		flag := strings.SplitN(errMsg, ": ", 2)[1]
+		return locale.NewInputError(
+			locale.Tl("command_flag_needs_argument", "Flag needs an argument: [NOTICE]{{.V0}}[/RESET]", flag))
+	}
+
```
daved, line 719: Please include a comment with an example of the string being checked for as seen for another similar case in lines 677-678.

daved (changes requested):

internal/captain/command.go
daved: Also, please defensively handle indexing of the result of SplitN.

mitchell-as: I am under the impression that it is impossible for the result of SplitN to panic because of the `string.Contains()` call above it. When you mean defensively, do you mean in case the error message changes over time?

daved: I see; The worst case here would be an empty string. Feel free to check for empty string or not.

internal/captain/command.go
```diff
@@ -697,6 +697,13 @@ func setupSensibleErrors(err error) error {
 		return locale.NewInputError("command_flag_invalid_value", "", flagText, msg)
 	}
 
+	// pflag error of the form "flag needs an argument: <flag>, called at: "
+	if strings.Contains(errMsg, "flag needs an argument: ") {
+		flag := strings.SplitN(errMsg, ": ", 2)[1]
+		return locale.NewInputError(
+			locale.Tl("command_flag_needs_argument", "Flag needs an argument: [NOTICE]{{.V0}}[/RESET]", flag))
```
daved, line 704: I'm fairly sure that the call to `Tl` is redundant.

daved (changes requested):

internal/captain/command.go
daved: Nevermind, that looks like an API issue. I'll look into it and file a story if needed.

daved (approved):

PR #1934, opened 2022-07-12
Show more helpful help and error message for --totp flag.

mitchell-as:


internal/locale/locales/en-us.yaml
```diff
@@ -118,7 +118,7 @@ flag_state_auth_username_description:
 flag_state_auth_password_description:
   other: The password to authenticate with
 flag_state_auth_totp_description:
-  other: The TOTP code generated via two-factor authentication
+  other: The TOTP code generated via two-factor authentication (used in conjunction with username and password flags)
```
MDrakos, line 121: Maybe we should add something about how it _has_ to be used with the other flags. Below is just a suggestion.
```suggestion
  other: The TOTP code generated via two-factor authentication (must be used in conjunction with username and password flags)
```

MDrakos (changes requested):

internal/runners/auth/auth.go
```diff
@@ -62,7 +62,7 @@ type SignupParams struct {
 func (a *Auth) Run(params *AuthParams) error {
 	if !a.Authenticated() {
 		if err := params.verify(); err != nil {
-			return locale.WrapInputError(err, "err_auth_params", "Invalid authentication params")
```
MDrakos, line 65: Sorry I missed this on the last review. Why is this no longer an input error?

mitchell-as: It is already an input error, as returned by params.verify(). It's an extra call that doesn't need to be made.

PR #1935, opened 2022-07-12
 Fixed display of beta label for `state invite`

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1039" title="DX-1039" target="_blank">BugDX-1039</a>  Command flags causing title / beta warning to get skipped
  </td></table>
  <br />
 

Delay org and role setting until executing the command, rather than doing so during flag processing.

If done during flag processing, the 'not authenticated' error is shown before the beta warning can be displayed.

### DIFF
--- a/internal/runners/invite/invite.go
+++ b/internal/runners/invite/invite.go
@@ -51,7 +51,7 @@ func (i *invite) Run(params *Params) error {
 	org := params.Org
 	if org.String() == "" {
 		if err := (&org).Set(i.project.Owner()); err != nil {
-			return err // do not wrap, as it overwrites the more helpful message displayed to the user (e.g. "You are not authenticated")
+			return locale.WrapInputError(err, "err_invite_org_current", "Could not use the owner of your current project.")
 		}
 	}
 
--- a/internal/runners/invite/invite.go
+++ b/internal/runners/invite/invite.go
@@ -18,8 +18,8 @@ import (
 const MaxParallelRequests = 10
 
 type Params struct {
-	Org       Org
-	Role      Role
+	Org       string
+	Role      string
 	EmailList string
 }
 
@@ -48,14 +48,28 @@ func (i *invite) Run(params *Params) error {
 		return locale.NewInputError("err_no_projectfile", "Must be in a project directory.")
 	}
 
-	org := params.Org
+	var org Org
+	if params.Org != "" {
+		err := org.Set(params.Org)
+		if err != nil {
+			return err
+		}
+	}
+
 	if org.String() == "" {
-		if err := (&org).Set(i.project.Owner()); err != nil {
+		if err := org.Set(i.project.Owner()); err != nil {
 			return locale.WrapInputError(err, "err_invite_org_current", "Could not use the owner of your current project.")
 		}
 	}
 
-	role := params.Role
+	var role Role
+	if params.Role != "" {
+		err := role.Set(params.Role)
+		if err != nil {
+			return err
+		}
+	}
+
 	if role == Unknown {
 		var err error
 		if role, err = i.promptForRole(); err != nil {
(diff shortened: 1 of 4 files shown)


MDrakos (changes requested):
I think this is pointing at a different issue. If we have to delay the setting of params then other commands may also be broken. Is there a way we can change how the header is printed so we can preserve the `Org` and `Role` types in the `Params`?

mitchell-as:
No, the processing of arguments happens way too early. I did check for other similar uses of this that would be problematic, but this command is the only one.

Naatan (changes requested):
This is too specific to the command being used. We'd have to work around this issue for every single command, and the workaround would mean we can't use typed flags anymore.

This will need to be addressed in the captain package rather than in the individual commands themselves.

Edit: Lol when I started responding no one had commented yet. And now I look like a douche.. :p

mitchell-as:
Okay, I made a change that may be controvertial. It uses `authentication.LegacyGet()`, but only because it was a bug that it wasn't already calling it like in the method below it.

If I wanted to replace `authentication.LegacyGet()` with a proper `authentication.Auth` pointer, I would have to modify the signatures of `model.FetchOrgByURLName()`, `model.FetchOrgMembers()`, `model.FetchOrgMember()`, and `model.InviteUserToOrg()`, and then propagate the changes to all callers of those 4 functions:

* secrets
* auth
* invite

It can be done, but it'll be more involved than just a simple bugfix.

What do you think @Naatan?

Naatan:
I think for the scope of the bug that feels appropriate. But you're right to feel uncomfortable, that's exactly why we named it that :p

internal/runners/invite/invite.go
```diff
@@ -51,7 +51,7 @@ func (i *invite) Run(params *Params) error {
 	org := params.Org
 	if org.String() == "" {
 		if err := (&org).Set(i.project.Owner()); err != nil {
-			return locale.WrapInputError(err, "err_invite_org_current", "Could not use the owner of your current project.")
+			return err // do not wrap, as it overwrites the more helpful message displayed to the user (e.g. "You are not authenticated")
```
Naatan, line 54: This is what `errs.Wrap()` is for. Non-localized errors are not shown to the user. But we should almost always wrap our errors as it implicitly tells us the stack trace.

Also it's not true that it overwrites anything, if there are two localized errors in the chain then both are shown. If that's not what you were seeing then either we have some type of bug or the error down the chain isn't localized.

Naatan (changes requested):

internal/runners/invite/invite.go
mitchell-as: Oh, this is actually a bug then: https://activestatef.atlassian.net/browse/DX-740

mitchell-as: Only one InputError is shown to the user, even if the error chain has multiple InputErrors. DX-740 mentions this. I've updated the PR to wrap errors, but to prefer any existing InputError for display to the user.

internal/runners/invite/invite.go
```diff
@@ -51,7 +51,10 @@ func (i *invite) Run(params *Params) error {
 	org := params.Org
 	if org.String() == "" {
 		if err := (&org).Set(i.project.Owner()); err != nil {
-			return locale.WrapInputError(err, "err_invite_org_current", "Could not use the owner of your current project.")
+			if locale.IsInputError(err) {
+				return locale.WrapError(err, "err_invite_org_current")
+			}
+			return locale.WrapInputError(err, "err_invite_org_current")
```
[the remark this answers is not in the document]
mitchell-as: Yes it is. I have removed it and made a comment in the automation test regarding the message currently displayed.

Naatan (approved):
Looks good! I've added an AC in DX-740 to revert the integration test changes once it is fixed.

MDrakos (approved):

PR #1938, opened 2022-07-14
Improve error message when uninstalling a non-existent package.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-887" title="DX-887" target="_blank">BugDX-887</a>  CLI - UNINSTALL: Very confusing error message if trying to uninstall non-existent package
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/packages.go
+++ b/internal/runners/packages/packages.go
@@ -163,6 +163,9 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
 	var commitID strfmt.UUID
 	commitID, err = model.CommitPackage(parentCommitID, operation, packageName, ns, packageVersion)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Naatan (commented):
While this is a good improvement, I think we also need to look into why `The following requirement was removed, but...` is not part of the error message printed to the user. I'm guessing that might be due to DX-740, but worth verifying (if it is - no need to fix / work around in this PR).

mitchell-as:
No, this is an error coming from the platform. If you look at the ticket in JIRA, you'll see the "Something Went Wrong" header, indicating a non-input error. I'm handling this specific platform error for this specific case of `state uninstall` to give an input error that the user is trying to uninstall something that does not exist.

MDrakos (approved):

PR #1940, opened 2022-07-14
Service autostarts

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1006" title="DX-1006" target="_blank">StoryDX-1006</a>  State-svc automatically starts with the OS
  </td></table>
  <br />


internal/osutils/autostart/autostart.go
```diff
@@ -1,50 +1,72 @@
...
+func New(name AppName, exec string, args []string, cfg Configurable) *App {
 	return &App{
-		Name: name,
-		Exec: exec,
-		cfg:  cfg,
+		Name:    name.String(),
+		Exec:    exec,
+		Args:    args,
+		cfg:     cfg,
+		options: data[name],
```
mitchell-as, line 40: It's possible this could fail by programmer error. Do we want to handle this properly? Do we have a way of handling this?

internal/osutils/autostart/autostart_linux.go
```diff
@@ -16,6 +16,25 @@ const (
 	autostartDir = ".config/autostart"
 )
 
+var data = map[AppName]options{
+	Tray: {
+		launchFileName: constants.TrayAppName,
```
mitchell-as, line 21: I like this use of struct initialization. It's clear to see what you're assigning the constants to.

internal/osutils/shortcut/shortcut_windows.go
```diff
@@ -96,6 +96,20 @@ func (s *Shortcut) setIcon(path string) error {
 	return nil
 }
 
+func (s *Shortcut) SetMinimized() error {
+	_, err := oleutil.PutProperty(s.dispatch, "WindowStyle", 7)
```
mitchell-as, line 100: Do we have to use this magic number? Is there a constant somewhere we can make use of? If not, please add a comment with what it means and/or the enum used by Windows/OLE (e.g. `// VK_MINIMIZE`, which I just made up)

internal/runners/prepare/prepare_linux.go
```diff
@@ -3,63 +3,48 @@ package prepare
...
-		return errs.Wrap(err, "Could not find icons directory")
+		r.reportError(locale.Tl(
+			"err_prepare_service_executable",
+			"Could not get service executable: {{.V0}}", err.Error(),
+		), err)
 	}
-	iconsPath := filepath.Join(iconsDir, constants.TrayIconFileName)
 
-	iconData, err := assets.ReadFileBytes(constants.TrayIconFileSource)
+	svcShortuct := autostart.New(autostart.Service, svcExec, []string{"start"}, r.cfg)
```
mitchell-as, line 41: Nit: spelling.

mitchell-as (changes requested):
Looks good with very minor requested changes.

internal/osutils/autostart/autostart.go
MDrakos: I've made it so that the only way you can get an `app` object is via the `New` function. I think this should cover it. Is there another way you can see?

internal/osutils/shortcut/shortcut_windows.go
MDrakos: It is essentially a magic number. Good point though I will try to make this more clear.

mitchell-as (dismissed):

activestate.windows.yaml
```diff
@@ -11,6 +11,8 @@ constants:
     value: state-update-dialog.exe
   - name: SYSTRAY_BUILDFLAGS
     value: -ldflags="-s -w -H=windowsgui"
+  - name: SVC_BUILDFLAGS
+    value: -ldflags="-s -w -H=windowsgui"
```
Naatan, line 15: Why was this necessary for this PR?

cmd/state-svc/main.go
```diff
@@ -99,6 +100,12 @@ func run(cfg *config.Instance) error {
 		return errs.Wrap(err, "Could not initialize outputer")
 	}
 
+	if mousetrap.StartedByExplorer() {
+		// Allow starting the installer via a double click
```
Naatan, line 104:
```suggestion
		// Allow starting the svc via a double click
```

internal/osutils/autostart/autostart.go
```diff
@@ -1,50 +1,78 @@
...
-	cfg  Configurable
+type AppName string
+
+func (a AppName) String() string {
+	return string(a)
+}
+
+const (
+	Tray    AppName = constants.TrayAppName
+	Service         = constants.SvcAppName
```
Naatan, line 19: We should move this out of autostart. This libs function should be all about facilitating autostarts, not about implementing specific versions of things that want to autostart. So svc and tray specific stuff should live in places that own that aspect, eg. cmd/state-tray and cmd/state-svc.

This config key would be another candidate for this.

internal/osutils/autostart/autostart.go
```diff
@@ -1,50 +1,78 @@
 package autostart
 
 import (
+	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/locale"
 )
 
 const ConfigKeyDisabled = "SystrayAutoStartDisabled"
```
Naatan, line 9:
```suggestion
const ConfigKeyDisabled = "systray.autostarted.disabled"
```
We should format these consistently.

Naatan: Also this config key seems too generic now. It says systray but effectively would also be used for the svc?

internal/osutils/autostart/autostart_mac.go
```diff
@@ -6,16 +6,27 @@ package autostart
...
+const launchFileSource = "com.activestate.platform.state.plist.tpl"
 
-func (a *App) enable() error {
+var data = map[AppName]options{
+	Tray: {
+		launchFileName: "com.activestate.platform.state-tray.plist",
+	},
+	Service: {
+		launchFileName: "com.activestate.platform.state-svc.plist",
+	},
```
Naatan, line 26: This seems like it can be calculated from app.Name or app.Exec. That way autostart doesn't need any opinionated boilerplate.

internal/osutils/autostart/autostart_mac.go
```diff
@@ -30,18 +41,26 @@ func (a *App) enable() error {
...
+
+	content, err := strutils.ParseTemplate(
+		string(asset),
+		map[string]interface{}{"Exec": a.Exec, "Args": strings.Join(a.Args, " ")})
+	if err != nil {
+		return errs.Wrap(err, "Could not parse %s", a.options.launchFileName)
+	}
+
+	err = fileutils.WriteFile(path, []byte(content))
 	if err != nil {
```
Naatan, line 57:
```suggestion
	if err = fileutils.WriteFile(path, []byte(content)); err != nil {
```

Try to avoid reusing err variables. It can give issues asserting nil values if the types satisfying the error interface don't match.

internal/osutils/autostart/autostart_mac.go
```diff
@@ -6,16 +6,27 @@ package autostart
...
 	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/ActiveState/cli/internal/strutils"
 	"github.com/mitchellh/go-homedir"
 )
 
-const launchFileMacOS = "com.activestate.platform.state-tray.plist"
+const launchFileSource = "com.activestate.platform.state.plist.tpl"
 
-func (a *App) enable() error {
+var data = map[AppName]options{
```
Naatan, line 20: Please name it something less generic.

internal/runners/prepare/prepare_darwin.go
```diff
@@ -19,13 +39,34 @@ func InstalledPreparedFiles(cfg autostart.Configurable) ([]string, error) {
...
+	if err != nil {
+		return nil, locale.WrapError(err, "err_svc_exec")
+	}
+
+	sc, err = autostart.New(autostart.Service, svcExec, []string{"start"}, cfg)
+	if err != nil {
+		return nil, locale.WrapError(err, "err_autostart_app")
+	}
+
+	path, err = sc.Path()
```
Naatan, line 65: This feels very similar if not identical to windows and linux. Does this need to be in prepareOS()? Curious if we can reduce the redundancy.

Naatan (changes requested):

activestate.windows.yaml
MDrakos: This prevents a window from opening when the service autostarts on Windows.

internal/osutils/autostart/autostart_mac.go
MDrakos: What exactly is giving issues with this pattern? Is it the language itself or something else? If a package like our `errs` package is forcing us into a certain pattern then we should file a story to determine why.

internal/osutils/autostart/autostart.go
```diff
@@ -1,77 +1,74 @@
...
-	if a.Name == Tray.String() {
-		if a.cfg.IsSet(ConfigKeyDisabled) {
+	if a.options.SetConfig {
+		if a.cfg.IsSet(a.options.ConfigKey) {
 			return nil
 		}
-		if err := a.cfg.Set(ConfigKeyDisabled, false); err != nil {
-			return errs.Wrap(err, "ConfigKeyDisabled=false failed")
+		if err := a.cfg.Set(a.options.ConfigKey, false); err != nil {
+			return errs.Wrap(err, "Could not set config key")
```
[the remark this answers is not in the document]
MDrakos: I can't find any place where the systray autostart config key is used. I think it was a remnant from when we set the tray to autostart on install but we no longer do that. I have removed the config key altogether. If we need this functionality back we can add it where it is needed ie. systray or state service

Naatan (dismissed):

MDrakos:
It looks like our `autostart` approach will not work for Linux without a desktop environment. Filled a follow-up story here: https://activestatef.atlassian.net/browse/DX-1067

Naatan (approved):

PR #1941, opened 2022-07-14
Improve `state use`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1046" title="DX-1046" target="_blank">StoryDX-1046</a>  As a user I can install a new runtime with state use
  </td></table>
  <br />
 

* All projects are stored in ~/Projects. (Configurable by env var for now.)
* `state use project` may be used if the project is already checked out. (Can omit org.)
* Switching to already checked out project does not re-check it out.
* Support `--path` and `--branch` options, just like `state activate`.

Still waiting on Product for messaging around what to print after running this. Currently we print "Switched to [name] located at [path]".

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -856,6 +856,8 @@ err_activate_create_project:
   other: Could not create new project
 err_invalid_namespace:
   other: "Invalid namespace: [NOTICE]{{.V0}}[/RESET]. Should be in the format of [NOTICE]org/project[/RESET], and can optionally contain a [NOTICE]#COMMIT_ID[/RESET] suffix. Names should be alphanumeric and may contain dashes and periods."
+err_invalid_project_name:
+  other: "Invalid project name: [NOTICE]{{.V0}}[/RESET]. Names should be alphanumeric, may contain dashes and periods, and can optionally contain a [NOTICE]#COMMIT_ID[/RESET] suffix."
 err_could_not_get_commit_behind_count:
   other: Could not obtain commit history properly
 err_must_create_project:
@@ -1868,11 +1870,17 @@ err_update_corrupt_install:
 arg_state_use_namespace:
   other: org/project
 arg_state_use_namespace_description:
-  other: The namespace of the project that you wish to use
-err_checkout_project:
+  other: The namespace of the project that you wish to use. If the project has already been checked out, you can simply use the project's name instead of the entire namespace.
+err_use_cannot_determine_projects_dir:
+  other: Cannot determine where to check out project to
+err_use_checkout_project:
   other: Could not checkout project {{.V0}}
-err_project_frompath:
+err_use_project_frompath:
   other: Could not create project files
+use_checkout_first:
+  other: "Please check it out first with [ACTIONABLE]state use org/{{.V0}}[/RESET]"
+err_use_project_not_checked_out:
+  other: "The project [ACTIONABLE]{{.V0}}[/ACTIONABLE] does not exist at [ACTIONABLE]{{.V1}}[/RESET]."
 err_language_by_commit:
   other: Could not get language from commit ID {{.V0}}
 err_make_language:
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1872,7 +1872,7 @@ flag_state_use_path_description:
 arg_state_use_namespace:
   other: org/project
 arg_state_use_namespace_description:
-  other: The namespace of the project that you wish to use. If the project has already been checked out, you can simply use the project's name instead of the entire namespace.
+  other: The org/project namespace of the project that you wish to use. If the project has already been checked out, you can simply use the project's name instead of the entire namespace.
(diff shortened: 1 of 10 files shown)


internal/runners/use/use.go
```diff
@@ -66,14 +69,44 @@ func (u *Use) Run(params *Params) error {
...
+		proj, err = project.FromPath(projectDir)
+		if err != nil {
+			wrapped := locale.WrapInputError(err, "err_use_project_not_checked_out", "", params.Namespace.Project, projectDir)
+			errs.AddTips(wrapped, locale.Tl("use_checkout_first", "", params.Namespace.Project))
+			return wrapped
+		}
+
+		if params.Branch != "" && proj.BranchName() != params.Branch {
+			return locale.NewInputError("err_conflicting_branch_while_checkedout", "", params.Branch, proj.BranchName())
+		}
```
MDrakos, line 109: I _think_ this is missing the following AC:
 - If namespace or projectName matches an already locally checked out project; use that instead of checking it out fresh

Wouldn't we have to check the config to see if a namespace matches a project checked out somewhere else on the system? This would mean that if I activated a project in a directory other than `~/Projects` I should be able `state use` that project without it being checked out again.

internal/locale/locales/en-us.yaml
```diff
@@ -1865,14 +1867,22 @@ err_update_corrupt_install:
...
+  other: The namespace of the project that you wish to use. If the project has already been checked out, you can simply use the project's name instead of the entire namespace.
+err_use_cannot_determine_projects_dir:
+  other: Cannot determine where to check out project to
+err_use_checkout_project:
   other: Could not checkout project {{.V0}}
-err_project_frompath:
+err_use_project_frompath:
   other: Could not create project files
+use_checkout_first:
+  other: "Please check it out first with [ACTIONABLE]state use org/{{.V0}}[/RESET]"
```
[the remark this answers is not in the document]
mitchell-as: Fair enough. I will add a "where 'org' is your organization's name" too.

internal/locale/locales/en-us.yaml
```diff
@@ -1865,14 +1867,22 @@ err_update_corrupt_install:
...
     Uninstallation instructions: {{.V0}}/troubleshooting/uninstall/
     Installation instructions: {{.V0}}/install/
+flag_state_use_path_description:
+  other: Where to install the project. The default location is in your ~/Projects directory
 arg_state_use_namespace:
   other: org/project
 arg_state_use_namespace_description:
-  other: The namespace of the project that you wish to use
-err_checkout_project:
+  other: The namespace of the project that you wish to use. If the project has already been checked out, you can simply use the project's name instead of the entire namespace.
```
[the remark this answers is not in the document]
mitchell-as: Fair enough.

internal/runners/use/use.go
mitchell-as: You are correct that this does not handle the case of a project checked out elsewhere. I have to add that. There is a unit test for this AC for projects checked out without using a custom path: https://github.com/ActiveState/cli/pull/1941/files#diff-5c34fd97485de0785b3179af8a6a7a4d5fe90ec253e26934b92640142105ceadR90

internal/runners/use/use.go
```diff
@@ -112,3 +145,22 @@ func (u *Use) Run(params *Params) error {
...
+
+func (u *Use) getLocalProjectPath(ns *project.Namespaced) string {
+	for namespace, paths := range projectfile.GetProjectMapping(u.config) {
+		if len(paths) == 0 {
+			continue
+		}
+		var namespaced project.Namespaced
+		err := namespaced.Set(namespace)
+		if err != nil {
+			logging.Debug("Cannot parse namespace: %v") // should not happen since this is stored
```
MDrakos, line 157: nit: If we make this `logging.Critical` it should get reported to rollbar. An error happening here would mean we're doing something wrong with storing our project mapping.

pkg/project/namespace_test.go
```diff
@@ -34,6 +34,22 @@ func TestParseNamespace_Invalid(t *testing.T) {
...
+	assert.Empty(t, parsed.CommitID)
+	assert.True(t, parsed.AllowOmitOwner)
+
+	parsed, err = ParseProjectNoOwner("project#a10-b11c12-d13e14-f15")
+	assert.NoError(t, err, "should be able to parse project part of namspace")
+	assert.Empty(t, parsed.Owner)
+	assert.Equal(t, parsed.Project, "project")
+	assert.Equal(t, *parsed.CommitID, strfmt.UUID("a10-b11c12-d13e14-f15"))
+	assert.True(t, parsed.AllowOmitOwner)
+}
```
MDrakos, line 51: nit: a lot of our unit tests are either split up where one tests is only testing one thing OR we use a table test (See [here](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests). Some Go tooling will also auto-generate the table tests for you. It's mostly personal preference though.

MDrakos (approved):
Looks good to me. Just a couple nits to consider.

PR #1942, opened 2022-07-15
Initial implementation of `state shell`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1048" title="DX-1048" target="_blank">StoryDX-1048</a>  We have a `state shell` command
  </td></table>
  <br />
 

Functions just like `state activate`, but ONLY starts a shell. There is a `prompt` alias for Windows users accustomed to that terminology.


internal/runners/shell/shell.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (u *Shell) Run(params *Params) error {
+	logging.Debug("Shell %v", params.Namespace)
+
+	projectDir := use.GetLocalProjectPath(params.Namespace, u.config)
+	if projectDir == "" {
+		err := locale.NewInputError("err_use_project_not_checked_out", "", params.Namespace.Project, projectDir)
+		errs.AddTips(err, locale.Tl("use_checkout_first", "", params.Namespace.Project))
```
daved, line 63: Are these KVs found in the locale file?

daved (changes requested):

internal/runners/shell/shell.go
mitchell-as: Yes, they were part of the `state use` improvements ticket.

daved (changes requested):

internal/runbits/use/use.go
```diff
@@ -0,0 +1,29 @@
+package use
```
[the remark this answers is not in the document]
mitchell-as: Okay, I moved it into pkg/project and did some refactoring to return a *Project struct with the ability to tell if an error is due to the project not existing locally, or some sort of projectfile parsing/loading error. The refactor touched other areas of the codebase, more changes may be needed upon review.

internal/runbits/activate/activation.go
```diff
@@ -6,20 +6,30 @@ import (
 	"path/filepath"
 	rt "runtime"
 
+	"github.com/ActiveState/cli/internal/analytics"
 	anaConst "github.com/ActiveState/cli/internal/analytics/constants"
```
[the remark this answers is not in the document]
mitchell-as: Ok.

pkg/project/project.go
```diff
@@ -358,6 +358,36 @@ func FromExactPath(path string) (*Project, error) {
 	return project, nil
 }
 
+// LocalProjectDoesNotExist is an error returned when a requested project is not checked out locally.
+type LocalProjectDoesNotExist struct{ error }
+
+// IsLocalProjectDoesNotExist checks if the error is a LocalProjectDoesNotExist.
+func IsLocalProjectDoesNotExist(err error) bool {
+	return errs.Matches(err, &LocalProjectDoesNotExist{})
```
Naatan, line 366: Personally I would prefer using this code directly instead of a helper function. It's concise and easy to understand, the helper just creates friction.

internal/runners/shell/shell.go
```diff
@@ -0,0 +1,90 @@
...
+			projectsDir, err2 := storage.ProjectsDir(u.config)
+			if err2 != nil {
+				return locale.WrapError(err2, "err_use_cannot_determine_projects_dir", "") // this error takes precedence
+			}
+			errs.AddTips(err, locale.Tl("use_checkout_first", "", params.Namespace.Project))
+			projectDir := filepath.Join(projectsDir, params.Namespace.Project)
+			return locale.WrapInputError(err, "err_use_project_not_checked_out", "", params.Namespace.Project, projectDir)
+		}
+		return locale.WrapError(err, "err_use_project_frompath") // error reading from project file
+	}
```
Naatan, line 73: This feels like it should be part of `FromNamespaceLocal`, because otherwise this is making unsafe assumptions about how that function is meant to operate.

ie. if `FromNamespaceLocal` can't figure out the namespace from the project dir then it should itself formulate a proper error message for this case.

internal/runners/shell/shell.go
```diff
@@ -0,0 +1,90 @@
...
+
+	rti, err := runtime.New(target.NewProjectTarget(proj, storage.CachePath(), nil, target.TriggerActivate), u.analytics, u.svcModel)
```
Naatan, line 75: This needs a new trigger. `TriggerActivate` is for `state activate`, which this is not. Suggest a new `TriggerShell`.

internal/runners/shell/shell.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		wrapped := locale.WrapInputError(err, "err_shell_load_runtime", "This project's runtime is not initialized.")
+		errs.AddTips(wrapped, locale.Tl("err_shell_load_runtime_tip", "Please run [ACTIONABLE]state use[/RESET] first."))
```
Naatan, line 78: Suggest dropping this for now, we'll want to tell them to use `state clone`, but we haven't committed to that name yet.

Added a note here:

https://activestatef.atlassian.net/browse/DX-1062

Naatan (changes requested):

pkg/project/project.go
mitchell-as: I was following this convention: https://github.com/ActiveState/cli/blob/e94d0b485e7775e975efa2ce6a27b15be86a5094/pkg/platform/runtime/ARCHITECTURE.md

cmd/state/internal/cmdtree/shell.go
```diff
@@ -0,0 +1,41 @@
...
+				Required:    true,
+				Value:       params.Namespace,
+			},
+		},
+		func(_ *captain.Command, _ []string) error {
+			return runner.Run(params)
+		},
+	)
+	cmd.SetGroup(EnvironmentGroup)
+	cmd.SetAliases("prompt")
```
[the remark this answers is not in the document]
mitchell-as: Looks like he has since agreed that this is okay.

internal/runners/shell/shell.go
mitchell-as: Okay.

internal/runners/use/use.go
```diff
@@ -77,24 +77,19 @@ func (u *Use) Run(params *Params) error {
...
 		} else if params.Namespace.Owner == "" {
-			projectsDir, err2 := storage.ProjectsDir(u.config)
-			if err2 != nil {
-				return locale.WrapError(err2, "err_use_cannot_determine_projects_dir", "") // this error takes precedence
-			}
-			errs.AddTips(err, locale.Tl("use_checkout_first", "", params.Namespace.Project))
-			projectDir := filepath.Join(projectsDir, params.Namespace.Project)
-			return locale.WrapInputError(err, "err_use_project_not_checked_out", "", params.Namespace.Project, projectDir)
+			// Note: use existing localized error message to workaround DX-740 for integration tests.
+			return locale.WrapInputError(err, "err_use_project_does_not_exist", err.Error())
```
Naatan, line 84: Nit: I'd prefer we drop the else on line 82, it makes for harder to read code and is unnecessary considering it returns on line 81.

pkg/project/project.go
```diff
@@ -358,6 +359,44 @@ func FromExactPath(path string) (*Project, error) {
...
+		}
+		var namespaced Namespaced
+		err := namespaced.Set(namespace)
+		if err != nil {
+			logging.Debug("Cannot parse namespace: %v") // should not happen since this is stored
+			continue
+		}
+		if (!ns.AllowOmitOwner && strings.ToLower(namespaced.String()) == strings.ToLower(ns.String())) ||
+			(ns.AllowOmitOwner && strings.ToLower(namespaced.Project) == strings.ToLower(ns.Project)) {
+			return FromPath(paths[0]) // just pick the first one
```
Naatan, line 386: Heads up; filed this story to follow up:

https://activestatef.atlassian.net/browse/DX-1075

Felt out of scope for this PR.

Naatan (changes requested):

internal/runners/shell/shell.go
```diff
@@ -60,23 +57,16 @@ func (u *Shell) Run(params *Params) error {
...
-			projectsDir, err2 := storage.ProjectsDir(u.config)
-			if err2 != nil {
-				return locale.WrapError(err2, "err_use_cannot_determine_projects_dir", "") // this error takes precedence
-			}
-			errs.AddTips(err, locale.Tl("use_checkout_first", "", params.Namespace.Project))
-			projectDir := filepath.Join(projectsDir, params.Namespace.Project)
-			return locale.WrapInputError(err, "err_use_project_not_checked_out", "", params.Namespace.Project, projectDir)
+		if project.IsLocalProjectDoesNotExistError(err) {
+			// Note: use existing localized error message to workaround DX-740 for integration tests.
+			return locale.WrapInputError(err, "err_shell_project_does_not_exist", err.Error())
```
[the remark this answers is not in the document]
mitchell-as: But I definitely want the integration tests to pass. They will fail if I don't have this workaround.

internal/runners/use/use.go
mitchell-as: It's an else if. I want a fall through on else.

internal/runners/shell/shell.go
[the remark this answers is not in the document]
mitchell-as: I've updated the ACs on DX-740 to remove these workarounds.

internal/runners/use/use.go
Naatan: My point is that if you drop the else (but keep the if) the logic is effectively the same but more readable.

Naatan (approved):

internal/locale/locales/en-us.yaml
```diff
@@ -1876,17 +1876,17 @@ flag_state_use_path_description:
 arg_state_use_namespace:
   other: org/project
 arg_state_use_namespace_description:
-  other: The org/project namespace of the project that you wish to use. If the project has already been checked out, you can simply use the project's name instead of the entire namespace.
-err_use_cannot_determine_projects_dir:
+  other: The org/project namespace of the project that you wish to use, or just the project name if previously used
```
daved, line 1879:
```suggestion
  other: The org/project namespace of the project that you wish to use, or just the project name if previously used.
```

daved (changes requested):

internal/locale/locales/en-us.yaml
mitchell-as: None of the other flag descriptions end in a period. This would stick out, no?

daved: Sure, go with what is best. I just scanned and thought the punctuation was missing.

daved (approved):

PR #1943, opened 2022-07-15
Remove sock file check from svc test

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1035" title="DX-1035" target="_blank">BugDX-1035</a>  State-svc's TestStartStop integration test keeps failing intermittently
  </td></table>
  <br />


internal/ipc/sockpath/sockpath.go
```diff
@@ -8,6 +8,7 @@ import (
 
 var (
 	sockpathExtension = "sock"
+	maxChannelLength  = 12
```
Naatan, line 11: This change doesn't seem relevant to the test? Please open a separate bug because otherwise this might not make it into the release. I definitely won't prioritize these changes for 0.34.1, but this particular change might make sense to include.

Naatan (changes requested):

internal/ipc/sockpath/sockpath.go
daved: This fixes a bug exposed by this PR's branch name length. If I remove the changes, I'll have to crate a new branch and PR for this same issue so that the failing tests will pass. Should I do so?

Naatan: Yes please. This PR is about fixing an integration test, whereas this fix is about fixing an end-user bug.

daved: If we prioritize [DX-1066](https://activestatef.atlassian.net/browse/DX-1066), then I can merge main into this branch and it will then pass tests so this PR can be left alone. We'll see how it goes at stand tomorrow.

Naatan (approved):

PR #1945, opened 2022-07-18
Made Projects directory configurable via "projects.directory" option.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1053" title="DX-1053" target="_blank">TaskDX-1053</a>  Project collection directory is configurable
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1868,7 +1868,7 @@ err_update_corrupt_install:
     Uninstallation instructions: {{.V0}}/troubleshooting/uninstall/
     Installation instructions: {{.V0}}/install/
 flag_state_use_path_description:
-  other: Where to install the project. The default location is in the {{.V0}} directory
+  other: Where to install the project. The default location is in the {{.V0}} directory, and it can be configured by running [ACTIONABLE]state config set projects.directory <path>[/RESET]
 arg_state_use_namespace:
   other: org/project
 arg_state_use_namespace_description:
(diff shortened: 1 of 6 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1868,7 +1868,7 @@ err_update_corrupt_install:
     Uninstallation instructions: {{.V0}}/troubleshooting/uninstall/
     Installation instructions: {{.V0}}/install/
 flag_state_use_path_description:
-  other: Where to install the project. The default location is in your ~/Projects directory
+  other: Where to install the project. The default location is in the {{.V0}} directory
```
MDrakos, line 1871: If the value of the project directory is changed in the config then is it still the default location? I think we could update this message to just say the current location.

Also, do we want to communicate to users how to change this value? Or is that something to be done at a later date?

MDrakos (changes requested):

internal/locale/locales/en-us.yaml
mitchell-as: When the project directory is changed in the config, this string will specify that newly configured location. Is that what you meant?

As far as communicating how to change this value, I presumed this would be in some sort of centralized documentation for all of our configuration options that has not yet been written. I suppose it might be a good idea to mention it there though.

MDrakos: What I mean is that the default is `~/Projects` but if I change that then this messaging says the default is whatever I changed it to. I admit that I'm being nit-picky here 😛

mitchell-as: Yes, whatever they changed it to becomes the "new default" with respect to this `--path` argument that's being documented. You are correct that it's not the "true" default in terms of out-of-box experience, but then there's always `state clean config`, right? :)

MDrakos (approved):

PR #1951, opened 2022-07-19
Show unstable warning, title, and banner if flag processing fails.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1039" title="DX-1039" target="_blank">BugDX-1039</a>  Command flags causing title / beta warning to get skipped
  </td></table>
  <br />

### DIFF
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -167,6 +167,21 @@ func NewCommand(name, title, description string, prime primer, flags []*Flag, ar
 		return err
 	})
 
+	// When there are errors processing flags, the command's runner is not called.
+	// In addition to performing the unstable feature check and printing of the unstable banner in
+	// the command's runner, we need to do it here too if there's an error processing flags.
+	// If the command is not unstable, simply return the flag error.
+	cmd.cobra.SetFlagErrorFunc(func(c *cobra.Command, err error) error {
+		if cmd.shouldWarnUnstable() {
+			if !condition.OptInUnstable(cmd.cfg) {
+				cmd.out.Notice(locale.Tr("unstable_command_warning", cmd.Name()))
+				return nil
+			}
+			cmd.outputTitleIfAny()
+		}
+		return err
+	})
+
 	cobraMapping[cmd.cobra] = cmd
 	return cmd
 }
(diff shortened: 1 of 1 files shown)


daved (changes requested):
Generally, this doesn't feel like the correct place to solve the issue. IMO, flags should be verified "best effort" before any network calls or other interesting things are expressed. So it should return a clear "your flag is empty" error or use a default value if the original implementer of that flag wishes it. Once there are no simple mistakes in the flags, checks like command stability can be run. That said, it is confusing why "You are not authenticated" is returned to the user. I suggest the control flow be understood better and ensuring that the correct error is shown or default value is used.

@Naatan Would you, please, weigh in on whether the story for this PR is best solved by preempting the flag parsing behavior?

mitchell-as:
Just so that we're all on the same page:

We currently do not handle the case where the user gives incorrect arguments to the state tool. We tell Cobra to execute our command and we handle any err returned. When flags are properly parsed, our command's `runner()` function is called. It is there that we can figure out if the command is unstable, and if so, require the "optin.unstable" config option to be set, print the unstable banner, etc.

The problem is when there are invalid flags given. If this is the case, then Cobra never calls our `runner()` function, so we are never able to figure out of the command in question is unstable, print the unstable banner, etc.

I stepped through the debugger and the function I'm hooking into appears to be the only place where I have access to the command the user is attempting to run, and only then can I determine if it's unstable, print the banner, etc.

If we want to have consistent behavior (of always requiring "optin.unstable" to be set for a command, printing the unstable banner, etc.), valid flags or not, then I think we have to handle it around the area that I'm proposing.

This is beside the point of the potentially confusing error message. (Hopefully that will go away when DX-740 is resolved.) Nathan wants to keep being able to use settable objects in Params{} to Cobra commands, so we need to be able to react to errors in setting (which is the case here when you're not authenticated).

daved:
> We currently do not handle the case where the user gives incorrect arguments to the state tool.
We should be handling bad flag and argument values at command parse time.

> If we want to have consistent behavior (of always requiring "optin.unstable" to be set for a command, printing the unstable banner, etc.), valid flags or not ...
I'm strongly suggesting that we fail fast on bad flag and arg values. If we do not go that route, then we should only be doing this in one place, before flags are parsed. @Naatan Thoughts?

> we need to be able to react to errors in setting (which is the case here when you're not authenticated).
How the provided value is handled in the runner is independent on whether the value is valid as a command line argument. We should be providing the user reasonable messaging so they can form their command properly.

daved (approved):

PR #1956, opened 2022-07-20
Print location of project being used.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1046" title="DX-1046" target="_blank">StoryDX-1046</a>  As a user I can install a new runtime with state use
  </td></table>
  <br />
 

This got lost somewhere in the process :(

### DIFF
--- a/internal/runners/use/use.go
+++ b/internal/runners/use/use.go
@@ -25,6 +25,7 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/runtime"
+	"github.com/ActiveState/cli/pkg/platform/runtime/setup"
 	"github.com/ActiveState/cli/pkg/platform/runtime/target"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/ActiveState/cli/pkg/projectfile"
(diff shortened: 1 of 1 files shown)


mitchell-as:
Still waiting on @rawktron for the correct messaging to communicate with users.

internal/runners/use/use.go
```diff
@@ -138,9 +140,11 @@ func (u *Use) Run(params *Params) error {
 		return locale.WrapError(err, "err_use_default", "Could not configure your project as the global default.")
 	}
 
-	u.out.Print(fmt.Sprintf("[NOTICE]%s[/RESET] [ACTIONABLE]%s[/RESET]",
+	u.out.Print(fmt.Sprintf("[NOTICE]%s[/RESET] [ACTIONABLE]%s[/RESET] %s [ACTIONABLE]%s[/RESET]",
 		locale.Tl("use_notice_switched_to", "Switched to"),
-		params.Namespace.Project),
+		params.Namespace.Project,
+		locale.Tl("use_notice_located_at", "located at"),
+		setup.ExecDir(projectTarget.Dir())),
```
MDrakos, line 147: I think this can be accomplished without the `fmt.Sprintf`. The locale string would be something like:
`"[NOTICE]{{.V0}}[/RESET] [ACTIONABLE]{{.V1}}[/RESET]{{.V2}}[ACTIONABLE]{{.V3}}[/RESET]"`

You can find examples of this in our `en-us.yaml` file

MDrakos (changes requested):

MDrakos (approved):

PR #1960, opened 2022-07-22
Convert Windows directory separators to '/' when using with bash.exe.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-542" title="DX-542" target="_blank">BugDX-542</a>  Integration test working directory on Windows is not correct
  </td></table>
  <br />
 

Also use correct suite tag.

The previous PR did not actually fix this because the suite tag was wrong.


mitchell-as:
@MDrakos ping

MDrakos (approved):

PR #1963, opened 2022-07-25
Verify checksum of downloaded artifacts.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1064" title="DX-1064" target="_blank">StoryDX-1064</a>  Validate checksums of artifacts
  </td></table>
  <br />
 

Camel builds do not provide artifact checksums, so skip validating those.

Note: we also validate the checksum of artifacts in the cache in the event a bad actor wants to substitute their own artifacts for ours. In that case, the validation will fail silently and a replacement will be downloaded from the platform.

### DIFF
--- a/pkg/platform/runtime/artifactcache/artifactcache.go
+++ b/pkg/platform/runtime/artifactcache/artifactcache.go
@@ -106,6 +106,20 @@ func (cache *ArtifactCache) Store(a artifact.ArtifactID, archivePath string) err
 	cache.mutex.Lock()
 	defer cache.mutex.Unlock()
 
+	// Replace an existing artifact in the cache.
+	// This would really only happen if a checksum validation fails and the artifact is silently
+	// re-downloaded from the platform.
+	if existingArtifact, found := cache.artifacts[a]; found {
+		path := existingArtifact.ArchivePath
+		logging.Debug("Replacing cached artifact '%s'", path)
+		err := os.Remove(path)
+		if err != nil {
+			return errs.Wrap(err, "Unable to overwrite existing artifact '%s'", path)
+		}
+		delete(cache.artifacts, existingArtifact.Id)
+		cache.currentSize -= existingArtifact.Size
+	}
+
 	stat, err := os.Stat(archivePath)
 	if err != nil {
 		return errs.Wrap(err, "Unable to stat artifact '%s'. Does it exist?", archivePath)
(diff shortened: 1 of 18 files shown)


pkg/platform/runtime/artifactcache/artifactcache.go
```diff
@@ -106,6 +106,20 @@ func (cache *ArtifactCache) Store(a artifact.ArtifactID, archivePath string) err
...
+	if existingArtifact, found := cache.artifacts[a]; found {
+		path := existingArtifact.ArchivePath
+		logging.Debug("Replacing cached artifact '%s'", path)
+		err := os.Remove(path)
+		if err != nil {
+			return errs.Wrap(err, "Unable to overwrite existing artifact '%s'", path)
+		}
+		delete(cache.artifacts, existingArtifact.Id)
+		cache.currentSize -= existingArtifact.Size
+	}
```
Naatan, line 121: I think if checksum validation failed it should error out and not reach this point. But pending Pete's call here: https://activestate.slack.com/archives/CFN2A6P9N/p1658859008105439

pkg/platform/runtime/setup/setup.go
```diff
@@ -579,10 +579,37 @@ func (s *Setup) downloadArtifactWithProgress(unsignedURI string, targetFile stri
...
+// platform, and returns an error if the verification fails.
+func (s *Setup) verifyArtifact(archivePath string, a artifact.ArtifactDownload) error {
+	if a.Checksum != "" {
+		logging.Debug("Validating checksum for %s", archivePath)
+	} else {
+		logging.Debug("Skipping checksum validation for %s because the Platform did not provide a checksum to validate against.")
+		return nil
+	}
+
+	checksum, err := fileutils.Hash(archivePath)
```
Naatan, line 592: We should maybe rename `Hash` to `Sha256Hash` or something more explicit. Since this usage cares strongly about the hashing mechanism used, whereas that function does not. Seems we could implicitly break runtime checksum validation by making changes to `fileutils.Hash()` without realizing that it's used in such opinionated ways.

Naatan (changes requested):
Really nice work! Just one fairly nitpicky change and possibly a change to how we handle errors depending on Pete's response.

pkg/platform/runtime/artifactcache/artifactcache.go
mitchell-as: I can edit the comment to be more explicit about this happening if the cached artifact's checksum fails (e.g. replaced by a bad actor). If the validation fails for a downloaded artifact, the entire process will error out.

mitchell-as: Pete says this behavior as programmed is good (hit the cache first, then silently redownload from Platform if checksum fails; if Platform artifact checksum validation fails, simply error out reporting that).

Naatan (approved):

PR #1964, opened 2022-07-26
BE-1031-there-is-a-standalone-tool-for-unpacking-and-relocating-artifacts-in-installers

rickprice:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/BE-1031" title="BE-1031" target="_blank">StoryBE-1031</a>  There is a standalone tool for unpacking and relocating artifacts in installers
  </td></table>
  <br />
 

Custom installer to handle offline installs.


cmd/state-custom-installer/docker/centos-7/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM docker-registry.activestate.build/activestate/centos-7.6-builder:1.0.66 AS custom-installer-builder
```
icanhasmath, line 1: FROM docker-registry.activestate.build/activestate/centos-7.6-builder:1.0.67 AS custom-installer-builder

icanhasmath: Needs an increment.

cmd/state-custom-installer/docker/centos-8/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM docker-registry.activestate.build/activestate/centos-8-builder:2.0.12 AS offline-install-builder
```
icanhasmath, line 1: Increment here too. 2.0.13

cmd/state-custom-installer/docker/centos-7/runCompile.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+cd cli
+git pull
+git checkout BE-1031-there-is-a-standalone-tool-for-unpacking-and-relocating-artifacts-in-installers
```
icanhasmath, line 5: If this is an alternate branch, there should be a tag instead.

cmd/state-custom-installer/docker/centos-8/runCompile.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+cd cli
+git pull
+git checkout BE-1031-there-is-a-standalone-tool-for-unpacking-and-relocating-artifacts-in-installers
```
icanhasmath, line 5: Same as above.

cmd/state-custom-installer/docker/centos-7/Dockerfile
icanhasmath: Probably make the centos-7 / Centos-8 a configurable option to the Docker build, as we're basically duplicating the code except for this line.

cmd/state-custom-installer/runBuildInteractive.sh
```diff
@@ -0,0 +1,15 @@
...
+rm -rf build/custom-installer
+rm -rf /tmp/artifacts-*
+rm -rf /tmp/FrederickTest
+gofmt -s -w *.go
+go build -o build/custom-installer github.com/ActiveState/cli/cmd/state-custom-installer
+gozip -c build/custom-installer LICENSE.txt artifacts.tar.gz
+# ./build/offline-installer install /tmp/FrederickTestII
+./build/custom-installer install /tmp/FrederickTest
+./build/custom-installer install /tmp/FrederickTest
+./build/custom-installer uninstall /tmp/FrederickTest
```
icanhasmath, line 15: FredrickTest is not general enough.

cmd/state-custom-installer/runUnInstall.go
```diff
@@ -0,0 +1,61 @@
...
+// NOTE: From: internal/runners/clean/uninstall.go
+
+func runUnInstall(out output.Outputer, params *Params) error {
+	// This is the deploy id constant: DeployID RcIdentification = RcIdentification{
+	installToDir := params.path
+	// Should look for license file to see if it exists before we delete the install directory
+	uninstallFileCheck := filepath.Join(installToDir, "LICENSE.txt")
+
+	if !fileutils.FileExists(uninstallFileCheck) {
+		return errs.New("Doesn't seem like given directory is an install directory")
```
icanhasmath, line 26: Change to something like this:
`ActiveState license not found. Please specify a valid install directory.`

cmd/state-custom-installer/runUnInstall.go
```diff
@@ -0,0 +1,61 @@
...
+
+	err = removeEnvPaths(cfg)
+	if err != nil {
+		fmt.Println(err)
+		return err
+	}
+
+	fmt.Println("")
+	fmt.Println("")
+	fmt.Println("Runtime de-installation completed.")
```
icanhasmath, line 42: `Uninstall Complete`

cmd/state-custom-installer/runUnInstall.go
icanhasmath: Though, if the process deletes the LICENSE.txt, and hasn't completed the user will be stuck.  Probably we'll need the user to be allowed to proceed with file removal anyway.  `Are you sure you want to uninstall ? y/[N]`

vendor/github.com/ActiveState/gozip/LICENSE.txt
```diff
@@ -0,0 +1,5 @@
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
```
icanhasmath, line 1: I thought we would fork the repository into its own space like we did for Pillow.  I don't think we should be forking this into TheHomeRepot.

icanhasmath (changes requested):
Lots of comments, please review them, and fix up the comment to be clear as to what the Custom installer is, and what components are being added.

cmd/state-custom-installer/docker/centos-7/runCompile.sh
rickprice: No, waiting for assimilation to Master.

cmd/state-custom-installer/runBuildInteractive.sh
rickprice: Yeah, makes sense.

vendor/github.com/ActiveState/gozip/LICENSE.txt
rickprice: I'm wondering if this is just the way DX is doing things, but I will happily check, having two copies isn't necessarily a great idea.

cmd/state-custom-installer/runUnInstall.go
rickprice: Okay, changed the behaviour.

vendor/github.com/ActiveState/gozip/LICENSE.txt
Naatan: Were you unable to use https://github.com/ActiveState/cli/blob/master/internal/unarchiver/zip.go ?

I'd rather we update that to add the functionality you need, rather than add a new dependency.

cmd/state-custom-installer/docker/centos-7/runCompile.sh
rickprice: Changed the script to assume we are running from Master.

cmd/state-custom-installer/docker/centos-8/runCompile.sh
rickprice: This code was rolled into the Centos-7 code, and was deleted.

vendor/github.com/ActiveState/gozip/LICENSE.txt
rickprice: I can look at the code to use the internal unzip code, and will copy across what is needed from Gozip, as needed. I expect that things should just work for the most part without needing to copy anything. Gozip is just a thin eggshell over the internal Go zip library anyway.

Naatan (changes requested):
Please assign me for review when you're ready for me to have a look.

vendor/github.com/ActiveState/gozip/LICENSE.txt
rickprice: I changed the custom-installer code to use the basic Golang primitives instead of using the Gozip library.

cmd/state-custom-installer/docker/centos/runBuild.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
Naatan, line 1: Please use the activestate.yaml for scripts (this goes for all .sh scripts introduced in this PR).

Note you can create a new `activestate.installer.yaml` or `activestate.<whatever>.yaml` if you want to keep it organized. It'll automatically get merged together at runtime.

cmd/state-custom-installer/docker/centos/runCompile.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+cd cli
+git pull
+# git checkout BE-1031-there-is-a-standalone-tool-for-unpacking-and-relocating-artifacts-in-installers 
+go mod vendor
```
Naatan, line 6: This should not happen in a script. The vendored files should be committed to the repo, not pulled in at build time.

cmd/state-custom-installer/docker/centos/runCompile.sh
```diff
@@ -0,0 +1,12 @@
...
+sudo /state-tool/bin/state run preprocess
```
Naatan, line 7: Don't use sudo. State tool is meant to run at user level.

cmd/state-custom-installer/docker/centos/runCompile.sh
```diff
@@ -0,0 +1,12 @@
...
+cd ..
+sudo chown -R as-build:as-build cli
+cd cli/cmd/state-custom-installer/docker
+go build -o build/custom-installer github.com/ActiveState/cli/cmd/state-custom-installer
+sudo cp build/custom-installer /custom-installer-data
```
Naatan, line 12: Why not just output to `/custom-installer-data` directly if that's where you want it?

cmd/state-custom-installer/runBuildInteractive.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
Naatan, line 1: This would be better serviced by an integration test, see for example: https://github.com/ActiveState/cli/blob/820f9e2c27f107c46fc5cf136d52ac2371000927/cmd/state-installer/test/integration/installer_int_test.go#L28

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/ActiveState/cli/pkg/platform/runtime/target"
+	"github.com/ActiveState/cli/pkg/project"
+	"github.com/go-openapi/strfmt"
+)
+
+func runInstall(out output.Outputer, params *Params) error {
+
+	tempDir, err := ioutil.TempDir("", "artifacts-")
+	if err != nil {
+		fmt.Println(err)
```
Naatan, line 35: This should probably return the error instead. Right now you're printing it and then continuing execution of this function.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// #FIX: Need to uncomment this for production
```
Naatan, line 38: Should drop this before merging.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+	defer os.RemoveAll(tempDir)
+
+	fmt.Printf("Temp directory is: %s\n", tempDir)
```
Naatan, line 41: Use the outputer you passed as an arg to this function instead.

This goes for any place where you are using `fmt.Print..`.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+
+	installToDir := params.path
+	artifactsTarGz := "artifacts.tar.gz"
+	assetsPath := filepath.Join(tempDir, "assets")
+	artifactsPath := filepath.Join(tempDir, "artifacts")
+	licenseFilePath := filepath.Join(assetsPath, "LICENSE.txt")
```
Naatan, line 47: I recommend recording these as constants at the top of this file somewhere.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Double check if installation directory exists already
+	if fileutils.DirExists(installToDir) {
+		empty, err := fileutils.IsEmptyDir(installToDir)
+		if err == nil && !empty {
```
Naatan, line 52: `err != nil` is never raised.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+			configureEnvironmentAccepted, err := getConfirmation("Installation directory is not empty, install anyway?")
+			if err != nil {
+				return err
```
Naatan, line 55: Always wrap your errors, otherwise it'll be very difficult to trace where an error came from.

eg. `errs.Wrap(err, 'could not get confirmation to install to non-empty dir')

Without this you might get an error like "Could not write to stdin" without any clue as to what caused it.

I won't comment on every err being returned in this PR, but I recommend going over them and wrapping them all.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+
+	if err := os.Mkdir(assetsPath, os.ModePerm); err != nil {
+		return err
+	}
+
+	if err := os.Mkdir(artifactsPath, os.ModePerm); err != nil {
+		return err
+	}
+
+	err = unzip(os.Args[0], assetsPath)
```
Naatan, line 72: Use `params` instead. Once you're in the command execution you generally should not be using `os.Args` anymore.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return err
+	}
+
+	eulaAccepted, err := showEULAAndGetAcceptance(licenseFilePath)
+	if err != nil {
+		return err
+	}
+
+	if !(eulaAccepted) {
```
Naatan, line 82:
```suggestion
	if !eulaAccepted {
```

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+		return errors.New("\n\nEnd user didn't accept the EULA")
```
Naatan, line 83: This should be an input error; eg. 

```
locale.NewInputError("err_eula_not_accepted", "End user didn't accept the EULA")
```

Note the second argument is the fallback string. You can instead just use:

```
locale.NewInputError("err_eula_not_accepted")
```

And add an entry for `err_eula_not_accepted` to `internale/locale/locales/en-us.yaml`.

cmd/state-custom-installer/custom-installer.go
```diff
@@ -0,0 +1,114 @@
...
+		logging.CurrentHandler().SetVerbose(true)
+	}
+
+	runErr := run()
+	if runErr != nil {
+		errMsg := errs.Join(runErr, ": ").Error()
+		if locale.IsInputError(runErr) {
+			logging.Debug("state-offline-installer errored out due to input: %s", errMsg)
+		} else {
+			multilog.Critical("state-offline-installer errored out: %s", errMsg)
```
Naatan, line 42: Please take care to actually use input errors when an error is caused by user input. Otherwise, we're going to get a lot of superfluous errors reported on our rollbar.

Naatan: Additionally, you'll need to call:

```
rollbar.SetupRollbar("TOKEN")
```

Otherwise error reporting won't work. Seeing as this is an offline installer I'm guessing you might not want error reporting at all, in this case just don't use `multilog` and use `logging` instead.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+		err := errs.Wrap(err, "Could not unpack artifact %s", archivePath)
+		return err
+	}
+	fmt.Println("Stage 2 of 2 Finished: Decompressing artifacts into: ", artifactsPath)
+
+	owner := "voidOwner"
+	name := "voidProject"
+	commitID := strfmt.UUID("000000000-0000-0000-0000-000000000000")
+
+	offlineTarget := target.NewOfflineTarget(owner, name, commitID, installToDir, artifactsPath)
```
Naatan, line 107: Feel free to drop owner/name/commit or anything else from the offline target that you don't need. We introduced that for you guys, you're the only consumers.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+
+	analytics := blackhole.New()
+	mockProgress := &OfflineProgressOutput{}
+	logfile, _ := buildlogfile.New(outputhelper.NewCatcher())
```
Naatan, line 111: Please don't ever leave errors unhandled. Go does not have exceptions, errors are your lifeline.

cmd/state-custom-installer/runInstall.go
```diff
@@ -0,0 +1,188 @@
...
+	eventHandler := events.NewRuntimeEventHandler(mockProgress, nil, logfile)
+
+	rti, err := runtime.New(offlineTarget, analytics, nil)
+	if err != nil && runtime.IsNeedsUpdateError(err) {
+		err = rti.Update(nil, eventHandler)
+		if err != nil {
+			fmt.Println("Had an installation error")
+			logging.Warning("Had an installation error: %v", err)
+			fmt.Println(err)
+			// return error
```
Naatan, line 121: Returning seems appropriate..

cmd/state-custom-installer/runUnInstall.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/ActiveState/cli/internal/subshell/sscommon"
+)
+
+// NOTE: From: internal/runners/clean/uninstall.go
+
+func runUnInstall(out output.Outputer, params *Params) error {
+	// This is the deploy id constant: DeployID RcIdentification = RcIdentification{
+	installToDir := params.path
+	// Should look for license file to see if it exists before we delete the install directory
+	uninstallFileCheck := filepath.Join(installToDir, "LICENSE.txt")
```
Naatan, line 23: This isn't much of a verification imo, `LICENSE.txt` is a very common filename and does not uniquely identify our runtime installation dir. Maybe also check if it contains 'ActiveState' or something?

cmd/state-custom-installer/runUnInstall.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+	}
+
+	cfg, err := config.New()
+	if err != nil {
+		return err
+	}
+
+	err = removeEnvPaths(cfg)
```
Naatan, line 43: It looks like uninstall is ONLY removing env config but otherwise leaves the install dir itself intact. Is that intended?

cmd/state-custom-installer/runUnInstall.go
```diff
@@ -0,0 +1,70 @@
...
+	// return errors.New("\n\nThis is an error")
+	isAdmin, err := osutils.IsAdmin()
+	if err != nil {
+		return errs.Wrap(err, "Could not determine if running as Windows administrator")
+	}
+
+	// remove shell file additions
+	s := subshell.New(cfg)
+	if err := s.CleanUserEnv(cfg, sscommon.DeployID, !isAdmin); err != nil {
+		return errs.Wrap(err, "Failed to remove State Tool deployment PATH")
```
Naatan, line 66:
```suggestion
		return errs.Wrap(err, "Failed to remove runtime PATH")
```

cmd/state-custom-installer/util.go
```diff
@@ -0,0 +1,152 @@
...
+	"golang.org/x/term"
+	"io"
+	"os"
+	"path"
+	"path/filepath"
+)
+
+func showEULAAndGetAcceptance(filename string) (bool, error) {
+	for {
+		fmt.Println("Do you accept the Activestate End User Agreement? (d to display) [y/n/d]:")
```
Naatan, line 17: https://github.com/ActiveState/cli/blob/fe7211dfc67ced5052a2b3f2b8243f83fedcb890/internal/prompt/prompt.go#L23

cmd/state-custom-installer/util.go
```diff
@@ -0,0 +1,152 @@
+package main
```
Naatan, line 1: Nit: a generic util file like this is a breeding ground for unorganized code. Recommend splitting the logic contained into separate files with distinctive naming.

Naatan (changes requested):

Naatan:
Note the 
image

cmd/state-custom-installer/custom-installer.go
```diff
@@ -0,0 +1,114 @@
+package main
```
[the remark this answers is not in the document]
rickprice: Changed it back again to state-offline-installer.

.gitignore
```diff
@@ -14,3 +14,7 @@ obj/
 bin/
 installers/msi-language/Product.wxs
 node_modules
+
+cmd/state-offline-installer/build
+cmd/state-offline-installer/artifacts.tar.gz
```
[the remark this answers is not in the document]
rickprice: Moved things to the top level build directory.

cmd/state-custom-installer/docker/centos/Dockerfile
```diff
@@ -0,0 +1,41 @@
+ARG CENTOS_VERSION=centos-7.6-builder:1.0.67
```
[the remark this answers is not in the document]
rickprice: Moved.

activestate.yaml
```diff
@@ -73,17 +73,21 @@ scripts:
...
+          if [ -n "${pkgs}" ]; then
+            echo Installing packages "${pkgs}"
+            sudo apt-get update && sudo apt-get install ${pkgs} -y
+          fi
+        elif [ hash yum &> /dev/null ]; then
+          yum install -y gtk3-devel libappindicator-gtk3 webkitgtk4
         fi
       }
-      aptin libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev
+      aptin
```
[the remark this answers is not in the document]
rickprice: Changes made.

cmd/state-custom-installer/runInstall.go
rickprice: Code re-architected

cmd/state-custom-installer/custom-installer.go
```diff
@@ -0,0 +1,114 @@
...
+
+	params := newParams()
+	cmd := captain.NewCommand(
+		path.Base(os.Args[0]), "", "", p, nil, nil,
+		func(ccmd *captain.Command, args []string) error {
+			out.Print(ccmd.UsageText())
+			return nil
+		},
+	)
+	if isZip(os.Args[0]) {
```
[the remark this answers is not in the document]
rickprice: We determine if our executable has the backpack attached or not, if not, then we must be the uninstaller only. So, basically, if we have no backpack, then assume we are the standalone uninstaller download. Otherwise, give them both options as usual. Without a backpack we can't install anything anyway.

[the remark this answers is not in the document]
Naatan: That wasn't my point though. You can get the behaviour you want and address the issue I raised. The issue is that the command registration should not be dependent on what input is passed, instead, it's the execution of the command that should consider the input.

If I run `state-offline-installer --help` it will give me help output for only one or the other command, and the actual execution will behave in ways that the user can't predict based on the help output. That's not the only issue, but a good demonstration of why this is putting the carriage in front of the horse.

Command registration should be independent of input. Command execution may consider input. We should design our commands accordingly.

cmd/state-offline-installer/offline-installer.go
```diff
@@ -0,0 +1,114 @@
...
+	})
+	if err != nil {
+		return errs.Wrap(err, "Could not initialize outputer")
+	}
+
+	p := primer.New(nil, out, nil, nil, nil, nil, nil, nil, nil, nil)
+
+	params := newParams()
+	cmd := captain.NewCommand(
+		path.Base(os.Args[0]), "", "", p, nil, nil,
```
Naatan, line 64: Please provide all the args. They are used for `--help` output amongst other things.

cmd/state-offline-installer/runInstall.go
```diff
@@ -0,0 +1,187 @@
...
+	out.Print(fmt.Sprintf("Stage 2 of 2 Start: Decompressing artifacts into: %s", artifactsPath))
+	err = unpackArtifact(unarchiver.NewTarGz(), archivePath, artifactsPath)
+	if err != nil {
+		return errs.Wrap(err, "Could not unpack artifact %s", archivePath)
+	}
+	out.Print(fmt.Sprintf("Stage 2 of 2 Finished: Decompressing artifacts into: %s", artifactsPath))
+
+	offlineTarget := target.NewOfflineTarget(installToDir, artifactsPath)
+
+	// FIX: Need to do this more correctly with a New function
```
Naatan, line 112: We have a rule on our repo which is that we are not allowed to put todo's in code comments as our code is not a task tracker. I'd say either address this, remove it, or file a jira ticket to follow up at a later date.

cmd/state-offline-installer/runInstall.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+const artifactsTarGZName = "artifacts.tar.gz"
+const assetsPathName = "assets"
+const artifactsPathName = "artifacts"
+const licenseFileName = "LICENSE.txt"
+
+func runInstall(out output.Outputer, params *Params) error {
+
+	analytics := blackhole.New()
```
Naatan, line 36: Are analytics out of scope for this iteration?

cmd/state-offline-installer/runInstall.go
```diff
@@ -0,0 +1,187 @@
...
+	configureEnvironmentAccepted, err := prompter.Confirm("Setup", "Setup environment for installed project?", &default_boolean_answer)
+	if err != nil {
+		return errs.Wrap(err, "Error getting confirmation")
+	}
+
+	if configureEnvironmentAccepted {
+		// This is the code to setup the environment when we install...
+		cfg, err := config.New()
+		if err != nil {
+			return errs.Wrap(err, "Error configuring environment")
```
Naatan, line 138:
```suggestion
			return errs.Wrap(err, "Error initializing config")
```

cmd/state-offline-installer/offline-installer.go
```diff
@@ -0,0 +1,114 @@
...
+
+	out, err := output.New("", &output.Config{
+		OutWriter: os.Stdout,
+		ErrWriter: os.Stderr,
+	})
+	if err != nil {
+		return errs.Wrap(err, "Could not initialize outputer")
+	}
+
+	p := primer.New(nil, out, nil, nil, nil, nil, nil, nil, nil, nil)
```
Naatan, line 60: I would strongly suggest using primer the way it was intended. I see you are using config, prompter, analytics, .. throughout your code. Ideally you configure that only once on the primer, and then you pass the primer struct along, rather than initialize it again in your code.

Doing it this way can lead to race conditions. eg. you could initialize config twice which could lead to writes/reads happening in async without awareness of one another.

cmd/state-offline-installer/runInstall.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		env, err := rti.Env(false, false)
+		if err != nil {
+			return errs.Wrap(err, "Error setting environment")
```
Naatan, line 143:
```suggestion
			return errs.Wrap(err, "Error retrieving environment")
```

cmd/state-offline-installer/runInstall.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		sshell := subshell.New(cfg)
+
+		if rt.GOOS == "windows" {
+			contents, err := assets.ReadFileBytes("scripts/setenv.bat")
+			if err != nil {
+				return errs.Wrap(err, "Error reading file bytes")
+			}
+			err = fileutils.WriteFile(filepath.Join(installToDir, "setenv.bat"), contents)
```
Naatan, line 153: Note: better not to reuse the err variable as it can break nil type assertions when two different errors pass a different type (both of which will satisfy the error interface).

cmd/state-offline-installer/runInstall.go
```diff
@@ -0,0 +1,187 @@
...
+
+		binPath := filepath.Join(offlineTarget.Dir(), "bin")
+		if err := fileutils.MkdirUnlessExists(binPath); err != nil {
+			return locale.WrapError(err, "err_deploy_binpath", "Could not create bin directory.")
+		}
+
+		namespace_to_use := project.NewNamespace("owner", "project", "commitID")
+
+		// Write global env file
+		err = sshell.SetupShellRcFile(binPath, env, *namespace_to_use)
```
Naatan, line 172: This will lead to `[owner/project]` being part of their prompt. I think you'll want to update `SetupShellRCFile()` to respect a nil value for the namespace and skip that prompt logic.

cmd/state-offline-installer/runInstall.go
```diff
@@ -0,0 +1,187 @@
...
+			if err != nil {
+				return errs.Wrap(err, "Error reading file bytes")
+			}
+			err = fileutils.WriteFile(filepath.Join(installToDir, "setenv.bat"), contents)
+			if err != nil {
+				return locale.WrapError(err, "err_deploy_write_setenv", "Could not create setenv batch scriptfile at path: {{.V0}}", installToDir)
+			}
+		}
+
+		err = sshell.WriteUserEnv(cfg, env, sscommon.DeployID, true)
```
Naatan, line 159: `sscommon.DeployID` is for `state deploy`. You'll want to create your own ID for the offline installer.

Otherwise `state deploy` and the offline installer could end up conflicting with one another.

cmd/state-offline-installer/test-for-install-directory.go
```diff
@@ -0,0 +1,22 @@
...
+	licenseFilePath := filepath.Join(installDirectory, licenseFileName)
+
+	// read the whole file at once
+	b, err := ioutil.ReadFile(licenseFilePath)
+	if err != nil {
+		return false
+	}
+
+	//check whether s contains substring text
+	return bytes.Contains(b, []byte(checkFileText))
```
Naatan, line 21: https://github.com/ActiveState/cli/blob/263212252142ab44d21b4ac0c849335c1b37be2e/internal/fileutils/fileutils.go#L1006

cmd/state-offline-installer/zip.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return errs.Wrap(err, "Prepare for unpacking failed")
+	}
+	defer f.Close()
+	return ua.Unarchive(f, i, targetDir)
+}
+
+// Unzip unzips the file zippath and puts it in destination
+func unzip(zippath string, destination string) (err error) {
+	r, err := zip.OpenReader(zippath)
```
Naatan, line 24: Are you unable to use the unarchiver? 

https://github.com/ActiveState/cli/blob/cbaa27607e1fc2c8063e11d9bfe10caa146ec0d3/internal/unarchiver/zip.go#L26

deployments/docker/state-offline-installer/centos/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+USER root
+
+# Install required basic commands
+RUN yum install -y git go curl
+RUN yum install -y gtk3-devel libappindicator-gtk3 webkitgtk4
+ENV GOPATH=/go
+ENV PATH=$PATH:/go/bin
+
+# Install gozip which we use to add files to our installer
+RUN go install github.com/ActiveState/gozip/cmd/gozip@latest
```
Naatan, line 14: I can't find any reference to this being used?

deployments/docker/state-offline-installer/centos/Dockerfile
```diff
@@ -0,0 +1,41 @@
+ARG CENTOS_VERSION=centos-7.6-builder:1.0.67
```
Naatan, line 1: What is this Docker accomplishing exactly? I feel like this should exist entirely within the activestate.yaml (see the build scripts we have).

Relying on Docker feels like working around using the State Tool in this case.

deployments/docker/state-offline-installer/centos/runCompile.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+cd cli
+git pull
+go mod vendor
+sudo /state-tool/bin/state run preprocess
+cd ..
+sudo chown -R as-build:as-build cli
+cd cli/cmd/state-offline-installer/docker
+go build -o /offline-installer-data/offline-installer github.com/ActiveState/cli/cmd/state-offline-installer
```
Naatan, line 10: `go mod vendor` should not be part of this, we commit our vendored modules, so by the time you're building this should already be in place.

Also this definitely should be an activestate.yaml script.

Naatan (changes requested):

activestate.offlineinstall.yaml
```diff
@@ -0,0 +1,38 @@
...
+    language: bash
+    description: Builds the project with the specified OS as the target OS. (valid values darwin, linux, windows)
+    value: |
+      set -e
+      $constants.SET_ENV
+      export GOOS=${1}
+      export GOARCH=amd64
+      [ -z "${2}" ] && >&2 echo "must provide output file as 2nd arg" && exit 1
+      go build -tags "internal $GO_BUILD_TAGS" -o ${2} $constants.CLI_BUILDFLAGS $constants.OFFLINE_INSTALL_PKGS
+      go build -tags "internal $GO_BUILD_TAGS" -o ${2} $constants.CLI_BUILDFLAGS $constants.OFFLINE_UNINSTALL_PKGS
```
Naatan, line 38: Suggest dropping this. `build-for` was introduced for frankly a very negligible use-case. I'd like to get rid of it.

cmd/state-offline-installer/offline-installer.go
```diff
@@ -67,28 +69,26 @@ func run() error {
...
-						Value:       &params.path,
-						Required:    true,
-					},
-				},
-				func(ccmd *captain.Command, args []string) error {
-					logging.Debug("Running CmdInstall")
-					return runInstall(out, params)
+	cmd.AddChildren(
+		captain.NewCommand(
+			cmdInstall,
```
Naatan, line 74: It's a little awkward that you have to run `./installer install` rather than just `./installer`. Why not get rid of the subcommand and run off of the main command?

cmd/state-offline-installer/offline-installer.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+			func(ccmd *captain.Command, args []string) error {
+				logging.Debug("Running CmdInstall")
+				return runInstall(out, params)
+			},
+		),
+	)
+	cmd.AddChildren(
+		captain.NewCommand(
+			cmdUnInstall,
```
Naatan, line 94: Suggest dropping this. We shouldn't provide (and thus support & maintain) two different methods of running the uninstaller unless there's a strong reason to.

cmd/state-offline-installer/runBuildInteractive.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
Naatan, line 1: This should be an activestate.yaml script.

cmd/state-offline-installer/runInstall.go
```diff
@@ -77,41 +78,55 @@ func runInstall(out output.Outputer, params *Params) error {
...
 
-	if !eulaAccepted {
-		return locale.NewInputError("err_eula_not_accepted", "End user didn't accept the EULA")
-	}
+	out.Print(fmt.Sprintf("Stage 1 of 3 Finished: Decompressing assets into: %s", assetsPath))
 
-	out.Print(fmt.Sprintf("Stage 1 of 2 Start: Decompressing assets into: %s", assetsPath))
-	err = unzip(params.backpackZipFile, assetsPath)
+	tos := prompts.NewOfflineFileTOS(licenseFilePath)
+	accepted, err := prompts.PromptTOS(tos, out, prompt)
```
Naatan, line 96: Nit: I suggest simplifying this to something like:

```
accepted, err := prompts.PromptOfflineFileTOS(out, prompt)
```

Which can under the hood do what you have here.

cmd/state-offline-installer/runInstall.go
```diff
@@ -77,41 +78,55 @@ func runInstall(out output.Outputer, params *Params) error {
...
 	if err != nil {
-		return errs.Wrap(err, "Error unzipping assets")
+		return errs.Wrap(err, "Error with TOS acceptance")
+	}
+	if !accepted {
+		return locale.NewInputError("tos_not_accepted", "")
```
Naatan, line 101:
```
tos_not_accepted:
  other: You may not create an ActiveState Platform account without accepting the Terms of Service
```

That locale entry doesn't look appropriate for this use-case.

cmd/state-offline-installer/runInstall.go
```diff
@@ -77,41 +78,55 @@ func runInstall(out output.Outputer, params *Params) error {
...
 
+	out.Print(fmt.Sprintf("Stage 2 of 3 Finished: Decompressing artifacts into: %s", artifactsPath))
+
+	out.Print(fmt.Sprintf("Stage 3 of 3 Start: Installing artifacts from: %s", artifactsPath))
 	offlineTarget := target.NewOfflineTarget(installToDir, artifactsPath)
 
-	// FIX: Need to do this more correctly with a New function
-	mockProgress := &OfflineProgressOutput{out: out}
-	// mockProgress := offlineProgressOutput.New(out)
+	mockProgress := newOfflineProgressOutput(out)
```
Naatan, line 129: Why call this `mockProgress`? It's not a mock.

cmd/state-offline-uninstaller/offline-uninstaller.go
```diff
@@ -0,0 +1,91 @@
...
+	cmd := captain.NewCommand(
+		path.Base(os.Args[0]), "", "", p, nil, nil,
+		func(ccmd *captain.Command, args []string) error {
+			out.Print(ccmd.UsageText())
+			return nil
+		},
+	)
+	cmd.AddChildren(
+		captain.NewCommand(
+			cmdUnInstall,
```
Naatan, line 71: As with the installer; the sub-command seems unnecessary. Just use the top-level command.

pkg/cmdlets/auth/signup.go
```diff
@@ -34,7 +28,8 @@ type signupInput struct {
 
 // Signup will prompt the user to create an account
 func Signup(cfg keypairs.Configurable, out output.Outputer, prompt prompt.Prompter, auth *authentication.Auth) error {
-	accepted, err := promptTOS(cfg.ConfigPath(), out, prompt)
+	tos := prompts.NewOnlineTOS()
+	accepted, err := prompts.PromptTOS(tos, out, prompt)
```
Naatan, line 32: As noted on the other use of `PromptTOS()`, this doesn't seem like it needs to live here. This logic can live in the prompts package with a convenient function that consumers like this would call without the need to supply the tos text.

pkg/cmdlets/prompts/TOS.go
```diff
@@ -0,0 +1,97 @@
...
+	TOS
+	tosFilePath string
+}
+
+func NewOfflineFileTOS(tosFilePath string) *offlineFileTOS {
+	return &offlineFileTOS{tosFilePath: tosFilePath}
+}
+
+func (tos *offlineFileTOS) GetTOSText() (string, error) {
+	b, err := os.ReadFile(tos.tosFilePath)
```
Naatan, line 57: This path only exists on dev machines. Once built and shipped it won't know how to find this on users computers.

Instead use embed: https://pkg.go.dev/embed

Naatan (changes requested):

cmd/state-offline-installer/main.go
```diff
@@ -63,52 +61,23 @@ func run() error {
 
 	params := newParams()
 	cmd := captain.NewCommand(
-		path.Base(os.Args[0]), "", "", p, nil, nil,
+		cmdInstall,
```
Naatan, line 64: Note; for clarity this should probably match the name of the executable. eg. "offline-installer".

cmd/state-offline-installer/runInstall.go
```diff
@@ -142,7 +148,17 @@ func runInstall(out output.Outputer, params *Params) error {
 	}
 	out.Print(fmt.Sprintf("Stage 3 of 3 Finished: Installing artifacts from: %s", artifactsPath))
 
-	configureEnvironmentAccepted, err := prompt.Confirm("Setup", "Setup environment for installed project?", &default_boolean_answer)
+	// Copy license file
+	contents, err := assets.ReadFileBytes(TOSPath)
```
Naatan, line 152: https://github.com/ActiveState/cli/pull/1964#discussion_r951973863

cmd/state-offline-installer/runInstall.go
```diff
@@ -78,6 +80,19 @@ func runInstall(out output.Outputer, params *Params) error {
...
+	if err != nil {
+		return errs.Wrap(err, "Error with TOS acceptance")
+	}
+	if !accepted {
+		return locale.NewInputError("offline_tos_not_accepted", "License not accepted")
+	}
+
+	// FIX: This is where I would suggest placing asking the EU for the installation directory
+	//     if installToDir == "" {
+	//     installToDir, err = prompter.Input("Choose installation directory", "Where would you like to install the software?", &installToDir, prompt.InputRequired)
```
Naatan, line 93: Please record this somewhere appropriate (eg. jira). Code comments are not issue trackers.

cmd/state-offline-installer/runInstall.go
Naatan: Also note that when you use embed you shouldn't need to write this file to disk at all. Its bytes will be in memory for you to access.

If we need the license file in the installed dir then that should IMO be part of the runtime itself, not something we bake into it after the fact.

pkg/cmdlets/prompts/TOS.go
```diff
@@ -3,16 +3,18 @@ package prompts
...
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/prompt"
+	"github.com/ActiveState/cli/internal/assets"
 )
 
+const TOSPath = "offline/LICENSE.txt"
```
Naatan, line 16: A license and a TOS are two different things.

Naatan (changes requested):

cmd/state-offline-installer/runInstall.go
```diff
@@ -107,6 +94,15 @@ func runInstall(out output.Outputer, params *Params) error {
 
 	out.Print(fmt.Sprintf("Stage 1 of 3 Finished: Decompressing assets into: %s", assetsPath))
 
+	fmt.Println(licenseFileAssetPath)
```
Naatan, line 97: Debug statement

cmd/state-offline-installer/runInstall.go
```diff
@@ -107,6 +94,15 @@ func runInstall(out output.Outputer, params *Params) error {
...
+	accepted, err := prompts.PromptOfflineLicense(out, prompter, licenseFileAssetPath)
```
Naatan, line 98: I suspect this should be a TOS and not a license prompt (never seen a license prompt). But we can get the UX pass surface that issue.

cmd/state-offline-installer/runInstall.go
```diff
@@ -149,13 +145,9 @@ func runInstall(out output.Outputer, params *Params) error {
 	out.Print(fmt.Sprintf("Stage 3 of 3 Finished: Installing artifacts from: %s", artifactsPath))
 
 	// Copy license file
-	contents, err := assets.ReadFileBytes(TOSPath)
-	if err != nil {
-		return errs.Wrap(err, "Error reading file bytes for TOS")
-	}
-	err = fileutils.WriteFile(licenseFilePath, contents)
+	err = fileutils.CopyFile(licenseFileAssetPath, licenseFilePath)
```
Naatan, line 148: @shaunmlowry I feel like if we want to install runtimes with a license file then the license should be part of the runtime, rather than something we manually account for. Thoughts?

pkg/cmdlets/prompts/License.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return "", errs.Wrap(err, "Unable to open TOS file")
+	}
+
+	return string(b), nil
+}
+
+func PromptOfflineLicense(out output.Outputer, prompt prompt.Prompter, filepath string)(bool,error) {
+    legalText := newOfflineLicense(filepath)
+    return PromptTOS(legalText,out,prompt)
```
Naatan, line 32: PromptTOS is prompting for a license? Again; these are different things. If we want to prompt for a license instead of a TOS then we should not funnel this through TOS specific logic.

I suspect we want a TOS though, and all this license stuff is a red herring.

Naatan (changes requested):

cmd/state-offline-installer/runInstall.go
shaunmlowry: Maybe we'd get there at some point, but trying to get another artifact (the license) through the build planner into the final artifact risks delaying build plan support for installers even further

pkg/cmdlets/prompts/License.go
shaunmlowry: The installer is produced by the service and it's access to the service that requires a TOS.  The License is for the use of the runtime, regardless of how you received it and neither the installer nor runtime require access to the service.

Naatan: Note this particular comment is about intermixing TOS and License in the implementation. Regardless of which one is the right one, a license should not invoke something that's specific to a TOS, or vice versa.

cmd/state-offline-installer/install.go
```diff
@@ -0,0 +1,291 @@
...
+		err = fileutils.WriteFile(filepath.Join(path, "setenv.bat"), contents)
+		if err != nil {
+			return locale.WrapError(err,
+				"err_deploy_write_setenv",
+				"Could not create setenv batch scriptfile at path: {{.V0}}",
+				path)
+		}
+	}
+
+	err = r.shell.WriteUserEnv(r.cfg, env, sscommon.OfflineInstallID, true)
```
Naatan, line 241: Note this was using `sscommon.DeployID` before. I had to introduce a new one for the offline installer. We should not be reusing those ID's as that would create conflicts between these two different methods of "deployment".

cmd/state-offline-installer/install.go
```diff
@@ -0,0 +1,291 @@
...
+			"Could not write environment information to your shell configuration.")
+	}
+
+	binPath := filepath.Join(path, "bin")
+	if err := fileutils.MkdirUnlessExists(binPath); err != nil {
+		return locale.WrapError(err, "err_deploy_binpath", "Could not create bin directory.")
+	}
+
+	// Write global env file
+	err = r.shell.SetupShellRcFile(binPath, env, nil)
```
Naatan, line 254: Note I dropped the namespace param here and refactored `SetupShellRCFile` to not alter the PS1 if no namespace was provided.

Naatan (approved):

icanhasmath (approved):

cmd/state-offline-installer/runInstall.go
Naatan: I'm ok with it. Most of the burden will be on the thing that produces the offline installer payload rather than the installer itself.

PR #1971, opened 2022-07-26
Use user SID when installing State Tool as admin on Windows.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-905" title="DX-905" target="_blank">BugDX-905</a>  _prepare []: prepare error, message: Could not add state protocol, error: Could not create state protocol registry key: The parameter is incorrect.  Stacktrace: D:/a/cli/cli/internal/logging/logging.go:github.com/ActiveState/cli/internal/logging.Error:288
  </td></table>
  <br />
 

The group SID does not exist in the registry, but the user SID does.

### DIFF
--- a/internal/runners/prepare/prepare_windows.go
+++ b/internal/runners/prepare/prepare_windows.go
@@ -108,8 +108,8 @@ func setStateProtocol() error {
 		if err != nil {
 			return locale.WrapError(err, "err_prepare_username", "Could not get current username")
 		}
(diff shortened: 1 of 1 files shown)


Naatan:
@mitchell-as Do you know if this will work both when running as admin and as non-admin?

mitchell-as:
Yes, non-admin uses a different registry key that exists for sure (HKEY_LOCAL_MACHINE\...). Admin is the case I tested.

Naatan (approved):

PR #1972, opened 2022-07-26
Do not error attempting to write completions script.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-904" title="DX-904" target="_blank">BugDX-904</a>  _prepare []: prepare error, message: Could not generate completions script, error received: Writing completion data failed., error: Writing completion data failed: Could not write completions script: os.OpenFile /etc/bash_completion.d/state failed: access
  </td></table>
  <br />
 

It's likely because we don't have admin privileges. This is a minor thing and not worth reporting an error over, especially to Rollbar.


Naatan (commented):
Might be able to address the privilege issue: https://serverfault.com/a/506707

daved (approved):

mitchell-as:
Filed a follow-up story as DX-1079, as it's outside the scope for this bug. We simply want to stop logging it to rollbar.

PR #1973, opened 2022-07-27
Allow user to select from multiple projects/paths with `state use` and `state shell`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1075" title="DX-1075" target="_blank">StoryDX-1075</a>  When using `state use <projectName>` - prompt if multiple projects match
  </td></table>
  <br />

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,15 +5,12 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"path/filepath"
 	"regexp"
 	"runtime"
-	"sort"
 	"strings"
 
 	"github.com/ActiveState/cli/internal/constraints"
 	"github.com/ActiveState/cli/internal/errs"
-	"github.com/ActiveState/cli/internal/installation/storage"
 	"github.com/ActiveState/cli/internal/keypairs"
 	"github.com/ActiveState/cli/internal/language"
 	"github.com/ActiveState/cli/internal/locale"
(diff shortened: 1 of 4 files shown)


pkg/project/project.go
```diff
@@ -20,6 +21,7 @@ import (
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/internal/prompt"
```
daved, line 24: Personally, I don't think we should do any prompting in this package. Currently, there is some prompting for secrets, but that is old code and I suggest that we shouldn't use that to justify adding more dependency on prompting. I suggest moving the prompting into a new package within the runbits directory. Generally, I'd imagine this function returning all projects and then the runbits package prompting based on those options. If that leaves this function fully or near fully useless, then it may be worth trying to remove it unless there are other callers needing it. @Naatan, please jump in if I'm suggesting a poor direction.

daved (changes requested):

pkg/project/project.go
Naatan: I agree, prompt should ideally only be used from runners or runbits. Generally speaking anything that produces output would fall under this rule.

mitchell-as: Committed a new runbits package.

daved (approved):

PR #1974, opened 2022-07-27
New `state checkout` command.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1062" title="DX-1062" target="_blank">StoryDX-1062</a>  state checkout command exists and can check out new projects
  </td></table>
  <br />


internal/runners/checkout/checkout.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	proj, err := project.FromPath(projectDir)
+	if err != nil {
+		return locale.WrapError(err, "err_project_frompath")
+	}
+
+	if params.Branch != "" && proj.BranchName() != params.Branch {
+		return locale.NewInputError("err_conflicting_branch_while_checkedout", "", params.Branch, proj.BranchName())
+	}
```
daved, line 91: Under what conditions can this occur? It might be obvious, but I'm not digging atm.

mitchell-as: I don't think this can happen. It should be removed I think.

daved (approved):

PR #1977, opened 2022-07-27
Use .bash_profile on Mac

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1054" title="DX-1054" target="_blank">BugDX-1054</a>  state not available in path when using bash as default shell on Mac with terminal
  </td></table>
  <br />

### DIFF
--- a/internal/subshell/bash/bash.go
+++ b/internal/subshell/bash/bash.go
@@ -22,6 +22,9 @@ var rcFileName = ".bashrc"
 
 func init() {
 	escaper = osutils.NewBashEscaper()
+	if runtime.GOOS == "darwin" {
+		rcFileName = ".bash_profile"
+	}
 }
 
 // SubShell covers the subshell.SubShell interface, reference that for documentation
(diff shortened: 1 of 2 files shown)


internal/subshell/bash/bash.go
```diff
@@ -22,6 +22,9 @@ var rcFileName = ".bashrc"
 
 func init() {
 	escaper = osutils.NewBashEscaper()
+	if runtime.GOOS == "darwin" {
+		rcFileName = ".bash_profile"
```
mitchell-as, line 26: Nit: maybe comment why .bash_profile is being used instead of .bashrc on macOS.

mitchell-as (approved):

internal/subshell/bash/bash.go
MDrakos: Good point!

mitchell-as (approved):

PR #1978, opened 2022-07-27
Remove local projects if their activestate.yaml file no longer exists or if there is a namespace mismatch.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-198" title="DX-198" target="_blank">BugDX-198</a>  deleting projects should update `state projects`
  </td></table>
  <br />
 

Also stabilize project sorting.

### DIFF
--- a/pkg/projectfile/projectfile.go
+++ b/pkg/projectfile/projectfile.go
@@ -1372,8 +1372,19 @@ func CleanProjectMapping(cfg ConfigGetter) {
 			for namespace, paths := range projects {
 				var removals []int
 				for i, path := range paths {
-					if !fileutils.DirExists(path) {
+					configFile := filepath.Join(path, constants.ConfigFileName)
+					if !fileutils.DirExists(path) || !fileutils.FileExists(configFile) {
 						removals = append(removals, i)
+						continue
+					}
+					// Only remove the project if the activestate.yaml is parseable and there is a namespace
+					// mismatch.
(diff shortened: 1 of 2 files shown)


pkg/projectfile/projectfile.go
```diff
@@ -1372,8 +1372,19 @@ func CleanProjectMapping(cfg ConfigGetter) {
...
+					configFile := filepath.Join(path, constants.ConfigFileName)
+					if !fileutils.DirExists(path) || !fileutils.FileExists(configFile) {
 						removals = append(removals, i)
+						continue
+					}
+					// Only remove the project if the activestate.yaml is parseable and there is a namespace
+					// mismatch.
+					// (We do not want to punish anyone for a syntax error when manually editing the file.)
+					if proj, err := parse(configFile); err == nil && proj.Init() == nil {
+						projNamespace := fmt.Sprintf("%s/%s", proj.Owner(), proj.Name())
```
daved, line 1384: I believe that you can use `proj.Namespace().String()` or `proj.NamespaceString()`.

daved (changes requested):

pkg/projectfile/projectfile.go
mitchell-as: Unfortunately, no. This `Project` is a struct for YAML. It is not the struct for the project we usually expect :(

daved (approved):

PR #1981, opened 2022-07-28
Attempt to fix randomly hanging integration test on Windows CI.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1083" title="DX-1083" target="_blank">BugDX-1083</a>  Nightly integration tests: Checkout randomly times out on Windows.
  </td></table>
  <br />

### DIFF
--- a/test/integration/checkout_int_test.go
+++ b/test/integration/checkout_int_test.go
@@ -71,7 +71,7 @@ func (suite *CheckoutIntegrationTestSuite) TestCheckoutWithFlags() {
 		e2e.WithArgs("checkout", "ActiveState-CLI/Python3#6d9280e7-75eb-401a-9e71-0d99759fbad3", "--path", python3Dir),
 		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
 	)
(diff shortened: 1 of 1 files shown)


test/integration/checkout_int_test.go
```diff
@@ -71,7 +71,7 @@ func (suite *CheckoutIntegrationTestSuite) TestCheckoutWithFlags() {
 		e2e.WithArgs("checkout", "ActiveState-CLI/Python3#6d9280e7-75eb-401a-9e71-0d99759fbad3", "--path", python3Dir),
 		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
 	)
-	cp.ExpectExitCode(0)
+	cp.Expect("Checked out")
```
daved, line 74: Can both the expectexitcode and this call be used successfully?

daved (changes requested):

test/integration/checkout_int_test.go
mitchell-as: Yes, it looks like that works.

daved (approved):

PR #1982, opened 2022-07-28
Error on commit ID mismatch in `state use`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1070" title="DX-1070" target="_blank">BugDX-1070</a>  USE command: `state use <org>/<proj>#<commit id>` not failed if there is checked out project with any other commit id.
  </td></table>
  <br />


Naatan (changes requested):
I think we should error out instead. It's not at all uncommon for people not to read these messages, and then they're in a state that they aren't aware of.

Naatan (approved):

PR #1984, opened 2022-07-28
Fix failing integration tests on macOS

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1086" title="DX-1086" target="_blank">BugDX-1086</a>  Nightly failure: Testdeploy/Prepare failing
  </td></table>
  <br />


test/integration/deploy_int_test.go
```diff
@@ -303,7 +303,12 @@ func (suite *DeployIntegrationTestSuite) AssertConfig(ts *e2e.Session, targetID
 		homeDir, err := os.UserHomeDir()
 		suite.Require().NoError(err)
 
-		bashContents := fileutils.ReadFileUnsafe(filepath.Join(homeDir, ".bashrc"))
+		configFile := ".bashrc"
```
mitchell-as, line 306: Do you think there is a way to call some public function to ask for the config file and that function performs this logic and returns the given file? It seems pretty redundant to have three of these checks (1 in non-test code, and 2 in these tests). If we add another similar test we'd have to remember to do this too.

mitchell-as (changes requested):

test/integration/deploy_int_test.go
MDrakos: Yeah, that makes sense. I've updated it to use the subshell logic which is where the switch between `.bashrc` and `.bash_profile` actually happens. There's still repeated code in these places but we're not required to know any difference between the platforms.

mitchell-as (approved):
This looks a lot better. Kudos!

PR #1986, opened 2022-07-28
Fixed `state use reset --non-interactive`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1082" title="DX-1082" target="_blank">BugDX-1082</a>  USE command: `state use reset -n` not resetting the default project + non message to user printed.
  </td></table>
  <br />
 

Also removed redundant `--force` argument.

Nathan confirmed we can remove the `--force` argument.

### DIFF
--- a/internal/runners/use/reset.go
+++ b/internal/runners/use/reset.go
@@ -42,7 +42,7 @@ func (u *Reset) Run(params *ResetParams) error {
 		return err
 	}
 	if !ok {
-		return locale.NewInputError("err_resert_aborted", "Reset aborted by user")
+		return locale.NewInputError("err_reset_aborted", "Reset aborted by user")
 	}
 
 	reset, err := globaldefault.ResetDefaultActivation(u.subshell, u.config)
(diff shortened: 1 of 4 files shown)


internal/runners/use/reset.go
```diff
@@ -38,15 +35,14 @@ func NewReset(prime primeable) *Reset {
...
-			return nil
-		}
+	defaultChoice := params.Force
+	ok, err := u.prompt.Confirm(locale.T("confirm"),
+		locale.Tl("use_reset_confirm", "You are about to reset your default project runtime. Continue?"), &defaultChoice)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return locale.NewInputError("err_resert_aborted", "Reset aborted by user")
```
daved, line 45:
```suggestion
		return locale.NewInputError("err_reset_aborted", "Reset aborted by user")
```

daved (changes requested):

daved (approved):

PR #1993, opened 2022-08-02
Namespace should be optional for `state shell`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1091" title="DX-1091" target="_blank">StoryDX-1091</a>  `state shell` should work on current dir and global default
  </td></table>
  <br />
 

If no namespace is given, start a shell in the current project or the default project.

### DIFF
--- a/internal/runners/shell/shell.go
+++ b/internal/runners/shell/shell.go
@@ -1,6 +1,8 @@
 package shell
 
 import (
+	"os"
+
 	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/locale"
@@ -58,12 +60,25 @@ func New(prime primeable) *Shell {
 func (u *Shell) Run(params *Params) error {
 	logging.Debug("Shell %v", params.Namespace)
 
-	proj, err := runbitsProject.FromNamespaceLocal(params.Namespace, u.config, u.prompt)
-	if err != nil {
-		if runbitsProject.IsLocalProjectDoesNotExistError(err) {
-			return locale.WrapInputError(err, "err_shell_project_does_not_exist", "Local project does not exist.")
+	var proj *project.Project
+	var err error
+	if params.Namespace.Owner != "" || params.Namespace.Project != "" {
+		proj, err = runbitsProject.FromNamespaceLocal(params.Namespace, u.config, u.prompt)
+		if err != nil {
+			if runbitsProject.IsLocalProjectDoesNotExistError(err) {
+				return locale.WrapInputError(err, "err_shell_project_does_not_exist", "Local project does not exist.")
+			}
+			return locale.WrapError(err, "err_shell", "Unable to run shell")
 		}
-		return locale.WrapError(err, "err_shell", "Unable to run shell")
+	} else {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return locale.WrapInputError(err, "err_shell_getwd_fail", "Cannot determine the current working directory.")
+		}
+		proj, err = project.FromPath(cwd)
+		if err != nil {
+			return locale.WrapInputError(err, "err_shell_cannot_determine_project", "Cannot determine the project to start a shell/prompt in.")
+		}
 	}
 
 	if cid := params.Namespace.CommitID; cid != nil && *cid != proj.CommitUUID() {
--- a/internal/runners/shell/shell.go
+++ b/internal/runners/shell/shell.go
@@ -1,8 +1,6 @@
 package shell
 
 import (
-	"os"
-
 	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/locale"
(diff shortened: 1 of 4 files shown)


internal/runners/shell/shell.go
```diff
@@ -58,12 +60,25 @@ func New(prime primeable) *Shell {
...
+			return locale.WrapError(err, "err_shell", "Unable to run shell")
+		}
+	} else {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return locale.WrapInputError(err, "err_shell_getwd_fail", "Cannot determine the current working directory.")
+		}
+		proj, err = project.FromPath(cwd)
+		if err != nil {
+			return locale.WrapInputError(err, "err_shell_cannot_determine_project", "Cannot determine the project to start a shell/prompt in.")
```
Naatan, line 80: I think you can replace most/all of this with: https://github.com/ActiveState/cli/blob/a2e3ebf27dfe163a69e7172fc4770ef14b71c5bb/pkg/projectfile/projectfile.go#L774

mitchell-as: Awesome, thanks. You saved me some time.

test/integration/shell_int_test.go
```diff
@@ -51,6 +51,31 @@ func (suite *ShellIntegrationTestSuite) TestShell() {
...
+
+	cp = ts.SpawnWithOpts(
+		e2e.WithArgs("shell"),
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+	)
+	cp.Expect("Activated")
+	cp.WaitForInput()
+	cp.SendLine("exit")
+	cp.ExpectExitCode(0)
+}
```
MDrakos, line 77: nit: There could be a few more tests that cover the cases in the ACs. This is testing that it sources the global default but we could also have tests for the current project, namespace, current project when there is currently a global default and potentially more.

MDrakos (commented):
Looks good. Just leaving a comment for my nit and I'm not sure if this will auto merge when I approve.

test/integration/shell_int_test.go
mitchell-as: This behavior hinges on `projectfile.GetProjectFilePath()`, which I'm almost positive is tested as a side-effect in other integration tests like `state activate`. I didn't think the additional explicit tests are worth ballooning the integration test runtime even more. What do you think @Naatan ?

mitchell-as:
Tagging David for review since Mike is out.

internal/runners/shell/shell.go
```diff
@@ -58,12 +59,25 @@ func New(prime primeable) *Shell {
 func (u *Shell) Run(params *Params) error {
 	logging.Debug("Shell %v", params.Namespace)
 
-	proj, err := runbitsProject.FromNamespaceLocal(params.Namespace, u.config, u.prompt)
-	if err != nil {
-		if runbitsProject.IsLocalProjectDoesNotExistError(err) {
-			return locale.WrapInputError(err, "err_shell_project_does_not_exist", "Local project does not exist.")
+	var proj *project.Project
+	var err error
+	if params.Namespace.Owner != "" || params.Namespace.Project != "" {
```
Naatan, line 64:
```suggestion
	if params.Namespace.Project != "" {
```

The owner on its own doesn't seem like it would work.

internal/runners/shell/shell.go
```diff
@@ -58,12 +59,25 @@ func New(prime primeable) *Shell {
...
+		proj, err = runbitsProject.FromNamespaceLocal(params.Namespace, u.config, u.prompt)
+		if err != nil {
+			if runbitsProject.IsLocalProjectDoesNotExistError(err) {
+				return locale.WrapInputError(err, "err_shell_project_does_not_exist", "Local project does not exist.")
+			}
+			return locale.WrapError(err, "err_shell", "Unable to run shell")
```
Naatan, line 70: This is too generic of an error. Plus we weren't attempting to do anything with the shell yet.

internal/runners/shell/shell.go
```diff
@@ -58,12 +59,25 @@ func New(prime primeable) *Shell {
...
+		}
+	} else {
+		projectFile, err := projectfile.GetProjectFilePath()
+		if err != nil {
+			return locale.WrapInputError(err, "err_shell_cannot_determine_project", "Cannot determine the project to start a shell/prompt in.")
+		}
+		proj, err = project.FromPath(projectFile)
```
Naatan, line 77:
```suggestion
		proj, err = project.GetOnce()
```

test/integration/shell_int_test.go
Naatan: I agree this test is a little bare. It'd be nice if we can test the "project from CWD" and "project from namespace" variations. 

While technically this is a little redundant with activate, there is enough plumbing here that they aren't identical.

It would also be good if we can test that it actually is in a subshell. The current test just "waits for input" which doesn't assert whether it is now in a subshell.

Naatan (changes requested):

test/integration/shell_int_test.go
mitchell-as: Okay, I added a cwd test. The `cp.Expect("Activated")` is the assertion that we're in a subshell, and the `cp.SendLine("exit")` and subsequent `cp.ExpectExitCode(0)` indirectly verify we've exited the subshell. (If we were not in a subshell, those lines would fail.)

The "project from namespace" variation tests already exist.

daved (approved):

daved (approved):

mitchell-as:
@Naatan ping. I cannot merge this without your approval overriding a previous change request.

Naatan (approved):

PR #1996, opened 2022-08-04
Add reporting of cli args of caller that starts svc

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1030" title="DX-1030" target="_blank">BugDX-1030</a>  state-svc foreground []: state-svc errored out: Could not start service: Failed to start server: flisten in use
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -103,7 +103,7 @@ func run(cfg *config.Instance) error {
 	if mousetrap.StartedByExplorer() {
 		// Allow starting the svc via a double click
 		captain.DisableMousetrap()
-		return runStart(out, "svc-start mouse")
+		return runStart(out, "svc-start:mouse")
 	}
 
 	p := primer.New(nil, out, nil, nil, nil, nil, cfg, nil, nil, an)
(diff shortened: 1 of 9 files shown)


cmd/state-svc/main.go
```diff
@@ -151,30 +153,41 @@ func run(cfg *config.Instance) error {
 			cmdForeground,
 			"Starting the ActiveState Service",
 			"Start the ActiveState Service (Foreground)",
-			p, nil, nil,
+			p, nil,
+			[]*captain.Argument{
+				{
+					Name:        "Arg text",
+					Description: "Argument text of caller",
```
Naatan, line 160: The description should answer the question "Why should I provide this?".

Naatan (changes requested):
Can we also improve the error of `flisten in use` so that it conveys that this error is partly expected? Otherwise this will continue to look like something that needs fixing.

Naatan (changes requested):
You missed this comment I think: https://github.com/ActiveState/cli/pull/1996#pullrequestreview-1065554573

daved:
@Naatan Sorry about missing that. I also added a test.

cmd/state-svc/service.go
```diff
@@ -58,6 +58,9 @@ func (s *service) Start() error {
 	s.ipcSrv = ipc.NewServer(s.ctx, spath, reqHandlers...)
 	err = s.ipcSrv.Start()
 	if err != nil {
+		if errors.Is(err, ipc.ErrInUse) {
+			return errors.New("An existing server instance appears to be in use")
```
Naatan, line 62: This should still wrap imo.

Naatan (changes requested):

Naatan (approved):

PR #1998, opened 2022-08-05
Explicitly set role for invites.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-946" title="DX-946" target="_blank">BugDX-946</a>  CLI - INVITE: Fail to send invite
  </td></table>
  <br />
 

At the time of this commit, there is a Platform bug attempting to infer a default value for role.


daved (approved):
I'm surprised we permit a user to do anything "as admin", but if that's fine, this is good to go.

mitchell-as:
The Platform performs the necessary checks for us. If you don't have permission, you cannot send an invite, regardless of the request.

mitchell-as:
Dashboard just noted that the 'AsOwner' field has been deprecated in favor of 'Role'.

(As a side note, I've been told that the changes will be backward compatible with existing state tool versions.)

daved (dismissed):

PR #2012, opened 2022-08-08
Added zig-cache to gitignore.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1116" title="DX-1116" target="_blank">TaskDX-1116</a>  Add zig-cache to gitignore
  </td></table>
  <br />
 

I'm targetting it for v0.34.1 even though the executors are in v0.35 because these files will not disappear when switching between branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ obj/
 bin/
 installers/msi-language/Product.wxs
 node_modules
+zig-cache
+


.gitignore
```diff
@@ -14,3 +14,5 @@ obj/
 bin/
 installers/msi-language/Product.wxs
 node_modules
+zig-cache
```
daved, line 17: This is a rather awkward change given the history of the project. This would mean that the zig cache dir is ignored before it exists. The only reason it exists is because we jump around in (release) time. My suggestions in order of preference, but still 100% up to you, are 1. delete the directory, 2. ignore globally, 3. use `**/zig-cache/` as the ignore line.

daved (changes requested):

.gitignore
mitchell-as: I'll go with option 1. Thanks.

PR #2020, opened 2022-08-10
If a project is already checked out, check entire owner/project namespace, not just project.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1120" title="DX-1120" target="_blank">BugDX-1120</a>  Same Project Name In Two Different Organizations With Same Local Checkout Path
  </td></table>
  <br />
 

Otherwise two projects of the same name but with different orgs would pass the test.

### DIFF
--- a/pkg/cmdlets/checkout/path.go
+++ b/pkg/cmdlets/checkout/path.go
@@ -1,6 +1,7 @@
 package checkout
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -32,7 +33,7 @@ func ensureProjectPath(cfg *config.Instance, namespace *project.Namespaced, pref
 	}
 
 	// Validate that target path doesn't contain a config for a different namespace
-	if err := validatePath(namespace.Project, targetPath); err != nil {
+	if err := validatePath(namespace, targetPath); err != nil {
 		return "", errs.Wrap(err, "Could not validate target path: %s", targetPath)
 	}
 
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
This seems like it would just error out rather than use the right path?

If I understand correctly running `state activate owner2/projectName` has the chance of picking up `owner1/projectName`. This change seems to ensure that we end up using the right project or otherwise error out, which his a good change, but we want it to detect the right project so no error happens at all.

Note stole this PR from Mike as I'm waiting on it to draft the v0.34.1 RC.

mitchell-as:
The kicker is when the *projectName/* directory exists in the current working directory. Running `state activate owner2/projectName` when it hasn't ever been activated before, but `state activate owner1/projectName` was run earlier and the *projectName/* folder exists, THIS is when we want to error out with a namespace mismatch, because we cannot checkout `owner2/projectName` for the first time.

This is what was happening in the video.

Naatan:
@mitchell-as I think I understand, so because you ran activate from a directory which already contained a `projectName` directory with a different activestate.yaml, it ended up using that?

In that case I think we are still not giving the right behavior. I would expect to see an error along the lines of "Target directory already contains a project", or I think with our current implementation where we care about empty dirs it should say "Target directory is not empty".

If it just says "Namespace mismatch" then the user is pretty much in the dark on how it got to that error.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2020#issuecomment-1212351907

Naatan (approved):
Looking good :)

mitchell-as:
If there is a project name mismatch, the user will see "The project at the target path does not match the namespace you provided. Expected: <org/project1>, Actual: <org/project2>."

If there is an owner name mismatch, the user will see "Project directory at <path> is not empty. When activating a new project the project directory must be empty."

Naatan (approved):

PR #2021, opened 2022-08-10
Respect GitHub API rate limit

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1123" title="DX-1123" target="_blank">BugDX-1123</a>  Workflow helpers' GitHub API calls are sometimes rate limited
  </td></table>
  <br />

### DIFF
--- a/scripts/internal/workflow-helpers/ghtransport.go
+++ b/scripts/internal/workflow-helpers/ghtransport.go
@@ -1,0 +1,157 @@
+package workflow_helpers
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/google/go-github/v45/github"
+)
+
+/*
+Code based on: https://github.com/hashicorp/terraform-provider-github/blob/fa73654b66e37b1fd8d886141d9c2974e24ba42f/github/transport.go#L42-L109
+Sourced via: https://github.com/google/go-github/issues/431
+
+MIT License
+
+Copyright (c) 2020 GitHub
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


scripts/internal/workflow-helpers/ghtransport.go
```diff
@@ -0,0 +1,157 @@
...
+	writeDelay = 720 * time.Millisecond // https://github.com/google/go-github/issues/431#issuecomment-248767702
+)
+
+// rateLimitTransport implements GitHub's best practices
+// for avoiding rate limits
+// https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits
+type rateLimitTransport struct {
+	transport        http.RoundTripper
+	delayNextRequest bool
+	responseBody     []byte
```
daved, line 53: I think this field is unused in this file.

scripts/internal/workflow-helpers/ghtransport.go
```diff
@@ -0,0 +1,157 @@
...
+		time.Sleep(writeDelay)
+	}
+
+	rlt.delayNextRequest = isWriteMethod(req.Method)
+
+	resp, err := rlt.transport.RoundTrip(req)
+	if err != nil {
+		rlt.unlock(req)
+		return resp, err
+	}
```
daved, line 79: Maybe `defer rlt.unlock(req)` instead.

daved (changes requested):

scripts/internal/workflow-helpers/ghtransport.go
daved: It ensures that the unlock is called. I think there is a return in this function that does not call unlock. It also means that future work does not have to step so carefully.

Naatan: Thing is, this code is taken from https://github.com/hashicorp/terraform-provider-github/blob/fa73654b66e37b1fd8d886141d9c2974e24ba42f/github/transport.go

It was authored about 3 years ago and seems to have been "stable" since then (judging by no further commits alone). Not understanding the full implications of unlocking the req I'd rather use it as is, as there might've been a reason for it.

Add to that the fact that this is only for our automations and not for any code we ship, I feel like it's safer to leave it as is.

daved (approved):

PR #2024, opened 2022-08-11
Build a more portable state-exec on Linux

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1126" title="DX-1126" target="_blank">BugDX-1126</a>  SHELL/ACTIVATE: Linux - any attempt to activate the project `use` or `activate` we have error message `Illegal instruction (core dumped)`
  </td></table>
  <br />

### DIFF
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -104,6 +104,10 @@ scripts:
       $constants.SET_ENV
 
       go build -tags "$GO_BUILD_TAGS" -o $BUILD_TARGET_DIR/$constants.BUILD_TARGET $constants.CLI_BUILDFLAGS $constants.CLI_PKGS
+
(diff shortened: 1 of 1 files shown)


Naatan:
I just said this as speculative on Slack, we should probably discuss/investigate before we action removing UPX. Though removing it for state-exec definitely makes sense as we had never evaluated it for zig.

mitchell-as:
I'm only testing to see if CI is to blame here.

daved (changes requested):
Let's also be explicit with the Mac build: `-target "x86_64-macosx"`

mitchell-as:
Good call.

daved (approved):

PR #2025, opened 2022-08-12
Fixed inability to run `state deploy uninstall` from cwd on Windows.

mitchell-as:
<table>
<td>
  </td></table>
  <br />
 

Windows doesn't let you remove the dir as its currently "in use" by the shell/prompt.

### DIFF
--- a/internal/runners/deploy/uninstall/uninstall.go
+++ b/internal/runners/deploy/uninstall/uninstall.go
@@ -2,6 +2,7 @@ package uninstall
 
 import (
 	"os"
+	"path"
 	"runtime"
 
 	"github.com/ActiveState/cli/internal/config"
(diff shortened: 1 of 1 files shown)


internal/runners/deploy/uninstall/uninstall.go
```diff
@@ -61,6 +62,15 @@ func (u *Uninstall) Run(params *Params) error {
...
 		path = cwd
+		if runtime.GOOS == "windows" {
+			err = os.Chdir(filepath.Dir(path)) // cannot os.RemoveAll() in current working directory, so leave it
+			if err != nil {
+				return locale.WrapInputError(
+					err,
+					"err_deploy_uninstall_cannot_chdir",
+					"Cannot switch away from current working directory. Please supply `--path` argument")
+			}
+		}
```
MDrakos, line 73: Should this be an error message rather than us changing the directory for the user? Usually, on Windows, I would expect a "directory in use" error message rather than a program handling it for me.

mitchell-as: I don't think so, because `state deploy <org/namespace>` without any additional arguments deploys to the current working directory. I think it would be nice to allow for the reverse, hence the need to do this on Windows.

MDrakos (approved):

PR #2029, opened 2022-08-12
Update gopsutil pkg

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1124" title="DX-1124" target="_blank">BugDX-1124</a>  TestUpdateToCurrent failure in nightly
  </td></table>
  <br />
 

Nathan edit:
Due to the random nature of the issue this fix is entirely experimental, success will have to be measured by observing CI results going forward.


Naatan:
@daved can you explain how this addresses the CI failure?

daved:
There have been changes to the dependency that may help here. It's not clear that this is something we can solve easily, so this solution may not work. I suggest we land this, then try something more thorough if it comes up again.

Naatan (approved):

PR #2030, opened 2022-08-15
Print state tool update version during an update to the latest version.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1099" title="DX-1099" target="_blank">StoryDX-1099</a>  UPDATE: When run `state update` no version number provided like it done in install flow.
  </td></table>
  <br />

### DIFF
--- a/internal/runners/update/update.go
+++ b/internal/runners/update/update.go
@@ -64,6 +64,10 @@ func (u *Update) Run(params *Params) error {
 		return nil
 	}
 
+	if params.Version == "" {
+		u.out.Notice(locale.Tl("updating_version", "Updating State Tool to version {{.V0}}", up.Version))
+	}
+
 	// Handle switching channels
 	var installPath string
 	if params.Channel != "" && params.Channel != constants.BranchName {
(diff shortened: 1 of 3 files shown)


internal/runners/update/update.go
```diff
@@ -64,6 +64,10 @@ func (u *Update) Run(params *Params) error {
 		return nil
 	}
 
+	if params.Version == "" {
+		u.out.Notice(locale.Tl("updating_version", "Updating State Tool to version {{.V0}}", up.Version))
+	}
```
MDrakos, line 69: I think we need to update the output on line 49 above rather than add another `Updating` message

MDrakos (changes requested):

internal/runners/update/update.go
mitchell-as: I thought about that, but if the update check takes some time, the user is left "hanging" and wondering what is going on until we get a version to print to the output. Or is this a non-issue?

mitchell-as: Never mind, you were on the right track. I've pushed an update to our messaging around updates.

MDrakos (approved):
Looks good!

PR #2035, opened 2022-08-17
Raise an error for invalid boolean or integer config values.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1151" title="DX-1151" target="_blank">BugDX-1151</a>  `state config set` accepts invalid value
  </td></table>
  <br />
 

Do not silently default to "false" or 0.

### DIFF
--- a/internal/runners/config/set.go
+++ b/internal/runners/config/set.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/errs"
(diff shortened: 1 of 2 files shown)


internal/runners/config/set.go
```diff
@@ -38,9 +38,17 @@ func (s *Set) Run(params SetParams) error {
 	}
 	switch option.Type {
 	case configMediator.Bool:
-		value = cast.ToBool(params.Value)
+		var err error
+		value, err = strconv.ParseBool(params.Value)
```
MDrakos, line 42: The `cast` library also has a `ToBoolE` function but since params.Value is not an interface I'm not sure we would get any benefit from using it.

MDrakos (approved):

PR #2038, opened 2022-08-17
Renamed `state checkout` to `state get` and make --path a positional argument

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1142" title="DX-1142" target="_blank">StoryDX-1142</a>  There is a `state get` command
  </td></table>
  <br />


cmd/state/internal/cmdtree/get.go
```diff
@@ -0,0 +1,46 @@
...
+
+func newGetCommand(prime *primer.Values) *captain.Command {
+	params := &get.Params{
+		Namespace: &project.Namespaced{AllowOmitOwner: true},
+	}
+
+	cmd := captain.NewCommand(
+		"get",
+		"",
+		locale.Tl("get_description", "Checkout the given project and setup its runtime"),
```
MDrakos, line 19: Should this be `get` the given project?

MDrakos (approved):
LGTM!

cmd/state/internal/cmdtree/get.go
mitchell-as: No, I asked about that, but Nathan said to keep the current messaging that uses "checkout".

PR #2039, opened 2022-08-17
State switch

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1144" title="DX-1144" target="_blank">StoryDX-1144</a>  We have a `state switch` command
  </td></table>
  <br />


cmd/state/internal/cmdtree/branch.go
```diff
@@ -5,6 +5,7 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runners/branch"
+	"github.com/ActiveState/cli/internal/runners/swtch"
```
mitchell-as, line 8: You're using 'switch' everywhere else, so why 'swtch'? There doesn't appear to be a name collision as far as I can see...

cmd/state/internal/cmdtree/branch.go
```diff
@@ -47,25 +48,28 @@ func newBranchAddCommand(prime *primer.Values) *captain.Command {
...
+				Name:        locale.Tl("switch_arg_identifier", "identifier"),
+				Description: locale.Tl("switch_arg_identifier_description", "The commit or branch to switch to"),
+				Value:       &params.Identifier,
 				Required:    true,
 			},
 		},
 		func(_ *captain.Command, _ []string) error {
 			return runner.Run(params)
 		})
+	cmd.SetHidden(true)
```
mitchell-as, line 72: Do you mind adding a comment as to why this is hidden?

internal/runners/swtch/switch.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return locale.WrapError(err, "err_refresh_runtime")
+	}
+
+	s.out.Print(locale.Tl("branch_switch_success", "Successfully switched to {{.V0}}: [NOTICE]{{.V1}}[/RESET]", idType, params.Identifier))
+
+	return nil
+}
+
+func resolveIdentifierCommitID(project *mono_models.Project, idParam string) (identifier, error) {
```
mitchell-as, line 113: What about tags? I'm not familiar enough with the model to know if the JIRA ticket's AC of commit -> branch -> tag is being honored. The first two are being done for sure. The last I don't know about.

internal/uuidutils/uuidutils.go
```diff
@@ -0,0 +1,19 @@
+package uuidutils
```
mitchell-as, line 1: Do we really need this when we have `strfmt.IsUUID()`? https://pkg.go.dev/github.com/go-openapi/strfmt#IsUUID. This may be an opportunity for refactoring.

test/integration/switch_int_test.go
```diff
@@ -0,0 +1,111 @@
+package integration
```
mitchell-as, line 1: Please also add a test for tags if that's not already covered by the branch or commitID tests.

mitchell-as (changes requested):
The idea to use an interface for getting the commit ID from objects derived from either the given UUID or branch name is quite clever. Kudos!

cmd/state/internal/cmdtree/branch.go
MDrakos: `switch` is a keyword in go so you cannot name a package `switch`

internal/runners/swtch/switch.go
MDrakos: As per our discussion during stand yesterday. We are removing tags from this story and will follow up here: https://activestatef.atlassian.net/browse/DX-1159

test/integration/switch_int_test.go
MDrakos: See my comment above re: tags

cmd/state/internal/cmdtree/branch.go
mitchell-as: lol, oops

internal/runners/swtch/switch.go
```diff
@@ -105,18 +111,17 @@ func (s *Switch) Run(params SwitchParams) error {
 		return locale.WrapError(err, "err_refresh_runtime")
 	}
 
-	s.out.Print(locale.Tl("branch_switch_success", "Successfully switched to {{.V0}}: [NOTICE]{{.V1}}[/RESET]", idType, params.Identifier))
+	s.out.Print(locale.Tl("branch_switch_success", "Successfully switched to {{.V0}}: [NOTICE]{{.V1}}[/RESET]", identifier.Type(), params.Identifier))
```
mitchell-as, line 114: Nice change!

mitchell-as (approved):

cmd/state/internal/cmdtree/switch.go
```diff
@@ -0,0 +1,36 @@
...
+				Description: locale.Tl("switch_arg_identifier_description", "The commit or branch to switch to"),
+				Value:       &params.Identifier,
+				Required:    true,
+			},
+		},
+		func(_ *captain.Command, _ []string) error {
+			return runner.Run(params)
+		})
+
+	cmd.SetGroup(PlatformGroup)
```
Naatan, line 33: I think the environment management group would be more appropriate.

internal/runners/swtch/switch.go
```diff
@@ -0,0 +1,133 @@
...
+	primer.Outputer
+	primer.Projecter
+	primer.Configurer
+	primer.Analyticer
+	primer.SvcModeler
+}
+
+type identifier interface {
+	CommitID() strfmt.UUID
+	Type() string
```
Naatan, line 42: It seems you are only using this for localization, so I recommend naming it Locale() and passing it through the locale package.

Otherwise this should probably be either more strongly typed or have constants to compare the value to as we don't want to be doing manual string matching.

internal/runners/swtch/switch.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return nil
+}
+
+func resolveIdentifierCommitID(project *mono_models.Project, idParam string) (identifier, error) {
+	var resolveErr error
+	if strfmt.IsUUID(idParam) {
+		return commitIdentifier{strfmt.UUID(idParam)}, nil
+	}
+	resolveErr = locale.NewError("invalid_uuid_val", "Invalid UUID {{.V0}} value.", idParam)
```
Naatan, line 124: This doesn't seem to be used.

internal/runners/swtch/switch.go
```diff
@@ -0,0 +1,133 @@
...
+
+	branch, err := model.BranchForProjectByName(project, idParam)
+	if err == nil {
+		return branchIdentifier{branch: branch}, nil
+	}
+	resolveErr = locale.WrapError(err, "err_identifier_branch", "Project does not have a branch named {{.V0}}", idParam)
```
Naatan, line 130: Can we flip this around? We almost always check `err != nil`, rather than the inverse. Feels error-prone to have it this way as it makes grokking more difficult.

test/integration/switch_int_test.go
```diff
@@ -0,0 +1,111 @@
...
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	err := ts.ClearCache()
+	suite.Require().NoError(err)
+
+	suite.PrepareActiveStateYAML(ts, "ActiveState-CLI", "Branches")
+	pjfilepath := filepath.Join(ts.Dirs.Work, constants.ConfigFileName)
+
+	cp := ts.SpawnWithOpts(e2e.WithArgs("pull"), e2e.AppendEnv(constants.DisableRuntime+"=false"))
```
Naatan, line 32: I don't think you need the runtime to test this?

Naatan: Also can we get rid of the pull and just bake the commitID into the test? For tests it's better that way, costs less time and gives fewer things that can break. Right now you could implicitly break the test by adding a new commit to the project.

Naatan (changes requested):
Looks great! Just minor points from me.

activestate.yaml
```diff
@@ -148,26 +149,32 @@ scripts:
     description: Builds the State Executor application
     language: bash
     value: |
-      zig build-exe --strip -static -O ReleaseFast -target "x86_64-linux" "./cmd/${BUILD_EXEC_TARGET}/${BUILD_EXEC_TARGET}.zig"
-      mv "$BUILD_EXEC_TARGET" "${BUILD_TARGET_PREFIX_DIR}"
+      set -e
+      $constants.SET_ENV
 
+      zig build-exe --strip -static -O ReleaseFast -target "x86_64-linux" "./cmd/${BUILD_EXEC_TARGET}/${BUILD_EXEC_TARGET}.zig"
+      mv "$BUILD_EXEC_TARGET" "$constants.BUILD_TARGET_PREFIX_DIR"
```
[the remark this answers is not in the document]
MDrakos: No, but I couldn't run `state run build-all` locally without these fixes

test/integration/switch_int_test.go
MDrakos: We're using the runtime in order to ensure that it is refreshed. We don't test the runtime, just the output associated with refreshing it. I don't believe that mechanism will give us the desired output without both enabling the runtime and the initial pull to install the runtime files.

Naatan: I could definitely be wrong here, but I think you can safely disable it.

While I agree it would be good to test that the runtime did in fact change it is also quite costly in terms of how long it takes to test. Seeing as we already have tests for this mechanic I would advocate taking it out of this one. 

Perhaps we can set up something for testing purposes that allows us to assert this without the cost of setting up a full runtime. I've filed a story to investigate: https://activestatef.atlassian.net/browse/DX-1173

Naatan (changes requested):
Looks good! But slight followup to https://github.com/ActiveState/cli/pull/2039#discussion_r950393058

Naatan (approved):

PR #2043, opened 2022-08-18
State export env

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1148" title="DX-1148" target="_blank">StoryDX-1148</a>  There is a command to easily export the location of my python interpreter (eg. state export env)
  </td></table>
  <br />

### DIFF
--- a/internal/runners/export/apikey.go
+++ b/internal/runners/export/apikey.go
@@ -4,7 +4,6 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/output"
-	"github.com/ActiveState/cli/internal/primer"
 )
 
 // APIKeyProvider describes the behavior required to obtain a new api key.
@@ -35,15 +34,6 @@ func prepareAPIKeyRunParams(params APIKeyRunParams) (APIKeyRunParams, error) {
 type APIKey struct {
 	keyPro APIKeyProvider
 	out    output.Outputer
-}
-
-type primeable interface {
-	primer.Auther
-	primer.Outputer
-	primer.Configurer
-	primer.Projecter
-	primer.Analyticer
-	primer.SvcModeler
 }
 
 // NewAPIKey is a convenience construction function.
--- a/internal/runners/export/apikey.go
+++ b/internal/runners/export/apikey.go
@@ -42,6 +42,8 @@ type primeable interface {
 	primer.Outputer
 	primer.Configurer
 	primer.Projecter
+	primer.Analyticer
+	primer.SvcModeler
 }
 
 // NewAPIKey is a convenience construction function.
--- a/internal/runners/export/env.go
+++ b/internal/runners/export/env.go
@@ -1,10 +1,7 @@
 package export
 
 import (
-	"strings"
-
 	"github.com/ActiveState/cli/internal/analytics"
-	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/runbits/runtime"
@@ -35,15 +32,6 @@ func NewEnv(prime primeable) *Env {
 func (e *Env) Run() error {
 	rt, _, err := runtime.NewFromProject(e.project, target.TriggerActivate, e.analytics, e.svcModel, e.out, e.auth)
 	if err != nil {
-		if errs.Matches(err, &model.ErrNoMatchingPlatform{}) {
-			branches, err := model.BranchNamesForProjectFiltered(e.project.Owner(), e.project.Name(), e.project.BranchName())
-			if err == nil && len(branches) > 1 {
-				return locale.NewInputError("err_alternate_branches", "", e.project.BranchName(), strings.Join(branches, "\n - "))
-			}
-		}
-		if !authentication.LegacyGet().Authenticated() {
-			return locale.WrapError(err, "err_export_env_auth", "Could not update runtime files. If this is a private project ensure that you are authenticated.")
-		}
 		return locale.WrapError(err, "err_export_new_runtime", "Could not get new runtime")
 	}
 
@@ -52,9 +40,7 @@ func (e *Env) Run() error {
 		return locale.WrapError(err, "err_env_get_env", "Could not get runtime environment")
 	}
 
-	for k, v := range env {
-		e.out.Print((locale.Tl("env_output_env", "{{.V0}}: {{.V2}}", k, v)))
-	}
+	e.out.Print(env)
 
 	return nil
 }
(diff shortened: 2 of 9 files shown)


internal/runners/export/apikey.go
```diff
@@ -42,6 +42,8 @@ type primeable interface {
 	primer.Outputer
 	primer.Configurer
 	primer.Projecter
+	primer.Analyticer
```
mitchell-as, line 45: Since `primeable` is now used elsewhere in this package, please hoist it out into a more generic file, maybe export.go? (Not sure exactly).

test/integration/export_int_test.go
```diff
@@ -78,6 +79,22 @@ func (suite *ExportIntegrationTestSuite) TestExport_Config() {
...
+	defer ts.Close()
+
+	suite.PrepareActiveStateYAML(ts)
+	asyData := fmt.Sprintf(`project: "https://platform.activestate.com/ActiveState-CLI/Export?branch=main&commitID=5397f645-da8a-4591-b106-9d7fa99545fe"`)
+	ts.PrepareActiveStateYAML(asyData)
+	cp := ts.SpawnWithOpts(
+		e2e.WithArgs("export", "env"),
+		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+	)
+	cp.Expect(`PATH: `)
```
mitchell-as, line 94: Please consider adding an additional test(s) that asserts `ACTIVESTATE_ACTIVATED` is not present, since one of the ACs is that state tool env vars should not be present. (e.g. `suite.Assert().NotContains(cp.TrimmedSnapshot(), "ACTIVESTATE_ACTIVATED")`

mitchell-as (changes requested):
Yay, an integration test! Just a couple of minor things, but this looks good.

mitchell-as (approved):

internal/runners/export/env.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (e *Env) Run() error {
+	rt, _, err := runtime.NewFromProject(e.project, target.TriggerActivate, e.analytics, e.svcModel, e.out, e.auth)
+	if err != nil {
+		if errs.Matches(err, &model.ErrNoMatchingPlatform{}) {
+			branches, err := model.BranchNamesForProjectFiltered(e.project.Owner(), e.project.Name(), e.project.BranchName())
+			if err == nil && len(branches) > 1 {
+				return locale.NewInputError("err_alternate_branches", "", e.project.BranchName(), strings.Join(branches, "\n - "))
+			}
```
Naatan, line 42: Can we move this into the runtime package? Seems like a useful check to have on all runtime usage. And having it implemented in two places like this is going to lead to inconsistencies.

internal/runners/export/env.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		if !authentication.LegacyGet().Authenticated() {
```
Naatan, line 44: We shouldn't introduce any new code that uses legacy methods. You are already passing the auth struct via the primer, please use that instead.

internal/runners/export/env.go
```diff
@@ -0,0 +1,60 @@
...
+			return locale.WrapError(err, "err_export_env_auth", "Could not update runtime files. If this is a private project ensure that you are authenticated.")
```
Naatan, line 45: This also seems like something we can move into the runtime package.

internal/runners/export/env.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		return locale.WrapError(err, "err_export_new_runtime", "Could not get new runtime")
```
Naatan, line 47:
```suggestion
		return locale.WrapError(err, "err_export_new_runtime", "Could not initialize runtime")
```

Keep in mind locale errors are user-facing errors.

internal/runners/export/env.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	env, err := rt.Env(false, true)
+	if err != nil {
+		return locale.WrapError(err, "err_env_get_env", "Could not get runtime environment")
+	}
+
+	for k, v := range env {
+		e.out.Print((locale.Tl("env_output_env", "{{.V0}}: {{.V2}}", k, v)))
```
Naatan, line 56: This isn't compatible with json output. Print a struct instead, eg. https://github.com/ActiveState/cli/blob/b64c36f1f85ab916831b79aa43a8bc7499ca00c0/internal/runners/show/show.go#L239

Naatan (changes requested):

internal/runners/export/env.go
MDrakos: Oh right! Thanks for pointing this out.

internal/runners/export/env.go
```diff
@@ -35,26 +32,15 @@ func NewEnv(prime primeable) *Env {
...
-			branches, err := model.BranchNamesForProjectFiltered(e.project.Owner(), e.project.Name(), e.project.BranchName())
-			if err == nil && len(branches) > 1 {
-				return locale.NewInputError("err_alternate_branches", "", e.project.BranchName(), strings.Join(branches, "\n - "))
-			}
-		}
-		if !authentication.LegacyGet().Authenticated() {
-			return locale.WrapError(err, "err_export_env_auth", "Could not update runtime files. If this is a private project ensure that you are authenticated.")
-		}
-		return locale.WrapError(err, "err_export_new_runtime", "Could not get new runtime")
+		return locale.WrapError(err, "err_export_new_runtime", "Could not initialize new runtime")
```
Naatan, line 35:
```suggestion
		return locale.WrapError(err, "err_export_new_runtime", "Could not initialize  runtime")
```

This wouldn't be a new runtime. I guess you're referring to `NewFromProject()` which refers to the struct it's initializing; not something users need to be concerned with.

internal/runners/export/env.go
```diff
@@ -35,26 +32,15 @@ func NewEnv(prime primeable) *Env {
...
 
 	env, err := rt.Env(false, true)
 	if err != nil {
 		return locale.WrapError(err, "err_env_get_env", "Could not get runtime environment")
 	}
 
-	for k, v := range env {
-		e.out.Print((locale.Tl("env_output_env", "{{.V0}}: {{.V2}}", k, v)))
-	}
+	e.out.Print(env)
```
Naatan, line 43: Please make sure you test that this looks as expected with the plain outputter, I don't think we designed the env struct with this in mind (but it might work anyway!).

Naatan (changes requested):

internal/runners/export/env.go
MDrakos: Yep, I've tested it. Here's a sample of the printout:

```
PATH: /Users/mikedrakos/Library/Caches/activestate/05c7ff6d/exec 
PYTHONIOENCODING: utf-8 
PYTHONPATH: /Users/mikedrakos/Library/Caches/activestate/05c7ff6d/_runtime_temp/d0c09e08-e89e-5253-8305-af6770d64992/ActivePython-3.6.6.0000-macosx10.9-x86_64-ae178372/Library/Frameworks/Python.framework/Versions/Current/lib/python3.6/site-packages 
```

Naatan (approved):

PR #2044, opened 2022-08-19
Handle multiple e-mail addresses passed as separate command line arguments

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1134" title="DX-1134" target="_blank">BugDX-1134</a>  INVITE: Only first user gets invited
  </td></table>
  <br />
 

I chose to handle additional e-mail address arguments instead of raising an error. If we want to force e-mails to be a single CSV argument, we can do that. Let me know.

### DIFF
--- a/cmd/state/internal/cmdtree/invite.go
+++ b/cmd/state/internal/cmdtree/invite.go
@@ -31,7 +31,7 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
 		},
 		[]*captain.Argument{
 			{
-				Name:        "email1[,email2,..]",
+				Name:        "email1,[email2,..]",
 				Description: locale.Tl("invite_arg_emails", "Email addresses to send the invitations to"),
 				Required:    true,
 				Value:       &params.EmailList,
@@ -40,5 +40,5 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
 		func(ccmd *captain.Command, _ []string) error {
 			return inviteRunner.Run(&params)
 		},
-	).SetGroup(PlatformGroup).SetUnstable(true)
+	).SetGroup(PlatformGroup).SetUnstable(true).SetHasVariableArguments()
 }
--- a/cmd/state/internal/cmdtree/invite.go
+++ b/cmd/state/internal/cmdtree/invite.go
@@ -31,7 +31,7 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
 		},
 		[]*captain.Argument{
 			{
-				Name:        "email1,[email2,..]",
+				Name:        "email1[,email2,..]",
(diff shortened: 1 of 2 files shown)


cmd/state/internal/cmdtree/invite.go
```diff
@@ -31,13 +34,18 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
...
-				Name:        "email1,[email2,..]",
+				Name:        "email1[,email2,..]",
 				Description: locale.Tl("invite_arg_emails", "Email addresses to send the invitations to"),
 				Required:    true,
 				Value:       &params.EmailList,
 			},
 		},
-		func(ccmd *captain.Command, _ []string) error {
+		func(ccmd *captain.Command, args []string) error {
+			if len(args) > 1 {
```
Naatan, line 44: We should maybe also guard against people giving both comma-separated emails as well as space separated (I'd just error out if they do this).

cmd/state/internal/cmdtree/invite.go
```diff
@@ -31,13 +34,18 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
...
 			{
-				Name:        "email1,[email2,..]",
+				Name:        "email1[,email2,..]",
 				Description: locale.Tl("invite_arg_emails", "Email addresses to send the invitations to"),
 				Required:    true,
 				Value:       &params.EmailList,
 			},
 		},
-		func(ccmd *captain.Command, _ []string) error {
+		func(ccmd *captain.Command, args []string) error {
```
Naatan, line 43: I am going back and forth on this but overall I think I'd personally prefer having this logic in the runner rather than the cmdtree. The idea is that cmdtree is all about command registration, and the runner is all about running the command.

Admittedly arg parsing does feel like a bit of a grey area. Perhaps worth for us to discuss during stand or weekly and have a stronger opinion on going forward.

Naatan (changes requested):

mitchell-as:
During stand Pete said just error out when there are multiple command line arguments. My initial guess to handle them turned out to be wrong.

Naatan (changes requested):
I think we got some wires crossed during stand. My takeaway was that we DO want to support space separation, but we want to error out if the user gives a mixture of both comma-separated as well as space-separated.

cmd/state/internal/cmdtree/invite.go
mitchell-as: I think you're right. I'll move it.

mitchell-as:
Okay, sorry for the confusion. Hopefully this addresses our final resolution.

Naatan (approved):

PR #2045, opened 2022-08-19
Sanitize e-mail list a bit to avoid sending empty e-mail addresses to the Platform

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1127" title="DX-1127" target="_blank">BugDX-1127</a>  INVITE: Using comma results in strange error
  </td></table>
  <br />

### DIFF
--- a/internal/runners/invite/invite.go
+++ b/internal/runners/invite/invite.go
@@ -63,7 +63,8 @@ func (i *invite) Run(params *Params) error {
 		}
 	}
 
-	emails := strings.Split(params.EmailList, ",")
+	emailList := strings.Trim(strings.ReplaceAll(params.EmailList, ",,", ","), ",")
+	emails := strings.Split(emailList, ",")
 
 	if err := org.CanInvite(len(emails)); err != nil {
 		return locale.WrapError(err, "err_caninvite", "Cannot invite users to {{.V0}}.", org.String())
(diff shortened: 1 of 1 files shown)


internal/runners/invite/invite.go
```diff
@@ -63,7 +64,9 @@ func (i *invite) Run(params *Params) error {
 		}
 	}
 
-	emails := strings.Split(params.EmailList, ",")
+	re := regexp.MustCompile(",,+")
```
MDrakos, line 67: Nit: even though you only use the regex on the next line it might be nice to name it something more descriptive

internal/runners/invite/invite.go
```diff
@@ -63,7 +64,9 @@ func (i *invite) Run(params *Params) error {
...
+	emailList := strings.Trim(re.ReplaceAllString(params.EmailList, ","), ",")
+	emails := strings.Split(emailList, ",")
```
MDrakos, line 69: Would iterating through `emails` and removing any blank entries cover more potential combinations than the regex?

MDrakos (dismissed):
Makes sense to me. Just a couple of comments to think about

internal/runners/invite/invite.go
mitchell-as: That's fair. I will change it.

internal/runners/invite/invite.go
mitchell-as: I don't think so. We're splitting on ',' so the only way to get a blank e-mail is if there is a leading, trailing, or double ','. All of those cases have been covered.

MDrakos (approved):

PR #2048, opened 2022-08-19
Don't bother trying to uninstall deployment in cwd on Windows.

mitchell-as:
<table>
<td>
  </td></table>
  <br />
 

Just tell the user to cd elsewhere and supply a path argument.

The `os.Chdir()` method worked in git bash, but did not work in cmd.exe. No matter what I try, a directory cannot be removed out from under cmd.exe.

### DIFF
--- a/internal/runners/deploy/uninstall/uninstall.go
+++ b/internal/runners/deploy/uninstall/uninstall.go
@@ -56,7 +56,6 @@ func (u *Uninstall) Run(params *Params) error {
 		if runtime.GOOS == "windows" {
 			if err != nil {
 				return locale.NewInputError(
-					err,
 					"err_deploy_uninstall_cannot_chdir",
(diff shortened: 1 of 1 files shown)


internal/runners/deploy/uninstall/uninstall.go
```diff
@@ -54,6 +53,11 @@ func (u *Uninstall) Run(params *Params) error {
 
 	path := params.Path
 	if path == "" {
+		if runtime.GOOS == "windows" {
+			return locale.NewInputError(
+				"err_deploy_uninstall_cannot_chdir",
+				"Cannot remove deployment in current working directory. Please cd elsewhere and then supply `--path` argument")
```
Naatan, line 59:
```suggestion
				"Cannot remove deployment in current working directory. Please cd elsewhere and run this command again with the '--path' flag.")
```

Naatan (changes requested):

Naatan (approved):

PR #2053, opened 2022-08-22
Ask installed state tool for its version in order to check if update flow should be performed.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1096" title="DX-1096" target="_blank">StoryDX-1096</a>  Running installer should update if installed state tool is older
  </td></table>
  <br />
 

Calling the updater routine was only doing so for the currently running installer (which is up to date), not the install state tool.

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -477,6 +477,9 @@ func noArgs() bool {
 	return len(os.Args[1:]) == 0
 }
 
+var branchRegex = regexp.MustCompile("Branch (\\S+)")
+var versionRegex = regexp.MustCompile("Version (\\S+)")
+
 func shouldUpdateInstalledStateTool() bool {
 	logging.Debug("Checking if installed state tool is an older version.")
 
@@ -489,7 +492,6 @@ func shouldUpdateInstalledStateTool() bool {
 		return true // probably corrupted install
 	}
 
-	branchRegex := regexp.MustCompile("Branch (\\S+)")
 	branchMatch := branchRegex.FindSubmatch(output)
 	if len(branchMatch) == 0 {
 		logging.Debug("Could not read state tool branch.")
@@ -500,7 +502,6 @@ func shouldUpdateInstalledStateTool() bool {
 		return false // do not update, require --force
 	}
 
-	versionRegex := regexp.MustCompile("Version (\\S+)")
 	versionMatch := versionRegex.FindSubmatch(output)
 	if len(versionMatch) == 0 {
 		logging.Debug("Could not read state tool version.")
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -3,7 +3,9 @@ package main
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime/debug"
 	"strings"
 	"time"
@@ -28,7 +30,6 @@ import (
 	"github.com/ActiveState/cli/internal/rollbar"
 	"github.com/ActiveState/cli/internal/runbits/panics"
 	"github.com/ActiveState/cli/internal/subshell"
-	"github.com/ActiveState/cli/internal/updater"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/ActiveState/cli/pkg/sysinfo"
 )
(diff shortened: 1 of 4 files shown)


cmd/state-installer/cmd.go
```diff
@@ -482,3 +476,40 @@ func determineLegacyUpdate(stateToolInstalled bool, packagedStateExe, payloadPat
...
+	stateExe := constants.StateCmd + exeutils.Extension
+	cmd := exec.Command(stateExe, "--version")
+	cmd.Env = os.Environ()
+	output, err := cmd.Output()
+	if err != nil {
+		logging.Debug("Could not determine state tool version.")
+		return true // probably corrupted install
+	}
+
+	branchRegex := regexp.MustCompile("Branch (\\S+)")
```
daved, line 492: Please move this to the global space so that it trips at startup rather than at this point in the control flow. Generally, "must" functions are best defined at a point where they will trip as soon as possible.

daved (changes requested):

cmd/state-installer/cmd.go
mitchell-as: Good to know, thanks.

daved (dismissed):

cmd/state-installer/cmd.go
```diff
@@ -260,17 +261,10 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
...
-		up, err := checker.CheckFor("", "")
-		if err != nil {
-			logging.Debug("Could not check for update.") // not a show-stopper; move on
-		}
-		if up != nil {
-			logging.Debug("Update available. Switching to update flow.")
-			params.isUpdate = true
-		}
+	// If the state tool is already installed, but out of date, continue with update flow.
+	if stateToolInstalled && !params.isUpdate && !params.force && shouldUpdateInstalledStateTool() {
```
Naatan, line 265:
```suggestion
	if stateToolInstalled && ((!params.isUpdate && !params.force) || shouldUpdateInstalledStateTool()) {
```

Note also filed a story to clean this up:

https://activestatef.atlassian.net/browse/DX-1188

cmd/state-installer/cmd.go
```diff
@@ -482,3 +476,41 @@ func determineLegacyUpdate(stateToolInstalled bool, packagedStateExe, payloadPat
 func noArgs() bool {
 	return len(os.Args[1:]) == 0
 }
+
+var branchRegex = regexp.MustCompile("Branch (\\S+)")
+var versionRegex = regexp.MustCompile("Version (\\S+)")
```
Naatan, line 481: No regex should be required, use json output:

```
state --version --output json
```

cmd/state-installer/cmd.go
```diff
@@ -482,3 +476,41 @@ func determineLegacyUpdate(stateToolInstalled bool, packagedStateExe, payloadPat
...
+
+func shouldUpdateInstalledStateTool() bool {
+	logging.Debug("Checking if installed state tool is an older version.")
+
+	stateExe := constants.StateCmd + exeutils.Extension
+	cmd := exec.Command(stateExe, "--version")
```
Naatan, line 487: Please use one of the exeutils methods.

cmd/state-installer/cmd.go
```diff
@@ -482,3 +476,41 @@ func determineLegacyUpdate(stateToolInstalled bool, packagedStateExe, payloadPat
...
 	return len(os.Args[1:]) == 0
 }
+
+var branchRegex = regexp.MustCompile("Branch (\\S+)")
+var versionRegex = regexp.MustCompile("Version (\\S+)")
+
+func shouldUpdateInstalledStateTool() bool {
+	logging.Debug("Checking if installed state tool is an older version.")
+
+	stateExe := constants.StateCmd + exeutils.Extension
```
Naatan, line 486: Please use the path that was returned here: 

https://github.com/ActiveState/cli/pull/2053/files#diff-2dd975f5d9eec0de10d0d721cf03ea5c32713c0df58a631ce2f711d519575608R233

Naatan (changes requested):

cmd/state-installer/cmd.go
mitchell-as: I did, but they were not returning stdout :(

cmd/state-installer/cmd.go
mitchell-as: That link doesn't take me anywhere.

cmd/state-installer/cmd.go
mitchell-as: I don't like this because I think we should only call out to `state --version` when absolutely necessary (not already in an update flow, and user is not explicitly forcing install). With your proposed solution, I think we're going to call out if we're already in an update flow OR the user is forcing the install.

Naatan (changes requested):

cmd/state-installer/cmd.go
```diff
@@ -477,37 +476,34 @@ func noArgs() bool {
 	return len(os.Args[1:]) == 0
 }
 
-var branchRegex = regexp.MustCompile("Branch (\\S+)")
-var versionRegex = regexp.MustCompile("Version (\\S+)")
+type stateToolVersionOutput struct {
+	Branch  string `json:"branch"`
+	Version string `json:"version"`
+}
```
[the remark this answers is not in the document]
mitchell-as: That's cool. I didn't know it was there.

Naatan (approved):

PR #2057, opened 2022-08-23
Validate e-mail addresses before sending Platform request.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1180" title="DX-1180" target="_blank">BugDX-1180</a>  INVITE: Some email addresses failed to be used for invite
  </td></table>
  <br />
 

It is debateable whether to fail before attempting to send invites, or during the attempt. I chose the latter.

### DIFF
--- a/pkg/platform/model/organizations.go
+++ b/pkg/platform/model/organizations.go
@@ -1,6 +1,7 @@
 package model
 
 import (
+	"net/mail"
 	"strings"
 
 	"github.com/go-openapi/strfmt"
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
Parsing emails is notoriously difficult. I'd rather have the API handle this.

Is this the 500 Internal Server Error issue? If we've determined we're not mangling any user input then I think there's no bug for us to address. It should be a bug with Dashboard instead.

mitchell-as:
No, https://activestatef.atlassian.net/browse/DX-1178 is the 500 error. Still, I agree that the Platform should be handling e-mail validation. I assumed since both this issue and DX-1178 passed triage, there's something to do here. I'm happy to close this though.

PR #2060, opened 2022-08-23
Rebase PYTHONPATH on runtime directory, not temp runtime directory, for Camel builds.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1179" title="DX-1179" target="_blank">BugDX-1179</a>  PYTHONPATH is set up with _runtime_temp directory
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/implementations/camel/prepare_mac.go
+++ b/pkg/platform/runtime/setup/implementations/camel/prepare_mac.go
@@ -1,3 +1,4 @@
+//go:build darwin
 // +build darwin
 
 package camel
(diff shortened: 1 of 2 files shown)


Naatan (changes requested):
This bug is a regression, so we have to find out what broke it and then address that. I would not expect to see new code introduced as part of the fix.

The logic for using the correct dir appears to live here:

https://github.com/ActiveState/cli/blob/26d220756a6ed379b1e847d91659a600c0467701/pkg/platform/runtime/setup/setup.go#L506

Not sure why it's apparently not being triggered anymore though.

Regardless of the fix, we need to understand what caused the regression as it could be impacting other things we haven't yet discovered.

mitchell-as:
This is where the PYTHONPATH is being constructed with the _runtime_temp directory:
https://github.com/ActiveState/cli/blob/26d220756a6ed379b1e847d91659a600c0467701/pkg/platform/runtime/setup/setup.go#L486
The actual installation location does not yet exist. The function you linked to is only called at the end of the current function I linked to.
Do you happen to know when this regression occured version-wise?

mitchell-as:
I went back to v0.32.1 and it's still happening there. I cannot go any further back due to deprecations.

PR #2063, opened 2022-08-24
subshell/cmd.cleanPath() should be able to clean multiple path entries at once

mitchell-as:
<table>
<td>
  </td></table>
  <br />

### DIFF
--- a/internal/subshell/cmd/env.go
+++ b/internal/subshell/cmd/env.go
@@ -76,7 +76,7 @@ func cleanPath(keyValue, oldEntry string) string {
 
 	var newValue []string
 	for _, entry := range strings.Split(keyValue, string(os.PathListSeparator)) {
-		if _, ok := oldEntries[filepath.Clean(entry)]; ok {
+		if oldEntries[filepath.Clean(entry)] {
 			continue
 		}
 		newValue = append(newValue, entry)
--- a/internal/subshell/cmd/env.go
+++ b/internal/subshell/cmd/env.go
@@ -69,9 +69,14 @@ func (c *CmdEnv) unset(keyName, oldValue string) error {
 }
 
 func cleanPath(keyValue, oldEntry string) string {
+	oldEntries := make(map[string]bool)
+	for _, entry := range strings.Split(oldEntry, string(os.PathListSeparator)) {
+		oldEntries[filepath.Clean(entry)] = true
+	}
+
 	var newValue []string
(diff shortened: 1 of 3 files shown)


internal/subshell/cmd/env.go
```diff
@@ -69,9 +69,14 @@ func (c *CmdEnv) unset(keyName, oldValue string) error {
...
 func cleanPath(keyValue, oldEntry string) string {
+	oldEntries := make(map[string]bool)
+	for _, entry := range strings.Split(oldEntry, string(os.PathListSeparator)) {
+		oldEntries[filepath.Clean(entry)] = true
+	}
+
 	var newValue []string
 	for _, entry := range strings.Split(keyValue, string(os.PathListSeparator)) {
-		if filepath.Clean(entry) == filepath.Clean(oldEntry) {
+		if _, ok := oldEntries[filepath.Clean(entry)]; ok {
```
MDrakos, line 79: Nit:
```suggestion
		if oldEntries[filepath.Clean(entry)] {
```

Since you're using `map[string]bool` you can do this. If you want to keep your old pattern then I _think_ `map[string]struct{}` is more performant.

internal/subshell/cmd/env.go
```diff
@@ -69,9 +69,14 @@ func (c *CmdEnv) unset(keyName, oldValue string) error {
 }
 
 func cleanPath(keyValue, oldEntry string) string {
+	oldEntries := make(map[string]bool)
+	for _, entry := range strings.Split(oldEntry, string(os.PathListSeparator)) {
+		oldEntries[filepath.Clean(entry)] = true
+	}
```
MDrakos, line 75: The environment we get for cleaning is from the config, so doesn't this mean that we're not storing it correctly if we have to do splitting here?

mitchell-as: For a `state deploy` on Windows, it's adding 4 directories to %PATH%. Since `oldEntry` is a string, it's a ';'-separated list of those directories. We are storing config values correctly in a map -- it's just that the values may be lists like this one.

internal/subshell/cmd/env.go
mitchell-as: Thanks for the tip. I did not know `map[string]bool` was special.

MDrakos (approved):

mitchell-as:
Since Windows tends to be really sticky about files and directories being in use, I'm defensively removing the user env changes first before the deployment itself. Please re-review.

MDrakos (approved):

PR #2065, opened 2022-08-25
Print all user-facing installer errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-957" title="DX-957" target="_blank">StoryDX-957</a>  Our installer handles locale vs internal errors
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -193,9 +193,7 @@ func main() {
 	an.Event(AnalyticsFunnelCat, "pre-exec")
 	err = cmd.Execute(processedArgs[1:])
 	if err != nil {
-		isInputError := locale.IsInputError(err)
-
-		if isInputError {
+		if locale.IsInputError(err) {
 			an.EventWithLabel(AnalyticsCat, "input-error", errs.JoinMessage(err))
 			multilog.Error("Installer input error: " + errs.JoinMessage(err))
 		} else {
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ActiveState/cli/internal/runbits/panics"
 	"github.com/ActiveState/cli/internal/subshell"
 	"github.com/ActiveState/cli/internal/updater"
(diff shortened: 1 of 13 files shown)


daved (approved):
I'm approving for now, but it feels like there may be a more fundamental issue here (I may be completely wrong here). It seems like we might need something like locale.JoinInputMessage so callers are simpler. I've not dug into this PR or the relevant errs/locale code, so feel free to disregard if my concern doesn't fit. If it does, please file a story to improve this solution going forward. @Naatan Thoughts?

Naatan (changes requested):
Can we consolidate the logic between `state` and now `state-installer`? Conceptually there should be no difference in how they spit out errors.

ie. move most of this: https://github.com/ActiveState/cli/blob/aa97c06ecdbc94acf7e43a61a9833c212a13e98f/cmd/state/errors.go#L77

Into an internal package (maybe `cmdlets/errors` or something like that).

cmd/state-installer/cmd.go
```diff
@@ -200,11 +201,9 @@ func main() {
 			multilog.Critical("Installer error: " + errs.JoinMessage(err))
 		}
 
-		exitCode = errs.UnwrapExitCode(err)
+		exitCode, err = errors.Unwrap(err)
 		an.EventWithLabel(AnalyticsFunnelCat, "fail", err.Error())
-		if !errs.IsSilent(err) {
-			out.Error(err.Error())
-		}
```
Naatan, line 207: This part still seems relevant. This was handling the case where we run the installer with a subshell, and then something the user did in the subshell errored. This is not an error on our end so we should not print it.

Naatan (changes requested):

cmd/state-installer/cmd.go
mitchell-as: Oh, I erroneously thought that was handled here: https://github.com/ActiveState/cli/blob/8ed0bef228d492ce2203e325226f80fd84ed3268/pkg/cmdlets/errors/errors.go#L94

Naatan (approved):

PR #2074, opened 2022-08-30
Added `state use show` command with integration test.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1198" title="DX-1198" target="_blank">StoryDX-1198</a>  USE: User can print current default that defined by `state use <namespace>`
  </td></table>
  <br />

### DIFF
--- a/internal/runners/use/show.go
+++ b/internal/runners/use/show.go
@@ -1,0 +1,40 @@
+package use
+
+import (
+	"github.com/ActiveState/cli/internal/config"
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/pkg/project"
+)
+
+type Show struct {
+	out output.Outputer
+	cfg *config.Instance
+}
+
+func NewShow(prime primeable) *Show {
+	return &Show{
+		prime.Output(),
+		prime.Config(),
+	}
+}
+
+func (s *Show) Run() error {
+	projectDir := s.cfg.GetString(constants.GlobalDefaultPrefname)
+	if projectDir == "" {
+		return locale.NewInputError("err_use_show_no_default_project", "No default project is set.")
+	}
+
+	proj, err := project.FromPath(projectDir)
+	if err != nil {
+		return locale.WrapError(err, "err_use_show_get_project", "Could not get default project.")
+	}
+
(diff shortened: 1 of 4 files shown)


test/integration/use_int_test.go
```diff
@@ -160,6 +160,53 @@ func (suite *UseIntegrationTestSuite) TestReset() {
...
+	cp = ts.SpawnWithOpts(e2e.WithArgs("use", "show"))
+	cp.ExpectLongString("The default project to use is ActiveState-CLI/Python3, located at")
+	if runtime.GOOS != "windows" {
+		projectDir := filepath.Join(ts.Dirs.Work, "Python3")
+		cp.ExpectLongString(projectDir)
+	} else {
+		// Windows sometimes uses shortened paths, sometimes not.
+		// Just verify the last two components of the path are there and assume all is well,
+		// especially if the Linux and macOS tests are passing.
+		projectDir := filepath.Join(filepath.Base(ts.Dirs.Work), "Python3")
```
MDrakos, line 196: nit: We have `fileutils.GetLongPath` and `fileutils.GetShortPath`, one of them might help with this.

MDrakos (dismissed):

test/integration/use_int_test.go
mitchell-as: That's good to know, but there's still the matter of knowing *which* one Windows is going to use for a given path.

internal/runners/use/show.go
```diff
@@ -0,0 +1,40 @@
...
+
+	proj, err := project.FromPath(projectDir)
+	if err != nil {
+		return locale.WrapError(err, "err_use_show_get_project", "Could not get default project.")
+	}
+
+	s.out.Print(locale.Tl("use_show", "The default project to use is {{.V0}}, located at {{.V1}}",
+		proj.NamespaceString(),
+		projectDir,
+	))
```
Naatan, line 37: We'll want this to be JSON interpretable also. Try something like this: 

https://github.com/ActiveState/cli/blob/37e1f4f6b6dbffed4ba4716f365c8967b8b4834d/internal/runners/export/jwt.go#L32

test/integration/use_int_test.go
Naatan: We could just verify that either longpath or shortpath matches. That's a safer assertion than just using the base dir.

Naatan (changes requested):

Naatan (approved):

PR #2075, opened 2022-08-30
Try waiting 50% longer between GitHub API requests.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1174" title="DX-1174" target="_blank">BugDX-1174</a>  GitHub API still rate limiting
  </td></table>
  <br />
 

This parameter can be tuned, but start here.

The default wait interval is based on an assumption of only one job hitting the GitHub API at a time. As soon as there are concurrent jobs, rate limiting is more probable.

### DIFF
--- a/scripts/internal/workflow-helpers/ghtransport.go
+++ b/scripts/internal/workflow-helpers/ghtransport.go
@@ -69,7 +69,7 @@ func (rlt *rateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// If you're making a large number of POST, PATCH, PUT, or DELETE requests
 	// for a single user or client ID, wait at least one second between each request.
 	if rlt.delayNextRequest {
-		delay := multiplier * writeDelay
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
It doesn't seem like the multiplier does anything but obscure the actual wait time, ie. the delay will ALWAYS be `multiplier * writeDelay` as far as I can tell. Suggest instead just increasing writeDelay.

mitchell-as:
The 720 milliseconds is a baseline number, and I think it has some value as a reference point. If we need to increase it to 200% for example, then it can be interpreted as "at any given time we likely have 2 jobs accessing the GitHub API at the same time". To me that means a lot more than 1440 milliseconds.

Ultimately, if we keep having to increase it, it seems like a good idea to easily reference the baseline number we've deviated from.

Just my opinion.

Naatan (approved):

PR #2079, opened 2022-09-01
Implemented prototype for releasing remote-installer

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1168" title="DX-1168" target="_blank">StoryDX-1168</a>  We can deploy releases of individual components
  </td></table>
  <br />


activestate.yaml
```diff
@@ -326,6 +329,12 @@ scripts:
     description: Deploys update files to S3. This steps is automated by CI and should never be ran manually unless you KNOW WHAT YOU'RE DOING.
     value: |
       go run scripts/ci/s3-deployer/main.go build/installers us-east-1 state-tool update/state
+  - name: deploy-remote-installer
+    language: bash
+    value: |
+      set -e
+      $constants.SET_ENV
+      go run scripts/ci/s3-deployer/main.go $BUILD_TARGET_DIR/remote-installer us-east-1 state-tool remote-installer
```
MDrakos, line 337: It doesn't look like we're doing any compression of the remove installer anywhere. Is that something we want to do before deploying it?

MDrakos (approved):
Looks good to me!

activestate.yaml
Naatan: No, if anything I'm thinking of removing UPX as it's only causing issues. Plus for State Tool it's not saving any bandwidth because the UPX compression is nullified by the tar.gz compression.

PR #2089, opened 2022-09-06
Checkout logic should not care about existing checkouts

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1206" title="DX-1206" target="_blank">BugDX-1206</a>  TestRemoteInstallIntegrationTestSuite/TestInstall integration test failure
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/findproject/project.go
+++ b/internal/runbits/findproject/project.go
@@ -1,4 +1,4 @@
-package project
+package findproject
 
 import (
 	"sort"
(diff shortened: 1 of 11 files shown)


internal/runbits/findproject/project.go
```diff
@@ -20,6 +20,38 @@ func IsLocalProjectDoesNotExistError(err error) bool {
...
+		return FromPath(path, ns)
+	}
+
+	// Priority #2 - Namespace
+	if ns != nil {
+		return FromNamespaceLocal(ns, cfg, prompt)
+	}
+
+	// Priority #3 - Env
+	pj, err := project.FromEnv()
```
daved, line 35: This might be confusing if we ever have a FromEnvVar method. Maybe "FromDisk"? Disregard if I'm misunderstanding, and generally feel free to skip this concern.

internal/testhelpers/tagsuite/tagsuite.go
```diff
@@ -57,6 +57,7 @@ const (
 	Secrets         = "secrets"
 	Switch          = "switch"
 	Shell           = "shell"
+	FindProject     = "findproject"
```
daved, line 60: Is this used?

daved (approved):

internal/runbits/findproject/project.go
Naatan: It seems appropriate to me; an env var is one of the possible sources this function will use.

internal/testhelpers/tagsuite/tagsuite.go
Naatan: Ah nope, thanks for spotting. I had started implementing tests but it was too much of a scope creep, so I filed https://activestatef.atlassian.net/browse/DX-1222

daved (approved):

Naatan:
Reopened PR here: https://github.com/ActiveState/cli/pull/2090

I had created this one under the wrong jira ID.

PR #2098, opened 2022-09-08
Fix prompt for bash and zsh

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1220" title="DX-1220" target="_blank">BugDX-1220</a>  `state activate` does not modify PS1
  </td></table>
  <br />
 

Note my particular issue was not fixed as it is caused by powerlevel10k, which doesn't respect your PS1 (by design).


daved (approved):
I'm concerned that advanced users may not want their prompt modified. Do we have a way to prevent this change and make the data (org/proj) available? If not, I suggest it would feel more considerate.

Naatan:
> I'm concerned that advanced users may not want their prompt modified. Do we have a way to prevent this change and make the data (org/proj) available? If not, I suggest it would feel more considerate.

This is a bugfix, so I don't think that's appropriate for us to get into for this PR. However, I'd encourage you to file a Jira story so we can discuss it on its own merit.

PR #2099, opened 2022-09-08
Fix namespace validation

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1225" title="DX-1225" target="_blank">BugDX-1225</a>  `state shell` fails if no path/namespace given
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/findproject/project.go
+++ b/internal/runbits/findproject/project.go
@@ -46,7 +46,7 @@ func FromPath(path string, ns *project.Namespaced) (*project.Project, error) {
 		return nil, &LocalProjectDoesNotExist{locale.WrapInputError(err, "err_project_frompath_notexist", "", path)}
 	}
 
-	if ns != nil && ((ns.Owner != "" && pj.Namespace().Owner != ns.Owner) || pj.Namespace().Project != ns.Project) {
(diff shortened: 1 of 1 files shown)


internal/runbits/findproject/project.go
```diff
@@ -46,7 +46,7 @@ func FromPath(path string, ns *project.Namespaced) (*project.Project, error) {
 		return nil, &LocalProjectDoesNotExist{locale.WrapInputError(err, "err_project_frompath_notexist", "", path)}
 	}
 
-	if ns != nil && ((ns.Owner != "" && pj.Namespace().Owner != ns.Owner) || pj.Namespace().Project != ns.Project) {
+	if ns != nil && ns.IsValid() && ((ns.Owner != "" && pj.Namespace().Owner != ns.Owner) || pj.Namespace().Project != ns.Project) {
```
daved, line 49: nit: This could be cleaned up a lot by having IsValid do the nil check, and having a method on ns like namespace1.Matches(namespace2).

daved (approved):

internal/runbits/findproject/project.go
Naatan: I've filed this story to follow up:

https://activestatef.atlassian.net/browse/DX-1231

PR #2100, opened 2022-09-09
Fix state shell not finding project, and test assertions

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1234" title="DX-1234" target="_blank">BugDX-1234</a>  Nightlies: TestShellIntegrationTestSuite/TestCd (Activate) failing
  </td></table>
  <br />

### DIFF
--- a/internal/installation/appinfo.go
+++ b/internal/installation/appinfo.go
@@ -1,9 +1,13 @@
 package installation
 
 import (
+	"os"
 	"path/filepath"
+	"strings"
 
+	"github.com/ActiveState/cli/internal/condition"
 	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/environment"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/osutils"
 )
@@ -42,6 +46,14 @@ func newExecFromDir(baseDir string, exec executableType) (string, error) {
 		}
 	} else {
 		path = filepath.Dir(osutils.Executable())
+	}
+
+	// Work around dlv debugger giving an unexpected executable path
+	if !condition.BuiltViaCI() && len(os.Args) > 1 && strings.Contains(os.Args[0], "__debug_bin") {
+		rootPath := filepath.Clean(environment.GetRootPathUnsafe())
+		if rootPath == filepath.Clean(path) {
+			path = filepath.Join(path, "build")
+		}
 	}
 
 	return filepath.Join(path, execData[exec]), nil
--- a/internal/runners/shell/shell.go
+++ b/internal/runners/shell/shell.go
@@ -1,8 +1,6 @@
 package shell
 
 import (
-	"fmt"
-
 	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/locale"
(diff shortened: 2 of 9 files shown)


test/integration/shell_int_test.go
```diff
@@ -110,12 +111,11 @@ func (suite *ShellIntegrationTestSuite) TestCwdShell() {
 func (suite *ShellIntegrationTestSuite) TestCd() {
 	suite.OnlyRunForTags(tagsuite.Shell)
 
-	ts := e2e.New(suite.T(), false)
+	ts := e2e.New(suite.T(), true)
```
daved, line 114: Is this still valid?

pkg/project/namespace.go
```diff
@@ -82,7 +82,7 @@ func (ns *Namespaced) Type() string {
 
 // IsValid returns whether or not the namespace is set sufficiently.
 func (ns *Namespaced) IsValid() bool {
-	return ns != nil && ns.Owner != "" && ns.Project != ""
+	return ns != nil && ns.Project != ""
```
daved, line 85: Sneak.

internal/runners/shell/shell.go
```diff
@@ -81,6 +76,7 @@ func (u *Shell) Run(params *Params) error {
 		return locale.WrapInputError(err, "err_shell_runtime_new", "Could not start a shell/prompt for this project.")
 	}
 
+	fmt.Printf(`rti: %v, proj: %v`, rti, proj)
```
daved, line 79: Debug? If so, you might want to scan for more of these since I'm not digging to hard. Or should I do the scan?

internal/installation/appinfo.go
```diff
@@ -44,6 +48,14 @@ func newExecFromDir(baseDir string, exec executableType) (string, error) {
...
 	}
 
+	// Work around dlv debugger giving an unexpected executable path
+	if !condition.BuiltViaCI() && len(os.Args) > 1 && strings.Contains(os.Args[0], "__debug_bin") {
+		rootPath := filepath.Clean(environment.GetRootPathUnsafe())
+		if rootPath == filepath.Clean(path) {
+			path = filepath.Join(path, "build")
+		}
+	}
+
```
daved, line 58: This feels like a sort of pollution. Is it OK?

daved (changes requested):

test/integration/shell_int_test.go
Naatan: Nope, thanks for catching!

pkg/project/namespace.go
Naatan: ?

internal/runners/shell/shell.go
Naatan: Will do a scan, thanks for picking up on these.

internal/installation/appinfo.go
Naatan: It's a bit awkward but edge case and safe enough IMO.

pkg/project/namespace.go
daved: It looks like a change from another PR, so I thought it was being sneaked in.

daved (approved):

pkg/project/namespace.go
Naatan: Ah. I guess it sort of is in that I found this issue while running the `state use` test suite to see if the bug was fixed (the issue being that `state use shell cli` would not work).

PR #2104, opened 2022-09-12
Handle `state clean uninstall -n` correctly and report abortion of uninstall process

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1196" title="DX-1196" target="_blank">BugDX-1196</a>  UNINSTALL: `state clean uninstall -n` not uninstalling ST and have one row message print that confuse users.
  </td></table>
  <br />

### DIFF
--- a/internal/runners/clean/uninstall.go
+++ b/internal/runners/clean/uninstall.go
@@ -28,7 +28,8 @@ type Uninstall struct {
 }
 
 type UninstallParams struct {
-	Force bool
+	Force          bool
+	NonInteractive bool
 }
 
 type primeable interface {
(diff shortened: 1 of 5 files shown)


internal/runners/clean/uninstall.go
```diff
@@ -59,12 +60,13 @@ func (u *Uninstall) Run(params *UninstallParams) error {
...
-		ok, err := u.confirm.Confirm(locale.T("confirm"), locale.T("uninstall_confirm"), new(bool))
+		defaultChoice := params.NonInteractive
+		ok, err := u.confirm.Confirm(locale.T("confirm"), locale.T("uninstall_confirm"), &defaultChoice)
 		if err != nil {
-			return err
+			return locale.WrapError(err, "err_uninstall_confirm", "Could not confirm uninstall choice")
 		}
 		if !ok {
-			return nil
+			return locale.NewInputError("err_uninstall_aborted", "Uninstall aborted by user")
```
MDrakos, line 69: Situations like this have come up in the past and we generally defaulted to _not_ returning an error here as the state tool is doing what the user asked of it. It appears that the AC says otherwise though so I'm just voicing the concern here.

MDrakos (approved):

internal/runners/clean/uninstall.go
mitchell-as: I think I agree in principle. Mark has been doing a lot of testing lately and this lack of affirmation that process was canceled by the user seems to be an inconsistency. I guess it boils down to *how* to present this information to the user. Do we print a message to the user from the command itself and then return nil? Or do we return something (e.g. error) that delegates to the error handler what to ultimately do.

MDrakos: Cancellation confirmations should definitely be consistent across the state tool. I'm not sure what purpose the error code here would provide outside of automation, but since this is from an interactive prompt I don't think it would be helpful. I'm fine with the code as is but we may want to bring the discussion to the rest of the team.

mitchell-as: There is now a ticket to review UX for input errors: DX-1243

PR #2105, opened 2022-09-12
Attempt to prevent panic caused by reusing an already completed progressbar

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1187" title="DX-1187" target="_blank">BugDX-1187</a>  CHECKOUT: panic error just after successful checkout - `panic: *mpb.Progress instance can't be reused after it's done!`
  </td></table>
  <br />
 

The screenshot and traceback seems to indicate an ArtifactStartEvent was received after an ArtifactCompleteEvent that completed installation.

This is not actually a user-facing error, since the runtime setup completes successfully. It's just that the reporting of it is possibly erroneous. Log it as an error instead of returning one. The primary objective is to prevent a Go panic.


Naatan (approved):
Since we are already aiming to refactor this mechanic in its entirety and get rid of this race condition in the process, I think this is a good interim solution!

PR #2106, opened 2022-09-13
Handle `state clean config -n` correctly and report abortion of clean process.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1239" title="DX-1239" target="_blank">BugDX-1239</a>  `state clean config -n` does nothing
  </td></table>
  <br />

### DIFF
--- a/internal/runners/clean/config.go
+++ b/internal/runners/clean/config.go
@@ -31,7 +31,8 @@ type Config struct {
 }
 
 type ConfigParams struct {
-	Force bool
+	Force          bool
+	NonInteractive bool
 }
 
 func NewConfig(prime primeable) *Config {
@@ -53,12 +54,13 @@ func (c *Config) Run(params *ConfigParams) error {
 	}
 
 	if !params.Force {
-		ok, err := c.confirm.Confirm(locale.T("confirm"), locale.T("clean_config_confirm"), new(bool))
+		defaultChoice := params.NonInteractive
+		ok, err := c.confirm.Confirm(locale.T("confirm"), locale.T("clean_config_confirm"), &defaultChoice)
 		if err != nil {
-			return err
+			return locale.WrapError(err, "err_clean_config_confirm", "Could not confirm clean config choice")
 		}
 		if !ok {
-			return nil
+			return locale.NewInputError("err_clean_config_aborted", "Cleaning of config aborted by user")
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/runners/clean/config.go
```diff
@@ -53,12 +54,13 @@ func (c *Config) Run(params *ConfigParams) error {
 	}
 
 	if !params.Force {
-		ok, err := c.confirm.Confirm(locale.T("confirm"), locale.T("clean_config_confirm"), new(bool))
+		defaultChoice := params.NonInteractive
+		ok, err := c.confirm.Confirm(locale.T("confirm"), locale.T("clean_config_confirm"), &defaultChoice)
```
Naatan, line 58: This is already handled: https://github.com/ActiveState/cli/blob/52511befe457f17b63048fdc07b78be9d50c861f/internal/prompt/prompt.go#L149

You can just pass `p.BoolP(true)`. The default is used when ran under non-interactive mode.

In fact I think you can reduce this PR to pretty much this one change. You shouldn't need to pass NonInteractive to the runner.

Naatan (changes requested):

internal/runners/clean/config.go
mitchell-as: But we want the default to be `false` if running interactively. If I pass `p.BoolP(true)`, then running `state clean config` prompts me with `Y/n` (Y) as the default. What am I missing?

Naatan: > But we want the default to be false if running interactively

I think you're asserting that based on what the code currently does? I'm not seeing anything on the Jira ticket about this.

Imo the default should just be true, regardless of what mode you're running under. This is how we've handled all non-interactive prompts so far. The reason why there are some prompts with the wrong default is because the prompts preceded the non-interactive mechanic, so not all of them are well curated towards the non-interactive mechanic.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2106#discussion_r970034241

mitchell-as:
Okay, I changed the default to be Yes. The other changes are still valid. One locale id was incorrect, updated another to be more descriptive, for consistency we should return input error when confirmation is not given, and updated an affected unit test.

Naatan (approved):

PR #2109, opened 2022-09-13
Installer only starts subshell in interactive terminal

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1242" title="DX-1242" target="_blank">BugDX-1242</a>  State Tool Installer hangs when installing without a pseudo-tty
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -45,7 +45,6 @@ type Params struct {
 	command         string
 	force           bool
 	isUpdate        bool
-	isInteractive   bool
 	activate        *project.Namespaced
 	activateDefault *project.Namespaced
 }
@@ -128,7 +127,6 @@ func main() {
 	an.Event(AnalyticsFunnelCat, "start")
 
 	params := newParams()
-	params.isInteractive = terminal.IsTerminal(int(os.Stdin.Fd()))
 	cmd := captain.NewCommand(
 		"state-installer",
 		"",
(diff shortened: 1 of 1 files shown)


cmd/state-installer/cmd.go
```diff
@@ -386,7 +387,7 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 			an.EventWithLabel(AnalyticsFunnelCat, "forward-activate-default-err", err.Error())
 			return errs.Silence(errs.Wrap(err, "Could not activate %s, error returned: %s", params.activateDefault.String(), errs.JoinMessage(err)))
 		}
-	case !params.isUpdate && os.Getenv(constants.InstallerNoSubshell) != "true":
+	case !params.isUpdate && terminal.IsTerminal(int(os.Stdin.Fd())) && os.Getenv(constants.InstallerNoSubshell) != "true":
```
mitchell-as, line 390: Does calling int() on a uintptr do the correct cast? I'm coming from C/C++ land where casting pointers to/from ints is tricky.

mitchell-as (approved):
This looks good with the one caveat I mentioned.

cmd/state-installer/cmd.go
MDrakos: I'm not sure in all of the cases but I _believe_ that in this case it is fine. `os.Stdin.Fd()` simply casts the `int` of the file descriptor to a `uintptr`.

PR #2110, opened 2022-09-14
`state use show` shows a better error message if the default project has been deleted

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1237" title="DX-1237" target="_blank">BugDX-1237</a>  USE: `state use show` failed to show info if the project it pointed to was deleted.
  </td></table>
  <br />

### DIFF
--- a/internal/runners/use/show.go
+++ b/internal/runners/use/show.go
@@ -3,9 +3,11 @@ package use
 import (
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/output"
-	"github.com/ActiveState/cli/internal/runbits/findproject"
+	"github.com/ActiveState/cli/pkg/project"
+	"github.com/ActiveState/cli/pkg/projectfile"
 )
 
 type Show struct {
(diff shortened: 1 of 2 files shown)


internal/runners/use/show.go
```diff
@@ -39,8 +39,13 @@ func (s *Show) Run() error {
 		return locale.NewInputError("err_use_show_no_default_project", "No default project is set.")
 	}
 
-	proj, err := project.FromPath(projectDir)
+	proj, err := findproject.FromPath(projectDir, nil)
```
Naatan, line 42: This should use `project.FromPath`, not `findproject`. `findproject` is for if the question about the location is a little fuzzy, in this case we know exactly where the project should be.

Naatan (changes requested):

internal/runners/use/show.go
```diff
@@ -41,6 +43,11 @@ func (s *Show) Run() error {
 
 	proj, err := project.FromPath(projectDir)
 	if err != nil {
+		if errs.Matches(err, &projectfile.ErrorNoProject{}) {
+			return locale.WrapError(err,
+				"err_use_show_default_project_does_not_exist",
+				"The default project no longer exists. Please either check it out again or run [ACTIONABLE]state use reset[/RESET].")
```
Naatan, line 49:
```suggestion
				"The default project no longer exists. Please either check it out again with [ACTIONABLE]state checkout[/RESET] or run [ACTIONABLE]state use reset[/RESET] to unset your default project.")
```

Naatan (changes requested):

Naatan (approved):

PR #2113, opened 2022-09-19
Rebase PYTHONPATH on ${INSTALLDIR} for Camel builds.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1179" title="DX-1179" target="_blank">StoryDX-1179</a>  PYTHONPATH is set up with _runtime_temp directory
  </td></table>
  <br />
 

It will be expanded later, like other env vars and file transformations.

### DIFF
--- a/pkg/platform/runtime/setup/implementations/camel/prepare_mac.go
+++ b/pkg/platform/runtime/setup/implementations/camel/prepare_mac.go
@@ -1,3 +1,4 @@
+//go:build darwin
 // +build darwin
 
 package camel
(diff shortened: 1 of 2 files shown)


daved (changes requested):
Does this need to be applied to Linux and Windows as well? Also, is there any potential for a partial replacement to occur in this transformation `strings.Replace(sitePackages, installRoot, "${INSTALLDIR}", 1)`?

mitchell-as:
This does not need to be applied on Linux or Windows (there is no special handling of PYTHONPATH, and the unit test checking for temp runtime directory in PYTHONPATH passes on both of those platforms).

I have updated the code to perform replacement only if the path starts with the install root. I seriously doubt partial replacement can occur, but I suppose it doesn't hurt to be safe. Thanks for the suggestion.

daved (approved):

PR #2114, opened 2022-09-20
Show a "default project no longer exists" error message for state shell too when applicable

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1252" title="DX-1252" target="_blank">BugDX-1252</a>  SHELL: When project that set as default deleted `state shell` execution should have this as a possible reason for failing.
  </td></table>
  <br />


internal/locale/locales/en-us.yaml
```diff
@@ -1896,6 +1896,8 @@ arg_state_use_namespace_description:
   other: The org/project namespace of the project that you wish to use, or just the project name if previously used
 err_use_commit_id_mismatch:
   other: "Cannot switch to the given commit ID. Please use [ACTIONABLE]`state checkout`[/RESET] to check out a specific commit ID and then try again."
+err_use_default_project_does_not_exist:
```
MDrakos, line 1899: Is it possible for this error message to include the namespace of the default project we were expecting?

MDrakos (approved):
Looks good to me, just one question.

PR #2120, opened 2022-09-22
Reimplement executor in Go

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1250" title="DX-1250" target="_blank">StoryDX-1250</a>  Replace zig executor with go executor
  </td></table>
  <br />


pkg/platform/runtime/executor/init.go
```diff
@@ -149,62 +178,15 @@ func (f *Executor) createExecutor(targeter Targeter, sockPath string, env map[st
...
-	fwStr, err := strutils.ParseTemplate(string(fwBytes), tplParams)
-	if err != nil {
-		return errs.Wrap(err, "Could not parse %s template", boxFile)
+	if err := fileutils.CopyFile(srcExec, target); err != nil {
+		return locale.WrapError(err, "err_copyexecutor_fail", "Could not copy {{.V0}} to {{.V1}}", srcExec, target)
 	}
 
-	if err = ioutil.WriteFile(target, []byte(fwStr), 0755); err != nil {
-		return locale.WrapError(err, "Could not create executor for {{.V0}} at {{.V1}}.", exe, target)
+	if err := os.Chmod(target, 0o755); err != nil {
```
Naatan, line 187:
```suggestion
	if err := os.Chmod(target, 0755); err != nil {
```

pkg/platform/runtime/executor/init.go
```diff
@@ -123,24 +132,44 @@ func (f *Executor) Cleanup() error {
...
+	return strings.Contains(string(fileContents), "state-exec") ||
+		execmeta.IsMetaFile(fileContents) ||
+		// deprecated
+		strings.Contains(string(fileContents), executorDenoter) ||
+		strings.Contains(string(fileContents), shimDenoter)
+}
+
+func copyExecutor(destDir, exe, srcExec string) error {
+	name := filepath.Base(exe)
+	target := filepath.Clean(filepath.Join(destDir, name))
```
Naatan, line 163: This appears to be old code that is no longer used.

Naatan (changes requested):

cmd/state-exec/comm.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/ActiveState/cli/internal/svcmsg"
+)
+
+const (
+	network  = "unix"
+	msgWidth = 1024
+)
+
+func sendMsgToService(sockPath string, hb *svcmsg.Heartbeat) error {
```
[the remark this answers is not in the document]
daved: There's no desire to make this function reusable. The main reason it is a function is to continue the logic in a contained place to keep the run func clean.

cmd/state-exec/main.go
```diff
@@ -0,0 +1,91 @@
...
+	logr.Debug("hello")
+	defer logr.Debug("run goodbye")
+
+	hb, err := newHeartbeat()
+	if err != nil {
+		return err
+	}
+	logr.Debug("message data - pid: %s, exec: %s", hb.ProcessID, hb.ExecPath)
+
+	meta, err := newExecutorMeta(hb.ExecPath)
```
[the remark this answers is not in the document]
daved: The executor uses its own path to get the meta data so that it can know how to contact the service (sockpath). Included in that data is the project name and organization. We could, instead, send the same location through the socket and have the service lookup (from disk or cache), but it's quite trivial for it to just be done in one type (and, therefore, one place). In other words, splitting it up will make it more complex and also require more processing.

cmd/state-exec/meta.go
```diff
@@ -0,0 +1,75 @@
...
+func tranformedEnv(current []string, updates []string) []string {
+	for _, update := range updates {
+		if strings.HasPrefix(update, pathEnvVarPrefix) {
+			pathUpdate := update[len(pathEnvVarPrefix):]
+			if pathCurrent, ok := getEnvVarValue(current, pathEnvVarPrefix); ok {
+				pathUpdate += pathListSeparator + pathCurrent
+			}
+			update = pathEnvVarPrefix + pathUpdate
+		}
+		current = append(current, update)
```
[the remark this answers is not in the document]
daved: There are a tremendous amount of trade-offs occurring in this handling. If we don't have the entry twice, we'd have to update the existing entry. I did try that, and can go that route, but it didn't seem worth the time when the second entry overrides the first nicely.

internal/globaldefault/default.go
```diff
@@ -77,8 +78,8 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 	}
 
 	target := target.NewProjectTarget(proj, storage.GlobalBinDir(), nil, target.TriggerActivate)
-	fw := executor.NewWithBinPath(BinDir())
-	if err := fw.Update(target, env, exes); err != nil {
+	execInit := executor.NewInit(BinDir())
```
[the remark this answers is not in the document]
daved: New is used when the returned type matches the package name. The type was called executor.Executor before, but it's not an executor. It's closer to a set of executors, and I felt it is more specifically an initialization of executors. So, executor.Init felt right.

cmd/state-exec/meta.go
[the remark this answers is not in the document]
Naatan: Unless there is some strong technical reservation you have about updating the existing entry I would favour that. In my mind it should just be a matter of iterating over the env slice and updating the one that matches the PATH entry, which ought to be simple and results in a less messy end-result.

internal/globaldefault/default.go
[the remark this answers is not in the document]
Naatan: Let's instead rename `executor` to `executors`? `executors.New()` ought to address your concern. `NewInit()` does not convey to me the thing that you want it to convey.

pkg/platform/runtime/executor/init.go
daved: I am not sure how changing this will affect users if they update from an older version which used the denoter messages, so I opted to include them. Otherwise, I'm fairly sure that most of the code is still in use.

cmd/state-exec/comm.go
```diff
@@ -13,23 +13,23 @@ const (
 )
 
 func sendMsgToService(sockPath string, hb *svcmsg.Heartbeat) error {
-	ef := "send msg to service: %w"
+	efmt := "send msg to service: %w"
```
Naatan, line 16: Please write custom errors for each. The point of this is to allow us to trace the error, if multiple share the same error then that becomes more difficult.

Recall we used to include stacktraces. We removed those when you argued that those were unnecessary if we provide wrapped errors from which we can extrapolate the stack.

internal/svcctl/comm.go
```diff
@@ -78,17 +78,18 @@ func HeartbeatHandler(reporter RuntimeUsageReporter) ipc.RequestHandler {
...
+				multilog.Error("Heartbeat: Could not convert pid string (%s) to int in heartbeat handler: %s", hb.ProcessID, err)
 			}
 
 			var headless, commit, namespace string
 
 			metaFilePath := filepath.Join(filepath.Dir(hb.ExecPath), execmeta.MetaFileName)
 			if metaData, err := execmeta.NewFromFile(metaFilePath); err != nil {
-				multilog.Critical("Could not create meta data from filepath (%s): %s", metaFilePath, err)
+				multilog.Critical("Heartbeat Failure: Could not create meta data from filepath (%s): %s", metaFilePath, err)
+				return
```
Naatan, line 92: Can drop the else below now that this one is returning.

pkg/platform/runtime/executor/init.go
Naatan: I don't follow. Whether or not they used this is irrelevant, no code is using any of this anymore. It's dead code, we can drop it.

Naatan:
https://github.com/ActiveState/cli/pull/2120#discussion_r992808207

cmd/state-exec/comm.go
daved: Darn. Heh. I was hoping this would fly under the radar a bit. My short reply is that the error messages produced by this approach are very clear and contain the entire stack for digging into issues.

Longer reply, if you care for a preview of a Weekly topic: During Innovation Days I worked on communicating why this approach provides improved output, and is easier to use in the code. Two docs have been posted in the Weekly agenda that cover "confessing" versus "blaming" when annotating errors. The first is a visualization of the "alignment of responsibilities", and the other is code examples.

pkg/platform/runtime/executor/init.go
daved: I might be wrong, but... If we don't detect old executors, then the new executors will error out on install. Detecting that they are files under our control allows us to be free to remove the old ones in favor of the new ones.

cmd/state-exec/comm.go
Naatan: We descoped that from this PR though, meaning this PR still has to follow our best practices and not make assumptions about the future or the underlying errors.

pkg/platform/runtime/executor/init.go
Naatan: My bad, I though this wasn't being used at all.

Can we put this somewhere that makes it clear that this is legacy code? And rename any external invocations to have a `Legacy()` suffix? Want to make sure we can drop this in the future.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2120#discussion_r993696151

Naatan (dismissed):

Naatan (approved):

PR #2122, opened 2022-09-23
LE-817 Has the offline installer report telemetry

shaunmlowry:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/LE-817" title="LE-817" target="_blank">StoryLE-817</a>  Offline installers send telemetry
  </td></table>
  <br />


cmd/state-offline-installer/install.go
```diff
@@ -60,17 +63,46 @@ func NewRunner(prime primeable) *runner {
...
+func (r *runner) Event(eventType string, installerDimensions *dimensions.Values) {
+	go r.analytics.Event(ac.CatRuntimeUsage, eventType, installerDimensions)
+}
+
+func (r *runner) EventWithLabel(eventType string, msg string, installerDimensions *dimensions.Values) {
+	go r.analytics.EventWithLabel(ac.CatRuntimeUsage, eventType, msg, installerDimensions)
+}
+
+func (r *runner) handleFailure(err error, msg string, installerDimensions *dimensions.Values) error {
+	r.EventWithLabel("failure", msg, installerDimensions)
```
MDrakos, line 86: Do you want to do this one in a goroutine as well?

cmd/state-offline-installer/install.go
```diff
@@ -60,17 +63,46 @@ func NewRunner(prime primeable) *runner {
...
+	return errs.Wrap(err, msg)
 }
 
 func (r *runner) Run(params *Params) error {
+	installerDimensions := &dimensions.Values{
+		ProjectID: params.config.ProjectID,
+		CommitID:  params.config.CommitID,
+		Trigger:   p.StrP(target.TriggerCliOfflineInstaller.String()),
+	}
+	r.analytics.Event(ac.CatRuntimeUsage, "start", installerDimensions)
```
MDrakos, line 96: Right now I believe that `CatRuntimeUsage` is only used with heartbeats. I don't think that should be an issue but it's something to be mindful of. Maybe check with DW if that will affect anything?

MDrakos (commented):
Makes sense to me. I didn't find anything unusual here.

cmd/state-offline-installer/install.go
shaunmlowry: It's what @Naatan recommended we use for compatibility.  DW are on board with this too

cmd/state-offline-installer/install.go
shaunmlowry: That would be a good idea ;)

cmd/state-offline-uninstaller/uninstall.go
```diff
@@ -54,11 +67,48 @@ func newParams() *Params {
...
+	r.analytics.EventWithLabel(ac.CatRuntimeUsage, eventType, msg, installerDimensions)
+}
+
+func (r *runner) handleFailure(err error, msg string, installerDimensions *dimensions.Values) error {
+	r.EventWithLabel("failure", msg, installerDimensions)
+	return errs.Wrap(err, msg)
+}
+
+func (r *runner) handleFailureNewErr(msg string, installerDimensions *dimensions.Values) error {
+	r.EventWithLabel("failure", msg, installerDimensions)
```
Naatan, line 84: I'm not a fan of submitting failure messages to analytics, much less to the heartbeat category. These should be reported to rollbar. If we also want to surface these in our metrics then I'd like to see a use-case for it.

The runtime-use category is stable, so we should avoid pushing anything new to it that's experimental or without a strong use-case.

If your concern is that you might loose out on data, or that you want this data in the data warehouse, then I would recommend using a new category that we can mark as unstable for the time being.

cmd/state-offline-uninstaller/uninstall.go
```diff
@@ -54,11 +67,48 @@ func newParams() *Params {
...
+	if err := json.Unmarshal([]byte(configData), &installerConfig); err != nil {
+		return errs.Wrap(err, "Failed to decode config file")
+	}
+
+	installerDimensions := &dimensions.Values{
+		ProjectID: installerConfig.ProjectID,
+		CommitID:  installerConfig.CommitID,
+		Trigger:   p.StrP(target.TriggerCliOfflineUninstaller.String()),
+	}
+	r.analytics.Event(ac.CatRuntimeUsage, "start", installerDimensions)
```
Naatan, line 107: Multiple issues here:

* CatRuntimeUsage does not have a "start" action, see https://docs.google.com/document/d/12-IHGV6iBGE12p61cwKvc16YzahplKxJYSh30HKaBRk/edit#heading=h.u1lsd920bwhf
* Why are we tracking an uninstall as runtime usage? I imagine you would want to fire whatever event you guys landed on for deducting the runtime usage.
* This is currently running before the validation of the install directory (line 114).

cmd/state-offline-uninstaller/uninstall.go
```diff
@@ -67,26 +117,28 @@ func (r *runner) Run(params *Params) error {
...
 	}
 
 	r.out.Print("Removing installation directory")
 	err = os.RemoveAll(params.path)
 	if err != nil {
-		return errs.Wrap(err, "Error removing installation directory")
+		return r.handleFailure(err, "Error removing installation directory", installerDimensions)
 	}
 
+	r.analytics.Event(ac.CatRuntimeUsage, "success", installerDimensions)
```
Naatan, line 140: "Success" is not an action of CatRuntimeUsage. See https://docs.google.com/document/d/12-IHGV6iBGE12p61cwKvc16YzahplKxJYSh30HKaBRk/edit#heading=h.u1lsd920bwhf

cmd/state-offline-uninstaller/uninstall.go
Naatan: Note that for just tracking the runtime usage you do not need to make any special calls to the analytics mechanic, you just need to ensure that you pass it a valid instance of the synchronous analytics client here: https://github.com/ActiveState/cli/blob/ec562261f7b30dcebad79f30ab327f217090ebdc/cmd/state-offline-installer/install.go#L150

internal/analytics/client/sync/reporters/pixel.go
```diff
@@ -14,7 +16,13 @@ type PixelReporter struct {
...
 
 func NewPixelReporter() *PixelReporter {
-	return &PixelReporter{"https://state-tool.s3.amazonaws.com/pixel"}
+	var pixelUrl string
+
+	// Attempt to get the value for the pixel URL from the environment.  Fall back to default if that fails
+	if pixelUrl = os.Getenv(constants.AnalyticsPixelOverrideEnv); pixelUrl == "" {
+		pixelUrl = constants.DefaultAnalyticsPixel
+	}
+	return &PixelReporter{pixelUrl}
```
Naatan, line 25: What is the use-case of this? I think it might already be addressed:

https://github.com/ActiveState/cli/blob/ec562261f7b30dcebad79f30ab327f217090ebdc/internal/constants/constants.go#L124

Naatan (changes requested):
In addition to my comments, would it be possible to write a simple integration test?

https://github.com/ActiveState/cli/blob/660ec0b1f0a7877f2918b8f385fba946e960aafa/test/integration/analytics_int_test.go#L33

That may have to be a separate story though, as we still need an offline installer payload that we can use in tests.

internal/analytics/client/sync/reporters/pixel.go
shaunmlowry: That env var allows us to override which pixel we send data to during testing, the one you referenced allows you to use a TestReporter instead of a PixelReporter which doesn't use a pixel at all.

cmd/state-offline-installer/install.go
shaunmlowry: Offline installer now has its own category

cmd/state-offline-uninstaller/uninstall.go
shaunmlowry: Offline installer now has its own category.  We expect the installers to fail a lot and want to make it trivial for DW to gather statistics on failure modes in Athena

cmd/state-offline-uninstaller/uninstall.go
shaunmlowry: Offline installer now has its own category

cmd/state-offline-uninstaller/uninstall.go
shaunmlowry: Offline installer now uses its own category

internal/analytics/client/sync/reporters/pixel.go
Naatan: I understand that, I'm saying we have a way of debugging whether analytics are firing already. I'd prefer not to have multiple ways of doing the same thing.

I don't think it's your job to test if the pixel mechanic specifically works. What you care about is that the events are triggered. Whether pixels work ought to be out of scope for what you're working on.

cmd/state-offline-uninstaller/uninstall.go
Naatan: That only addresses the first point.

cmd/state-offline-installer/install.go
```diff
@@ -199,15 +208,15 @@ func (r *runner) Run(params *Params) error {
...
 			uninstallDir,
 			uninstallerDestName,
 			uninstallerDestName,
 			uninstallerDestName,
-			params.path,
+			installDir,
 			uninstallerDestName,
 		)
-		err := os.WriteFile(filepath.Join(params.path, "uninstall.bat"), []byte(batch), 0755)
+		err = os.WriteFile(filepath.Join(installDir, "uninstall.bat"), []byte(batch), 0755)
```
Naatan, line 219: Is this fixing something unrelated to this PR? I feel uncertain reviewing this without knowing the context.

Naatan (changes requested):
Unresolved two comments as they are not fully addressed. In addition to this, please consider that the runtime mechanic will still be firing heartbeats as that is what it does and that is what the offline-installer uses. So if you proceed this way then it means DW will have to untangle those events which adds complexity on their end and can lead to issues with tracking (not unlikely in my experience).

I would advocate not introducing new categories or actions unless we absolutely have to. Or if we are introducing new categories and actions to only use them for new use-cases (ie. don't use it for runtime tracking).

cmd/state-offline-uninstaller/uninstall.go
shaunmlowry: 2nd point - we now send both a category=runtime-use, action=delete and a category=offline-installer, action=success, trigger=offline-uninstaller event on successful uninstall as discussed
3rd point - this event is to denote the user initiating the uninstall process, to be later correlated with the success or failure event that indicates the final outcome

internal/analytics/client/sync/reporters/pixel.go
shaunmlowry: DW specifically asked for us to be able to redirect events to a secondary pixel for them to test processing the data at their end without interfering with data sent to the production pixel.  This addresses that use case.

cmd/state-offline-installer/install.go
shaunmlowry: It's tangentially related in that the uninstaller couldn't ever send a success event without this mechanism, as it would always fail trying to delete itself on windows.  Windows doesn't allow you to delete currently running executables as it pages them directly from their location on disk rather than copying them either to memory or a page file like UNIX systems do.

cmd/state-offline-uninstaller/uninstall.go
Naatan: 3rd point is still a problem imo. Because the curated error that is handled on line 114 will never be able to fire because it will run into the new uncurated error you're introducing in this PR.

I don't feel very strongly about this, I'm just pointing out that the code as written effectively makes the error handling on line 114 pointless.

cmd/state-offline-installer/install.go
```diff
@@ -89,62 +116,144 @@ func (r *runner) Run(params *Params) error {
...
+			return r.handleFailure(err, "Error creating uninstall directory", installerDimensions)
+		}
+		uninstallerDestName := fmt.Sprintf("%s-%s-%s.exe", *config.ProjectID, *config.CommitID, uninstallerFileNameRoot)
+
+		uninstallerSrc = filepath.Join(assetsPath, uninstallerFileNameRoot+".exe")
+		uninstallerDest = filepath.Join(uninstallDir, uninstallerDestName)
+
+		// create batch script
+		batch := fmt.Sprintf(
+			"@echo off\ncopy %s\\%s %%TEMP%%\\%s >nul 2>&1\n%%TEMP%%\\%s %s & del %%TEMP%%\\%s >nul 2>&1 & echo You can safely ignore the following error message:\n",
```
Naatan, line 211: Can we split this into its own bat file (you can use https://pkg.go.dev/embed to bring it in), and make it a bit more verbose (whitespace, new lines, comments)?

Seeing as the intend is for DevX to eventually take over I'd want this to be in a more maintainable form. This takes a lot of effort to unpack currently.

cmd/state-offline-installer/install.go
```diff
@@ -89,62 +116,144 @@ func (r *runner) Run(params *Params) error {
...
+			uninstallDir,
+			uninstallerDestName,
+			uninstallerDestName,
+			uninstallerDestName,
+			installDir,
+			uninstallerDestName,
```
Naatan, line 217: Fwiw you can reuse params in sprintf: https://stackoverflow.com/a/53026802/490321

cmd/state-offline-installer/install.go
```diff
@@ -89,62 +116,144 @@ func (r *runner) Run(params *Params) error {
...
+	var uninstallerDest string
+
+	/* Manually Install uninstaller */
+	if rt.GOOS == "windows" {
+		/* shenanigans because windows won't let you delete an executable that's running */
+		installDir, err := filepath.Abs(params.path)
+		if err != nil {
+			return r.handleFailure(err, "Error determining absolute install directory", installerDimensions)
+		}
+		uninstallDir := filepath.Join(installDir, "uninstall-data")
```
Naatan, line 200: Nit: I think `params.path` and `installDir` are effectively the same right? Can we standardize on one or the other? It's a bit confusing at the moment.

Naatan (changes requested):
Only thing I'm still asking changes for is the batch script, rest is just for your consideration.

Note this is getting sufficiently complex that we really ought to get some integration tests set up. I'm not confident dev testing and code reviews are going to suffice at this point.

cmd/state-offline-installer/install.go
Naatan: For your consideration, it might even be better to reuse this code:

https://github.com/ActiveState/cli/blob/65113461adf92c91713c04c92b2782d0c04995be/internal/runners/clean/run_win.go#L120

Which uses this batch file:

https://github.com/ActiveState/cli/blob/904d0f66f51ff2a91bfb7b33ffeebdb3fdc2a0e0/internal/assets/contents/scripts/removePaths.bat#L1-L0

cmd/state-offline-installer/install.go
shaunmlowry: params.path is whatever the user entered (e.g `perl`, `../perl`, `\perl`) and installDir is the absolute path (e.g `C:\users\User\perl`, `/opt/perl`, `E:\perl`).  The former is OK in the context of the installation process but we need the latter in the context of the uninstall.  We'll need to use the former in the next phase of work (UX improvement) where we may need to echo it back in its raw form to the user.

Naatan:
Closing in favour of https://github.com/ActiveState/cli/pull/2122

PR #2123, opened 2022-09-27
async.EventWithLabel should be run as a goroutine.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1170" title="DX-1170" target="_blank">BugDX-1170</a>  Noticing slowdown at "Setting config: projects"
  </td></table>
  <br />
 

sendEvent blocks while sending analytics, but async.EventWithLabel does not care about the return value, so run sendEvent within a goroutine.

### DIFF
--- a/internal/analytics/client/async/client.go
+++ b/internal/analytics/client/async/client.go
@@ -75,10 +75,12 @@ func (a *Client) Event(category string, action string, dims ...*dimensions.Value
 
 // EventWithLabel logs an event with a label to google analytics
 func (a *Client) EventWithLabel(category string, action string, label string, dims ...*dimensions.Values) {
-	err := a.sendEvent(category, action, label, dims...)
-	if err != nil {
-		multilog.Error("Error during analytics.sendEvent: %v", errs.Join(err, ":"))
-	}
+	go func() {
+		err := a.sendEvent(category, action, label, dims...)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/analytics/client/async/client.go
```diff
@@ -75,10 +75,12 @@ func (a *Client) Event(category string, action string, dims ...*dimensions.Value
...
-	err := a.sendEvent(category, action, label, dims...)
-	if err != nil {
-		multilog.Error("Error during analytics.sendEvent: %v", errs.Join(err, ":"))
-	}
+	go func() {
+		err := a.sendEvent(category, action, label, dims...)
+		if err != nil {
+			multilog.Error("Error during analytics.sendEvent: %v", errs.Join(err, ":"))
+		}
+	}()
```
daved, line 83: This call will already lead to a goroutine (https://github.com/ActiveState/cli/pull/2123/files#diff-cd900b80f4e1a877bb85c9f8872215cb3107ee30100fd5aa64204201bcfc4bdfL126-L133), so I don't think it is necessary. Further, the resolver package is using the sync package client, so it would remain unaffected. I don't have much experience with these two packages, but maybe the async client needs to be held by the relevant struct in the resolver package, and then use its EventWithLabel method.

daved (changes requested):

PR #2128, opened 2022-10-03
BE-1554 Fix some pip shim issues

icanhasmath:
Add a check to see if an argument has been passed to pip and
display a help message with possible arguments.
Add Help to the list of pip arguments.
Fix the path that is displayed in the configure message. Use raw
data to deal with unicode issue when translating displayed path
for windows.  Use os.path.join for universal path separators.


Naatan (approved):
Thanks!

PR #2130, opened 2022-10-03
Add more context to error message

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1204" title="DX-1204" target="_blank">BugDX-1204</a>  Rollbar: Could not start service
  </td></table>
  <br />
 

The other rollbar errors related to `Unable to start service` are mostly due to a restrive environment preventing the service to start.

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1976,6 +1976,13 @@ svcctl_wait_timeout:
     Please try running your command again.
 
     If this problem persists please report it on our forums: [ACTIONABLE]{{.V2}}[/RESET]
+svcctl_file_not_found:
+  other: |
+    Could not find the State Tool service executable.
+
+    Please try running your command again.
+
+    If this problem persists please reinstall the State Tool.
 err_autostart_app:
   other: Could not create new autostart app
 install_remote_title:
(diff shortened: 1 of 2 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1976,6 +1976,13 @@ svcctl_wait_timeout:
     Please try running your command again.
 
     If this problem persists please report it on our forums: [ACTIONABLE]{{.V2}}[/RESET]
+svcctl_file_not_found:
+  other: |
+    Could not find the State Tool service executable.
+
+    Please try running your command again.
+
+    If this problem persists please reinstall the State Tool.
```
Naatan, line 1985: I think mentioning the forums here would be better. Do we know for a fact that reinstalling would resolve these issues?

Naatan (changes requested):

internal/locale/locales/en-us.yaml
MDrakos: Fair point and no we don't, reinstall could end up with the same result.

mitchell-as (approved):

Naatan (approved):

PR #2131, opened 2022-10-04
Ensure verifypr workflow is called before target-version-pr.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1129" title="DX-1129" target="_blank">BugDX-1129</a>  Verify PR runs before target-version-pr
  </td></table>
  <br />


Naatan (changes requested):
I'm afraid you have them inverted: verifypr needs to run after target-version-pr. I worded this poorly in the jira ticket.

mitchell-as:
Okay, the target.yml became redundant so I removed it. Let me know if you want the action to run only on 'opened' and not also on 'reopened'. The latter was used for testing.

Naatan (changes requested):
Looks good! But please rename the workflow to something like "Target & Verify".

mitchell-as:
Let me know if I should change the filename too.

Naatan (approved):

PR #2137, opened 2022-10-05
Do not report to rollbar a failure to remove old installer files

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1270" title="DX-1270" target="_blank">BugDX-1270</a>  state-installer: Failed to remove old executable (reported many times per user)
  </td></table>
  <br />
 

Could be permission issues that are not our responsibility.

### DIFF
--- a/cmd/state-installer/installer_windows.go
+++ b/cmd/state-installer/installer_windows.go
@@ -12,7 +12,6 @@ import (
 	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/installation"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/subshell"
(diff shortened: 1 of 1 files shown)


daved (approved):

Naatan (changes requested):
This does not appear to address the issue. It's just muting it by not reporting it anymore.

mitchell-as:
The failure is occuring in Go's `os.Remove()`, which is outside our capabilities of handling. I suspect it's a permission issue, so there's nothing we can do about it, right? How else can we avoid polluting rollbar with this type of error?

Naatan:
I didn't realize this was for the ".old" removal, which is entirely on me for grokking too much. I agree in this case that a debug is appropriate. 

My fear was that we were silencing a legitimate error, but this particular one is a "known" issue that we already have a mechanism for to address (ie. try again next update).

Naatan (approved):

PR #2141, opened 2022-10-05
Display project path in error message

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1254" title="DX-1254" target="_blank">StoryDX-1254</a>  When applicable, display the org/project and path of a project not found
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/findproject/project.go
+++ b/internal/runbits/findproject/project.go
@@ -2,6 +2,7 @@ package findproject
 
 import (
 	"sort"
+	"strings"
 
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
@@ -79,7 +80,8 @@ func FromNamespaceLocal(ns *project.Namespaced, cfg projectfile.ConfigGetter, pr
 			logging.Debug("Cannot parse namespace: %v") // should not happen since this is stored
 			continue
 		}
-		if ns.ProjectIsEqual(namespaced) {
+		if !ns.AllowOmitOwner && strings.EqualFold(strings.ToLower(namespaced.String()), strings.ToLower(ns.String())) ||
+			(ns.AllowOmitOwner && strings.EqualFold(strings.ToLower(namespaced.Project), strings.ToLower(ns.Project))) {
 			matchingProjects[namespace] = paths
 			matchingNamespaces = append(matchingNamespaces, namespace)
 		}
@@ -129,7 +131,8 @@ func FromNamespaceLocal(ns *project.Namespaced, cfg projectfile.ConfigGetter, pr
 			continue
 		}
 
-		if ns.ProjectIsEqual(namespaced) && len(paths) > 0 {
+		if !ns.AllowOmitOwner && strings.EqualFold(strings.ToLower(namespaced.String()), strings.ToLower(ns.String())) ||
+			(ns.AllowOmitOwner && strings.EqualFold(strings.ToLower(namespaced.Project), strings.ToLower(ns.Project))) && len(paths) > 0 {
 			return nil, &LocalProjectDoesNotExist{
 				locale.NewInputError("err_findproject_notfound", "", ns.Project, paths[0]),
 			}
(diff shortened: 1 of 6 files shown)


internal/runbits/findproject/project.go
```diff
@@ -119,7 +119,16 @@ func FromNamespaceLocal(ns *project.Namespaced, cfg projectfile.ConfigGetter, pr
...
 
-		return project.FromPath(path)
+		proj, err := project.FromPath(path)
+		if err != nil {
+			if errs.Matches(err, &projectfile.ErrorNoProject{}) {
+				return nil, &LocalProjectDoesNotExist{
+					locale.WrapInputError(err, "err_local_project_not_checked_out", "", ns.Project),
+				}
+			}
+			return nil, locale.WrapError(err, "err_project_frompath", "Could not load project from path: {{.V0}}", path)
```
mitchell-as, line 129: This locale key already exists, so please reconcile the two.

mitchell-as (changes requested):
This looks good other than one small issue. Please flag Nathan for final review though.

mitchell-as (dismissed):

internal/runbits/findproject/project.go
```diff
@@ -119,7 +119,16 @@ func FromNamespaceLocal(ns *project.Namespaced, cfg projectfile.ConfigGetter, pr
...
 
-		return project.FromPath(path)
+		proj, err := project.FromPath(path)
+		if err != nil {
+			if errs.Matches(err, &projectfile.ErrorNoProject{}) {
+				return nil, &LocalProjectDoesNotExist{
+					locale.WrapInputError(err, "err_local_project_not_checked_out", "", ns.Project),
+				}
+			}
+			return nil, locale.WrapError(err, "err_findproject_frompath", "Could not load project from path: {{.V0}}", path)
```
Naatan, line 129: We should mention the namespace here as well imo.

internal/runbits/findproject/project.go
```diff
@@ -68,7 +68,7 @@ func FromNamespaceLocal(ns *project.Namespaced, cfg projectfile.ConfigGetter, pr
 
 	matchingProjects := make(map[string][]string)
 	matchingNamespaces := make([]string, 0)
-	for namespace, paths := range projectfile.GetProjectMapping(cfg) {
+	for namespace, paths := range projectfile.GetStaleProjectMapping(cfg) {
```
Naatan, line 71: This doesn't feel right. We might be prompting them to use a project that no longer exists.. that feels very strange.

I don't think we need to be doing this at all. If a user deleted a project then we can't find it. Offering the original location feels like it's just going to cause confusion and buggy code.

In terms of the story I would consider this scenario as "not applicable".

Naatan (changes requested):

internal/runbits/findproject/project.go
MDrakos: Ahh right, I hadn't thought about the prompting. Update this function to store the stale project listing and only use it when we can't find a project locally.

pkg/project/namespace.go
```diff
@@ -93,6 +94,15 @@ func (ns *Namespaced) Validate() error {
 	return nil
 }
 
+func (ns *Namespaced) Equal(other *Namespaced) bool {
+	if ns == nil || other == nil {
+		return false
+	}
+
+	return !ns.AllowOmitOwner && strings.EqualFold(strings.ToLower(other.String()), strings.ToLower(ns.String())) ||
```
Naatan, line 102: This has multiple issues:

* You're not checking whether AllowOmitOwner matches. And `Equal()` makes no implications about this behavior.
* Either namespace can include commit ID's. The current behavior would want the commit ID's  to match also. 

I suggest naming this something a bit more explicit. eg. `OwnerIsEqual()` and `ProjectIsEqual()`.

Naatan (changes requested):

pkg/project/namespace.go
MDrakos: Rather than having a convenience function I've moved the checks back to where they were initially.

PR #2144, opened 2022-10-06
Update state export _docs

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1266" title="DX-1266" target="_blank">StoryDX-1266</a>  Update `state export _docs` to denote unstable commands, generate table-of-contents, handle examples
  </td></table>
  <br />

### DIFF
--- a/internal/assets/contents/usage.tpl
+++ b/internal/assets/contents/usage.tpl
@@ -12,12 +12,15 @@ Aliases:
     {{.Cobra.NameAndAliases}}
 {{- end}}
 
+{{- if .Cmd.Examples  }}
+
 Examples:
 {{- range  .Cmd.Examples }} 
     {{ . -}}
-{{ end }}
+{{- end }}
+{{- end }}
 
-{{childCommands .Cmd}}
+{{- childCommands .Cmd}}
 {{- if .Cobra.HasAvailableFlags}}
 
 Flags:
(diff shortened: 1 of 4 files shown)


internal/runners/export/docs/docs.md.tpl
```diff
@@ -5,14 +5,24 @@
 
 ## {{.Group.String}}{{end}}
 
-### {{.NameRecursive}}
+### {{.NameRecursive}} {{ if .Unstable }}(unstable) {{ end }}
```
Naatan, line 8: I've added a new AC that I had forgotten to add after our original discussions:

> Unstable commands are grouped together below stable commands

That's my bad, but we will want to still address it for this PR.

internal/assets/contents/usage.tpl
```diff
@@ -11,11 +11,11 @@ Usage:
...
     {{.Cobra.NameAndAliases}}
 {{- end}}
-{{- if .Cobra.HasExample}}
 
 Examples:
-    {{.Cobra.Example}}
-{{- end}}
+{{- range  .Cmd.Examples }} 
+    {{ . -}}
+{{ end }}
```
Naatan, line 18: I think this also needs a `{{- if .Cmd.Examples }}`

Naatan (changes requested):

internal/assets/contents/usage.tpl
MDrakos: Oh right. Good catch.

Naatan (approved):

PR #2145, opened 2022-10-06
Stop enabling runtimes for integration tests that don't need them.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1173" title="DX-1173" target="_blank">StoryDX-1173</a>  We have a way of integration testing runtimes without the cost of setting up a full runtime
  </td></table>
  <br />
 

I didn't see a need to assert that runtime setup was skipped for every single removed instance of enabling the runtime. I just sprinkled a few choice assertions for unique commands like activate and checkout.

### DIFF
--- a/internal/runbits/runtime/runtime.go
+++ b/internal/runbits/runtime/runtime.go
@@ -53,8 +53,5 @@ func NewFromProject(
 			return nil, locale.WrapError(err, "err_update_runtime", "Could not update runtime installation.")
 		}
 	}
-	if rti.Disabled() {
-		out.Notice(locale.T("notice_runtime_disabled"))
-	}
 	return rti, nil
 }
--- a/internal/runbits/runtime/runtime.go
+++ b/internal/runbits/runtime/runtime.go
@@ -26,7 +26,7 @@ func NewFromProject(
 	out output.Outputer,
 	auth *authentication.Auth) (*rt.Runtime, error) {
 	projectTarget := target.NewProjectTarget(proj, storage.CachePath(), nil, trigger)
-	rti, err := rt.New(projectTarget, an, svcModel, out)
+	rti, err := rt.New(projectTarget, an, svcModel)
 	if err != nil {
 		if !rt.IsNeedsUpdateError(err) {
 			return nil, locale.WrapError(err, "err_activate_runtime", "Could not initialize a runtime for this project.")
@@ -52,6 +52,9 @@ func NewFromProject(
 			}
 			return nil, locale.WrapError(err, "err_update_runtime", "Could not update runtime installation.")
 		}
+	}
+	if rti.Disabled() {
+		out.Notice(locale.T("notice_runtime_disabled"))
 	}
 	return rti, nil
 }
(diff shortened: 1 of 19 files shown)


Naatan (changes requested):
Can we do the assertion still for all the tests that were updated? Your reasoning makes sense but I'd still like to include it as it will help prevent regressions (eg. we want to make sure that for example `checkout` has the "source a runtime" behavior, but we don't need to assert the runtime behavior itself).

internal/runbits/runtime/runtime.go
```diff
@@ -53,5 +53,8 @@ func NewFromProject(
 			return nil, locale.WrapError(err, "err_update_runtime", "Could not update runtime installation.")
 		}
 	}
+	if rti.Disabled() {
+		out.Notice(locale.T("notice_runtime_disabled"))
+	}
```
Naatan, line 58: Let's put this in the same place as the behavior: https://github.com/ActiveState/cli/blob/cfbe41918961fc22599d6f8fb2bc1f1efe55314d/pkg/platform/runtime/runtime.go#L101

Naatan (changes requested):

internal/runbits/runtime/runtime.go
mitchell-as: But we never get to this code path because a disabled runtime never returns the NeedsUpdateError error that would trigger a call to Update().

pkg/platform/runtime/runtime.go
```diff
@@ -66,9 +68,10 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
 }
 
 // New attempts to create a new runtime from local storage.  If it fails with a NeedsUpdateError, Update() needs to be called to update the locally stored runtime.
-func New(target setup.Targeter, an analytics.Dispatcher, svcm *model.SvcModel) (*Runtime, error) {
+func New(target setup.Targeter, an analytics.Dispatcher, svcm *model.SvcModel, out output.Outputer) (*Runtime, error) {
 	if strings.ToLower(os.Getenv(constants.DisableRuntime)) == "true" {
-		return &Runtime{disabled: true, target: target}, nil
+		out.Notice(locale.Tl("notice_runtime_disabled", "Skipping runtime setup because it was disabled by an environment variable"))
```
Naatan, line 73: Can we use `fmt.Fprintf(os.Stderr, ..)` here? It's been a conscious decision to keep the outputer out of the runtime mechanic. This is situation obviously calls for it, but seeing as the use-case is internal I think it's acceptable to use `fmt` here.

I worry that if we pass the outputer in that we will start using it elsewhere and break the original design goals (which is to use the event handles to handle output).

Naatan (changes requested):

internal/runbits/runtime/runtime.go
Naatan, line 58: We can drop this one now

Naatan (changes requested):

internal/runbits/runtime/runtime.go
mitchell-as: Oops, thanks.

Naatan (approved):

PR #2147, opened 2022-10-10
Initial test case for all supported shells.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1072" title="DX-1072" target="_blank">StoryDX-1072</a>  Develop test cases around supported shells
  </td></table>
  <br />
 

Also allow test cases to run commands in particular shells.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,9 @@ jobs:
           if [ -x "$(command -v apt-get)" ]; then
             sudo apt-get update
             sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish zsh tcsh -y
-            echo "ZSH_DISABLE_COMPFIX=true" > ~/.zshrc
+            # Prevent zsh insecure directory warning.
+            sudo chmod -R 755 /usr/local/share/zsh/site-functions
+            sudo chown -R root:root /usr/local/share/zsh/site-functions
           fi
 
           printenv
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
 
           if [ -x "$(command -v apt-get)" ]; then
             sudo apt-get update
-            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish -y
+            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish zsh tcsh -y
           fi
 
           printenv
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,6 +71,7 @@ jobs:
           if [ -x "$(command -v apt-get)" ]; then
             sudo apt-get update
             sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish zsh tcsh -y
+            echo "ZSH_DISABLE_COMPFIX=true" > ~/.zshrc
           fi
 
           printenv
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,8 +72,9 @@ jobs:
             sudo apt-get update
             sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish zsh tcsh -y
             # Prevent zsh insecure directory warning.
-            sudo chmod -R 755 /usr/local/share/zsh/site-functions
-            sudo chown -R root:root /usr/local/share/zsh/site-functions
+            sudo chmod -R 755 /usr/share/zsh/vendor-completions /usr/share/zsh
+            sudo chown -R root:root /usr/share/zsh/vendor-completions /usr/share/zsh
+            touch ~/.zshrc
           fi
 
           printenv
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
 
           if [ -x "$(command -v apt-get)" ]; then
             sudo apt-get update
-            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev -y
+            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish -y
           fi
 
           printenv
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,11 +70,7 @@ jobs:
 
           if [ -x "$(command -v apt-get)" ]; then
             sudo apt-get update
-            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev -y
-            # Install supported shells for the integration test that tests supported shells.
-            sudo apt-get install fish zsh tcsh -y
-            # Prevent zsh compinit prompt on startup.
-            sudo compaudit | xargs chmod g-w
+            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish zsh tcsh -y
           fi
 
           printenv
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,11 @@ jobs:
 
           if [ -x "$(command -v apt-get)" ]; then
             sudo apt-get update
-            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish zsh tcsh -y
+            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev -y
+            # Install supported shells for the integration test that tests supported shells.
+            sudo apt-get install fish zsh tcsh -y
+            # Prevent zsh compinit prompt on startup.
+            sudo compaudit | xargs chmod g-w
           fi
 
           printenv
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,6 +81,12 @@ jobs:
         run: |
           echo "${PSScriptRoot}/.github/deps/${{ runner.os }}/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
+      - # == Setup macOS ==
+        name: Setup (macOS)
+        shell: bash
+        run: brew install fish
+        if: runner.os == 'macOS'
+
       - # === Preprocess ===
         name: Preprocess
         shell: bash
--- a/internal/testhelpers/e2e/shell.go
+++ b/internal/testhelpers/e2e/shell.go
@@ -1,0 +1,31 @@
+package e2e
+
+import (
+	"strings"
+
+	"github.com/ActiveState/cli/internal/osutils"
+)
+
+type Shell string
+
+const (
+	Bash Shell = "bash"
+	Zsh        = "zsh"
+	Tcsh       = "tcsh"
+	Fish       = "fish"
+	Cmd        = "cmd.exe"
+)
+
+// QuoteCommand constructs and returns a command line string from the given list of arguments.
+// The returned string can be passed to the given shell for evaluation.
+func QuoteCommand(shell Shell, args ...string) string {
+	escaper := osutils.NewBashEscaper()
+	if shell == Cmd {
+		escaper = osutils.NewCmdEscaper()
+	}
+	quotedArgs := make([]string, len(args))
+	for i, arg := range args {
+		quotedArgs[i] = escaper.Quote(arg)
+	}
+	return strings.Join(quotedArgs, " ")
+}
--- a/internal/testhelpers/e2e/spawn.go
+++ b/internal/testhelpers/e2e/spawn.go
@@ -1,5 +1,11 @@
 package e2e
 
+import (
+	"strings"
+
+	"github.com/ActiveState/cli/internal/errs"
+)
+
 type SpawnOptions func(*Options) error
 
 func WithArgs(args ...string) SpawnOptions {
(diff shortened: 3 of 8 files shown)


mitchell-as:
Putting this up for early review. I expect more churn.

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,104 @@
...
+	defer ts.Close()
+
+	var shells []string
+	if runtime.GOOS == "linux" {
+		shells = []string{"bash", "fish"}
+	} else if runtime.GOOS == "darwin" {
+		shells = []string{"bash", "zsh", "tcsh"}
+	} else if runtime.GOOS == "windows" {
+		shells = []string{"cmd"}
+	}
```
Naatan, line 30: Nit: Would look cleaner as a switch.

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,104 @@
...
+			cp = ts.SpawnInBash(args)
+		} else if shell == "zsh" {
+			cp = ts.SpawnInZsh(args)
+		} else if shell == "tcsh" {
+			cp = ts.SpawnInTcsh(args)
+		} else if shell == "fish" {
+			cp = ts.SpawnInFish(args)
+		} else if shell == "cmd" {
+			cp = ts.SpawnInCmd(args)
+		}
```
Naatan, line 51: This seems like it could benefit from some sort of helper function, you're basically repeating it on line 64.

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,104 @@
...
+		if shell == "cmd" {
+			cp.SendLine("echo %COMSPEC%")
+			cp.Expect("cmd.exe") // verify the expected shell is running
+		} else if shell == "fish" {
+			cp.SendLine("echo $fish_pid")
+			cp.ExpectRe("\\d+") // verify the expected shell is running
+		} else {
+			cp.SendLine("echo $0")
+			cp.Expect(shell) // verify the expected shell is running
+		}
```
Naatan, line 93: This feels a bit awkward imo. I think it would look much cleaner if we wrote one test per shell and simply had convenience functions that the tests can use to reduce boilerplate.

Naatan (changes requested):

daved (changes requested):

.github/workflows/build.yml
```diff
@@ -70,7 +70,7 @@ jobs:
 
           if [ -x "$(command -v apt-get)" ]; then
             sudo apt-get update
-            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev -y
+            sudo apt-get install libgtk-3-dev libappindicator3-dev libwebkit2gtk-4.0-dev fish -y
```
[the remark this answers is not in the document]
mitchell-as: That would mean fish is installed each time we run `state activate`, which I don't think we want. I just want fish to be available for running the integration tests on CI.

test/integration/shells_int_test.go
mitchell-as: Thanks for the e2e.Shell tip. I've incorporated that idea into something more generic than I had. It's not as fancy as your suggestion, but it's an improvement at least.

test/integration/shells_int_test.go
Naatan: You could put some common assertions into a helper function which each test calls out to, eg. `ts.assertShellHasRuntime(cp)` or something like that.

mitchell-as: Is this something that is required to pass the PR? I did this and it still felt too awkward for me. It also introduces the element of order. The assertions have to all happen in the right order. It becomes harder/more problematic to refactor and/or add assertions down the road.

If this is required for your acceptance, I will do it. I'm just voicing my concerns first.

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,96 @@
...
+	cp := ts.SpawnWithOpts(e2e.WithArgs("checkout", "ActiveState-CLI/small-python"))
+	cp.Expect("Checked out project")
+	cp.ExpectExitCode(0)
+
+	for _, shell := range shells {
+		// Run the checkout in a particular shell.
+		cp := ts.SpawnInShell(
+			shell,
+			e2e.WithArgs("checkout", "ActiveState-CLI/small-python", string(shell)),
+		)
```
daved, line 42: It would be nice to set this up in a subtest so that only specific shells can be tested locally, and so that the output messages are organized more thoroughly.
```
	suite.T().Run(fmt.Sprintf("using_%s", shell), func(t *testing.T) {
```
I'm fairly sure that the design I had suggested would've allowed for that layer of nesting to be avoided, in case it's something you want to avoid (it is generally recommended to avoid nesting, but testing seems to be much looser regarding certain practices like this).

daved (changes requested):

test/integration/shells_int_test.go
mitchell-as: That's a fair point. The checkout command checks a project out to a shell-specific folder, and I was using that to identify where in the loop a test might have been failing, but an explicit marker is probably better.

mitchell-as:
By the way, I think this is ready for final review. I think it's in a relatively stable place now.

test/integration/shells_int_test.go
```diff
@@ -35,59 +36,61 @@ func (suite *ShellsIntegrationTestSuite) TestShells() {
...
-		// Run the checkout in a particular shell.
-		cp := ts.SpawnInShell(
-			shell,
-			e2e.WithArgs("checkout", "ActiveState-CLI/small-python", string(shell)),
-		)
-		cp.Expect("Checked out project")
-		if shell != e2e.Cmd {
-			cp.ExpectExitCode(0)
-		}
+		suite.T().Run(fmt.Sprintf("using_"+string(shell)), func(t *testing.T) {
```
daved, line 39: Making use of printf verbs, `fmt.Sprintf("using_%s", shell)` should work. The "%s" verb will cause the fmt package to check for a String() method, or string compatibility.

daved (changes requested):

daved (approved):

mitchell-as:
@Naatan ready for final review when you have the chance

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,99 @@
...
+func (suite *ShellsIntegrationTestSuite) TestShells() {
+	suite.OnlyRunForTags(tagsuite.Critical, tagsuite.Shell)
+
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	var shells []e2e.Shell
+	switch runtime.GOOS {
+	case "linux":
+		shells = []e2e.Shell{e2e.Bash, e2e.Fish}
```
Naatan, line 26: Linux has zsh and tcsh too.

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,99 @@
...
+	case "darwin":
+		shells = []e2e.Shell{e2e.Bash, e2e.Zsh, e2e.Tcsh}
```
Naatan, line 28: Mac has Fish shell too

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,99 @@
...
+	case "windows":
+		shells = []e2e.Shell{e2e.Cmd}
+	}
+
+	// Checkout the first instance. It doesn't matter which shell is used.
+	cp := ts.SpawnWithOpts(e2e.WithArgs("checkout", "ActiveState-CLI/small-python"))
```
Naatan, line 34: Suggest using `ACTIVESTATE_CLI_DISABLE_RUNTIME=false` here so the runtime checkout logic is centralized and isn't likely to interfere with the test assertions further down.

test/integration/shells_int_test.go
```diff
@@ -0,0 +1,99 @@
...
+			// use.
+			cp = ts.SpawnInShell(
+				shell,
+				e2e.WithArgs("shell", "small-python"),
+				e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
+			)
+			cp.Expect("Multiple project paths")
+
+			// Just pick the first one and verify the selection prompt works.
+			cp.SendLine("\n")
```
Naatan, line 60:
```suggestion
			cp.SendLine("")
```

This seems redundant.

mitchell-as:
@Naatan ping

internal/testhelpers/e2e/spawn.go
```diff
@@ -45,3 +51,29 @@ func BackgroundProcess() SpawnOptions {
...
+		if len(opts.Options.Args) == 0 {
+			return errs.New("e2e.WithShell must come after e2e.WithArgs")
+		}
+		opts.Options.CmdName = string(shell)
+		shellArg := "-c"
+		if shell == Cmd {
+			shellArg = "/k"
+		}
+		cmd := s.Exe + " " + strings.Join(opts.Options.Args, " ")
+		opts.Options.Args = []string{shellArg, cmd} // e.g. -c "state activate project/org"
```
Naatan, line 76:
```suggestion
		opts.Options.Args = append([]string{shellArg, opts.Options.CmdName}, opts.Options.Args...) // e.g. -c "state activate project/org"
```

Note I also replaced `s.Exe` with `opts.Options.CmdName`; s.Exe is not always used for all spawn commands, but CmdName should be far as I can tell.

Naatan (changes requested):

test/integration/shells_int_test.go
mitchell-as: But it's not installed. Are we supposed to install it on CI for this test?

internal/testhelpers/e2e/spawn.go
mitchell-as: That won't work because `opts.Options.CmdName` is the name of the shell we're running. Also, all arguments need to be grouped into a single quoted argument. Appending args into a `[]string` will unquote/ungroup them and you'll wind up with an incorrect command like `bash -c state activate project/org` instead of `bash -c "state activate project/org"`.

Naatan: I think just string joining and using quotes is gonna be very error prone.

My solution might not be appropriate, but I'd still like us to address these issues in this PR:

- Passing arguments via strings.Join and manually quoting it is going to create a lot of gotcha scenarios. Can we find a way to do this cleaner?
- On that point; `bash -c` has some issues, as raised during our stand. I think perhaps a more appropriate solution would be to use a custom SpawnWithCmd(bash). Then once spawned you can use ts.Send(opts.Options.CmdName), though we'd still be facing the arg parsing problem raised first.
- We can't just assume s.Exe is the command that's being spawned

test/integration/shells_int_test.go
Naatan: Yes. The AC's said which shells we want to cover with tests. CI not having a shell installed doesn't alter the AC's, but we can certainly talk if it creates a big scope creep.

Naatan (changes requested):
See comment responses

internal/testhelpers/e2e/spawn.go
mitchell-as: I've addressed the first point with the proper shell quoting of arguments. I've addressed point three by allowing the tester to specify an exe, but the `WithShell()` option silently invokes the shell with that exe.

As for option 2, I see two competing use cases:

1. Launch a single command in a particular shell and then exit.
2. Launch a shell, run 1 or more commands in it, and then exiting the shell (a lot like the `state activate` or `state shell` workflow).

I believe this ticket and this functionality should be targetted at (1). Therefore I am not in favor of having the test writer launch the shell, `Send()`ing input to it, and then `Send()`ing an exit command. It also feels like scope creep for this ticket. For more elaborate scenarios, maybe we should be leveraging different GitHub shells for running our integration tests instead of assuming bash for everything.

Naatan: The way I see it not addressing point 2 creates three issues:

- It makes for an error-prone API, because as a developer I can run `WithShell()` and start sending a bunch of input (ie. `tp.Send()`) which it won't respect implicitly because it only runs a single command.
- It means that we're not actually emulating what a user would do, which is the point of our integration tests. A user isn't going to run `bash -c whatever`, they will open a bash shell and then run `whatever`.
- The existing issue regarding loading RC files with `-c`

I don't think this is adding scope creep to the story, as the point of the story is to test that we actually can run the supported shells on the supported platforms, which the current iteration of the PR is not adequately doing.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2147#discussion_r998503103

internal/testhelpers/e2e/spawn.go
mitchell-as: Fair enough. The existing `SpawnInShell()` function threw me off. I have removed it in favor of `SpawnShell()` which spawns an interactive shell that you can send commands to and read output from.

internal/testhelpers/e2e/shell.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+type Shell string
+
+const (
+	Bash Shell = "bash"
+	Zsh        = "zsh"
+	Tcsh       = "tcsh"
+	Fish       = "fish"
+	Cmd        = "cmd.exe"
```
Naatan, line 16: Might be good if we hook this up to the constants in subshell, eg. https://github.com/ActiveState/cli/blob/5dc25c28a3dd85bddbfbc22a8fd68f5205025e1b/internal/subshell/bash/bash.go#L42-L42

internal/testhelpers/e2e/shell.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// QuoteCommand constructs and returns a command line string from the given list of arguments.
+// The returned string can be passed to the given shell for evaluation.
+func QuoteCommand(shell Shell, args ...string) string {
```
Naatan, line 21: Would this work the same on cmd.exe?

Naatan (approved):
Looks good to me! My 2 comments are nits, address them if you think they're relevant enough.

internal/testhelpers/e2e/shell.go
mitchell-as: The behavior is not exactly the same with cmd.exe vs bash, but the idea is the same. As you saw in the the function body, cmd.exe has its own quoter.

internal/testhelpers/e2e/shell.go
mitchell-as: These are actually executable names, not shell names (although they are mostly equivalent). These strings are passed to Spawn(), so they need to exist on $PATH. There's no guarantee that a given shell's `Name` string will exist on $PATH.

PR #2148, opened 2022-10-11
Fire initial runtime-use event from the state tool

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1290" title="DX-1290" target="_blank">StoryDX-1290</a>  Our runtime fires the initial heartbeat from the runtime process
  </td></table>
  <br />


MDrakos (approved):
I assume that offline installer means "not the state tool"? Because if this is actually meant to be run without an internet connection then why do we care about the runtime events firing?

Naatan:
> I assume that offline installer means "not the state tool"? Because if this is actually meant to be run without an internet connection then why do we care about the runtime events firing?

Offline only means no connection is required, not that we won't utilize it if it exists.

PR #2149, opened 2022-10-12
Implemented integration test for the offline installer

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1295" title="DX-1295" target="_blank">StoryDX-1295</a>  Offline installers have integration tests
  </td></table>
  <br />


internal/fileutils/fileutils.go
```diff
@@ -949,6 +959,19 @@ func ListDirSimple(sourcePath string, includeDirs bool) []string {
 	return result
 }
 
+// ListFilesUnsafe lists filepaths under the given sourcePath non-recursively
+func ListFilesUnsafe(sourcePath string) []string {
+	result := []string{}
+	files, err := ioutil.ReadDir(sourcePath)
+	if err != nil {
+		panic(err)
```
mitchell-as, line 967: In other panics, we add a sprintf message. Would that be helpful here?

internal/subshell/cmd/env.go
```diff
@@ -85,7 +85,7 @@ func cleanPath(keyValue, oldEntry string) string {
 }
 
 // setUserEnv sets a variable in the user environment and saves the original as a backup
```
mitchell-as, line 87: This comment needs updating.

test/integration/offinstall_int_test.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		tp.Expect("TEST REPLACEMENT", 5*time.Second)
+		tp.ExpectExitCode(0)
+	}
+
+	{ // Uninstall
+		tp := ts.SpawnCmdWithOpts(
+			suite.uninstallerPath,
+			e2e.WithArgs(targetDir),
+			e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false"),
```
mitchell-as, line 89: Does the uninstaller need a runtime enabled?

Naatan: Yes, it needs to access the runtime in order to find out what files to delete and whatnot.

activestate.yaml
```diff
@@ -68,6 +68,9 @@ scripts:
       export GO111MODULE=on
       goflags="${GOFLAGS}"; unset GOFLAGS
       GOFLAGS="${goflags}"
+      pushd ~
+      go install github.com/sanderhahn/gozip/cmd/gozip@latest
+      popd
```
shaunmlowry, line 73: 1. This should be vendored (or at least not using latest) so the build is reproducible
2. You should be using our fork of gozip from here: https://github.com/ActiveState/gozip

shaunmlowry (changes requested):

activestate.yaml
Naatan: Why do we have a fork? We've made no changes to it.

mitchell-as (dismissed):
This looks fine from my perspective.

mitchell-as (dismissed):

PR #2150, opened 2022-10-12
Offline installer reports runtime usage

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1293" title="DX-1293" target="_blank">StoryDX-1293</a>  Offline installers send telemetry
  </td></table>
  <br />
 

This is continuing the work started in https://github.com/ActiveState/cli/pull/2122


cmd/state-offline-installer/main.go
```diff
@@ -24,11 +25,22 @@ import (
 	"github.com/ActiveState/cli/pkg/cmdlets/errors"
 )
 
+type Params struct {
+	path string
+}
+
+func newParams() *Params {
+	return &Params{
+		path: "/tmp",
```
mitchell-as, line 34: I would prefer a more OS-agnostic value here, even if it's just a placeholder.

mitchell-as (changes requested):

cmd/state-offline-installer/main.go
Naatan: Yeah in fact it shouldn't be here at all. It's not how defaults work.

mitchell-as (approved):

PR #2153, opened 2022-10-13
Add debug output to better assess svc behavior

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1209" title="DX-1209" target="_blank">BugDX-1209</a>  flisten file does not exist
  </td></table>
  <br />


daved:
@Naatan There's not much I can do with this bug for now, so improving debug prints and waiting for another occurrence seems like the best bet.

Naatan (approved):

PR #2162, opened 2022-10-18
Recognize ACTIVESTATE_HOME as fallback if HOME is not set

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1276" title="DX-1276" target="_blank">StoryDX-1276</a>  When HOME is not set we error out and instruct the user
  </td></table>
  <br />
 

If that still doesn't work, show an informative error to the user.

Also route all fetching of user's home directory through one function. It was previously obtained in 3 different ways.

### DIFF
--- a/internal/osutils/user/user.go
+++ b/internal/osutils/user/user.go
@@ -1,0 +1,15 @@
+package user
+
+import (
+	"os"
+)
+
+// HomeDir returns the user's homedir
+func HomeDir() (string, error) {
+	dir, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return dir, nil
+}
--- a/internal/osutils/user/user.go
+++ b/internal/osutils/user/user.go
@@ -10,6 +10,7 @@ import (
 // from locale/errors.go because importing locale for NewInputError creates an import cycle.
 // Instead, return this error that looks like a LocalizedError.
 type HomeDirNotFoundError struct {
+	wrapped error
 }
 
 const homeDirNotFoundErrorMessage = "Could not proceed because your HOME environment variable is unset. " +
@@ -30,6 +31,10 @@ func (e *HomeDirNotFoundError) InputError() bool {
 	return true
 }
 
+func (e *HomeDirNotFoundError) Unwrap() error {
+	return e.wrapped
+}
+
 // HomeDir returns the user's homedir
 func HomeDir() (string, error) {
 	dir, err := os.UserHomeDir()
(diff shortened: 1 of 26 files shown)


internal/osutils/user/user.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const homeDirNotFoundErrorMessage = "Could not proceed because your HOME environment variable is unset. " +
+	"Please ensure that your HOME environment variable is set. " +
+	"Alternatively if you do not or cannot set this variable you can instead use the ACTIVESTATE_HOME variable." +
+	"\n\n" +
+	"This variable is used by the State Tool to determine things like the installation directory, config directory and cache directory."
+
+func (e *HomeDirNotFoundError) Error() string {
+	return homeDirNotFoundErrorMessage
```
daved, line 22: Please consider having the error message be something simple and possibly overly informative (i.e. for developers only). It also needs to implement the Unwrap method: `func (e *HomeDirNotFound) Unwrap() error {` so that the dev messaging contains the cause.

internal/osutils/user/user.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (e *HomeDirNotFoundError) UserError() string {
+	return homeDirNotFoundErrorMessage
```
daved, line 26: The message looks great for users. I think it is warranted to use localization.

daved (changes requested):

internal/osutils/user/user.go
mitchell-as: As mentioned in a comment in user.go:

```// HomeDirNotFoundError is an error that implements the ErrorLocalier and ErrorInput interfaces
// from locale/errors.go because importing locale for NewInputError creates an import cycle.
// Instead, return this error that looks like a LocalizedError.
```

As much as I'd like to use localization, I don't know how. Nathan suggested I implement the two functions that I did to make it look like a localized error.

Naatan:
Please flag me for final review.

daved (approved):

daved:
Added story https://activestatef.atlassian.net/browse/DX-1347

internal/osutils/user/user.go
```diff
@@ -0,0 +1,49 @@
...
+// HomeDir returns the user's homedir
+func HomeDir() (string, error) {
+	dir, err := os.UserHomeDir()
+	if err != nil {
+		var exists bool
+		dir, exists = os.LookupEnv(constants.HomeEnvVarName)
+		if !exists {
+			return "", &HomeDirNotFoundError{err}
+		}
+	}
```
Naatan, line 47: This should use the ACTIVESTATE_HOME as an override, not a fallback.

That's on me for not putting it in the AC's better. It does in fact say override in the story in one place and fallback in another -.-

Naatan (changes requested):
In addition to the code I commented on, please note the automated testing requirements:

* Test whether ACTIVESTATE_HOME overrides HOME
* Test whether running without either of the HOME env vars results in the curated error

Naatan (changes requested):

Naatan (approved):

PR #2163, opened 2022-10-18
Report input errors to pixel

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1309" title="DX-1309" target="_blank">StoryDX-1309</a>  We report input errors to the pixel
  </td></table>
  <br />

### DIFF
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ActiveState/cli/internal/table"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"github.com/thoas/go-funk"
 )
 
 func init() {
@@ -355,6 +356,17 @@ func (c *Command) Description() string {
 
 func (c *Command) Flags() []*Flag {
 	return c.flags
+}
+
+func (c *Command) ActiveCommands(args []string) []*Command {
+	var cmds []*Command
+	cmds = append(cmds, c)
+	for _, child := range c.Children() {
+		if funk.Contains(args, child.Name()) {
+			cmds = append(cmds, child)
+		}
+	}
+	return cmds
 }
 
 func (c *Command) ActiveFlags() []*Flag {
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -30,7 +30,6 @@ import (
 	"github.com/ActiveState/cli/internal/table"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"github.com/thoas/go-funk"
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


pkg/cmdlets/errors/errors.go
```diff
@@ -117,6 +108,37 @@ func Unwrap(err error) (int, error) {
...
+
+	// Log error if this isn't a user input error
+	if !locale.IsInputError(err) {
+		multilog.Critical("Returning error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
+	} else {
+		var flagNames []string
+		for _, flag := range flags {
+			flagNames = append(flagNames, fmt.Sprintf("--%s", flag.Name))
+		}
+		logging.Debug("Returning input error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
```
Naatan, line 136:
```suggestion
		logging.Debug("Reporting input error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
```

pkg/cmdlets/errors/errors.go
```diff
@@ -117,6 +108,37 @@ func Unwrap(err error) (int, error) {
...
+		an.Event(anaConst.CatDebug, anaConst.ActInputError, &dimensions.Values{
+			Trigger: p.StrP(strings.Join(flagNames, ", ")),
```
Naatan, line 138: This should contain the command as well, not just the flags.

eg. "state --version", not just "--version".

Naatan (changes requested):

internal/captain/command.go
```diff
@@ -357,6 +358,17 @@ func (c *Command) Flags() []*Flag {
...
+func (c *Command) ActiveCommands(args []string) []*Command {
+	var cmds []*Command
+	cmds = append(cmds, c)
+	for _, child := range c.Children() {
+		if funk.Contains(args, child.Name()) {
+			cmds = append(cmds, child)
+		}
+	}
+	return cmds
+}
```
Naatan, line 370: This is error prone since you're not respecting the order of the arguments.

Run-command uses `c.UseFull()`, can we use that here as well?

https://github.com/ActiveState/cli/blob/5dc25c28a3dd85bddbfbc22a8fd68f5205025e1b/internal/captain/command.go#L561-L561

Naatan (changes requested):

internal/captain/command.go
MDrakos: `c.UseFull()` works from the bottom-up while in this case, we are working top-down. Was still able to get this working but we do have to identify the lowest level child command.

Naatan (approved):

PR #2166, opened 2022-10-19
Deploy uninstall analytics event

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1291" title="DX-1291" target="_blank">StoryDX-1291</a>  state deploy uninstall fires runtime-use: delete 
  </td></table>
  <br />


internal/runners/deploy/uninstall/uninstall.go
```diff
@@ -66,19 +76,35 @@ func (u *Uninstall) Run(params *Params) error {
...
 	}
 
+	contents, err := fileutils.ReadFile(store.MarkerFile())
+	if err != nil {
+		return locale.WrapError(err, "err_deploy_uninstall_marker", "Unable to read marker file at [ACTIONABLE]{{.V0}}[RESET]. Deployment may be corrupted.", store.MarkerFile())
+	}
+
+	var namespace string
+	var commitID string
+	lines := strings.Split(string(contents), "\n")
```
mitchell-as, line 93: I'm wondering if maybe we can put this logic into a method in `store` so that we don't have to do any interpretetation of the store outside of the store package. I'm thinking if we modify the store's marker file down the road, we might have issues here that we didn't foresee.

mitchell-as (changes requested):
This looks good other than I think we should leave interpetation of the contents of the marker file to the store package itself, rather than outside the package.

pkg/platform/runtime/store/store.go
```diff
@@ -53,7 +52,7 @@ func New(installPath string) *Store {
 	}
 }
 
-func (s *Store) markerFile() string {
+func (s *Store) MarkerFile() string {
```
mitchell-as, line 55: I don't think this has to be exposed anymore, right? All marker file reading is now internal to the `store` package, yes?

mitchell-as (changes requested):

pkg/platform/runtime/store/store.go
MDrakos: Yes, you're right. Thanks for catching that!

mitchell-as (dismissed):
nit: you may want to change "markerfile" back to "markerFile", which is what it was originally.

mitchell-as (dismissed):
Perfect. I approve.

internal/runners/deploy/uninstall/uninstall.go
```diff
@@ -66,19 +74,33 @@ func (u *Uninstall) Run(params *Params) error {
...
+	namespace, err = store.Namespace()
+	if err != nil {
+		return locale.WrapError(err, "err_deploy_uninstall_namespace", "Unable to read namespace from marker file.")
+	}
+
+	var commitID string
+	commitID, err = store.CommitID()
+	if err != nil {
+		return locale.WrapError(err, "err_deploy_uninstall_commit_id", "Unable to read commit ID from marker file.")
+	}
```
Naatan, line 95: Analytics should never break a command. If we can't collect these then we can't report them, but it should not affect the action the user was taking.

Naatan: Also can we put this and the analytics call itself into a helper function? This is a lot of analytics code obfuscating the rest of the logic.

pkg/platform/runtime/store/store.go
```diff
@@ -180,6 +179,34 @@ func (s *Store) StoreRecipe(recipe *inventory_models.Recipe) error {
...
 
+func (s *Store) CommitID() (string, error) {
+	contents, err := fileutils.ReadFile(s.markerFile())
+	if err != nil {
+		return "", locale.WrapError(err, "err_deploy_uninstall_marker", "Unable to read marker file at [ACTIONABLE]{{.V0}}[RESET]. Deployment may be corrupted.", s.markerFile())
+	}
+
+	lines := strings.Split(string(contents), "\n")
+	if len(lines) < 1 {
+		return "", locale.NewError("err_deploy_uninstall_marker", "Marker file is incomplete, cannot determine commitID")
```
Naatan, line 190: This isn't really an error. We should 100% expect this to happen as older installations wouldn't have this.

pkg/platform/runtime/store/store.go
```diff
@@ -110,14 +109,14 @@ func (s *Store) MarkerIsValid(commitID strfmt.UUID) bool {
...
-func (s *Store) MarkInstallationComplete(commitID strfmt.UUID) error {
+func (s *Store) MarkInstallationComplete(commitID strfmt.UUID, namespace string) error {
 	markerFile := s.markerFile()
 	markerDir := filepath.Dir(markerFile)
 	err := fileutils.MkdirUnlessExists(markerDir)
 	if err != nil {
 		return errs.Wrap(err, "could not create completion marker directory")
 	}
-	err = fileutils.WriteFile(markerFile, []byte(fmt.Sprintf("%s\n%s", commitID.String(), constants.Version)))
+	err = fileutils.WriteFile(markerFile, []byte(strings.Join([]string{commitID.String(), constants.Version, namespace}, "\n")))
```
Naatan, line 119: Can we just make this JSON? This is starting to become a bit much for just line separating it and relying purely on the positioning.

Naatan (changes requested):

internal/runners/deploy/uninstall/uninstall.go
MDrakos: I can move the sourcing of the commitID and namespace to a helper function but I assumed we wanted to fire the analytics event after successfully uninstalling the deployment, which would mean that the file we're using would be gone.

pkg/platform/runtime/store/store.go
MDrakos: I thought about that as well but making this change would likely break older installations. We would either need some transitionary error messages or be able to support both formats which seems like too much just to support this analytics call.

internal/runners/deploy/uninstall/uninstall.go
Naatan: You could read the file when the store is initialized rather than when the function is called. Or use 2 helper functions instead of 1. Main point here was getting the analytics code to be minimal inside the run function.

pkg/platform/runtime/store/store.go
Naatan: We can transition from the old format when we encounter it. There are no permutations other than "is the old format" so that seems relatively trivial.

pkg/platform/runtime/store/store.go
```diff
@@ -187,7 +187,8 @@ func (s *Store) CommitID() (string, error) {
 
 	lines := strings.Split(string(contents), "\n")
 	if len(lines) < 1 {
-		return "", locale.NewError("err_deploy_uninstall_marker", "Marker file is incomplete, cannot determine commitID")
+		// Older installations may not have a marker file
```
Naatan, line 190: So then wouldn't the error on line 185 have triggered? I think what you're addressing here is that the file exists but doesn't have the data you're looking for.

Naatan (changes requested):

internal/runners/deploy/uninstall/uninstall.go
MDrakos: I've moved the sourcing of the info to a helper function which is called before we delete the marker file. The analytics call is still in the runner as the helper would simply be wrapping that.

pkg/platform/runtime/store/marker.go
```diff
@@ -0,0 +1,141 @@
...
+		logging.Debug("Could not match State Tool version in %s, expected: %s, got: %s", marker, constants.Version, marker.Version)
+		return false
+	}
+
+	return true
+}
+
+func (s *Store) parseMarker() (*Marker, error) {
+	markerFile := s.markerFile()
+	if !fileutils.FileExists(markerFile) {
```
Naatan, line 52:
```suggestion
	if !s.HasMarker() {
```

pkg/platform/runtime/store/marker.go
```diff
@@ -0,0 +1,141 @@
...
+	marker := &Marker{}
+	err = json.Unmarshal(contents, marker)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not unmasrshal marker file")
+	}
+
+	return marker, nil
+}
+
+func (s *Store) updateMarker(contents []byte) (*Marker, error) {
```
Naatan, line 74: Can we add a comment explaining what this does?

Naatan (changes requested):

Naatan (approved):

PR #2167, opened 2022-10-19
Fix failing offline installer tests

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1322" title="DX-1322" target="_blank">BugDX-1322</a>  Offline installer tests are all failing on master
  </td></table>
  <br />


internal/osutils/privileges_win.go
```diff
@@ -1,12 +1,22 @@
...
+	"os"
+
+	"github.com/ActiveState/cli/internal/constants"
 	"golang.org/x/sys/windows"
 )
 
+const EnvOverrideAdmin = "ACTIVESTATE_CLI_ISADMIN_OVERRIDE"
+
 func IsAdmin() (bool, error) {
+	// Work around weird CI bug: DX-1329
```
mitchell-as, line 16: Would you please put a one-sentence summary so I or anyone else does not have to search JIRA for this ticket?

test/integration/e2eissues_int_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/ActiveState/cli/internal/testhelpers/e2e"
+)
+
+/*
+Tests in this file are failing due to bugs in the test framework. They are skipped so as not to break our CI.
+The intend is to collect e2e bugs here so that we can test that they are fixed once we refactor the framework.
+*/
+
+func TestMultipleSends(t *testing.T) {
+	t.Skip("")
```
mitchell-as, line 19: Should we have a skip reason? Or is the comment above enough?

test/integration/offinstall_int_test.go
```diff
@@ -65,22 +61,21 @@ func (suite *OffInstallIntegrationTestSuite) TestInstallAndUninstall() {
 	defaultInstallDir := filepath.Join(defaultInstallParentDir, "IntegrationTest")
 
 	env := []string{
-		"ACTIVESTATE_CLI_DISABLE_RUNTIME=false",
-		"HOME=" + homeDir,
+		constants.DisableRuntime + "=false",
+		constants.IsAdminOverrideEnvVarName + "=false",
+		"VERBOSE=true",
```
mitchell-as, line 66: Just double-checking that HOME isn't needed anymore, as it was removed.

test/integration/offinstall_int_test.go
```diff
@@ -127,15 +122,18 @@ func (suite *OffInstallIntegrationTestSuite) TestInstallAndUninstall() {
...
-		tp.Expect("Enter an installation directory to uninstall")
-		tp.SendLine(defaultInstallDir)
+		tp.Expect("continue?")
+		tp.SendLine("y")
 		tp.Expect("Uninstall Complete", 5*time.Second)
 		tp.Expect("Press enter to exit")
 		tp.SendLine("")
 		tp.ExpectExitCode(0)
 
+		fmt.Println(tp.Snapshot())
```
mitchell-as, line 135: I don't think you meant to keep this.

mitchell-as (changes requested):

test/integration/e2eissues_int_test.go
Naatan: I feel like the comment above is enough, since the intend of this entire test suite is to be skipped until we're ready to run them again.

test/integration/offinstall_int_test.go
Naatan: Yep. On linux the current home detection doesn't use the HOME env var at all, so this just caused inconsistency in the tests.

mitchell-as (approved):

PR #2168, opened 2022-10-19
Do not allow deploying in the root directory.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1289" title="DX-1289" target="_blank">BugDX-1289</a>  State Deploy Failure On Ubuntu
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	rt "runtime"
 	"strings"
 	"sync"
 
@@ -163,6 +164,13 @@ func NewWithModel(target Targeter, msgHandler Events, model ModelProvider, an an
 
 // Update installs the runtime locally (or updates it if it's already partially installed)
 func (s *Setup) Update() error {
+	// Do not allow users to deploy runtimes to the root directory (this can easily happen in docker
+	// images). Note that runtime targets are fully resolved via fileutils.ResolveUniquePath(), so
+	// paths like "/." and "/opt/.." resolve to simply "/" at this time.
+	if (rt.GOOS != "windows" && s.target.Dir() == "/") {
+		return locale.NewInputError("err_runtime_setup_root", "Cannot set up a runtime in the root directory. Please try in another directory.")
+	}
+
 	// Update all the runtime artifacts
 	artifacts, err := s.updateArtifacts()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


MDrakos (approved):
Looks fine to me but you may want to get Nathan and/or Pete to take a look at the error message

Naatan (changes requested):
The person encountering the bug did not seem to have ran deploy with a custom target path, meaning that something in our implementation defaulted to the root. I suspect this might be a case of the `HOME` env var not being set, but worth digging into that.

mitchell-as:
@Naatan This is what happens in the docker file mentioned in the ticket. The command is run at root (/) so the deployment path was '.', which is root (/).

Naatan:
@mitchell-as I see.. in that case can we just move the logic to here:

https://github.com/ActiveState/cli/blob/6adb01cf7a668ad1797e35108ea24b1b5e0794de/pkg/platform/runtime/setup/setup.go#L166-L166

That way it'll apply to all runtime usage. I don't think we're ever going to want to install to root.

Naatan:
Also I think we need to resolve the path, as right now we wouldn't catch something like `/` or `../` from `/opt`.

mitchell-as:
We are resolving it. `NewCustomTarget()` is calling `fileutils.ResolveUniqPath()`, which does what you are expecting. That is why I was comfortable with `== "/"`.

mitchell-as:
Okay, moved the check into setup. Not sure about the wording though. Feel free to suggest something better.

pkg/platform/runtime/setup/setup.go
```diff
@@ -163,6 +164,13 @@ func NewWithModel(target Targeter, msgHandler Events, model ModelProvider, an an
 
 // Update installs the runtime locally (or updates it if it's already partially installed)
 func (s *Setup) Update() error {
+	// Do not allow users to deploy runtimes to the root directory (this can easily happen in docker
+	// images). Note that runtime targets are fully resolved via fileutils.ResolveUniquePath(), so
+	// paths like "/." and "/opt/.." resolve to simply "/" at this time.
+	if (rt.GOOS != "windows" && s.target.Dir() == "/") {
+		return locale.NewInputError("err_runtime_setup_root", "Cannot set up a runtime in the root directory. Please try in another directory.")
```
Naatan, line 171:
```suggestion
		return locale.NewInputError("err_runtime_setup_root", "Cannot set up a runtime in the root directory. Please specify or run from a user-writable directory.")
```

Naatan (changes requested):

Naatan (approved):

PR #2169, opened 2022-10-19
Do not forward io.Writer errors to Rollbar.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1273" title="DX-1273" target="_blank">BugDX-1273</a>  state-installer []: Writing colored output failed: write /dev/stdout: The pipe is being closed
  </td></table>
  <br />

### DIFF
--- a/internal/output/plain.go
+++ b/internal/output/plain.go
@@ -101,7 +101,7 @@ func (f *Plain) write(writer io.Writer, value interface{}) {
 func (f *Plain) writeNow(writer io.Writer, value string) {
 	_, err := colorize.Colorize(wordWrap(value), writer, !f.cfg.Colored)
 	if err != nil {
-		multilog.Log(logging.ErrorNoStacktrace, rollbar.Error)("Writing colored output failed: %v", err)
+		logging.ErrorNoStacktrace("Writing colored output failed: %v", err)
 	}
 }
 
--- a/internal/terminal/terminal.go
+++ b/internal/terminal/terminal.go
@@ -11,7 +11,7 @@ import (
 // fdSupportsColors implements a heuristic checking whether a file descriptor supports colors
 func fdSupportsColors(fd int, lookupEnv func(string) (string, bool)) bool {
 	if runtime.GOOS == "windows" {
-		return true
+		return terminal.IsTerminal(fd)
 	}
 	termValue, ok := lookupEnv("TERM")
 	if !ok {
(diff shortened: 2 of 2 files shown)


daved (dismissed):

mitchell-as:
Upon further review this is not correct, because io.Writer is an interface and we cannot make any assuptions about an error coming from `io.Writer.Write()`.

mitchell-as:
Upon re-review of this ticket's ACs, I'm going with the "when all else fails, silence the error" part. This particular error is a Windows I/O error 232: ERROR_NO_DATA. I don't know if it's worth specifically handling this case, as it becomes a slippery slop of what types of I/O errors to handle, and if we handle it here, should we be handling these types of io.Writer.Write() errors elsewhere in the codebase?

Naatan (changes requested):
Did you check the first AC? 

> Ensure that our code for detecting whether colored output is supported is working as expected

It seems that the issue might stem from us never passing `true` for `checkTerminal`:

https://github.com/ActiveState/cli/blob/3a08029a3f8fcfecd3e89d5d2b18b86a1e394935/cmd/state/output.go#L33-L33

eg.

https://github.com/ActiveState/cli/blob/3a08029a3f8fcfecd3e89d5d2b18b86a1e394935/cmd/state/output.go#L58-L58

https://github.com/ActiveState/cli/blob/3a08029a3f8fcfecd3e89d5d2b18b86a1e394935/cmd/state/output.go#L76-L76

I think passing `true` here might address the issue.

mitchell-as:
Yes I did. I've been running the state tool a lot in Git Bash on Windows (which doesn't have color support) and have not run into anything weird. I *suppose* it is possible that a color escape sequence could be interpreted as an EOF sequence, so I've made a change to not always assume color is available on Windows, but I am not confident that this is the solution to the rollbar bug.

Naatan:
> Yes I did. I've been running the state tool a lot in Git Bash on Windows (which doesn't have color support) and have not run into anything weird. I _suppose_ it is possible that a color escape sequence could be interpreted as an EOF sequence, so I've made a change to not always assume color is available on Windows, but I am not confident that this is the solution to the rollbar bug.

I see. Do you think it makes sense to include the changes I suggested? It still seems relevant to me.

mitchell-as:
`checkTerminal` is `true` by default: https://github.com/ActiveState/cli/blob/3a08029a3f8fcfecd3e89d5d2b18b86a1e394935/cmd/state/output.go#L31, so `terminal.StdoutSupportsColors()` was always being called, hence my downstream change to not always return `true` for Windows.

Naatan:
Oh I see, I had it inverted. Ok makes sense then, thanks for explaining!

Naatan (approved):

PR #2170, opened 2022-10-19
Make inability to stop state-svc when installing an input error (due …

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1279" title="DX-1279" target="_blank">BugDX-1279</a>  state-installer []: Installer error: Failed to stop running services: ... Could not kill child pro
  </td></table>
  <br />
 

…to lack of permissions).

Do not report this to rollbar.

### DIFF
--- a/internal/installmgr/stop.go
+++ b/internal/installmgr/stop.go
@@ -143,7 +143,7 @@ func killProcess(proc *process.Process, name string) error {
 		for _, c := range children {
 			err = c.Kill()
 			if err != nil {
-				return errs.WrapInputError(err, "err_stop_svc_permissions")
+				return locale.WrapInputError(err, "err_stop_svc_permissions")
 			}
 		}
 	} else {
@@ -152,7 +152,7 @@ func killProcess(proc *process.Process, name string) error {
 
 	err = proc.Kill()
 	if err != nil {
-		return errs.WrapInputError(err, "err_stop_svc_permissions")
+		return locale.WrapInputError(err, "err_stop_svc_permissions")
 	}
 
 	logging.Debug("Stopped %s process with SIGKILL", name)
--- a/internal/installmgr/stop.go
+++ b/internal/installmgr/stop.go
@@ -143,7 +143,7 @@ func killProcess(proc *process.Process, name string) error {
 		for _, c := range children {
 			err = c.Kill()
 			if err != nil {
-				return errs.Wrap(err, "Could not kill child process of %s", name)
+				return errs.WrapInputError(err, "err_stop_svc_permissions")
 			}
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


MDrakos (commented):
I like what you've added to the error message but do we know _for sure_ that we're getting this error due to insufficient permissions?

mitchell-as:
Well, the ticket wants the error message I replaced to be this new one about permissions. Also, the only things that error in `stopSvc()` are due to permissions (cannot list processes, cannot get the path of a process, cannot kill a process, ...), so I think it's a reasonable assumption that if there is an error stopping the service, it's due to permissions.

MDrakos:
Fair enough. I'll approve but please get another set of eyes on this PR as I feel we aren't being transparent here.

I recall there being other errors related to the service or updating where we say that the process failed and then ask the user if they are running the tool in a way that could prevent it from functioning properly. That may be an alternative.

MDrakos (dismissed):

Naatan (changes requested):
This is assuming that the error returned is an access denied error. It shouldn't be an assumption, we should actually assert it and only then give them the input error.

Naatan (changes requested):
I'm sorry but this is still making assumptions about the cause of the error. Ideally, I'd see an `errors.Is()` or an `errors.As()` in here. If the only way to do this is with string matching then let's chat about this during stand tomorrow.

mitchell-as:
I'm not trying to start an argument. I'd only like to point out that if we've detected a running state-svc process, and if we cannot kill it via a 3rd-party package, I think it's a reasonable assumption that it's due to a permission issue.

Since we want to be all pedantic about it, I'll have to go through the 3rd party library to identify how processes are killed on each of the three platforms, identify which Go calls are made to do the killing, and then go parse the Go docs to figure out if there are specific permission related errors to test for. Then I'll have to go and try to test this trying to make sure that we only return a permission error for one case, but a standard error for another case (which is to-be-determined). I personally do not think this is a reasonable way to spend one's time for a bug of this magnitude, however I have been very wrong in the past about a lot of things, so I might be missing something obvious. Sorry if that is the case.

Naatan:
> I'm not trying to start an argument. I'd only like to point out that if we've detected a running state-svc process, and if we cannot kill it via a 3rd-party package, I think it's a reasonable assumption that it's due to a permission issue.
> 
> Since we want to be all pedantic about it, I'll have to go through the 3rd party library to identify how processes are killed on each of the three platforms, identify which Go calls are made to do the killing, and then go parse the Go docs to figure out if there are specific permission related errors to test for. Then I'll have to go and try to test this trying to make sure that we only return a permission error for one case, but a standard error for another case (which is to-be-determined). I personally do not think this is a reasonable way to spend one's time for a bug of this magnitude, however I have been very wrong in the past about a lot of things, so I might be missing something obvious. Sorry if that is the case.

Yeah this is what I was trying to caution us about by saying "let's chat in stand". Basically I am asking to see if there is an error type we can check for, which should be relatively easy to find out. If that's not possible and we need something more complex then we ought to go back to the drawing board.

> I think it's a reasonable assumption that it's due to a permission issue

Personally, I don't feel like I have enough knowledge about the things that can go wrong when killing processes to make this a reasonable assumption from my end. Do you feel your experience is sufficient enough that this is a reasonable assumption on your end? I'm happy to defer to your better judgement here, it's just that to me this looks like a risky assumption.

PR #2171, opened 2022-10-19
Do not report inability to get state-svc executable path to rollbar.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1284" title="DX-1284" target="_blank">BugDX-1284</a>  state-installer [-t -f --source-installer]: Could not get executable path for state-svc process, error: bad call to lsof
  </td></table>
  <br />
 

This is likely due to a permissions issue. Try to stop the process anyway. Upon the likely failure, this will ultimately be reported as an input error per DX-1279.

### DIFF
--- a/internal/installmgr/stop.go
+++ b/internal/installmgr/stop.go
@@ -83,14 +83,14 @@ func stopSvc(installPath string) error {
 		svcName := constants.ServiceCommandName + exeutils.Extension
 		if n == svcName {
 			exe, err := p.Exe()
-			if err != nil {
-				multilog.Error("Could not get executable path for state-svc process, error: %v", err)
-				continue
-			}
-
-			if !strings.Contains(strings.ToLower(exe), "activestate") {
-				multilog.Error("Found state-svc process in unexpected directory: %s", exe)
-				continue
+			if err == nil {
+				if !strings.Contains(strings.ToLower(exe), "activestate") {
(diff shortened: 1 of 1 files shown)


daved (approved):

internal/installmgr/stop.go
```diff
@@ -83,14 +83,14 @@ func stopSvc(installPath string) error {
...
-				multilog.Error("Found state-svc process in unexpected directory: %s", exe)
-				continue
+			if err == nil {
+				if !strings.Contains(strings.ToLower(exe), "activestate") {
+					multilog.Error("Found state-svc process in unexpected directory: %s", exe)
+					continue
+				}
+			} else {
+				logging.Debug("Could not get executable path for state-svc process, error: %v", err) // permissions issue
+				exe = "<unknown>"
```
Naatan, line 93: Couple issues I see here:

- It seems `continue` would be appropriate here? At the moment it's going to stop the process despite not being able to identify it. That's very dangerous.
- The exe value isn't used, so pointless to set it.
- We are again making assumptions about an error being due to permissions. If we have to make this assumption somewhere so be it, but please add it as low in the stack as possible and communicate it with a typed error

daved: This also caught my eye. After looking a while, I thought it did make sense, but I may have just been being lazy.

PR #2172, opened 2022-10-20
svcModel is set on construction and only nil on CI.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1305" title="DX-1305" target="_blank">BugDX-1305</a>  Rollbar: exec: error during analytics.sendEvent: Could not send analytics event, not connected to state-svc yet
  </td></table>
  <br />
 

There was some refactoring in this area about a year ago and the svcModel is now being set on the construction of the analytics client, so that is the reasoning behind this change. The rollbar errors are from an earlier version of the state tool prior to this refactor where svcModel was being set sometime after construction when analytics events could be coming in before an svcModel was available.

### DIFF
--- a/internal/analytics/client/async/client.go
+++ b/internal/analytics/client/async/client.go
@@ -93,7 +93,7 @@ func (a *Client) Wait() {
 }
 
 func (a *Client) sendEvent(category, action, label string, dims ...*dimensions.Values) error {
-	if condition.InUnitTest() {
+	if a.svcModel == nil { // this is only true on CI
 		return nil
 	}
 
--- a/internal/analytics/client/async/client.go
+++ b/internal/analytics/client/async/client.go
@@ -93,6 +93,10 @@ func (a *Client) Wait() {
 }
 
 func (a *Client) sendEvent(category, action, label string, dims ...*dimensions.Values) error {
+	if condition.InUnitTest() {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


internal/analytics/client/async/client.go
```diff
@@ -93,6 +93,10 @@ func (a *Client) Wait() {
 }
 
 func (a *Client) sendEvent(category, action, label string, dims ...*dimensions.Values) error {
+	if condition.InUnitTest() {
+		return nil
+	}
```
daved, line 98: It may be worth making this more general with `if a.svcModel == nil {`

daved (changes requested):

internal/analytics/client/async/client.go
daved: Also, to note (I feel like I might have mentioned this before), I would normally leave this as a nit message and approve the PR so that, if you don't care to make the change, you can merge right away. Though, I've seen your PRs with auto-merging, so I am hesitant and mark my nits with "request changes". Just be aware that changes are not necessarily expected when the only comments made are nits.

mitchell-as: Fair enough. I've also annotated it with a comment.

daved (approved):

PR #2177, opened 2022-10-21
VerifyPR fixes for RC's

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1333" title="DX-1333" target="_blank">TaskDX-1333</a>  Fix verifyPR issues
  </td></table>
  <br />

### DIFF
--- a/scripts/internal/workflow-helpers/jira.go
+++ b/scripts/internal/workflow-helpers/jira.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ActiveState/cli/internal/testhelpers/secrethelper"
 	"github.com/andygrunwald/go-jira"
 	"github.com/blang/semver"
+	"github.com/google/go-github/v45/github"
 )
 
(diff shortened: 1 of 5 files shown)


scripts/internal/workflow-helpers/jira.go
```diff
@@ -101,3 +102,15 @@ func IsMergedStatus(status string) bool {
 	}
 	return false
 }
+
+func FetchJiraIDsInCommits(commits []*github.RepositoryCommit) []string {
```
MDrakos, line 106: It doesn't look like this function is used anywhere except in a test. Is there any reason to keep it?

MDrakos (approved):

scripts/internal/workflow-helpers/jira.go
Naatan: Yes, to help me debug issues if/when they occur :p

PR #2181, opened 2022-10-24
Log svc comms err in executor rather than fail

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1334" title="DX-1334" target="_blank">BugDX-1334</a>  state-exec failed because it can't reach state-svc
  </td></table>
  <br />


Naatan (approved):

Naatan (changes requested):
Let's chat after stand. I'm not necessarily against what's happening here, but this is not what I thought we talked about. And so I'm missing some perspective.

cmd/state-exec/main.go
```diff
@@ -85,7 +85,11 @@ func run() error {
 
 	logr.Debug("communications - sock: %s", meta.SockPath)
 	if err := sendMsgToService(meta.SockPath, hb); err != nil {
-		return fmt.Errorf("cannot send message to service: %w", err)
+		logr.Debug("                 sock - error: %v", err)
+
+		if onCI() { // halt control flow on CI only
+			return fmt.Errorf("cannot send message to service: %w", err)
+		}
```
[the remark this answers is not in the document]
daved: We did speak about this after stand and everything was fine for this PR.

pkg/platform/model/svc.go
```diff
@@ -38,7 +41,21 @@ func (m *SvcModel) EnableDebugLog() {
...
+
+	err := m.client.RunWithContext(ctx, request, resp)
+	if err != nil {
+		reqError := &gqlclient.RequestError{}
+		if errors.As(err, &reqError) {
+			logging.Debug(
+				"svc client gql request failed - query: %q, vars: %q",
+				reqError.Request.Query(),
+				jsonFromMap(reqError.Request.Vars()),
+			)
```
Naatan, line 53: This feels like it could easily end up leaking sensitive information via log files and rollbar. Maybe we can make it only report the vars for local builds and while running in tests?

pkg/platform/model/svc.go
```diff
@@ -133,3 +150,20 @@ func (m *SvcModel) ConfigChanged(ctx context.Context, key string) error {
...
+}
+
+func stringFromMap(m map[string]interface{}) string {
+	var s, sep string
+	for k, v := range m {
+		s += fmt.Sprintf("%s%s: %#v", sep, k, v)
+		sep = ", "
+	}
+	return s
+}
```
Naatan, line 169: Doesn't look like these are being used?

Naatan (changes requested):

pkg/platform/model/svc.go
daved: These are used to make the debugging output more human readable - json first, then a basic string fallback.

Naatan (approved):

PR #2186, opened 2022-10-25
Replace fixVersion=Any with Next Feasible and Next Unscheduled

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1260" title="DX-1260" target="_blank">StoryDX-1260</a>  fixVersion "Any" is replaced with something less implicit
  </td></table>
  <br />

### DIFF
--- a/scripts/ci/propagate-pr/main.go
+++ b/scripts/ci/propagate-pr/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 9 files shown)


MDrakos (approved):
This makes sense to me.

It looks like the version `Any` is still available on our Jira tickets. Should we remove that? I imagine it could cause issues once this is merged.

Naatan:
> This makes sense to me.
> 
> It looks like the version `Any` is still available on our Jira tickets. Should we remove that? I imagine it could cause issues once this is merged.

Yep.. I'll follow up with that soon once these changes are live.

PR #2187, opened 2022-10-26
Test remote installer

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1339" title="DX-1339" target="_blank">TaskDX-1339</a>  We have test plans covering the remote installer
  </td></table>
  <br />
 

Tests surfaced some small issues, which I decided to address while I was at it as otherwise it'd balloon into a timesink to address individually.

### DIFF
--- a/cmd/state-remote-installer/main.go
+++ b/cmd/state-remote-installer/main.go
@@ -198,7 +198,7 @@ func execute(out output.Outputer, prompt prompt.Prompter, cfg *config.Instance, 
 		return errs.Wrap(err, "Could not run installer")
 	}
 
-	out.Print(locale.Tl("remote_install_exit_prompt", "Installation complete. Press enter to exit."))
+	out.Print(locale.Tl("remote_install_exit_prompt", "Press ENTER to exit."))
 	fmt.Scanln(p.StrP("")) // Wait for input from user
 
 	return nil
--- a/cmd/state-remote-installer/main.go
+++ b/cmd/state-remote-installer/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/debug"
+	"syscall"
 	"time"
 
 	"github.com/ActiveState/cli/internal/analytics"
@@ -143,7 +144,9 @@ func main() {
 		}
 
 		exitCode, err = errors.Unwrap(err)
-		out.Error(err)
+		if err != nil {
+			out.Error(err)
+		}
 		return
 	}
 }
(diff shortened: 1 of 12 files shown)


cmd/state-remote-installer/main.go
```diff
@@ -193,12 +196,17 @@ func execute(out output.Outputer, prompt prompt.Prompter, cfg *config.Instance,
 	env := []string{
 		constants.InstallerNoSubshell + "=true",
 	}
-	_, _, err = exeutils.ExecuteAndPipeStd(filepath.Join(tmpDir, constants.StateInstallerCmd+exeutils.Extension), args, env)
+	_, cmd, err := exeutils.ExecuteAndPipeStd(filepath.Join(tmpDir, constants.StateInstallerCmd+exeutils.Extension), args, env)
 	if err != nil {
+		if cmd != nil && cmd.ProcessState.Sys().(syscall.WaitStatus).Exited() {
```
MDrakos, line 201: Isn't there a possibility of a panic calling `.Exited()` if the type check failed?

test/integration/remote_installer_int_test.go
```diff
@@ -15,23 +17,75 @@ import (
...
+			cp.Expect("Press ENTER to exit")
+			cp.SendLine("")
+			cp.ExpectExitCode(0)
+
+			suite.Require().FileExists(stateExePath)
+
+			cp = ts.SpawnCmdWithOpts(
+				stateExePath,
+				e2e.WithArgs("--version"),
+				e2e.AppendEnv(constants.InstallPathOverrideEnvVarName+"="+installPath),
```
MDrakos, line 77: nit: This shouldn't be needed here

MDrakos (changes requested):
Do we want to start phasing out the install scripts integration tests with this PR or wait?

cmd/state-remote-installer/main.go
Naatan: I thought so too, but this is what Go does:

https://github.com/golang/go/blob/master/src/os/exec_posix.go#L102

test/integration/remote_installer_int_test.go
Naatan: It's needed to keep the tests from conflicting with one another. Without this, it always installs in a global directory, which given the asynchronous nature of tests can cause all sorts of issues.

Naatan: Oh you meant it's not needed for `--version`; I don't think that's true either though, as we rely on the install path when calling the svc.

MDrakos: Ahh okay, makes sense 👍

MDrakos (approved):

PR #2188, opened 2022-10-26
Update all available RC files

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1310" title="DX-1310" target="_blank">StoryDX-1310</a>  We configure all shells that were found on the system
  </td></table>
  <br />

### DIFF
--- a/internal/globaldefault/default.go
+++ b/internal/globaldefault/default.go
@@ -53,7 +53,7 @@ func Prepare(cfg DefaultConfigurer, subshell subshell.SubShell) error {
 		"PATH": binDir,
 	}
 
-	if err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true); err != nil {
+	if err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, true); err != nil {
 		return locale.WrapError(err, "err_globaldefault_update_env")
 	}
 
@@ -106,7 +106,7 @@ func ResetDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer) (
 	}
 
 	envUpdates := map[string]string{}
-	err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true)
+	err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, true)
 	if err != nil {
 		return false, locale.WrapError(err, "err_globaldefault_update_env")
 	}
--- a/internal/globaldefault/default.go
+++ b/internal/globaldefault/default.go
@@ -28,7 +28,7 @@ func BinDir() string {
 	return storage.GlobalBinDir()
 }
 
-func Prepare(cfg DefaultConfigurer, subshell subshell.SubShell) error {
+func Prepare(cfg DefaultConfigurer, shell subshell.SubShell) error {
 	logging.Debug("Preparing globaldefault")
 	binDir := BinDir()
 
@@ -53,8 +53,17 @@ func Prepare(cfg DefaultConfigurer, subshell subshell.SubShell) error {
 		"PATH": binDir,
 	}
 
-	if err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, true); err != nil {
+	if err := shell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, true); err != nil {
 		return locale.WrapError(err, "err_globaldefault_update_env")
+	}
+
+	// Configure available shells
+	for _, shell := range subshell.AvailableShells() {
+		err = shell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, false)
+		if err != nil {
+			logging.Error("Could not update PATH for shell %s: %v", shell.Shell(), err)
+			continue
+		}
 	}
 
 	return nil
--- a/internal/globaldefault/default.go
+++ b/internal/globaldefault/default.go
@@ -53,17 +53,10 @@ func Prepare(cfg DefaultConfigurer, shell subshell.SubShell) error {
 		"PATH": binDir,
 	}
 
-	if err := shell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, true); err != nil {
-		return locale.WrapError(err, "err_globaldefault_update_env")
-	}
-
 	// Configure available shells
-	for _, shell := range subshell.AvailableShells() {
-		err = shell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, false)
-		if err != nil {
-			logging.Error("Could not update PATH for shell %s: %v", shell.Shell(), err)
-			continue
-		}
+	err = subshell.ConfigureAvailableShells(cfg, envUpdates, sscommon.DefaultID, true)
+	if err != nil {
+		return locale.WrapError(err, "err_globaldefault_update_env")
 	}
 
 	return nil
@@ -100,7 +93,7 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 	return nil
 }
 
-func ResetDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer) (bool, error) {
+func ResetDefaultActivation(shell subshell.SubShell, cfg DefaultConfigurer) (bool, error) {
 	logging.Debug("Resetting globaldefault")
 
 	projectDir := cfg.GetString(constants.GlobalDefaultPrefname)
@@ -115,7 +108,9 @@ func ResetDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer) (
 	}
 
 	envUpdates := map[string]string{}
-	err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true, true)
+
+	// Configure available shells
+	err := subshell.ConfigureAvailableShells(cfg, envUpdates, sscommon.DefaultID, true)
 	if err != nil {
 		return false, locale.WrapError(err, "err_globaldefault_update_env")
 	}
--- a/internal/globaldefault/default.go
+++ b/internal/globaldefault/default.go
@@ -28,7 +28,7 @@ func BinDir() string {
 	return storage.GlobalBinDir()
 }
 
-func Prepare(cfg DefaultConfigurer, shell subshell.SubShell) error {
+func Prepare(cfg DefaultConfigurer) error {
 	logging.Debug("Preparing globaldefault")
 	binDir := BinDir()
 
@@ -63,9 +63,9 @@ func Prepare(cfg DefaultConfigurer, shell subshell.SubShell) error {
 }
 
 // SetupDefaultActivation sets symlinks in the global bin directory to the currently activated runtime
-func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, runtime *runtime.Runtime, proj *project.Project) error {
+func SetupDefaultActivation(cfg DefaultConfigurer, runtime *runtime.Runtime, proj *project.Project) error {
 	logging.Debug("Setting up globaldefault")
-	if err := Prepare(cfg, subshell); err != nil {
+	if err := Prepare(cfg); err != nil {
 		return locale.WrapError(err, "err_globaldefault_prepare", "Could not prepare environment.")
 	}
 
@@ -93,7 +93,7 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 	return nil
 }
 
-func ResetDefaultActivation(shell subshell.SubShell, cfg DefaultConfigurer) (bool, error) {
+func ResetDefaultActivation(cfg DefaultConfigurer) (bool, error) {
 	logging.Debug("Resetting globaldefault")
 
 	projectDir := cfg.GetString(constants.GlobalDefaultPrefname)
--- a/internal/globaldefault/default.go
+++ b/internal/globaldefault/default.go
@@ -54,7 +54,8 @@ func Prepare(cfg DefaultConfigurer) error {
 	}
 
 	// Configure available shells
-	err = subshell.ConfigureAvailableShells(cfg, envUpdates, sscommon.DefaultID, true)
+	shell := subshell.New(cfg)
+	err = subshell.ConfigureAvailableShells(shell, cfg, envUpdates, sscommon.DefaultID, true)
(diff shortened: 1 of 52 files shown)


test/integration/prepare_int_test.go
```diff
@@ -92,7 +92,7 @@ func (suite *PrepareIntegrationTestSuite) AssertConfig(target string) {
 		suite.Require().NoError(err)
 
 		subshell := subshell.New(cfg)
-		rcFile, err := subshell.RcFile()
+		rcFile, err := subshell.RcFile(false)
```
mitchell-as, line 95: I'm seeing a lot of churn in this diff related to the function signature change. I wonder if it may be more advantageous to have a new function specifically for creating the file if it already exists. (e.g. `subshell.RcFileEnsureExists()`) I'm just thinking out loud though. Not sure if this will help or just cause more churn elsewhere.

test/integration/use_int_test.go
```diff
@@ -126,6 +127,10 @@ func (suite *UseIntegrationTestSuite) TestReset() {
 	cp.Expect("Checked out project")
 	cp.ExpectExitCode(0)
 
+	zsh := &zsh.SubShell{}
```
mitchell-as, line 130: Technically this will be created on Windows. Do we care? Perhaps we do.
Please also consider adding a comment annotating why you're creating this file and what you intend to do with it. It's a bit jarring to see a zshrc created out of nowhere for a reset test.

internal/subshell/fish/fish.go
```diff
@@ -95,13 +95,21 @@ func (v *SubShell) WriteCompletionScript(completionScript string) error {
...
 
-func (v *SubShell) RcFile() (string, error) {
+func (v *SubShell) RcFile(create bool) (string, error) {
 	homeDir, err := fileutils.HomeDir()
 	if err != nil {
 		return "", errs.Wrap(err, "IO failure")
 	}
 
-	return filepath.Join(homeDir, ".config/fish/config.fish"), nil
+	rcFilePath := filepath.Join(homeDir, ".config/fish/config.fish")
```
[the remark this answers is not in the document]
MDrakos: I've implemented a helper function in `fileutils` however there is still repetition in calling that helper function.

test/integration/prepare_int_test.go
MDrakos: That's a good point. After looking at this again I was able to get rid of the boolean and move the logic to a different function. Let me know what you think.

test/integration/use_int_test.go
```diff
@@ -127,9 +127,18 @@ func (suite *UseIntegrationTestSuite) TestReset() {
...
-	zshRcFile, err := zsh.RcFile(true)
-	suite.NoError(err)
+	// Create a zsh RC file to ensure that the path is updated for both bash and zsh.
+	var zshRcFile string
+	var err error
+	if runtime.GOOS != "windows" {
+		zsh := &zsh.SubShell{}
+		zshRcFile, err = zsh.RcFile()
+		suite.NoError(err)
+		if !fileutils.FileExists(zshRcFile) {
```
mitchell-as, line 137: Can you use your new fileutils.TouchFileUnlessExists function?

internal/subshell/subshell_lin_mac.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/ActiveState/cli/internal/subshell/bash"
+	"github.com/ActiveState/cli/internal/subshell/cmd"
+	"github.com/ActiveState/cli/internal/subshell/fish"
+	"github.com/ActiveState/cli/internal/subshell/tcsh"
+	"github.com/ActiveState/cli/internal/subshell/zsh"
+)
+
+func init() {
+	supportedShells = []SubShell{
```
mitchell-as, line 15: This is clever :)

mitchell-as (changes requested):
Almost there!

test/integration/prepare_int_test.go
mitchell-as: I think it's a reasonable trade-off. Nice :)

mitchell-as (dismissed):

internal/subshell/subshell.go
```diff
@@ -52,7 +54,7 @@ type SubShell interface {
 	SetBinary(string)
 
 	// WriteUserEnv writes the given env map to the users environment
-	WriteUserEnv(sscommon.Configurable, map[string]string, sscommon.RcIdentification, bool) error
+	WriteUserEnv(sscommon.Configurable, map[string]string, sscommon.RcIdentification, bool, bool) error
```
Naatan, line 57: I think the bool flag "create" here has the wrong connotation. The logic that this is facilitating is "only if the user has it". I'd rather see the implementation convey that logic.

Additionally, you are still calling WriteRCFile() even when create=false, which is banking on WriteRCFile() implicitly handing this scenario.

internal/subshell/subshell.go
```diff
@@ -140,6 +142,45 @@ func New(cfg sscommon.Configurable) SubShell {
...
+func availableShells() []SubShell {
+	var shells []SubShell
+	for _, shell := range supportedShells {
+		rcFile, err := shell.RcFile()
+		if err != nil {
+			logging.Error("Could not determine rc file for shell %s: %v", shell.Shell(), err)
+			continue
+		}
+
+		if !fileutils.FileExists(rcFile) {
```
Naatan, line 154:
```suggestion
		if !fileutils.TargetExists(rcFile) {
```

This could reasonably be a symlink.

internal/subshell/subshell.go
```diff
@@ -140,6 +142,45 @@ func New(cfg sscommon.Configurable) SubShell {
...
+		}
+
+		shells = append(shells, shell)
+	}
+
+	return shells
+}
+
+func ConfigureAvailableShells(cfg sscommon.Configurable, env map[string]string, identifier sscommon.RcIdentification, isAdmin bool) error {
+	activeShell := New(cfg)
```
Naatan, line 165: This feels backwards. You refactored a lot of code such that we don't have to pass subshell around, but are now constructing it ONLY for this function.

If we can't reasonably put this function as a method of SubShell then we should instead pass it as an argument.

internal/subshell/subshell.go
```diff
@@ -140,6 +142,45 @@ func New(cfg sscommon.Configurable) SubShell {
...
+	for _, s := range availableShells() {
+		isActive := false
+		if s.Shell() == activeShell.Shell() {
+			isActive = true
+		}
+
+		err := s.WriteUserEnv(cfg, env, identifier, isAdmin, isActive)
```
Naatan, line 172: I think it would be cleaner if we introduced something like `subshell.IsAvailable()`.

So you'd end up with:

```
if s.IsAvailable() {
  s.WriteUserEnv(...)
}
```

internal/subshell/subshell_lin_mac.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/ActiveState/cli/internal/subshell/bash"
+	"github.com/ActiveState/cli/internal/subshell/cmd"
+	"github.com/ActiveState/cli/internal/subshell/fish"
+	"github.com/ActiveState/cli/internal/subshell/tcsh"
+	"github.com/ActiveState/cli/internal/subshell/zsh"
+)
+
+func init() {
```
Naatan, line 14: I don't think you need an init to do what you're doing here.

test/integration/use_int_test.go
```diff
@@ -126,6 +127,17 @@ func (suite *UseIntegrationTestSuite) TestReset() {
 	cp.Expect("Checked out project")
 	cp.ExpectExitCode(0)
 
+	// Create a zsh RC file to ensure that the path is updated for both bash and zsh.
+	var zshRcFile string
+	var err error
+	if runtime.GOOS != "windows" {
+		zsh := &zsh.SubShell{}
```
Naatan, line 134: This is assuming that the test is ran as "not zsh". Suggest adding `SHELL=bash` to the env to make that assumption explicit.

Naatan (changes requested):

internal/globaldefault/default.go
```diff
@@ -53,17 +53,20 @@ func Prepare(cfg DefaultConfigurer, subshell subshell.SubShell) error {
 		"PATH": binDir,
 	}
 
-	if err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true); err != nil {
+	// Configure available shells
+	shell := subshell.New(cfg)
```
Naatan, line 57: This was receiving it as an argument before, not constructing it here.

The point of the change I requested is that we're not manually constructing subshell every time and instead pass it around.

internal/globaldefault/default.go
```diff
@@ -106,7 +109,10 @@ func ResetDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer) (
 	}
 
 	envUpdates := map[string]string{}
-	err := subshell.WriteUserEnv(cfg, envUpdates, sscommon.DefaultID, true)
+
+	// Configure available shells
+	shell := subshell.New(cfg)
```
Naatan, line 114: Pass by arg instead

internal/runners/deploy/deploy.go
```diff
@@ -206,7 +206,9 @@ func (d *Deploy) configure(namespace project.Namespaced, rtTarget setup.Targeter
 
 	d.output.Notice(output.Heading(locale.Tr("deploy_configure_shell", d.subshell.Shell())))
 
-	err = d.subshell.WriteUserEnv(d.cfg, env, sscommon.DeployID, userScope)
+	// Configure available shells
+	shell := subshell.New(d.cfg)
+	err = subshell.ConfigureAvailableShells(shell, d.cfg, env, sscommon.DefaultID, userScope)
```
Naatan, line 211:
```suggestion
	err = subshell.ConfigureAvailableShells(d.subshell, d.cfg, env, sscommon.DefaultID, userScope)
```

internal/subshell/subshell.go
```diff
@@ -68,6 +66,9 @@ type SubShell interface {
 	// RcFile return the path of the RC file
 	RcFile() (string, error)
 
+	// EnsureRcFile ensures that the RC file exists
+	EnsureRcFile() error
```
Naatan, line 70: Can we name this `EnsureRcFileExists`? This has me going.. "ensure RC file .. and then?" :p

image

internal/subshell/tcsh/tcsh.go
```diff
@@ -48,17 +49,11 @@ func (v *SubShell) SetBinary(binary string) {
 }
 
 // WriteUserEnv - see subshell.SubShell
-func (v *SubShell) WriteUserEnv(cfg sscommon.Configurable, env map[string]string, envType sscommon.RcIdentification, _ bool, create bool) error {
+func (v *SubShell) WriteUserEnv(cfg sscommon.Configurable, env map[string]string, envType sscommon.RcIdentification, _ bool) error {
```
Naatan, line 52: Why does the bool argument still exist?

internal/subshell/tcsh/tcsh.go
```diff
@@ -99,6 +94,15 @@ func (v *SubShell) RcFile() (string, error) {
...
 }
 
+func (v *SubShell) EnsureRcFile() error {
+	rcFile, err := v.RcFile()
+	if err != nil {
+		return errs.Wrap(err, "Could not determine rc file")
+	}
+
+	return fileutils.TouchFileUnlessExists(rcFile)
+}
```
Naatan, line 104: Nit: Maybe put the logic for this in sscommon? ie. have this call something like `sscommon.EnsureRCFilExists(v)`.

test/integration/use_int_test.go
```diff
@@ -121,13 +122,28 @@ func (suite *UseIntegrationTestSuite) TestReset() {
...
+		zshRcFile, err = zsh.RcFile()
+		suite.NoError(err)
+		err = fileutils.TouchFileUnlessExists(zshRcFile)
+		suite.NoError(err)
+	}
+
 	cp = ts.SpawnWithOpts(
 		e2e.WithArgs("use", "ActiveState-CLI/Python3"),
+		e2e.AppendEnv("SHELL=bash"),
```
Naatan, line 146: Only the test with the zsh assertion needs this.

test/integration/use_int_test.go
```diff
@@ -121,13 +122,28 @@ func (suite *UseIntegrationTestSuite) TestReset() {
...
+	)
 	cp.Expect("Skipping runtime setup")
 	cp.Expect("Checked out project")
 	cp.ExpectExitCode(0)
 
+	// Create a zsh RC file to ensure that the path is updated for both bash and zsh.
+	var zshRcFile string
+	var err error
+	if runtime.GOOS != "windows" {
+		zsh := &zsh.SubShell{}
```
Naatan, line 137: This test case (TestReset) is very over-loaded. Let's not make this problem worse and instead address the shell tests with their own test rather than piggy backing off of this one.

Naatan (changes requested):

test/integration/use_int_test.go
MDrakos: I believe that setting `SHELL` here ensures that `bash` is the active shell so that we can tests that the `.zshrc` is also updated.

internal/subshell/tcsh/tcsh.go
MDrakos: I tried that initially but none of these shell packages can import `sscommon` without an import cycle.

internal/subshell/tcsh/tcsh.go
MDrakos: That argument is `isAdmin` for the `cmd` shell. It was there originally.

test/integration/use_int_test.go
```diff
@@ -158,37 +143,67 @@ func (suite *UseIntegrationTestSuite) TestReset() {
...
+	cp.Expect("No global default project to reset")
+	cp.ExpectExitCode(0)
+
+	if runtime.GOOS != "windows" {
+		suite.NotContains(string(fileutils.ReadFileUnsafe(rcfile)), ts.Dirs.DefaultBin, "PATH still has default project in it")
+	}
+}
+
+func (suite *UseIntegrationTestSuite) TestUseShellUpdates() {
+	suite.OnlyRunForTags(tagsuite.Use)
```
Naatan, line 171: Let's move this to shell_int_test.go and use `tagsuite.Shell`. "state use" isn't what we're testing here, it's just what we're using to facilitate the test.

Naatan (changes requested):

Naatan (approved):

PR #2189, opened 2022-10-26
Initial support for bash on Windows.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-595" title="DX-595" target="_blank">StoryDX-595</a>  We support bash & zsh shells on Windows
  </td></table>
  <br />
 

zsh support has been de-scoped, as installing it on Windows is non-trivial. More non-trivial than via WSL, which we do not support.


internal/subshell/cmd/cmd.go
```diff
@@ -149,7 +149,7 @@ func (v *SubShell) Activate(prj *project.Project, cfg sscommon.Configurable, out
 	if prj != nil {
 		env := sscommon.EscapeEnv(v.env)
 		var err error
-		if v.rcFile, err = sscommon.SetupProjectRcFile(prj, "config.bat", ".bat", env, out, cfg); err != nil {
+		if v.rcFile, err = sscommon.SetupProjectRcFile(prj, "config.bat", ".bat", env, out, cfg, false); err != nil {
```
mitchell-as, line 152: Note: my initial thought was to pass in the subshell object (`v`), but this introduces an import cycle in `sscommon`, so I just pass a boolean.

cmd/state/main.go
```diff
@@ -246,6 +246,10 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 		if childCmd != nil {
 			cmdName = childCmd.UseFull() + " "
 		}
+		nonInteractiveError := &prompt.NonInteractiveError{}
+		if errors.As(err, &nonInteractiveError) {
```
MDrakos, line 250: Is there a reason to assign the error struct to a variable here?

internal/globaldefault/default.go
```diff
@@ -72,7 +73,8 @@ func SetupDefaultActivation(subshell subshell.SubShell, cfg DefaultConfigurer, r
 		return locale.WrapError(err, "err_globaldefault_rtexes", "Could not retrieve runtime executables")
 	}
 
-	env, err := runtime.Env(false, false)
+	bashifyPaths := rt.GOOS == "windows" && subshell.Shell() != "cmd"
```
MDrakos, line 76: Could this be a function of the subshell itself? (ie. `subshell.BashifyPaths()`)?

pkg/platform/runtime/runtime.go
```diff
@@ -152,6 +152,13 @@ func (r *Runtime) Env(inherit bool, useExecutors bool) (map[string]string, error
 	env := envDef.GetEnv(inherit)
 
 	execDir := filepath.Clean(setup.ExecDir(r.target.Dir()))
+	if bashifyPaths {
+		var err error
+		execDir, err = osutils.BashifyPath(execDir)
+		if err != nil {
+			return nil, errs.Wrap(err, "Could not bashify path: %s", execDir)
+		}
+	}
```
MDrakos, line 161: This could potentially be an `osutils` function

test/integration/shells_int_test.go
```diff
@@ -66,7 +66,6 @@ func (suite *ShellsIntegrationTestSuite) TestShells() {
 			}
 
 			// Verify the runtime is functioning properly.
-			cp.WaitForInput()
```
MDrakos, line 69: Was this removed for debugging or is it no longer needed?

internal/runners/deploy/deploy.go
```diff
@@ -173,7 +173,7 @@ func (d *Deploy) install(rtTarget setup.Targeter) error {
 	// Todo Remove with https://www.pivotaltracker.com/story/show/178161240
 	// call rti.Environ as this completes the runtime activation cycle:
 	// It ensures that the analytics event for failure / success are sent
-	_, _ = rti.Env(false, false)
+	_, _ = rti.Env(false, false, false)
```
MDrakos, line 176: There are a lot of cases here where we are defaulting to `false` for `bashifyPath` is there a reason why we don't want to bashify these paths?

MDrakos: When we have a lot of arguments that are just booleans it may be worth considering comments to explain them. Though I realize that's a lot of comment writing in this case so I'll leave it up to you.

MDrakos (changes requested):
Makes sense to me so far. Just some more minor comments to consider.

cmd/state/main.go
mitchell-as: Yes. Go wouldn't compile without it. I didn't want to use the variable :(

internal/globaldefault/default.go
mitchell-as: I'm not sure about that since it doesn't make sense for the cmd subshell?

MDrakos: I would think that a function like that for `cmd` would just return false, but I'll leave this up to you.

internal/runners/deploy/deploy.go
MDrakos: That's fair. Again, comments might be helpful.

MDrakos (approved):

internal/prompt/prompt.go
```diff
@@ -75,7 +79,7 @@ func (p *Prompt) InputAndValidate(title, message string, defaultResponse *string
 			logging.Debug("Selecting default choice %s for Input prompt %s in non-interactive mode", *defaultResponse, title)
 			return *defaultResponse, nil
 		}
-		return "", locale.NewInputError("err_non_interactive_prompt", message)
+		return "", &NonInteractiveError{locale.NewInputError("err_non_interactive_prompt", message)}
```
Naatan, line 82: `err_non_interactive_prompt` does not appear to exist.

cmd/state/main.go
```diff
@@ -246,6 +246,10 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
...
+			err = locale.WrapInputError(err, locale.T("err_non_interactive_terminal"))
+		}
```
Naatan, line 252: This feels way too specific to live in main.go. Let's just raise this in prompt?

internal/assets/contents/activestate.yaml.perl.tpl
```diff
@@ -44,4 +44,8 @@ events:
   # This is the ACTIVATE event, it will run whenever a new virtual environment is created (eg. by running `state activate`)
   # On Linux and macOS this will be ran as part of your shell's rc file, so you can use it to set up aliases, functions, environment variables, etc.
   - name: ACTIVATE
+    if: ne .Shell "cmd"
+    value: {{.LangExe}} $scripts.activationMessage.path._posix()
+  - name: ACTIVATE
+    if: eq .Shell "cmd"
```
Naatan, line 50: What's the relevance to this PR?

internal/locale/locales/en-us.yaml
```diff
@@ -1718,6 +1718,8 @@ cve_needs_authentication:
   other: You need to be authenticated in order to access vulnerability information about your project.
 auth_tip:
   other: Run `state auth` to authenticate.
+err_non_interactive_terminal:
+  other: You are running the State Tool from a non-interactive terminal, but interactive input was needed. Please re-run the State Tool from an interactive terminal. If you are using Git Bash on Windows for example, prefix your command with [ACTIONABLE]`winpty`[/RESET].
```
Naatan, line 1722: Can we link to some online article explaining the issue being worked around here?

internal/runbits/activation/activation.go
```diff
@@ -40,7 +40,8 @@ func ActivateAndWait(
 		}
 	}
 
-	ve, err := venv.GetEnv(false, true, projectDir)
+	bashifyPaths := rt.GOOS == "windows" && ss.Shell() != "cmd"
```
Naatan, line 43: Can we put `ss.Shell() != "cmd"` into subshell itself, and expose a method instead? eg. `ss.BashifyPaths()`.

internal/subshell/bash/bash.go
```diff
@@ -63,6 +63,10 @@ func (v *SubShell) WriteUserEnv(cfg sscommon.Configurable, env map[string]string
 		return errs.Wrap(err, "RcFile failure")
 	}
 
+	if _, pathExists := env["PATH"]; pathExists && runtime.GOOS == "windows" {
+		env["PATH"] = filepath.ToSlash(env["PATH"])
+	}
```
Naatan, line 68: Couple of issues:

- This isn't bashifying the path, it's just replacing the path separator characters. I can see you the actual bashify logic is in rcfile.go, so what is this trying to do?
- Comment explaining what it does would be helpful here
- You are calling `filepath.ToSlash()` on a value that is itself not a path, but contains a variety of paths. Especially on windows this could create some weird bugs.

I imagine we will have to unpack this and bashify each individual path. If this is the case keep in mind that `osutils.BashifyPath()` calls out to `bash -c`. So calling it individually for each path would make this very slow. We might want to implement a variation of this that can bashify multiple paths with one call.

internal/subshell/bash/bash.go
```diff
@@ -147,8 +151,9 @@ func (v *SubShell) Activate(proj *project.Project, cfg sscommon.Configurable, ou
 	// available project files require more intensive modification of shell envs
 	if proj != nil {
 		env := sscommon.EscapeEnv(v.env)
+		bashifyPaths := runtime.GOOS == "windows" && v.Shell() != "cmd"
```
Naatan, line 154: We're in bash.go. The `v.Shell() != "cmd"` assertion seems redundant.

pkg/platform/runtime/runtime.go
Naatan, line 161: Isn't this redundant? Ultimately this env info will be handled by subshell. I feel like the bashify stuff should probably be centralized in the subshell package.

Naatan (changes requested):

internal/assets/contents/activestate.yaml.perl.tpl
mitchell-as: If you activate a project in the Git Bash shell on Windows, the ACTIVATE event will try and use a cmd-style path instead of a bash-style path, resulting in a "path not found" type of error.

internal/prompt/prompt.go
mitchell-as: It already exists in the locale file and has been used in existing non-interactive prompt error messages.

internal/assets/contents/activestate.yaml.perl.tpl
Naatan: I see. Then I would ask for a change here too; bash support for windows mean that the end-user doesn't need to work around limitations with _posix(). So what you did here should be a non-issue if we supported bash.

mitchell-as: I wanted to do this, but the ScriptExpander doesn't have access to the underlying subshell. This would possibly be a significant change. I'll have to revisit this (it's been almost a week since I looked into it).

pkg/platform/runtime/runtime.go
mitchell-as: No, we need to catch it here. This is where we "// Override PATH entry with exec path", so that exec path needs to be explicitly bashified if needed. There is no notion of subshell at this point.

Naatan: I understand, what I meant is that the env that this returns is ultimately used through the subshell package. If not; can you point me at the code that's causing bash incompatibility that the subshell package is not addressing?

mitchell-as: Never mind. It looks like setting up executors appears to use an env without subshell, so we probably still need to bashify here in runtime env setup: https://github.com/ActiveState/cli/blob/5080107ce157556e7bbffb0314afdf9a8185896c/internal/globaldefault/default.go#L75

Naatan: For executors it shouldn't matter, they're executables. They should not care what shell invoked them. 

> Are you expecting the subshell to iterate through all of its environment variables, look for paths that should be bashified, and then do so? 

I think so. Open to suggestions but this seems like the way to do it.

> but I thought it'd be less error-prone to bashify paths at their source rather than downstream.

I feel like this is more error prone as now everything dealing with this data is dealing with bashified paths, rather than JUST the component that cares about this (ie. subshell).

internal/subshell/bash/bash.go
mitchell-as: This only happens when installing the State Tool from Git Bash on Windows, or when running state deploy. The incoming PATH contains only State Tool paths, so we are not modifying any existing Windows PATH entries. In my experiments with the installer, swapping '\' for '/' was enough, but this is not the case for deploy, so I will go back to splitting and bashifying each path component (which has only ever been 1 or 2 in my experiments).

Naatan: > In my experiments with the installer, swapping '\' for '/' was enough

This is not an assumption we can make. There will for certain be situations where this is not true.

internal/assets/contents/activestate.yaml.perl.tpl
mitchell-as: It looks like I managed to work in the subshell into the Expander context.

pkg/platform/runtime/runtime.go
mitchell-as: Subshells are now responsible for bashifying their environment variables. The bash subshell now does this.

internal/subshell/bash/bash.go
mitchell-as: Okay, each path component is now bashified.

mitchell-as:
Second attempt. I expect more churn.

internal/osutils/osutils.go
```diff
@@ -32,22 +33,28 @@ func CmdString(c *exec.Cmd) string {
...
-	if absolutePath[0:1] == "/" {
+	if len(absolutePath) > 0 && absolutePath[0:1] == "/" {
 		// Already the format we want
 		return absolutePath, nil
 	}
 
-	if absolutePath[1:2] != ":" {
+	if len(absolutePath) < 2 || absolutePath[1:2] != ":" {
 		// Check for windows style paths
 		return "", errs.New("Unrecognized absolute path format: %s", absolutePath)
```
Naatan, line 43: Wouldn't this almost certainly happen? Seems we would need to drop this if we want windows compatibility.

internal/osutils/osutils.go
```diff
@@ -32,22 +33,28 @@ func CmdString(c *exec.Cmd) string {
...
 	}
 
+	if strings.Contains(absolutePath, ";") {
+		return "", errs.New("Path is a list: %s", absolutePath)
+	}
```
Naatan, line 48: This will need to account for escaped characters if we feel we really need to assert this.

```
❯ touch "/tmp/foo;bar"
❯ ls /tmp/ | grep foo
foo;bar
```

internal/osutils/osutils.go
```diff
@@ -71,6 +78,24 @@ func winPathToLinPath(name string) (string, error) {
 	return path, nil
 }
 
+// BashifyPath takes a windows %PATH% list and turns it into a bash style PATH list.
+// e.g. C:\foo;C:\bar becomes /c/foo:/c/bar
+func BashifyPathEnv(pathList string) (string, error) {
+	if runtime.GOOS != "windows" {
+		return pathList, nil // already bashified
+	}
```
Naatan, line 86: Please use osutils_windows.go

internal/osutils/osutils.go
```diff
@@ -71,6 +78,24 @@ func winPathToLinPath(name string) (string, error) {
...
+
+	dirs := strings.Split(pathList, ";")
+	for i, dir := range dirs {
+		path, err := BashifyPath(dir)
```
Naatan, line 90: This will cause it to call `bash -c` in a loop for each path found, which will be slow and process intensive.

Might be worth researching some alternative solutions, but at the very least we could just resolve the first part of the path and apply the result of that to all matching path entries. In most cases that'd just be `C:\`.

internal/prompt/prompt.go
```diff
@@ -68,14 +70,25 @@ func (p *Prompt) Input(title, message string, defaultResponse *string, flags ...
 	}, flags...)
 }
 
+// interactiveInputError returns the proper input error for a non-interactive prompt.
+// If the terminal cannot show prompts (e.g. Git Bash on Windows), the error mentions this.
+// Otherwise, the error simply states the prompt cannot be resolved in non-interactive mode.
+// The "message" argument is the prompt's user-facing message.
+func interactiveInputError(message string) error {
+	if !term.IsTerminal(int(os.Stdin.Fd())) {
```
Naatan, line 78: What's the distinction between a non-interactive prompt and a non-interactive terminal? This seems like it might be worth passing along like we do non-interactive, rather than resolving it through use of a global.

internal/subshell/bash/bash.go
```diff
@@ -131,6 +140,20 @@ func (v *SubShell) SetupShellRcFile(targetDir string, env map[string]string, nam
...
+	for key, value := range env {
+		var bashified string
+		var err error
+		if key == "PATH" {
+			bashified, err = osutils.BashifyPathEnv(value)
+		} else {
+			bashified, err = osutils.BashifyPath(value)
+		}
+		if err != nil {
+			continue // ignore error because this probably isn't a path or path list to bashify
```
Naatan, line 152: This'll probably go away once line 149 is addressed but please actually assert this, don't rely on probability, or if you do; make sure it has a failsafe (ie. log it).

internal/subshell/bash/bash.go
```diff
@@ -131,6 +140,20 @@ func (v *SubShell) SetupShellRcFile(targetDir string, env map[string]string, nam
 
 // SetEnv - see subshell.SetEnv
 func (v *SubShell) SetEnv(env map[string]string) {
+	for key, value := range env {
+		var bashified string
+		var err error
+		if key == "PATH" {
+			bashified, err = osutils.BashifyPathEnv(value)
+		} else {
+			bashified, err = osutils.BashifyPath(value)
```
Naatan, line 149: This would cause us to attempt bashifying every single env var? I can't see that not causing issues. 

Perhaps we are over-complicating this. What if we just run `bash -c printenv` with our windows env passed in, then the result would be the completely bashified bash env, at least in theory.

internal/subshell/sscommon/rcfile.go
```diff
@@ -329,6 +329,12 @@ func SetupProjectRcFile(prj *project.Project, templateName, ext string, env map[
 
 	currExec := osutils.Executable()
 	currExecAbsDir := filepath.Dir(currExec)
+	if bashifyPaths {
+		currExec, err = osutils.BashifyPath(currExec)
```
Naatan, line 333: Seeing as we're calling BashifyPath from multiple locations; perhaps it would make sense if the function itself took care of reducing callouts to bash by matching against previous requests.

pkg/project/expander.go
```diff
@@ -89,8 +90,19 @@ func ExpandFromProject(s string, p *Project) (string, error) {
...
+	expansion := &Expansion{Project: p, BashifyPaths: true}
+	return expansion.ApplyWithMaxDepth(s, 0)
+}
+
 func ExpandFromScript(s string, script *Script) (string, error) {
-	expansion := &Expansion{Project: script.project, Script: script}
+	expansion := &Expansion{
+		Project:      script.project,
+		Script:       script,
+		BashifyPaths: runtime.GOOS == "windows" && script.LanguageSafe()[0] != language.Batch,
```
Naatan, line 104:
```suggestion
		BashifyPaths: runtime.GOOS == "windows" && (script.LanguageSafe()[0] == language.Bash || script.LanguageSafe()[0] == language.Sh),
```

pkg/project/project.go
```diff
@@ -608,14 +610,23 @@ func (e *Event) Name() string { return e.event.Name }
 
 // Value returned with all secrets evaluated
 func (e *Event) Value() (string, error) {
+	if e.BashifyPaths {
+		return ExpandFromProjectBashifyPaths(e.event.Value, e.project)
+	}
```
Naatan, line 615: This would lead to us bashifying values that aren't paths. I don't think we need to be expanding this value for bash paths for bash compatibility to work?

pkg/project/project.go
```diff
@@ -608,14 +610,23 @@ func (e *Event) Name() string { return e.event.Name }
...
 
 // Scope returns the scope property of the event
 func (e *Event) Scope() ([]string, error) {
 	result := []string{}
 	for _, s := range e.event.Scope {
-		v, err := ExpandFromProject(s, e.project)
+		var v string
+		var err error
+		if e.BashifyPaths {
+			v, err = ExpandFromProjectBashifyPaths(s, e.project)
```
Naatan, line 626: Like Value; this would lead to us bashifying values that aren't paths. I don't think we need to be expanding this value for bash paths for bash compatibility to work?

Naatan (changes requested):

pkg/project/project.go
mitchell-as: This only provides a context such that if a `.path()` is encountered, it is bashified. It will not bashify just any value.

pkg/project/project.go
mitchell-as: Like previously, this only instructs how to handle a `.path()` that was encountered.

internal/subshell/bash/bash.go
mitchell-as: It *attempts* to bashify every single env var that matches the "C:" pattern for filenames on Windows. This is because there are other environment variables declared by the runtime that may need to be bashified (e.g. PYTHONPATH). If the env var cannot be bashified, an error is thrown and ignored.

mitchell-as: After thinking about this some more, perhaps only PATH needs to be bashified. Presumably any other utilities know they're running on Windows and handle paths properly.

internal/prompt/prompt.go
mitchell-as: Non-interactive prompt is when you pass `-n` and do not want any interaction at all. Non-interactive terminal is when you don't pass `-n` because you wan't prompts, but cannot receive prompts because your terminal is non-interactive.

internal/subshell/sscommon/rcfile.go
mitchell-as: I'm no longer iteratively calling BashifyPath, so this is a non-issue.

internal/osutils/osutils.go
mitchell-as: This was only needed for attempting to bashify any environment variable that looked like a path. Some env vars are empty or were single characters, so I was getting panics without this test. Anyway, I'm not doing this anymore, so I've removed the check.

internal/osutils/osutils.go
mitchell-as: This is no longer needed.

internal/osutils/osutils.go
mitchell-as: I've switched to doing a simple string translation. We only need it for bash to resolve executables in PATH, so we don't need fully resolved or simplified paths. The translation results in a valid path that can be read from disk.

internal/subshell/bash/bash.go
mitchell-as: Yes, this has gone away.

internal/subshell/bash/bash.go
mitchell-as: Okay, I was overcomplicating this. As mentioned in a previous comment, I'm doing a simple string translation of PATH only, because we only need it to help bash find runtime executables.

internal/prompt/prompt.go
Naatan: I don't think that distinction is accurate. Yes the user can supply `-n` which would set non-interactive, but we also detect it:

https://github.com/ActiveState/cli/blob/4b38cc68f36736428be2ad5e42a94c9e431b414f/cmd/state/main.go#L98-L98

So this check should ideally rely on https://github.com/ActiveState/cli/blob/4b38cc68f36736428be2ad5e42a94c9e431b414f/internal/prompt/prompt.go#L40-L40

internal/subshell/bash/bash.go
Naatan: I don't think that's true. You're right in that any env var can contain path values. And while we ourselves might only rely on PATH, the user may have scripts or other automations that rely on the values in the other env variables.

What about my `bash -c printenv` suggestion? That seems to me to be a silver bullet solution.

Naatan: Note I haven't yet reviewed the string translation you refer to as I'm writing this; but that sounds like a red flag to me. There's no way I can think of that we can know what a windows path looks like in a bash environment. The mounting point can differ depending on their bash setup.

pkg/project/project.go
Naatan: I see. I think this would be better addressed at the source:

https://github.com/ActiveState/cli/blob/4b38cc68f36736428be2ad5e42a94c9e431b414f/pkg/project/expander.go#L274-L274

Your solution at the moment appears to be trying to update the path once it has already been evaluated, I think?

internal/prompt/prompt.go
mitchell-as: The `terminal.IsTerminal()` call is part of a multi-line conditional statement. One line above is a check for `-n` or `--non-interactive`. If that option is present, the entire `isInteractive` bool becomes false, regardless of whether or not the terminal is actuall interactive. This is why the separate check exists in the prompt package.

internal/subshell/bash/bash.go
mitchell-as: `bash -c printenv` does not do any translations/bashification. I still see "C:\Users\username" and similar paths. I'm not sure what you were expecting?

internal/osutils/osutils_windows.go
```diff
@@ -27,6 +28,22 @@ func CmdExitCode(cmd *exec.Cmd) (code int) {
...
+		// Calling out to bash like in BashifyPath is too slow. Just do a simple string transformation.
+		// The resulting paths will be valid in bash, they just may not be fully resolved/simplified.
+		// This is okay because bash really just needs to find executables to run.
+		vol := strings.ToLower(filepath.VolumeName(dir)) // "C:\foo bar\baz" -> "c:"
+		dir = dir[len(vol):]                             // "C:\foo bar\baz" -> "\foo bar\baz"
+		vol = strings.Replace(vol, ":", "", 1)           // "c:" -> "c"
+		dirs[i] = "/" + vol + filepath.ToSlash(dir)      // "C:\foo bar\baz" -> "/c/foo bar/baz"
+	}
+	return strings.Join(dirs, ":") // bash uses ':' while Windows uses ';'
+}
```
Naatan, line 45: This is assuming drive letters are mounted as `/c/`. I've definitely used bash version on Windows where this was not true (I think cygwin is an example of this, they might use `/mnt/c/` or `/drive/c` or something along those lines.

String replacement alone simply cannot sufficiently address this use-case. Please consider the solution I suggested here: https://github.com/ActiveState/cli/pull/2189#discussion_r1012042815

Naatan (changes requested):
Please see unresolved conversations also.

pkg/project/project.go
mitchell-as: No. This is when there is a path call inside the script itself. For example, see the changes at the top of this PR, and also review the activestate.yaml.*.tpl templates, which have `.path()` calls in them.

Naatan: As far as I'm aware the code I linked is what resolves the path call inside the script itself.

internal/subshell/bash/bash.go
Naatan: I'm not seeing this? 
image

internal/prompt/prompt.go
Naatan: I'm not understanding how that's an issue. The user specifying it as non-interactive ought to be respected. It shouldn't encounter the error you're trying to raise in that condition anyway.

internal/subshell/bash/bash.go
mitchell-as: I'm using Git Bash, which gives me this: 
VirtualBox_Windows 10_03_11_2022_17_41_53

It looks like your Bash environment is doing something extra.

internal/prompt/prompt.go
mitchell-as: I respected existing behavior with regards to the user specifying or not specifying non-interactive. The only thing I added was the case where the State Tool is prompting for input, but the user's terminal does not allow it. In this special case, we give a more informative error message. Otherwise, it's business as usual.

mitchell-as: (I clicked Delete instead of Edit *sigh*)
I am respecting the existing behavior regarding the user specifying interactive vs non-interactive. The only special case I'm handling is when the State Tool wants to prompt the user, but the terminal doesn't allow it. In that case I'm giving a more informative error message. Otherwise, it's business as usual.

internal/osutils/osutils.go
mitchell-as: I switched back from doing a simple string translation. Now we leverage MinGW/MSYS2/WSL's auto-translation of %PATH%.

internal/prompt/prompt.go
mitchell-as: Removed IsTerminal() test in favor of existing `isInteractive`. Still use a separate message for Windows as opposed to macOS and Linux, but prefer "interactive mode" verbage as opposed to "interactive terminal".

internal/subshell/bash/bash.go
mitchell-as: As we discovered last week, MinGW/MSYS2/WSL is auto-translating %PATH% to a bashified equivalent, but the State Tool is writing an unbashified form to a shell RC file for later sourcing, which does not take advantage of this auto-translation. We're now running `bash -c "echo \"$PATH\"` in order to write the auto-bashified %PATH% to the shell rc file.

internal/osutils/osutils_windows.go
mitchell-as: I am using a variant of the suggested solution in that comment. See my last comment there.

internal/osutils/osutils_windows.go
```diff
@@ -29,6 +30,19 @@ func CmdExitCode(cmd *exec.Cmd) (code int) {
 	return cmd.ProcessState.Sys().(Status).ExitStatus()
 }
 
+// BashifyPath takes a windows %PATH% list and turns it into a bash style PATH list.
+// e.g. C:\foo;C:\bar becomes /c/foo:/c/bar
+// Leverages MinGW/MSYS2/WSL's PATH transformation when it invokes a Unix command.
+func BashifyPathEnv(pathList string) (string, error) {
+	cmd := exec.Command("bash", "-c", `echo -n "$PATH"`)
```
Naatan, line 37: I thought we had to funnel this through cmd otherwise it wouldn't do the path translation?

internal/subshell/bash/bash.go
```diff
@@ -64,6 +65,15 @@ func (v *SubShell) WriteUserEnv(cfg sscommon.Configurable, env map[string]string
 		return errs.Wrap(err, "RcFile failure")
 	}
 
+	if path, pathExists := env["PATH"]; pathExists && runtime.GOOS == "windows" {
+		// Either the State Tool is being installed on Windows, or a runtime is being deployed on Windows.
```
Naatan, line 69:
```suggestion
```

Nit: This feels oddly opinionated for a package that should not care about what consumers use it for.

internal/subshell/bash/bash.go
```diff
@@ -132,6 +142,14 @@ func (v *SubShell) SetupShellRcFile(targetDir string, env map[string]string, nam
 
 // SetEnv - see subshell.SetEnv
 func (v *SubShell) SetEnv(env map[string]string) {
+	if path, pathExists := env["PATH"]; pathExists && runtime.GOOS == "windows" {
+		if bashified, err := osutils.BashifyPathEnv(path); err == nil {
+			env["PATH"] = bashified
+		} else {
+			multilog.Error("Unable to bashify PATH: %v", err)
```
Naatan, line 149: Wouldn't this almost certainly cause issues if we just ignore without warning the user? I think we'll need to return this error, which sadly means updating all use of `SetEnv`, but I don't see any other way around that at the moment.

Naatan (changes requested):
Home stretch!

internal/osutils/osutils_windows.go
mitchell-as: No, that is not necessary. If this fails, the integration test will catch it (`state` will not be found on the PATH).

internal/subshell/bash/bash.go
mitchell-as: Okay, I've refactored SetEnv to allow for throwing an error. The only questionable result is that subshell.New() cannot throw this error. However, at that point `PATH` is unmodified and an error should never be thrown. Also, whenever a runtime is activated, a new environment is set and that is where we can notify the user of an error. I've documented this in a comment in subshell.New().

internal/subshell/subshell.go
```diff
@@ -127,11 +127,11 @@ func New(cfg sscommon.Configurable) SubShell {
...
 		switch runtime.GOOS {
 		case "windows":
-			return &cmd.SubShell{}
+			subs = &cmd.SubShell{}
 		case "darwin":
-			return &zsh.SubShell{}
+			subs = &zsh.SubShell{}
 		default:
-			return &bash.SubShell{}
+			subs = &bash.SubShell{}
```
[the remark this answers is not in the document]
mitchell-as: Reverted and filed a bug, which has been triaged.

Naatan (approved):

PR #2190, opened 2022-10-26
Print used command name

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1346" title="DX-1346" target="_blank">BugDX-1346</a>  PROMPT: `state prompt` shows help and errors for `state shell` it will confuse the user
  </td></table>
  <br />
 

This differs slightly from the `Expected Behaviour` in the ticket

### DIFF
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -30,7 +30,6 @@ import (
 	"github.com/ActiveState/cli/internal/table"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"github.com/thoas/go-funk"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


Naatan (changes requested):
Ranging over os.Args makes assumptions about the contents of args and how captain uses os.Args. If we need the "UsedName()" functionality I'd rather we be more determinative with the result.

I think we're over-complicating things though. Can we just drop the command from the locale and say "This command" instead? That will also address another issue I just noticed; we don't account for sub-commands with this message.

Naatan (approved):

PR #2192, opened 2022-10-27
Include trigger in concurrency jobs

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1351" title="DX-1351" target="_blank">TaskDX-1351</a>  Concurrency jobs should include the trigger
  </td></table>
  <br />


Naatan (changes requested):
It seems I targeted the wrong version for this story. Changed it to 36, which has the concurrency changes.

Note the story is talking about adding something to the concurrency identifier, not implementing concurrency.

PR #2200, opened 2022-10-31
Add attempt analytics event

MDrakos:


Naatan:
@MDrakos changed reviewer to @daved as he authored this code. Please flag me for final review.

cmd/state-exec/main.go
```diff
@@ -65,10 +65,22 @@ func run() error {
 	}
 	logr.Debug("message data - pid: %s, exec: %s", hb.ProcessID, hb.ExecPath)
 
+	a, err := newAttempt()
+	if err != nil {
+		return fmt.Errorf("cannot create new attempt: %w", err)
+	}
+	logr.Debug("message data - exec: %s", a.ExecPath)
+
 	meta, err := newExecutorMeta(hb.ExecPath)
```
daved, line 74: The reason that the heartbeat type is being created early is to obtain the execpath. If, instead, the attempt type is constructed first, the execpath can be passed into the heartbeat construction func later (closer to where it is used - the debugging output should be able to be moved as well), and passed into the newExecutorMeta func here.

cmd/state-exec/main.go
```diff
@@ -65,10 +65,22 @@ func run() error {
...
+	// Send attempt event regardless of whether we can get the meta data.
+	logr.Debug("communications - sock: %s", meta.SockPath)
+	if msgErr := sendMsgToService(meta.SockPath, a); msgErr != nil {
```
daved, line 77: This is a bit confusing. If acquiring the meta data fails due to not finding the meta file, then the sock path value will not be set. I'm not sure that there is a great way to report an attempt if acquiring the meta data fails aside from communicating directly with whatever data store holds the usage info (i.e. bypassing the service).

There is a small chance that the data is partially corrupt, and does contain the sockpath, but it's a very narrow case.

daved (changes requested):

cmd/state-exec/main.go
MDrakos: This crossed my mind as well but at the time I decided to not address it. I'll leave it up to @Naatan if he wants us to come up with some way to deal with this situation.

Naatan: Not sure if I'm fully in the loop. But sending it without the metadata is pointless. Without the metadata we can't do anything with the event.

cmd/state-exec/main.go
```diff
@@ -59,16 +59,28 @@ func run() error {
...
 	defer logr.Debug("run goodbye")
 
-	hb, err := newHeartbeat()
+	a, err := newAttempt()
+	if err != nil {
+		return fmt.Errorf("cannot create new attempt: %w", err)
+	}
+	logr.Debug("message data - exec: %s", a.ExecPath)
+
+	hb, err := newHeartbeat(a.ExecPath)
```
daved, line 68: nit: I'd prefer to see the heartbeat construction closer to its usage below around line 91.

cmd/state-exec/main.go
daved: This doesn't seem to be addressed.  I think that this approach effectively means that the ExecutorMeta error is being ignored. I suggest printing the communication debug and sending the msg to svc within a conditional block of `if meta.SockPath != "" {`.

daved (changes requested):

daved (approved):

pkg/platform/runtime/runtime.go
```diff
@@ -183,27 +188,21 @@ func (r *Runtime) recordCompletion(err error) {
...
 	}
 
-	dims := &dimensions.Values{
-		Trigger:          p.StrP(r.target.Trigger().String()),
-		Headless:         p.StrP(strconv.FormatBool(r.target.Headless())),
-		CommitID:         p.StrP(r.target.CommitUUID().String()),
-		ProjectNameSpace: p.StrP(project.NewNamespace(r.target.Owner(), r.target.Name(), r.target.CommitUUID().String()).String()),
-		InstanceID:       p.StrP(instanceid.ID()),
-	}
+	dims.Headless = p.StrP(strconv.FormatBool(r.target.Headless()))
```
Naatan, line 205: This feels like it should be part of `usageDims()`. If this was just to respect the dimensions section in the documentation then we can just update that to also mention headless.

test/integration/analytics_int_test.go
```diff
@@ -377,6 +377,45 @@ func (suite *AnalyticsIntegrationTestSuite) TestInputError() {
...
+	events := parseAnalyticsEvents(suite, ts)
+
+	var foundAttempts int
+	for _, e := range events {
+		if strings.Contains(e.Category, "runtime") && strings.Contains(e.Action, "attempt") {
+			foundAttempts++
+		}
+	}
+
+	if foundAttempts < 2 {
```
Naatan, line 414: Unlike with heartbeats we should be able to get a consistent number here, so please assert `== 2`.

Naatan (changes requested):

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -170,6 +171,17 @@ func (r *Resolver) RuntimeUsage(ctx context.Context, pid int, exec string, dimen
 	return &graph.RuntimeUsageResponse{Received: true}, nil
 }
 
+func (r *Resolver) RuntimeAttempt(ctx context.Context, exec string, dimensionsJSON string) (*graph.RuntimeAttemptResponse, error) {
```
Naatan, line 174: This endpoint isn't needed, we already reach out to the svc for analytics. Just call the analytics event from the async client in the state tool.

The reason heartbeats have their own endpoint is because it needs to do things beside just send an analytics event.

Naatan (changes requested):
Changes in the following locations don't seem needed:

pkg/platform/runtime/runtime.go
```diff
@@ -173,8 +173,12 @@ func (r *Runtime) recordCompletion(err error) {
 		return
 	}
 	r.completed = true
+
 	logging.Debug("Recording runtime completion: %v", err == nil)
 
+	// Record attempt regarless of success
+	r.recordAttempt()
```
Naatan, line 180: Sorry for not flagging this before; but this should fire at the start of the "attempt". ie. the second the runtime is starting to be used. Now it is firing after the actual attempt has happened, the only distinction is that it doesn't trigger if the attempt was unsuccessful. 

The intent of the event is to signal that an attempt IS being made. Not that an attempt WAS made. This is important because any number of things could have happened in between.

test/integration/analytics_int_test.go
```diff
@@ -377,6 +377,45 @@ func (suite *AnalyticsIntegrationTestSuite) TestInputError() {
...
+	var foundAttempts int
+	for _, e := range events {
+		if strings.Contains(e.Category, "runtime") && strings.Contains(e.Action, "attempt") {
+			foundAttempts++
+		}
+	}
+
+	if foundAttempts == 2 {
+		suite.Fail("Should find multiple runtime attempts")
+	}
```
Naatan, line 416: Since the executor attempt is in a bit of an awkward place; it would be good if we can explicitly assert that one of the attempts has trigger=exec.

Naatan (changes requested):

pkg/platform/runtime/runtime.go
MDrakos: No problem, I think the `recordAttempt` calls are in the right place now.

Naatan (changes requested):
I think we can move the attempt to a single call alongside the "start" here:

https://github.com/ActiveState/cli/blob/3b27894f31a7a6218f48ecc894af731fe49ffae3/pkg/platform/runtime/runtime.go#L73-L73

Naatan (approved):

PR #2202, opened 2022-11-01
Revert executor PRs for now as we want to mature this code more

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1369" title="DX-1369" target="_blank">TaskDX-1369</a>  Revert executor work from v35
  </td></table>
  <br />
 

Please review the conflict resolutions in particular:

image


test/integration/use_int_test.go
```diff
@@ -42,8 +42,12 @@ func (suite *UseIntegrationTestSuite) TestUse() {
 	cp.ExpectExitCode(0)
 
 	// Verify runtime works.
-	pythonExe := filepath.Join(ts.Dirs.DefaultBin, "python3"+osutils.ExeExt)
-	cp = ts.SpawnCmd(pythonExe, "--version")
+	pythonExe := filepath.Join(ts.Dirs.DefaultBin, "python3")
```
daved, line 45: If the intent is to return to the .bat/.sh executors, I believe that lines like this need code that looks more like: https://github.com/ActiveState/cli/blob/508375609b0c87794c7885a71f71c26fad5a482d/test/integration/use_int_test.go#L29-L32

daved (changes requested):
I only spotted one area that may have an issue. If the relevant e2e tests are running, then maybe it's well enough to just go with it as-is.

daved (approved):

PR #2204, opened 2022-11-03
Print success message if the cache was successfully cleaned.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1358" title="DX-1358" target="_blank">BugDX-1358</a>  `state clean cache` missing message when execution finished successfully
  </td></table>
  <br />

### DIFF
--- a/internal/runners/clean/cache.go
+++ b/internal/runners/clean/cache.go
@@ -76,7 +76,13 @@ func (c *Cache) removeCache(path string, force bool) error {
 	}
 
 	logging.Debug("Removing cache path: %s", path)
-	return removeCache(c.path)
+	err := removeCache(c.path)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


internal/runners/clean/cache.go
```diff
@@ -76,7 +76,13 @@ func (c *Cache) removeCache(path string, force bool) error {
 	}
 
 	logging.Debug("Removing cache path: %s", path)
-	return removeCache(c.path)
+	err := removeCache(c.path)
+	if err != nil {
+		return err
```
MDrakos, line 81: We should wrap this error

MDrakos (changes requested):

MDrakos (dismissed):

MDrakos (approved):

PR #2211, opened 2022-11-03
Increase logging on "Already in activated state"

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1315" title="DX-1315" target="_blank">StoryDX-1315</a>  Increase logging on "Already in activated state"
  </td></table>
  <br />

### DIFF
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -48,6 +48,13 @@ func ActivationPID(cfg Configurable) int32 {
 			return -1
 		}
 
+		name, err := pproc.Ppid()
+		if err != nil {
+			multilog.Log(logging.ErrorNoStacktrace, rollbar.Error)(procInfoErrMsgFmt, err)
+		} else {
+			logging.Debug("Parent process name: %s", name)
+		}
+
 		pid = ppid
 		ppid, err = pproc.Ppid()
 		if err != nil {
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -50,7 +50,7 @@ func ActivationPID(cfg Configurable) int32 {
 
 		name, err := pproc.Ppid()
 		if err != nil {
-			multilog.Log(logging.ErrorNoStacktrace, rollbar.Error)(procInfoErrMsgFmt, err)
+			logging.Error(procInfoErrMsgFmt, err)
 		} else {
 			logging.Debug("Parent process name: %s", name)
 		}
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -28,10 +28,13 @@ func ActivationPID(cfg Configurable) int32 {
 	procInfoErrMsgFmt := "Could not detect process information: %v"
 
 	for pid != 0 && pid != ppid {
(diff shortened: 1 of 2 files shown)


internal/runners/activate/activate.go
```diff
@@ -121,6 +122,8 @@ func (r *Activate) run(params *ActivateParams) error {
 				return nil
 			}
 
+			logging.Debug("Currently activated project: %s", activated)
+			logging.Debug("Activated environment: %s", strings.Join(os.Environ(), "\n"))
```
Naatan, line 126: This is going to blow up our log file, plus might leak sensitive info for end-users.

Naatan (changes requested):
I think this is focussing too much on the effect and not the cause.

Instead I think it makes more sense to debug the actual "is activated" mechanic:

https://github.com/ActiveState/cli/blob/9e551969ae64d9a806698b25109de5583e5bf988/internal/process/process.go#L24-L24

Things like the PID, and the matching process it found (its name would be good) are good candidates.

internal/process/process.go
```diff
@@ -28,10 +28,13 @@ func ActivationPID(cfg Configurable) int32 {
 	procInfoErrMsgFmt := "Could not detect process information: %v"
 
 	for pid != 0 && pid != ppid {
+		logging.Debug("Current PID: %d, Parent PID: %d", pid, ppid)
 		pidFileName := ActivationPIDFileName(cfg.ConfigPath(), int(pid))
+		logging.Debug("Looking for activation pid file: %s", pidFileName)
 		if fileutils.FileExists(pidFileName) {
 			return pid
 		}
+		logging.Debug("Activation pid file not found")
```
MDrakos, line 37: I'm having trouble wrapping my head around this function. We return "successfully" early on Line 35 _before_ initializing the process and attempting to get the PID, so all that verification below seems pointless. I could add a name check and debug there but that wouldn't be for the case we are trying to verify. Initializing the process earlier makes this function more error-prone but we could ignore those errors.

daved: I don't think it would produce the same results since it is updating pid and ppid, then looping. The -1 returns collapse the loop before the inspection finds itself back to process zero, but it all does seem reasonably useful.

MDrakos: I see what you mean. I've added a name debug but I'm not sure how helpful it will be.

internal/process/process.go
```diff
@@ -45,6 +48,13 @@ func ActivationPID(cfg Configurable) int32 {
 			return -1
 		}
 
+		name, err := pproc.Ppid()
+		if err != nil {
+			multilog.Log(logging.ErrorNoStacktrace, rollbar.Error)(procInfoErrMsgFmt, err)
```
Naatan, line 53: Don't want to log to rollbar for something we only looked up for debugging purposes.

Naatan (changes requested):

Naatan (approved):

PR #2218, opened 2022-11-04
Non-exe artifacts use exe extension (executors)

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1370" title="DX-1370" target="_blank">StoryDX-1370</a>  Don't create executors for non .exe files on Windows
  </td></table>
  <br />

### DIFF
--- a/cmd/state-exec/main.go
+++ b/cmd/state-exec/main.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/ActiveState/cli/cmd/state-exec/internal/logr"
@@ -90,6 +91,12 @@ func run() error {
 
 	logr.Debug("cmd - running: %s", meta.MatchingBin)
 	if err := runCmd(meta); err != nil {
+		logr.CallIfDebugIsSet(func() {
+			logr.Debug(
+				"      running - failed: bins (%s)",
+				strings.Join(meta.ExecMeta.Bins, ", "),
+			)
+		})
 		return fmt.Errorf("cannot run command: %w", err)
 	}
 
--- a/cmd/state-exec/main.go
+++ b/cmd/state-exec/main.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/ActiveState/cli/cmd/state-exec/internal/logr"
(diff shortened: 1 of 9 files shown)


cmd/state-exec/main.go
```diff
@@ -90,6 +90,7 @@ func run() error {
 
 	logr.Debug("cmd - running: %s", meta.MatchingBin)
 	if err := runCmd(meta); err != nil {
+		logr.Debug("      running - failed: bins (%v)", meta.ExecMeta.Bins)
```
Naatan, line 93: This seems like a local debugging statement. If you want to commit it can you phrase it in a way that others reading it can understand what it's saying?

internal/exeutils/exeutils.go
```diff
@@ -179,4 +179,8 @@ func DecodeCmd(cmd string) (string, []string) {
 	default:
 		return "sh", []string{"-c", cmd}
 	}
-}
\ No newline at end of file
+}
+
+func WinExeExt(path string) string {
+	return path + Extension
+}
```
Naatan, line 186: Please drop this and just use `path + exeutils.Extension`. This isn't saving us much and is just obfuscating logic.

pkg/platform/runtime/executors/executors.go
```diff
@@ -55,15 +55,16 @@ func (es *Executors) ExecutorSrc() (string, error) {
...
+	for _, dest := range exes {
+		alias := filepath.Base(dest)
+		if rt.GOOS == "windows" {
+			ext := filepath.Ext(alias)
+			if ext != "" && ext != exeutils.Extension { // for non-.exe executables like pip.bat
+				alias = strings.TrimSuffix(alias, ext) + exeutils.Extension // setup alias pip.exe -> pip.bat
 			}
-			exes = append(exes, exe+exeutils.Extension) // Double up on the ext so only the first on gets dropped
 		}
+		executors[alias] = dest
```
Naatan, line 67: This is starting to become a bit hard to read, can you clean it up a bit? Maybe split it out into a function that clearly denotes what it's doing.

Naatan (changes requested):

cmd/state-exec/main.go
daved: The message is additive to the previous debug, so it should read nicely.

Naatan (approved):

PR #2224, opened 2022-11-07
Fix installer panic

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1388" title="DX-1388" target="_blank">BugDX-1388</a>  Installer panics if user ran command that exited non-zero
  </td></table>
  <br />


mitchell-as (dismissed):
Oops.

PR #2229, opened 2022-11-08
Remove IsAdmin() environment variable override

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1329" title="DX-1329" target="_blank">BugDX-1329</a>  IsAdmin detection broken on CI
  </td></table>
  <br />
 

Runtimes are not installed to the admin `%PATH%` even if the user has admin privileges (which is the case on CI), so when uninstalling the runtime, the user scope needs to be used. Previously, the environment variable would override `IsAdmin()` to return `false` on CI (and thus uninstall with user scope despite having admin privileges).

### DIFF
--- a/cmd/state-offline-uninstaller/uninstall.go
+++ b/cmd/state-offline-uninstaller/uninstall.go
@@ -214,7 +214,9 @@ func (r *runner) validateTargetPath(path string) error {
 
 func (r *runner) removeEnvPaths() error {
 	// remove shell file additions
-	if err := r.shell.CleanUserEnv(r.cfg, sscommon.OfflineInstallID, false); err != nil {
+	// Note: on Windows, runtimes are not installed to the Admin %PATH%, even if the user has admin
+	// privileges, so call CleanUserEnv with user scope.
+	if err := r.shell.CleanUserEnv(r.cfg, sscommon.OfflineInstallID, true); err != nil {
 		return errs.Wrap(err, "Failed to remove runtime PATH")
 	}
 
--- a/cmd/state-offline-uninstaller/uninstall.go
+++ b/cmd/state-offline-uninstaller/uninstall.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ActiveState/cli/internal/rtutils/p"
 	"github.com/ActiveState/cli/pkg/project"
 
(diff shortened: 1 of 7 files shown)


cmd/state-offline-uninstaller/uninstall.go
```diff
@@ -214,13 +213,10 @@ func (r *runner) validateTargetPath(path string) error {
...
 func (r *runner) removeEnvPaths() error {
-	isAdmin, err := osutils.IsAdmin()
-	if err != nil {
-		return errs.Wrap(err, "Could not determine if running as Windows administrator")
-	}
-
 	// remove shell file additions
-	if err := r.shell.CleanUserEnv(r.cfg, sscommon.OfflineInstallID, !isAdmin); err != nil {
+	// Note: on Windows, runtimes are not installed to the Admin %PATH%, even if the user has admin
+	// privileges, so call CleanUserEnv with user scope.
```
Naatan, line 218: This doesn't sound right to me. We pass `isAdmin` to `WriteUserEnv()`, and it's respected there.

If `CleanUserEnv()` doesn't handle this input the same as `WriteUserEnv()` then we should address that head-on. The current change feels like we're working around the issue from the consuming code, which is error-prone.

Naatan (changes requested):

cmd/state-offline-uninstaller/uninstall.go
mitchell-as: The offline installer is not calling `WriteUserEnv()` as far as I can see. I can look into making it do so, but I was under the impression that we were intentionally not running the offline installer in full admin mode because setting a global default as administrator is not supported (https://github.com/ActiveState/cli/blob/d2503b72f9fa33f73b14f2be7c9e97d894cbc088/internal/globaldefault/default.go#L40).

mitchell-as: After investigation, it was a bug in `ConfigureAvailableShells`. Thanks for suggesting something was amiss.

mitchell-as:
I checked the other calls to `ConfigureAvailableShells` in our code-base, and they look okay. Most use `true` for user-scope.

Naatan (approved):
Nice catch!

PR #2233, opened 2022-11-09
Switch only allows commit IDs from branch history

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1353" title="DX-1353" target="_blank">BugDX-1353</a>  `state switch <commit_id>` executed successfully with commit_id from another project from any organization.
  </td></table>
  <br />


mitchell-as (approved):
I like the test!

PR #2236, opened 2022-11-15
Expansion integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1385" title="DX-1385" target="_blank">StoryDX-1385</a>  We test the performance of as.yaml expansion
  </td></table>
  <br />


internal/process/process.go
```diff
@@ -27,16 +27,23 @@ func ActivationPID(cfg Configurable) int32 {
 
 	procInfoErrMsgFmt := "Could not detect process information: %v"
 
+	seen := map[int32]bool{}
```
MDrakos, line 30: These integration tests triggered an issue on Windows with PID recycling. The loop below would continue forever until the tests timed out. I can remove these fixes and file a blocking story if we want to discuss the solution.

Naatan: Nice catch! But yeah let's track this via its own story/PR. If for no other reason than that it will make it easier to keep track of the "why" on these code changes.

test/integration/performance_yaml_test.go
```diff
@@ -0,0 +1,165 @@
+package integration
```
Naatan, line 1: Can we rename this to `performance_expansion_int_test.go`?

test/integration/testdata/yaml/activestate.yaml
```diff
@@ -0,0 +1,89 @@
+project: https://platform.activestate.com/ActiveState-CLI/Yaml-Test/?branch=main&commitID=0476ac66-007c-4da7-8922-d6ea9b284fae
```
Naatan, line 1: Rather than have this hard coded on the filesystem it'd be nice if we can generate this on the fly from the test. That way the code for the actual test we're performing is all in one place also.

So the individual tests would be doing something like:

```
suite.testScriptPerformance(`echo "Hello World"`, "Hello World", DefaultSamples, baseline)
```

test/integration/performance_yaml_test.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (suite *PerformanceYamlIntegrationTestSuite) testScriptPerformance(scriptName, expect string, samples int, max time.Duration) time.Duration {
+	suite.OnlyRunForTags(tagsuite.Performance)
+	ts := e2e.New(suite.T(), true)
+	defer ts.Close()
+
+	suite.startSvc(ts)
+
+	ts.LoginAsPersistentUser()
```
Naatan, line 124: Maybe make this conditional? It'd make a lot of the performance tests very slow themselves, even though they mostly don't need auth.

test/integration/performance_yaml_test.go
```diff
@@ -0,0 +1,165 @@
...
+	// Start svc first, as we don't want to measure svc startup time which would only happen the very first invocation
+	stdout, stderr, err := exeutils.ExecSimple(ts.SvcExe, []string{"start"}, []string{})
+	suite.Require().NoError(err, fmt.Sprintf("Full error:\n%v\nstdout:\n%s\nstderr:\n%s", errs.JoinMessage(err), stdout, stderr))
+}
+
+func (suite *PerformanceYamlIntegrationTestSuite) TestYamlPerformance() {
+	suite.OnlyRunForTags(tagsuite.Performance)
+	baseline := DefaultMaxTime
+	suite.Run("CallScript", func() {
+		avg := suite.testScriptPerformance("call-script", "Hello World", DefaultSamples, DefaultMaxTime)
```
Naatan, line 50: We should calculate the baseline with a clean activestate.yaml that has no other scripts or utilized merged yamls, otherwise that would invalidate part of the performance tests (namely the merged yaml one).

Naatan (changes requested):

test/integration/testdata/yaml/activestate.yaml
MDrakos: It proved to be a bit more complex when we want to support tests calling other scripts, using constants and calling scripts from merged activestate.yaml files.

PR #2237, opened 2022-11-15
Do not log "bad call to lsof" to Rollbar.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1284" title="DX-1284" target="_blank">BugDX-1284</a>  state-installer [-t -f --source-installer]: Could not get executable path for state-svc process, error: bad call to lsof
  </td></table>
  <br />

### DIFF
--- a/internal/installmgr/stop.go
+++ b/internal/installmgr/stop.go
@@ -85,6 +85,11 @@ func stopSvc(installPath string) error {
 		if n == svcName {
 			exe, err := p.Exe()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


daved (approved):

daved:
We could try again on this error. My guess is that a second attempt will confirm shutdown. Though, I didn't look at much of the code context, so maybe ignoring is sufficient to be sure the svc stopped.

PR #2242, opened 2022-11-15
Always show either "Installing Runtime" or "Updating Runtime" when installing or updating a runtime

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1396" title="DX-1396" target="_blank">StoryDX-1396</a>  In case state command trigger runtime update we provide suitable messages.
  </td></table>
  <br />
 

Explicit calls to `runbits.RefreshRuntime()` work when we know a runtime update needs to happen. This function would display "Installing Runtime" or "Updating Runtime" to the user. However, implicit updates (e.g. after `state update`) would not call this function, and a silent runtime update would occur.

This PR adds a new event to the runtime that can be used to always print either "Installing Runtime" or "Updating Runtime" to the user when installing or updating a runtime, regardless of whether it's an explicit update or implicit update.

### DIFF
--- a/cmd/state-offline-installer/progress.go
+++ b/cmd/state-offline-installer/progress.go
@@ -27,6 +27,10 @@ func newOfflineProgressOutput(out output.Outputer) *offlineProgressOutput {
 	return &offlineProgressOutput{out: out}
 }
 
+func (mpo *offlineProgressOutput) UpdateStarted(bool) error {
+	return nil
+}
+
 func (mpo *offlineProgressOutput) BuildStarted(total int64) error {
 	return nil
 }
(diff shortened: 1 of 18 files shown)


MDrakos (approved):
Nice work! This makes sense to me.

Naatan (changes requested):
This feels like overkill for what we're looking for. It's a lot of added complexity for what amounts to a simple message meant to indicate that we are sourcing their runtime.

I think this PR can just be a single addition here:

https://github.com/ActiveState/cli/blob/9b010694dfeca42fc68750ea9c3e0aebbd506985/internal/runbits/progressbar/progressbar.go#L104-L104

That prints something like "Setting up runtime". Making the distinction between install and update is nice if we can assert this, but doesn't warrant the added complexity just for this message imo.

mitchell-as:
Okay, how about this? BuildStarted does not always fire, but SolverStart seems to.

Naatan (approved):

PR #2244, opened 2022-11-16
We warn macOS users not to use bash

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1398" title="DX-1398" target="_blank">StoryDX-1398</a>  We warn macOS users not to use bash
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -348,7 +348,7 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 	}
 
 	ss := subshell.New(cfg)
-	if filepath.Base(ss.Shell()) == bash.Name && runtime.GOOS == "darwin" {
+	if ss.Shell() == bash.Name && runtime.GOOS == "darwin" {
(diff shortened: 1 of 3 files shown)


mitchell-as (changes requested):

cmd/state-installer/cmd.go
```diff
@@ -345,6 +347,11 @@ func postInstallEvents(out output.Outputer, cfg *config.Instance, an analytics.D
 		return locale.WrapError(err, "err_state_exec")
 	}
 
+	ss := subshell.New(cfg)
+	if filepath.Base(ss.Shell()) == bash.Name && runtime.GOOS == "darwin" {
```
mitchell-as, line 351: Please remove the `filepath.Base()` call and just leave it at `ss.Shell() == bash.Name`

mitchell-as (changes requested):

cmd/state-installer/cmd.go
MDrakos: 🤦‍♂️

mitchell-as (approved):

PR #2246, opened 2022-11-17
Try specific tests for DX-1370

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1370" title="DX-1370" target="_blank">StoryDX-1370</a>  Don't create executors for non .exe files on Windows
  </td></table>
  <br />

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -41,7 +41,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
-func main() { // bump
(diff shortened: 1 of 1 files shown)


daved:
This PR demonstrates that the activation tests are passing. v0.36.0-rc1 was failing on main. We suspect (strongly now that this PR passes) that the svc-stop behavior is causing this due to tests running in parallel on the same machine during CI.

PR #2249, opened 2022-11-24
Handle aggressive PID recycling

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1431" title="DX-1431" target="_blank">BugDX-1431</a>  Handle aggressive PID recycling by Windows.
  </td></table>
  <br />

### DIFF
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -44,6 +44,7 @@ func ActivationPID(cfg Configurable) int32 {
 		}
 		if seen[ppid] {
 			logging.Debug("Parent process PID has already been seen")
+			return -1
 		}
 
 		pproc, err := process.NewProcess(ppid)
@@ -61,7 +62,6 @@ func ActivationPID(cfg Configurable) int32 {
 			logging.Debug("Parent process name: %s", name)
 		}
 
-		seen[pid] = true
 		seen[ppid] = true
 		pid = ppid
 		ppid, err = pproc.Ppid()
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -27,17 +27,23 @@ func ActivationPID(cfg Configurable) int32 {
 
 	procInfoErrMsgFmt := "Could not detect process information: %v"
 
+	seen := map[int32]bool{}
 	for pid != 0 && pid != ppid {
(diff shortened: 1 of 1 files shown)


internal/process/process.go
```diff
@@ -27,18 +27,24 @@ func ActivationPID(cfg Configurable) int32 {
...
 			return pid
 		}
 		logging.Debug("Activation pid file not found")
 
 		if ppid == 0 {
+			logging.Debug("Parent PID is 0")
 			return -1
 		}
+		if seen[ppid] {
+			logging.Debug("Parent process PID has already been seen")
```
Naatan, line 46: Don't we want a `continue` here?

internal/process/process.go
```diff
@@ -48,13 +54,15 @@ func ActivationPID(cfg Configurable) int32 {
...
 
-		name, err := pproc.Ppid()
+		name, err := pproc.Name()
 		if err != nil {
 			logging.Error(procInfoErrMsgFmt, err)
 		} else {
 			logging.Debug("Parent process name: %s", name)
 		}
 
+		seen[pid] = true
```
Naatan, line 64: I think we should drop this one. We technically have not "seen" this one since we only look at the ppid. Sure the ppid that comes around might equal the pid, but in that case line 65 would already have handled that.

Naatan (changes requested):

internal/process/process.go
MDrakos: Whoops missed that. I believe we want to return here. We've already seen this parent ID which means it's been recycled so we're in the loop. Also continuing here would not change the `pid` and `ppid` values so the loop would get stuck again.

Naatan (changes requested):
As discussed on Zoom we want to get a better understanding for how this loop happens before we merge this.

MDrakos:
After another discussion on zoom, we've decided to leave this guard in and try to reproduce the integration test error from https://github.com/ActiveState/cli/pull/2236 once the new integration tests framework is ready. The Jira ticket has been updated with information about the error and reproduction instructions.

Naatan (approved):

PR #2250, opened 2022-11-25
Add performance integration test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1385" title="DX-1385" target="_blank">StoryDX-1385</a>  We test the performance of as.yaml expansion
  </td></table>
  <br />

### DIFF
--- a/test/integration/performance_expansion_int_test.go
+++ b/test/integration/performance_expansion_int_test.go
@@ -26,7 +26,7 @@ const (
 	DefaultMaxTime        = 1000 * time.Millisecond
 	DefaultSamples        = 10
 	DefaultVariance       = 0.75
-	DefaultSecretsMaxTime = 2100 * time.Millisecond
+	DefaultSecretsMaxTime = 3500 * time.Millisecond
 	// Add other configuration values on per-test basis if needed
 )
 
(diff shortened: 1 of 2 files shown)


MDrakos:
Had to close the other PR as a bunch of unrelated commits were picked up. Reference for comments here: https://github.com/ActiveState/cli/pull/2236

test/integration/performance_expansion_int_test.go
```diff
@@ -0,0 +1,329 @@
...
+				Language: "bash",
+			},
+			expect:  "Hello World",
+			samples: DefaultSamples,
+			max:     DefaultMaxTime,
+		})
+		variance := float64(avg) + (float64(avg) * DefaultVariance)
+		baseline = time.Duration(variance)
+	})
+
```
Naatan, line 60: Can we add a `suite.Require().NotEqual(baseline, DefaultMaxTime)` here to ensure that the baseline has been overwritten?

I feel like I've ran into some issues with `suite.Run()` being ran concurrently. Certainly the underlying implementation could be doing this. I don't think it's intended to be used in such a way that multiple `suite.Run()` calls impact each other.

MDrakos: Good to know!

Naatan (approved):

PR #2252, opened 2022-11-29
Added new "autostart.svc" config parameter to toggle state-svc autostart

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1403" title="DX-1403" target="_blank">TaskDX-1403</a>  Users can opt-out of having state-svc run on startup
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -11,20 +11,17 @@ import (
 	"syscall"
 	"time"
 
-	svcAutostart "github.com/ActiveState/cli/cmd/state-svc/autostart"
+	"github.com/ActiveState/cli/cmd/state-svc/autostart"
 	anaSync "github.com/ActiveState/cli/internal/analytics/client/sync"
 	"github.com/ActiveState/cli/internal/captain"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/events"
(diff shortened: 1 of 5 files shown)


cmd/state-svc/main.go
```diff
@@ -100,6 +104,22 @@ func run(cfg *config.Instance) error {
...
+			configMediator.AddListener(constants.AutostartSvcConfigKey, func() {
+				if cfg.GetBool(constants.AutostartSvcConfigKey) {
+					as.Enable()
+				} else {
+					as.Disable()
+				}
+			})
+		} else {
+			multilog.Error("state-svc could not find its autostart")
+		}
```
daved, line 118: Please consider hiding this within the cmd/state-svc/autostart package as a function named something like "NewAutostartAutostart". I realize there's a stutter, and am open to abbreviating or simplifying in some manner. The intention of that naming convention is to emphasize that the type returned is internal/osutils/autostart.Autostart (i.e. from another package) rather than a type named Autostart that is local to the cmd/state-svc/autostart package (which would normally have a construction func named "New", so maybe "NewAutostart" is enough in your opinion.).

daved (changes requested):

cmd/state-svc/main.go
mitchell-as: Thanks for the suggestion. I took this a step further and moved the entire block into the autostart package and exposed a single function: `RegisterConfigListener()`. We don't need a persistent object here, so we side-step the potentially awkward naming. I hope this is okay with you.

PR #2254, opened 2022-11-30
macOS LaunchAgent should launch state-svc directly

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1394" title="DX-1394" target="_blank">StoryDX-1394</a>  macOS ventura doesn't like our state-svc start script
  </td></table>
  <br />
 

Also use a proper identifier.

$HOME appears to be available anyway, and launching it through sh shows "sh" as the identifier, which is not descriptive. Using the executable directly shows something more informative and also takes the user to its location after the "get info" button is clicked.

Note: this PR does not meet the ACs as written, but I put a comment in the ticket why we cannot meet them with LaunchAgents like this.

Screenshot 2022-11-30 at 3 54 44 PM
Screenshot 2022-11-30 at 3 53 28 PM

### DIFF
--- a/internal/assets/contents/com.activestate.platform.state.plist.tpl
+++ b/internal/assets/contents/com.activestate.platform.state.plist.tpl
@@ -11,10 +11,6 @@
           <string>{{.Args}}</string>
           {{- end}}
         </array>
-        {{- if .Interactive }}
-        <key>ProcessType</key>
-        <string>Interactive</string>
-        {{- end}}
         <key>RunAtLoad</key>
         <true/>
         <key>KeepAlive</key>
--- a/internal/assets/contents/com.activestate.platform.state.plist.tpl
+++ b/internal/assets/contents/com.activestate.platform.state.plist.tpl
@@ -3,7 +3,7 @@
 <plist version="1.0">
     <dict>
         <key>Label</key>
-        <string>com.activestate.StateToolService</string>
+        <string>{{.Label}}</string>
         <key>ProgramArguments</key>
         <array>
           <string>{{.Exec}}</string>
--- a/internal/assets/contents/com.activestate.platform.state.plist.tpl
+++ b/internal/assets/contents/com.activestate.platform.state.plist.tpl
@@ -11,6 +11,10 @@
           <string>{{.Args}}</string>
           {{- end}}
         </array>
+        {{- if .Interactive }}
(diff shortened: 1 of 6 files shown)


internal/assets/contents/com.activestate.platform.state.plist.tpl
```diff
@@ -3,23 +3,17 @@
...
-            {{- else}}
-            <string>{{.Exec}}</string>
-            {{- end}}
+          <string>{{.Exec}}</string>
+          {{- if .Args }}
+          <string>{{.Args}}</string>
+          {{- end}}
         </array>
-        <key>ProcessType</key>
-        <string>Interactive</string>
```
Naatan, line 19: We also use this file to create the state-tray autostart. I think that one likely relies on this setting. Should make this a conditional instead perhaps?

The other changes are probably fine for state-tray, but to be safe we should verify that.

Naatan (changes requested):

internal/assets/contents/com.activestate.platform.state.plist.tpl
mitchell-as: Okay, currently this file is only being used by state-svc autostart. Whenever we go back to adding support for state-tray, then it can start utilizing this file too, as well as a conditional property I added.

Naatan: > Whenever we go back to adding support for state-tray

The state-tray is "supported", it's just in beta. Frankly, we probably should remove it but that's not for this ticket to address. Point is it's part of our codebase right now and thus we should address it.

Naatan (changes requested):
See comment response.

internal/assets/contents/com.activestate.platform.state.plist.tpl
mitchell-as: I built the state-tray macOS app bundle and it has no mention of ProcessType in its plist file, so it doesn't look like it references the plist template file in this PR. I have removed the conditional I added in the previous commit.

Naatan (approved):

mitchell-as:
Sorry, I was wrong about this. I've made some modifications to support both state-svc and state-tray autostart plist files.

Naatan (approved):

PR #2255, opened 2022-12-01
Do not error attempting to kill a state-svc process that has already finished.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1386" title="DX-1386" target="_blank">BugDX-1386</a>  state update failed and has a double error message with the need for help part in the middle.
  </td></table>
  <br />

### DIFF
--- a/internal/installmgr/stop.go
+++ b/internal/installmgr/stop.go
@@ -1,6 +1,8 @@
 package installmgr
 
 import (
+	"errors"
+	"os"
 	"runtime"
 	"strings"
 	"syscall"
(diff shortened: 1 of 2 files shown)


MDrakos (commented):
Won't this still show the `Something went wrong` header twice? Also, do we have any ideas as to why we cannot stop the service? That seems to be the issue described in the ticket.

mitchell-as:
Good catch. I don't see an easy way to eliminate the double header because the updater executes the installer and pipes its stdout and stderr directly to the terminal without having a way to read it or determine whether or not the err was a locale.InputError. I've reverted the tips fix and just addressed the state-svc killing failure. Let me know if you think there's a better way to address this ticket. It's a bit ambiguous as to what exactly the fix should be.

MDrakos (approved):
Nice! I think this covers it.

PR #2257, opened 2022-12-02
Fix autostart output on WSL

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1432" title="DX-1432" target="_blank">BugDX-1432</a>  `--: command not found` when starting WSL console
  </td></table>
  <br />

### DIFF
--- a/internal/osutils/autostart/autostart_linux.go
+++ b/internal/osutils/autostart/autostart_linux.go
@@ -88,6 +88,11 @@ func (a *app) enableOnServer() error {
 		exec += " " + esc.Quote(arg)
 	}
 
+	// Some older versions of the State Tool used a different ID for the autostart entry.
+	err = sscommon.CleanRcFile(profile, sscommon.InstallID)
+	if err != nil {
+		return errs.Wrap(err, "Could not clean old autostart entry from %s", profile)
+	}
 	return sscommon.WriteRcData(exec, profile, sscommon.AutostartID)
 }
 
--- a/internal/osutils/autostart/autostart_linux.go
+++ b/internal/osutils/autostart/autostart_linux.go
@@ -1,6 +1,7 @@
 package autostart
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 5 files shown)


internal/osutils/autostart/autostart_linux.go
```diff
@@ -133,9 +140,13 @@ func (a *app) disable() error {
 	if err != nil {
 		return errs.Wrap(err, "Could not find ~/.profile")
 	}
+	// Some older versions of the State Tool used a different ID for the autostart entry.
 	if fileutils.FileExists(profile) {
 		return sscommon.CleanRcFile(profile, sscommon.InstallID)
```
Naatan, line 145: I guess we're doing it this way in order to clean up the old ones? Can we clearly mark this as Legacy if so? ie. `InstallIDLegacy`.

internal/osutils/autostart/autostart_linux.go
```diff
@@ -81,12 +82,18 @@ func (a *app) enableOnServer() error {
 	}
 
 	esc := osutils.NewBashEscaper()
-	exec := esc.Quote(a.Exec)
+	// Avoid output to the console when starting up
+	exec := fmt.Sprintf("%s > /dev/null 2>&1", a.Exec)
```
Naatan, line 86: This is the wrong place to address showing the title IMO. If there are real errors such as the current one I wouldn't mind that it is explicit and end-users thus have something to report to us.

Instead I think we can just drop printing titles from the state-svc altogether. It's not meant to be run by users manually anyway.

Naatan (changes requested):

internal/osutils/autostart/autostart_linux.go
MDrakos: We still use `InstallID` for installations. So far it's only been using the `--` prefix in the `.profile` that has surfaced an issue.

Naatan: Oh interesting. In that case can we rename these constants so that it's clear they're related but for slightly different use-cases?

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2257#discussion_r1038535376

internal/osutils/autostart/autostart_linux.go
MDrakos: I _believe_ this already is the case with the old one being `InstallID` and the new one being `AutostartID`. I'm open to any suggestions to make these more clear.

MDrakos:
Forgot to request a re-review. See comment: https://github.com/ActiveState/cli/pull/2257#discussion_r1038591407

Naatan (approved):

PR #2261, opened 2022-12-05
Shorten stacktrace format.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1374" title="DX-1374" target="_blank">TaskDX-1374</a>  Our logging format is concise
  </td></table>
  <br />
 

Reduce data sent to Rollbar, but keep it meaningful.

### DIFF
--- a/internal/osutils/stacktrace/stacktrace.go
+++ b/internal/osutils/stacktrace/stacktrace.go
@@ -33,6 +33,13 @@ type Frame struct {
 // for purpose of performance optimisation.
 var FrameCap = 20
 
+var environmentRootPath string
+
+func init() {
+	// Note: ignore any error. It cannot be logged due to logging's dependence on this package.
+	environmentRootPath, _ = environment.GetRootPath()
+}
+
 // String returns a string representation of a stacktrace
 // For example:
 //   ./package/file.go:123:file.func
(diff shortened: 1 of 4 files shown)


internal/osutils/stacktrace/stacktrace.go
```diff
@@ -31,10 +34,33 @@ type Frame struct {
...
 func (t *Stacktrace) String() string {
 	result := []string{}
 	for _, frame := range t.Frames {
-		result = append(result, fmt.Sprintf(`%s:%s:%d`, frame.Path, frame.Func, frame.Line))
+		// Shorten path from its absolute path.
+		path := frame.Path
+		if strings.HasPrefix(path, build.Default.GOROOT) {
+			// Convert "/path/to/go/distribution/file" to "<go>/file".
+			path = strings.Replace(path, build.Default.GOROOT, "<go>", 1)
+		} else if root, err := environment.GetRootPath(); err == nil {
```
Naatan, line 49: `environment.GetRootPath()` will be the same each time it is called during the same runtime. We should cache it.

internal/osutils/stacktrace/stacktrace.go
```diff
@@ -31,10 +34,33 @@ type Frame struct {
...
+			// Convert "/path/to/cli/file" to "./file".
+			if relPath, err2 := filepath.Rel(root, path); err2 == nil {
```
Naatan, line 51: Nit: Calling this err2 isn't necessary.

Naatan (changes requested):
This is currently not addressing the first AC of the story. The story is to reduce the size of both the logging format as well as the stacktrace format.

Naatan (dismissed):

Naatan (approved):

PR #2262, opened 2022-12-06
Send state-svc logs along with state tool logs to Rollbar.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1375" title="DX-1375" target="_blank">TaskDX-1375</a>  Include state-svc log in rollbar reports
  </td></table>
  <br />
 

Create a way for state-svc to send over its logs on request.

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -150,6 +150,17 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 
 	svcmodel := model.NewSvcModel(svcPort)
 
+	// Give the logger a way to fetch the state-svc log tail. This cannot be done inside the logger
+	// package itself because importing pkg/platform/model creates an import cycle.
+	logging.SetSvcTailProvider(func() string {
+		ctx, cancel := context.WithTimeout(context.Background(), model.SvcTimeoutMinimal)
+		defer cancel()
+		if tail, err := svcmodel.FetchLogTail(ctx); err == nil {
+			return tail
+		}
+		return ""
+	})
+
 	// Retrieve project file
 	pjPath, err := projectfile.GetProjectFilePath()
 	if err != nil && errs.Matches(err, &projectfile.ErrorNoProjectFromEnv{}) {
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -150,16 +150,21 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 
 	svcmodel := model.NewSvcModel(svcPort)
 
-	// Give the logger a way to fetch the state-svc log tail. This cannot be done inside the logger
+	// Amend Rollbar data to also send the state-svc log tail. This cannot be done inside the rollbar
 	// package itself because importing pkg/platform/model creates an import cycle.
-	logging.SetSvcTailProvider(func() string {
+	rollbar.SetLogDataAmender(func(logData string) string {
 		ctx, cancel := context.WithTimeout(context.Background(), model.SvcTimeoutMinimal)
 		defer cancel()
-		tail, err := svcmodel.FetchLogTail(ctx)
+		svcLogData, err := svcmodel.FetchLogTail(ctx)
 		if err != nil {
-			return fmt.Sprintf("Could not fetch state-svc log: %v", err)
+			svcLogData = fmt.Sprintf("Could not fetch state-svc log: %v", err)
 		}
-		return tail
+		logData += "\nstate-svc log:\n"
+		if len(svcLogData) == logging.TailSize {
+			logData += "<truncated>\n"
+		}
+		logData += svcLogData
+		return logData
 	})
 
 	// Retrieve project file
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -155,10 +155,11 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	logging.SetSvcTailProvider(func() string {
 		ctx, cancel := context.WithTimeout(context.Background(), model.SvcTimeoutMinimal)
 		defer cancel()
-		if tail, err := svcmodel.FetchLogTail(ctx); err == nil {
-			return tail
+		tail, err := svcmodel.FetchLogTail(ctx)
+		if err != nil {
+			return fmt.Sprintf("Could not fetch state-svc log: %v", err)
 		}
-		return ""
+		return tail
 	})
 
 	// Retrieve project file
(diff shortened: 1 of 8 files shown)


cmd/state/main.go
```diff
@@ -150,6 +150,17 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
...
 
+	// Give the logger a way to fetch the state-svc log tail. This cannot be done inside the logger
+	// package itself because importing pkg/platform/model creates an import cycle.
+	logging.SetSvcTailProvider(func() string {
+		ctx, cancel := context.WithTimeout(context.Background(), model.SvcTimeoutMinimal)
+		defer cancel()
+		if tail, err := svcmodel.FetchLogTail(ctx); err == nil {
+			return tail
+		}
+		return ""
```
daved, line 161: It might be worth displaying the error if one is encountered: `return fmt.Sprintf("svc log tail not found: %v", err)`.

internal/logging/logging.go
```diff
@@ -237,6 +237,26 @@ func ReadTail() string {
...
+
+// ReadSvcTail returns the tail end of the state-svc log, just like ReadTail().
+// This exists in order to have Rollbar send both state logs and state-svc logs together.
+// If no provider function was given earlier, this function returns the empty string, so calling it
+// from state-svc is safe. (i.e. state-svc is sending its own error log to Rollbar.)
+func ReadSvcTail() string {
+	if svcTailProvider != nil {
+		return svcTailProvider()
+	}
+	return ""
```
daved, line 257: Please consider returning something like "no svc log tail provider set" for the sake of debugging.

internal/logging/logging.go
```diff
@@ -237,6 +237,26 @@ func ReadTail() string {
...
+	svcTailProvider = f
+}
+
+// ReadSvcTail returns the tail end of the state-svc log, just like ReadTail().
+// This exists in order to have Rollbar send both state logs and state-svc logs together.
+// If no provider function was given earlier, this function returns the empty string, so calling it
+// from state-svc is safe. (i.e. state-svc is sending its own error log to Rollbar.)
+func ReadSvcTail() string {
+	if svcTailProvider != nil {
+		return svcTailProvider()
```
daved, line 255: I suggest that the truncation should be moved here in the same way that the ring buffer does its work in this package.

daved (changes requested):

cmd/state/main.go
mitchell-as: I'm not sure about that. The user is already going to get a display that something went wrong, so this would just be added noise that's more confusing to them.

internal/logging/logging.go
mitchell-as: I don't follow. The state tool supplies a svcTailProvider that asks the service to send its truncated log tail (which is the contents of the service's own ring buffer). Are you asking the service to send its entire log over the wire and have the truncation happen here in the state tool process?

internal/logging/logging.go
mitchell-as: Okay, I will use a named constant here.

cmd/state/main.go
mitchell-as: Oh wait, I see what you mean. This formatted error will be uploaded to rollbar. Got it. That's a good idea.

internal/logging/logging.go
daved: Ah, yeah, that sounds rough. In that case I'd send the truncation length as an argument, but I'll leave it to you to decide if that's worthwhile. Favoring the truncation being done before being transferred is surely preferable, imo.

mitchell-as: Right, and it's also a constant value, just like in the state tool: https://github.com/ActiveState/cli/blob/f569d74340062440556372e7a539a3a33a5e87ff/internal/logging/logging.go#L219

internal/rollbar/rollbar.go
```diff
@@ -112,6 +111,13 @@ func logToRollbar(critical bool, message string, args ...interface{}) {
 	if len(logData) == logging.TailSize {
 		logData = "<truncated>\n" + logData
 	}
+	if svcLogData := logging.ReadSvcTail(); svcLogData != logging.SvcTailProviderNotSet {
```
daved, line 114: My thought was to have the value included for explicitness in case something is going wrong somewhere, so please consider including the SvcTailProviderNotSet msg in the logData.

internal/logging/logging.go
daved: I see what's going on... I think I just don't like the way that the truncation is being communicated in the rollbar package. No worries: If it's worked fine so far, let's leave it. IOW, ignore this thread.

daved (changes requested):

internal/rollbar/rollbar.go
mitchell-as: Fair enough.

daved (dismissed):

cmd/state/main.go
```diff
@@ -150,6 +150,18 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
...
+	// package itself because importing pkg/platform/model creates an import cycle.
+	logging.SetSvcTailProvider(func() string {
+		ctx, cancel := context.WithTimeout(context.Background(), model.SvcTimeoutMinimal)
+		defer cancel()
+		tail, err := svcmodel.FetchLogTail(ctx)
+		if err != nil {
+			return fmt.Sprintf("Could not fetch state-svc log: %v", err)
+		}
+		return tail
+	})
```
Naatan, line 163: I don't think there's merit in linking "give me the svc log" with the logging library. ie. when we report to rollbar we can call the model function to retrieve the svc log then and there, it doesn't need to be routed through the logging lib itself.

To put it differently; the responsibility of the logging lib is logging of the current runtime. Not providing access to logs of other runtimes.

Naatan: Perhaps what we can do instead is introduce a mechanic to the rollbar library that allows you to amend the logged data before it is sent.

Naatan (changes requested):

cmd/state/main.go
mitchell-as: That's reasonable. I've added such a function to the rollbar library and utilize it within cmd/state/main.go. (Keep in mind there's still an import cycle trying to use the svcModel within the Rollbar package).

Naatan (changes requested):

Naatan (approved):
Beautiful :)

PR #2264, opened 2022-12-07
Offline Installer can install multiple distributions

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1298" title="DX-1298" target="_blank">BugDX-1298</a>  Offline Installer can install multiple distributions
  </td></table>
  <br />


Naatan (approved):
Nicely done!

PR #2268, opened 2022-12-07
Offline installer does not error on same directory

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1301" title="DX-1301" target="_blank">BugDX-1301</a>  Offline installer errors on installing to existing dir
  </td></table>
  <br />


test/integration/offinstall_int_test.go
```diff
@@ -171,6 +171,77 @@ func (suite *OffInstallIntegrationTestSuite) TestInstallNoPermission() {
 	tp.ExpectExitCode(1)
 }
 
+func (suite *OffInstallIntegrationTestSuite) TestInstallTwice() {
+	suite.OnlyRunForTags(tagsuite.OffInstall)
+
+	// Clean up env after test (windows only for now)
```
mitchell-as, line 177: Why is this Windows only? Since this test runs on all platforms, is there a risk the environment will not be cleaned up for macOS and Linux? Will there be any ill effects?

MDrakos: Fair point. This was copied from another test that already existed. I've updated all of the tests to clean up after themselves

test/integration/offinstall_int_test.go
```diff
@@ -57,6 +57,14 @@ func (suite *OffInstallIntegrationTestSuite) TestInstallAndUninstall() {
 		defer func() {
 			suite.Require().NoError(env.Set("PATH", origPath))
 		}()
+	} else {
```
mitchell-as, line 60: Now that you're cleaning up on all platforms, would you please change the above comment to something like "Clean up env after test"?

test/integration/offinstall_int_test.go
```diff
@@ -182,6 +190,14 @@ func (suite *OffInstallIntegrationTestSuite) TestInstallMultiple() {
...
```
mitchell-as, line 193: Same here. Please update the comment above.

test/integration/offinstall_int_test.go
mitchell-as: Okay great. Please update the comment then.

mitchell-as (changes requested):

mitchell-as (approved):

PR #2269, opened 2022-12-08
Offline installer does not give state tool errors

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1302" title="DX-1302" target="_blank">BugDX-1302</a>  Offline installer gives state tool errors
  </td></table>
  <br />


cmd/state-offline-installer/main.go
```diff
@@ -38,6 +38,7 @@ func main() {
 
 	// Handle things like panics, exit codes and the closing of globals
 	defer func() {
+		panics.RecoverLocale = "offline_installer_recover_message"
```
daved, line 41: This feels like shenanigans. Please consider changing the function definition to take a locale key. A default value can be provided in the panics pkg with special keys being defined in calling pkgs.

daved (changes requested):

daved (approved):

daved:
An alternative could have been for there to be a new function added that permits the use of a l10n key, and the original function could simply call the new function with a default key. Approved as-is, though.

PR #2270, opened 2022-12-11
Do not use temp dir when updating artifacts

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1404" title="DX-1404" target="_blank">StoryDX-1404</a>  Runtime appears to still be using a temp dir
  </td></table>
  <br />


pkg/platform/runtime/setup/setup.go
```diff
@@ -229,21 +229,21 @@ func (s *Setup) updateArtifacts() ([]artifact.ArtifactID, error) {
 		}
 
 		// Retrieve environment definitions for artifact
-		envDef, err := as.EnvDef(unpackedDir)
+		envDef, err := as.EnvDef(tempUnpackedDir)
```
MDrakos, line 232: This is the only function where we pass the `_runtime_temp` dir directly and I believe for now it will have to stay that way.  There are two implementations of this function, one for the alternative builder and one for camel. The alternative builder just wants the `runtime.json` file while camel relies a bit more on the directory structure.

MDrakos:
Added AC to runtime refactor here: https://activestatef.atlassian.net/browse/DX-846

PR #2274, opened 2022-12-13
Fixed display of error message.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1428" title="DX-1428" target="_blank">BugDX-1428</a>  Unhelpful error message when uninstalling on macOS
  </td></table>
  <br />


MDrakos (changes requested):
The image linked in the story also has a color tag that is not supported. Note the `[WARNING]` in the output. https://github.com/ActiveState/cli/blob/master/internal/colorize/colorize.go#L20

PR #2275, opened 2022-12-13
Convert various rollbar errors to input errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1426" title="DX-1426" target="_blank">TaskDX-1426</a>  Convert various rollbar reports to input errors
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1690,6 +1690,8 @@ err_no_projectfile:
   other: No activestate.yaml file exists in the current working directory or its parent directories.
 err_project_not_found:
   other: Could not find project at [NOTICE]{{.V0}}[/RESET]
+err_projectfile_parse:
+  other: Unable to parse {{.V0}}. Please try checking out your project again.
 err_non_interactive_prompt:
   other: Prompt "{{.V0}}" cannot be resolved in non-interactive mode.
 err_read_projectfile:
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1691,7 +1691,7 @@ err_no_projectfile:
 err_project_not_found:
   other: Could not find project at [NOTICE]{{.V0}}[/RESET]
 err_projectfile_parse:
-  other: Unable to parse {{.V0}}. Please try checking out your project again.
+  other: Unable to parse {{.V0}}
 err_non_interactive_prompt:
   other: Prompt "{{.V0}}" cannot be resolved in non-interactive mode.
 err_read_projectfile:
--- a/pkg/platform/authentication/auth.go
+++ b/pkg/platform/authentication/auth.go
@@ -65,7 +65,7 @@ func LegacyGet() *Auth {
 			os.Exit(1)
 		}
 		defer cfg.Close()
-		
+
 		persist = New(cfg)
 		if err := persist.Sync(); err != nil {
 			logging.Warning("Could not sync authenticated state: %s", err.Error())
@@ -189,6 +189,9 @@ func (s *Auth) AuthenticateWithModel(credentials *mono_models.Credentials) error
 		case *apiAuth.PostLoginRetryWith:
 			return errs.AddTips(&ErrTokenRequired{locale.WrapInputError(err, "err_auth_fail_totp")}, tips...)
 		default:
+			if os.IsTimeout(err) {
+				return locale.NewInputError("err_api_auth_timeout", "Timed out waiting for authentication response. Please try again.")
+			}
 			multilog.Error("Authentication API returned %v", err)
 			return errs.AddTips(locale.WrapError(err, "err_api_auth", "Authentication failed: {{.V0}}", err.Error()), tips...)
 		}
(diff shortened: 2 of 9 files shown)


pkg/platform/model/vcs.go
```diff
@@ -424,8 +424,15 @@ func AddChangeset(parentCommitID strfmt.UUID, commitMessage string, changeset Ch
 
 	res, err := mono.New().VersionControl.AddCommit(params, authentication.ClientAuth())
 	if err != nil {
-		multilog.Error("AddCommit Error: %s", err.Error())
-		return nil, locale.WrapError(err, "err_add_commit", "", api.ErrorMessageFromPayload(err))
+		switch err.(type) {
+		case *version_control.AddCommitBadRequest:
+		case *version_control.AddCommitConflict:
+		case *version_control.AddCommitForbidden:
```
daved, line 430: Should these be ignored?

pkg/platform/authentication/auth.go
```diff
@@ -189,6 +189,9 @@ func (s *Auth) AuthenticateWithModel(credentials *mono_models.Credentials) error
 		case *apiAuth.PostLoginRetryWith:
 			return errs.AddTips(&ErrTokenRequired{locale.WrapInputError(err, "err_auth_fail_totp")}, tips...)
 		default:
+			if os.IsTimeout(err) {
+				return locale.NewInputError("err_api_auth_timeout", "Timed out waiting for authentication response. Please try again.")
+			}
```
daved, line 194: @Naatan This feels like it's stretching the idea of InputError. Is this OK?

daved (changes requested):

pkg/platform/model/vcs.go
mitchell-as: I feel like these are also input errors. The BadRequest is if the user submits a bad request (not exactly sure what this entails). Conflict is what was reported to Rollbar -- the user tried to remove a package that did not exist. And Forbidden is an authentication error.

daved: If I'm understanding correctly, the case should then be `case badReq, conflict, forbidden:` since the current form will return the payload without error.

daved: `fallthrough` may also work, but in all my years writing Go I can't think of a time I preferred or even ended up using it.

pkg/platform/authentication/auth.go
Naatan: I think this is a valid input error. Input error for us basically means an error due to a user-side variable that we simply cannot address ourselves.

pkg/platform/model/vcs.go
mitchell-as: Thanks for the tip. I come from C/C++ land where fallthrough is implicit without a `break`. I was expecting the same in Go, but I see that is not the case.

daved (dismissed):

internal/locale/locales/en-us.yaml
```diff
@@ -1690,6 +1690,8 @@ err_no_projectfile:
   other: No activestate.yaml file exists in the current working directory or its parent directories.
 err_project_not_found:
   other: Could not find project at [NOTICE]{{.V0}}[/RESET]
+err_projectfile_parse:
+  other: Unable to parse {{.V0}}. Please try checking out your project again.
```
Naatan, line 1694: We should include the parsing error and remove the blurb about "checking out again". It failing to parse is a legitimate input error that the user is expected to address because the activestate.yaml is meant to be edited by the user.

pkg/platform/model/recipe.go
```diff
@@ -179,7 +181,9 @@ func FetchRecipe(commitID strfmt.UUID, owner, project string, hostPlatform *stri
 
 		orderBody, _ := json.Marshal(params.Order)
 		serr := resolveSolverError(err)
-		multilog.Error("Solver returned error: %s, order: %s", errs.JoinMessage(err), string(orderBody))
+		if !locale.IsInputError(serr) {
```
Naatan, line 184: Are the underlying errors we care about for this story actually input errors?

Naatan (changes requested):

internal/locale/locales/en-us.yaml
mitchell-as: The parse error is shown, but I will remove the blurb about checking out the project again.

```
Script Output
─────────────
% build/state run build
 x Unable to parse /Users/test/cli/activestate.yaml
 x Project file `/Users/test/cli/activestate.yaml` could not be parsed, the parser produced the following error: yaml: line 3: did not find expected key
```

pkg/platform/model/recipe.go
mitchell-as: I believe so. The conflict errors are probably user errors (attempting to remove a non-installed package). Forbidden (authentication) also seems like one. I can leave this bit out if you're not comfortable. The ticket says to use one's best judgement, so this is subjective. I can defer to you.

Naatan: Let's try it.. we can always tune it further :)

Naatan (approved):

PR #2280, opened 2022-12-14
Removed "Platforms" and "Languages" keys from activestate.yaml

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1430" title="DX-1430" target="_blank">TaskDX-1430</a>  Remove old as.yaml properties
  </td></table>
  <br />


internal/constraints/constraints.go
```diff
@@ -400,21 +390,7 @@ func osIsConstrained(constraintOSes string) bool {
...
-				if strings.HasPrefix(name, "-") {
-					return true
-				}
-				constrained = false // can't return here because an exclude might still occur
-			}
-		}
-	}
-
-	return constrained
+	return false // project.Platforms no longer exists
```
MDrakos, line 393: Is this function still needed? Can we remove the platform constraint as well?

pkg/project/expander_test.go
```diff
@@ -182,13 +171,9 @@ func TestExpandProjectSecret(t *testing.T) {
...
 
-	expanded, err := project.ExpandFromProject("${platform.os}", prj)
+	expanded, err := project.ExpandFromProject("${project.name()}", prj)
 	assert.NoError(t, err, "Ran without failure")
-	if runtime.GOOS != "darwin" {
-		assert.Equal(t, runtime.GOOS, expanded, "Expanded platform variable")
-	} else {
-		assert.Equal(t, "macos", expanded, "Expanded platform variable")
-	}
+	assert.Equal(t, "general", expanded, "Expanded platform variable")
```
MDrakos, line 176: This error string should be updated as we are no longer expanding the platform variable

MDrakos (changes requested):
Looks good to me, just some minor comments

internal/constraints/constraints.go
mitchell-as: This is why I filed https://activestatef.atlassian.net/browse/DX-1455. I'd rather remove it there since platform constraints still technically exist.

pkg/project/expander_test.go
mitchell-as: Thanks for catching this.

MDrakos (dismissed):

pkg/project/expander_test.go
```diff
@@ -130,14 +123,10 @@ func TestExpandTopLevel(t *testing.T) {
 func TestExpandProjectPlatformOs(t *testing.T) {
```
Naatan, line 123: The update to this test seems to now be asserting something that is different from the intend. I think we can just drop this test? It was specifically testing the platform property, which has been removed, so it no longer serves a purpose.

Naatan (changes requested):

pkg/project/expander_test.go
mitchell-as: Good catch.

Naatan (approved):

PR #2281, opened 2022-12-14
Clean temp dir from environment definition

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1404" title="DX-1404" target="_blank">StoryDX-1404</a>  Runtime appears to still be using a temp dir
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/implementations/alternative/artifact.go
+++ b/pkg/platform/runtime/setup/implementations/alternative/artifact.go
@@ -34,6 +34,9 @@ func (as *ArtifactSetup) PrepareEnvDef(tmpDir, installDir string, constants envd
 	ed = ed.ExpandVariables(constants)
 
 	// Ensure that the replacement values are set to the correct directory
+	// In some cases the environment definition will contain values that point
+	// to the temporary directory that the artifact was unpacked to. In this case
+	// we need to replace those values with the actual install directory.
 	ed.ReplacePath(filepath.Join(tmpDir, ed.InstallDir), installDir)
 
 	return ed, nil
--- a/pkg/platform/runtime/setup/implementations/alternative/artifact.go
+++ b/pkg/platform/runtime/setup/implementations/alternative/artifact.go
@@ -23,7 +23,23 @@ func NewArtifactSetup(artifactID artifact.ArtifactID, store *store.Store) *Artif
 	return &ArtifactSetup{artifactID, store}
 }
 
-func (as *ArtifactSetup) EnvDef(tmpDir string) (*envdef.EnvironmentDefinition, error) {
+func (as *ArtifactSetup) PrepareEnvDef(tmpDir, installDir string, constants envdef.Constants) (*envdef.EnvironmentDefinition, error) {
+	// Source the environment definition from the extracted archive
+	ed, err := as.newEnvDef(tmpDir)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not load environment definitions for artifact.")
+	}
+
+	// Expand the environment definition variables
+	ed = ed.ExpandVariables(constants)
+
+	// Ensure that the replacement values are set to the correct directory
+	ed.ReplacePath(filepath.Join(tmpDir, ed.InstallDir), installDir)
+
+	return ed, nil
+}
+
+func (as *ArtifactSetup) newEnvDef(tmpDir string) (*envdef.EnvironmentDefinition, error) {
 	path := filepath.Join(tmpDir, constants.RuntimeDefinitionFilename)
 	e, err := envdef.NewEnvironmentDefinition(path)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


MDrakos:
In an attempt to fix the integration tests, I ended up closing the original PR. Here's the link for reference: https://github.com/ActiveState/cli/pull/2270

pkg/platform/runtime/setup/implementations/alternative/artifact.go
```diff
@@ -23,7 +23,23 @@ func NewArtifactSetup(artifactID artifact.ArtifactID, store *store.Store) *Artif
 	return &ArtifactSetup{artifactID, store}
 }
 
-func (as *ArtifactSetup) EnvDef(tmpDir string) (*envdef.EnvironmentDefinition, error) {
+func (as *ArtifactSetup) PrepareEnvDef(tmpDir, installDir string, constants envdef.Constants) (*envdef.EnvironmentDefinition, error) {
```
Naatan, line 26: Can you add a comment explaining the intend of this function? By the naming it sounds like this is some sort of staged function that you're only meant to call at a specific time.

pkg/platform/runtime/setup/implementations/alternative/artifact.go
```diff
@@ -23,7 +23,23 @@ func NewArtifactSetup(artifactID artifact.ArtifactID, store *store.Store) *Artif
...
+	ed, err := as.newEnvDef(tmpDir)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not load environment definitions for artifact.")
+	}
+
+	// Expand the environment definition variables
+	ed = ed.ExpandVariables(constants)
+
+	// Ensure that the replacement values are set to the correct directory
+	ed.ReplacePath(filepath.Join(tmpDir, ed.InstallDir), installDir)
```
Naatan, line 37: I understand there's a chicken and egg problem here. Can you add a comment explaining why this is necessary this way? Looking at the code my first thought is "why use a temp dir here at all?".

pkg/platform/runtime/setup/implementations/alternative/artifact.go
```diff
@@ -23,7 +23,23 @@ func NewArtifactSetup(artifactID artifact.ArtifactID, store *store.Store) *Artif
...
-func (as *ArtifactSetup) EnvDef(tmpDir string) (*envdef.EnvironmentDefinition, error) {
+func (as *ArtifactSetup) PrepareEnvDef(tmpDir, installDir string, constants envdef.Constants) (*envdef.EnvironmentDefinition, error) {
+	// Source the environment definition from the extracted archive
+	ed, err := as.newEnvDef(tmpDir)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not load environment definitions for artifact.")
+	}
+
+	// Expand the environment definition variables
+	ed = ed.ExpandVariables(constants)
```
Naatan, line 34: Why was this moved into the envdef function but ApplyFileTransformations wasn't?

pkg/platform/runtime/setup/setup.go
```diff
@@ -229,21 +229,19 @@ func (s *Setup) updateArtifacts() ([]artifact.ArtifactID, error) {
...
-		envDef, err := as.EnvDef(unpackedDir)
+		envDef, err := as.PrepareEnvDef(tempUnpackedDir, s.store.InstallPath(), cnst)
 		if err != nil {
 			return errs.Wrap(err, "Could not collect env info for artifact")
 		}
 
-		// Expand environment definitions using constants
-		envDef = envDef.ExpandVariables(cnst)
-		err = envDef.ApplyFileTransforms(filepath.Join(unpackedDir, envDef.InstallDir), cnst)
+		err = envDef.ApplyFileTransforms(filepath.Join(tempUnpackedDir, envDef.InstallDir), cnst)
```
Naatan, line 237: What would happen if we removed tempdir from the equation and we just gave this the install path?

This still feels like it's not architected right. Something here is out of order..

Naatan (changes requested):

pkg/platform/runtime/setup/implementations/alternative/artifact.go
MDrakos: I agree the naming isn't perfect. This could still be named `EnvDef` and act as a constructor with added steps.

pkg/platform/runtime/setup/implementations/alternative/artifact.go
MDrakos: This function operates on the environment definition itself whereas ApplyFileTransforms operates on the files passed to it and uses the environment definition. The idea with the prepare function was to do all the preparatory steps required on the environment definition so that it's ready to be used with artifact installation.

pkg/platform/runtime/setup/setup.go
MDrakos: I agree it still doesn't feel right. The correct solution will likely require a larger changeset. That said we can use the install directory but only after we've moved the files. So now we are doing the file relocation after the files have been moved to the install path.

Naatan:
fyi haven't forgotten about this one.. just want to find some time to properly evaluate the code to decide on a way forward.

PR #2282, opened 2022-12-15
Update show output

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1417" title="DX-1417" target="_blank">StoryDX-1417</a>  You should be able to get the executable location on non-default/used runtimes
  </td></table>
  <br />


internal/runners/show/show.go
```diff
@@ -217,14 +224,30 @@ func (s *Show) Run(params Params) error {
 		return locale.WrapError(err, "err_show_secrets", "Could not get secret information")
 	}
 
+	projectDir := filepath.Dir(s.project.Path())
+	if fileutils.IsSymlink(projectDir) {
```
mitchell-as, line 228: I like that you anticipated a possible symlink here.

internal/runners/show/show.go
```diff
@@ -217,14 +224,30 @@ func (s *Show) Run(params Params) error {
...
 		LastCommit:   commit,
+		Location:     projectDir,
 	}
 
+	projectTarget := target.NewProjectTarget(s.project, storage.CachePath(), nil, "").Dir()
+	if projectTarget != "" {
+		rd.Executables = setup.ExecDir(projectTarget)
+	}
+
+	fileutils.IsSymlink(filepath.Dir(s.project.Path()))
```
mitchell-as, line 249: I think you meant to either make this part of a conditional statement or omit it completely.

test/integration/show_int_test.go
```diff
@@ -24,11 +25,29 @@ func (suite *ShowIntegrationTestSuite) TestShow() {
...
+
+	cp = ts.Spawn("show")
 	cp.Expect(`Name`)
 	cp.Expect(`Show`)
 	cp.Expect(`Organization`)
 	cp.Expect(`cli-integration-tests`)
+	cp.Expect(`Namespace`)
+	cp.Expect(`cli-integration-tests/Show`)
+	cp.Expect(`Location`)
+	// MacOS seems to have issues matching this directory
```
mitchell-as, line 42: Could you try a `cp.ExpectLongString()` first? I'm reluctant to omit macOS so quickly, especially after that PythonPath issue with showing tempdir paths.

mitchell-as (changes requested):

internal/runners/show/show.go
MDrakos: Whoops, thanks for catching that!

test/integration/show_int_test.go
MDrakos: Unfortunately, that didn't work. Good idea though. Not sure what's going on with the expect on MacOS

mitchell-as: Okay, maybe @Naatan has an idea given he's been working on refactoring termtest.

mitchell-as (approved):

test/integration/show_int_test.go
Naatan: Is it not matching even though it should or is it not matching cause it comes up with a different direction altogether?

For the former; yeah that's an issue with our current termtest implementation. No solution until I can land my rewrite.

For the latter; try using: https://github.com/ActiveState/cli/blob/79f31adbaf3a18072127fb863ec69a45660faceb/internal/fileutils/fileutils_lin_mac.go#L35-L35

mitchell-as: When I looked at the test output, the directory name looked fine, it was just too long, and the terminal appeared to wrap and indent the output.

MDrakos: `ResolveUniquePath` worked! Thanks @Naatan @mitchell-as

MDrakos: `ResolveUniquePath` worked! Thanks @Naatan @mitchell-as

mitchell-as (approved):
Yay!

PR #2283, opened 2022-12-15
Remove state-tray and update dialog

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1438" title="DX-1438" target="_blank">TaskDX-1438</a>  Drop state-tray from state tool distribution
  </td></table>
  <br />
 

We don't work on them anymore.


internal/runners/prepare/prepare_windows.go
```diff
@@ -143,17 +120,7 @@ func setStateProtocol() error {
...
-	trayExec, err := installation.TrayExec()
-	if err != nil {
-		return nil, locale.WrapError(err, "err_tray_exec")
-	}
-
-	sc := shortcut.New(shortcutDir, constants.TrayAppName, trayExec)
-	files = append(files, filepath.Dir(sc.Path()))
-
-	return files, nil
+	return []string{}, nil
```
daved, line 123: This functions doesn't seem useful any longer.

internal/runners/clean/run_win.go
```diff
@@ -87,14 +87,9 @@ func removeInstall(logFile string, cfg configurable) error {
...
 
-	trayExec, err := installation.TrayExec()
-	if err != nil {
-		return locale.WrapError(err, "err_tray_exec")
-	}
-
 	transitionalStateTool := cfg.GetString(installation.CfgTransitionalStateToolPath)
 	var aggErr error
-	for _, exec := range []string{svcExec, trayExec} {
+	for _, exec := range []string{svcExec} {
```
daved, line 92: The loop should be able to be dropped.

internal/runners/prepare/prepare_linux.go
```diff
@@ -70,22 +46,7 @@ func prependHomeDir(path string) (string, error) {
...
-
-	iconsDir, err := prependHomeDir(constants.IconsDir)
-	if err != nil {
-		multilog.Error("Could not find icons directory: %v", err)
-	} else {
-		files = append(files, filepath.Join(iconsDir, constants.TrayIconFileName))
-	}
-
-	return files, nil
+	return []string{}, nil
```
daved, line 49: This function may not be necessary.

daved (changes requested):

daved (dismissed):

Naatan (changes requested):
Woah the tray was a lot more infectious than I would've though!

Reading through this I realized we should have had an additional AC to cover people updating from versions that still had the tray (our update mechanism doesn't handle deleted files).

I've added the following AC to the jira story:

* Remove state-tray remnants on update
  * Stop running instances of state-tray
  * Remove state-tray and state-update-dialog in install dir
  * Remove application file on macOS
  * Remove autostart files

Sorry for the scope creep.

cmd/state-installer/installer.go
```diff
@@ -133,18 +129,6 @@ func (i *Installer) Install() (rerr error) {
...
-	if trayRunning {
-		trayExec, err := installation.TrayExecFromDir(binDir)
-		if err != nil {
-			return locale.WrapError(err, "err_tray_exec_dir", "", binDir)
-		}
-
-		if _, err := exeutils.ExecuteAndForget(trayExec, []string{}); err != nil {
-			multilog.Error("Could not start state-tray: %s", errs.JoinMessage(err))
-		}
-	}
```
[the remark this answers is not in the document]
mitchell-as: I don't follow. What are we supposed to keep?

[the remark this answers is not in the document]
Naatan: My bad, I commented on the wrong code. I meant the mechanic that stops the state-tray if it is found to be running, prior to performing the install.

cmd/state-installer/installer.go
```diff
@@ -83,6 +80,11 @@ func (i *Installer) Install() (rerr error) {
 		return locale.WrapInputError(err, "err_update_corrupt_install", constants.DocumentationURL)
 	}
 
+	err = legacytray.DetectAndRemove(i.path, i.cfg)
+	if err != nil {
+		multilog.Error("Unable to detect and/or remove legacy tray. Will try again next update. Error: %v", err)
```
mitchell-as, line 85: There is a question of what exactly to do in this case. If the user attempts to uninstall the State Tool after a failed removal here, then state-tray will remain. On the other hand, if we error out during the update, that might cause them trouble.

Naatan: Could we just add this to the uninstaller as well? And have that one error out if it fails.

mitchell-as: It got a bit messy, but I've moved the legacytray package into a shared area and use it from both the installer and uninstaller.

Naatan (changes requested):

cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -219,6 +219,15 @@ func (suite *InstallerIntegrationTestSuite) TestStateTrayRemoval() {
...
 	cp.ExpectExitCode(0)
+
+	// Uninstall.
+	cp = ts.SpawnCmdWithOpts(stateExec, e2e.WithArgs("clean", "uninstall", "--force"))
+	if runtime.GOOS == "windows" {
+		cp.ExpectLongString("Deletion of State Tool has been scheduled.")
+	} else {
+		cp.ExpectLongString("Successfully removed State Tool and related files")
+	}
+	cp.ExpectExitCode(0)
```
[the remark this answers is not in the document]
mitchell-as: I don't remember why exactly I added this. I've removed it.

Naatan (approved):

PR #2288, opened 2022-12-19
`state languages install` does not modify remote project

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1443" title="DX-1443" target="_blank">StoryDX-1443</a>  `state languages install` should not modify remote project
  </td></table>
  <br />
 

Also updates the `platforms` sub commands


Naatan (changes requested):
I think we can instead use this function:

https://github.com/ActiveState/cli/blob/79f31adbaf3a18072127fb863ec69a45660faceb/internal/runners/packages/packages.go#L50-L50

That way all these order operations would function exactly the same way. Additionally it would result in only a single commit rather than two (remove, then add).

internal/runbits/requirements/requirements.go
```diff
@@ -0,0 +1,334 @@
...
+	primer.SvcModeler
+}
+
+type configurable interface {
+	keypairs.Configurable
+}
+
+const latestVersion = "latest"
+
+func ExecuteRequirementOperation(prime primeable, requirementName, requirementVersion string, operation model.Operation, nsType model.NamespaceType) (rerr error) {
```
Naatan, line 61: This seems like you're introducing a ton of new code.. but I'm guessing this is actually just mostly copying ExecutePackageOperation? Unless github is being weird again it does not appear that ExecutePackageOperation went anywhere though? Can you clarify what's happening?

Naatan (changes requested):

internal/runbits/requirements/requirements.go
MDrakos: Whoops. Forgot to remove the `executePackageOperation` function. The new `ExecuteRequirementOperation` is just a modified version of the original package operation function.

internal/runners/platforms/add.go
```diff
@@ -43,13 +43,17 @@ func (a *Add) Run(ps AddRunParams) error {
 		return locale.NewInputError("err_no_project")
 	}
 
-	err = model.CommitPlatform(
-		a.project,
+	commit, err := model.CommitPlatform(
```
Naatan, line 46: I think this should also use ExecuteRequirementOperation.

pkg/platform/model/vcs.go
```diff
@@ -675,89 +675,135 @@ func (cs indexedCommits) countBetween(first, last string) (int, error) {
...
-		return err
-	}
+	return AddCommit(parentCommitID, msg, op, NewNamespacePlatform(), platformID, "")
 
-	return UpdateBranchCommit(branch.BranchID, commit.CommitID)
 }
 
 // CommitLanguage commits a single language to the platform
-func CommitLanguage(pj ProjectInfo, op Operation, name, version string) error {
+func CommitLanguage(commitID strfmt.UUID, op Operation, name, version string) (*mono_models.Commit, error) {
```
Naatan, line 703: This can be dropped I think.

pkg/platform/model/vcs.go
```diff
@@ -675,89 +675,135 @@ func (cs indexedCommits) countBetween(first, last string) (int, error) {
 }
 
 // CommitPlatform commits a single platform commit
-func CommitPlatform(pj ProjectInfo, op Operation, name, version string, word int) error {
+func CommitPlatform(parentCommitID strfmt.UUID, op Operation, name, version string, word int) (*mono_models.Commit, error) {
```
Naatan, line 678: This can be dropped if we move the platform logic to also use ExecuteRequirementOperation.

internal/runbits/requirements/requirements.go
```diff
@@ -79,24 +90,32 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
 		}()
 	} else {
 		language, err := model.LanguageByCommit(pj.CommitUUID())
-		if err == nil {
-			langName = language.Name
+		if err != nil {
+			return locale.WrapError(err, "err_package_get_language", "Could not get language from project")
+		}
```
Naatan, line 95: You're introducing a new error here. This might be desirable for the language operation, but not so for regular packages. I think this will break the behavior on line 109 as it can never get there through this path.

internal/runbits/requirements/requirements.go
```diff
@@ -162,11 +182,8 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
 	}
 
 	var commitID strfmt.UUID
-	commitID, err = model.CommitPackage(parentCommitID, operation, packageName, ns, packageVersion)
+	commitID, err = model.CommitRequirement(parentCommitID, operation, requirementName, requirementVersion, ns)
 	if err != nil {
-		if operation == model.OperationRemoved && strings.Contains(err.Error(), "does not exist") {
-			return locale.WrapInputError(err, "err_package_remove_does_not_exist", "Package is not installed: {{.V0}}", packageName)
-		}
```
Naatan, line 169: This seems like a valuable assertion, why drop it?

Naatan (changes requested):

internal/runbits/requirements/requirements.go
MDrakos: This was just a reaction to an unhandled error. Updated now to at least log the error.

internal/runbits/requirements/requirements.go
```diff
@@ -58,7 +58,7 @@ type configurable interface {
 
 const latestVersion = "latest"
 
-func ExecuteRequirementOperation(prime primeable, requirementName, requirementVersion string, operation model.Operation, nsType model.NamespaceType) (rerr error) {
+func ExecuteRequirementOperation(prime primeable, requirementName, requirementVersion string, bitWidth int, operation model.Operation, nsType model.NamespaceType) (rerr error) {
```
Naatan, line 61: Adding the bitWidth here seems to move even more responsibility to this humongous function. And it's an arg that's ONLY relevant to platform requirements. Why not calculate the platform ID before calling this function and call it with the resolved ID?

ie. move this logic to your runner:

https://github.com/ActiveState/cli/pull/2288/files/a074423c8e6b85288c5b0754a88370094465f2f0..ac973fcdd57b19b39bb2ed7b8742627ace0fbb52#diff-d9612958a129acac4c7a4c03e241b6b912c286635b188a24975afe794fcc94f5R750

Then just run `ExecuteRequirementOperation` with the platform ID.

internal/runbits/requirements/requirements.go
```diff
@@ -90,19 +90,23 @@ func ExecuteRequirementOperation(prime primeable, requirementName, requirementVe
...
 		language, err := model.LanguageByCommit(pj.CommitUUID())
-		if err != nil {
-			return locale.WrapError(err, "err_package_get_language", "Could not get language from project")
+		if err == nil {
+			langName = language.Name
+			if nsType == model.NamespacePackage || nsType == model.NamespaceBundle {
+				ns = model.NewNamespacePkgOrBundle(langName, nsType)
+			}
+		} else {
+			logging.Error("Could not get language from project: %v", err)
```
Naatan, line 99:
```suggestion
			logging.Debug("Could not get language from project: %v", err)
```

This will happen by design for certain use-cases (eg. adding packages when no language has been defined yet).

internal/runbits/requirements/requirements.go
```diff
@@ -81,22 +92,34 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
 		language, err := model.LanguageByCommit(pj.CommitUUID())
 		if err == nil {
 			langName = language.Name
-			ns = model.NewNamespacePkgOrBundle(langName, nsType)
+			if nsType == model.NamespacePackage || nsType == model.NamespaceBundle {
```
Naatan, line 95: Should do this conditional before line 92. There's no point looking up the language unless we are looking at one of these namespaces it seems like.

internal/runbits/requirements/requirements.go
```diff
@@ -81,22 +92,34 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
...
 	}
 
-	var validatePkg = operation == model.OperationAdded
+	switch nsType {
+	case model.NamespaceLanguage:
+		ns = model.NewNamespaceLanguage()
+	case model.NamespacePlatform:
+		ns = model.NewNamespacePlatform()
+	}
+
```
Naatan, line 111: This feels like it should be more closely tied to the logic happening on line 96. Perhaps this switch case is where you look up the language if encountering a package / bundle.

internal/runbits/requirements/requirements.go
```diff
@@ -81,22 +92,34 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
...
+	var validatePkg = operation == model.OperationAdded && ns.Type() == model.NamespacePackage
```
Naatan, line 112: Should this also address bundles?

internal/runners/languages/install.go
```diff
@@ -4,24 +4,30 @@ import (
...
 	"github.com/ActiveState/cli/internal/primer"
+	"github.com/ActiveState/cli/internal/runbits/requirements"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/project"
 )
 
 type Update struct {
-	out     output.Outputer
-	project *project.Project
+	prime primeable
```
Naatan, line 14: This feels inconsistent with how we usually write these. I'm pretty sure we only interpret primeable for the constructor for our other runners.

internal/runners/platforms/remove.go
```diff
@@ -14,19 +14,19 @@ type RemoveRunParams struct {
 
 // Remove manages the removeing execution context.
 type Remove struct {
-	*project.Project
+	prime primeable
```
Naatan, line 17: Inconsistent with how we use primeable in runners (handle it in the constructor).

internal/runners/platforms/remove.go
```diff
@@ -35,9 +35,12 @@ func (r *Remove) Run(ps RemoveRunParams) error {
 		return nil
 	}
 
-	return model.CommitPlatform(
-		r.Project,
-		model.OperationRemoved,
-		params.name, params.version, params.BitWidth,
-	)
+	err = requirements.ExecuteRequirementOperation(r.prime, params.name, params.version, params.BitWidth, model.OperationRemoved, model.NamespacePlatform)
+	if err != nil {
```
Naatan, line 39: Nit: Reusing err vars always triggers red flags for me. I don't recall the exact issue but I recall this can cause issues with asserting nil if the error types the var handles don't match.

Naatan (changes requested):

internal/runbits/requirements/requirements.go
MDrakos: Resolving the platform ID in the runner will result in output like `Installing requirement <UUID>`. I would prefer not to have another argument in this function however in this case it is only passed on to lower-level functions. Resolving the ID early could lead to more complexity in order to get the UX right.

internal/runbits/requirements/requirements.go
```diff
@@ -42,28 +43,29 @@ func (pv *PackageVersion) Set(arg string) error {
...
+	Project             *project.Project
+	Auth                *authentication.Auth
+	Config              *config.Instance
+	Analytics           analytics.Dispatcher
+	SvcModel            *model.SvcModel
+	RequirementName     string
+	RequirementVersion  string
+	RequirementBitWidth int
+	Operation           model.Operation
+	NsType              model.NamespaceType
```
Naatan, line 60: This is mixing different param types. Traditionally we use params for command input as well as struct instances passed down through the runner. 

Think of params as the things you send to a constructor. The rest should just remain arguments.

Apologies though; it didn't register to me that you are not working on a traditional runner but a runbit here. Perhaps it makes sense if you make this a struct with a constructor, which you then call `ExecuteRequirementOperation`.

I realize this is slightly getting into the refactor space. I don't intend for this change request to be more than a trivial amount of line changes. But I want to make sure we at least leave this code in a better place than we found it.

internal/runbits/requirements/requirements.go
```diff
@@ -78,25 +91,38 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
...
 		}()
 	} else {
+		out.Notice(locale.Tl("operating_message", "", pj.NamespaceString(), pj.Dir()))
+	}
+
+	switch params.NsType {
+	case model.NamespacePackage, model.NamespaceBundle:
+		if pj == nil {
+			break
+		}
```
Naatan, line 101: Can't be non-nil due to line 82.

internal/runbits/requirements/requirements.go
```diff
@@ -78,25 +91,38 @@ func executePackageOperation(prime primeable, packageName, packageVersion string
...
-		out.Notice(locale.Tl("operating_message", "", pj.NamespaceString(), pj.Dir()))
+	case model.NamespaceLanguage:
+		ns = model.NewNamespaceLanguage()
+	case model.NamespacePlatform:
+		ns = model.NewNamespacePlatform()
 	}
 
-	var validatePkg = operation == model.OperationAdded
+	var validatePkg = params.Operation == model.OperationAdded && (ns.Type() == model.NamespacePackage || ns.Type() == model.NamespaceBundle)
 	if !ns.IsValid() {
```
Naatan, line 116: We should assert the NsType here as well (ie. is package or bundle), cause the logic below is irrelevant for platforms or languages.

internal/runners/packages/install.go
```diff
@@ -1,27 +1,74 @@
...
+}
+
 // InstallRunParams tracks the info required for running Install.
 type InstallRunParams struct {
 	Package PackageVersion
 }
 
 // Install manages the installing execution context.
 type Install struct {
-	prime primeable
```
Naatan, line 15: Sorry.. I didn't realize that using prime in a non-standard way was a pre-existing issue rather than one you introduced.

pkg/platform/model/checkpoints.go
```diff
@@ -32,16 +32,16 @@ type Language struct {
...
 
 // GetRequirement searches a commit for a requirement by name.
-func GetRequirement(commitID strfmt.UUID, namespace, requirement string) (*gqlModel.Requirement, error) {
+func GetRequirement(commitID strfmt.UUID, namespace Namespace, requirement string) (*gqlModel.Requirement, error) {
 	chkPt, _, err := FetchCheckpointForCommit(commitID)
 	if err != nil {
 		return nil, err
 	}
 
-	chkPt = FilterCheckpointPackages(chkPt)
```
Naatan, line 41: Can we get rid of this function? Seems the only other use-case for it could probably also use `FilterCheckpointNamespace`.

Naatan (changes requested):

pkg/platform/model/checkpoints.go
```diff
@@ -32,16 +32,16 @@ type Language struct {
 }
 
 // GetRequirement searches a commit for a requirement by name.
-func GetRequirement(commitID strfmt.UUID, namespace Namespace, requirement string) (*gqlModel.Requirement, error) {
+func GetRequirement(commitID strfmt.UUID, nsType Namespace, requirement string) (*gqlModel.Requirement, error) {
```
Naatan, line 35: a Namespace and a namespace type are two distinct things. I fear this is sufficiently confusing to create bugs.

pkg/platform/model/vcs.go
```diff
@@ -139,6 +140,10 @@ func (t NamespaceType) Prefix() string {
 	return t.prefix
 }
 
+func (t NamespaceType) Matchable() NamespaceMatchable {
+	return t.matchable
+}
```
Naatan, line 145: This mechanic feels redundant to me. Can't we just match the actual types? ie. `nsTypeIHave == NamespacePackage`.

Naatan (changes requested):

pkg/platform/model/vcs.go
MDrakos: For the `FilterCheckpointNamespace` function the namespace we get back is a string. `NamespaceMatch()` and the `NamespaceMatchable` type make that comparison fairly straightforward.

Naatan (approved):

PR #2291, opened 2022-12-19
Fix TestAutostartConfigEnableDisable

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1468" title="DX-1468" target="_blank">BugDX-1468</a>  `TestAutostartConfigEnableDisable` still failing
  </td></table>
  <br />


mitchell-as (changes requested):
This seems excessive. All that we're waiting for is for a file to be removed. Can we make a loop that checks `as.IsEnabled()` every *x* seconds and times out after a minute or so with a failure? That way if it's fast, we don't lose too much time, but we can still accomodate any slowness.

MDrakos:
> Can we make a loop that checks `as.IsEnabled()` every _x_ seconds and times out after a minute or so with a failure?

Great idea!

mitchell-as (approved):
Looks like black magic to me. Nice!

PR #2294, opened 2022-12-21
Another fix for autostart integration test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1476" title="DX-1476" target="_blank">BugDX-1476</a>  TestAutostartConfigEnableDisable still failing part 3
  </td></table>
  <br />

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -330,7 +330,7 @@ jobs:
             TIMEOUT=60m
           fi
           set +e
-          SHELL='' go test -timeout $TIMEOUT -v `go list ./... | grep "integration\|automation"` &> build/integration-tests.out
+          SHELL='' go test -p 1 -timeout $TIMEOUT -v `go list ./... | grep "integration\|automation" | grep -v -e "svc"` >> build/integration-tests.out
           CODE=$?
           cat build/integration-tests.out | gotestfmt
           exit $CODE
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -332,7 +332,7 @@ jobs:
             TIMEOUT=60m
           fi
           set +e
-          SHELL='' go test $TEST_FLAGS -timeout $TIMEOUT -v `go list ./... | grep "integration\|automation" | grep -v -e "svc"` >> build/integration-tests.out
+          SHELL='' go test $TEST_FLAGS -timeout $TIMEOUT -v `go list ./... | grep "integration\|automation" | grep -v -e "svc"` &> build/integration-tests.out
           CODE=$?
           cat build/integration-tests.out | gotestfmt
           exit $CODE
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -320,6 +320,7 @@ jobs:
             fi
           else
             TEST_SUITE_TAGS="all"
+            # On scheduled runs, we only run one test at a time
             TEST_FLAGS="-p 1"
           fi
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -320,6 +320,7 @@ jobs:
             fi
           else
             TEST_SUITE_TAGS="all"
+            TEST_FLAGS="-p 1"
           fi
 
           echo "Running integration tests with tags: $TEST_SUITE_TAGS (empty means every test not specifically tagged)"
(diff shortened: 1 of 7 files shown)


MDrakos:
I was able to get all of the integration tests to pass with the `-p 1` flag on this PR

.github/workflows/build.yml
```diff
@@ -320,6 +320,8 @@ jobs:
             fi
           else
             TEST_SUITE_TAGS="all"
+            # On scheduled runs, we only run one test at a time
+            TEST_FLAGS="-p 1"
```
MDrakos, line 324: Leaving this here but I'm not sure if we want to keep it.

Naatan: Let's drop it. This seems like it would slow down all our tests significantly and if we don't have a strong use-case for it it doesn't seem worth it.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -207,40 +208,61 @@ func (suite *SvcIntegrationTestSuite) TestAutostartConfigEnableDisable() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
+	autostartDir := filepath.Join(ts.Dirs.Config, "autostart")
+	err := fileutils.Mkdir(autostartDir)
+	suite.Require().NoError(err)
+	err = os.Setenv("_TEST_AUTOSTART_DIR", autostartDir)
```
Naatan, line 214: Please add it to constants.go and follow our naming convention (ie. prefix with ACTIVESTATE).

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -207,40 +208,61 @@ func (suite *SvcIntegrationTestSuite) TestAutostartConfigEnableDisable() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
+	autostartDir := filepath.Join(ts.Dirs.Config, "autostart")
```
Naatan, line 211: I think you should use something under `ts.Dirs.Work`. Config is a curated directory which autostart is not part of.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -207,40 +208,61 @@ func (suite *SvcIntegrationTestSuite) TestAutostartConfigEnableDisable() {
...
+	err := fileutils.Mkdir(autostartDir)
+	suite.Require().NoError(err)
+	err = os.Setenv("_TEST_AUTOSTART_DIR", autostartDir)
+	suite.Require().NoError(err)
+	defer os.Unsetenv("_TEST_AUTOSTART_DIR")
+
+	if runtime.GOOS == "linux" {
+		err = fileutils.Touch(filepath.Join(autostartDir, ".profile"))
```
Naatan, line 219: Do we not write to .profile if it doesn't exist? That seems like a bug and not something we should work around in the test.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -207,40 +208,61 @@ func (suite *SvcIntegrationTestSuite) TestAutostartConfigEnableDisable() {
...
 			toggled, err := as.IsEnabled()
-			suite.Require().NoError(err)
-			if suite.Equal(expect, toggled) {
-				return
+			if err != nil {
+				suite.FailNow("failed to check autostart", err.Error())
+				return false
+			}
+			if expect == toggled {
+				return true
```
Naatan, line 265: Why not do the assertion here? You have failNow above, so this function is not against failing the suite.

Feels weird having this return a bool here, that is then funnelled into a `suite.Require().True()` in the consumer above this one.

Naatan (changes requested):

cmd/state-svc/test/integration/svc_int_test.go
MDrakos: We create the file if it doesn't exist so this isn't necessary.

cmd/state-svc/test/integration/svc_int_test.go
MDrakos: Doing the assertion here creates a lot of unwanted output every time the assertion fails, which I was trying to avoid.

Naatan (approved):

PR #2295, opened 2023-01-04
Secrets should use primer auth instead of legacy auth.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1427" title="DX-1427" target="_blank">BugDX-1427</a>  Race condition in `state secrets get`
  </td></table>
  <br />
 

[DX-1427]: https://activestatef.atlassian.net/browse/DX-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DX-1427]: https://activestatef.atlassian.net/browse/DX-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DX-1427]: https://activestatef.atlassian.net/browse/DX-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -215,7 +215,7 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	conditional := constraints.NewPrimeConditional(auth, pj, sshell.Shell())
 	project.RegisterConditional(conditional)
 	project.RegisterExpander("mixin", project.NewMixin(auth).Expander)
-	project.RegisterExpander("secrets", project.NewSecretPromptingExpander(secretsapi.Get(), prompter, cfg))
+	project.RegisterExpander("secrets", project.NewSecretPromptingExpander(secretsapi.Get(), prompter, cfg, auth))
 
 	// Run the actual command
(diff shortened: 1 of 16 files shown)


daved (approved):
nit: The pattern I've found useful (and relatively common) for function signatures is to have mechanical dependencies first, then data deps, anything being "written to", then any values being worked on if any. So, for example:
```
func Do(ctx context.Context, key string, w io.Writer, v *usefulThing) *usefulThing {
```
So, for this PR, I'd suggest having auth be towards the beginning of the func params, if not first in many (most?) cases.

Naatan (approved):

mitchell-as:
Seeing as how this is all going to be refactored (authentication will likely be moving into the service), I don't think it's worth spending time on this. I'll keep it in mind for future code.

PR #2297, opened 2023-01-05
Update yaml expansion performance tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1481" title="DX-1481" target="_blank">BugDX-1481</a>  Nightly failure: TestPerformanceYamlIntegrationTestSuite sporadically failing
  </td></table>
  <br />


MDrakos:
Ran the performance tests a few times and other than some random failures it appears that the delay between the first and second runs fixes the issues. If this persists on the nightlies we can look at increasing the `variance` for more leeway and the `samples` to help lower the average.

mitchell-as (approved):

PR #2302, opened 2023-01-10
`state revert` reverts changes from a commit, rather than reverting to the commit itself.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1424" title="DX-1424" target="_blank">StoryDX-1424</a>  state revert should revert "a" commit rather than "to" a commit.
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/vcs.go
+++ b/pkg/platform/model/vcs.go
@@ -888,23 +888,30 @@ func GetRevertCommit(from, to strfmt.UUID) (*mono_models.Commit, error) {
 	return res.Payload, nil
 }
 
-func RevertCommitWithinHistory(from, to strfmt.UUID) (*mono_models.Commit, error) {
-	ok, err := CommitWithinCommitHistory(from, to)
+func RevertCommitWithinHistory(from, to, latest strfmt.UUID) (*mono_models.Commit, error) {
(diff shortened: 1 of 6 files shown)


pkg/platform/model/vcs.go
```diff
@@ -888,23 +888,30 @@ func GetRevertCommit(from, to strfmt.UUID) (*mono_models.Commit, error) {
 	return res.Payload, nil
 }
 
-func RevertCommitWithinHistory(from, to strfmt.UUID) (*mono_models.Commit, error) {
-	ok, err := CommitWithinCommitHistory(from, to)
+func RevertCommitWithinHistory(from, to, latest strfmt.UUID) (*mono_models.Commit, error) {
+	ok, err := CommitWithinCommitHistory(latest, from)
```
daved, line 892: Maybe we should also check if "to" is also within the commit history.

daved (approved):

pkg/platform/model/vcs.go
mitchell-as: If the 'to' commit is the initial commit, then the platform will return an error because it cannot fetch any prior history, and thus cannot verify if the commit is within its history.

PR #2303, opened 2023-01-11
State service autostarts as an .app on macOS

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1435" title="DX-1435" target="_blank">StoryDX-1435</a>  The state-svc autostarts as an .app on macOS
  </td></table>
  <br />


cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -38,11 +39,15 @@ func (suite *InstallerIntegrationTestSuite) TestInstallFromLocalSource() {
...
 
+	dir, err := ioutil.TempDir("", "system*")
+	suite.NoError(err)
+
 	// Run installer with source-path flag (ie. install from this local path)
 	cp := ts.SpawnCmdWithOpts(
 		suite.installerExe,
 		e2e.WithArgs(target),
 		e2e.AppendEnv(constants.DisableUpdates+"=false"),
+		e2e.AppendEnv(fmt.Sprintf("%s=%s", constants.OverwriteDefaultSystemPathEnvVarName, dir)),
```
mitchell-as, line 50: Seeing this, I wonder if we can use something similar for the failing autostart unit tests so that they can run concurrently, but not overwrite each other. Thanks for giving me the idea :)

mitchell-as (approved):
Nicely done. I like how you merged the ideas of autostart and app into one place.

internal/installation/app/app.go
```diff
@@ -0,0 +1,66 @@
+package app
```
Naatan, line 1: Can we move this into the top level internal dir.

cmd/state-installer/installer.go
```diff
@@ -100,6 +102,17 @@ func (i *Installer) Install() (rerr error) {
 		return errs.Wrap(err, "Failed to copy installation files to dir %s. Error received: %s", i.path, errs.JoinMessage(err))
 	}
 
+	// Install the state service as an app if necessary
+	app, err := i.installSvcApp()
+	if err != nil {
+		return errs.Wrap(err, "Installation of service app failed.")
+	}
+
+	err = app.EnableAutostart()
```
Naatan, line 111: App and autostart should be separate concepts that have no awareness of each (ie. neither should import the other). Instead this linking happens in the consuming code.

So here you might do something like:

```
autostart.New(app.Path())
```

cmd/state-installer/installer.go
```diff
@@ -156,6 +169,25 @@ func (i *Installer) sanitizeInput() error {
 	return nil
 }
 
+func (i *Installer) installSvcApp() (*app.App, error) {
```
Naatan, line 172: As discussed; it's a bit of a worry that we are generating this in the installer, rather than at build time. Not going to iterate on this at this time, but worth commenting in the code.

internal/assets/contents/com.activestate.platform.app.plist.tpl
```diff
@@ -0,0 +1,16 @@
...
+        <key>CFBundleExecutable</key>
+        <string>{{.Exec}}</string>
+        <key>CFBundleIconFile</key>
+        <string>{{.Icon}}</string>
+        <key>CFBundleIdentifier</key>
+        <string>com.activestate.platform.{{.Exec}}</string>
+        <key>NSHighResolutionCapable</key>
+        <true/>
+        <key>LSUIElement</key>
+        <true/>
```
Naatan, line 14: These seem irrelevant for CLI applications. Maybe we need to have some sort of setting for the App package that says "this is a UI app" vs "this is a CLI app".

internal/installation/app/app_darwin.go
```diff
@@ -0,0 +1,244 @@
...
+var targetDirs = []string{
+	"/Contents",
+	"/Contents/MacOS",
+	"/Contents/Resources",
+}
+
+func (a *App) install() error {
+	// Create all of the necessary directories and files in a temporary directory
+	// Then move the temporary directory to the final location which for macOS will be the Applications directory
+	tmpDir, err := ioutil.TempDir("", "state-svc-")
```
Naatan, line 36: `state-svc-` seems to be assuming we are setting up a svc app, but at this level we should not have any knowledge about what app we are configuring. Maybe use a.Name.

internal/installation/app/app_darwin.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return errs.Wrap(err, "Could not create temporary directory")
+	}
+	defer os.RemoveAll(tmpDir)
+
+	tmpAppPath := filepath.Join(tmpDir, fmt.Sprintf("%s.app", a.Name))
+	err = fileutils.Mkdir(tmpAppPath)
+	if err != nil {
```
Naatan, line 44: Please don't reuse err vars.

internal/installation/app/app_darwin.go
```diff
@@ -0,0 +1,244 @@
...
+
+	installDir, err := installation.ApplicationInstallPath()
+	if err != nil {
+		return errs.Wrap(err, "Could not get installation path")
+	}
+
+	err = fileutils.MoveAllFiles(tmpDir, installDir)
+	if err != nil {
+		return errs.Wrap(err, "Could not move .app to Applications directory")
+	}
```
Naatan, line 84: Can we try to use go/embed and put this whole file structure under assets? Then it should just be a matter of copying a single directory and doing some post-processing stuff.

Naatan (changes requested):

internal/assets/contents/com.activestate.platform.app.plist.tpl
MDrakos: We will want to keep `LSUIElement` as that denotes that the app runs in the background and there isn't an icon in the doc. Added an option for that as well.

internal/app/app_darwin.go
```diff
@@ -0,0 +1,189 @@
...
+
+	info, err := appDir.Stat()
+	if err != nil {
+		return errs.Wrap(err, "Could not stat app directory asset")
+	}
+
+	err = generateAppDir(tmpAppPath, info.Name())
+	if err != nil {
+		return errs.Wrap(err, "Could not generate app directory")
+	}
```
MDrakos, line 51: This didn't end up where I would have liked it. Unfortunately, the embedded file system doesn't have enough functionality to simply copy one directory to another. So I can reflect the directory structure in our assets are recreate it from that.

cmd/state-installer/installer.go
```diff
@@ -108,8 +110,7 @@ func (i *Installer) Install() (rerr error) {
 		return errs.Wrap(err, "Installation of service app failed.")
 	}
 
-	err = app.EnableAutostart()
-	if err != nil {
+	if err = autostart.Enable(app.Exec, svcAutostart.Options); err != nil {
```
Naatan, line 113: Let's move this into `installSvcApp`?

internal/app/app.go
```diff
@@ -0,0 +1,41 @@
+package app
+
+import (
+	"github.com/ActiveState/cli/internal/constants"
+	configMediator "github.com/ActiveState/cli/internal/mediators/config"
+)
+
+func init() {
+	configMediator.RegisterOption(constants.AutostartSvcConfigKey, configMediator.Bool, configMediator.EmptyEvent, configMediator.EmptyEvent)
```
Naatan, line 9: This should be in the autostart package.

internal/app/app_darwin.go
Naatan: It doesn't have a convenient function for it, but you should be able to do it by iterating over `fs.ReadDir()` and then copying the individual files it comes across and entering into subdirs. You can probably repurpose this function to allow supplying the embed fs:

https://github.com/ActiveState/cli/blob/34b6f33e804560166c6df72bf97ddb24f3ddf9b2/internal/fileutils/fileutils.go#L690-L690

internal/app/app_darwin.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+const (
+	execFileSource   = "exec.sh.tpl"
+	launchFileSource = "com.activestate.platform.app.plist.tpl"
+	iconFile         = "icon.icns"
+	assetAppDir      = "placeholder.app"
+)
+
+func (a *App) install() error {
```
Naatan, line 24: This still has too much logic for my taste, I responded to your comment about using go:embed, I still think that's the way to go.

internal/osutils/autostart/autostart_darwin.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func installPath(name string) (string, error) {
+	dir, err := user.HomeDir()
+	if err != nil {
+		return "", errs.Wrap(err, "Could not get home directory")
+	}
+	path := filepath.Join(dir, "Applications", fmt.Sprintf("%s.app", name))
+	return path, nil
+}
```
Naatan, line 104: This seems overly opinionated for autostart. We should just supply it the app path, it should not care about what it is.

Naatan (changes requested):

internal/app/app.go
MDrakos: Turns out it already is and this is unnecessary.

internal/osutils/autostart/autostart_darwin.go
MDrakos: Updated this to use a function from the `installation` package and not assume the `Applications` directory. This change doesn't require us to update the autostart API

internal/fileutils/fileutils.go
```diff
@@ -681,6 +696,56 @@ func WriteTempFileToDir(dir, pattern string, data []byte, perm os.FileMode) (str
 	return f.Name(), nil
 }
 
+func CopyFilesFromAssets(assetDir, dst string) error {
```
Naatan, line 699: I'd like to keep `fileutils` unopinionated about the rest of our codebase. Granted this is a ship that has sailed, but I'd like to get back to that point eventually so we can treat this as a standalone package that we might even open source.

Instead of giving this package opinions about assets, we could give the asset package a way of interacting with it via a struct rather than on the package top level. Then just turn this function into one that uses an interface with `ReadDir()`.

internal/assets/assets.go
```diff
@@ -5,6 +5,11 @@ import (
 	iofs "io/fs"
 )
 
+const (
+	// PlaceholderFileName is the name of the file that is used to populate the directory structure of the embedded assets.
+	PlaceholderFileName = "placeholder"
```
Naatan, line 10: Is this to work around git not tracking empty folders? This might be more appropriate:

https://stackoverflow.com/a/932982/490321

internal/fileutils/fileutils.go
```diff
@@ -681,6 +696,56 @@ func WriteTempFileToDir(dir, pattern string, data []byte, perm os.FileMode) (str
...
+			if err != nil {
+				return errs.Wrap(err, "os.Remove %s failed", path)
+			}
+		}
+
+		return nil
+	})
+	if walkErr != nil {
+		return errs.Wrap(walkErr, "filepath.Walk %s failed", dst)
+	}
```
Naatan, line 744: This feels pointless. We are copying all files and then walking over the files we copied to remove these.. why not skip copying these in the first place (line ~721)?

Naatan (changes requested):

internal/assets/assets.go
MDrakos: This is a workaround for assets. Without the placeholder files, it will not copy the directory structure.

internal/fileutils/fileutils.go
MDrakos: 🤦‍♂️

internal/fileutils/fileutils.go
```diff
@@ -696,14 +696,18 @@ func WriteTempFileToDir(dir, pattern string, data []byte, perm os.FileMode) (str
...
-	entries, err := assets.ReadDir(assetDir)
+type DirReader interface {
+	ReadDir(string) ([]os.DirEntry, error)
+}
+
+func CopyFilesDirReader(reader DirReader, src, dst string) error {
+	entries, err := reader.ReadDir(src)
 	if err != nil {
-		return errs.Wrap(err, "asset.ReadDir %s failed", assetDir)
+		return errs.Wrap(err, "asset.ReadDir %s failed", src)
```
Naatan, line 706:
```suggestion
		return errs.Wrap(err, "reader.ReadDir %s failed", src)
```

internal/fileutils/fileutils.go
```diff
@@ -713,34 +717,25 @@ func CopyFilesFromAssets(assetDir, dst string) error {
...
-			err := CopyAsset(srcPath, destPath)
+		case os.ModeSymlink:
+			err := CopySymlink(srcPath, destPath)
 			if err != nil {
-				return errs.Wrap(err, "CopyFile %s:%s failed", srcPath, destPath)
+				return errs.Wrap(err, "CopySymlink %s:%s failed", srcPath, destPath)
+			}
+		default:
+			if entry.Name() == assets.PlaceholderFileName {
+				continue
```
Naatan, line 731: This is opinionating it towards assets again. Perhaps provide an argument where you can pass in a custom filter function.

internal/runners/clean/run_lin_mac.go
```diff
@@ -47,6 +49,12 @@ func (u *Uninstall) runUninstall(params *UninstallParams) error {
 		aggErr = locale.WrapError(aggErr, "uninstall_remove_executables_err", "Failed to remove all State Tool files in installation directory")
 	}
 
+	err = removeApp()
+	if err != nil {
```
Naatan, line 53:
```suggestion
	if err := removeApp(); err != nil {
```

internal/runners/clean/run_lin_mac.go
```diff
@@ -127,6 +135,25 @@ func removeInstall(cfg *config.Instance) error {
...
+func removeApp() error {
+	svcExec, err := installation.ServiceExec()
+	if err != nil {
+		return locale.WrapError(err, "err_svc_exec")
+	}
+
+	svcApp, err := app.New(constants.SvcAppName, svcExec, []string{"start"}, svcApp.Options)
+	if err != nil {
+		return locale.WrapError(err, "err_autostart_app")
+	}
```
Naatan, line 147: Seems this has to stay in sync with the install step or this will break, maybe centralize this somewhere?

Naatan (changes requested):

Naatan (approved):

PR #2304, opened 2023-01-11
Do not raise errors if update checks are blocked or the service is not available.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1479" title="DX-1479" target="_blank">BugDX-1479</a>  Rollbar: MUST ADDRESS: Error does not have localization: Failed to check for update
  </td></table>
  <br />
 

This is a networking issue, not the State Tool's.


Naatan (changes requested):
Looks good! But please add a log entry so it's clear why we're not getting the update if ever we run into this.

Naatan (changes requested):
Comment is helpful too! But I asked for a log entire (logging.Warning seems appropriate).

mitchell-as:
🤦

Naatan (approved):

PR #2307, opened 2023-01-11
Show executable runtime directories in `state projects`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1493" title="DX-1493" target="_blank">StoryDX-1493</a>  `state projects` includes runtime executable dir
  </td></table>
  <br />
 

There are separate output formats for plain text and JSON.

### DIFF
--- a/internal/runners/projects/projects.go
+++ b/internal/runners/projects/projects.go
@@ -4,11 +4,14 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/ActiveState/cli/internal/installation/storage"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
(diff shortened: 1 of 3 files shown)


internal/runners/projects/projects.go
```diff
@@ -18,15 +21,32 @@ type projectWithOrg struct {
 	Name           string   `json:"name"`
 	Organization   string   `json:"organization"`
 	LocalCheckouts []string `json:"local_checkouts,omitempty"`
+	Executables    []string `json:"executables,omitempty"`
 }
 
 type projectWithOrgs []projectWithOrg
 
+func newProjectWithOrg(name, org string, checkouts []string) *projectWithOrg {
```
MDrakos, line 29: I don't think you'll need the pointer return type here as the slice value is a pointer to the backing array.

MDrakos (changes requested):

internal/runners/projects/projects.go
mitchell-as: You have eagle eyes. Thanks for noticing that.

MDrakos (approved):

PR #2324, opened 2023-01-25
DO NOT MERGE: Remove offline installer from repo

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1407" title="DX-1407" target="_blank">TaskDX-1407</a>  Offline installer uses its own github project
  </td></table>
  <br />


daved:
Marking as draft to prevent merging. This branch is only needed temporarily (the newest commit will be pointed to by the offline-installer repo). Once the offline-installer repo is refactored, this branch can be safely removed. A new branch/pr should be established which touches far less files.

PR #2325, opened 2023-01-26
State checkout has a `--set-cache` flag

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1486" title="DX-1486" target="_blank">StoryDX-1486</a>  State checkout takes a flag that specifies the runtime dir
  </td></table>
  <br />


cmd/state/internal/cmdtree/checkout.go
```diff
@@ -24,6 +24,11 @@ func newCheckoutCommand(prime *primer.Values) *captain.Command {
 				Description: locale.Tl("flag_state_checkout_branch_description", "Defines the branch to checkout"),
 				Value:       &params.Branch,
 			},
+			{
+				Name:        locale.Tl("flag_state_checkout_cache", "set-cache"),
+				Description: locale.Tl("flag_state_checkout_cache_description", "Defines the cache to use"),
```
mitchell-as, line 29: This seems pretty vague. Is this the language we want to use?

pkg/projectfile/projectfile.go
```diff
@@ -972,9 +975,51 @@ func createCustom(params *CreateParams, lang language.Language) (*Project, error
...
+	}
+
+	fileContents, err := strutils.ParseTemplate(string(tplContents), data)
+	if err != nil {
+		return errs.Wrap(err, "Could not parse %s", tplName)
+	}
+
+	// Username on Windows is  DOMAIN\username, we only want the username
+	// rather than trimming the domain, we just take the last part of the
+	// home directory's path
```
mitchell-as, line 1011: Some users have different usernames than their home directory (not just on Windows, but on any platform). Some usernames may also have spaces in them. Is this going to cause any trouble with the logic here?

mitchell-as (approved):
This looks fine in principle to me, but I made some comments worth considering.

pkg/projectfile/projectfile.go
MDrakos: Good point. Updated the username handling for windows and also ensure that the username has no spaces on any platforms.

mitchell-as (dismissed):

pkg/projectfile/projectfile.go
```diff
@@ -972,9 +975,46 @@ func createCustom(params *CreateParams, lang language.Language) (*Project, error
...
+		return errs.Wrap(err, "Could not read asset")
+	}
+
+	fileContents, err := strutils.ParseTemplate(string(tplContents), data)
+	if err != nil {
+		return errs.Wrap(err, "Could not parse %s", tplName)
+	}
+
+	// Trim any non-alphanumeric characters from the username
+	nonAlphanumericRegex := regexp.MustCompile(`[^a-zA-Z0-9 ]+`)
```
Naatan, line 1010: Please put this at the top in the root of the scope.

pkg/projectfile/projectfile.go
```diff
@@ -972,9 +975,46 @@ func createCustom(params *CreateParams, lang language.Language) (*Project, error
 		return nil, err
 	}
 
+	if params.Cache != "" {
+		createErr := createCacheFile(params.Directory, params.Cache)
```
Naatan, line 979: Can we name this `createHostFile()`? Cache is just the first usage of it. But I foresee we may want to use this for other things in the future.

internal/assets/contents/activestate.yaml.cache.tpl
```diff
@@ -0,0 +1 @@
+cache: {{ .Cache }}
```
Naatan, line 1:
```suggestion
# It is recommended that you do not commit this file as it contains configuration that is specific to your machine
cache: {{ .Cache }}
```

Naatan (changes requested):

internal/assets/contents/activestate.yaml.cache.tpl
```diff
@@ -0,0 +1,2 @@
+# It is recommended that you do not commit this file as it contains configuration that is specific to your machine
+cache: {{ .Cache }}
```
Naatan, line 2:
```suggestion
cache: {{ .Cache }}

```

Naatan (changes requested):
Looks good! Really a minor change just to add a line break at the end. Normally not something I'd nitpick about but that file goes out to users.

Naatan (approved):

PR #2327, opened 2023-01-26
Do not use terminal escape sequences in non-interactive terminals or the Windows command prompt.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1527" title="DX-1527" target="_blank">BugDX-1527</a>  Occasional panic errors and continues gibberish during runtime progress on Windows
  </td></table>
  <br />

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -82,7 +82,7 @@ func main() {
 
 	// Set up our output formatter/writer
 	outFlags := parseOutputFlags(os.Args)
-	out, err := initOutput(outFlags, "")
+	out, err := initOutput(outFlags, "", subshell.New(cfg).Shell())
 	if err != nil {
 		multilog.Critical("Could not initialize outputer: %s", errs.JoinMessage(err))
 		os.Stderr.WriteString(locale.Tr("err_main_outputer", err.Error()))
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -82,7 +82,8 @@ func main() {
 
 	// Set up our output formatter/writer
 	outFlags := parseOutputFlags(os.Args)
-	out, err := initOutput(outFlags, "", subshell.New(cfg).Shell())
+	shellName, _ := subshell.DetectShell(cfg)
+	out, err := initOutput(outFlags, "", shellName)
 	if err != nil {
 		multilog.Critical("Could not initialize outputer: %s", errs.JoinMessage(err))
 		os.Stderr.WriteString(locale.Tr("err_main_outputer", err.Error()))
--- a/internal/output/progress.go
+++ b/internal/output/progress.go
@@ -9,11 +9,12 @@ import (
 const moveCaretBackEscapeSequence = "\x1b[%dD" // %d is the number of characters to move back
 
 type Spinner struct {
-	frame    int
-	frames   []string
-	out      Outputer
-	stop     chan struct{}
-	interval time.Duration
+	frame         int
+	frames        []string
+	out           Outputer
+	stop          chan struct{}
+	interval      time.Duration
+	reportedError bool
 }
 
 var _ Marshaller = &Spinner{}
@@ -27,7 +28,7 @@ func StartSpinner(out Outputer, msg string, interval time.Duration) *Spinner {
 	if out.Config().Interactive {
 		frames = []string{`|`, `/`, `-`, `\`}
 	}
-	d := &Spinner{0, frames, out, make(chan struct{}, 1), interval}
+	d := &Spinner{0, frames, out, make(chan struct{}, 1), interval, false}
 
 	if msg != "" {
 		d.out.Fprint(d.out.Config().ErrWriter, strings.TrimSuffix(msg, " ")+" ")
--- a/internal/output/progress.go
+++ b/internal/output/progress.go
@@ -1,9 +1,12 @@
 package output
 
 import (
+	"fmt"
 	"strings"
 	"time"
 )
+
+const moveCaretBackEscapeSequence = "\x1b[%dD" // %d is the number of characters to move back
 
 type Spinner struct {
 	frame    int
@@ -47,7 +50,11 @@ func (d *Spinner) moveCaretBack() int {
 		prevPos = len(d.frames) - 1
 	}
 	prevFrame := d.frames[prevPos]
-	d.moveCaretBackInTerminal(len(prevFrame))
+	if d.out.Config().ShellName != "cmd" { // cannot use subshell/cmd.Name due to import cycle
+		d.moveCaretBackInTerminal(len(prevFrame))
+	} else {
+		d.moveCaretBackInCommandPrompt(len(prevFrame))
+	}
 
 	return len(prevFrame)
 }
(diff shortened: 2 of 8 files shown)


internal/output/progress.go
```diff
@@ -96,3 +100,7 @@ func (d *Spinner) Stop(msg string) {
 
 	d.out.Fprint(d.out.Config().ErrWriter, "\n")
 }
+
+func isInteractive(out Outputer) bool {
+	return out.Config().Interactive && terminal.IsTerminal(int(os.Stdin.Fd())) && (runtime.GOOS != "windows" || os.Getenv("SHELL") != "")
```
Naatan, line 105: We can't rely on `runtime.GOOS` as Windows also has bash shells. Windows and cmd.exe are not the same.

Naatan (changes requested):
Interactivity isn't the issue here. My forgetting to address cmd.exe when I implemented this is. There's a lib I found that should handle it cross-platform:

https://github.com/atomicgo/cursor

BUT it has the same issue as your code; it assumes windows==cmd.exe, whereas in reality windows also runs bash, and it's not an uncommon use-case.

That said; you can probably use the lib above to figure out how to do it with cmd.exe. I realize this is more of a missed functionality issue than a bug though, so if this turns into a timesink we could descope the windows implementation and utilize interactive=false for windows for the bug (but please only if it turns into a timesink).

Regardless of which route we go; it's not the job of the output package to assert interactive or detect shells. So we'll want to find a way to pass that in externally, this seems like a reasonable location:

https://github.com/ActiveState/cli/blob/4e7692e37d5d9602945a34fb352faed519d21c8e/cmd/state/output.go#L59

For shell detection we can use the logic here:

https://github.com/ActiveState/cli/blob/4e7692e37d5d9602945a34fb352faed519d21c8e/internal/subshell/subshell.go#L91

I suggest splitting that out into its own function so we don't have to do `subshell.New()`.

There's a lot in my response here.. happy to jump on zoom to go over things in greater detail.

internal/output/progress.go
mitchell-as: FYI: that's what the `os.Getenv("SHELL") != ""` is for. Bash on Windows will have a non-empty "SHELL" env.

Naatan: > FYI: that's what the `os.Getenv("SHELL") != ""` is for. Bash on Windows will have a non-empty "SHELL" env.

Oh I had not caught onto that, that's clever! Still, I'd rather we centralized this behavior in subshell. Such that wherever we detect the subshell it's always the same mechanic.

mitchell-as:
I ended up using `subshell.New()` because it doesn't look very expensive and the result is cleaner. I could switch it to `subshell.DetectShellBinary()` and check the file basename for cmd.exe if you prefer.

cmd/state/main.go
```diff
@@ -82,7 +82,7 @@ func main() {
 
 	// Set up our output formatter/writer
 	outFlags := parseOutputFlags(os.Args)
-	out, err := initOutput(outFlags, "")
+	out, err := initOutput(outFlags, "", subshell.New(cfg).Shell())
```
Naatan, line 85: I'd still like to advocate splitting the shell detection logic out. It's true that `New()` isn't very costly, but it's still doing more than we need it to. Plus we don't know how it will evolve in the future, nor will we be sensitive to this use-case in the future unless we make it explicit.

internal/output/progress.go
```diff
@@ -50,7 +50,11 @@ func (d *Spinner) moveCaretBack() int {
 		prevPos = len(d.frames) - 1
 	}
 	prevFrame := d.frames[prevPos]
-	d.out.Fprint(d.out.Config().ErrWriter, fmt.Sprintf(moveCaretBack, len(prevFrame)))
+	if d.out.Config().ShellName != "cmd" { // cannot use subshell/cmd.Name due to import cycle
```
Naatan, line 53: Let's use this instead, minimizing hard coded values:

https://github.com/ActiveState/cli/blob/2f23dcf71b56d5f188119d2e5fc4f4eaa80a6175/internal/subshell/cmd/cmd.go#L33

internal/output/progress_unix.go
```diff
@@ -0,0 +1,8 @@
+//go:build linux || darwin
+// +build linux darwin
+
+package output
+
+func (d *Spinner) moveCaretBackInCommandPrompt(n int) {
+	// No-op
```
Naatan, line 7: Let's send a rollbar error if this happens, since it should never happen.

internal/output/progress_win.go
```diff
@@ -0,0 +1,50 @@
...
+	handle := syscall.Handle(os.Stdout.Fd())
+
+	var csbi consoleScreenBufferInfo
+	_, _, _ = procGetConsoleScreenBufferInfo.Call(uintptr(handle), uintptr(unsafe.Pointer(&csbi)))
+
+	var cursor coord
+	cursor.x = csbi.cursorPosition.x + short(-n)
+	cursor.y = csbi.cursorPosition.y
+
+	_, _, _ = procSetConsoleCursorPosition.Call(uintptr(handle), uintptr(*(*int32)(unsafe.Pointer(&cursor))))
```
Naatan, line 49: Are we purposely ignoring the errors here? If so please comment why.

Naatan (changes requested):

internal/output/progress.go
mitchell-as: My comment next to this conditional addresses this. I wanted to use the constant, but I cannot due to an import cycle.

internal/output/progress_unix.go
mitchell-as: I thought about that, but then we'll hammer rollbar on every tick.

internal/output/progress_win.go
mitchell-as: One of them should not be ignored. The other can be for the same reason above (logging errors on every tick). I've added a comment.

mitchell-as:
Let me know if I missed something extracting the shell detection logic into its own function.

internal/output/progress.go
Naatan: Ok.. not that big a deal. Let's go with this.

internal/output/progress_unix.go
Naatan: Could just only send it the first time. Record it to something like `Spinner.recorded = true`.

internal/subshell/subshell.go
```diff
@@ -117,20 +107,20 @@ func New(cfg sscommon.Configurable) SubShell {
...
+			path = resolveBinaryPath("zsh")
 			subs = &zsh.SubShell{}
 		default:
-			binary = "bash"
+			path = resolveBinaryPath("bash")
 			subs = &bash.SubShell{}
 		}
-		binary = resolveBinaryPath(binary)
+
 	}
```
Naatan, line 120: This should be part of the shell detection.

Naatan (changes requested):
Please also see my replies to previous comments.

internal/output/progress_unix.go
mitchell-as: Aye, but is there a chance we'll be creating a large number of these progress objects? We'd still be in the same boat.

Naatan (approved):
Nice, the subshell stuff looks much more readable!

PR #2332, opened 2023-01-30
Clear interactive progress ticker after completion.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1535" title="DX-1535" target="_blank">BugDX-1535</a>  Spinner isn't cleared during runtime progress
  </td></table>
  <br />

### DIFF
--- a/internal/output/progress.go
+++ b/internal/output/progress.go
@@ -92,6 +92,13 @@ func (d *Spinner) Stop(msg string) {
 	d.stop <- struct{}{}
 	close(d.stop)
 
+	if d.out.Config().Interactive {
+		nMoved := d.moveCaretBack()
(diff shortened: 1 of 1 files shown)


internal/output/progress.go
```diff
@@ -92,6 +92,13 @@ func (d *Spinner) Stop(msg string) {
 	d.stop <- struct{}{}
 	close(d.stop)
 
+	if d.out.Config().Interactive {
```
Naatan, line 95:
```suggestion
    // We're done, so remove the last spinner frame
	if d.out.Config().Interactive {
```

Naatan (changes requested):

Naatan (approved):

PR #2334, opened 2023-01-30
Handle unhandled err in installFromBuildLog func

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1531" title="DX-1531" target="_blank">BugDX-1531</a>  Panic during runtime progress on Windows
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -617,6 +617,9 @@ func (s *Setup) installFromBuildLog(buildResult *model.BuildResult, artifacts ar
 	defer cancel()
 
 	buildLog, err := buildlog.New(ctx, artifacts, s.eventHandler, *buildResult.Recipe.RecipeID, logFilePath)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


Naatan (approved):

daved:
@Naatan 
Returning the error using `errs.Wrap` results in the user-facing output of
``` x Could not install dependencies```
The causes for this flow are:
1. Cannot connect to the build streamer
2. Could not handle buildstarted event
3. Could not write to websocket

I think a clearer message is worthwhile.

Naatan:
@daved I think the current message is appropriate. The underlying issue is entirely too technical to communicate to the user in a way that is useful.

API's being unreachable is going to bubble up in awkward ways in the state tool, we definitely haven't optimized for that. That can be something we look into in the future, but it's separate from the issue at hand.

Naatan (approved):

Naatan:
@daved went ahead and merged it as I'm blocked by this.

PR #2336, opened 2023-01-31
Improve executor svc message error text for CI only

daved:


daved:
Ignore this in favor of #2337

PR #2338, opened 2023-01-31
Fix state-svc logs not appearing in integration tests

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1542" title="DX-1542" target="_blank">BugDX-1542</a>  state-svc logs always empty on CI
  </td></table>
  <br />
 

Upon fixing the issue I realized how silly the old implementation was. Integration tests already create their own config dir, so there would only ever be logs relevant to the test in question. So instead of any sort of "latest log" logic I rewrote it to just dump all the logs in the log directory.

Additionally I added some small improvements to make the logs more digestible and generally useful.

### DIFF
--- a/internal/testhelpers/e2e/session.go
+++ b/internal/testhelpers/e2e/session.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strings"
 	"testing"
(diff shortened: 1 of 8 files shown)


internal/testhelpers/e2e/session.go
```diff
@@ -614,45 +568,36 @@ func (s *Session) SvcLog() string {
...
-	var newest time.Time
+
+	var sectionStart, sectionEnd string
+	sectionStart = "\n=== "
+	if os.Getenv("GITHUB_ACTIONS") == "true" {
+		sectionStart = "##[group]"
+		sectionEnd = "##[endgroup]"
+	}
+
+	result := "Logs:\n"
```
MDrakos, line 584: nit: For some reason, this line appears to be indented in the test failures. It's not immediately clear to me why so this might be more trouble to "fix" than it's worth.

MDrakos (approved):
Nice improvement!

PR #2341, opened 2023-02-01
Fix state-svc exiting prematurely

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1544" title="DX-1544" target="_blank">BugDX-1544</a>  Executors are failing to make svc connection
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/internal/resolver/resolver.go
+++ b/cmd/state-svc/internal/resolver/resolver.go
@@ -2,15 +2,17 @@ package resolver
 
 import (
 	"encoding/json"
-	"github.com/ActiveState/cli/cmd/state-svc/internal/rtusage"
+	"runtime/debug"
 	"sort"
 	"time"
 
 	"github.com/ActiveState/cli/cmd/state-svc/internal/deprecation"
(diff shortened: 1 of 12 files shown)


cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -88,6 +90,8 @@ func (r *Resolver) Close() error {
 func (r *Resolver) Query() genserver.QueryResolver { return r }
 
 func (r *Resolver) Version(ctx context.Context) (*graph.Version, error) {
+	defer handlePanics(recover(), debug.Stack())
```
daved, line 93: There's a good chance that this will not work as expected. The context in which `recover` and `debug.Stack` are called is the moment that the defer is called. This means that it should be `nil`. Please consider: https://go.dev/play/p/dfuoRApHgeC - The panic value is not handled and halts the program abruptly.

Please also consider: https://go.dev/play/p/nic95m6QXbT - This shows that an anonymous function permits the function being checked within the defer statement to have the expected context of "at the end of the function".

daved (changes requested):

daved (approved):

PR #2343, opened 2023-02-02
Don't filter out noop artifacts too early; we still need to build them

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1551" title="DX-1551" target="_blank">BugDX-1551</a>  INSTALL: Package install failed with `Failed to download artifact unknown (4088b386-0a8f-539e-9ea1-5ffa6dd15702)`
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -22,7 +22,6 @@ import (
 	"github.com/ActiveState/cli/internal/rtutils/p"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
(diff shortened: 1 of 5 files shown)


pkg/platform/runtime/runtime.go
```diff
@@ -250,16 +249,6 @@ func (r *Runtime) ExecutableDirs() (envdef.ExecutablePaths, error) {
 	return env.ExecutableDirs()
 }
 
-// Artifacts returns a map of artifact information extracted from the recipe
-func (r *Runtime) Artifacts() (map[artifact.ArtifactID]artifact.ArtifactRecipe, error) {
```
Naatan, line 254: Wasn't being used.

MDrakos (approved):

PR #2344, opened 2023-02-06
Fixed panic for remote project listings.

mitchell-as:


MDrakos (commented):
@mitchell-as I'm confused about how this fixes the panic in the story. How is the projects runner being triggered when you run `state switch`? This looks like a valid panic fix but I'm unsure if it fixes the one in the story.

PR #2347, opened 2023-02-06
Ensure download progress is counted on cached artifacts

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1550" title="DX-1550" target="_blank">BugDX-1550</a>  INSTALL: Package installation failed with `Received events that appear to be out of order. `
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/runtime/progress/progress.go
+++ b/internal/runbits/runtime/progress/progress.go
@@ -344,12 +344,9 @@ Progress bars status:
 Still expecting:
  - Builds: %v
  - Downloads: %v
- - Installs: %v
-Event log:
-%s`,
+ - Installs: %v`,
 				strings.Join(debugMsg, "\n"),
 				p.buildsExpected, p.downloadsExpected, p.installsExpected,
-				strings.Join(p.dbgEventLog, " > "),
 			)
 
 			if pending > 0 {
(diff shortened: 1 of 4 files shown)


internal/runbits/runtime/progress/progress.go
```diff
@@ -344,12 +344,9 @@ Progress bars status:
 Still expecting:
  - Builds: %v
  - Downloads: %v
- - Installs: %v
-Event log:
-%s`,
```
Naatan, line 349: Log was being flooded, and the values weren't helpful.

test/integration/checkout_int_test.go
```diff
@@ -24,25 +24,42 @@ type CheckoutIntegrationTestSuite struct {
 func (suite *CheckoutIntegrationTestSuite) TestCheckout() {
 	suite.OnlyRunForTags(tagsuite.Checkout)
 
-	ts := e2e.New(suite.T(), false)
+	ts := e2e.New(suite.T(), true)
```
MDrakos, line 27: Can this be reverted back to `false`?

internal/runbits/runtime/progress/progress.go
MDrakos: Would we want some sort of environment variable or other configuration option to get these logs?

MDrakos (dismissed):

MDrakos (approved):

PR #2349, opened 2023-02-06
Add env vars that allow tweaking the soft limit notifications

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1566" title="DX-1566" target="_blank">TaskDX-1566</a>  Make testing of soft limit toast notification more robust
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/rtusage/rtusage.go
+++ b/internal/runbits/rtusage/rtusage.go
@@ -53,15 +53,6 @@ func NotifyRuntimeUsage(cfg *config.Instance, data dataHandler, orgName string) 
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


internal/runbits/rtusage/rtusage.go
```diff
@@ -53,15 +53,6 @@ func NotifyRuntimeUsage(cfg *config.Instance, data dataHandler, orgName string)
 		return
 	}
 
-	if time.Now().Sub(cfg.GetTime(CfgKeyLastNotify)).Minutes() < float64(60) {
```
Naatan, line 56: Note I moved this down so that it only checks and records the timestamp if there is an actual notification to be sent.

daved (approved):

PR #2352, opened 2023-02-07
Filter out artifact that can't be build

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1550" title="DX-1550" target="_blank">BugDX-1550</a>  INSTALL: Package installation failed with `Received events that appear to be out of order. `
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/vcs.go
+++ b/pkg/platform/model/vcs.go
@@ -71,6 +71,9 @@ const (
 	// NamespacePackageMatch is the namespace used for language package requirements
 	NamespacePackageMatch = `^language\/(\w+)$`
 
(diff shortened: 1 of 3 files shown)


MDrakos (approved):
This makes sense but I wonder what effect this will have if a build artifact fails. In that case, the build should fail but it looks like with this PR we wouldn't handle that event.

Naatan:
> This makes sense but I wonder what effect this will have if a build artifact fails. In that case, the build should fail but it looks like with this PR we wouldn't handle that event.

It wouldn't fail to build on those artifacts, as that namespace basically means "this isn't a real artifact, it's not going to send build events as it won't ever be built".

PR #2358, opened 2023-02-08
Update fixVersion when opening PR

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1518" title="DX-1518" target="_blank">TaskDX-1518</a>  When a PR is opened we update fixVersion 
  </td></table>
  <br />

### DIFF
--- a/scripts/internal/workflow-helpers/jira.go
+++ b/scripts/internal/workflow-helpers/jira.go
@@ -1,6 +1,7 @@
 package workflow_helpers
 
 import (
+	"io/ioutil"
 	"os"
 	"regexp"
 	"sort"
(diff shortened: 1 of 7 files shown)


scripts/internal/workflow-helpers/jira.go
```diff
@@ -97,53 +103,53 @@ func FetchAvailableVersions(jiraClient *jira.Client) ([]semver.Version, error) {
...
+		return result[i].LT(result[j].Version)
 	})
 
 	return result, nil
 }
 
 var VersionMaster = semver.MustParse("0.0.0")
 
-func ParseTargetFixVersion(issue *jira.Issue, availableVersions []semver.Version) (semver.Version, *jira.FixVersion, error) {
+func ParseTargetFixVersion(issue *jira.Issue, availableVersions []Version) (target Version, original *jira.FixVersion, err error) {
```
mitchell-as, line 118: The return value names do not appear to be used. Perhaps you can remove them?

mitchell-as (changes requested):
One minor quibble. Please correct me if I'm wrong. Otherwise this looks good to me.

scripts/internal/workflow-helpers/jira.go
mitchell-as: I meant the named return values `target`, `original`, and `err`, not the actual return values. You are not setting `target`, `original`, or `err`, but returning actual values everywhere in this function. Am I misunderstanding named return values in Go?

mitchell-as (changes requested):
See my response to your response.

scripts/internal/workflow-helpers/jira.go
Naatan: Oh I see. I named it so that the API is self-explanatory. I found that without the naming I was interpreting the second argument as being the same as the first but wrapped in jira.FixVersion, whereas in actuality it's the jira fixVersion that it used to calculate the returned version from.

So the point of naming in this context is documentation. I'm not looking to use those variables functionally.

Naatan:
See my response to your response to my response. 😅

mitchell-as (approved):

PR #2359, opened 2023-02-09
Fix logging race condition and rotation errors

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1578" title="DX-1578" target="_blank">BugDX-1578</a>  Logging race condition
  </td></table>
  <br />

### DIFF
--- a/internal/logging/rotate.go
+++ b/internal/logging/rotate.go
@@ -28,7 +28,8 @@ func rotateLogs(files []fs.FileInfo, timeCutoff time.Time, amountCutoff int) []f
 	for prefix := range prefixes {
 		c := 0
 		for _, file := range files {
-			if strings.HasPrefix(file.Name(), prefix) && strings.HasSuffix(file.Name(), FileNameSuffix) {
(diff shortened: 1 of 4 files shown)


internal/logging/rotate.go
```diff
@@ -28,7 +28,8 @@ func rotateLogs(files []fs.FileInfo, timeCutoff time.Time, amountCutoff int) []f
 	for prefix := range prefixes {
 		c := 0
 		for _, file := range files {
-			if strings.HasPrefix(file.Name(), prefix) && strings.HasSuffix(file.Name(), FileNameSuffix) {
+			currentPrefix := LogPrefixRx.FindString(filepath.Base(file.Name()))
```
MDrakos, line 31: nit: I'm pretty sure the `filepath.Base` call is unnecessary. The `FileInfo` interface doesn't expose the full path. I ran into this in another PR using the `fs.FileInfo` 🤔

MDrakos (dismissed):

MDrakos (approved):

PR #2361, opened 2023-02-09
Fixed error detection of Win32 API calls.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1579" title="DX-1579" target="_blank">BugDX-1579</a>  CLONE - CLONE - Spinner isn't cleared during runtime progress
  </td></table>
  <br />
 

I wasn't interpreting Win32 API call results properly, so this fix is needed in order to not hammer rollbar with nonexistent errors.


mitchell-as:
The Target & Verify automation is failing and I don't know why :(

Naatan (approved):

PR #2363, opened 2023-02-09
BuildPlanner integration

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1568" title="DX-1568" target="_blank">TaskDX-1568</a>  Finish and integrate buildplanner implementation
  </td></table>
  <br />


pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -69,6 +48,35 @@ func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
 	return res, nil
 }
 
+func buildTerminals(nodeID strfmt.UUID, lookup map[strfmt.UUID]interface{}, result *[]strfmt.UUID) {
+	targetArtifact, ok := lookup[nodeID].(*model.Artifact)
```
Naatan, line 52: You're doing two actions here that can fail:

* lookup[nodeID]
* lookup[nodeID].(*model.Artifact)

I think the conditional is only addressing the first, and if `lookup[nodeID]` does not exist it will probably panic.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -69,6 +48,35 @@ func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
 	return res, nil
 }
 
+func buildTerminals(nodeID strfmt.UUID, lookup map[strfmt.UUID]interface{}, result *[]strfmt.UUID) {
```
Naatan, line 51: Please add comment explaining what this does.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -69,6 +48,35 @@ func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
...
+		logging.Debug("NodeID %s does not resolve to an artifact", nodeID)
+		return
+	}
+
+	if model.IsStateToolArtifact(targetArtifact.MimeType) {
+		*result = append(*result, targetArtifact.NodeID)
+		return
+	}
+
+	step, ok := lookup[targetArtifact.GeneratedBy].(*model.Step)
```
Naatan, line 63: Two actions that can fail here as well.

Naatan (changes requested):

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: I don't believe it will panic if there is no entry in the map, the conversion will simply fail. See here: https://go.dev/play/p/I8GmYcNDzQo

Let me know if that addresses your concerns, if not I'm happy to add more checks.

Naatan: Nice, that does indeed address my concern, thanks for being thorough!

Naatan (approved):
Great work!

PR #2367, opened 2023-02-10
Handle panics in sighandler.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1470" title="DX-1470" target="_blank">BugDX-1470</a>  We're not catching panics happening in captain runners
  </td></table>
  <br />

### DIFF
--- a/internal/sighandler/awaiting.go
+++ b/internal/sighandler/awaiting.go
@@ -45,8 +45,8 @@ func (as *Awaiting) WaitForFunc(f func() error) error {
 	as.Resume()
 
 	go func() {
-		defer close(errCh)
 		defer func() { panics.HandlePanics(recover(), debug.Stack()) }()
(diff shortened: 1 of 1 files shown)


internal/sighandler/awaiting.go
```diff
@@ -43,6 +46,7 @@ func (as *Awaiting) WaitForFunc(f func() error) error {
 
 	go func() {
 		defer close(errCh)
+		defer func() { panics.HandlePanics(recover(), debug.Stack()) }()
```
daved, line 49: It might be better to place the panic handling handling first. I'm not certain, but I think it should handle a panic if errCh is already closed somehow.

daved (changes requested):

daved (approved):

PR #2368, opened 2023-02-10
Handle 403 Forbidden errors during authentication.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1480" title="DX-1480" target="_blank">BugDX-1480</a>  Rollbar: Authentication API returned response status code does not match any response statuses defined for this endpoint in the swagger spec (status 403)
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/authentication/auth.go
+++ b/pkg/platform/authentication/auth.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ActiveState/cli/internal/profile"
 	"github.com/ActiveState/cli/internal/rollbar"
 	"github.com/ActiveState/cli/internal/singleton/uniqid"
+	"github.com/ActiveState/cli/pkg/platform/api"
 	"github.com/ActiveState/cli/pkg/platform/api/mono"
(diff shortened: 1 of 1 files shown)


pkg/platform/authentication/auth.go
```diff
@@ -196,6 +197,9 @@ func (s *Auth) AuthenticateWithModel(credentials *mono_models.Credentials) error
 			if os.IsTimeout(err) {
 				return locale.NewInputError("err_api_auth_timeout", "Timed out waiting for authentication response. Please try again.")
 			}
+			if api.ErrorCode(err) == 403 {
+				return locale.NewInputError("err_auth_forbidden", "You are not allowed to login now. Please try again later.")
+			}
```
MDrakos, line 202: nit: Do we want to update this error message to provide more context to the user? If this isn't in the swagger spec then the mono API isn't returning it so it could be something to do with the user's network configuration. We also may want them to try to report it to the forums.

MDrakos (approved):

PR #2369, opened 2023-02-10
Fixed error message if `state import file` fails.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1503" title="DX-1503" target="_blank">BugDX-1503</a>  Error message is not actionable for state import 
  </td></table>
  <br />
 

Differentiate between I/O errors and Platform errors. That makes the message more actionable for users.

### DIFF
--- a/internal/runners/packages/import.go
+++ b/internal/runners/packages/import.go
@@ -122,7 +122,7 @@ func (i *Import) Run(params *ImportRunParams) error {
 
 	changeset, err := fetchImportChangeset(reqsimport.Init(), params.FileName, lang.Name)
 	if err != nil {
-		return locale.WrapError(err, "err_obtaining_change_request", "Could not process change set: {{.V0}}.", api.ErrorMessageFromPayload(err))
+		return err
 	}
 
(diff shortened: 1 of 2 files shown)


internal/runners/packages/import.go
```diff
@@ -122,7 +122,7 @@ func (i *Import) Run(params *ImportRunParams) error {
 
 	changeset, err := fetchImportChangeset(reqsimport.Init(), params.FileName, lang.Name)
 	if err != nil {
-		return locale.WrapError(err, "err_obtaining_change_request", "Could not process change set: {{.V0}}.", api.ErrorMessageFromPayload(err))
+		return err
```
MDrakos, line 125: We should still wrap this error. We can come up with a nicer error message than what was here before or use `errs.Wrap` if you don't want it to be included in the default state tool output.

MDrakos (changes requested):

internal/runners/packages/import.go
mitchell-as: That's a fair point. Thanks for reminding me about `errs.Wrap`. I did want to avoid showing this error to the user.

MDrakos (approved):

PR #2371, opened 2023-02-10
Check for authentication early in `state revert`.

mitchell-as:
Erroring out due to lack of authentication after the user answers prompts is backwards.

### DIFF
--- a/internal/runners/revert/revert.go
+++ b/internal/runners/revert/revert.go
@@ -69,6 +69,10 @@ func (r *Revert) Run(params *Params) error {
 	if r.project == nil {
 		return locale.NewInputError("err_no_project")
 	}
+	if !r.auth.Authenticated() {
+		// Authentication is required in order to make a revert commit. Fail now rather than later.
+		return locale.NewInputError("err_api_not_authenticated")
+	}
 	r.out.Notice(locale.Tl("operating_message", "", r.project.NamespaceString(), r.project.Dir()))
 	commitID := strfmt.UUID(params.CommitID)
 


internal/runners/revert/revert.go
```diff
@@ -72,6 +72,10 @@ func (r *Revert) Run(params *Params) error {
 	if !strfmt.IsUUID(params.CommitID) {
 		return locale.NewInputError("err_invalid_commit_id", "Invalid commit ID")
 	}
+	if !r.auth.Authenticated() {
+		// Authentication is required in order to make a revert commit. Fail now rather than later.
+		return locale.NewInputError("err_api_not_authenticated")
+  }
```
daved, line 78: This looks like it needs to be formatted. Also, I somewhat dispute the need for this. If the API changes and permits a revert commit without authentication, then this change will prevent that policy decision from reaching users without further intervention on our part. @Naatan, thoughts?

daved (changes requested):

internal/runners/revert/revert.go
Naatan: I agree we should not outright require it here when we in fact allow unauthenticated commits, though this has me wondering where the original error came from.

daved: @damienprystay Is there intention behind reverts requiring authentication while commits do not? If not, then we probably need to file a story with DB to remove that requirement on reverts.

damienprystay: Which API endpoint are you referring to when talking about reverting? The endpoint to calculate a revert commit from two commit IDs (https://platform.activestate.com/swagger-ui/#/Version%20Control/getRevertCommit) doesn't require authentication per se: you can call it anonymously, although doing so for a commit that's deemed to be "private" will result in a 403.

Or are you talking about the call to save the revert commit once calculated? That should be the same (https://platform.activestate.com/swagger-ui/#/Version%20Control/addCommit) endpoint as for creating a "normal" commit, so the auth requirements there should be identical.

I can hopefully be more specific if you can point me at the endpoint(s) in question and the specifics of the behaviour you're seeing.

daved: @damienprystay I think we should look through the control flow more to see if we have an auth check deeper in the stack before troubling you more, so please disregard for now. @mitchell-as Please take a peek, or I can do it if you are occupied.

mitchell-as: @damienprystay It's the latter: addCommit. I am not authenticated and attempting to add a revert commit to a public repository, and it's requesting authentication to do so.

damienprystay: "Requesting authentication" as in `addCommit` is returning a 401, as opposed to a 403 Forbidden? The endpoint allows both authed and unauthed access (per the rather opaque `security` item in the [swagger spec](https://github.com/ActiveState/TheHomeRepot/blob/587f49df3280c5996694f20ae420fd5c0ea13bb4/service/api/swagger/swagger.yaml#L2916)), so it shouldn't 401 on simply absent auth – the one way I can see a 401 occurring is if the request _does_ provide the Authorization header, but it's invalid in some way (the header's malformed, or the token's malformed or expired or references a nonexistent user).

If it _is_ a 403 that's coming back, there are a [few reasons](https://github.com/ActiveState/TheHomeRepot/blob/587f49df3280c5996694f20ae420fd5c0ea13bb4/service/api/vcs.go#L27) that can happen, in particular specifying a commit from a private project as the parent, or attempting to use a restricted requirement namespace.

Happy to help diagnose specifics given the details of a request that's failing.

mitchell-as: Thanks for the detailed information. After going through the debugger, I found that immediately prior to making the API call, the state tool asserts the client is authenticated, which is obviously false. This assertion was masked by a lot of plumbing that I only discovered with the help of a debugger. Sorry for the noise.

daved (approved):

PR #2378, opened 2023-02-13
Always refresh JWT before attempting to use it

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1594" title="DX-1594" target="_blank">BugDX-1594</a>  Errors in state-svc log file - `Could not verify JWT: JWTExpired`
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/internal/rtusage/rtusage.go
+++ b/cmd/state-svc/internal/rtusage/rtusage.go
@@ -47,8 +47,8 @@ func (c *Checker) Check(organizationName string) (*model.RuntimeUsage, error) {
 		return cached.(*model.RuntimeUsage), nil
 	}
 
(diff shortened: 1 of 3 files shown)


MDrakos (approved):
This makes sense, but will this change also ensure the JWT is refreshed when `AnalyticsEvent` is called in the resolver? That is the function that produced the error in the linked Jira ticket.

Naatan:
> This makes sense, but will this change also ensure the JWT is refreshed when `AnalyticsEvent` is called in the resolver? That is the function that produced the error in the linked Jira ticket.

Thanks for catching that!

Fwiw I don't like the current solution, but I'm trying to make minimal changes. Filed a followup ticket.

MDrakos (approved):
Looks good!

PR #2381, opened 2023-02-14
Check if paths were actually removed to show correct message.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1548" title="DX-1548" target="_blank">BugDX-1548</a>  CLEAN: `state clean config/uninstall` generates a success message even if failed to delete the `config.db` file.
  </td></table>
  <br />
 

Windows' rmdir and del commands do not set %ERRORLEVEL%. Instead, check if the directories still exist.

Also, rmdir and del are equivalent.

For more information, see https://stackoverflow.com/questions/11137702/rd-exits-with-errorlevel-set-to-0-on-error-when-deletion-fails-etc (note that RD is an alias for RMDIR)

### DIFF
--- a/internal/assets/contents/scripts/removePaths.bat
+++ b/internal/assets/contents/scripts/removePaths.bat
@@ -30,14 +30,20 @@ echo "Waiting for process %exe% with PID %pid% to end..." >> %logfile%
     )
 
 echo "Process %exe% has ended" >> %logfile%
+set success=true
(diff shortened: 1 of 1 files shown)


daved (dismissed):

mitchell-as:
@daved Please re-review when you have the chance.

daved (approved):

PR #2389, opened 2023-02-22
Handle bad command values in subshell activation

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1529" title="DX-1529" target="_blank">BugDX-1529</a>  `state shell` under bash on Windows fails
  </td></table>
  <br />

### DIFF
--- a/internal/subshell/bash/bash.go
+++ b/internal/subshell/bash/bash.go
@@ -176,7 +176,11 @@ func (v *SubShell) Activate(proj *project.Project, cfg sscommon.Configurable, ou
 		directEnv = sscommon.EnvSlice(v.env)
 	}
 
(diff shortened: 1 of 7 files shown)


daved:
@Naatan The messaging was not signed-off on. It shows as:
```
$ state activate
█ Creating a Virtual Environment for your Project's Packages

WARNING: The following scripts collide with existing commands and should be renamed: install, test.
 x Could not activate runtime environment
 x Could not activate a new subshell
 x Command '/usr/binge/bash' is not an executable
█ Need More Help?

 • Run → `state activate --help` for general help
 • Ensure that the targeted command is a valid executable
 • Checking environment vars like SHELL may help resolve this
```
Do we want testing for this? If so, should there be focus on the value of SHELL?

Naatan (approved):
Short and sweet!

PR #2391, opened 2023-02-23
Add trigger token for executor usage

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1588" title="DX-1588" target="_blank">StoryDX-1588</a>  Separate trigger for `state exec` and executors
  </td></table>
  <br />


daved:
@Naatan In the AC, the `state exec` command is to get a new token as "exec-cmd". However this would then not match with other commands like activate which do not use that suffix. I added a new token for executors simply names "executor". I'm happy to redo this to match AC more exactly if the odd "-cmd" suffix is desirable, or the new token can be renamed to be more distinct from the existing "exec" token.

Naatan (changes requested):
Please stick to the original AC's. I agree the naming inconsistency is awkward, but ultimately inconsequential and it's better than having DW have to handle runtime tracking using a different trigger for some version or date cutoff.

I'm always happy to chat on Slack if AC's are posing problems. Rather we have a quick chat and adjust the AC's before investing time on implementation.

pkg/platform/runtime/target/target.go
```diff
@@ -74,8 +76,15 @@ func (t Trigger) IndicatesUsage() bool {
...
+var (
+	triggerExecPrefix     = TriggerExec.String() + ": "
+	triggerExecutorPrefix = TriggerExecutor.String() + ": "
+)
+
 func (t Trigger) IsExecTrigger() bool {
-	return strings.HasPrefix(string(t), string(TriggerExec)+": ")
+	trigger := t.String()
+	return strings.HasPrefix(trigger, triggerExecutorPrefix) ||
+		strings.HasPrefix(trigger, triggerExecPrefix)
```
Naatan, line 87: This logic seems to only cover the `state exec` command.. so we shouldn't address executors here.

Naatan (changes requested):

Naatan (approved):

PR #2395, opened 2023-02-24
Add shorthand ASY handling

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1519" title="DX-1519" target="_blank">StoryDX-1519</a>  Shorthand notation for activestate.yaml
  </td></table>
  <br />


daved:
@Naatan 
## Caveats with the current implementation

### A little wordy of a solution, but rational within Go's type system
The type system requires a bit of juggling, and I'll elaborate if you'd like. If the approach is acceptable, I will comment the new "Fields" suffixed types so that it is clear why they exist.

### Line number in error messages may shift slightly
In order for the data processing to be handled by the exact rules as provided by the YAML library we use, and so that the error messages are coming from the same source, I chose to unmarshal, then marshal/unmarshal the data. This can create line counts that are slightly off since the user may have chose to use a single line array, but the default marshaler went with a multi-line array.
```
- arraySingleLine: [my array]
- arrayMultiLine:
  - my
  - array
```
Further, depending on the offensiveness of the bad value, the error message may seem wrong.
```
=== RUN   TestParseShorthandBadData/array_in_name
    nameval_test.go:64:
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: unmarshal errors:
          line 2: cannot unmarshal !!seq into string
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: unmarshal errors:
          line 3: cannot unmarshal !!seq into string
=== RUN   TestParseShorthandBadData/array_in_value
    nameval_test.go:64:
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: unmarshal errors:
          line 3: cannot unmarshal !!seq into string
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: unmarshal errors:
          line 2: cannot unmarshal !!seq into string
=== RUN   TestParseShorthandBadData/new_field_in_name
    nameval_test.go:64:
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: line 3: block sequence entries are not allowed in this context
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: unmarshal errors:
          line 3: cannot unmarshal !!seq into map[string]interface {}
=== RUN   TestParseShorthandBadData/new_field_in_value
    nameval_test.go:64:
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: line 4: block sequence entries are not allowed in this context
         Project file `junk/path` could not be parsed, the parser produced the following error: yaml: line 3: block sequence entries are not allowed in this context
```

## Alternative
If the current approach is not acceptable, we may have to develop our own unmarshaling mechanisms so that we can pre-parse the data. That would mean abiding by whatever YAML parsing rules there are, and (likely) digging into reflection to pull out the dynamic shorthand key-val entry that does not match with the fields defined in the structure. Apologies if I'm not seeing something more suitable and obvious.

pkg/projectfile/nameval.go
```diff
@@ -0,0 +1,63 @@
...
+	if len(data) == 1 {
+		ssData, err := sanitizeMap(data)
+		if err != nil {
+			return err
+		}
+
+		for k, v := range ssData {
+			nv.Name, nv.Value = k, v
+			break
+		}
```
Naatan, line 29: I think we can address the caveats by simply doing a regular unmarshal when we encounter a "name" field.

We could represent the default struct without a `UnmarshalYAML()` method, and then use composition to create another struct that has `UnmarshalYAML()`, when this one encounters "name" we just defer to the default marshal behaviour on the embedded struct.

Hit me up on Slack so we can zoom if I'm not being clear. Key point is I think we can still achieve the shortcut mechanic you want without sacrificing error reporting.

I do agree manually unmarshalling the actual data is something we want to avoid. We are effectively piggybacking off UnmarshalYAML to basically direct the unmarshalling mechanic one of two ways.

Naatan (changes requested):

Naatan (changes requested):
I don't see how the latest changes address my previous review. Let's chat on Zoom today when we both have some time, maybe we can do some pair programming to figure this out, or if nothing else give me the perspective to understand why this solution is appropriate.

pkg/projectfile/nameval.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (nv *NameVal) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	data := make(map[string]interface{})
+	if err := unmarshal(&data); err != nil {
+		return err
+	}
+
+	if len(data) == 1 {
+		ssData, err := sanitizeMap(data)
```
Naatan, line 19: Can we rename this to "ensureStringMap()" and add comments to the function explaining that the reason for this function is to preserve yaml errors.

pkg/projectfile/nameval.go
```diff
@@ -0,0 +1,53 @@
...
+	Value string `yaml:"value"`
+}
+
+func (nv *NameVal) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	data := make(map[string]interface{})
+	if err := unmarshal(&data); err != nil {
+		return err
+	}
+
+	if len(data) == 1 {
```
Naatan, line 18:
```suggestion
    // If we only have 1 map entry then we know we're dealing with a <name>: <value> field
	if len(data) == 1 {
```

Naatan (changes requested):

PR #2401, opened 2023-02-28
Uninstall on Windows should remove the same files as on macOS/Linux

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1614" title="DX-1614" target="_blank">BugDX-1614</a>  Reinstall is fail if previous version was uninstalled and all config/cache folders remain
  </td></table>
  <br />
 

The previous fix was erroneously tailored towards the integration tests. It was not deleting the installer exe and install root stub.

### DIFF
--- a/internal/runners/clean/run_win.go
+++ b/internal/runners/clean/run_win.go
@@ -111,7 +111,7 @@ func removeInstall(logFile string, params *UninstallParams, cfg *config.Instance
 	// This is because Windows often thinks the installation.InstallDirMarker and
 	// constants.StateInstallerCmd files are still in use.
 	branchDir := filepath.Dir(filepath.Dir(stateExec))
-	if condition.OnCI() && !params.All {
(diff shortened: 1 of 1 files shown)


mitchell-as:
It feels a bit icky, but I'm not sure how best to go about reconciling the differences between installation structure on Windows vs Linux/macOS. Let me know if you have any ideas. Otherwise, we can go with something like this.

mitchell-as:
Never mind. After talking to Nathan about this, I'm going to try something else. All three platforms use the same install structure for integration tests. Windows is doing something different.

MDrakos (changes requested):
This fixes our integration tests CI issues but could legitimately mess with users attempting to uninstall the state tool as part of their CI steps... can we use `condition.InTest` instead?

MDrakos (approved):

PR #2403, opened 2023-03-02
Expand autostart behavior on Linux

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1606" title="DX-1606" target="_blank">StoryDX-1606</a>  Autostart should always update existing entries
  </td></table>
  <br />

### DIFF
--- a/internal/osutils/autostart/autostart.go
+++ b/internal/osutils/autostart/autostart.go
@@ -15,7 +15,7 @@ func (a AppName) String() string {
 	return string(a)
 }
 
-type app struct {
+type App struct {
 	Name    string
 	Exec    string
 	Args    []string
@@ -39,8 +39,8 @@ type Configurable interface {
 	IsSet(string) bool
 }
 
-func New(name AppName, exec string, args []string, options Options, cfg Configurable) (*app, error) {
-	return &app{
+func New(name AppName, exec string, args []string, options Options, cfg Configurable) (*App, error) {
+	return &App{
 		Name:    name.String(),
 		Exec:    exec,
 		Args:    args,
@@ -49,10 +49,19 @@ func New(name AppName, exec string, args []string, options Options, cfg Configur
 	}, nil
 }
 
-func (a *app) Enable() error {
+func (a *App) Enable() error {
 	return a.enable()
 }
 
-func (a *app) Disable() error {
+func (a *App) Disable() error {
 	return a.disable()
+}
+
+// IsEnabled is provided for testing only.
+func (a *App) IsEnabled() (bool, error) {
+	return a.isEnabled()
+}
+
+func (a *App) InstallPath() (string, error) {
+	return a.installPath()
 }
--- a/internal/osutils/autostart/autostart_linux.go
+++ b/internal/osutils/autostart/autostart_linux.go
@@ -29,11 +29,11 @@ func (a *app) enable() error {
 		return nil
 	}
 
-	if a.onDesktop() {
-		// The user is installing while in a desktop environment. Install an autostart shortcut file.
-		return a.enableOnDesktop()
+	// Enable for both Desktop and Server. Only error out for Desktop if on Desktop.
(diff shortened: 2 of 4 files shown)


daved:
@Naatan The AC state "Desktop logic should fail gracefully if we’re not on a desktop environment", and it makes sense to me to similarly fail gracefully for the server logic to fail if not on a server environment, so I included that case. If the server logic should be a hard fail in all cases, I can revert that commit.

Naatan (changes requested):
I'm not seeing the "fail gracefully" behaviour here? eg. on a server environment where HOME is not set, `enableOnDesktop` would almost certainly error out.

Basically, we want to fail if we absolutely cannot configure any autostart behaviour.

daved:
@Naatan Sanity check, but, if it passes, feel free to do a full review. Also, I'm wondering if there should be a desktop-autostart removal test, and I'd like to call attention to the log and continue behavior in `*App.enable()`.

internal/osutils/autostart/autostart_linux.go
```diff
@@ -5,76 +5,32 @@ import (
...
-func (a *app) enable() error {
-	enabled, err := a.IsEnabled()
-	if err != nil {
-		return errs.Wrap(err, "Could not check if app autostart is enabled")
-	}
-	if enabled {
-		return nil
+func (a *App) enable() error {
+	if err := legacyDisableOnDesktop(a.options.LaunchFileName); err != nil {
+		logging.Error("Cannot properly disable autostart (desktop): %v", err)
```
Naatan, line 23: You can error out for real on this. We need to get more comfortable with throwing errors at users so we can actually fix them rather than quietly ignore them to have them bubble up in different ways.

internal/osutils/autostart/autostart_linux.go
```diff
@@ -5,76 +5,32 @@ import (
...
-	}
-
-	scutOpts := shortcut.SaveOpts{
-		Name:        a.Name,
-		GenericName: a.options.GenericName,
-		Comment:     a.options.Comment,
-		Keywords:    a.options.Keywords,
-		IconData:    iconData,
-		IconPath:    iconsPath,
+		return err
```
Naatan, line 28: Wrap

internal/osutils/autostart/autostart.go
```diff
@@ -49,10 +49,19 @@ func New(name AppName, exec string, args []string, options Options, cfg Configur
...
 	return a.enable()
 }
 
-func (a *app) Disable() error {
+func (a *App) Disable() error {
 	return a.disable()
 }
+
+// IsEnabled is provided for testing only.
+func (a *App) IsEnabled() (bool, error) {
```
Naatan, line 61: You're already violating the comment:

https://github.com/ActiveState/cli/pull/2403/files#diff-fe1d6fb736a8d64ee905da5161a57eae198ad600fe9e62e48390d0ad205dd88fR26

Why not just rename `isEnabled` to `IsEnabled`?

internal/osutils/autostart/autostart_linux.go
```diff
@@ -176,6 +91,11 @@ func (a *app) IsEnabled() (bool, error) {
 	return false, nil
 }
 
+func (a *App) installPath() (string, error) {
+	// do not report ~/.profile as installed, as it would be removed on uninstall.
```
Naatan, line 95: This feels like the wrong place to be addressing that logic. Please write this in a way that's explicit.

Naatan (changes requested):

internal/osutils/autostart/autostart.go
daved: This comment is for external callers only. It forwards to the isEnabled method because each OS has its own implementation. I've improved the linux isEnabled comment to emphasize this.

daved:
Closing in favor of #2456 - The autostart package was changed enough to make restarting the PR more expedient than resolving the merge conflicts.

PR #2407, opened 2023-03-06
Fixed localization messages

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1632" title="DX-1632" target="_blank">BugDX-1632</a>  Commit messages are empty/incorrect when installing via CLI
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -992,16 +992,16 @@ field_version:
   other: Version
 field_bitwidth:
   other: Bit Width
-commit_message_add_platform:
+commit_message_added_platform:
   other: "Added platform: {{.V0}} {{.V1}}bit {{.V2}}"
 commit_message_updated_platform:
   other: "Updated platform: {{.V0}} to {{.V1}}"
 commit_message_removed_platform:
   other: "Removed platform: {{.V0}} {{.V1}}bit {{.V2}}"
-commit_message_add_language:
+commit_message_added_language:
   other: "Added language: {{.V0}} {{.V1}}"
 commit_message_updated_language:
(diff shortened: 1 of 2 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -994,10 +994,14 @@ field_bitwidth:
...
 commit_message_add_platform:
   other: "Added platform: {{.V0}} {{.V1}}bit {{.V2}}"
+commit_message_updated_platform:
+  other: "Updated platform: {{.V0}} to {{.V1}}"
 commit_message_removed_platform:
   other: "Removed platform: {{.V0}} {{.V1}}bit {{.V2}}"
 commit_message_add_language:
   other: "Added language: {{.V0}} {{.V1}}"
+commit_message_updated_language:
+  other: "Updated language: {{.V0}} {{.V1}}"
```
MDrakos, line 1004:
```suggestion
  other: "Updated language: {{.V0}} to {{.V1}}"
```
Just to keep this in line with the other `updated` message

pkg/platform/model/vcs.go
```diff
@@ -717,11 +717,11 @@ func languageCommitMessage(op Operation, name, version string) string {
 	var msgL10nKey string
 	switch op {
 	case OperationAdded:
-		msgL10nKey = locale.T("commit_message_add_language")
+		msgL10nKey = "commit_message_add_language"
```
MDrakos, line 720: nit:
```suggestion
		msgL10nKey = "commit_message_added_language"
```
To keep these all in the same tense. The same goes for `packages` and `platforms`

MDrakos (changes requested):

PR #2408, opened 2023-03-06
Moved initial runtime heartbeat event into the service.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1590" title="DX-1590" target="_blank">TaskDX-1590</a>  Initial runtime use event should be sent via svc
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/internal/rtwatcher/watcher.go
+++ b/cmd/state-svc/internal/rtwatcher/watcher.go
@@ -121,4 +121,5 @@ func (w *Watcher) Watch(pid int, exec string, dims *dimensions.Values) {
 	dims.Sequence = p.IntP(-1) // sequence is meaningless for heartbeat events
 	e := entry{pid, exec, dims}
 	w.watching = append(w.watching, e)
+	go w.RecordUsage(e) // initial event
 }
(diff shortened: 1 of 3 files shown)


cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -121,4 +121,5 @@ func (w *Watcher) Watch(pid int, exec string, dims *dimensions.Values) {
 	dims.Sequence = p.IntP(-1) // sequence is meaningless for heartbeat events
 	e := entry{pid, exec, dims}
 	w.watching = append(w.watching, e)
+	go w.RecordUsage(e) // initial event
```
mitchell-as, line 124: Instead of emitting the analytics event in the location mentioned by the ticket: https://sourcegraph.com/github.com/ActiveState/cli@d8d2fa6c465c7b8e8456e02ca1bcdd2b1cc77423/-/blob/cmd/state-svc/internal/resolver/resolver.go?L197:20&popover=pinned, emit it from the function called by that function: https://sourcegraph.com/github.com/ActiveState/cli@d8d2fa6c465c7b8e8456e02ca1bcdd2b1cc77423/-/blob/cmd/state-svc/internal/resolver/resolver.go?L206

Naatan (approved):

PR #2409, opened 2023-03-06
Add bad command arg value err messaging

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1577" title="DX-1577" target="_blank">BugDX-1577</a>  Wrong error message when used argument is not exist
  </td></table>
  <br />


daved:
@Naatan I'm not fully convinced of the merit of this bug. If resolving this does seem reasonable, would you like some unit tests to verify that the correct arg will be shown to users as the offending arg?

daved:
Abandoned in favor of a future refactor of the captain pkg.

PR #2417, opened 2023-03-07
Added `state refresh` command.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1626" title="DX-1626" target="_blank">StoryDX-1626</a>  We have a command like `state sync` or `state refresh` that only does the runtime installation step and nothing else
  </td></table>
  <br />


daved (approved):

daved:
@mitchell-as Nit/suggestion: If there is any way to diff the directory contents really easily, it would be nice to assert that in the test.

PR #2418, opened 2023-03-08
Log autoupdate failures instead of halting execution.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1640" title="DX-1640" target="_blank">BugDX-1640</a>  Auto update should not interrupt state tool
  </td></table>
  <br />

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -232,8 +232,10 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 
 	if childCmd != nil && !childCmd.SkipChecks() {
 		// Auto update to latest state tool version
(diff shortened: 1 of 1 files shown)


MDrakos (changes requested):
I don't believe the `autoUpdate` function will notify the user that the auto-update failed. Rather than this failing silently maybe we should provide a message to the user that their auto-update failed and remind them that they can run `state update` to try again. You may want to contact Pete and/or Nathan for copy.

Naatan (approved):

mitchell-as:
@MDrakos I got an OK from Nathan and GitHub won't let me merge without your approval.

MDrakos (approved):

PR #2422, opened 2023-03-09
Sort and group commands in docs more intuitively

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1658" title="DX-1658" target="_blank">BugDX-1658</a>  Command reference docs aren't sorted
  </td></table>
  <br />

### DIFF
--- a/internal/runners/export/docs/docs.go
+++ b/internal/runners/export/docs/docs.go
@@ -2,6 +2,7 @@ package docs
 
 import (
 	_ "embed"
+	"sort"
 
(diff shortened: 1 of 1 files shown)


MDrakos:
The export docs command relies on access to the state `Command` object. The instantiation of this object and the functions used to do it are all internal to `cmd/state` which scripts can't access. In order to move this out of a state tool command we would need to move those internal libraries so I've left the command where it is for now.

mitchell-as (approved):

PR #2428, opened 2023-03-13
Initial dotprogress indicator for non-interactive mode.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1449" title="DX-1449" target="_blank">StoryDX-1449</a>  Runtime progress has a non-interactive mode
  </td></table>
  <br />

### DIFF
--- a/cmd/state/output.go
+++ b/cmd/state/output.go
@@ -55,7 +55,7 @@ func initOutput(flags outputFlags, formatName string, shellName string) (output.
 		OutWriter:   os.Stdout,
 		ErrWriter:   os.Stderr,
 		Colored:     !flags.DisableColor(),
-		Interactive: term.IsTerminal(int(os.Stdin.Fd())),
+		Interactive: !flags.NonInteractive && term.IsTerminal(int(os.Stdin.Fd())),
 		ShellName:   shellName,
 	})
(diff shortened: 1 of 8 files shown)


cmd/state/output.go
```diff
@@ -55,7 +55,7 @@ func initOutput(flags outputFlags, formatName string, shellName string) (output.
 		OutWriter:   os.Stdout,
 		ErrWriter:   os.Stderr,
 		Colored:     !flags.DisableColor(),
-		Interactive: term.IsTerminal(int(os.Stdin.Fd())),
+		Interactive: !flags.NonInteractive && term.IsTerminal(int(os.Stdin.Fd())),
```
daved, line 58: I'd like to verify that this conditional is consciously different than the "isInteractive" variable in the main.go file. Normally it is nice to calculate this value once, then pass it around to all interested locations.

daved (approved):

cmd/state/output.go
mitchell-as: I've created a ticket for this question: https://activestatef.atlassian.net/browse/DX-1667

Naatan (approved):
Sweet and simple!

PR #2429, opened 2023-03-13
Update to Go 1.20

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-668" title="DX-668" target="_blank">TaskDX-668</a>  Update Go version
  </td></table>
  <br />


daved:
I've added a story to look into the need for the timeout increases: https://activestatef.atlassian.net/browse/DX-1671

activestate.yaml
```diff
@@ -47,14 +47,35 @@ constants:
...
-      go install github.com/ActiveState/gozip/cmd/gozip@e455986
-      popd
-      $scripts.install-deps-os
+      oldGOFLAGS="${GOFLAGS}"
+      # strip mod flag in CI only
+      if [ -z ${CI+x} ]; then
+        export GOFLAGS="$(echo ${GOFLAGS} | sed 's#-mod=vendor \?##')"
+      fi
+      if ! type "go" &> /dev/null; then
+        echo "go is not installed. Please install Go version 1.11 or above."
```
Naatan, line 58:
```suggestion
        echo "go is not installed. Please install Go version 1.20 or above."
```

activestate.yaml
```diff
@@ -47,14 +47,35 @@ constants:
...
-      GOFLAGS="${goflags}"
-      pushd ~
-      go install github.com/ActiveState/gozip/cmd/gozip@e455986
-      popd
-      $scripts.install-deps-os
+      oldGOFLAGS="${GOFLAGS}"
+      # strip mod flag in CI only
+      if [ -z ${CI+x} ]; then
+        export GOFLAGS="$(echo ${GOFLAGS} | sed 's#-mod=vendor \?##')"
+      fi
```
Naatan, line 56: Assuming this is necessary for Go 1.20; please add a comment explaining why we're doing this.

Naatan (changes requested):

activestate.yaml
daved: oof

activestate.yaml
daved: I'm not exactly sure what change occurred to require this, but it is a known issue: https://github.com/golang/go/issues/45811

IIUC, we can vendor the source needed to build these executables, then ensure that the vendored source is used by setting the mod flag to =vendor.

I've created a new story to look into our install-deps setup: https://activestatef.atlassian.net/browse/DX-1673

Naatan: Can't we just drop -mod=vendor? I thought the recent Go versions use the vendor dir if it exists, no need for the flag?

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2429#discussion_r1140495561

daved:
@Naatan Since the last review, the changes have been to drop mod=vendor, and the constants-generator script is now silent except when passed -v (which is always the case in CI).

The silent output was for two reasons: 1. Time from start was added to the logging output and was a little busy looking 2. It's not really necessary for normal development use (i.e. `state run preprocess` output isn't usually helpful, for myself anyway)

scripts/constants-generator/main.go
```diff
@@ -2,53 +2,80 @@ package main
...
+	d := time.Since(start)
+	fmt.Fprintf(os.Stderr, "[%08dμs] ", d.Microseconds())
+	fmt.Fprintf(os.Stderr, format, as...)
+	fmt.Fprintln(os.Stderr)
+}
+
+func log(format string, as ...any) {
+	if !verbose {
+		return
+	}
```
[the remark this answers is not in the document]
daved: It wasn't for performance. The info that was added was to help debug, but it's also busy looking. And, I rarely ever read the output. By the time I'm interested in some of the output, `state --version` meets the need.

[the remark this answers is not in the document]
daved: The `-v` flag will now add the time-from-start content, and `-s` will silence the output. So, `state run preprocess -s` will be have minimal output.

Naatan (approved):

PR #2431, opened 2023-03-14
Add `--to` option for `state revert`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1506" title="DX-1506" target="_blank">StoryDX-1506</a>  State revert has a `--to` flag to get back to the state of the target commit
  </td></table>
  <br />
 

Allow reverting to a given commit, like in the Dashboard.

### DIFF
--- a/internal/runners/revert/revert.go
+++ b/internal/runners/revert/revert.go
@@ -73,15 +73,15 @@ func (r *Revert) Run(params *Params) error {
 	if !strfmt.IsUUID(params.CommitID) {
 		return locale.NewInputError("err_invalid_commit_id", "Invalid commit ID")
 	}
-	if params.CommitID == r.project.CommitID() && params.To {
-		return locale.NewInputError("err_revert_to_current_commit", "Cannot revert to current commit")
+	latestCommit := r.project.CommitUUID()
+	if params.CommitID == latestCommit.String() && params.To {
+		return locale.NewInputError("err_revert_to_current_commit", "The commit to revert to cannot be the latest commit")
 	}
 	r.out.Notice(locale.Tl("operating_message", "", r.project.NamespaceString(), r.project.Dir()))
 	commitID := strfmt.UUID(params.CommitID)
 
-	var targetCommit *mono_models.Commit
+	var targetCommit *mono_models.Commit // the commit to revert the contents of, or the commit to revert to
 	var fromCommit, toCommit strfmt.UUID
-	latestCommit := r.project.CommitUUID()
 	if !params.To {
 		priorCommits, err := model.CommitHistoryPaged(commitID, 0, 2)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/platform/model/vcs.go
```diff
@@ -989,7 +997,11 @@ func GetCommit(commitID strfmt.UUID) (*mono_models.Commit, error) {
 	params.SetCommitID(commitID)
 	params.SetHTTPClient(retryhttp.DefaultClient.StandardClient())
 
-	res, err := authentication.Client().VersionControl.GetCommit(params, authentication.ClientAuth())
+	client := mono.New()
+	if authentication.LegacyGet().Authenticated() {
```
mitchell-as, line 1001: This is not ideal, but necessary in order to avoid authentication being required to perform a revert. This `LegacyGet()` is used elsewhere throughout this file in similar places. It just was never added here when it should have been.

I could get rid of it by passing `*authentication.Auth` pointers to this function, but that would affect other unrelated callers like *internal/runners/pull/pull.go*. We have a ticket to refactor out `LegacyGet()`.

internal/runners/revert/revert.go
```diff
@@ -72,28 +73,50 @@ func (r *Revert) Run(params *Params) error {
 	if !strfmt.IsUUID(params.CommitID) {
 		return locale.NewInputError("err_invalid_commit_id", "Invalid commit ID")
 	}
+	if params.CommitID == r.project.CommitID() && params.To {
+		return locale.NewInputError("err_revert_to_current_commit", "Cannot revert to current commit")
```
MDrakos, line 77: I think this error need a bit more context

internal/runners/revert/revert.go
```diff
@@ -72,28 +73,50 @@ func (r *Revert) Run(params *Params) error {
...
 	r.out.Notice(locale.Tl("operating_message", "", r.project.NamespaceString(), r.project.Dir()))
 	commitID := strfmt.UUID(params.CommitID)
 
-	priorCommits, err := model.CommitHistoryPaged(commitID, 0, 2)
-	if err != nil {
-		return locale.WrapError(err, "err_revert_get_commit", "Could not fetch commit details for commit with ID: {{.V0}}", params.CommitID)
-	}
-	if priorCommits.TotalCommits < 2 {
-		return locale.NewInputError("err_revert_no_history", "Cannot revert commit {{.V0}}: no prior history", params.CommitID)
+	var targetCommit *mono_models.Commit
```
MDrakos, line 82: I think I need some help understanding what the `targetCommit` represents in this code. We have a `from` and `to` commit so what is the target commit? It seems to only be used for sourcing author information and printing, except in the `to` case where it becomes the `toCommit`...

MDrakos (changes requested):
Looks good to me! Just one minor note and a question.

internal/runners/revert/revert.go
mitchell-as: `targetCommit` is either the commit whose contents are being reverted, or the commit being reverted to. It depends on the context of course. Yes, it is very confusing :/

MDrakos (approved):

PR #2434, opened 2023-03-15
Defer `state pull --set-project` from changing namespace until merge succeeds.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1624" title="DX-1624" target="_blank">BugDX-1624</a>  PULL: The link in the YAML file was modified despite the command’s fails
  </td></table>
  <br />


mitchell-as:
That's a good idea. Thanks for the suggestion. I was under the impression `state pull` needed an updated activestate.yaml to work.

MDrakos (approved):

PR #2440, opened 2023-03-20
Enable PYTHONUTF8 in Git Bash on Windows for default Python projects.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1633" title="DX-1633" target="_blank">BugDX-1633</a>  Windows `git bash` project activation trigger error
  </td></table>
  <br />
 

Python cannot detect a UTF-8 environment within a Git Bash terminal, so the UTF-8 characters in the activation message cannot be printed by Python. Provide a hint.

We can either set an environment variable, or modify the actual Python script to add:

```
import sys
sys.stdout.reconfigure(encoding='utf-8')
```

I chose the environment variable because it's less invasive.

### DIFF
--- a/internal/assets/contents/activestate.yaml.python.tpl
+++ b/internal/assets/contents/activestate.yaml.python.tpl
@@ -2,13 +2,12 @@ scripts:
   - name: activationMessage
     language: {{.Language}}
     value: |
-      # -*- coding: utf-8 -*-
       import textwrap
       print(textwrap.dedent("""
         Quick Start
-        ───────────
-        • To add a package to your runtime, type "state install <package name>"
-        • Learn more about how to use the State Tool, type "state learn"
+        -----------
+        * To add a package to your runtime, type "state install <package name>"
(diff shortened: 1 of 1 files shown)


daved (dismissed):

Naatan (changes requested):
We discussed this during stand and agreed that a more appropriate solution would be to try and remove the UTF8 dependency from the script.

The perspective there being that this yaml code is meant to be an onboarding experience. And so us having additional chunks of code in there to work around bugs or facilitate advanced use-cases would detract from that experience.

Naatan (approved):

Naatan (changes requested):
Oh.. I think we can drop this? 

https://github.com/ActiveState/cli/pull/2440/files#diff-d578f534a6a7c90608b33507b2fe2531d46837a66000f77e834781b1f6da3088L5

mitchell-as:
Yep, you're right.

PR #2445, opened 2023-03-21
Send analytics event on config set

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1649" title="DX-1649" target="_blank">StoryDX-1649</a>  As a PM I can see what config keys users are changing
  </td></table>
  <br />

### DIFF
--- a/internal/runners/config/set_test.go
+++ b/internal/runners/config/set_test.go
@@ -3,6 +3,7 @@ package config
 import (
 	"testing"
 
+	"github.com/ActiveState/cli/internal/analytics/client/blackhole"
 	"github.com/ActiveState/cli/internal/config"
 	configMediator "github.com/ActiveState/cli/internal/mediators/config"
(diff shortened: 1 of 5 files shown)


internal/runners/config/set_test.go
```diff
@@ -15,7 +16,7 @@ func TestSetUnknownKey(t *testing.T) {
 	cfg.Set("unknown", nil)
 
 	outputer := outputhelper.NewCatcher()
-	set := Set{outputer, cfg, nil}
+	set := Set{outputer, cfg, nil, blackhole.New()}
```
mitchell-as, line 19: I would imagine we want some sort of test that verifies the correct analytics "Set" and "Unset" events are emitted, right? Or is it out of scope?

mitchell-as (changes requested):

mitchell-as (approved):

PR #2450, opened 2023-03-22
Fix failing integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1687" title="DX-1687" target="_blank">BugDX-1687</a>  "Expected '{dir}' to be a valid directory" failure message across multiple integration tests
  </td></table>
  <br />

### DIFF
--- a/internal/app/app_darwin.go
+++ b/internal/app/app_darwin.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/ActiveState/cli/internal/assets"
+	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/installation"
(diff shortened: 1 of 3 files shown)


daved (approved):
If this can be improved by using a directory within the e2e tmp dirs, and then setting some envvar for the relevant logic to use, I suggest going that route in the future (new story).

internal/app/app_darwin.go
```diff
@@ -52,9 +52,12 @@ func (a *App) install() error {
 		return errs.Wrap(err, "Could not create info file")
 	}
 
-	installDir, err := installation.ApplicationInstallPath()
-	if err != nil {
-		return errs.Wrap(err, "Could not get installation path")
+	installDir := os.Getenv("ACTIVESTATE_CLI_APPINSTALLDIR_OVERRIDE")
```
daved, line 55: I think using the added constant is appropriate here and other places where this magic value exists.

daved (approved):

PR #2451, opened 2023-03-22
Define our own global `--help` flag.

mitchell-as:
Don't let Cobra auto-define its own for us.

### DIFF
--- a/cmd/state/internal/cmdtree/cmdtree.go
+++ b/cmd/state/internal/cmdtree/cmdtree.go
@@ -230,6 +230,7 @@ func newGlobalOptions() *globalOptions {
 
 func newStateCommand(globals *globalOptions, prime *primer.Values) *captain.Command {
 	opts := state.NewOptions()
+	var help bool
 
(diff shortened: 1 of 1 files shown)


cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -290,6 +291,13 @@ func newStateCommand(globals *globalOptions, prime *primer.Values) *captain.Comm
 				Description: locale.T("flag_state_version_description"),
 				Value:       &opts.Version,
 			},
+			{
+				Name:        "help",
+				Description: locale.Tl("flag_help", "Help for this command"),
+				Shorthand:   "h",
+				Persist:     true,
```
daved, line 298: Will this "override" the output of the help descriptions/messages of subcommands?

mitchell-as: Yes, this is a single, universal message for all of our commands and subcommands.

daved (approved):

PR #2455, opened 2023-03-23
Windows uninstaller should allow the user to confirm exit before clos…

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1689" title="DX-1689" target="_blank">BugDX-1689</a>  Uninstall pop-up window closed without providing any info to the user or the path to the uninstall log.
  </td></table>
  <br />
 

…ing popup cmd prompt.

Otherwise the user cannot see its contents.

### DIFF
--- a/cmd/state/internal/cmdtree/clean.go
+++ b/cmd/state/internal/cmdtree/clean.go
@@ -42,6 +42,14 @@ func newCleanUninstallCommand(prime *primer.Values, globals *globalOptions) *cap
 				Description: locale.T("flag_state_clean_uninstall_force_description"),
 				Value:       &params.Force,
 			},
+			{
+				// This option is only used on Windows to prevent the popup cmd from closing before the
+				// user can take note of the printed log path.
+				Name:        "confirm-exit",
+				Description: "Prompts the user to press enter before exiting",
+				Hidden:      true, // this is not a user-facing flag
+				Value:       &params.ConfirmExit,
+			},
 		},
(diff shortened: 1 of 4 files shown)


cmd/state/internal/cmdtree/clean.go
```diff
@@ -42,6 +42,14 @@ func newCleanUninstallCommand(prime *primer.Values, globals *globalOptions) *cap
 				Description: locale.T("flag_state_clean_uninstall_force_description"),
 				Value:       &params.Force,
 			},
+			{
+				// This option is only used on Windows to prevent the popup cmd from closing before the
+				// user can take note of the printed log path.
+				Name:        "confirm-exit",
+				Description: "Prompts the user to press enter before exiting",
+				Hidden:      true, // this is not a user-facing flag
+				Value:       &params.ConfirmExit,
```
MDrakos, line 51: It doesn't appear that this value is being used anywhere.

MDrakos (changes requested):

cmd/state/internal/cmdtree/clean.go
mitchell-as: 🤦 Long day yesterday. Sorry about that...

MDrakos: No worries

MDrakos (approved):

PR #2456, opened 2023-03-24
Remove desktop-specific autostart logic for linux

daved:


internal/osutils/autostart/autostart_linux.go
```diff
@@ -171,16 +102,25 @@ func isEnabled(exec string, opts Options) (bool, error) {
 }
 
 func autostartPath(name string, _ Options) (string, error) {
-	dir, err := prependHomeDir(autostartDir)
+	// do not report ~/.profile as installed, as it would be removed on uninstall.
+	return "", nil
```
MDrakos, line 106: Just want to make sure we've verified that the `~/.profile` entries for autostart will be removed on uninstall.

MDrakos (approved):
Looks good to me. Just one thing to check on.

internal/osutils/autostart/autostart_linux.go
daved: I didn't check the current calls to this package, but under my last PR the uninstall process called "Disable" as well as removing any entry returned from this method (previously named "InstallPath"). The disable code for the .profile mod was untouched and is here: https://github.com/ActiveState/cli/pull/2456/files#diff-fe1d6fb736a8d64ee905da5161a57eae198ad600fe9e62e48390d0ad205dd88fR68-R81

internal/osutils/autostart/autostart_linux.go
```diff
@@ -5,76 +5,31 @@ import (
...
 
 const (
-	autostartDir  = ".config/autostart"
 	autostartFile = ".profile"
 )
 
 func enable(exec string, opts Options) error {
+	if err := legacyDisableOnDesktop(exec, opts); err != nil {
+		return errs.Wrap(err, "Could not properly disable autostart (desktop): %v", err)
+	}
```
Naatan, line 23: This feels like the wrong place. Can invoke this logic through the installer instead? So it's removed at install time, and not during regular use.

internal/osutils/autostart/autostart_linux.go
Naatan: Please rewrite this logic so it's not implicit. Imo any time we're implicitly returning some value that embeds specialized knowledge of consuming code we should revisit what we're doing.

internal/osutils/autostart/autostart_linux.go
```diff
@@ -171,16 +102,25 @@ func isEnabled(exec string, opts Options) (bool, error) {
...
+// https://activestatef.atlassian.net/browse/DX-1677
+func legacyDisableOnDesktop(exec string, opts Options) error {
+	dir, err := prependHomeDir(".config/autostart")
 	if err != nil {
-		return "", errs.Wrap(err, "Could not find autostart directory")
+		return errs.Wrap(err, "Could not find autostart directory")
 	}
-	path := filepath.Join(dir, name)
+
+	path := filepath.Join(dir, exec)
```
Naatan, line 116: Are you sure this is the correct path? We were using `opts.LaunchFileName` before:

https://github.com/ActiveState/cli/pull/2456/files#diff-fe1d6fb736a8d64ee905da5161a57eae198ad600fe9e62e48390d0ad205dd88fL49

Naatan (changes requested):

internal/osutils/autostart/autostart_linux.go
daved: The recent rework has the disable logic using "exec":
https://github.com/ActiveState/cli/blob/086ad2976bdd8d7b01b8e03305030d561cf883da/internal/osutils/autostart/autostart_linux.go#L117
... which is passed as "name":
https://github.com/ActiveState/cli/blob/086ad2976bdd8d7b01b8e03305030d561cf883da/internal/osutils/autostart/autostart_linux.go#L178
In other words, "LaunchFileName" is not used for disabling. @MDrakos This feels like it might be a bug; Is this OK?

internal/osutils/autostart/autostart_linux.go
daved: We had a discussion which emphasized the need to ensure this method does not have any knowledge about how it will be called. I think I reached that intention by updating the comment. Please let me know if it's sufficient.

internal/osutils/autostart/autostart_linux.go
daved: This is a Linux-specific behavior and the placement allows it to not affect the outer cross-platform surface. If that is, in fact, desirable, each OS will need to have an implementation for a new function in autostart.go.

Alternately, if the relevant file exists, we can report it via autostartPath, but that feels even more janky in that it relies on uninstallation for cleanup to occur. Though, I may be misunderstanding the expected flow.

In most cases, the current approach will look for a file, not find it, then take no action - so it's low cost and should, in more cases, ensure the file system is as we intend.

This seems complex enough that we might want to have a quick VC (along with @MDrakos) so that everyone is on the same page with the logic.

internal/osutils/autostart/autostart_linux.go
MDrakos: `LaunchFileName` is the `state-svc.desktop` file which we should be removing. The `exec` argument that is being passed to `autoStartPath` is the name of the state service executable. This looks correct to me.

daved: For the legacy desktop uninstall, we do want to focus on the LaunchFileName. I've updated the logic and performed some other cleanup so that the primary functions read more simply, and all errors are now wrapped.

internal/osutils/autostart/autostart_linux.go
daved: There are legacy .profile mod cleanups and they fire in roughly the same places as the legacy desktop removals. The .profile mod cleanups have been moved into their own legacy function and are documented in the same Jira story as the legacy desktop removals.

Naatan: I'm not following why my requested change can't be addressed. My point is that this is effectively install/update logic, and should live in the installer and update code. This code already has platform specific logic, and even if not we can always introduce it. Go makes this dead simple so not sure I follow why this would require some larger refactor.

internal/osutils/autostart/autostart_linux.go
```diff
@@ -13,26 +13,26 @@ import (
...
+	profile, err := profilePath()
 	if err != nil {
-		return errs.Wrap(err, "Could not check if app autostart is already enabled")
+		return errs.Wrap(err, "Could not get profile path")
 	}
-	if enabled {
-		return nil
+
+	if err := legacyRemoveAutostartEntry(profile); err != nil {
+		return errs.Wrap(err, "Could not clean up legacy autostart entry (server)")
```
Naatan: https://github.com/ActiveState/cli/pull/2456#discussion_r1157445613

Naatan (changes requested):
I still feel like this logic is invoked from the wrong place. This is update logic and so it should be invoked from the installer. Having it mingled into the autostart code feels ripe for misinterpretation and regressions.

I realize this is building on what was there before; ie. there was already code in autostart that was effectively performing update logic. But just because we didn't catch it before doesn't mean we can't catch it now.

internal/osutils/autostart/autostart_linux.go
```diff
@@ -96,6 +81,23 @@ func autostartPath(name string, opts Options) (string, error) {
 	return "", nil
 }
 
+func upgrade(exec string, opts Options) error {
```
Naatan, line 84: It doesn't look like you're using `exec`?

Naatan (changes requested):

internal/osutils/autostart/autostart_linux.go
daved: Yeah, that's too defensive of a definition decision, on my part, for an internal package. The param is removed.

PR #2458, opened 2023-03-24
Add analytics events to update flow

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1663" title="DX-1663" target="_blank">TaskDX-1663</a>  As a Tech lead I can identify the success rate of auto updates
  </td></table>
  <br />

### DIFF
--- a/cmd/state/autoupdate.go
+++ b/cmd/state/autoupdate.go
@@ -27,6 +27,10 @@ import (
 	"github.com/thoas/go-funk"
 )
 
+type ErrStateExe struct{ *locale.LocalizedError }
+
+type ErrExecuteRelaunch struct{ *errs.WrapperError }
+
 const CfgKeyLastCheck = "auto_update_lastcheck"
 
 func init() {
@@ -101,9 +105,16 @@ func autoUpdate(args []string, cfg *config.Instance, an analytics.Dispatcher, ou
 
 	code, err := relaunch(args)
 	if err != nil {
+		var msg string
+		if errs.Matches(err, &ErrStateExe{}) {
+			msg = "Could not locate state executable for relaunch"
+		}
+		if errs.Matches(err, &ErrExecuteRelaunch{}) {
+			msg = "Could not execute relaunch"
+		}
 		an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateRelaunch, anaConst.AutoUpdateLabelFailed, &dimensions.Values{
 			Version: p.StrP(up.Version),
-			Error:   p.StrP("Failed to relaunch after update"),
+			Error:   p.StrP(msg),
 		})
 		return true, errs.Silence(errs.WrapExitCode(err, code))
 	}
@@ -191,12 +202,12 @@ func shouldRunAutoUpdate(args []string, cfg *config.Instance, an analytics.Dispa
 func relaunch(args []string) (int, error) {
 	exec, err := installation.StateExec()
 	if err != nil {
-		return -1, locale.WrapError(err, "err_state_exec")
+		return -1, &ErrStateExe{locale.WrapError(err, "err_state_exec")}
 	}
 
 	code, _, err := exeutils.ExecuteAndPipeStd(exec, args[1:], []string{fmt.Sprintf("%s=true", constants.ForwardedStateEnvVarName)})
 	if err != nil {
-		return code, errs.Wrap(err, "Forwarded command after auto-updating failed. Exit code: %d", code)
+		return code, &ErrExecuteRelaunch{errs.Wrap(err, "Forwarded command after auto-updating failed. Exit code: %d", code)}
 	}
 
 	return code, nil
--- a/cmd/state/autoupdate.go
+++ b/cmd/state/autoupdate.go
@@ -6,6 +6,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ActiveState/cli/internal/analytics"
+	anaConst "github.com/ActiveState/cli/internal/analytics/constants"
+	"github.com/ActiveState/cli/internal/analytics/dimensions"
 	"github.com/ActiveState/cli/internal/condition"
 	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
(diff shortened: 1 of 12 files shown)


cmd/state/autoupdate.go
```diff
@@ -19,17 +22,35 @@ import (
...
+	anaLabelForward         = "forward"
+	anaLabelUnitTest        = "unittest"
+	anaLabelConflict        = "conflict"
+	anaLabelDisabledEnv     = "disabled_env"
+	anaLabelDisabledConfig  = "disabled_config"
+	anaLabelCI              = "ci"
+	anaLabelFreshInstall    = "fresh-install"
+	anaLabelLocked          = "locked"
+	anaLabelTooFreq         = "too-frequent"
+	anaLabelSpecificVersion = "specific-version"
```
MDrakos, line 46: These aren't in the doc but I added them as there are error cases for them. See below where they are used.

internal/updater/fetcher.go
```diff
@@ -4,54 +4,78 @@ import (
...
 	isEmpty, err := fileutils.IsEmptyDir(targetDir)
 	if err != nil {
-		return errs.Wrap(err, "Could not verify if target dir is empty")
+		msg := "Could not verify if target dir is empty"
+		f.event(update.Version, msg)
+		return errs.Wrap(err, msg)
 	}
 	if !isEmpty {
-		return errs.Wrap(err, "Target dir is not empty: %s", targetDir)
+		msg := fmt.Sprintf("Target dir is not empty: %s", targetDir)
```
Naatan, line 57:
```suggestion
		msg := "Target dir is not empty"
```

internal/updater/fetcher.go
```diff
@@ -4,54 +4,78 @@ import (
...
-		return errs.Wrap(err, "Unarchiving failed")
+		msg := "Unarchiving failed"
+		f.event(update.Version, msg)
+		return errs.Wrap(err, msg)
 	}
 
 	return nil
 }
 
+func (f *Fetcher) event(version, msg string) {
```
Naatan, line 72:
```suggestion
func (f *Fetcher) analyticsEvent(version, msg string) {
```

internal/updater/updater.go
```diff
@@ -19,12 +22,21 @@ import (
...
 
-const CfgKeyInstallVersion = "state_tool_installer_version"
+const (
+	CfgKeyInstallVersion = "state_tool_installer_version"
+
+	// Ananlytics labels
+	anaLabelSuccess     = "success"
+	anaLabelFailed      = "failed"
+	anaLabelAvailable   = "available"
+	anaLabelUnavailable = "unavailable"
```
Naatan, line 38: Should be in the analytics constants package.

internal/updater/updater.go
```diff
@@ -184,3 +204,15 @@ func (u *AvailableUpdate) InstallWithProgress(installTargetPath string, progress
 
 	return proc, nil
 }
+
+func (u *AvailableUpdate) event(label, version, msg string) {
```
Naatan, line 208:
```suggestion
func (u *AvailableUpdate) analyticsEvent(label, version, msg string) {
```

internal/updater/updater.go
```diff
@@ -184,3 +204,15 @@ func (u *AvailableUpdate) InstallWithProgress(installTargetPath string, progress
...
+	dims := &dimensions.Values{
+		TargetVersion: p.StrP(version),
+	}
+
+	if msg != "" {
+		dims.Error = p.StrP(msg)
+	}
+
+	u.an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateDownload, label, dims)
```
Naatan, line 217: Not all events here are `ActUpdateDownload` I believe.

Naatan (changes requested):
In addition to the lines I commented on, I think it makes sense to also move the labels (the error messages) to the analytics constants package.

internal/updater/checker.go
```diff
@@ -119,22 +125,37 @@ func (u *Checker) GetUpdateInfo(desiredChannel, desiredVersion string) (*Availab
...
+		u.an.EventWithLabel(anaConst.CatConfig, anaConst.ActUpdateCheck, anaLabelFailed, &dimensions.Values{
+			Version: p.StrP(desiredVersion),
+			Error:   p.StrP("Could not fetch update info"),
+		})
 		return nil, errs.Wrap(err, "Could not fetch update info from %s", infoURL)
 	}
 
-	info := &AvailableUpdate{}
+	info := &AvailableUpdate{
+		an: u.an,
```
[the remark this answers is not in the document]
MDrakos: I agree it's awkward but I'm not sure any of the alternatives are better. We could pass the analytics instance to the `AvailableUpdate`'s methods or create a new constructor. Either way, since this object is responsible for downloading and installing the update I think we want the analytics in there.

cmd/state/autoupdate.go
```diff
@@ -91,43 +78,46 @@ func autoUpdate(args []string, cfg *config.Instance, an analytics.Dispatcher, ou
...
 	code, err := relaunch(args)
 	if err != nil {
-		an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateRelaunch, anaLabelFailed, &dimensions.Values{
+		var msg string
+		if errs.Matches(err, &ErrStateExe{}) {
+			msg = anaConst.UpdateErrorExecutable
+		}
+		if errs.Matches(err, &ErrExecuteRelaunch{}) {
+			msg = anaConst.UpdateErrorRelaunch
+		}
```
Naatan, line 112: Nit: should be `switch` or `else`.

internal/updater/updater.go
```diff
@@ -205,7 +197,11 @@ func (u *AvailableUpdate) InstallWithProgress(installTargetPath string, progress
 	return proc, nil
 }
 
-func (u *AvailableUpdate) event(label, version, msg string) {
+func (u *AvailableUpdate) analyticsEvent(action, label, version, msg string) {
+	if u.an == nil {
+		return
+	}
```
Naatan, line 203: I don't think this should ever be true? In which case if we were to encounter this we'd want it to error out, even panic.

Granted we should do nil checks but I feel like doing it at this level is excessive.

Naatan (changes requested):

cmd/state/autoupdate.go
```diff
@@ -106,8 +106,7 @@ func autoUpdate(args []string, cfg *config.Instance, an analytics.Dispatcher, ou
 		var msg string
 		if errs.Matches(err, &ErrStateExe{}) {
 			msg = anaConst.UpdateErrorExecutable
-		}
-		if errs.Matches(err, &ErrExecuteRelaunch{}) {
+		} else {
```
Naatan, line 109:
```suggestion
		} else if errs.Matches(err, &ErrExecuteRelaunch{}) {
```

Naatan (changes requested):

Naatan (approved):

PR #2466, opened 2023-03-30
Windows uninstaller prompts user to keep cache/config or delete every…

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1688" title="DX-1688" target="_blank">StoryDX-1688</a>  Uninstall State Tool supports `state clean uninstall --all` flag
  </td></table>
  <br />
 

…thing.

### DIFF
--- a/internal/runners/clean/uninstall.go
+++ b/internal/runners/clean/uninstall.go
@@ -16,23 +16,24 @@ import (
 	"github.com/ActiveState/cli/internal/svcctl"
 )
 
-type confirmAble interface {
+type promptable interface {
 	Confirm(title, message string, defaultChoice *bool) (bool, error)
+	Select(title, message string, choices []string, defaultChoice *string) (string, error)
 }
 
 type Uninstall struct {
-	out     output.Outputer
-	confirm confirmAble
-	cfg     *config.Instance
-	ipComm  svcctl.IPCommunicator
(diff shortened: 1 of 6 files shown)


internal/runners/clean/uninstall.go
```diff
@@ -62,13 +63,30 @@ func (u *Uninstall) Run(params *UninstallParams) error {
...
+	if err != nil {
+		return errs.Wrap(err, "Could not verify installation")
+	}
+
+	if params.Prompt {
+		choices := []string{
+			locale.Tl("uninstall_prompt", "Uninstall the State Tool, but keep runtime cache and configuration files"),
+			locale.Tl("uninstall_prompt_all", "Completely uninstall the State Tool, including runtime cache and configuration files"),
+		}
+		selection, err := u.prompt.Select("", "", choices, new(string))
```
mitchell-as, line 76: The header for this command, "Uninstall" is already displayed, so there's no need to adorn the prompt with a header.

MDrakos (changes requested):
The code seems fine, but nothing in the story says this should be Windows only or a flag. It appears that we want to prompt by default. Is there a discussion that happened that is not reflected in the ticket?

MDrakos (approved):

PR #2475, opened 2023-04-04
Transition startup entry on update

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1435" title="DX-1435" target="_blank">StoryDX-1435</a>  The state-svc autostarts as an .app on macOS
  </td></table>
  <br />


internal/osutils/autostart/autostart_darwin.go
```diff
@@ -20,6 +20,11 @@ const (
 )
 
 func enable(exec string, opts Options) error {
+	err := cleanLegacyFile(exec, opts)
+	if err != nil {
+		return errs.Wrap(err, "Could not clean legacy file")
+	}
```
Naatan, line 26: We just started moving this logic to the installer, might be worth waiting for David's PR and piggy backing off his changes:

https://github.com/ActiveState/cli/pull/2456/files/f7027b520621686fcac30ec61731c2c03c58955b..479edd891a2492b2f2cd7d9af55e1b8839a96e92

internal/osutils/autostart/legacy_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return false, errs.Wrap(err, "Could not open plist file")
+	}
+
+	decoder := plist.NewDecoder(reader)
+	var p LegacyPlist
+	err = decoder.Decode(&p)
+	if err != nil {
+		return false, errs.Wrap(err, "Could not decode plist file")
+	}
```
Naatan, line 30: Wouldn't these error conditions mean it's not a plist file? Are we not causing errors here for users who aren't upgrading from this legacy state?

Naatan (changes requested):
In addition to the lines tagged; why is this PR showing so many vendor changes?

internal/osutils/autostart/legacy_darwin.go
MDrakos: Without knowing more about plist files I don't think we can make that assumption. 

For users upgrading who aren't in a legacy state the return on line 32 should be false. We still have the `ProgramArguments` key in the file, which is an array but there should only be one entry in the non-legacy case.

MDrakos:
> In addition to the lines tagged; why is this PR showing so many vendor changes?

I added the library `howett.net/plist` to work with plist files then ran `go mod tidy` and `go mod vendor`

Naatan (changes requested):
This one still needs to be addressed:

https://github.com/ActiveState/cli/pull/2475/files#r1161891070

Recognize you may be blocked for now, but I'm not the blocker so bouncing it back ;)

MDrakos:
> This one still needs to be addressed:
> 
> https://github.com/ActiveState/cli/pull/2475/files#r1161891070
> 
> Recognize you may be blocked for now, but I'm not the blocker so bouncing it back ;)

Whoops looks like the PR is merged, so I was able to piggyback my changes with the new `Update` function. Tested out on my local machine and it's working as expected.

Naatan (approved):

PR #2480, opened 2023-04-07
Rework `state init`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1715" title="DX-1715" target="_blank">StoryDX-1715</a>  As a user I can rely on `state init` to be stable and intuitive
  </td></table>
  <br />
 

* Create and auto-push.
* Disambiguate "python".
* Drop "--skeleton" flag and config.
* Infer language from "state use" if possible.
* Refactored unit tests involving init.


internal/runners/initialize/init.go
```diff
@@ -33,19 +32,36 @@ type RunParams struct {
...
+		if projectDir := config.GetString(constants.GlobalDefaultPrefname); projectDir != "" {
+			if proj, err := project.FromPath(projectDir); err == nil {
+				if commitID := proj.CommitUUID(); commitID != "" {
+					if lang, err := model.FetchLanguageForCommit(commitID); err == nil {
+						params.Language = fmt.Sprintf("%s@%s", lang.Name, lang.Version)
+					}
+				}
+			}
+		}
+	}
```
MDrakos, line 63: nit: is there any way we can reduce the nesting? This could also be a helper function.

test/integration/init_int_test.go
```diff
@@ -19,54 +23,109 @@ type InitIntegrationTestSuite struct {
...
+	suite.runInitTest(true, "python", "python3")
+	suite.runInitTest(true, "python@3.10.0", "python3")
+	suite.runInitTest(true, "python@2.7.18", "python2")
 }
 
-func (suite *InitIntegrationTestSuite) runInitTest(addPath bool, config string, language string, args ...string) {
+func (suite *InitIntegrationTestSuite) runInitTest(addPath bool, lang string, expectedConfigLanguage string, args ...string) {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
+	ts.LoginAsPersistentUser()
```
MDrakos, line 49: Are we removing these newly created projects after the test completes? I _believe_ we are...

MDrakos (approved):
Nice work! Just a nit and a question for you.

PR #2481, opened 2023-04-10
Fix analytics integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1725" title="DX-1725" target="_blank">BugDX-1725</a>  Nightlies: Failing: TestAnalyticsIntegrationTestSuite/TestHeapEvents
  </td></table>
  <br />

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,7 +96,7 @@ jobs:
       - # === Install Deps ===
         name: Install Deps
         shell: bash
-        timeout-minutes: 1
+        timeout-minutes: 10
         run: state run install-deps
 
       - # === Preprocess ===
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       - # === Install Deps ===
         name: Install Deps
         shell: bash
-        timeout-minutes: 1
+        timeout-minutes: 10
```
mitchell-as, line 99: Do you have an instance where install deps took more than 1 minute? I'm asking because I never found one myself and I'd rather be notified after 1 minute to retry a failing job than after 10 minutes when I think it's happily running integration tests.

MDrakos: Yes, my previous PR was consistently failing on Windows because a `go install` step was taking longer than 1 minute.

mitchell-as (approved):

PR #2483, opened 2023-04-10
Rework `state init`

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1715" title="DX-1715" target="_blank">StoryDX-1715</a>  As a user I can rely on `state init` to be stable and intuitive
  </td></table>
  <br />
 

* Create and auto-push.
* Disambiguate "python".
* Drop "--skeleton" flag and config.
* Infer language from "state use" if possible.
* Refactored unit tests involving init.


MDrakos (approved):

cmd/state/internal/cmdtree/init.go
```diff
@@ -23,18 +24,14 @@ func newInitCommand(prime *primer.Values) *captain.Command {
...
-				Name:      "skeleton",
-				Shorthand: "",
-				Description: locale.Tr(
-					"flag_state_init_skeleton_description",
-					initialize.RecognizedSkeletonStyles(),
-				),
-				Value: &params.Style,
+				// Hidden flag for legacy Komodo support
+				Name:   "skeleton",
+				Value:  &unused,
```
Naatan, line 33: fwiw you could use `p.StrP("")` here.

internal/runners/initialize/init.go
```diff
@@ -33,19 +32,49 @@ type RunParams struct {
...
+		return ""
+	}
+	commitID := defaultProj.CommitUUID()
+	if commitID == "" {
+		return ""
+	}
+	lang, err := model.FetchLanguageForCommit(commitID)
+	if err != nil {
+		return ""
+	}
```
Naatan, line 69: To keep this logic centralized maybe we should move this to convenience methods under the globaldefault package? Mainly just thinking out loud.

internal/runners/initialize/init.go
```diff
@@ -33,19 +32,49 @@ type RunParams struct {
...
+	return fmt.Sprintf("%s@%s", lang.Name, lang.Version)
+}
+
+func sanitize(params *RunParams, config projectfile.ConfigGetter) error {
+	if params.Language == "" {
+		params.Language = inferLanguage(config)
```
Naatan, line 75: I recognize the code is already doing this, but ideally we should never override params, and consider them read-only. params should always be the thing the user passed in.

Suggest fixing this in the code for this runner while we're at it.

internal/runners/initialize/init.go
```diff
@@ -33,19 +32,49 @@ type RunParams struct {
...
+	}
+	commitID := defaultProj.CommitUUID()
+	if commitID == "" {
+		return ""
+	}
+	lang, err := model.FetchLanguageForCommit(commitID)
+	if err != nil {
+		return ""
+	}
+	return fmt.Sprintf("%s@%s", lang.Name, lang.Version)
```
Naatan, line 70: It feels a bit awkward that we're interpolating this just to then parse it again a bit later. Feels like we could just return the language struct value here and use it accordingly in the consuming code.

internal/runners/initialize/init.go
```diff
@@ -55,6 +84,15 @@ func sanitize(params *RunParams) error {
...
 	}
 
+	// Disambiguate "python", defaulting to "python3" if no version was given.
+	if langParts[0] == "python" {
+		if params.version == "" || strings.HasPrefix(params.version, "3") {
+			langParts[0] = "python3"
+		} else if strings.HasPrefix(params.version, "2") {
+			langParts[0] = "python2"
+		}
+	}
```
Naatan, line 94: We should use language constants here, eg. `language.Python3.String()`.

Also, since we are already interpreting the input here I think we should map this to a language constant, rather than to a string that we then redundantly infer with `language.MakeByName()`.

test/integration/init_int_test.go
```diff
@@ -19,54 +23,109 @@ type InitIntegrationTestSuite struct {
...
+func (suite *InitIntegrationTestSuite) TestInit_DisambiguatePython() {
 	suite.OnlyRunForTags(tagsuite.Init)
-	suite.runInitTest(false, sampleYAMLPython3, "python3@1.0")
+	suite.runInitTest(true, "python", "python3")
+	suite.runInitTest(true, "python@3.10.0", "python3")
+	suite.runInitTest(true, "python@2.7.18", "python2")
 }
 
-func (suite *InitIntegrationTestSuite) runInitTest(addPath bool, config string, language string, args ...string) {
+func (suite *InitIntegrationTestSuite) runInitTest(addPath bool, lang string, expectedConfigLanguage string, args ...string) {
```
Naatan, line 46: With all the projects this is now creating on the platform we should call `ts.DeleteUUIDProjects()` as a cleanup. eg. https://github.com/ActiveState/cli/blob/d35115ca6fd927436a27ccf27d53db87758c1865/test/integration/delete_uuid_projects_int_test.go#L19-L19

It seems vital that this is called before `ts.Close()`.

Naatan (changes requested):

internal/runners/initialize/init.go
mitchell-as: The `sanitize` function is changing params for use in a later function. Any local values stored in this function will be discarded. Should I spend the time doing a more involved refactor?

internal/runners/initialize/init.go
mitchell-as: Again, we're "sanitizing" parameters for use in a later function. I agree it's a bit awkward at first glance, but a more involved refactor may be out of scope here.

test/integration/init_int_test.go
mitchell-as: That is happening behind the scenes here:
https://github.com/ActiveState/cli/pull/2483/files#diff-43151b4055cc3065eb74749285d1a52636e027a42052a542e98c1992950a6bf2R63

internal/runners/initialize/init.go
```diff
@@ -175,6 +201,25 @@ func run(params *RunParams, out output.Outputer) (string, error) {
...
+	if err != nil {
+		return "", locale.NewInputError("err_no_default_branch")
+	}
+
+	err = model.UpdateProjectBranchCommitWithModel(platformProject, branch.Label, commitID)
+	if err != nil {
+		return "", locale.WrapError(err, "err_init_push", "Failed to push to the newly created Platform project at {{.V0}}", params.Namespace.String())
+	}
+
+	projectfile.StoreProjectMapping(config, params.Namespace.String(), filepath.Dir(proj.Source().Path()))
```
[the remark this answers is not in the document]
mitchell-as: There are multiple separate conditional branches in the `push` runner when it comes to creating projects and pushing changes. The code here spans over those branches. There is no readable way to reconcile the two into one shared function. I tried.

internal/runners/initialize/init.go
mitchell-as: It seems a bit specialized, so I'm not sure it's worth hoisting up.

internal/runners/initialize/init.go
mitchell-as: I made some changes to use constants, but I'm not sure I accurately reflected your request. Please let me know if I missed anything.

internal/runners/initialize/init.go
Naatan: Yes please. Since this story is effectively about stabilising the init command I think it makes sense to make these architectural improvements.

internal/runners/initialize/init.go
Naatan: Yeah that's a very fair point, but seeing as the story is about stabilizing I think it's fair game.

internal/runners/initialize/init.go
```diff
@@ -85,11 +85,11 @@ func sanitize(params *RunParams, config projectfile.ConfigGetter) error {
...
-	if langParts[0] == "python" {
+	if strings.ToLower(langParts[0]) == language.Python3.Requirement() {
 		if params.version == "" || strings.HasPrefix(params.version, "3") {
-			langParts[0] = "python3"
+			langParts[0] = language.Python3.String()
 		} else if strings.HasPrefix(params.version, "2") {
-			langParts[0] = "python2"
+			langParts[0] = language.Python2.String()
 		}
 	}
```
Naatan: While reviewing I noticed we have a function that looks like it might address this entire block of logic:

https://github.com/ActiveState/cli/blob/d35115ca6fd927436a27ccf27d53db87758c1865/internal/language/language.go#L122-L122

internal/runners/initialize/init.go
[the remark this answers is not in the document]
Naatan: Bah ok, thanks for digging!

test/integration/init_int_test.go
Naatan: Whoops, missed that!

mitchell-as:
Round 2. I'm okay if there's more churn.

internal/runners/initialize/init.go
```diff
@@ -50,184 +47,138 @@ func New(prime primeable) *Initialize {
...
-	return fmt.Sprintf("%s@%s", lang.Name, lang.Version)
+	return lang.Name, lang.Version, true
 }
 
-func sanitize(params *RunParams, config projectfile.ConfigGetter) error {
-	if params.Language == "" {
-		params.Language = inferLanguage(config)
+func (r *Initialize) Run(params *RunParams) error {
+	if err := params.Namespace.Validate(); err != nil {
```
Naatan, line 71: I could be remembering wrong; but I think captain already does this `Validate()` call, so this would be redundant.

internal/runners/initialize/init.go
```diff
@@ -50,184 +47,138 @@ func New(prime primeable) *Initialize {
...
 		}
 	}
 
-	params.language = language.Supported{language.MakeByName(langParts[0])}
-	if !params.language.Recognized() {
-		return language.UnrecognizedLanguageError(
-			params.Language,
-			language.RecognizedSupportedsNames(),
-		)
+	if fileutils.FileExists(filepath.Join(path, constants.ConfigFileName)) {
```
Naatan, line 86:
```suggestion
	if fileutils.TargetExists(filepath.Join(path, constants.ConfigFileName)) {
```

Naatan: May as well support symlinks here.

internal/runners/initialize/init.go
```diff
@@ -50,184 +47,138 @@ func New(prime primeable) *Initialize {
...
 	}
 
-	// Fail if target dir already has an activestate.yaml
-	if fileutils.FileExists(filepath.Join(params.Path, constants.ConfigFileName)) {
-		absPath, err := filepath.Abs(params.Path)
-		if err != nil {
-			return errs.Wrap(err, "IO failure")
-		}
-		return locale.NewInputError("err_init_file_exists", "", absPath)
+	if _, err := fileutils.PrepareDir(path); err != nil {
```
Naatan, line 90:
```suggestion
	if err := fileutils.MkdirUnlessExists(path); err != nil {
```

If you're not using the return value then you just need `MkdirUnlessExists`.

internal/runners/initialize/init.go
```diff
@@ -50,184 +47,138 @@ func New(prime primeable) *Initialize {
...
+		if inferred {
+			return locale.WrapError(err, "err_init_lang", "", languageName, languageVersion)
+		} else {
+			return locale.WrapInputError(err, "err_init_lang", "", languageName, languageVersion)
+		}
 	}
-
-	proj, err := project.FromPath(params.Path)
+	err = lang.Validate()
 	if err != nil {
```
Naatan:
```suggestion
	if err := lang.Validate(); err != nil {
```

Naatan (changes requested):

Naatan (approved):

mitchell-as:
@Naatan please force merge this one too.

PR #2484, opened 2023-04-10
Fix error output on package installation

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1711" title="DX-1711" target="_blank">BugDX-1711</a>  Starting 0.38 version, non-interactive ssh connection to the machine running ST missing messages
  </td></table>
  <br />
 

I'm not certain about this fix. The `dotprogress.Close()` function shouldn't be erroring every time on package installation but the real fix is [here](https://activestatef.atlassian.net/browse/DX-1731) which I believe is outside of the scope of this bug.


internal/runbits/runtime/progress/dotprogress.go
```diff
@@ -27,7 +27,7 @@ func NewDotProgressIndicator(out output.Outputer) *DotProgressDigester {
 
 func (d *DotProgressDigester) Handle(event events.Eventer) error {
 	switch event.(type) {
-	case events.Start:
+	case events.Start, events.SolveStart:
```
Naatan, line 30:
```suggestion
	case events.SolveStart:
```

Having it trigger on both is going to trigger the logic twice. Looking at the interactive version of this code the spinner is starting on SolveStart, not Start. So just listening to SolveStart seems appropriate.

.github/workflows/build.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       - # === Install Deps ===
         name: Install Deps
         shell: bash
-        timeout-minutes: 1
+        timeout-minutes: 10
```
Naatan, line 99:
```suggestion
```

Unrelated to this PR. Also if this is taking more than a minute then we need to get the time down, not increase the timeout.

Naatan (changes requested):

.github/workflows/build.yml
mitchell-as: It's a CI thing. Sometimes it's slow downloading and installing gozip, sometimes not. I wasn't experiencing any problems until yesterday, when it took me 5 tries before it worked on a 1 minute timeout.

Naatan: In my experience when it takes more than a minute it is going to timeout. Have you found that giving it more time sometimes works?

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2484/files#r1163129011

There might still be merit to this, but it's unrelated to this PR, and that conversation hasn't concluded. Suggest filing a jira story if this is something you'd like to address.

Naatan (approved):

PR #2487, opened 2023-04-12
Added CI and Interactive analytics dimensions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1694" title="DX-1694" target="_blank">TaskDX-1694</a>  As a PM I can see on the DW what percentage of our use is through CI
  </td></table>
  <br />

### DIFF
--- a/internal/analytics/client/async/client.go
+++ b/internal/analytics/client/async/client.go
@@ -28,7 +28,7 @@ import (
 type Client struct {
 	svcModel         *model.SvcModel
 	auth             *authentication.Auth
-	output           string
+	output           output.Outputer
 	projectNameSpace string
 	eventWaitGroup   *sync.WaitGroup
 	sessionToken     string
@@ -44,11 +44,7 @@ func New(svcModel *model.SvcModel, cfg *config.Instance, auth *authentication.Au
 		eventWaitGroup: &sync.WaitGroup{},
 	}
 
-	o := string(output.PlainFormatName)
-	if out.Type() != "" {
-		o = string(out.Type())
-	}
-	a.output = o
+	a.output = out
 	a.projectNameSpace = projectNameSpace
 	a.auth = auth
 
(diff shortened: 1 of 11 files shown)


internal/analytics/client/sync/client.go
```diff
@@ -49,7 +50,7 @@ type Client struct {
 var _ analytics.Dispatcher = &Client{}
 
 // New initializes the analytics instance with all custom dimensions known at this time
-func New(cfg *config.Instance, auth *authentication.Auth) *Client {
+func New(cfg *config.Instance, auth *authentication.Auth, out output.Outputer) *Client {
```
MDrakos, line 53: The `out` argument is not being used anywhere in this PR, all instances of `sync.New()` is just passing `nil`

internal/analytics/client/async/client.go
```diff
@@ -51,6 +53,8 @@ func New(svcModel *model.SvcModel, cfg *config.Instance, auth *authentication.Au
 	a.output = o
 	a.projectNameSpace = projectNameSpace
 	a.auth = auth
+	a.ci = condition.OnCI()
+	a.interactive = out.Config().Interactive
```
MDrakos, line 57: Determining interactive in the main state tool function is a bit more elaborate. Do we want to include the extra checks? See here: https://github.com/ActiveState/cli/blob/master/cmd/state/main.go#L101

MDrakos (changes requested):

internal/analytics/client/sync/client.go
mitchell-as: For now the ticket is only interested in knowing if the state tool is running on CI and interactively. However, in developing the PR I thought we might eventually want to know if the installer or some other tool with synchronous analytics is being run on CI and interactively, so I created the plumbing for it without connecting anything.

internal/analytics/client/async/client.go
mitchell-as: There is an open ticket to reconcile the differences: https://activestatef.atlassian.net/browse/DX-1667. This is possibly the third case now where we use `output`'s determination of interactive over `main`'s, so majority wins to me.

MDrakos (approved):

internal/analytics/client/sync/client.go
Naatan: I think it makes sense, but why not actually use it? Instead of passing in nil, just pass in the outputer. I think we have it available in all the situations where you're currently passing in nil.

Naatan (dismissed):
Looks good to me. Some discussion here that you may want to consider:

https://github.com/ActiveState/cli/pull/2487/files#r1167042187

mitchell-as:
Okay, I enabled it. Analytics will now receive 'interactive' for sync clients like installers. (It was already getting 'CI', which does not depend on output).

Naatan (approved):

mitchell-as:
@Naatan would you please force-merge this? It's hung on Target & Verify.

PR #2489, opened 2023-04-13
Update error reporting analytics

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-495" title="DX-495" target="_blank">TaskDX-495</a>  As a tech-lead I can diagnose the failure reasons of commands
  </td></table>
  <br />

### DIFF
--- a/pkg/cmdlets/errors/errors.go
+++ b/pkg/cmdlets/errors/errors.go
@@ -128,9 +128,9 @@ func ReportError(err error, cmd *captain.Command, an analytics.Dispatcher) {
 	_, hasMarshaller := err.(output.Marshaller)
 
 	cmdName := cmd.Name()
-	childCmd, err := cmd.Find(os.Args[1:])
-	if err != nil {
-		logging.Error("Could not find child command: %v", err)
+	childCmd, findErr := cmd.Find(os.Args[1:])
+	if findErr != nil {
+		logging.Error("Could not find child command: %v", findErr)
 	}
 
 	var flagNames []string
--- a/pkg/cmdlets/errors/errors.go
+++ b/pkg/cmdlets/errors/errors.go
@@ -146,16 +146,23 @@ func ReportError(err error, cmd *captain.Command, an analytics.Dispatcher) {
 
 	// Log error if this isn't a user input error
 	var action string
+	errorMsg := err.Error()
 	if !locale.IsInputError(err) {
 		multilog.Critical("Returning error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
 		action = anaConst.ActCommandError
 	} else {
 		action = anaConst.ActCommandInputError
+		for err != nil {
+			if locale.IsInputErrorNonRecursive(err) {
+				errorMsg = locale.ErrorMessage(err)
+				break
+			}
+			err = errors.Unwrap(err)
+		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/cmdlets/errors/errors.go
```diff
@@ -127,33 +127,37 @@ func ReportError(err error, cmd *captain.Command, an analytics.Dispatcher) {
...
-		trigger = append(trigger, flagNames...)
-
-		logging.Debug("Reporting input error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
-		an.Event(anaConst.CatDebug, anaConst.ActInputError, &dimensions.Values{
-			Trigger: p.StrP(strings.Join(trigger, " ")),
-		})
+		action = anaConst.ActCommandInputError
 	}
 
+	logging.Debug("Reporting error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
```
Naatan, line 156:
```suggestion
	logging.Debug("Reporting error:\n%s\nCreated at:\n%s", errs.Join(err, "\n").Error(), stack)
```

Realize you were just copying this over, but seems this was a debugging statement that probably shouldn't have been pushed in the first place. Line 150 already addresses logging non-input errors. And there's no real value in logging input errors to the log file imo.

pkg/cmdlets/errors/errors.go
```diff
@@ -127,33 +127,37 @@ func ReportError(err error, cmd *captain.Command, an analytics.Dispatcher) {
...
+	an.EventWithLabel(anaConst.CatDebug, action, strings.Join(label, " "), &dimensions.Values{
+		Error: p.StrP(err.Error()),
```
Naatan, line 158: If `action == anaConst.ActCommandInputError` this needs to be the topmost input error, rather than the topmost error (the two may not be the same).

Naatan (changes requested):

pkg/cmdlets/errors/errors.go
```diff
@@ -146,16 +146,23 @@ func ReportError(err error, cmd *captain.Command, an analytics.Dispatcher) {
...
 		action = anaConst.ActCommandError
 	} else {
 		action = anaConst.ActCommandInputError
+		for err != nil {
+			if locale.IsInputErrorNonRecursive(err) {
+				errorMsg = locale.ErrorMessage(err)
+				break
+			}
+			err = errors.Unwrap(err)
+		}
```
Naatan: You forgot to update `errorMsg`.

Also, you can probably simplify this by using the first slice entry returned by `locale.UnwrapError()`.

Naatan (changes requested):

pkg/cmdlets/errors/errors.go
MDrakos: Do we want to update the error to a special static message if we exit the for loop without encountering an input error? `errorMsg` is set on line 149 and updated when we encounter the input error.

The first entry in `locale.UnwrapError()` is not guaranteed to be an input error, just a localized one.

Naatan: Ah I forgot about that distinction! 

Looking at it again, I think your logic is sound! Makes sense to use the non-input error if we didn't encounter one (which shouldn't happen if the input error functions work as intended).

Naatan (approved):

PR #2491, opened 2023-04-13
Move log rotation into state-svc.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1516" title="DX-1516" target="_blank">TaskDX-1516</a>  Move log rotation to state-svc
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -185,8 +185,8 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 
 	logFile := logging.FilePath()
 	logging.Debug("Logging to %q", logFile)
-	logging.StartRotateLogTimer()
-	defer logging.StopRotateLogTimer()
+	stopTimer := logging.StartRotateLogTimer()
+	defer stopTimer()
 
 	p := NewService(ctx, cfg, an, auth, logFile)
 
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -185,6 +185,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 
 	logFile := logging.FilePath()
 	logging.Debug("Logging to %q", logFile)
+	go logging.RotateLogTimer()
 
 	p := NewService(ctx, cfg, an, auth, logFile)
 
@@ -235,6 +236,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 }
 
 func runStart(out output.Outputer, argText string) error {
+	go logging.RotateLogTimer()
 	if _, err := svcctl.EnsureStartedAndLocateHTTP(argText); err != nil {
 		if errors.Is(err, ipc.ErrInUse) {
 			out.Print("A State Service instance is already running in the background.")
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -186,6 +186,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 	logFile := logging.FilePath()
 	logging.Debug("Logging to %q", logFile)
 	logging.StartRotateLogTimer()
+	defer logging.StopRotateLogTimer()
 
 	p := NewService(ctx, cfg, an, auth, logFile)
 
@@ -236,7 +237,6 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 }
 
 func runStart(out output.Outputer, argText string) error {
-	logging.StartRotateLogTimer()
 	if _, err := svcctl.EnsureStartedAndLocateHTTP(argText); err != nil {
 		if errors.Is(err, ipc.ErrInUse) {
 			out.Print("A State Service instance is already running in the background.")
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -185,7 +185,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 
 	logFile := logging.FilePath()
 	logging.Debug("Logging to %q", logFile)
-	go logging.RotateLogTimer()
+	logging.StartRotateLogTimer()
 
 	p := NewService(ctx, cfg, an, auth, logFile)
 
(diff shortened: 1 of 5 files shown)


cmd/state-svc/main.go
```diff
@@ -185,6 +185,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 
 	logFile := logging.FilePath()
 	logging.Debug("Logging to %q", logFile)
+	go logging.RotateLogTimer()
```
daved, line 188: nit: It's usually recommended for packages to abstract away goroutines. So `logging.RotateLogTimer()` would be non-blocking and manage the goroutine and whatever backpressure (i.e. errors, channel outputs) that is required, which is none in this case (IIUC).

internal/logging/rotate.go
```diff
@@ -39,3 +45,46 @@ func rotateLogs(files []fs.FileInfo, timeCutoff time.Time, amountCutoff int) []f
...
+		}
+	}
+
+	rotateLogsOnDisk()
+	for {
+		tick := time.Tick(timeout)
+		select {
+		case <-tick:
+			rotateLogsOnDisk()
+		}
```
daved, line 88: Please consider using `time.After`. Also consider the note in the docs regarding cleanup (I think we should be good to go using After, but give it your own judgment): https://pkg.go.dev/time#After

daved (changes requested):

cmd/state-svc/main.go
mitchell-as: I agree. Thanks for pointing this out.

daved (dismissed):

cmd/state-svc/main.go
```diff
@@ -185,6 +185,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 
 	logFile := logging.FilePath()
 	logging.Debug("Logging to %q", logFile)
+	logging.StartRotateLogTimer()
```
Naatan, line 188: This should be closed at shutdown, otherwise, we're shutting down with an active goroutine running.

cmd/state-svc/main.go
```diff
@@ -235,6 +236,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 }
 
 func runStart(out output.Outputer, argText string) error {
+	logging.StartRotateLogTimer()
```
Naatan, line 239:
```suggestion
```

Start will invoke foreground, and will exit almost immediately.

cmd/state-svc/test/integration/svc_int_test.go
```diff
@@ -249,6 +253,72 @@ func (suite *SvcIntegrationTestSuite) checkEnabled(exec string, opts autostart.O
...
+	for i := 1; i <= numFooFiles; i++ {
+		logFile := filepath.Join(logDir, fmt.Sprintf("foo-%d%s", i, logging.FileNameSuffix))
+		err := fileutils.Touch(logFile)
+		suite.Require().NoError(err, "could not create dummy log file")
+		err = os.Chtimes(logFile, thirtyDaysOld, thirtyDaysOld)
+		suite.Require().NoError(err, "must be able to change file modification times")
+	}
+
+	// Override state-svc log duration timer duration from 1 minute to 10 seconds for this test.
+	logRotateTimerDuration := 10
```
Naatan, line 275: Can we bring this down as much as possible? This adds 10 seconds to the test, which is pretty significant.

internal/constants/constants.go
```diff
@@ -166,6 +166,10 @@ const RuntimeUsageSilenceTimeOverrideEnvVarName = "ACTIVESTATE_OVERRIDE_RTUSAGE_
 // SvcAuthPollingRateEnvVarName is used to override the default polling rate for syncing the authenticated state with the svc
 const SvcAuthPollingRateEnvVarName = "ACTIVESTATE_SVC_AUTH_POLLING_RATE"
 
+// StateSvcLogRotateDurationEnvVarName is the environment variable used to override the default
+// log rotation timer duration (60 seconds).
+const SvcLogRotateTimerDurationEnvVarName = "ACTIVESTATE_CLI_LOG_ROTATE_TIMER_DURATION"
```
Naatan, line 171: Please make this take the value in ms rather than seconds (for consistency with other similar env vars).

internal/logging/rotate.go
```diff
@@ -39,3 +45,46 @@ func rotateLogs(files []fs.FileInfo, timeCutoff time.Time, amountCutoff int) []f
...
+	for _, file := range rotate {
+		if err := os.Remove(filepath.Join(logDir, file.Name())); err != nil {
+			Error("Could not clean up old log: %s, error: %v", file.Name(), err)
+		}
+	}
+}
+
+// RotateLogListener rotates logs on a timer.
+func StartRotateLogTimer() {
+	timeout := 1 * time.Minute
```
Naatan, line 74: Nit: this is an interval, not a timeout.

Naatan (changes requested):

cmd/state-svc/test/integration/svc_int_test.go
mitchell-as: I'll try bringing it down to 4 seconds, but intermittent CI slowness may cause more nightly test failure troubles.

internal/logging/rotate.go
mitchell-as: Thanks. I was looking for the right word here.

internal/logging/rotate.go
```diff
@@ -69,22 +69,30 @@ func rotateLogsOnDisk() {
...
+	stopTimer = make(chan bool)
 	go func() {
 		rotateLogsOnDisk()
 		for {
 			select {
-			case <-time.After(timeout):
+			case <-time.After(interval):
 				rotateLogsOnDisk()
+			case <-stopTimer:
+				return
```
Naatan:
```suggestion
			case <-stopTimer:
				return
			case <-time.After(interval):
				rotateLogsOnDisk()
```

I realize both of these triggering in the same loop is pretty unlikely, but still worth being mindful of.

cmd/state-svc/main.go
```diff
@@ -186,6 +186,7 @@ func runForeground(cfg *config.Instance, an *anaSync.Client, auth *authenticatio
 	logFile := logging.FilePath()
 	logging.Debug("Logging to %q", logFile)
 	logging.StartRotateLogTimer()
+	defer logging.StopRotateLogTimer()
```
Naatan, line 189: Can we make `StartRotateLogTimer()` return the stop function and invoke it here? 

Realize the logging lib is riddled with globals already, but still rather avoid adding to that technical debt.

Naatan (changes requested):

Naatan (approved):

mitchell-as:
@Naatan Please force merge this one too.

PR #2493, opened 2023-04-13
Don't download and install artifacts that originate from an internal namespace

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1740" title="DX-1740" target="_blank">BugDX-1740</a>  Failed to download artifact cryptography-rust-vendor (6e2397b4-70bc-5352-91f6-3e6f5c69abc5)
  </td></table>
  <br />
 

I'm not pleased with how dirty this is, but this is for a hotfix so I'm trying to touch minimal code. Also this code will all soon be replaced with buildplan specific code which should incidentally also solve issues of this variety.

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -569,7 +569,7 @@ func (s *Setup) setupArtifactSubmitFunction(a artifact.ArtifactDownload, ar *art
 		// If artifact has no valid download, just count it as completed and return
 		if strings.HasPrefix(a.UnsignedURI, "s3://as-builds/noop/") ||
 			// Internal namespace artifacts are not to be downloaded
-			ar.Namespace == inventory_models.NamespaceCoreTypeInternal {
+			(ar != nil && ar.Namespace == inventory_models.NamespaceCoreTypeInternal) {
 			logging.Debug("Skipping setup of noop artifact: %s", a.ArtifactID)
 			if _, expected := expectedArtifactInstalls[a.ArtifactID]; expected {
 				if err := s.eventHandler.Handle(events.ArtifactDownloadSkipped{a.ArtifactID}); err != nil {
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -49,6 +49,7 @@ import (
 	"github.com/faiface/mainthread"
 	"github.com/gammazero/workerpool"
 	"github.com/go-openapi/strfmt"
+	"github.com/thoas/go-funk"
 )
 
 // MaxConcurrency is maximum number of parallel artifact installations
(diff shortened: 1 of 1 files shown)


Naatan:
@mdrohmann hope you don't mind I flagged you for review as I want some eyes on whether this fix is appropriate (from an API pov). Incidentally, you know this go code so win-win.. ;)

mdrohmann (dismissed):
This seems legit as a hotfix.  FWIU the internal packages in the internal namespace shouldn't appear in the runtime closure.  But to be clear: That might be a correct assumption, but not a safe one.

There is no formal contract for that...

I wonder if it is safer to just ignore the error that actually occurred.  I can't really tell why the ArtifactDownload was not expected, and why that would be a problem...

Naatan:
> I wonder if it is safer to just ignore the error that actually occurred. I can't really tell why the ArtifactDownload was not expected, and why that would be a problem...

The reason it wasn't expected is because the expectation was based on the artifacts returned from this function:

https://github.com/ActiveState/cli/blob/6d188a424a66bdf2e572811c5536bed2b02f3bd9/pkg/platform/runtime/artifact/download.go#L26-L26

Which apparently did not include the internal artifact. I think it would've worked if it was coming from a build in progress as builds in progress do not use this function.

A design principle with the new progress indicator was to not just ignore things that look like failures, such that we can identify and fix issues, which is why we're now talking about this. In a way this bug is a signal of success.. :p Although obviously, it would've been better if we could have caught this before it went live.

mdrohmann:
Hmm... That is indeed odd, that it would not appear in the BuildStatus response, but you'd get a build status update for the file...
Do we still have access to the commit that caused the problem?

Naatan:
> Hmm... That is indeed odd, that it would not appear in the BuildStatus response, but you'd get a build status update for the file... Do we still have access to the commit that caused the problem?

I've been testing with `f94b7ab8-dd56-41e9-94fe-95181d3685a6`.

mdrohmann:
Strange:  It appears in the build status response:

```
    {
      "artifact_id": "6c998ac0-40fc-5e28-91f0-dce8d06eb26e",
      "build_state": "succeeded",
      "build_timestamp": "2023-01-23T06:27:29.658Z",
      "checksum": "134d75747e3322332ad0ed4d3a0a185cd6d898ee6f990ab472f43c4927e9429c",
      "dependency_ids": null,
      "ingredient_version_id": "4a4d39b6-54ab-56fc-8a0d-b00a4dbe5882",
      "log_uri": "s3://as-builds/production/internal/cryptography-rust-vendor/36.0.1/1/6c998ac0-40fc-5e28-91f0-dce8d06eb26e/logs.jsonl",
      "mime_type": "application/x.artifact",
      "platform_id": "96b7e6f2-bebf-564c-bc1c-f04482398f38",
      "uri": "s3://as-builds/production/internal/cryptography-rust-vendor/36.0.1/1/6c998ac0-40fc-5e28-91f0-dce8d06eb26e/artifact.tar.gz"
    },
```

No idea, why it wouldn't be added to the `downloads` array...

PR #2494, opened 2023-04-14
Unify asy conditionals and variables data structure

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1661" title="DX-1661" target="_blank">StoryDX-1661</a>  As a user I can access the same variables in scripts as I can in conditionals
  </td></table>
  <br />


pkg/projget/projget.go
```diff
@@ -0,0 +1,58 @@
+package projget
```
Naatan, line 1: Why is this in its own package?

internal/constraints/constraints.go
```diff
@@ -72,62 +42,36 @@ func NewConditional(a *authentication.Auth) *Conditional {
...
+		// registration is needed at greater depths, this will need to be
+		// reworked (and may not be possible without expansive refactoring).
+		switch d1Val.Type().Kind() {
+		case reflect.Func:
+			c.RegisterFunc(f.Name, d1Val.Interface())
+
+		default:
+			c.RegisterParam(f.Name, d1Val.Interface())
+		}
 	}
```
Naatan, line 74: If there is one part of this PR that needs thorough commenting it's this one.. Please revisit this code such that someone can read through it and understand what's happening without hitting the docs.

pkg/projectfile/vars/vars.go
```diff
@@ -0,0 +1,143 @@
...
+			JWT:   auth.BearerToken(),
+		},
+	}
+}
+
+type Vars struct {
+	Project *Project
+	OS      *OS
+	Shell   string
+	Mixin   func() *Mixin
```
Naatan, line 128: I'm not seeing why this needs reflection. Why can't we give these values enough information to identify and act on their type differences? eg. add an interface with `IsMixing()` for example.

Naatan (changes requested):
Haven't done a full review as I feel like we can land this without any need for reflection. Figure we should address that before addressing the rest of the review.

pkg/projget/projget.go
daved: A story has been filed to remove this package: https://activestatef.atlassian.net/browse/DX-1808

Naatan (changes requested):

daved:
Closing in favor of #2568.

PR #2495, opened 2023-04-17
Installer should overwrite State Service.app.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1722" title="DX-1722" target="_blank">BugDX-1722</a>  Install 0.38.0-RC1 fails with "Installation of service app failed"
  </td></table>
  <br />


MDrakos (dismissed):
Is there an easy way to integration test this?

mitchell-as:
Yes, it looks like there is!

MDrakos (approved):
Nice work!

PR #2498, opened 2023-04-18
Start-story improvements

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1584" title="DX-1584" target="_blank">TaskDX-1584</a>  start-story improvements
  </td></table>
  <br />

### DIFF
--- a/scripts/internal/workflow-helpers/jira.go
+++ b/scripts/internal/workflow-helpers/jira.go
@@ -1,6 +1,7 @@
 package workflow_helpers
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
 	"regexp"
@@ -195,6 +196,34 @@ func UpdateJiraFixVersion(client *jira.Client, issue *jira.Issue, versionID stri
 	issueUpdate.Fields.FixVersions = []*jira.FixVersion{{ID: versionID}}
 	_, response, err := client.Issue.Update(issueUpdate)
 	res, err2 := ioutil.ReadAll(response.Body)
+	if err2 != nil {
+		res = []byte(err2.Error())
+	}
+	if err != nil {
+		return errs.Wrap(err, string(res))
+	}
+	return nil
+}
+
+func UpdateJiraStatus(client *jira.Client, issue *jira.Issue, statusName string) error {
+	transitions, _, err := client.Issue.GetTransitions(issue.ID)
+	if err != nil {
+		return errs.Wrap(err, "failed to get Jira transitions")
+	}
+
+	var transition *jira.Transition
+	for _, t := range transitions {
+		if t.To.Name == statusName {
+			transition = &t
+			break
+		}
+	}
+	if transition == nil {
+		return errs.New("failed to find a Jira transition that changes the status to %s for issue %s", statusName, issue.Key)
+	}
+
+	response, err := client.Issue.DoTransition(issue.ID, transition.ID)
+	res, err2 := io.ReadAll(response.Body)
 	if err2 != nil {
 		res = []byte(err2.Error())
 	}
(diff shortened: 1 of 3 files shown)


scripts/internal/workflow-helpers/jira.go
```diff
@@ -203,3 +204,31 @@ func UpdateJiraFixVersion(client *jira.Client, issue *jira.Issue, versionID stri
...
+			transition = &t
+			break
+		}
+	}
+	if transition == nil {
+		return errs.New("failed to find a Jira transition that changes the status to %s for issue %s", statusName, issue.Key)
+	}
+
+	response, err := client.Issue.DoTransition(issue.ID, transition.ID)
+	res, err2 := io.ReadAll(response.Body)
```
mitchell-as, line 226: Are you sure `response` is a valid object if `err` is not `nil`? It's probably worth mentioning this in a comment because it looks wrong at first glance.

scripts/start-story/main.go
```diff
@@ -105,6 +106,10 @@ func run() error {
 	if err != nil {
 		return errs.Wrap(err, "failed to checkout base ref, stdout:\n%s\nstderr:\n%s", stdout, stderr)
 	}
+	stdout, stderr, err = exeutils.ExecSimpleFromDir(environment.GetRootPathUnsafe(), "git", []string{"pull", "--rebase"}, nil)
```
mitchell-as, line 109: I might be misremembering, but can `git pull --rebase` hang due to an interactive conflict/merge process? If it cannot hang, are merge conflicts going to affect this action?

scripts/start-story/main.go
```diff
@@ -115,6 +120,14 @@ func run() error {
 	}
 	finish()
 
+	if meta.JiraIssue.Fields.Status.Name == "To Do" || meta.JiraIssue.Fields.Status.Name == "Pending" {
```
mitchell-as, line 123: Nit: might want to make these strings constants. We'll probably never change them in JIRA, but if we do, changing one constant is easier than search/replace.

mitchell-as (changes requested):

scripts/start-story/main.go
Naatan: It shouldn't since this would run in non-interactive mode. Merge conflicts would be raised as an error as the command would exit non-zero.

scripts/start-story/main.go
```diff
@@ -120,7 +121,7 @@ func run() error {
 	}
 	finish()
 
-	if meta.JiraIssue.Fields.Status.Name == "To Do" || meta.JiraIssue.Fields.Status.Name == "Pending" {
+	if meta.JiraIssue.Fields.Status.Name == wh.JiraStatusTodo || meta.JiraIssue.Fields.Status.Name == wh.JiraStatusPending {
 		finish = wc.PrintStart("Updating jira issue to In Progress")
 		if err := wh.UpdateJiraStatus(jiraClient, meta.JiraIssue, "In Progress"); err != nil {
```
mitchell-as, line 126: You'll want to use a constant for "In Progress" too.

scripts/start-story/main.go
```diff
@@ -151,7 +152,7 @@ func fetchMeta(ghClient *github.Client, jiraClient *jira.Client, jiraIssueID str
 		return Meta{}, errs.New("Target issue has fixVersion '%s', which has either been archived or released\n", fixVersion.Name)
 	}
 
-	if jiraIssue.Fields.Status.Name != "To Do" && jiraIssue.Fields.Status.Name != "In Progress" && jiraIssue.Fields.Status.Name != "Pending" {
+	if !funk.ContainsString([]string{wh.JiraStatusTodo, wh.JiraStatusInProgress, wh.JiraStatusPending}, jiraIssue.Fields.Status.Name) {
 		return Meta{}, errs.New("Story is in the %s state, but only 'To Do', 'In Progress' and 'Pending' are valid states to start a story from.", jiraIssue.Fields.Status.Name)
```
mitchell-as, line 156: Maybe use constants in this error message too.

mitchell-as (changes requested):
Almost there!

scripts/start-story/main.go
Naatan: Woops, thanks for catching!

mitchell-as (approved):

PR #2503, opened 2023-04-24
Add messaging mechanic

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1742" title="DX-1742" target="_blank">StoryDX-1742</a>  The State Tool Can Display a Message of the Month/Day/Week to users on selected commands
  </td></table>
  <br />


cmd/state-svc/internal/messages/messages.go
```diff
@@ -0,0 +1,171 @@
...
+const ConfigKeyLastReport = "messages.last_reported"
+
+type Messages struct {
+	cfg        *config.Instance
+	auth       *auth.Auth
+	baseParams *ConditionParams
+	poll       *poller.Poller
+}
+
+func New(cfg *config.Instance, auth *auth.Auth) (*Messages, error) {
```
daved, line 31: nit: I suggest reworking the params. From auth, only the email is needed. Instead of config, the func can require an interface with GetStringMap (though, I'm wondering where SetStringMap is, and if it's needed). It would also be nice for sysinfo work to be performed by the caller and sent into this function, and same with semver. This would remove 4 direct dependencies, and reduce controller noise within this construction function.

cmd/state/internal/messenger/messenger.go
```diff
@@ -0,0 +1,99 @@
+package messenger
```
daved, line 1: I suggest for this to be in cmd/state/internal/cmdtree/intercepts/messenger.

daved (changes requested):
My main concern is that I don't see where the config is being updated when reporting. Sorry if I missed it.

messages.json
```diff
@@ -0,0 +1,6 @@
+[
```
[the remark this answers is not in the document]
Naatan: Woops, no.. that was just me testing stuff.

cmd/state-svc/internal/messages/messages.go
Naatan: For now, the cost outweighs the value to me. The value is forward compatibility. The cost is that walking through this code now has hurdles. As forward compatibility is likely a non-issue with the config and auth libs, I don't see merit in changing this at the moment.

daved:
Is the config update handled elsewhere? Or maybe I'm missing where it's happening.

Naatan:
> Is the config update handled elsewhere? Or maybe I'm missing where it's happening.

Can you elaborate? What config update? We check the config anew each time so in theory there should be no out of sync issues.

daved:
I found where we grab the "last reported" data from the config, but I didn't find where we set that data when something is reported.

Naatan:
> I found where we grab the "last reported" data from the config, but I didn't find where we set that data when something is reported.

Ohhhh gotcha! Yes I think you're right; I forgot that! Please mark this as "Changes Requested" so I can notify you properly when this has been addressed.

daved (changes requested):
Mentioned "last reported" setting in conversation.

daved (changes requested):

daved (approved):

daved (approved):

PR #2506, opened 2023-04-25
Comprehensive JSON support.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1721" title="DX-1721" target="_blank">StoryDX-1721</a>  As a user I can expect json output to be usable at all times
  </td></table>
  <br />
 

The idea is to have all `output.Print()` statements print an object that has both `MarshalOutput()` and `MarshalStructured()` methods that return plain and structured data, respectively. Anything else to print (i.e. bare strings) should be reduced to `output.Notice()` calls.

If a lone string is passed to `output.Print()` but structured output is requested (e.g. `-o json`), a "json output not supported" error will be printed (in structured format).

The following commands are, or were made to be, JSON/structured-compatible:

```
checkout
init
switch

activate
shell
use
use show

projects
projects remote
show

bundles install
bundles search
bundles uninstall
info
install
packages
search
uninstall

auth
branch
languages
organizations
platforms
platforms search
secrets
secrets get
cve
cve report

fork
history
pull
reset
revert

events
scripts

config get
config set
export config
export env
export jwt
export private-key
export recipe
update lock
```

Commands not on this list are either not JSON/structured-compatible and will raise an error when they attempt to print to the terminal, or they do not print anything at all and are okay (until they print something that may not be compatible). For example, `state use reset` is not explicitly JSON/structured-compatible because it only outputs notices to the user.


internal/runbits/refresh.go
```diff
@@ -27,7 +27,7 @@ func RefreshRuntime(auth *authentication.Auth, out output.Outputer, an analytics
 	}
 
 	if !changed && isCached {
-		out.Print(locale.Tl("pkg_already_uptodate", "Requested dependencies are already configured and installed."))
+		out.Notice(locale.Tl("pkg_already_uptodate", "Requested dependencies are already configured and installed."))
```
mitchell-as, line 30: `output.Notice()` is ignored by JSON/structured outputs. All of these types of informative messages need to be output as notices from now on, or they will contaminate JSON/structured output.

internal/runbits/requirements/requirements.go
```diff
@@ -78,12 +78,20 @@ func NewRequirementOperation(prime primeable) *RequirementOperation {
 
 const latestVersion = "latest"
 
-type RequirementOperationParams struct {
```
mitchell-as, line 81: This struct was unused so I removed it.

internal/runbits/requirements/requirements.go
```diff
@@ -261,18 +269,24 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
...
-	} else {
-		out.Print(locale.Tr(fmt.Sprintf("%s_%s", ns.Type(), operation), requirementName))
+	// Print the result
+	message := locale.Tr(fmt.Sprintf("%s_version_%s", ns.Type(), operation), requirementName, requirementVersion)
+	if requirementVersion == "" {
+		message = locale.Tr(fmt.Sprintf("%s_%s", ns.Type(), operation), requirementName)
 	}
-
-	out.Print(locale.T("operation_success_local"))
+	out.Print(&requirementsOutput{
```
mitchell-as, line 281: All `output.Print()` calls going forward should print an object that has `MarshalOutput()` and `MarshalStructured()` methods. That way, the correct thing is printed to the terminal depending on the output mode.

internal/runners/auth/auth.go
```diff
@@ -54,8 +54,17 @@ func (p AuthParams) verify() error {
 	return nil
 }
 
-type SignupParams struct {
```
mitchell-as, line 57: I moved this to signup.go where it belongs.

internal/runners/languages/languages.go
```diff
@@ -24,17 +24,16 @@ func NewLanguages(prime primeable) *Languages {
...
-	if f == output.PlainFormatName {
-		return l.Languages
-	}
-	return l
+func (o *languagesOutput) MarshalOutput(f output.Format) interface{} {
+	return o.Languages
+}
+
+func (o *languagesOutput) MarshalStructured(f output.Format) interface{} {
+	return o
```
mitchell-as, line 36: This results in `{"languages":[{...}]}`, which is unusual and inconsistent with most of our other output. However, perhaps backwards-compatibility is desirable here, so I left it as-is. I would recommend simply returning a list of languages though.

pkg/cmdlets/errors/errors.go
```diff
@@ -85,6 +85,21 @@ func trimError(msg string) string {
 	return strings.TrimRight(msg, " .")
 }
 
+func (o *OutputError) MarshalStructured(f output.Format) interface{} {
```
mitchell-as, line 88: Unwrap all errors and return a list of error strings to be converted to a `{"errors":["error1","error2",...],"code":x}` JSON object.

test/integration/init_int_test.go
```diff
@@ -128,6 +128,21 @@ func (suite *InitIntegrationTestSuite) TestInit_InferLanguageFromUse() {
...
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	// Rather than triggering another build/solve on the Platform, just verify the command is
+	// JSON-compatible.
+	cp := ts.Spawn("init", "-o", "json")
+	cp.ExpectLongString(`{"errors":["The following argument is required:`)
+	cp.Expect(`"code":`)
+	cp.ExpectNotExitCode(0)
+	AssertValidJSON(suite.T(), cp)
```
Naatan, line 143: This is just verifying that the error reporting is json compatible, rather than the command output itself. Also I don't think it's worth testing for error cases in integration tests unless it's some very specific error case we want to curate.

test/integration/invite_int_test.go
```diff
@@ -28,6 +28,20 @@ func (suite *InviteIntegrationTestSuite) TestInvite_NotAuthenticated() {
...
+	defer ts.Close()
+
+	url := "https://platform.activestate.com/ActiveState-CLI/Invite-Test?branch=main&commitID=eb8dd176-d557-4adc-8b79-7b17e3a03bd7"
+	suite.Require().NoError(fileutils.WriteFile(filepath.Join(ts.Dirs.Work, "activestate.yaml"), []byte("project: "+url)))
+
+	cp := ts.Spawn("invite", "test-user@test.com", "-o", "json")
+	cp.Expect(`{"errors":["You need to authenticate`)
+	cp.ExpectNotExitCode(0)
+	AssertValidJSON(suite.T(), cp)
+}
```
Naatan, line 43: As with the other test; this is just testing the error case, not worth having a test for.

Naatan (changes requested):

internal/output/json.go
```diff
@@ -103,3 +106,21 @@ func prepareJSONValue(v interface{}) interface{} {
...
+// toJsonError attempts to convert the given interface into a jsonError struct.
+// It accepts an error object, a list of string error messages, or a single string error message.
+// If it cannot perform the conversion, it returns a jsonError indicating so.
+func toJsonError(v interface{}) jsonError {
+	if err, ok := v.(error); ok {
+		return jsonError{[]string{err.Error()}, 1}
+	}
+	if strings, ok := v.([]string); ok {
+		return jsonError{strings, 1}
+	}
```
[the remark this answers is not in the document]
mitchell-as: It comes from unwrapping errors into a list and returning that list: https://github.com/ActiveState/cli/blob/53b093659b6bb19f963e7fac8114cab5039a11af/pkg/cmdlets/errors/errors.go#L88-L100

internal/runners/auth/auth.go
```diff
@@ -79,12 +88,12 @@ func (a *Auth) Run(params *AuthParams) error {
...
-		output.NewFormatter(data).
-			WithFormat(output.PlainFormatName, locale.T("logged_in_as", map[string]string{
-				"Name": a.Auth.WhoAmI(),
-			})),
-	)
+	a.Outputer.Print(&authOutput{
+		locale.T("logged_in_as", map[string]string{
+			"Name": data.Username,
+		}),
+		*data,
```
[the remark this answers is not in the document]
mitchell-as: `data` is a pointer and I cannot create authOutput struct members from a pointer. It has to be a value.

internal/runners/revert/revert.go
```diff
@@ -115,10 +116,12 @@ func (r *Revert) Run(params *Params) error {
...
 	}
-	r.out.Print(locale.Tl("revert_info", "You are about to revert{{.V0}} the following commit:", preposition))
-	commit.PrintCommit(r.out, targetCommit, orgs)
+	if !output.IsStructuredFormat(r.out.Type()) {
+		r.out.Print(locale.Tl("revert_info", "You are about to revert{{.V0}} the following commit:", preposition))
+		commit.PrintCommit(r.out, targetCommit, orgs)
+	}
 
-	defaultChoice := params.Force
+	defaultChoice := params.Force || !r.out.Config().Interactive
```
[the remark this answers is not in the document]
mitchell-as: I originally thought this too, but when you give a prompt a default value, it returns exactly that value if the prompt is not interactive. It does not return "true" by default, regardless of the default you give it. The current behavior makes sense to me.

pkg/cmdlets/commit/commit.go
```diff
@@ -169,3 +151,45 @@ func usernameForID(id strfmt.UUID, orgs []gmodel.Organization) string {
...
+func (o *commitsOutput) MarshalStructured(format output.Format) interface{} {
+	return o
+}
+
+func PrintCommits(out output.Outputer, commits []*mono_models.Commit, orgs []gmodel.Organization, lastRemoteID *strfmt.UUID) error {
+	data, err := newCommitsOutput(commits, orgs, lastRemoteID)
+	if err != nil {
+		return errs.Wrap(err, "Unable to fetch commit data")
+	}
+	out.Print(data)
```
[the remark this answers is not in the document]
mitchell-as: It does print as a single slice. This is a `commitsOutput` (not a `commitOutput`), which is a slice of the latter, and is output as a JSON list.

test/integration/init_int_test.go
mitchell-as: You're right. I will assert proper json error format in a unit test.

internal/runners/export/config/config.go
```diff
@@ -24,6 +24,30 @@ type primeable interface {
 	primer.Configurer
 }
 
+type mapOutput struct {
+	map_ map[string]string
```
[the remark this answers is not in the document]
mitchell-as: That's funny. My first thought here was to use generics, but my first (and only) attempt failed. I may try again, or I may try your `output.Prepare()` suggestion.

internal/runners/secrets/secrets.go
```diff
@@ -62,6 +57,37 @@ func NewList(client *secretsapi.Client, p listPrimeable) *List {
...
+	out  output.Outputer
+	data []*secretData
+}
+
+func (o *listOutput) MarshalOutput(format output.Format) interface{} {
+	return struct {
+		Data []*secretData `opts:"verticalTable" locale:","`
+	}{
+		o.data,
+	}
```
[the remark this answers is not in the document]
mitchell-as: It is a slice, but it has this weird structure to output it in a certain form/style (verticalTable).

internal/runners/auth/auth.go
```diff
@@ -54,8 +54,17 @@ func (p AuthParams) verify() error {
...
+	message string
+	userData
+}
+
+func (o *authOutput) MarshalOutput(format output.Format) interface{} {
+	return o.message
+}
+
+func (o *authOutput) MarshalStructured(format output.Format) interface{} {
+	return o.userData
```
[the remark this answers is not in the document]
mitchell-as: The good news: this ended up being a great idea, simplified most of the calls, and makes things easier for future commands to implement.

The bad news: another big diff :(

pkg/cmdlets/errors/errors.go
```diff
@@ -85,6 +85,21 @@ func trimError(msg string) string {
...
+		// It's possible the error came from cobra or something else low level that doesn't use localization
+		logging.Warning("Error does not have localization: %s", errs.JoinMessage(o.error))
+		rerrs = []error{o.error}
+	}
+	for _, errv := range rerrs {
+		message := trimError(locale.ErrorMessage(errv))
+		errors = append(errors, message)
+	}
+	return errors
+}
```
Naatan, line 101: Picking up from the discussion here:

https://github.com/ActiveState/cli/pull/2506#discussion_r1179664735

Can we make this return a slice of errors instead? Seems if it's only to satisfy the commented code linked above then the act of turning errors into strings is better handled when the error itself is reported.

internal/output/mediator.go
```diff
@@ -53,13 +53,17 @@ func (m *Mediator) Notice(v interface{}) {
...
 		if vt, ok := v.(StructuredMarshaller); ok {
 			return vt.MarshalStructured(format)
 		}
-		strv := fmt.Sprintf("%v", v)
+		multilog.Error("%s output not supported for message: %v", string(format), v)
 		return jsonError{
-			[]string{locale.Tl("err_no_structured_output", "{{.V0}} output not supported for message: {{.V1}}", string(format), strv)},
+			[]string{locale.Tl(
+				"err_no_structured_output",
+				"This command does not support the {{.V0}} output format. Please try again without that output flag",
```
Naatan, line 64:
```suggestion
				"This command does not support the {{.V0}} output format. Please try again without that output flag.",
```

internal/output/mediator_test.go
```diff
@@ -65,7 +65,7 @@ func Test_mediatorValue(t *testing.T) {
 				"unstructured",
 				JSONFormatName,
 			},
-			jsonError{[]string{"json output not supported for message: unstructured"}, 1},
+			jsonError{[]string{"This command does not support the json output format. Please try again without that output flag"}, 1},
```
Naatan, line 68: Nit: If we're reusing localization it's better to add it to en-us.yaml.

internal/runners/packages/list.go
```diff
@@ -170,5 +178,16 @@ func newFilteredRequirementsTable(requirements []*gqlModel.Requirement, filter s
...
+	// Sort the rows.
+	less := func(i, j int) bool {
+		a := rows[i].Pkg
+		b := rows[j].Pkg
+		if strings.ToLower(a) < strings.ToLower(b) {
+			return true
+		}
+		return a < b
+	}
+	sort.Slice(rows, less)
```
Naatan, line 190: Looks like you can drop `.sortByPkg()` since you moved the logic here.

internal/runners/export/config/config.go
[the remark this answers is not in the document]
Naatan: Ah, guessing there are some limitations around using generics inside struct definitions, the same as with methods? Hopefully, they will address those limitations soon, I keep running into them as well. I feel like most situations in which I want generics still aren't supported.

Naatan (changes requested):
Almost there, looking good!

internal/runners/packages/list.go
mitchell-as: Yes, that was my intent. I don't see any `.sortByPkg()` calls anywhere.

internal/runners/export/config/config.go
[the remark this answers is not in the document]
mitchell-as: Yes, I ran into issues with using generics in structs. I was hoping for C++-style template classes, but sadly no. Just functions and methods it looks like.

pkg/cmdlets/errors/errors.go
mitchell-as: Not quite. I changed it to return a JSON error object directly (with a list of error strings) and changed the JSON `Error()` method to handle this case. It's possible to call JSON's `Error()` method directly from code (not just an error handler), so we still need to consider at least incoming `error`, `[]byte`, and `string` types.

internal/runners/packages/list.go
Naatan: Ah, I see now you removed the entire file. I was looking for a `sortByPkg` method removal and not finding it.. :p

internal/output/json.go
```diff
@@ -59,7 +59,7 @@ func (f *JSON) Fprint(writer io.Writer, value interface{}) {
 	f.cfg.OutWriter.Write([]byte(nul + "\n")) // Terminate with NUL character so consumers can differentiate between multiple output messages
 }
 
-type jsonError struct {
+type JsonError struct {
```
Naatan, line 62: Now that we're making this a public struct I feel more opinionated about the naming; can we name this StructuredError instead? That'll keep the naming consistent. 

The fact is we currently only have JSON output so by definition structured errors are json errors, but that might not always be the case. I'd rather we future-proof the naming and thus clearly mark the boundaries between json and structured output.

Naatan (changes requested):
Looks good! One last nitpick.

internal/output/json.go
mitchell-as: Okay, I've changed the name, but I left it in the output/json.go. I wasn't sure where exactly else it can/should go. Let me know if you have a preference.

I also removed the `[]string` possibility for `toStructuredError()` since the errors cmdlet returns a `StructuredError` directly.

Naatan (dismissed):

PR #2511, opened 2023-04-27
Update retry client status code handling

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1723" title="DX-1723" target="_blank">BugDX-1723</a>  retryhttp should not retry on server errors.
  </td></table>
  <br />

### DIFF
--- a/internal/retryhttp/client.go
+++ b/internal/retryhttp/client.go
@@ -183,27 +183,26 @@ func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, err
 		return false, ctx.Err()
 	}
 
-	if v, ok := err.(*url.Error); ok {
-		// Don't retry if the error was due to too many redirects.
-		if redirectsErrorRe.MatchString(v.Error()) {
-			return false, nil
+	if err != nil {
+		if v, ok := err.(*url.Error); ok {
+			// Don't retry if the error was due to too many redirects.
+			if redirectsErrorRe.MatchString(v.Error()) {
+				return false, nil
+			}
+
+			// Don't retry if the error was due to an invalid protocol scheme.
+			if schemeErrorRe.MatchString(v.Error()) {
+				return false, nil
+			}
+
+			// Don't retry if the error was due to TLS cert verification failure.
+			if _, ok := v.Err.(x509.UnknownAuthorityError); ok {
+				return false, nil
+			}
 		}
 
-		// Don't retry if the error was due to an invalid protocol scheme.
-		if schemeErrorRe.MatchString(v.Error()) {
-			return false, nil
-		}
-
-		// Don't retry if the error was due to TLS cert verification failure.
-		if _, ok := v.Err.(x509.UnknownAuthorityError); ok {
-			return false, nil
-		}
-	}
-
-	// Don't retry if the response is nil.
-	// This can happen when the request is cancelled via a client timeout.
-	if resp == nil {
-		return false, err
+		// The error is likely recoverable so retry.
+		return true, err
 	}
 
 	return isRetryableStatus(resp.StatusCode), nil
--- a/internal/retryhttp/client.go
+++ b/internal/retryhttp/client.go
@@ -179,8 +179,7 @@ func transport() http.RoundTripper {
 // retryPolicy is a modified version of retryablehttp.DefaultRetryPolicy to handle
 // status codes differently.
 func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
-	// do not retry on context.Canceled or context.DeadlineExceeded
-	if ctx.Err() != nil {
+	if ctx.Err() != nil || err == nil {
 		return false, ctx.Err()
 	}
 
@@ -201,7 +200,12 @@ func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, err
 		}
 	}
 
-	// The error is likely recoverable so retry.
+	// Don't retry if the response is nil.
+	// This can happen when the request is cancelled via a client timeout.
+	if resp == nil {
+		return false, err
+	}
+
 	return isRetryableStatus(resp.StatusCode), nil
 }
 
--- a/internal/retryhttp/client.go
+++ b/internal/retryhttp/client.go
@@ -1,16 +1,20 @@
 package retryhttp
 
 import (
+	"context"
+	"crypto/x509"
 	"errors"
 	"net"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/thoas/go-funk"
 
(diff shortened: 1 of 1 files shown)


internal/retryhttp/client.go
```diff
@@ -37,6 +41,44 @@ var (
...
+		// to indicate when the server is available to start processing
+		// request from client.
+		http.StatusTooManyRequests,
+		// The server is unwilling to risk
+		// processing a request that might be replayed.
+		http.StatusTooEarly,
+		// 5XX Status codes
+		// The server is currently unable to handle the request. Without
+		// any other information, the client may retry the request.
+		http.StatusInternalServerError,
```
Naatan, line 69: We debated this while filing the story; I don't want to retry on 500 status codes. I don't care what the reasoning out there is that people think retrying on 500 is appropriate, for our platform it is not appropriate. 500 means the server encountered an issue the client cannot resolve, and retrying is just going to needlessly hammer the server.

internal/retryhttp/client.go
```diff
@@ -133,3 +176,40 @@ func transport() http.RoundTripper {
...
+
+// retryPolicy is a modified version of retryablehttp.DefaultRetryPolicy to handle
+// status codes differently.
+func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	// do not retry on context.Canceled or context.DeadlineExceeded
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+
+	if err != nil {
```
Naatan, line 188: Nit: just return early if `err == nil`.

internal/retryhttp/client.go
```diff
@@ -37,6 +41,44 @@ var (
...
+
+	// A regular expression to match the error returned by net/http when the
+	// configured number of redirects is exhausted. This error isn't typed
+	// specifically so we resort to matching on the error string.
+	redirectsErrorRe = regexp.MustCompile(`stopped after \d+ redirects\z`)
+
+	// A regular expression to match the error returned by net/http when the
+	// scheme specified in the URL is invalid. This error isn't typed
+	// specifically so we resort to matching on the error string.
+	schemeErrorRe = regexp.MustCompile(`unsupported protocol scheme`)
```
Naatan, line 53: Aren't these system errors? Meaning they would be localized and these regexes would only match on English systems.

internal/retryhttp/client.go
```diff
@@ -133,3 +176,40 @@ func transport() http.RoundTripper {
...
+			if _, ok := v.Err.(x509.UnknownAuthorityError); ok {
+				return false, nil
+			}
+		}
+
+		// The error is likely recoverable so retry.
+		return true, nil
+	}
+
+	return isRetryableStatus(resp.StatusCode), nil
```
Naatan, line 210: Wouldn't err be non-nil for status codes over the 200 series? Meaning this wouldn't hit due to the err check above.

Naatan (changes requested):

internal/retryhttp/client.go
MDrakos: I should have noted that these regexs are also from the retry library that we are using. I grabbed them as the function I added to check retries is modifying the behavior of the default one. 

These string values are coming from the `net/http` package which I _believe_ isn't localized but also since we are statically building our binary I don't think they would be presented in different languages.

https://cs.opensource.google/go/go/+/master:src/net/http/client.go;l=812?q=%22stopped%20after%22&ss=go%2Fgo:src%2Fnet%2Fhttp%2F

https://cs.opensource.google/go/go/+/master:src/net/http/transport.go;l=560?q=%22unsupported%20protocol%20scheme%22&ss=go%2Fgo:src%2Fnet%2Fhttp%2F

internal/retryhttp/client.go
MDrakos: That makes sense and from my research, the consensus is that if we know enough about the server and 500 isn't valid to retry then we shouldn't. If this was something that we explicitly decided then it should have been an AC.

internal/retryhttp/client.go
MDrakos: I think you're right... if that's the case then I don't know if the default retry client was ever retrying on 500 errors 😕 

https://github.com/ActiveState/cli/blob/DX-1723/vendor/github.com/hashicorp/go-retryablehttp/client.go#L440

Naatan: No I'm wrong, seems this should be ok:

image

internal/retryhttp/client.go
```diff
@@ -181,32 +180,28 @@ func transport() http.RoundTripper {
 // status codes differently.
 func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
 	// do not retry on context.Canceled or context.DeadlineExceeded
-	if ctx.Err() != nil {
+	if ctx.Err() != nil || err == nil {
```
Naatan, line 183:
```suggestion
	if ctx.Err() != nil {
```

I was wrong about this, status codes over 2xx do not equal a non-nil error.

Naatan (changes requested):

internal/retryhttp/client.go
```diff
@@ -133,3 +175,36 @@ func transport() http.RoundTripper {
 	}
 	return cleanhttp.DefaultPooledTransport()
 }
+
+// retryPolicy is a modified version of retryablehttp.DefaultRetryPolicy to handle
+// status codes differently.
+func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	// do not retry on context.Canceled or context.DeadlineExceeded
+	if ctx.Err() != nil {
```
Naatan, line 183:
```suggestion
	if ctx.Err() != nil || err == nil {
```

My bad once again, I was miss-interpreting. This snippet and the rest of the error handling here requires a bit of a double take :p 

Without this 187 can throw a nil pointer panic.

Naatan (changes requested):

internal/retryhttp/client.go
MDrakos: The use of the `ok` variable in the type check should avoid a panic https://go.dev/play/p/EE8uZufw6u2

However, the unit tests surfaced a panic with the `nil` response. If the client has exceeded its timeout it will return an error without making the response. It also appears that this error is not a "checkable" `Context deadline exceeded` error so I've added an explicit check for a `nil` response which should cover more bases for us anyways. [This](https://go.dev/play/p/EE8uZufw6u2) is where the error is being returned.

internal/retryhttp/client.go
```diff
@@ -133,3 +175,39 @@ func transport() http.RoundTripper {
...
+			}
+
+			// Don't retry if the error was due to TLS cert verification failure.
+			if _, ok := v.Err.(x509.UnknownAuthorityError); ok {
+				return false, nil
+			}
+		}
+
+		// The error is likely recoverable so retry.
+		return true, err
```
MDrakos, line 205: This change is also lifted from the default retry policy of the retry client library.

Naatan (approved):

PR #2518, opened 2023-05-02
Replace multierr package with native go solution

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1736" title="DX-1736" target="_blank">TaskDX-1736</a>  Get rid of the multierr package in favour of native solution
  </td></table>
  <br />


internal/errs/errs_private_test.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			encoded := encodeErrorForJoin(tt.err)
+			if !reflect.DeepEqual(encoded, tt.want) {
+				t.Fatalf("got %v, want %v", encoded, tt.want)
+			}
+			// assert.Equalf(t, tt.want, encoded, "EncodeError(%v)", tt.err)
```
mitchell-as, line 74: You can probably remove this if the `reflect.DeepEqual()` call is sufficient.

internal/logging/logging.go
```diff
@@ -255,16 +255,7 @@ func writeMessageDepth(depth int, level string, msg string, args ...interface{})
...
-	errw := err
-	for {
-		errw = errors.Unwrap(errw)
-		if errw == nil {
-			break
-		}
-		errMsg += ": " + errw.Error()
-	}
-	fmt.Fprintf(os.Stderr, "Error writing log message: %s\n", errMsg)
+	fmt.Fprintf(os.Stderr, "Error writing log message: %s\n", errs.JoinMessage(err))
```
mitchell-as, line 258: Very elegant! This feels like one of those moments in tetris where you prepared everything for that 1x4 piece and finally got it.

internal/rtutils/rtutils.go
```diff
@@ -4,10 +4,22 @@ import (
 	"fmt"
 	"runtime"
 	"time"
-
-	"github.com/hashicorp/go-multierror"
 )
 
+type packedErrors struct {
```
mitchell-as, line 9: It might be worth adding a comment that this was duplicated in another package in order to avoid an import cycle, just like the previous instance.

mitchell-as (approved):
This looks great. Two tiny nits for you to consider.

internal/errs/errs_private_test.go
Naatan: Ahh yep, thanks for catching it! I was testing both approaches and forgot to remove this one.

internal/logging/logging.go
Naatan: Lol! To me it felt more like "Why did I do it this way the first time around?!?!" :p

mitchell-as (approved):

PR #2520, opened 2023-05-02
Do not report "blocked access" to Rollbar

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1609" title="DX-1609" target="_blank">StoryDX-1609</a>  Don't report "blocked access" to rollbar
  </td></table>
  <br />


internal/locale/locales/en-us.yaml
```diff
@@ -2043,3 +2043,5 @@ runtime_limit_reached_action:
   other: Upgrade
 pjfile_deprecation_entry:
   other: " - '[ACTIONABLE]{{.V0}}[/RESET]' located at byte [ACTIONABLE]{{.V1}}[/RESET]"
+err_country_blocked:
+  other: Your country is blocked
```
mitchell-as, line 2047: Oof. That's a bit harsh. We might want to soften the language a bit, but I'm not sure how.

internal/rollbar/rollbar.go
```diff
@@ -23,9 +23,23 @@ type config interface {
 	Closed() bool
 }
 
-var currentCfg config
+type doNotReport []string
 
-var reportingDisabled bool
+func (d *doNotReport) Add(msg string) {
```
mitchell-as, line 28: If you accidentally pass `locale.T("does_not_exist")`, does it return an empty string? If so, would that end up blocking all rollbar reports? If so, we might want to ignore empty strings here and possibly report them to rollbar so we can fix.

mitchell-as (changes requested):
Nice solution :) I wrote some thoughts to consider.

internal/locale/locales/en-us.yaml
MDrakos: Agreed it is a bit rough, but it is from the ACs of the story. @Naatan do we want to consider a more friendly error message?

internal/rollbar/rollbar.go
MDrakos: Thank you for catching this! It should be addressed now and I've added some tests to verify.

mitchell-as (dismissed):

internal/locale/locales/en-us.yaml
Naatan: @rawktron ^

rawktron: "This service is not available in your region."

mitchell-as (approved):

PR #2524, opened 2023-05-03
Fix error unit test failures

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1810" title="DX-1810" target="_blank">BugDX-1810</a>  Error unit test failures on master
  </td></table>
  <br />

### DIFF
--- a/internal/output/json_test.go
+++ b/internal/output/json_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/ActiveState/cli/internal/locale"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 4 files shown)


internal/output/json_test.go
```diff
@@ -170,21 +171,21 @@ func TestJSON_Error(t *testing.T) {
 		expectedErr string
 	}{
 		{
-			"error string",
-			args{errors.New("hello")},
-			`{"errors":["hello"],"code":1}` + "\x00\n",
+			"localized error",
+			args{locale.NewError("", "hello")},
```
Naatan, line 175: `errs.New()` would be invalid as we only want to expose localized errors. As it stands if you pass a non-localized error it will panic if run from a local build, and log to rollbar if run from a CI build, whilst reporting the topmost error. That's probably not worth testing at this level though.

mitchell-as (approved):

PR #2526, opened 2023-05-03
Fix JSON output not printed

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1809" title="DX-1809" target="_blank">BugDX-1809</a>  `--output json`: command is not executed if the `-o json` flag is used.
  </td></table>
  <br />


cmd/state/internal/cmdtree/intercepts/messenger/messenger.go
```diff
@@ -26,7 +26,7 @@ func (m *Messenger) Interceptor(next captain.ExecuteFunc) captain.ExecuteFunc {
 	return func(cmd *captain.Command, args []string) error {
 		if m.out.Type() != output.PlainFormatName && m.out.Type() != output.SimpleFormatName {
```
mitchell-as, line 27: We can use the new `m.out.Type().IsStructured()` method here.

mitchell-as (changes requested):

PR #2532, opened 2023-05-08
Added missing dimensions to legacyDimensionMap.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1806" title="DX-1806" target="_blank">TaskDX-1806</a>  Add new dimensions to legacyDimensionMap
  </td></table>
  <br />


Naatan (changes requested):
In reviewing I think I found 2 more missing dimensions, see the docs here:

https://docs.google.com/document/d/1GiofKl6YYt9X70nBz93nRXtibChU3zXyFG7kavOJXGA/edit

Seems 22 and 23 are already "reserved", but the dimensions documented there aren't hooked up either.

I think we should add those as scope for this story, as it is about missing dimensions. Since they haven't been put into use yet I'll let you decide what numbers you wanna use (long as it's 22+) and I'll update the docs accordingly.

Naatan (approved):

PR #2537, opened 2023-05-09
Ensure API client headers are set

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1787" title="DX-1787" target="_blank">BugDX-1787</a>  State Tool should send its User Agent string to solver request
  </td></table>
  <br />

### DIFF
--- a/internal/gqlclient/gqlclient.go
+++ b/internal/gqlclient/gqlclient.go
@@ -3,9 +3,11 @@ package gqlclient
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"time"
 
+	"github.com/ActiveState/cli/internal/condition"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/profile"
@@ -53,7 +55,11 @@ func NewWithOpts(url string, timeout time.Duration, opts ...graphql.ClientOption
 }
 
 func New(url string, timeout time.Duration) *Client {
-	return NewWithOpts(url, timeout, graphql.WithHTTPClient(api.NewHTTPClient()))
+	httpClient := api.NewHTTPClient()
+	if condition.InUnitTest() {
+		httpClient = http.DefaultClient
+	}
+	return NewWithOpts(url, timeout, graphql.WithHTTPClient(httpClient))
 }
 
 // EnableDebugLog turns on debug logging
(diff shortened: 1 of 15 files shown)


pkg/platform/errors.go
```diff
@@ -0,0 +1,13 @@
...
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/rollbar"
+)
+
+type ErrCountryBlocked struct{ *locale.LocalizedError }
+
+func NewCountryBlockedError() *ErrCountryBlocked {
+	rollbar.DoNotReportMessages.Add(locale.T("err_country_blocked"))
+	return &ErrCountryBlocked{LocalizedError: locale.NewInputError("err_country_blocked")}
+}
```
MDrakos, line 13: This was moved from it's original location so I could import `retryhttp` in the `api` package without an import cycle.

internal/gqlclient/gqlclient.go
```diff
@@ -53,7 +55,11 @@ func NewWithOpts(url string, timeout time.Duration, opts ...graphql.ClientOption
 }
 
 func New(url string, timeout time.Duration) *Client {
-	return NewWithOpts(url, timeout, graphql.WithHTTPClient(retryhttp.DefaultClient.StandardClient()))
+	httpClient := api.NewHTTPClient()
+	if condition.InUnitTest() {
+		httpClient = http.DefaultClient
+	}
+	return NewWithOpts(url, timeout, graphql.WithHTTPClient(httpClient))
```
MDrakos, line 62: For some reason, the `retryhttp` client is having trouble connecting with the `platform.activestate.testing.tld` URL in unit tests. I spent some time trying to determine why and then just decided to go with this fix.

Naatan: Wouldn't this be better addressed inside `api.NewHTTPClient()`?

pkg/platform/errors.go
Naatan: Are you sure this is still relevant? I don't see this being called from `retryhttp`.

Naatan (changes requested):

pkg/platform/errors.go
MDrakos: It's being used in a different file but same package: https://github.com/ActiveState/cli/blob/DX-1787/internal/retryhttp/roundtripper.go#L33

internal/gqlclient/gqlclient.go
MDrakos: Yes, that's a much better place for it.

MDrakos:
This will also fix https://activestatef.atlassian.net/browse/DX-1825

Naatan (approved):

PR #2538, opened 2023-05-10
Match found package/bundle's letter case when installing.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1816" title="DX-1816" target="_blank">BugDX-1816</a>  BUNDLES: `search <bundle>` and `install/uninstall <bundle>`  to behave consistently
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -174,6 +174,10 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 			}
 			return locale.WrapInputError(err, "package_ingredient_alternatives", "", requirementName, strings.Join(suggestions, "\n"))
 		}
+		if name := packages[0].Ingredient.Name; name != nil && requirementName != *name {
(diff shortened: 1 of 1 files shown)


internal/runbits/requirements/requirements.go
```diff
@@ -174,6 +174,10 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 			}
 			return locale.WrapInputError(err, "package_ingredient_alternatives", "", requirementName, strings.Join(suggestions, "\n"))
 		}
+		if name := packages[0].Ingredient.Name; name != nil && requirementName != *name {
```
MDrakos, line 177: This makes sense, you could add an extra `strings.EqualFold` check to ensure that `requirementName` and `name` are actually the same in a case insensitive way... though `SearchIngredientsStrict` _should_ cover that. This would just be another check to ensure they actually match.

MDrakos (approved):

PR #2543, opened 2023-05-11
Use custom home dir for integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1785" title="DX-1785" target="_blank">TaskDX-1785</a>  We use a custom home dir for our integration tests
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/test/integration/installer_int_test.go
+++ b/cmd/state-installer/test/integration/installer_int_test.go
@@ -274,6 +274,10 @@ func (suite *InstallerIntegrationTestSuite) TestInstallerOverwriteServiceApp() {
 }
 
 func (suite *InstallerIntegrationTestSuite) SetupRCFile(ts *e2e.Session) {
+	if runtime.GOOS == "windows" {
+		return
+	}
+
 	cfg, err := config.New()
 	suite.Require().NoError(err)
 
--- a/cmd/state-installer/test/integration/installer_int_test.go
+++ b/cmd/state-installer/test/integration/installer_int_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/environment"
 	"github.com/ActiveState/cli/internal/fileutils"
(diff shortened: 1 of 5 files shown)


cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -284,7 +303,12 @@ func (suite *InstallerIntegrationTestSuite) AssertConfig(ts *e2e.Session) {
 			fname = ".zshrc"
 		}
 
-		bashContents := fileutils.ReadFileUnsafe(filepath.Join(homeDir, fname))
+		rcFile := filepath.Join(homeDir, fname)
+		if fileutils.FileExists(filepath.Join(ts.Dirs.HomeDir, filepath.Base(fname))) {
```
mitchell-as, line 307: Could we just test `fname` directly (instead of `filepath.Base(fname)`) and remove all usage of `homeDir` from line 295? I don't think we want to have any mention of the real home dir for fear of confusion or accidental contamination.

test/integration/deploy_int_test.go
```diff
@@ -155,9 +155,11 @@ func (suite *DeployIntegrationTestSuite) TestDeployPython() {
 		suite.T().Skipf("Skipping DeployIntegrationTestSuite when not running on CI, as it modifies bashrc/registry")
 	}
 
-	ts := e2e.New(suite.T(), false)
+	ts := e2e.New(suite.T(), true)
```
mitchell-as, line 158: Is this intentional? Or was it a residual change from testing/debugging?

test/integration/deploy_int_test.go
```diff
@@ -313,6 +333,10 @@ func (suite *DeployIntegrationTestSuite) AssertConfig(ts *e2e.Session, targetID
 		rcFile, err := subshell.RcFile()
 		suite.Require().NoError(err)
 
+		if fileutils.FileExists(filepath.Join(ts.Dirs.HomeDir, filepath.Base(rcFile))) {
```
mitchell-as, line 336: Note there is no mention of `user.HomeDir()` like in the previous instance. This is what I'd prefer to see.

mitchell-as (changes requested):
Looks like you're on the right path here.

Would you please also enable the installer, service, and deploy integration tests? You changed them and I'm not certain that the critical label covers them.

cmd/state-installer/test/integration/installer_int_test.go
MDrakos: I decided to go the other direction with this and unify the `AssertConfig` functions to rely on the subshell mechanic. That way the integration tests aren't assuming what shell they are running on. If you would still prefer we go the other way let me know.

test/integration/deploy_int_test.go
MDrakos: Whoops, that's a leftover from debugging.

test/integration/deploy_int_test.go
MDrakos: The other instance is still using `user.HomeDir()` under the hood, at least for the bash implementation of the `subshell`.

mitchell-as (approved):
Nice!

PR #2544, opened 2023-05-12
Remove go version from Target & Verify job

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1744" title="DX-1744" target="_blank">TaskDX-1744</a>  Drop Go version from job names
  </td></table>
  <br />

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -36,7 +36,7 @@ jobs:
         name: Install Go
         uses: actions/setup-go@v3
         with:
-          go-version: ${{ matrix.go-version }}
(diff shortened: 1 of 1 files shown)


daved (dismissed):

Naatan (changes requested):
This seems to drop the go version as a requirement. Whereas the story is about dropping it from the job name.

This is likely happening due to us using the "matrix" strategy, which I guess makes it try to use a unique name based on what we put into the matrix. In this case we don't need a matrix at all so the PR should probably just change the go-version input to be send directly to the job rather than through the matrix.

Naatan (approved):

PR #2547, opened 2023-05-15
Fix nightly failures due to home dir change

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1837" title="DX-1837" target="_blank">BugDX-1837</a>  Nightlies: PATH does not have your project in it
  </td></table>
  <br />

### DIFF
--- a/test/integration/use_int_test.go
+++ b/test/integration/use_int_test.go
@@ -144,7 +144,7 @@ func (suite *UseIntegrationTestSuite) TestReset() {
 	rcFilePath := filepath.Join(ts.Dirs.HomeDir, filepath.Base(rcfile))
 	if runtime.GOOS != "windows" && fileutils.FileExists(rcFilePath) {
 		suite.NoError(err)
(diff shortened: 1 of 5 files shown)


test/integration/use_int_test.go
```diff
@@ -139,9 +141,10 @@ func (suite *UseIntegrationTestSuite) TestReset() {
 	cfg, err := config.New()
 	suite.NoError(err)
 	rcfile, err := subshell.New(cfg).RcFile()
-	if runtime.GOOS != "windows" && fileutils.FileExists(rcfile) {
+	rcFilePath := filepath.Join(ts.Dirs.HomeDir, filepath.Base(rcfile))
```
daved, line 144: Is there some way to have the subshell use the desired path for the rc file rather than reconstructing it after copying contents? I think it's less fragile to make the subshell type configurable and minimize special logic in the tests themselves.

daved (changes requested):

daved (approved):

PR #2549, opened 2023-05-17
Mature buildscript (local order) representation

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1853" title="DX-1853" target="_blank">StoryDX-1853</a>  Mature the local order PR
  </td></table>
  <br />


pkg/platform/api/graphql/model/buildplanner/buildScript.go
```diff
@@ -35,57 +38,58 @@ func (o Operation) String() string {
 // TODO: We will likely need some sort of parser or other solution for the build graph
 func NewBuildScript() *BuildScript {
 	return &BuildScript{
-		Let: LetStatement{
-			Runtime: Runtime{
-				SolveLegacy: SolveLegacy{
-					Requirements: []Requirement{},
+		Let: &LetStatement{
```
mitchell-as, line 41: This will all be replaced by the Python-like build script implementation.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -0,0 +1,96 @@
+package buildscript
```
mitchell-as, line 1: Most of this will be replaced by the Python-like build script implementation.

internal/runbits/buildscript/buildscript.go
```diff
@@ -0,0 +1,86 @@
...
+			Description:  locale.Tl("buildscript_commit_description", "Update project due to build script change."),
+			Script:       script,
+		})
+		if err != nil {
+			return errs.Wrap(err, "Could not update project to reflect build script changes.")
+		}
+		newCommitID := strfmt.UUID(commit.CommitID)
+		commitID = &newCommitID
+
+		script, err = getRemoteBuildScript(proj, commitID, auth)
```
mitchell-as, line 68: This may not be necessary if the platform doesn't change the given script?

mitchell-as:
The test failures do not appear to have anything to do with this PR, so I think it's ready for its initial review.

internal/runbits/buildscript/buildscript.go
MDrakos: No, it shouldn't be. Can you verify that the platform won't change the script without making the request though? This will require more than just checking the requirements. We'll neet to ensure the timestamp doesn't change as well.

internal/runbits/buildscript/buildscript.go
```diff
@@ -0,0 +1,92 @@
...
+// pull in the new build script. Otherwise, if there are local build script changes, create a new
+// commit with them in order to update the remote one.
+func Sync(proj *project.Project, commitID *strfmt.UUID, out output.Outputer, auth *authentication.Auth) error {
+	logging.Debug("Synchronizing local build script")
+	file, err := buildscript.Get(proj.Dir())
+	if err != nil && !buildscript.IsDoesNotExistError(err) {
+		return errs.Wrap(err, "Could not get local build script")
+	}
+
+	script, err := getRemoteBuildScript(proj, commitID, auth)
```
MDrakos, line 36: We should update the naming of things in this PR. The BuildScript refers to the local order file whereas the BuildExpression is what the buildplanner sends and receives. So this function should be `getBuildExpression`

pkg/platform/api/graphql/model/buildplanner/buildScript.go
```diff
@@ -35,57 +38,58 @@ func (o Operation) String() string {
...
 
 type Solve struct {
-	BuildFlags    []string      `json:"build_flags,omitempty"`
-	CamelFlags    []string      `json:"camel_flags,omitempty"`
-	Platforms     []string      `json:"platforms"`
-	SolverVersion string        `json:"solver_version,omitempty"`
-	AtTime        string        `json:"at_time"`
-	Requirements  []Requirement `json:"requirements"`
+	BuildFlags    []string       `json:"build_flags,omitempty" yaml:"build_flags,omitempty"`
+	CamelFlags    []string       `json:"camel_flags,omitempty" yaml:"camel_flags,omitempty"`
```
MDrakos, line 68: These should still be included even if empty. See the build expression [here](https://docs.google.com/document/d/1WlnbG20q0Lq3tTKsbpZB5JsK_u9qcg1x4iG9A1AsMUo/edit#heading=h.skrjork327yz) for reference

MDrakos (changes requested):
Looking good to me. Just a few comments about naming.

internal/runbits/buildscript/buildscript.go
mitchell-as: I cannot verify that, and even so, there's no guarantee that it will not change in the future. Perhaps it's best to do this "just in case".

internal/runbits/buildscript/buildscript.go
mitchell-as: I've done some more thorough renaming here and elsewhere. Thanks for the feedback.

internal/runbits/refresh.go
```diff
@@ -15,7 +16,11 @@ import (
 
 // RefreshRuntime should be called after runtime mutations.
 func RefreshRuntime(auth *authentication.Auth, out output.Outputer, an analytics.Dispatcher, proj *project.Project, commitID strfmt.UUID, changed bool, trigger target.Trigger, svcm *model.SvcModel) (rerr error) {
-	target := target.NewProjectTarget(proj, &commitID, trigger)
+	err := buildscript.Sync(proj, &commitID, out, auth)
```
MDrakos, line 19: Thinking on this some more I would expect to see this function call, or something similar, more often. For example: If the user modifies their local order file and then runs `state activate` I would expect the runtime to be updated. However, I don't believe `RefreshRuntime` will be called in that case, so the changes to the local order file will not be picked up. I could be missing something but I just want to call it out so we can ensure we've got that covered.

MDrakos (changes requested):

internal/runbits/refresh.go
mitchell-as: `state activate` calls `runtime.NewFromProject()`, which syncs the build script. I did a search for all instances we call `runtime.New()` and this PR calls `buildscript.Sync()` for all of them except `state _prepare` and `state deploy`. I'm not sure if it's necessary for the former, and deployments are not projects, so syncing doesn't seem necessary there.

Does this assuage your concerns? Or am I not looking in the right place?

MDrakos: Great! I just wanted to make sure that we've covered all the uses of the runtime and you have done that.

PR #2552, opened 2023-05-18
State Upload

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1754" title="DX-1754" target="_blank">StoryDX-1754</a>  Implement `state upload` 
  </td></table>
  <br />


internal/captain/flags.go
```diff
@@ -0,0 +1,181 @@
...
+	if strings.Contains(s, "@") {
+		v := strings.Split(s, "@")
+		u.Version = strings.TrimSpace(v[1])
+		s = v[0]
+	}
+	if strings.Index(s, "/") == -1 {
+		return fmt.Errorf("invalid package name format: %s (expected '<namespace>/<name>[@version]')", s)
+	}
+	v := strings.Split(s, "/")
+	u.Namespace = strings.TrimSpace(strings.Join(v[0:len(v)-1], "/"))
```
daved, line 149: Maybe it's worth having a story to centralize a namespace type.

internal/runners/uploadingredient/uploadingredient.go
```diff
@@ -0,0 +1,343 @@
...
+	primer.Projecter
+	primer.Prompter
+}
+
+func New(prime primeable) *Runner {
+	client := gqlclient.NewWithOpts(
+		api.GetServiceURL(api.ServiceBuildPlan).String(), 0,
+		graphql.WithHTTPClient(http.DefaultClient),
+		graphql.UseMultipartForm(),
+	)
```
daved, line 63: I'm a bit surprised we're not passing around the gql client from higher in the stack into runners. If we are, why is there special setup here?

internal/runners/uploadingredient/uploadingredient.go
```diff
@@ -0,0 +1,343 @@
...
+	} else if reqVars.Namespace == "" {
+		reqVars.Namespace = model.NewSharedNamespace(r.project.Owner()).String()
+	}
+
+	// Validate namespace
+	prefix := r.project.Owner() + "/"
+	if !strings.HasPrefix(reqVars.Namespace, prefix) {
+		return locale.NewInputError("err_uploadingredient_namespace_invalid_org",
+			"Namespace should be prefixed '[ACTIONABLE]{{.V0}}[/RESET]', received: '[ACTIONABLE]{{.V1}}[/RESET]'.", prefix, reqVars.Namespace)
+	}
```
daved, line 109: This seems to not take care of non-personal orgs. Is that intentional? Org restrictions feel like something the backend should handle.

internal/runners/uploadingredient/uploadingredient.go
```diff
@@ -0,0 +1,343 @@
...
+		reqVars.Name = params.Name
+	} else if reqVars.Name == "" {
+		reqVars.Name = filepath.Base(params.Filepath)
+	}
+
+	ts := time.Now()
+	ingredients, err := model.SearchIngredientsStrict(reqVars.Namespace, reqVars.Name, true, false, &ts)
+	if err != nil && !errs.Matches(err, &model.ErrSearch404{}) { // 404 means namespace not found, which is fine
+		return locale.WrapError(err, "err_uploadingredient_search", "Could not search for ingredient")
+	}
```
daved, line 122: If this is done to check for a namespace, it feels super janky and like a story should be filed somewhere for a better solution. If it's to gather ingredients to edit one, taking the first in the slice seems possibly concerning. Also, can this chunk be moved inside of the `if params.Edit` block?

daved (changes requested):

internal/runners/uploadingredient/uploadingredient.go
Naatan: What do you mean by this? I don't see anything that implies either personal or non-personal.

daved: Ah, yeah, nevermind. I read this to mean that only the current user org could be uploaded to, but it's the org that contains the specific project that is being uploaded to. Sorry about that.

internal/gqlclient/gqlclient.go
```diff
@@ -1,38 +1,101 @@
...
-
-	"github.com/ActiveState/cli/internal/singleton/uniqid"
+	"github.com/pkg/errors"
 )
 
-type Request interface {
+type File struct {
+	Field string
+	Name  string
+	R     io.Reader
```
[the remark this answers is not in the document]
Naatan: This isn't the owner of the file pointer, it just consumes it. The closing mechanic is owned by

https://github.com/ActiveState/cli/blob/DX-1754/pkg/platform/api/graphql/request/publish.go#L147-L147

And it's closed a few lines below that.

internal/runners/uploadingredient/uploadingredient.go
Naatan: I don't follow how this is janky. Not finding any ingredients is a legitimate result that can happen, and a 404 is what this API responds with in that case.

Moving it inside the edit conditional isn't an option, as the result is used in the "else" block of that conditional.

internal/runners/uploadingredient/uploadingredient.go
daved: Why hold off on structuring the BP client into the application?

internal/runners/uploadingredient/uploadingredient.go
daved: https://softwareengineering.stackexchange.com/a/203493

internal/runners/uploadingredient/uploadingredient.go
Naatan: You mean in terms of future sprints? Fair; I've opened a ticket: https://activestatef.atlassian.net/browse/DX-1907

In terms of this ticket it feels significantly out of scope.

internal/runners/uploadingredient/uploadingredient.go
Naatan: That's because when the API was developed the namespaces were predetermined, and so supplying a namespace that didn't exist was likely a fair case for a 404.

Now that namespaces have become more variable we could probably revisit that, but then we also need to consider backward compatibility.

Frankly I just don't see this being worth the time investment. It might not fully respect the standards that are out there, but few API's do and spending time JUST to satisfy a standard is in my opinion a waste.

daved (dismissed):

PR #2553, opened 2023-05-19
Initial Python-style buildscript read/write support.

mitchell-as:
Translating to and from build expressions is unsupported until the model/AST is finalized.


pkg/platform/runtime/buildscript/buildexpression.go
```diff
@@ -0,0 +1,151 @@
+package buildscript
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/rtutils/p"
+	buildexpression "github.com/ActiveState/cli/pkg/buildexpression/parser"
```
mitchell-as, line 10: This currently only exists on #2554.

pkg/platform/runtime/buildscript/buildexpression.go
```diff
@@ -0,0 +1,151 @@
+package buildscript
```
mitchell-as, line 1: I recommend reviewing this file last, as it ties together build scripts with the work-in-progress build expressions on the Platform side.

internal/runbits/buildscript/buildscript.go
```diff
@@ -48,7 +48,7 @@ func Sync(proj *project.Project, commitID *strfmt.UUID, out output.Outputer, aut
 		// For now, if commitID is given, a mutation happened, so prefer the remote build script.
 		// Otherwise, prefer local changes.
 		if commitID == nil {
-			script = file.Script
+			//script = file.Script
```
mitchell-as, line 51: Will be hooked up in https://activestatef.atlassian.net/browse/DX-1858

pkg/platform/runtime/buildscript/buildexpression.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/ActiveState/cli/internal/rtutils/p"
+	buildexpression "github.com/ActiveState/cli/pkg/buildexpression/parser"
+	"github.com/ActiveState/cli/pkg/platform/api/graphql/model/buildplanner"
+)
+
+func FromBuildExpression(tree *buildexpression.Tree) *Script {
+	script := &Script{&Let{}, &In{}}
+	// tree.Root.Children() contains the following 7 nodes: { let : <binding> , in <in expr>
+
+	letBinding := tree.Root.Children()[1].Children()[3]
```
mitchell-as, line 18: Note that this and a lot of other index operations with magic numbers are due to how the parser produces the AST. It is subject to change until finalized.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -1,102 +1,119 @@
...
-	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/pkg/platform/api/graphql/model/buildplanner"
-	"gopkg.in/yaml.v2"
+	"bytes"
+	"io"
 )
 
-type DoesNotExistError struct{ error }
+type Script struct {
+	Let *Let `parser:"'let' ':' @@"`
```
mitchell-as, line 9: These struct tags utilize https://github.com/alecthomas/participle, a Go package for dead-simple parsing into structures.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -1,102 +1,119 @@
...
-	return file.Update(script)
+type In struct {
+	FuncCall *FuncCall `parser:"@@"`
+	Name     *string   `parser:"| @Ident"`
 }
 
-func create(dir string, script *model.BuildScript) (*File, error) {
-	if script == nil {
-		script = model.NewBuildScript()
+func (s *Script) Write(w io.Writer) {
```
mitchell-as, line 42: As noted in the technical spec, we do not care about preserving whitespace.

pkg/platform/runtime/buildscript/file.go
```diff
@@ -0,0 +1,113 @@
...
+		file, err = create(dir, nil)
+		if err != nil {
+			return errs.Wrap(err, "Could not create build script")
+		}
+	}
+	return file.update(script)
+}
+
+func create(dir string, script *model.BuildScript) (*File, error) {
+	return nil, errs.New("Writing not supported yet")
```
mitchell-as, line 72: This will be supported in https://activestatef.atlassian.net/browse/DX-1858

pkg/platform/runtime/buildscript/file.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	path := filepath.Join(dir, constants.BuildScriptFileName)
+	logging.Debug("Creating build script: %s", path)
+	if err := fileutils.WriteFile(path, data); err != nil {
+		return nil, errs.Wrap(err, "Could not write build script to file")
+	}
+
+	//return &File{path, script}, nil
+	return nil, errs.New("Writing not supported yet")
```
mitchell-as, line 90: This will be supported in https://activestatef.atlassian.net/browse/DX-1858

pkg/platform/runtime/buildscript/file.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (o *File) update(script *model.BuildScript) error {
+	if script == nil {
+		return errs.New("Build script to write is nil")
+	}
+	if o.Script != nil && o.Script.Equals(script) {
+		return nil
+	}
+
+	//o.Script = script
```
mitchell-as, line 111: This will be supported in https://activestatef.atlassian.net/browse/DX-1858

pkg/platform/runtime/buildscript/json.go
```diff
@@ -0,0 +1,91 @@
...
+		writeString(w, "$"+*s.In.Name)
+	}
+	w.Write([]byte(`}}`))
+}
+
+func (a *Assignment) WriteJson(w io.Writer) {
+	writeString(w, a.Key)
+	w.Write([]byte(":"))
+	a.Value.WriteJson(w)
+}
```
daved, line 42: Same as with the Write method, implementing different interfaces should make for more flexible usage. In this case, implementing json.Marshaler (https://pkg.go.dev/encoding/json#Marshaler) should result in much simpler handling when getting JSON output from a type. This might mean that the marshaling requires a call to the json package, but that addition should be minor compared to the reductions.

pkg/platform/runtime/buildscript/json.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (v *Value) WriteJson(w io.Writer) {
```
daved, line 44: Is this a pattern suggested by the participle package? If not, I suggest looking into whether ths "value" field can be an interface type where each of the potential types implement that interface. Based on my suggestions, something like:
```
type Valuer interface {
    fmt.Stringer
    json.Marshaler
}
```
Or, roughly based on existing methods:
```
type Valuer interface {
    io.WriterTo
    WriteJSONTo(io.Writer) (int, error)
}
```

daved (changes requested):
I do feel strongly, at the moment, that the changes I've suggested would help us keep things more standard. However, if the participle package is prescribing this sort of design, feel free to do whatever is best in your judgment. The code is readable and looks well-considered (Great work!). Similarly, if it's not worth the time now and you agree with a good portion of the suggestions here, it might be worth filing a story for a refactor that points to this review.

go.mod
```diff
@@ -74,6 +74,8 @@ require (
 	modernc.org/sqlite v1.11.2
 )
 
+require github.com/alecthomas/participle/v2 v2.0.0 // indirect
+
```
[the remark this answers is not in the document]
mitchell-as: I ran `go mod vendor`. If you have any other suggestions, please let me know. I don't know anything about go modules and vendoring.

pkg/platform/runtime/buildscript/testdata/complex.lo
```diff
@@ -0,0 +1,23 @@
...
+                name="language/perl"
+            }
+        ],
+        platforms=["12345"]
+    )
+
+in:
+   merge(
+        win_installer(win_runtime),
+        tar_installer(linux_runtime)
```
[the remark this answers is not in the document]
mitchell-as: Oh it's certainly possible. If I remember I'll bring it up in stand if the syntax should be solidified now, or if it's a WIP.

pkg/platform/runtime/buildscript/testdata/basic.lo
```diff
@@ -0,0 +1,22 @@
+let:
+  runtime = solve(
+    platforms = ["linux", "windows"],
```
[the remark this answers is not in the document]
mitchell-as: Unfortunately, I don't know. The design doc does not give any hint as to what the fields and parameters actually mean.

pkg/platform/runtime/buildscript/file.go
```diff
@@ -0,0 +1,113 @@
...
+func IsDoesNotExistError(err error) bool {
+	return errs.Matches(err, &DoesNotExistError{})
+}
+
+type File struct {
+	Path   string
+	Script *Script
+}
+
+func get(path string) (*File, error) {
```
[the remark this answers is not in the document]
mitchell-as: I'll make `NewScript()`, but I don't see the utility in `NewFile()` because we're only supposed to have one build script per project, like `project.Get()`.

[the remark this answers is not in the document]
daved: Maybe @Naatan will correct me... Get is a workaround for legacy reliance on globals. We intend to drop that in favor of structuring dependencies throughout the application. That is, project.Project would be called higher in the stack, and passed around where it's needed. Keep in mind that Get does forward to project.New.

The official docs suggest the structuring approach, and also to avoid the "get" term in naming: https://go.dev/doc/effective_go#package-names. It's worth adding that the docs recommend a more package-oriented approach to design than we tend to take at this time. It does result in more files and packages, but the code tends to be more modular.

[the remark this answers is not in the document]
Naatan: Get() in the project package is definitely legacy functionality, because it relies on globals. Seems this isn't relying on globals though, so I don't see the issue. We don't have a ban on the word "get", just on globals.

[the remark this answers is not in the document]
mitchell-as: Very well, I'll use NewFile as well then.

pkg/platform/runtime/buildscript/json.go
mitchell-as: Thanks for the pointer. I'll use the marshaler interface.

pkg/platform/runtime/buildscript/json.go
mitchell-as: I just wanted a way to efficiently concatenate a lot of strings piecemeal, and `bytes.Buffer` coupled with an `io.Writer` seemed to fit the bill. I've implemented your `String()` and `MarshalJSON()` interface ideas, so this should not be necessary anymore.

pkg/platform/runtime/buildscript/buildexpression.go
daved: It might be nice to include a link to some doc and/or a comment to clarify the basis for the magic numbers. To make them non-magical, consider defining some constants where the names act as clarification.

go.mod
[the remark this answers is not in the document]
daved: Try moving the path/version into the larger require directive block, dropping the `// indirect` portion. Then run `go mod tidy`. Whatever is the result should be good to go.

pkg/platform/runtime/buildscript/buildexpression.go
mitchell-as: After discussion, we're tossing out the custom build expression parser and lexer in favor of unmarshaling the build expression JSON directly and manipulating it. This file has been completely rewritten.

pkg/platform/runtime/buildscript/file.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func newFile(path string) (*File, error) {
+	data, err := fileutils.ReadFile(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, &DoesNotExistError{errs.New("Build script '%s' does not exist", path)}
+		}
+		return nil, errs.Wrap(err, "Could not read build script")
+	}
```
daved, line 35: I find it preferable to have disk ops isolated away from testable logic. In other words, letting disk operations be tested by tests that are more integration-focused. Consider moving this disk op into NewFile, and then changing newFile to be `newFile(data []byte) (*File, error)`. NewFile will probably then need to receive a full path, and the path construction would occur within UpdateOrCreate to be passed into both NewFile and file.update. This might also mean that the json test data would live in .go files as variables. Again, that would be more normal, in my opinion/experience, for type/package tests of lower level logic. The e2e tests would be the tests which rely on disk ops and json files.

daved (changes requested):

pkg/platform/runtime/buildscript/buildexpression.go
```diff
@@ -0,0 +1,195 @@
...
+	if m, ok := inValue.(map[string]interface{}); ok {
+		f, err := newFuncCall(m)
+		if err != nil {
+			return nil, errs.Wrap(err, "'in' object is not a function call")
+		}
+		in.FuncCall = f
+
+	} else if s, ok := inValue.(string); ok {
+		in.Name = p.StrP(strings.TrimPrefix(s, "$"))
+	}
```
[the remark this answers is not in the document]
mitchell-as: That results in double-typing the types.

```
switch inValue.(type) {
case map[string]interface{}:
  f, err := newFuncCall(inValue.(map[string]interface{}))
```

I did forget to write an `else` block with an error though.

pkg/platform/runtime/buildscript/json.go
```diff
@@ -0,0 +1,72 @@
+package buildscript
```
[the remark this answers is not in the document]
mitchell-as: If you have a concrete suggestion for me, I'm open to hearing it. Otherwise it's just vague hand-waving to me. I've chosen to break up the buildscript package into logical files dealing with buildscripts, buildexpressions, and json. I didn't want to have one monolithic buildscript.go file.

pkg/platform/runtime/buildscript/file.go
mitchell-as: `newFile()` still needs a path in order to construct a `File{}`, so your suggestion will not work. However, I've changed the tests to use `NewScript()` instead, and inlined the buildscripts and buildexpressions as you suggested. Integration tests in future PRs will do e2e testing of `NewFile()`, so it can be revisited then if necessary.

pkg/platform/runtime/buildscript/json.go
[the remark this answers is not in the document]
daved: The basic suggestion would be to have separate files for Assignment, Value, File, etc. Again, organizing by type rather than purpose (where JSON is a purpose). I'm not directly suggesting that. Rather, just grouping methods with the type would be sufficient in my mind. If this line of thinking still doesn't feel right to you, I'm happy to accept as-is.

pkg/platform/runtime/buildscript/buildexpression.go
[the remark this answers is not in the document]
daved:
```
switch v := inValue.(type) {
case map[string]interface{}:
  f, err := newFuncCall(v)
```

pkg/platform/runtime/buildscript/file.go
daved: If I understand correctly, the path field would no longer need to be held by the File type. In other words, both the write and update functions can not be methods. It feels to me that this would work towards simplifying File itself. However, feel free to disregard this thread.

pkg/platform/runtime/buildscript/buildexpression.go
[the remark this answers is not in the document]
mitchell-as: Well now I feel dumb. Thanks.

pkg/platform/runtime/buildscript/json.go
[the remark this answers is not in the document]
mitchell-as: I see. I am not in favor of that since build scripts are still in flux and would probably result in a lot of file churn. It's nice to have all of these structs present in one file for making quick changes.

pkg/platform/runtime/buildscript/file.go
mitchell-as: Thanks for the clarification. I have removed the `File` struct in favor of using functions that operate directly on paths.

pkg/platform/runtime/buildscript/buildexpression.go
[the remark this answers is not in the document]
daved: Oh, gee. Don't! Go has precious few niceties like this, and they're easy to miss.

daved (approved):

daved:
This looks really great, @mitchell-as!

mitchell-as:
Thanks for all the helpful feedback!

PR #2554, opened 2023-05-19
BuildExpression parser

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1863" title="DX-1863" target="_blank">TaskDX-1863</a>  We can parse a BuildExpression into an AST
  </td></table>
  <br />


pkg/buildexpression/parser/parser_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/ActiveState/cli/internal/environment"
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPaser_Parse(t *testing.T) {
```
mitchell-as, line 12: It's great that you have tests here, but they're only testing a lack of failure. You're not testing that the parser is producing the AST you expect it to.

pkg/buildexpression/parser/parser.go
```diff
@@ -0,0 +1,541 @@
...
+	case STRING:
+		err = p.parseString(assignmentNode)
+		msg = "Failed to parse string"
+	case IDENTIFIER:
+		err = p.parseIdentifier(assignmentNode)
+		msg = "Failed to parse identifier"
+	case IN, R_CURL:
+		// If we encounter an IN or R_CURL token then we've reached the end of the binding
+		return nil
+	default:
```
mitchell-as, line 238: Would you consider having an explicit case for lists here, and defaulting to an error for unexpected input?

pkg/buildexpression/parser/parser.go
```diff
@@ -0,0 +1,541 @@
...
+
+		var msg string
+		switch p.tok {
+		case STRING:
+			err = p.parseString(node)
+			msg = "Failed to parse string"
+		case IDENTIFIER:
+			err = p.parseIdentifier(node)
+			msg = "Failed to parse identifier"
+		default:
```
mitchell-as, line 337: Again, would you consider erroring out for unexpected input?

pkg/buildexpression/parser/parser.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+	if err != nil {
+		return errs.Wrap(err, msg)
+	}
+
+	return nil
+}
+
+func (p *Parser) isAssignment() bool {
+	return p.tok == IDENTIFIER && p.peek() == COLON && p.tok != IN
```
mitchell-as, line 522: Is it possible for `p.tok == IDENTIFIER` and `p.tok == IN` to both be true at once? If not, then you can remove the `p.tok != IN` test.

mitchell-as (changes requested):
This is a spirited first attempt at writing a parser for Platform build expressions. I am seriously impressed that you handwrote a lexer and parser for it. It's an amazing achievement.

That said, I have a number of concerns that I've pointed out in the comments. I know that this whole build planner/build scripts/build expressions saga has undergone so many different proposals, changes, adjustments that it's hard to see the end goal. I don't profess to understand what the end goal is, and my experience on this subject is a week old at best. However, I feel compelled to share some thoughts from my (perhaps naive) perspective.

Seeing as we've seemingly settled on something relatively simple, that is we do not care about preserving whitespace or comments, and seeing as we're offloading a lot of translation between build expressions (Platform) and build scripts (State Tool) to another ticket (https://activestatef.atlassian.net/browse/DX-1790), I honestly question the need for the lexer, parser, and AST. Please understand I am not allocating any sort of blame. I'm sure this approach was the best at one point in time (i.e. when we cared about preserving whitespace, allowing comments to be added, etc.), and it may very well still be, but I'm not so sure. (Again, please forgive me if I'm completely missing the point.)

Build Expressions are JSON objects and can be unmarshalled into JSON structures that can be operated on, since we no longer care about perserving changes like whitespace and additions like comments. The parser is littering the AST with character nodes (e.g. `{`, `:`, and `[`) that are of no use (because we're not trying to reconstruct input from the AST) and are a chore to skip over (e.g. https://github.com/ActiveState/cli/blob/925ab04e890c0f1e3aaadc95fd72b7bc8a8ea730/pkg/platform/runtime/buildscript/buildexpression.go#L22, https://github.com/ActiveState/cli/blob/925ab04e890c0f1e3aaadc95fd72b7bc8a8ea730/pkg/platform/runtime/buildscript/buildexpression.go#L27, and https://github.com/ActiveState/cli/blob/925ab04e890c0f1e3aaadc95fd72b7bc8a8ea730/pkg/platform/runtime/buildscript/buildexpression.go#L61). Of course, this is from my perspective over in build scripts. I don't know exactly how build expressions are intended to be used in conjunction with the build planner.

Once again, I'm not belittling this PR. Its handwritten lexer and parser is an amazing achievement and one I applaud whole-heartedly. Maintenance may be a burden, and seeing as all we're doing right now is parsing JSON, could we perhaps take advantage of Go's JSON unmarshaling features to make our lives easier?

MDrakos:
After a few discussions, we've simplified our approach to use the `encoding/json` package and Go maps. That is reflected in https://github.com/ActiveState/cli/pull/2559 so I will be closing this PR.

PR #2556, opened 2023-05-25
Update flags/args of init cmd

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1871" title="DX-1871" target="_blank">TaskDX-1871</a>  `state init` inconsistent with `state checkout`
  </td></table>
  <br />

### DIFF
--- a/cmd/state/internal/cmdtree/init.go
+++ b/cmd/state/internal/cmdtree/init.go
@@ -29,7 +29,6 @@ func newInitCommand(prime *primer.Values) *captain.Command {
 				Hidden: true,
 			},
 			{
-				// Supports Komodo
 				Name:        "language",
 				Description: locale.T("flag_state_init_language_description"),
 				Value:       &params.Language,
--- a/cmd/state/internal/cmdtree/init.go
+++ b/cmd/state/internal/cmdtree/init.go
@@ -23,21 +23,16 @@ func newInitCommand(prime *primer.Values) *captain.Command {
 		prime,
 		[]*captain.Flag{
 			{
-				Name:        "path",
-				Description: locale.T("flag_state_init_path_description"),
-				Value:       &params.Path,
-			},
-			{
 				// Hidden flag for legacy Komodo support
 				Name:   "skeleton",
 				Value:  p.StrP(""),
 				Hidden: true,
 			},
 			{
-				// Hidden flag for legacy Komodo support
-				Name:   "language",
-				Value:  &params.Language,
-				Hidden: true,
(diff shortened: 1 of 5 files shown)


cmd/state/internal/cmdtree/init.go
```diff
@@ -22,22 +22,17 @@ func newInitCommand(prime *primer.Values) *captain.Command {
...
 				Name:   "skeleton",
 				Value:  p.StrP(""),
 				Hidden: true,
 			},
 			{
-				// Hidden flag for legacy Komodo support
-				Name:   "language",
-				Value:  &params.Language,
-				Hidden: true,
+				// Supports Komodo
```
Naatan, line 32:
```suggestion
```

internal/locale/locales/en-us.yaml
```diff
@@ -1246,12 +1246,12 @@ arg_state_init_namespace:
...
-arg_state_init_language_description:
-  other: The language that this project should use, and optionally the version of that language
-flag_state_init_path_description:
-  other: Where to initialize the project
+flag_state_init_language_description:
+  other: "The Language[@version] that this project should use"
+arg_state_init_path:
+  other: path
+arg_state_init_path_description:
+  other: "Where to initialize the project (Default: Current working directory)"
```
Naatan, line 1254:
```suggestion
  other: "Where to initialize the project"
```

We don't seem to do this for other commands. Plus doing this will lead to bugs when we change the default behavior but forget the description cause it's not covered by our type system.

internal/output/progress_win.go
```diff
@@ -47,7 +47,7 @@ func (d *Spinner) moveCaretBackInCommandPrompt(n int) {
 		cursor.x = csbi.cursorPosition.x + short(-n)
 		cursor.y = csbi.cursorPosition.y
 
-		r2, _, err2 := procSetConsoleCursorPosition.Call(uintptr(handle), uintptr(*(*int32)(unsafe.Pointer(&cursor))))
+		r2, _, err2 := procSetConsoleCursorPosition.Call(uintptr(handle), uintptr(*((*uint32)(unsafe.Pointer(&cursor)))))
```
Naatan, line 50: This seems unrelated to this PR?

Naatan (changes requested):

internal/output/progress_win.go
mitchell-as: Looks like he pulled in this fix from v0.40.

Naatan (approved):

PR #2557, opened 2023-05-25
Show executables path after init

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1879" title="DX-1879" target="_blank">BugDX-1879</a>  INIT: Summary of the successful initialization of the project not includes path to executables
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1242,6 +1242,7 @@ init_success:
   other: |
     Project '[NOTICE]{{.V0}}[/RESET]' has been successfully initialized at '[NOTICE]{{.V1}}[/RESET]' and pushed to the Platform.
     You can start using your project with [ACTIONABLE]`state shell`[/RESET] and [ACTIONABLE]`state use`[/RESET].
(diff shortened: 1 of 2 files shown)


internal/output/progress_win.go
```diff
@@ -47,7 +47,7 @@ func (d *Spinner) moveCaretBackInCommandPrompt(n int) {
 		cursor.x = csbi.cursorPosition.x + short(-n)
 		cursor.y = csbi.cursorPosition.y
 
-		r2, _, err2 := procSetConsoleCursorPosition.Call(uintptr(handle), uintptr(*(*int32)(unsafe.Pointer(&cursor))))
+		r2, _, err2 := procSetConsoleCursorPosition.Call(uintptr(handle), uintptr(*((*uint32)(unsafe.Pointer(&cursor)))))
```
daved, line 50: I'm not sure how this got "autoformatted", but it did, and it does look worth keeping. This is the only place in the code that this is done, so it's not out of sync with anything else.

Naatan: Please fix your PR so this isn't in the history.

Naatan (changes requested):

Naatan (approved):

PR #2558, opened 2023-05-26
Move commitID from activestate.yaml into its own .activestate/commit file

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1852" title="DX-1852" target="_blank">StoryDX-1852</a>  The commit ID is tracked from .activestate/commit
  </td></table>
  <br />
 

Also add the commit file to .gitignore if necessary.


mitchell-as:
This is ready for initial review, as critical tests are passing (despite what it looks like on CI), and I'll be out on Monday.

pkg/projectfile/projectfile.go
```diff
@@ -706,21 +720,11 @@ func (p *Project) SetNamespace(owner, project string) error {
...
-func (p *Project) SetCommit(commitID string, headless bool) error {
-	pf := NewProjectField()
-	if err := pf.LoadProject(p.Project); err != nil {
-		return errs.Wrap(err, "Could not load activestate.yaml")
-	}
-	pf.SetCommit(commitID, headless)
-	if err := pf.Save(p.path); err != nil {
-		return errs.Wrap(err, "Could not save activestate.yaml")
+func (p *Project) SetCommit(commitID string) error {
+	if err := localcommit.Set(filepath.Dir(p.Path()), commitID); err != nil {
```
mitchell-as, line 724: Note to self: forgot to remove commitID from project field if it exists.

pkg/localcommit/localcommit.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return commitID, nil
+}
+
+func Set(projectDir, commitID string) error {
+	if !strfmt.IsUUID(commitID) {
+		return locale.NewError("err_commit_id_invalid", commitID)
+	}
+
+	updateGitIgnore := shouldAddToGitIgnore(projectDir)
```
Naatan, line 47: This should only happen at checkout and init time, not each time we set a commit.

pkg/localcommit/localcommit.go
```diff
@@ -0,0 +1,108 @@
...
+		return true // fresh checkout
+	}
+
+	for _, file := range files {
+		if file.Name() == ".git" {
+			return true // project is under Git revision control
+		}
+	}
+
+	return false
```
Naatan, line 76:
```suggestion
	return fileutils.DirExists(filepath.Join(projectDir, "git")
```

pkg/localcommit/localcommit.go
```diff
@@ -0,0 +1,108 @@
...
+type FileDoesNotExistError struct{ *locale.LocalizedError }
+
+func IsFileDoesNotExistError(err error) bool {
+	return errs.Matches(err, &FileDoesNotExistError{})
+}
+
+func Get(projectDir string) (string, error) {
+	configDir := filepath.Join(projectDir, constants.ProjectConfigDirName)
+	commitFile := filepath.Join(configDir, constants.CommitIdFileName)
+	if !fileutils.DirExists(configDir) || !fileutils.FileExists(commitFile) {
```
Naatan, line 24:
```suggestion
	if !fileutils.DirExists(configDir) || !fileutils.TargetExists(commitFile) {
```

pkg/projectfile/projectfile.go
```diff
@@ -102,6 +103,7 @@ type Project struct {
 	Cache         string        `yaml:"cache,omitempty"`
 	path          string        // "private"
 	parsedURL     projectURL    // parsed url data
+	commitID      string
```
Naatan, line 106: Commits are no longer a property of the projectile. And so there really shouldn't be any mention of commits in either the projectfile or the project packages other than to facilitate migrating.

For any commit logic that we keep around here for said migration I suggest we clearly prefix it as `deprecatedCommitID` so we don't accidentally use the wrong thing and it's clear to the reader that this isn't where the actual code lives.

pkg/projectfile/projectfile.go
```diff
@@ -1022,6 +1026,13 @@ func createCustom(params *CreateParams, lang language.Language) (*Project, error
 		return nil, errs.Wrap(err, "Could not parse %s", tplName)
 	}
 
+	if commitID != "" {
+		err := localcommit.Set(params.Directory, commitID)
+		if err != nil {
+			return nil, errs.Wrap(err, "Could not set local commit")
+		}
+	}
```
Naatan, line 1034: We should also unset it in the activestate.yaml.

pkg/projectfile/projectfile.go
```diff
@@ -464,6 +466,18 @@ func (p *Project) Init() error {
 	}
 	p.parsedURL = parsedURL
 
+	if p.parsedURL.CommitID != "" {
+		// Note: do not migrate to localcommit here. Writing to disk during Parse() feels wrong.
+		// Migrate when SetCommit() is called instead.
+		p.commitID = p.parsedURL.CommitID
+	} else {
+		if commitID, err := localcommit.Get(filepath.Dir(p.Path())); err == nil {
+			p.commitID = commitID
```
Naatan, line 475: As noted in my other comment; this is no longer the responsibility of projectfile.

Any code that is currently accessing `projectfile.CommitID()` will have to be transitioned to instead use `localcommit.Get()`. For what it's worth in Goland you can use refactoring tools to make this change in a matter of seconds.

Naatan (changes requested):

mitchell-as:
Well, pulling commit out of project/projectfile and requiring clients to use localcommit has ballooned the size of this PR substantially. It's ready for general review though. I expect churn.

pkg/localcommit/localcommit.go
```diff
@@ -0,0 +1,91 @@
...
+	return commitID, nil
+}
+
+func GetUUID(projectDir string) (strfmt.UUID, error) {
+	commitID, err := Get(projectDir)
+	if err != nil {
+		return "", errs.Wrap(err, "Unable to get local commit")
+	}
+	return strfmt.UUID(commitID), nil
+}
```
Naatan, line 47: Nit: strfmt.UUID already has a `.String()` method. We don't need both `Get()` and `GetUUID()`, We could just have `Get() (strfmt.UUID, error)` and instances that need the string just call `.String()`.

pkg/localcommit/localcommit.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	b, err := fileutils.ReadFile(commitFile)
+	if err != nil {
+		return "", locale.WrapError(err, "err_get_commit_file", "Could not read your project runtime's commit ID file")
+	}
+
+	commitID := string(b)
+	if !strfmt.IsUUID(commitID) {
+		return "", locale.NewError("err_commit_id_invalid", commitID)
```
Naatan, line 35:
```suggestion
		return "", locale.NewInputError("err_commit_id_invalid", commitID)
```

This would only happen if users manually modified it.

pkg/localcommit/localcommit.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return strfmt.UUID(commitID), nil
+}
+
+func Set(projectDir, commitID string) error {
+	if !strfmt.IsUUID(commitID) {
+		return locale.NewError("err_commit_id_invalid", commitID)
+	}
+
+	commitFile := filepath.Join(projectDir, constants.ProjectConfigDirName, constants.CommitIdFileName)
```
Naatan, line 54: nit: You're doing this in `Get()` as well. Suggest making it a private function.

pkg/localcommit/localcommit.go
```diff
@@ -0,0 +1,91 @@
...
+		newline = "\r" + newline
+	}
+	b = append(b, []byte(newline)...)
+	b = append(b, []byte(locale.T("commit_id_gitignore"))...)
+
+	err = fileutils.WriteFile(gitIgnore, b)
+	if err != nil {
+		return locale.WrapError(err, "err_commit_id_gitignore_add",
+			"Unable to add your project runtime's commit ID file to .gitignore")
+	}
```
Naatan, line 88: We should first check if gitignore is already covering `.activestate/commit`.

pkg/projectfile/projectfield.go
```diff
@@ -99,6 +71,13 @@ func (p *projectField) Save(path string) error {
 		return errs.Wrap(err, "ioutil.ReadFile %s failed", path)
 	}
 
+	// Strip legacy commitID parameter.
+	q := p.url.Query()
+	if q.Has("commitID") {
+		q.Del("commitID")
+		p.url.RawQuery = q.Encode()
+	}
```
Naatan, line 79: This feels like the wrong place. I could theoretically make changes to these fields and use these changes, just to have them be changed when I call Save(). Save should only propagate the changes, not itself make changes.

pkg/projectfile/projectfile.go
```diff
@@ -464,6 +465,21 @@ func (p *Project) Init() error {
...
+		if err := localcommit.Set(filepath.Dir(p.Path()), p.parsedURL.LegacyCommitID); err != nil {
+			return errs.Wrap(err, "Could not create local commit file")
+		}
+		pf := NewProjectField()
+		if err := pf.LoadProject(p.Project); err != nil {
+			return errs.Wrap(err, "Could not load activestate.yaml")
+		}
+		if err := pf.Save(p.path); err != nil {
+			return errs.Wrap(err, "Could not save activestate.yaml")
+		}
```
Naatan, line 480: This is initializing a new Project struct and saving that. This should instead operate on the one we're on.

internal/assets/contents/activestate.yaml.tpl
```diff
@@ -1,5 +1,5 @@
 project: {{.Project}}
-{{if and (.Private) (eq .CommitID "") }}
+{{if .Private }}
```
Naatan, line 2: This is to facilitate an older use-case. Ticket to get rid of it filed here:

https://activestatef.atlassian.net/browse/DX-1895

However, the change here is still troublesome as it would cause us to inject the `private` property on EVERY private project, whereas the intend is that it only happens for headless private projects. 

I think it's fine to just drop this template portion since it no longer serves a purpose as per the linked jira ticket.

Naatan (changes requested):

internal/constraints/constraints.go
```diff
@@ -113,7 +114,7 @@ func NewPrimeConditional(auth *authentication.Auth, pj projectable, subshellName
 		"Url":       pjURL,
 		"Commit":    pjCommit,
 		"Branch":    pjBranch,
-		"Path":      pjPath,
+		"Path":      pjDir,
```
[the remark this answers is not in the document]
mitchell-as: That was not my intent. From what I gather in https://github.com/ActiveState/cli/pull/2558/files#diff-0bc5f0643bb7ae3a8beadb795df406943b58af413f51b6e318c2ea83a63b89adL102-L104, "Path" expands to the directory of the project file. We already have a `project.Dir()` method, so I'm using it. I wanted to rename the variable to be more descriptive because it's a directory, not a path.

internal/runbits/requirements/requirements.go
```diff
@@ -115,7 +115,11 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 
 	switch nsType {
 	case model.NamespacePackage, model.NamespaceBundle:
-		language, err := model.LanguageByCommit(pj.CommitUUID())
+		commitUUID, err := localcommit.GetUUID(pj.Dir())
+		if err != nil && !localcommit.IsFileDoesNotExistError(err) {
```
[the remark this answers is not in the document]
mitchell-as: I thought so too, but it fails one of the `state install` integration tests.

I did do a review of these to the best of my abilities. Originally I assumed there always had to be a commit ID, ran the integration tests, and plugged the `IsFileDoesNotExist` errors in where there were failures.

internal/runners/initialize/init.go
```diff
@@ -158,8 +168,14 @@ func (r *Initialize) Run(params *RunParams) error {
 		return locale.WrapError(err, "err_init_commit", "Could not create initial commit")
 	}
 
-	if err := proj.SetCommit(commitID.String()); err != nil {
-		return locale.WrapError(err, "err_init_setcommit", "Could not store commit to project file")
+	if err := localcommit.Set(proj.Dir(), commitID.String()); err != nil {
+		return errs.Wrap(err, "Unable to create local commit file")
+	}
+	if emptyDir || fileutils.DirExists(filepath.Join(path, ".git")) {
```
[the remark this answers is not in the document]
mitchell-as: The AC says "We automatically add the .activestate/commit file to the .gitignore if the current project has a .git directory, **OR if the target directory is empty (eg. fresh checkout)**." (emphasis mine)

I cannot add that conditional logic to `localcommit.AddGitIgnore()` because it will not add it in the empty directory case (at the time of calling, it will have activestate.yaml in it).

If you'd like to change the AC to only add .gitignore if .git/ directory exists, then I can make the conditional inside `localcommit.AddGitIgnore()`.

internal/testhelpers/e2e/session.go
```diff
@@ -288,11 +288,24 @@ func (s *Session) SpawnCmdWithOpts(exe string, opts ...SpawnOptions) *termtest.C
...
+	require.NoError(s.t, fileutils.WriteFile(filepath.Join(s.Dirs.Work, constants.ProjectConfigDirName, constants.CommitIdFileName), []byte(commitID)))
+}
+
+// PrepareProject creates a very simple activestate.yaml file for the given org/project and, if a
+// commit ID is given, an .activestate/commit file.
+func (s *Session) PrepareProject(namespace, commitID string) {
+	yaml := fmt.Sprintf("project: https://%s/%s", constants.DefaultAPIHost, namespace)
+	require.NoError(s.t, fileutils.WriteFile(filepath.Join(s.Dirs.Work, constants.ConfigFileName), []byte(yaml)))
+	if commitID != "" {
+		require.NoError(s.t, fileutils.WriteFile(filepath.Join(s.Dirs.Work, constants.ProjectConfigDirName, constants.CommitIdFileName), []byte(commitID)))
```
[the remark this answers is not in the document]
mitchell-as: Sure, but wrapping `s.PrepareCommitIdFile(commitID)` in an `if commitID != ""` conditional.

pkg/localcommit/localcommit.go
mitchell-as: Very well. I was trying for a "minimal diff" approach, but this PR is already so big. Why not?

pkg/projectfile/projectfile.go
mitchell-as: Perhaps I'm missing something, but this is how it's done when setting namespace (https://github.com/ActiveState/cli/blob/4f1d3400cdffcf6b7feeee5ad4552129fb0ff40d/pkg/projectfile/projectfile.go#L702-L709) and branch (https://github.com/ActiveState/cli/blob/4f1d3400cdffcf6b7feeee5ad4552129fb0ff40d/pkg/projectfile/projectfile.go#L722-L735). The projectfile's project field does not appear to persist anywhere.

internal/runbits/requirements/requirements.go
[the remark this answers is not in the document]
Naatan: Do you recall what test it was that was failing? Seems either this test needs to be updated or there's another use-case that we forgot about.

internal/runners/initialize/init.go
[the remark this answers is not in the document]
Naatan: I think that AC still stands. I'd argue you could still assert "empty" by seeing if the dir ONLY contains the activestate.yaml, but this is a nit so entirely up to you on how/if you handle this feedback.

pkg/localcommit/localcommit.go
Naatan: Yeah that ship has sailed :p

pkg/projectfile/projectfile.go
Naatan: Oh my bad, I was reading `pf` as being `Project` but this is that special struct we made that ONLY looks at the project field. Ignore my comment above.

internal/runbits/requirements/requirements.go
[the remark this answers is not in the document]
mitchell-as: TestPackageIntegrationTestSuite/TestInstall_Empty
https://github.com/ActiveState/cli/actions/runs/5126684167/jobs/9221496100#step:24:766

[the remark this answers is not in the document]
mitchell-as: I think its the one where you state install from scratch.

internal/constraints/constraints.go
```diff
@@ -90,19 +90,20 @@ func NewPrimeConditional(auth *authentication.Auth, pj projectable, subshellName
...
 		pjNamespace = pj.NamespaceString()
 		pjURL = pj.URL()
-		pjCommit = pj.CommitID()
-		pjBranch = pj.BranchName()
-		pjPath = pj.Path()
-		if pjPath != "" {
-			pjPath = filepath.Dir(pjPath)
+		commitID, err := localcommit.Get(pj.Dir())
+		if err != nil && !localcommit.IsFileDoesNotExistError(err) {
+			multilog.Error("Unable to get local commit: %v", err)
```
Naatan, line 102:
```suggestion
			multilog.Error("Unable to get local commit: %v", errs.JoinMessage(err))
```

internal/constraints/constraints.go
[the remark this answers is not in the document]
Naatan: After looking again; I missed that it was doing `filepath.Dir()` before on the pjPath, so this actually simplifies it. My bad!

internal/runbits/requirements/requirements.go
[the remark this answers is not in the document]
Naatan: I happened to be playing with this code myself yesterday, and running into similar issues. The intend of the code is that it falls through to this logic over here:

https://github.com/ActiveState/cli/pull/2558/files#diff-e6c3f1c43aa6786ae16024b493aff78e55e41cf15d5cec53d76242974a4c30d0R138

Which is to say you're handling of the code is correct here, it's just that this function itself has become a big mess that's convoluted to deal with.

Filed a followup here: https://activestatef.atlassian.net/browse/DX-1897

Naatan (changes requested):
Looking good! Just one minor change to make rollbar reports more actionable.

mitchell-as:
Note: I rebased against v0.40 to get latest integration test fix.

internal/constraints/constraints.go
Naatan: Doesn't look like this made it in.

Naatan (changes requested):
In the future please merge instead of rebase, as I cannot see what has changed since the last review and now either need to review the whole thing again or assume that your rebase was clean and didn't introduce any new changes.

mitchell-as:
Ugh, sorry about that. I thought the commits would carry over as-is because rebasing doesn't affect review status. I will merge next time. Nothing new was introduced. It was a clean rebase.

You're right that the requested change did not carry over because I made the commit via GitHub and my local rebase didn't pull first :( I've remedied that and added more instances of `errs.JoinMessage()` in similar places.

Naatan (approved):

PR #2560, opened 2023-05-29
Support namespace and timestamp for commands and make json output more useful

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1887" title="DX-1887" target="_blank">StoryDX-1887</a>  Support `state install` by adding namespace and timestamp flags and giving json output
  </td></table>
  <br />

### DIFF
--- a/internal/captain/values.go
+++ b/internal/captain/values.go
@@ -175,7 +175,7 @@ func (p *PackageValueNoVersion) Set(s string) error {
 		return errs.Wrap(err, "PackageValue.Set failed")
 	}
 	if p.Version != "" {
-		return fmt.Errorf("Specifying a version is not supported, package format should be '[<namespace>/]<name>'", s)
+		return fmt.Errorf("Specifying a version is not supported, package format should be '[<namespace>/]<name>'")
 	}
 	return nil
 }
--- a/internal/rtutils/p/pointresolv.go
+++ b/internal/rtutils/p/pointresolv.go
@@ -56,6 +56,11 @@ func PboolP(v *bool) *bool {
 	return BoolP(*v)
 }
 
+// Pointer adds a pointer to a value
+func Pointer[V any](v V) *V {
+	return &v
+}
+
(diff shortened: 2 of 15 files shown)


internal/captain/values.go
```diff
@@ -122,35 +131,78 @@ type PackageValue struct {
...
+	p.Namespace = strings.TrimSpace(strings.Join(v[0:len(v)-1], "/"))
+	p.Name = strings.TrimSpace(v[len(v)-1])
+	return nil
+}
+
+func (p *PackageValue) Type() string {
+	return "package"
+}
+
+// PackageValueNoVersion is identical to PackageValue except that it does not support a version.
```
daved, line 168: For these "is identical to PackageValue" types, maybe it's possible to have some config fields in PackageValue instead:
```
type Package struct {
    Namespace string
    Name      string
    Version   string

    IsVersionForbidden  bool 
    IsNamespaceRequired bool
}

And then have the Set and Type logic behavior based on the values.

**Review by daved — CHANGES_REQUESTED** [review]

File: internal/captain/values.go (hunk 1 above)
Comment (Naatan) [reply internal/captain/values.go:168]: I tried this but it just makes things more difficult as it would require us to manually initialize these structs and configure them before we can use them. That's possible, but it feels better to centralize the boilerplate in this file rather than throughout our cmdtree.

File: internal/rtutils/p/pointresolv.go [hunk 2]
```diff
@@ -56,6 +56,11 @@ func PboolP(v *bool) *bool {
 	return BoolP(*v)
 }
 
+// Pointer adds a pointer to a value
+func Pointer[V any](v V) *V {
+	return &v
+}
+
```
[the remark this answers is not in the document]
Naatan: I've added it to a future sprint. Doesn't really change anything for this story though, just means that when we do that story we'll have to also update the code introduced here.

daved (approved):

PR #2567, opened 2023-06-02
Initial use of Python-like build scripts.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1858" title="DX-1858" target="_blank">StoryDX-1858</a>  Local orders use the Python representation
  </td></table>
  <br />
 

Hooking things up was relatively straightforward. I did have to change StageCommit to allow for a build expression update rather than an individual requirement update. Other than that, I discovered some minor errors in the build script implementation: null values and empty lists were not being handled properly, build script assignment order was not stable, and the "in" key was being duplicated on write because it is inside "let" in build expressions.

### DIFF
--- a/pkg/platform/api/graphql/model/buildplanner/buildExpression.go
+++ b/pkg/platform/api/graphql/model/buildplanner/buildExpression.go
@@ -60,7 +60,7 @@ type BuildExpression struct {
 	expression       map[string]interface{}
 	solveNode        *map[string]interface{}
 	requirementsNode []interface{}
-	bytes            []byte
+	raw              json.RawMessage
 }
 
 func NewBuildExpression(data []byte) (*BuildExpression, error) {
@@ -89,7 +89,7 @@ func NewBuildExpression(data []byte) (*BuildExpression, error) {
 		expression:       expression,
 		solveNode:        &solveNode,
 		requirementsNode: requirementsNode,
-		bytes:            data,
+		raw:              data,
 	}, nil
 }
 
@@ -247,5 +247,5 @@ func getFuncNode(expression map[string]interface{}, funcName string) (map[string
 }
 
 func (bx *BuildExpression) String() string {
-	return string(bx.bytes)
+	return string(bx.raw)
 }
--- a/pkg/platform/api/graphql/model/buildplanner/buildExpression.go
+++ b/pkg/platform/api/graphql/model/buildplanner/buildExpression.go
@@ -60,6 +60,7 @@ type BuildExpression struct {
 	expression       map[string]interface{}
 	solveNode        *map[string]interface{}
 	requirementsNode []interface{}
+	bytes            []byte
 }
 
 func NewBuildExpression(data []byte) (*BuildExpression, error) {
@@ -88,6 +89,7 @@ func NewBuildExpression(data []byte) (*BuildExpression, error) {
 		expression:       expression,
 		solveNode:        &solveNode,
 		requirementsNode: requirementsNode,
+		bytes:            data,
 	}, nil
 }
 
@@ -242,4 +244,8 @@ func getFuncNode(expression map[string]interface{}, funcName string) (map[string
 	}
 
 	return nil, funcNodeNotFoundError
+}
+
+func (bx *BuildExpression) String() string {
+	return string(bx.bytes)
 }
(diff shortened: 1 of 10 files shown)


pkg/platform/api/graphql/model/buildplanner/buildExpression.go
```diff
@@ -60,6 +60,7 @@ type BuildExpression struct {
 	expression       map[string]interface{}
 	solveNode        *map[string]interface{}
 	requirementsNode []interface{}
+	bytes            []byte
```
mitchell-as, line 63: bpModel.BuildExpression does have all the fields a BuildScript needs in order to convert from a bpModel.BuildExpression to a buildscript.BuildScript. Rather than extending BuildExpression, it's easier to work with its underlying JSON. If we make the incoming build expression's JSON available, we can construct a BuildScript from it.

mitchell-as, line 63: bpModel.BuildExpression does have all the fields a BuildScript needs in order to convert from a bpModel.BuildExpression to a buildscript.BuildScript. Rather than extending BuildExpression, it's easier to work with its underlying JSON. If we make the incoming build expression's JSON available, we can construct a BuildScript from it.

pkg/platform/runtime/buildscript/buildexpression.go
```diff
@@ -28,16 +29,17 @@ func NewScriptFromBuildExpression(expr []byte) (*Script, error) {
 	if !ok {
 		return nil, errs.New("'let' key is not a JSON object")
 	}
+	inValue, ok := letMap["in"]
+	if !ok {
+		return nil, errs.New("Build expression's 'let' object has no 'in' key")
+	}
+	delete(letMap, "in")
```
mitchell-as, line 36: Prevent duplication of "in" field when writing the build script. This may change depending on how we want the build script syntax to look in the end.

MDrakos: This comment is very helpful, so it should probably be in the code.

pkg/platform/api/graphql/model/buildplanner/buildExpression.go
MDrakos: nit: could we rename this to `raw` and use `json.RawMessage`? I feel it might better communicate what this field is. Though I'm not sure how easy it will be use `json.RawMessage`. If it's too much of a hassle or you disagree then please disregard this comment.

pkg/platform/runtime/buildscript/file.go
```diff
@@ -43,11 +43,15 @@ func UpdateOrCreate(dir string, newScript *model.BuildScript) error {
 		return nil
 	}
 
+	script, err = NewScriptFromBuildExpression([]byte(newScript.String()))
```
MDrakos, line 46: nit: script could also have a `Bytes()` function since it seems like you're only ever converting the string to a slice of bytes

pkg/platform/runtime/model/buildplanner.go
```diff
@@ -227,40 +227,46 @@ func removeEmptyTargets(bp *model.BuildPlan) {
...
+	Owner        string
+	Project      string
+	ParentCommit string
+	// Commits can have either an operation (e.g. installing a package)...
 	PackageName      string
 	PackageVersion   string
 	PackageNamespace vcsModel.Namespace
 	Operation        model.Operation
+	// ... or a script (e.g. from pull).
+	Script *model.BuildExpression
```
MDrakos, line 239: I believe this comment says that when we run `state pull` and the script needs to be updated we will end up submitting that to the BuildPlanner rather than the traditional commit operation parameters. Is that correct? 

Either way, I think we should make this comment a bit more clear. It took me a second to follow so someone else who's unfamiliar might benefit from something more detailed.

MDrakos (changes requested):
This looks great! Just some very minor nits/comments.

pkg/platform/runtime/buildscript/file.go
mitchell-as: When David reviewed my Build Script PR, he suggested the `String()` method because it's quite a common API convention, so I'm going with that :)

pkg/platform/runtime/model/buildplanner.go
mitchell-as: Yes, that is correct. I've updated the comment. Hopefully it's more helpful. Feel free to suggest something else if it's not.

mitchell-as:
Note: you can ignore the build failures because they are unrelated to this set of changes. (This branch needs to be merged with another one to fix them.)

MDrakos (approved):
Looks good!

PR #2570, opened 2023-06-05
Update workflows to use aws oidc auth.

gscho:


.github/workflows/build.yml
```diff
@@ -232,6 +237,14 @@ jobs:
         name: Generate Remote Install Deployment
         shell: bash
         run: state run generate-remote-install-deployment
+      
+      - # === Configure AWS credentials ==
+        name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: arn:aws:iam::062663976825:role/gha_oidc_activestate_cli
```
mitchell-as, line 245: These three fields should be variables, so that we can easily change them if needed (just like https://github.com/ActiveState/cli/pull/2570/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L260).

Are these sensitive? If so, we should make them secrets.

mitchell-as (changes requested):

.github/workflows/build.yml
gscho: They aren't secrets but given the choice between sharing our AWS account ID with the world vs not, I decided to make it a `secret` 👍

mitchell-as (approved):
Looks good to me!

PR #2582, opened 2023-06-12
Local orders (build scripts and commitID files)

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1859" title="DX-1859" target="_blank">TaskDX-1859</a>  Integrate local orders into versioned branches
  </td></table>
  <br />
 

Please see my last comment: https://github.com/ActiveState/cli/pull/2582#issuecomment-1662908678
This PR review is limited to the last 4 commits. All other commits have been previously reviewed in other PRs.


mitchell-as:
The only commits that have not been reviewed yet are the last 4: https://github.com/ActiveState/cli/pull/2582/files/ec0e9b44bf466b3244b80262131850e304b34402..18c1d2fb2ddd1a1224b77418cf503c40478c185c

All other commits have been reviewed as part of previous PRs merged into this feature branch, or they are commits from v40/v41 merges into this branch.

pkg/platform/runtime/buildexpression/merge/merge.go
```diff
@@ -56,22 +58,26 @@ func Merge(exprA *buildexpression.BuildExpression, exprB *buildexpression.BuildE
...
-	jsonB, err := getJsonMinusRequirements(exprB)
+	jsonB, err := getComparableJson(exprB)
 	if err != nil {
 		multilog.Error("Unable to get buildxpression minus requirements: %v", errs.JoinMessage(err))
 		return false
 	}
-	return reflect.DeepEqual(jsonA, jsonB)
+	logging.Debug("Checking for possibility of auto-merging build expressions")
+	logging.Debug("JsonA: %v", jsonA)
+	logging.Debug("JsonB: %v", jsonB)
```
Naatan, line 73: Guessing this was for local debugging? Or you want to keep this verbose for a while since this'll likely need some further debugging?

pkg/platform/runtime/buildexpression/merge/merge.go
```diff
@@ -91,24 +97,45 @@ func getJsonMinusRequirements(expr *buildexpression.BuildExpression) (map[string
...
 
 	return m, nil
 }
 
-// deleteRequirements recursively iterates over the given JSON map until it finds a "requirements"
-// key and deletes it and its value.
-func deleteRequirements(m *map[string]interface{}) bool {
+// deleteKey recursively iterates over the given JSON map until it finds the given key and deletes
+// it and its value.
+func deleteKey(m *map[string]interface{}, key string) bool {
```
Naatan, line 110: This should have some unit tests I think, especially for covering the recursive case.

pkg/platform/runtime/buildexpression/merge/merge.go
```diff
@@ -91,24 +97,45 @@ func getJsonMinusRequirements(expr *buildexpression.BuildExpression) (map[string
...
 		}
 	}
 	return false
 }
+
+// sortLists recursively iterates over the given JSON map looking for string lists, and sorts them.
+// This is needed because isAutoMergePossible() does a reflect.DeepEqual(), but build expression
+// list order does not matter.
+// This will not be necessary after DX-1939 is implemented.
+func sortLists(m *map[string]interface{}) {
```
Naatan, line 129: Same for this one; we should cover this with unit tests.

pkg/platform/runtime/buildscript/file.go
```diff
@@ -36,6 +36,9 @@ func newScriptFromFile(path, org, project string, auth *authentication.Auth) (*S
 	logging.Debug("Build script does not exist. Creating one.")
 	commitId, err := localcommit.Get(filepath.Dir(path))
 	if err != nil {
+		if localcommit.IsFileDoesNotExistError(err) {
+			return nil, nil // headless project
+		}
```
Naatan, line 41: Headless projects still have commits.

pkg/platform/runtime/buildscript/json.go
```diff
@@ -40,7 +41,17 @@ func (v *Value) MarshalJSON() ([]byte, error) {
...
+		// Buildexpression list order does not matter, so sorting is necessary for
+		// comparisons. Go's JSON marshaling is deterministic, so utilize that.
+		// This should not be necessary when DX-1939 is implemented.
+		list := make([]*Value, len(*v.List))
+		copy(list, *v.List)
+		sort.SliceStable(list, func(i, j int) bool {
+			b1, err1 := json.Marshal(list[i])
+			b2, err2 := json.Marshal(list[j])
+			return err1 == nil && err2 == nil && string(b1) < string(b2)
+		})
```
Naatan, line 53: This feels too hacky. We are effectively throwing our hands up and hoping for the best if an error occurs when marshalling. We're also marshalling list values redundantly, which I imagine could add up significantly for build expressions.

Is this still warranted considering you have `getComparableJson`? Can we handle this use-case differently?

Naatan (changes requested):

pkg/platform/runtime/buildexpression/merge/merge.go
mitchell-as: The latter. I've only ID'ed 3 cases in which build expressions may be different, but equal (as pointed out in DX-1939): different key-value order in objects, different list order, and different "at_time" values. There could be more, and I'd like to have this easily accessible for now. All the JSON is on one line and it's not too bad in terms of size, so it won't bloat up the logs like graphql queries and responses do.

pkg/platform/runtime/buildexpression/merge/merge.go
mitchell-as: It's not tested directly, but it is tested indirectly by the existing unit tests. I suppose I can add a direct, explicit test.

pkg/platform/runtime/buildscript/file.go
mitchell-as: Not the very first one, e.g. `state install` in an empty directory. I forgot that case too.

pkg/platform/runtime/buildscript/json.go
mitchell-as: > This feels too hacky. We are effectively throwing our hands up and hoping for the best if an error occurs when marshalling. We're also marshalling list values redundantly, which I imagine could add up significantly for build expressions.

No, this is a stable sort, so if there are errors marshaling (and there really shouldn't be), the order will remain the same. If that were to happen (and I cannot see how it would), then the worst that happens is that buildscript equality fails and we do a merge. That seems safe.

The redundant marshaling is happening in Go, so it's going to be fast. Moreover, this only happens when checking build script equality. I think the network call to fetch the remote build expression takes longer than the actual equality check.

> Is this still warranted considering you have `getComparableJson`? Can we handle this use-case differently?

`getComparableJson()` invokes this function because I need to handle different list order (https://github.com/ActiveState/cli/pull/2582/commits/d307bf98f29fc10557c8843f9d9a3a4acc214753#diff-1ebd8e286e3a9885c33bb17b9fdb14f4d2624f1d9dda84792bb018dd3352866eR171-R178). This will go away once we have hash checking in DX-1939.

Since we are currently comparing JSON directly at this time, we need everything to be ordered properly. Maps are fine, but lists are not. The only other way around it is to have a BuildScript struct method written in Go that iterates over all members, checking for equality with logic to handle out of order things. That's more throwaway code than this marshaling & string comparison method.

pkg/platform/runtime/buildexpression/merge/merge.go
Naatan: Yeah I'd prefer direct tests over indirect ones. Indirect testing is what integration tests are meant to cover.

pkg/platform/runtime/buildexpression/merge/merge.go
Naatan: Please add explicit tests.

pkg/platform/runtime/buildscript/file.go
Naatan: That might be an order of operations issue though. Even those projects have a commit, they just don't have a parent commit. I'm guessing this code gets invoked before we have created the commit? Is there any way we can change that so we don't need special handling here?

Naatan (changes requested):
Looks good! I'd just like to have a closer look at the no-commit issue here: https://github.com/ActiveState/cli/pull/2582#discussion_r1283427389

pkg/platform/runtime/buildscript/file.go
mitchell-as: Okay, I found it. When running `state install` in an empty directory, I made it so the commit ID file is written before the buildscript.

mitchell-as:
New test failures are due to a new Platform change that disallows passwords to contain usernames.

Naatan (approved):

mitchell-as:
@Naatan would you please force-merge this? I don't have the little checkbox and button.

PR #2584, opened 2023-06-12
List files that cannot be removed during uninstall.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1850" title="DX-1850" target="_blank">BugDX-1850</a>  UNINSTALL ST: Uninstall fail with errors if path contains unrecognized file
  </td></table>
  <br />

### DIFF
--- a/internal/runners/clean/run_lin_mac.go
+++ b/internal/runners/clean/run_lin_mac.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"strings"
 
 	svcApp "github.com/ActiveState/cli/cmd/state-svc/app"
 	"github.com/ActiveState/cli/internal/config"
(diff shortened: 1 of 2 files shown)


internal/runners/clean/run_lin_mac.go
```diff
@@ -46,9 +47,9 @@ func (u *Uninstall) runUninstall(params *UninstallParams) error {
 	}
 
 	err = removeInstall(u.cfg)
-	if errors.Is(err, errDirNotEmpty) {
+	if dirNotEmpty := (&dirNotEmptyError{}); errors.As(err, &dirNotEmpty) {
```
daved, line 50: Consider using errs.Matches.

internal/runners/clean/run_lin_mac.go
```diff
@@ -166,7 +169,8 @@ func removeEmptyDir(dir string) error {
 	}
 
 	if !empty {
-		return errDirNotEmpty
+		fileList := strings.Join(fileutils.ListDirSimple(dir, true), "\n - ")
```
daved, line 172: If we have a templating mechanism for looping to avoid formatting here, please use that.

daved (changes requested):

internal/runners/clean/run_lin_mac.go
mitchell-as: Thanks. I didn't know that existed.

daved: My original intention was to avoid handling the view formatting here at all. However, the localization library itself is limited to string inputs, so there's no good way to do this at this time. This or the original solution is fine. Though, this centralizes the handling some and that might be best.

daved (approved):

PR #2587, opened 2023-06-13
Build Planner implementation uses `commit` query

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1919" title="DX-1919" target="_blank">TaskDX-1919</a>  BuildPlan implementation uses new `commit` query
  </td></table>
  <br />


pkg/platform/model/buildplanner.go
```diff
@@ -100,29 +100,26 @@ func NewBuildPlannerModel(auth *authentication.Auth) *BuildPlanner {
 	}
 }
 
-func (bp *BuildPlanner) FetchBuildResult(commitID strfmt.UUID, owner, project string) (*BuildResult, error) {
+func (bp *BuildPlanner) FetchBuildResult(commitID strfmt.UUID) (*BuildResult, error) {
 	resp := &bpModel.BuildPlan{}
-	err := bp.client.Run(request.BuildPlan(owner, project, commitID.String()), resp)
+	err := bp.client.Run(request.BuildPlan(commitID.String()), resp)
```
Naatan, line 105: This looks wrong. We should always provide the project owner and name. According to this ticket the limitation should be addressed:

[PB-4359](https://activestatef.atlassian.net/browse/PB-4359)

pkg/platform/model/buildplanner.go
```diff
@@ -100,29 +100,26 @@ func NewBuildPlannerModel(auth *authentication.Auth) *BuildPlanner {
...
 		var errs []string
 		var isTransient bool
-		for _, se := range resp.Project.Commit.Build.SubErrors {
+		for _, se := range resp.Commit.Build.SubErrors {
 			errs = append(errs, se.Message)
 			isTransient = se.IsTransient
 		}
 		return nil, &BuildPlannerError{
-			wrapped:          locale.NewError("err_buildplanner", resp.Project.Commit.Build.Error),
+			wrapped:          locale.NewError("err_buildplanner", resp.Commit.Build.Error),
```
Naatan, line 122: This feels like it should be an input error.

I didn't really think about this when originally reviewing, but we want to make sure we don't end up logging build errors and solve errors to our rollbar, as they are not our responsibility.

Naatan (changes requested):

pkg/platform/model/buildplanner.go
MDrakos: That ticket is describing the `stageCommit` mutation. This function is for fetching the results of a build. We are now using the new top-level `commit` query which only requires a commit ID.

In this PR, I've changed how we fetch the build plan so that we only ever use the commit ID. This allows us to use the same function for both local and remote commits. Do we want to fetch the build result using the `project` query when the commit is not local?

Here's the Build Planners schema for reference: https://github.com/ActiveState/TheHomeRepot/blob/master/service/buildplanner/buildplan.graphql

pkg/platform/model/buildplanner.go
```diff
@@ -198,17 +195,28 @@ func (bp *BuildPlanner) FetchBuildResult(commitID strfmt.UUID, owner, project st
...
-			err := bp.client.Run(request.BuildPlan(owner, project, commitID), resp)
+			err := bp.client.Run(request.BuildPlan(commitID), resp)
 			if err != nil {
 				return nil, errs.Wrap(err, "failed to fetch build plan")
 			}
-			if resp.Project.Commit.Build.Status != bpModel.Planning {
+
+			if resp == nil {
+				continue
+			}
```
MDrakos, line 211: This seems to be happening intermittently when we poll the build planner. I will follow up with PB to see if this is a bug.

pkg/platform/model/buildplanner.go
Naatan: That still doesn't feel right though. How can we grab the build result without project info considering it needs the project info in order to resolve private ingredients?

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -65,30 +70,134 @@ func (o Operation) String() string {
...
 	Project *Project `json:"project"`
+	*Error
+}
+
+func (b *BuildPlanByProject) Build() (*Build, error) {
+	if b.Project == nil {
+		return nil, errs.New("Project is nil")
+	}
+
+	if b.Project.Error != nil && b.Project.Message != "" {
```
Naatan, line 96: This feels wrong; if Error isn't nil but Message is empty then we disregard the error? Same for the Commit error check further down. If this is intended please add a comment explaining.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -65,30 +70,134 @@ func (o Operation) String() string {
...
+		return nil, errs.New(b.Project.Commit.Message)
+	}
+
+	if b.Project.Commit.Type == NotFound {
+		return nil, locale.NewError("err_buildplanner_commit_not_found", "Build plan does not contain commit")
+	}
+
+	if b.Project.Commit.Build == nil {
+		if b.Project.Commit.Error != nil {
+			return nil, errs.New("Commit not found: %s", b.Project.Commit.Error.Message)
```
Naatan, line 114: This doesn't seem like it would ever trigger as line 105 would have already caught this.

pkg/platform/model/buildplanner.go
```diff
@@ -102,28 +103,36 @@ func NewBuildPlannerModel(auth *authentication.Auth) *BuildPlanner {
 }
 
 func (bp *BuildPlanner) FetchBuildResult(commitID strfmt.UUID, owner, project string) (*BuildResult, error) {
-	resp := &bpModel.BuildPlan{}
-	err := bp.client.Run(request.BuildPlan(owner, project, commitID.String()), resp)
+	logging.Debug("FetchBuildResult")
+	logging.Debug("CommitID: %s", commitID.String())
+	logging.Debug("Owner: %s", owner)
+	logging.Debug("Project: %s", project)
```
Naatan, line 109: Think we can drop these now?

pkg/platform/model/buildplanner.go
```diff
@@ -102,28 +103,36 @@ func NewBuildPlannerModel(auth *authentication.Auth) *BuildPlanner {
...
-	// Check for errors in the response
-	if resp.Project.Type == bpModel.NotFound {
-		return nil, locale.NewError("err_buildplanner_project_not_found", "Build plan does not contain project")
+	data, err := json.MarshalIndent(resp, "", "  ")
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not unmarshal JSON")
 	}
-	if resp.Project.Commit.Type == bpModel.NotFound {
-		return nil, locale.NewError("err_buildplanner_commit_not_found", "Build plan does not contain commit")
+	logging.Debug("Response:", string(data))
```
Naatan, line 120: Drop

pkg/platform/model/buildplanner.go
```diff
@@ -102,28 +103,36 @@ func NewBuildPlannerModel(auth *authentication.Auth) *BuildPlanner {
...
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not get build from response")
 	}
-	if resp.Project.Commit.Build.Type == bpModel.BuildResultPlanningError {
+
+	if build.PlanningError != nil {
 		var errs []string
 		var isTransient bool
-		for _, se := range resp.Project.Commit.Build.SubErrors {
+		for _, se := range build.SubErrors {
```
Naatan, line 130: This feels like something that should be handled in the `Build()` function (ie. the error contained could already address this use-case).

pkg/platform/model/buildplanner.go
```diff
@@ -160,56 +169,69 @@ func (bp *BuildPlanner) FetchBuildResult(commitID strfmt.UUID, owner, project st
...
-				return nil, errs.Wrap(err, "Build plan contains multiple recipe IDs")
-			}
-			res.RecipeID = strfmt.UUID(id.ID)
+	// This information will only be populated if the build is an alternate build.
+	// This is specified in the build planner queries.
+	for _, id := range build.BuildLogIDs {
+		if res.RecipeID != "" {
+			return nil, errs.Wrap(err, "Build plan contains multiple recipe IDs")
 		}
+		res.RecipeID = strfmt.UUID(id.ID)
```
Naatan, line 208:
```suggestion
		res.RecipeID = strfmt.UUID(id.ID)
		break
```

Or do we expect multiple matches and only want the last one?

pkg/platform/model/buildplanner.go
Naatan: Still a concern?

pkg/platform/model/buildplanner.go
```diff
@@ -272,24 +295,45 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
...
 
-	if resp.NotFoundError != nil {
-		return "", errs.New("Commit not found: %s", resp.NotFoundError.Message)
+	if resp.Commit == nil {
+		return "", errs.New("Staged commit is nil")
+	}
+
+	if resp.Commit.Build == nil {
+		if resp.Error != nil {
+			return "", errs.New("Commit not found: %s", resp.Error.Message)
```
Naatan, line 320: Do we actually know that this is a "not found" error?

pkg/platform/model/buildplanner.go
```diff
@@ -305,20 +349,22 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
 }
 
 func (bp *BuildPlanner) GetBuildExpression(owner, project, commitID string) (*buildexpression.BuildExpression, error) {
-	resp := &bpModel.BuildPlan{}
-	err := bp.client.Run(request.BuildExpression(owner, project, commitID), resp)
+	logging.Debug("GetBuildExpression")
```
Naatan, line 352: Drop

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -18,18 +18,41 @@ func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
...
+		// If there is an artifact for this terminal and its mime type is application/x-gozip-installer
+		// then the list of terminal target IDs should be the sources for the step that generated the
+		// installer. That is what we want to use to build the artifact map.
+		for _, nodeID := range terminal.NodeIDs {
+			artifact, ok := lookup[nodeID].(*model.Artifact)
+			if !ok {
+				continue
+			}
+
+			if artifact.MimeType == "application/x-gozip-installer" {
```
Naatan, line 41: This feels overly opinionated on specifically the gozip installer. If it were up to me this should never be part of the client-side logic and we should be looking for something more generic. Can you check with Tyler/Martin if there is a more appropriate thing that we can check for to mean "look up artifacts from this terminal".

Naatan (changes requested):

pkg/platform/model/buildplanner.go
MDrakos: Whoops, yes

pkg/platform/model/buildplanner.go
MDrakos: No, you're right. There should only be one.

pkg/platform/model/buildplanner.go
MDrakos: Not anymore, there are two error types: `Error` and `NotFoundError`. They both have a `message` field. I'm using a generic `Error` type to try and unmarshal both if we encounter them.

pkg/platform/model/buildplanner.go
MDrakos: This and other logging calls with function names could still be useful when tracing through logs, no?

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: The error is just a message. It's an embedded struct in the buildplanner types. If the error is nil then the message should also be.

I've updated how we access the message value.

pkg/platform/model/buildplanner.go
MDrakos: I've added a multilog call so we can know if this is happening. If we don't see it we can remove this. I've added extra error handling so hopefully this won't appear. In order to test properly I would have to disable it for a few integration test runs to see if it comes up.

pkg/platform/runtime/buildplan/buildplan.go
mdrohmann: Yeah, I agree.  If we add a different type of installer/packager like a docker container this won't work anymore.

My personal preference was that we have more specific terminals such that you don't need this graph walking at all.  So, the buildplan would have terminals with tags like this:

* `"platform:<UUID>:gozip-installer" == {"<UUID of gozip-artifact>"}`
* `"platform:<UUID>:state-tool-artifacts" == {"<UUID of artifact 1>", "<UUID of artifact 2>"}`

But I believe that Scott didn't like this for some reason.  @srobertson : Do you remember why you preferred that the clients would walk the graph here?

Assuming that there is a good reason, I would instead check for `artifact.MimeType != <State Tool Artifact Mime Type> || artifact.MimeType != <Camel Installer Mime Type>`, because these are the only ones the State Tool can handle.

So, basically you walk up the graph until you find what you are looking for oblivious of what it might be.

pkg/platform/api/buildplanner/model/buildplan.go
Naatan: Ok, but then Error being non-nil and message being empty IS an error condition which the code as written doesn't seem to address.

pkg/platform/model/buildplanner.go
Naatan: Ok yeah, if you think it's useful. Possibly add the input to the debug call? I assumed this was a local debugging statement, hence why I flagged it.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -54,8 +54,18 @@ const (
...
+
+	XArtifactMimeType            = "application/x.artifact"
+	XActiveStateArtifactMimeType = "application/x-activestate-artifacts"
+	XCamelInstallerMimeType      = "application/x-camel-installer"
 )
 
+func IsStateToolArtifact(mimeType string) bool {
+	return mimeType == XArtifactMimeType ||
+		mimeType == XActiveStateArtifactMimeType ||
+		mimeType == XCamelInstallerMimeType
```
Naatan, line 66: Are we sure this is a valid state tool artifact? IIRC we actually have offline installers for camel, just making sure that's not what we're matching here.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -93,29 +117,46 @@ func (b *BuildPlanByProject) Build() (*Build, error) {
 		return nil, errs.New("Project is nil")
 	}
 
-	if b.Project.Error != nil && b.Project.Message != "" {
-		return nil, errs.New(b.Project.Message)
+	if b.Project.Error != nil {
+		if b.Project.Error.Message != "" {
+			return nil, errs.New(b.Project.Message)
```
Naatan, line 122: Please wrap so we can trace the origin.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -93,29 +117,46 @@ func (b *BuildPlanByProject) Build() (*Build, error) {
...
 
 	if b.Project.Commit == nil {
 		return nil, errs.New("Commit is nil")
 	}
 
-	if b.Project.Commit.Error != nil && b.Project.Commit.Message != "" {
-		return nil, errs.New(b.Project.Commit.Message)
+	if b.Project.Commit.Error != nil {
+		if b.Project.Commit.Error.Message != "" {
+			return nil, errs.New(b.Project.Commit.Message)
```
Naatan, line 133: Wrap

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -124,16 +165,22 @@ func (b *BuildPlanByProject) CommitID() (strfmt.UUID, error) {
 		return "", errs.New("Project is nil")
 	}
 
-	if b.Project.Error != nil && b.Project.Message != "" {
-		return "", errs.New(b.Project.Message)
+	if b.Project.Error != nil {
+		if b.Project.Error.Message != "" {
+			return "", errs.New(b.Project.Message)
```
Naatan, line 170: Wrap

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -124,16 +165,22 @@ func (b *BuildPlanByProject) CommitID() (strfmt.UUID, error) {
...
 
 	if b.Project.Commit == nil {
 		return "", errs.New("Commit is nil")
 	}
 
-	if b.Project.Commit.Error != nil && b.Project.Commit.Message != "" {
-		return "", errs.New(b.Project.Commit.Message)
+	if b.Project.Commit.Error != nil {
+		if b.Project.Commit.Error.Message != "" {
+			return "", errs.New(b.Project.Commit.Message)
```
Naatan, line 181: Wrap

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -149,8 +196,11 @@ func (b *BuildPlanByCommit) Build() (*Build, error) {
 		return nil, errs.New("Commit is nil")
 	}
 
-	if b.Commit.Error != nil && b.Commit.Message != "" {
-		return nil, errs.New(b.Commit.Message)
+	if b.Commit.Error != nil {
+		if b.Commit.Error.Message != "" {
+			return nil, errs.New(b.Commit.Message)
```
Naatan, line 201: Wrap

Naatan (changes requested):

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: Yes, in the camel builds the artifacts with the mime type of `application/x-camel-installer` are what we should be installing. Here's an example of the artifact listing from a camel build plan

```
{
    "artifacts": [
        {
            "__typename": "ArtifactSucceeded",
            "nodeId": "3863e868-156c-5efb-8cc6-f15e961478ae",
            "mimeType": "application/x-activestate-builder",
            "generatedBy": "9792f42a-69f2-5577-b947-51c9b0570d93",
            "runtimeDependencies": [],
            "status": "SUCCEEDED",
            "logURL": "",
            "url": "s3://INVALID",
            "checksum": "816a927e49b8a5484d059afda25a1252517277f0"
        },
        {
            "__typename": "ArtifactSucceeded",
            "nodeId": "6fbb595e-ec6d-5347-ac9c-0c770b36472b",
            "mimeType": "text/plain",
            "generatedBy": "a1062e9b-b120-505e-900d-04c6aac1dac5",
            "runtimeDependencies": [],
            "status": "SUCCEEDED",
            "logURL": "https://s3.amazonaws.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e.log",
            "url": "https://dl.activestate.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e-packages.txt",
            "checksum": ""
        },
        {
            "__typename": "ArtifactSucceeded",
            "nodeId": "27f76f04-5bf4-5091-851f-4d6ff0bfffc1",
            "mimeType": "application/gzip",
            "generatedBy": "a1062e9b-b120-505e-900d-04c6aac1dac5",
            "runtimeDependencies": [],
            "status": "SUCCEEDED",
            "logURL": "https://s3.amazonaws.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e.log",
            "url": "https://dl.activestate.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e-tests.tar.gz",
            "checksum": ""
        },
        {
            "__typename": "ArtifactSucceeded",
            "nodeId": "5cdc504a-95d5-53c7-95c3-9605d468dbf0",
            "mimeType": "application/x-xar",
            "generatedBy": "a1062e9b-b120-505e-900d-04c6aac1dac5",
            "runtimeDependencies": [],
            "status": "SUCCEEDED",
            "logURL": "https://s3.amazonaws.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e.log",
            "url": "https://dl.activestate.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e.pkg",
            "checksum": ""
        },
        {
            "__typename": "ArtifactSucceeded",
            "nodeId": "58efa9c2-e49d-527c-a7d2-4adea0d410e3",
            "mimeType": "application/x-camel-installer",
            "generatedBy": "a1062e9b-b120-505e-900d-04c6aac1dac5",
            "runtimeDependencies": [],
            "status": "SUCCEEDED",
            "logURL": "https://s3.amazonaws.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e.log",
            "url": "https://dl.activestate.com/camel-builds/ActivePython/darwin-17.7.0/20200320T153029Z/ActivePython-2.7.16.0000-macosx10.9-x86_64-816a927e.tar.gz",
            "checksum": ""
        }
    ]
}
```

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: The `Error` here is just a struct with a `Message` field. There is no error (of the `error` type) to wrap.

Naatan: I think one of us is confused with what the other means, genuinely not sure which one :p What I meant is I would expect something like:

```
errs.New("API returned project error message: " + b.Project.Message)
```

And I'd expect them to each be unique for the flagged errors. The point being that when we see this error we can trace where it originated in the code. If we just give the error the API gave and don't wrap it then it may be impossible to trace where this error was triggered in the code.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2587#discussion_r1260303110

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: Ah okay, that makes sense. I thought you were asking to use `errs.Wrap()` which is where the confusion came from.

Naatan: Yeah it's a bit weird since you have an error, it's just not typed. And you want to wrap it. Technically you could certainly do an `errs.Wrap()` here but it feels a little pointless.

Naatan (changes requested):
All the errors you added have the exact same message. So if one happens we don't know which of 6 locations it came from.

MDrakos:
> All the errors you added have the exact same message. So if one happens we don't know which of 6 locations it came from.

Whoops, a couple of the error messages were the same. The rest were unique but maybe not unique enough. I've added identifiers to each error message though I'll admit my creativity is lacking. If you know what error message you would like to see please feel free to push a commit.

That said I think it would be useful to have a full stack trace in the error logs. I know we have something lke that now but I don't believe it helps much. When I've tried to use it I've also ended up relying on the error string.

Naatan (approved):

Naatan:
> That said I think it would be useful to have a full stack trace in the error logs. I know we have something lke that now but I don't believe it helps much. When I've tried to use it I've also ended up relying on the error string.

Yeah we started doing it in error messages with the intend of dropping the stack traces, as David was of the opinion that the Go way is to rely on the error message itself and thereby avoid paying for the extra processing of generating stack traces. 

At this point I don't see any reason to get rid of stack traces as they are incredibly helpful and their cost more than justifies the value they bring. But I also can see the merit of having error messages that are easily traceable as errors aren't always reported with full stack traces, regardless of whether they are available.

PR #2588, opened 2023-06-13
Fix state tool not detecting install path on Windows dev machine

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1811" title="DX-1811" target="_blank">BugDX-1811</a>  As a DevX developer I can no longer use my own build/state executable because state-svc cannot find its install location
  </td></table>
  <br />


mitchell-as (approved):
lol! So simple.

PR #2590, opened 2023-06-14
Notify the user if their build script has local changes.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1861" title="DX-1861" target="_blank">StoryDX-1861</a>  PoC: We detect when the buildscript is in a dirty state and warn the user
  </td></table>
  <br />
 

I'm ignoring the AC's "hashing" concept because we can get the build expression directly from the build planner, and there are existing routines to convert build scripts to/from build expressions for comparison (taking into account arbitrary sort orders). These routines are unit tested and are accurate.


mitchell-as:
Not sure how I feel about this PR. I'm submitting it now rather than later to see if I'm on the right track.

pkg/platform/runtime/runtime.go
```diff
@@ -63,11 +75,33 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
...
+				}
+			} else {
+				multilog.Error("Unable to get remote build expression: %v", errs.JoinMessage(err))
+			}
+		} else if !localcommit.IsFileDoesNotExistError(err) {
+			multilog.Error("Unable to get local commit: %v", errs.JoinMessage(err))
+		}
+	} else if !buildscript.IsDoesNotExistError(err) {
+		multilog.Error("Unable to get local build script: %v", errs.JoinMessage(err))
+	}
```
Naatan, line 98: I realized while reading this that we didn't cover for the offline use-case. Filed a followup here:

https://activestatef.atlassian.net/browse/DX-1934

It may be worth just doing that as this PR though, since much of the code being introduced here would effectively be partially throwaway. Up to you.

internal/runners/exec/exec.go
```diff
@@ -122,25 +121,22 @@ func (s *Exec) Run(params *Params, args ...string) (rerr error) {
...
 			return locale.WrapError(err, "err_activate_runtime", "Could not initialize a runtime for this project.")
 		}
 		pg := runbits.NewRuntimeProgressIndicator(s.out)
 		defer rtutils.Closer(pg.Close, &rerr)
-		if err := rt.Update(s.auth, pg); err != nil {
+		if err := rt.Update(pg); err != nil {
 			return locale.WrapError(err, "err_update_runtime", "Could not update runtime installation.")
 		}
+	} else if runtime.IsNeedsStageError(err) {
+		s.out.Notice(locale.T("notice_stage"))
```
Naatan, line 139: This feels like an awkward mix of patterns. Above we are checking `err != nil` and then handling cases of err not being nil. This else statement should IMO be part of those cases.

pkg/platform/runtime/runtime.go
```diff
@@ -47,12 +50,21 @@ func IsNeedsUpdateError(err error) bool {
...
+func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.SvcModel, auth *authentication.Auth) (*Runtime, error) {
 	rt := &Runtime{
 		target:    target,
 		store:     store.New(target.Dir()),
 		analytics: an,
 		svcm:      svcModel,
+		auth:      auth,
 	}
 
 	if !rt.store.MarkerIsValid(target.CommitUUID()) {
```
Naatan, line 70: I think we should probably make this a single function, eg. `rt.CacheIsValid()`, which then in turn calls out to `rt.store.MarkerIsValid()` and the new logic for the buildscript. That way this common behavior is organized as such.

pkg/platform/runtime/runtime.go
```diff
@@ -63,11 +75,33 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
 		}
 	}
 
+	if target.ProjectDir() == "" || auth == nil {
+		return rt, nil
+	}
```
Naatan, line 80: Auth shouldn't matter as it should b e optional for the buildplan API. If it's not then that needs to be addressed, but I'm fairly certain it is as we have already had a lot of back and forth over this with extensibility.

pkg/platform/runtime/runtime.go
```diff
@@ -63,11 +75,33 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
...
+		bp := model.NewBuildPlannerModel(auth)
+		if commitID, err := localcommit.Get(target.ProjectDir()); err == nil {
+			if expr, err := bp.GetBuildExpression(target.Owner(), target.Name(), commitID.String()); err == nil {
+				if !script.Equals(expr) {
+					return rt, &NeedsStageError{errs.New("Runtime changes should be staged")}
+				}
+			} else {
+				multilog.Error("Unable to get remote build expression: %v", errs.JoinMessage(err))
+			}
+		} else if !localcommit.IsFileDoesNotExistError(err) {
```
Naatan, line 93: This won't ever trigger cause line 85 is checking for non-nil errors, meaning this else would only get nil errors.

pkg/platform/runtime/runtime.go
Naatan, line 98: Please try to organize these conditionals better, maybe use a switch instead? This feels a bit error-prone in terms of reader comprehension.

pkg/platform/runtime/runtime.go
```diff
@@ -63,11 +75,33 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
...
+			multilog.Error("Unable to get local commit: %v", errs.JoinMessage(err))
```
Naatan, line 94: This can be an actual error, there's no recovery from this.

pkg/platform/runtime/runtime.go
```diff
@@ -63,11 +75,33 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
...
+		}
+	} else if !buildscript.IsDoesNotExistError(err) {
+		multilog.Error("Unable to get local build script: %v", errs.JoinMessage(err))
```
Naatan, line 97: This can also be an actual error.

Naatan (changes requested):

pkg/platform/runtime/runtime.go
mitchell-as: I was going by this line: https://github.com/ActiveState/cli/blob/mitchell/dx-1861/pkg/platform/model/buildplanner.go#L94
I will add an `auth != nil &&` bit to the conditional in order to allow for optional auth.

pkg/platform/runtime/runtime.go
mitchell-as: Looks okay to me. Must have been a trick of the light. Line 85 is checking for `err == nil`.

Naatan: My bad, I did indeed read that as `!= nil`.

pkg/platform/runtime/ARCHITECTURE.md
```diff
@@ -9,11 +9,15 @@ The general usage pattern is as follows:
...
-			return err
+		switch {
+			case runtime.IsNeedsUpdateError(err):
+				if err = rt.Update(messageHandler); err != nil {
+					return err
+				}
+			case runtime.IsNeedsStageError(err):
+				...
+			default:
+				return err
```
Naatan, line 20: Nit: Maybe slightly cleaner (this assumes dropping the `if` on line 11):

```
		switch {
			case err == nil:
				break
			case runtime.IsNeedsUpdateError(err):
				if err = rt.Update(messageHandler); err != nil {
					return err
				}
			case runtime.IsNeedsStageError(err):
				...
			default:
				return err
 		}
```

pkg/platform/runtime/store/store.go
```diff
@@ -267,3 +271,11 @@ func (s *Store) StoreBuildPlan(build *bpModel.Build) error {
 	}
 	return nil
 }
+
+func (s *Store) BuildExpression() ([]byte, error) {
+	return fileutils.ReadFile(s.buildExpressionFile())
+}
```
Naatan, line 277: I think the build expression itself should be tied to the commit ID. So we can catch the commit and the buildexpression being out of sync. Technically that "shouldn't" happen, but at the moment if it does happen it will be difficult to diagnose.

Naatan (changes requested):

pkg/platform/runtime/ARCHITECTURE.md
mitchell-as: I like it! Thanks.

pkg/platform/runtime/store/store.go
mitchell-as: @Naatan What do you propose here? I don't see a way to do this without an internet connection (to query the build expression for the current commit ID). Note we already have https://github.com/ActiveState/cli/blob/mitchell/dx-1861/internal/runbits/buildscript/buildscript.go#L37-L41 to detect changes when a runtime mutation occurs. I don't know if that helps.

Naatan: I meant we should store the build expression in a way that it also embeds the commit ID, so we can assure at the time that we source this information that the two are in sync.

Eg. the build_expression file could have the commit ID as the first line, and the first validation we do is ensuring that this commit matches the commit that's in the local store.

Again this shouldn't really be an issue if we did our job right, but considering how new this code is and how hard it is to oversee all possible permutations it feels better to have some safety guards here since it'd always break in very vague ways.

mitchell-as: I committed something that ties the two together. Let me know if you expected something else or if it needs improvement.

Naatan (approved):
Looks good! We may still want to iterate on hashing depending on what Martin comes back with, but we'll do this in a separate story.

PR #2594, opened 2023-06-20
Stream file uploads.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1906" title="DX-1906" target="_blank">StoryDX-1906</a>  `state upload` supports streaming the upload
  </td></table>
  <br />
 

Used this as a reference: https://medium.com/@owlwalks/sending-big-file-with-minimal-memory-in-golang-8f3fc280d2c

### DIFF
--- a/internal/gqlclient/gqlclient.go
+++ b/internal/gqlclient/gqlclient.go
@@ -15,7 +15,6 @@ import (
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/profile"
 	"github.com/ActiveState/cli/internal/singleton/uniqid"
 	"github.com/ActiveState/cli/internal/strutils"
@@ -191,7 +190,7 @@ func (c *Client) RunWithContext(ctx context.Context, request Request, response i
 	return nil
 }
 
-func (c *Client) runRaw(ctx context.Context, request *http.Request, response interface{}) error {
+func (c *Client) runRaw(ctx context.Context, request *http.Request, reqErrChan <-chan error, response interface{}) error {
 	var bearerToken string
 	if c.tokenProvider != nil {
 		bearerToken = c.tokenProvider.BearerToken()
@@ -205,9 +204,12 @@ func (c *Client) runRaw(ctx context.Context, request *http.Request, response int
 	}
 	request = request.WithContext(ctx)
 	c.Log(fmt.Sprintf(">> Raw Request: %s\n", request.URL.String()))
-	res, err := http.DefaultClient.Do(request)
-	if err != nil {
-		return err
+	res, responseErr := http.DefaultClient.Do(request)
+	if reqErr := <-reqErrChan; reqErr != nil {
+		return reqErr
+	}
+	if responseErr != nil {
+		return responseErr
 	}
 	defer res.Body.Close()
 	var buf bytes.Buffer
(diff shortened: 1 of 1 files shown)


mitchell-as:
This seems to work, but it'll be more comfortable when the integration tests are available.

internal/gqlclient/gqlclient.go
```diff
@@ -248,74 +248,76 @@ type JsonRequest struct {
 }
 
 func (c *Client) createMultiPartUploadRequest(gqlReq RequestWithFiles) (*http.Request, error) {
-	req, err := http.NewRequest("POST", c.url, nil)
+	bodyReader, body := io.Pipe()
```
Naatan, line 251: nit: Can we call the second one bodyWriter?

internal/gqlclient/gqlclient.go
```diff
@@ -248,74 +248,76 @@ type JsonRequest struct {
...
-		return nil, errs.Wrap(err, "Could not create form field map")
-	}
-	for n, f := range gqlReq.Files() {
-		if _, err := mapField.Write([]byte(fmt.Sprintf(`{"%d": ["%s"]}`, n, f.Field))); err != nil {
-			return nil, errs.Wrap(err, "Could not write map field")
+		// Operations
+		operations, err := mw.CreateFormField("operations")
+		if err != nil {
+			multilog.Error("Could not create form field operations: %v", err)
+			return
```
Naatan, line 276: These errors should not be dropped. I suggest utilizing a channel to catch them. You'd have to refactor the code such that the actual request happens within the scope of this function, as sharing channels between functions always ends up a bit messy.

Currently this happens here:

https://github.com/ActiveState/cli/blob/47e6eef1a732a2dc5ef32927952fe98d98055452/internal/gqlclient/gqlclient.go#L194-L194

I suggest moving the request logic to this function, and keeping the linked one just for handling the response.

Naatan (changes requested):

internal/gqlclient/gqlclient.go
mitchell-as: I changed the `createMultiPartUploadRequest()` function to return an error channel that can be checked immediately and from within `runRaw()` after submitting the graphql request. It seems to work okay, but let me know if I missed something.

Naatan: Sorry but that's what I was warning; sharing channels between functions always ends up being messy. Much of this code was proof of concept grade, so we're not beholden to the architecture, refactoring the code such that we don't share channels between functions feels reasonable.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2594#discussion_r1237627653

internal/gqlclient/gqlclient.go
mitchell-as: Okay, next time please word your request more forcefully. I took it to be a suggestion/recommendation, especially because we do this sort of thing (error channel passing) in runtime setup.

mitchell-as:
Okay, I think I made the changes you originally hoped to see. If not, let me know.

internal/gqlclient/gqlclient.go
```diff
@@ -307,21 +258,21 @@ func (c *Client) createMultiPartUploadRequest(gqlReq RequestWithFiles) (*http.Re
...
+				reqErrChan <- errs.Wrap(err, "Could not read file")
 				return
 			}
 		}
 	}()
 
 	varJson, err := json.Marshal(vars)
 	if err != nil {
-		return nil, errs.Wrap(err, "Could not marshal vars")
+		return errs.Wrap(err, "Could not marshal vars")
```
Naatan, line 275: Please move this above the goroutine, cause otherwise if and when this triggers the goroutine would be left dangling.

Naatan (changes requested):
Looking good, just one minor change.

Naatan (approved):

PR #2595, opened 2023-06-21
Initial implementation of `state projects delete`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1775" title="DX-1775" target="_blank">StoryDX-1775</a>  ST support delete project action
  </td></table>
  <br />
 

Since deleting projects via the MonoAPI is so simple, I didn't see a need to update the integration tests to use `state projects delete` when deleting projects created in tests. Also, there's the matter of ensuring project deletion even if errors occur; there's no guarantee if we switch to running `state projects delete` for cleanup.

### DIFF
--- a/internal/runners/projects/delete.go
+++ b/internal/runners/projects/delete.go
@@ -1,0 +1,60 @@
+package projects
+
+import (
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/internal/prompt"
+	"github.com/ActiveState/cli/pkg/platform/api/mono/mono_client/projects"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/project"
+)
+
+type DeleteParams struct {
+	Project *project.Namespaced
+}
+
+type Delete struct {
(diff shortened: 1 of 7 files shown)


internal/runners/projects/delete.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	monoParams := projects.NewDeleteProjectParams()
+	monoParams.SetOrganizationName(params.Project.Owner)
+	monoParams.SetProjectName(params.Project.Project)
+
+	_, err = d.auth.Client().Projects.DeleteProject(monoParams, d.auth.ClientAuth())
+	if err != nil {
+		return locale.WrapError(err, "err_projects_delete", "Unable to delete project")
+	}
```
daved, line 55: Please consider placing this in `pkg/platform/model/projects.go` and then calling that function here. I'm not sure if we have a requirement of reuse before placing this sort of logic in the model package, but I'd prefer the cleanliness of this logic being near other similar backend calls.

daved (changes requested):

internal/runners/projects/delete.go
mitchell-as: Good call. I didn't even realize that package existed.

daved (approved):

PR #2596, opened 2023-06-21
Add `state projects edit`

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1920" title="DX-1920" target="_blank">StoryDX-1920</a>  As a user I can edit my project name, repo and visibility
  </td></table>
  <br />


internal/runners/projects/edit.go
```diff
@@ -0,0 +1,179 @@
...
+func NewEdit(prime primeable) *Edit {
+	return &Edit{
+		auth:   prime.Auth(),
+		out:    prime.Output(),
+		prompt: prime.Prompt(),
+		config: prime.Config(),
+	}
+}
+
+func (e *Edit) Run(params EditParams) error {
```
mitchell-as, line 58: Do we need an `e.auth.Authenticated()` check here and error out ASAP if it fails?

internal/runners/projects/edit.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+
+	if params.Repository != "" {
+		editMsg += locale.Tl("edit_prompt_repo", "  - Repository: {{.V0}}\n", params.Repository)
+		editable.RepoURL = &params.Repository
+	}
+	editMsg += locale.Tl("edit_prompt_confirm", "Continue?")
+
+	edit, err := e.prompt.Confirm("", editMsg, p.BoolP(false))
```
mitchell-as, line 90: We should support `e.out.Config().Interactive` being a default choice.

internal/runners/projects/edit.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if err = e.editLocalCheckouts(params); err != nil {
+		return locale.WrapError(err, "err_edit_local_checkouts", "Could not edit local checkouts")
+	}
+
+	if err = e.updateProjectMapping(params); err != nil {
+		return locale.WrapError(err, "err_edit_project_mapping", "Could not update project mapping")
+	}
+
+	e.out.Print(locale.Tl("edit_success", "Project edited successfully"))
```
mitchell-as, line 112: Please use `e.out.Notice()` here, or it will output invalid JSON if JSON output mode is enabled.

internal/runners/projects/edit.go
```diff
@@ -0,0 +1,179 @@
...
+			localCheckouts = append(localCheckouts, checkouts...)
+		}
+	}
+
+	for _, checkout := range localCheckouts {
+		err := e.editLocalCheckout(params.Namespace.Owner, checkout, params)
+		if err != nil {
+			return errs.Wrap(err, "Could not edit local checkout at %s", checkout)
+		}
+	}
```
mitchell-as, line 132: Is there any reason to not combine these two loops into one that just calls `e.editLocalCheckout()` in the range over `localProjects`?

internal/runners/projects/edit.go
```diff
@@ -0,0 +1,179 @@
...
+	localProjects := projectfile.GetStaleProjectMapping(e.config)
+
+	var localCheckouts []string
+	for namespace, checkouts := range localProjects {
+		if namespace == params.Namespace.String() {
+			localCheckouts = append(localCheckouts, checkouts...)
+		}
+	}
+
+	ns, err := project.ParseNamespace(fmt.Sprintf("%s/%s", params.Namespace.Owner, params.ProjectName))
```
mitchell-as, line 169: `params.Namespace` is a `project.Namespaced`, which has a `.String()` method. Did you consider using it?

internal/runners/projects/edit.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return errs.Wrap(err, "Could not parse namespace")
+	}
+
+	for _, checkout := range localCheckouts {
+		projectfile.StoreProjectMapping(e.config, ns.String(), checkout)
+	}
```
mitchell-as, line 176: Again, can these two loops be combined? It just feels like extra code for no reason. I could be wrong though.

test/integration/projects_int_test.go
```diff
@@ -75,6 +77,87 @@ func (suite *ProjectsIntegrationTestSuite) TestJSON() {
...
+
+	// Verify the local checkouts have been updated
+	cp = ts.Spawn("projects")
+	cp.Expect(newName)
+	cp.ExpectExitCode(0)
+
+	// Change name back to original
+	cp = ts.Spawn("projects", "edit", fmt.Sprintf("ActiveState-CLI/%s", newName), "--name", originalName)
+	cp.Expect("You are about to edit")
+	cp.Send("y")
```
mitchell-as, line 123: After supporting non-interactive mode as mentioned earlier, I suggest changing this invocation to utilize it.

test/integration/projects_int_test.go
```diff
@@ -75,6 +77,87 @@ func (suite *ProjectsIntegrationTestSuite) TestJSON() {
 	//AssertValidJSON(suite.T(), cp) // list is too large to fit in terminal snapshot
 }
 
+func (suite *ProjectsIntegrationTestSuite) TestEdit_Name() {
```
mitchell-as, line 80: These two tests will fail sporadically due to race conditions when any two platforms are running them at the same time. Would it be reasonable to have platform-specific project names for these tests? (e.g. "Edit-Test" + runtime.GOOS).

mitchell-as (changes requested):
Most of this looks good. There are a few things to address I've made a note of.

internal/runners/projects/edit.go
MDrakos: This is creating a new namespace with the original owner and the newly modified name.

internal/runners/projects/edit.go
MDrakos: I tend to avoid nesting for loops as I find it more easily readable. However, if you feel strongly about it I can nest them.

test/integration/projects_int_test.go
MDrakos: The ACs in the ticket state that we should test the interactive scenarios.

internal/runners/projects/edit.go
mitchell-as: Just so we're on the same page,

```
	for namespace, checkouts := range localProjects {
		if namespace == params.Namespace.String() {
			err := e.editLocalCheckout(params.Namespace.Owner, checkout, params)
			if err != nil {
				return errs.Wrap(err, "Could not edit local checkout at %s", checkout)
			}
		}
	}
```

is what you mean by "nesting for loops"?

mitchell-as: Ohhhhh wait. I totally missed the trailing `...` in your append. Forget it. Its fine the way you have it.

test/integration/projects_int_test.go
MDrakos: Very good point! Thanks for catching that.

internal/runners/projects/edit.go
```diff
@@ -56,6 +56,10 @@ func NewEdit(prime primeable) *Edit {
 }
 
 func (e *Edit) Run(params EditParams) error {
+	if !e.auth.Authenticated() {
+		return locale.NewInputError("err_project_edit_not_authenticated", "In order to edit your project you need to be authenticated, please run '[ACTIONABLE]state auth[/RESET]' to authenticate.")
```
mitchell-as, line 60: Nit: comma splice

internal/runners/projects/edit.go
```diff
@@ -87,7 +91,7 @@ func (e *Edit) Run(params EditParams) error {
 	}
 	editMsg += locale.Tl("edit_prompt_confirm", "Continue?")
 
-	edit, err := e.prompt.Confirm("", editMsg, p.BoolP(false))
+	edit, err := e.prompt.Confirm("", editMsg, &e.out.Config().Interactive)
```
mitchell-as, line 94: I think you meant to use a `!` prefix here. The default should be false for interactive mode, and true for non-interactive mode. Also, is it safe to put the pointer to the actual mode here? Normally I would do something like `default := !e.out.Config().Interactive` and then `&default` in order to avoid possibly changing the underlying config value.

mitchell-as (changes requested):

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -122,6 +122,8 @@ func New(prime *primer.Values, args ...string) *CmdTree {
 	projectsCmd.AddChildren(
 		newRemoteProjectsCommand(prime),
 		newProjectsEditCommand(prime),
+		newRemoteProjectsCommand(prime),
```
mitchell-as, line 125: Looks like a duplicate from a merge.

mitchell-as (changes requested):
One tiny merge issue to address, and this should be done.

cmd/state/internal/cmdtree/cmdtree.go
MDrakos: Whoops, thanks for catching that.

PR #2597, opened 2023-06-22
Ensure runtime can be used on control error (e.g. needsupdate)

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1940" title="DX-1940" target="_blank">BugDX-1940</a>  Local order feature branch: runtime.Runtime panic
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -128,7 +128,9 @@ func (r *Runtime) validateCache() error {
 		if err != nil {
 			return errs.Wrap(err, "Unable to get remote build expression")
 		}
-		r.store.StoreBuildExpression(bpExpr, commitID)
+		if err := r.store.StoreBuildExpression(bpExpr, commitID); err != nil {
+			return errs.Wrap(err, "Unable to store build expression")
+		}
 		expr = bpExpr.String()
 	}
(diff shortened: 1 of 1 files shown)


pkg/platform/runtime/runtime.go
```diff
@@ -68,7 +68,7 @@ func newRuntime(target setup.Targeter, an analytics.Dispatcher, svcModel *model.
 
 	err := rt.validateCache()
 	if err != nil {
-		return nil, err // do not wrap; could be NeedsUpdateError, NeedsStageError, etc.
+		return rt, err
```
mitchell-as, line 71: Is my comment about not wrapping invalid? If so, that's fine. Otherwise, please leave it, as it's easy to change this into a wrap later during a refactor.

mitchell-as (approved):
Looks fine, but with a caveat/question.

pkg/platform/runtime/runtime.go
daved: The errs.Matches and other unwrapping error helpers should be able to inspect throughout the chain of errors, so wrapping shouldn't need to be avoided.

Naatan (approved):

PR #2598, opened 2023-06-22
Ensure empty {} JSON output for commands without any output.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1854" title="DX-1854" target="_blank">TaskDX-1854</a>  Standardize our empty JSON responses
  </td></table>
  <br />
 

Also mark more commands as not supporting structured output.

### DIFF
--- a/cmd/state/internal/cmdtree/invite.go
+++ b/cmd/state/internal/cmdtree/invite.go
@@ -40,5 +40,5 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
 		func(ccmd *captain.Command, args []string) error {
 			return inviteRunner.Run(&params, args)
 		},
-	).SetGroup(PlatformGroup).SetUnstable(true).SetHasVariableArguments()
+	).SetGroup(PlatformGroup).SetUnstable(true).SetHasVariableArguments().SetDoesNotSupportStructuredOutput()
 }
--- a/cmd/state/internal/cmdtree/invite.go
+++ b/cmd/state/internal/cmdtree/invite.go
@@ -12,7 +12,7 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
 
 	params := invite.Params{}
 
-	return captain.NewCommand(
+	cmd := captain.NewCommand(
 		"invite",
 		locale.Tl("invite_title", "Inviting New Members"),
 		locale.Tl("invite_description", "Invite new members to an organization"),
@@ -40,5 +40,11 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
 		func(ccmd *captain.Command, args []string) error {
 			return inviteRunner.Run(&params, args)
 		},
-	).SetGroup(PlatformGroup).SetUnstable(true).SetHasVariableArguments().SetDoesNotSupportStructuredOutput()
+	)
+	cmd.SetGroup(PlatformGroup)
+	cmd.SetUnstable(true)
+	cmd.SetHasVariableArguments()
+	cmd.SetDoesNotSupportStructuredOutput()
+
+	return cmd
 }
--- a/internal/runners/auth/logout.go
+++ b/internal/runners/auth/logout.go
@@ -28,10 +28,13 @@ func (l *Logout) Run() error {
 		return locale.WrapError(err, "err_auth_logout", "Failed to delete authentication key")
 	}
 	l.Outputer.Notice(output.Title(locale.Tl("authentication_title", "Authentication")))
-	if l.Auth.AvailableAPIToken() == "" {
-		l.Outputer.Notice(locale.T("logged_out"))
-	} else {
-		l.Outputer.Notice(locale.T("logout_still_have_api_token"))
+	message := locale.T("logged_out")
+	if l.Auth.AvailableAPIToken() != "" {
+		locale.T("logout_still_have_api_token")
 	}
+	l.Outputer.Print(output.Prepare(
+		message,
+		&struct{}{},
+	))
 	return nil
 }
--- a/internal/runners/auth/logout.go
+++ b/internal/runners/auth/logout.go
@@ -28,13 +28,10 @@ func (l *Logout) Run() error {
 		return locale.WrapError(err, "err_auth_logout", "Failed to delete authentication key")
 	}
 	l.Outputer.Notice(output.Title(locale.Tl("authentication_title", "Authentication")))
-	message := locale.T("logged_out")
-	if l.Auth.AvailableAPIToken() != "" {
-		locale.T("logout_still_have_api_token")
+	if l.Auth.AvailableAPIToken() == "" {
+		l.Outputer.Notice(locale.T("logged_out"))
+	} else {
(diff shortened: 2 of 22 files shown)


internal/runners/auth/logout.go
```diff
@@ -28,10 +28,13 @@ func (l *Logout) Run() error {
 		return locale.WrapError(err, "err_auth_logout", "Failed to delete authentication key")
 	}
 	l.Outputer.Notice(output.Title(locale.Tl("authentication_title", "Authentication")))
-	if l.Auth.AvailableAPIToken() == "" {
-		l.Outputer.Notice(locale.T("logged_out"))
-	} else {
-		l.Outputer.Notice(locale.T("logout_still_have_api_token"))
+	message := locale.T("logged_out")
+	if l.Auth.AvailableAPIToken() != "" {
+		locale.T("logout_still_have_api_token")
```
Naatan, line 33:
```suggestion
		message = locale.T("logout_still_have_api_token")
```

internal/runners/auth/logout.go
Naatan, line 1: We should just mark this one as does not support JSON IMO. It doesn't seem like it will ever produce json output, just the empty dict, at which point it's no different from looking at the exit code.

internal/runners/secrets/secrets.go
```diff
@@ -83,7 +83,7 @@ func (o *listOutput) MarshalStructured(format output.Format) interface{} {
 			out.HasValue = true
 		}
 
-		output = append(output, out)
+		output[i] = out
```
Naatan, line 86: Curious; what is this fixing?

internal/runners/secrets/set.go
```diff
@@ -75,6 +75,8 @@ func (s *Set) Run(params SetRunParams) error {
 		return err
 	}
 
+	s.out.Print(output.Prepare("", &struct{}{})) // {} for JSON
```
Naatan, line 78: Another example of a command that just does not produce any JSON output, so we should just mark it as does not support.

internal/runners/secrets/sync.go
```diff
@@ -63,7 +63,10 @@ func (s *Sync) Run() error {
 		return locale.WrapError(err, "secrets_err_sync", "Cannot synchronize secrets")
 	}
 
-	s.out.Notice(locale.Tr("secrets_sync_results_message", strconv.Itoa(updatedCount), org.DisplayName))
+	s.out.Print(output.Prepare(
+		locale.Tr("secrets_sync_results_message", strconv.Itoa(updatedCount), org.DisplayName),
+		&struct{}{},
+	))
```
Naatan, line 69: Should just mark as does not support.

internal/runners/update/unlock.go
```diff
@@ -60,7 +60,11 @@ func (u *Unlock) Run(params *UnlockParams) error {
 		return locale.WrapError(err, "err_update_projectfile", "Could not update projectfile")
 	}
 
-	u.out.Notice(locale.Tl("version_unlocked", "State Tool version unlocked"))
+	u.out.Print(output.Prepare(
+		locale.Tl("version_unlocked", "State Tool version unlocked"),
+		&struct{}{},
+	))
```
Naatan, line 66: Should just mark as does not support.

internal/runners/use/reset.go
```diff
@@ -56,7 +56,10 @@ func (u *Reset) Run(params *ResetParams) error {
 		return locale.NewInputError(locale.T("use_reset_notice_not_reset"))
 	}
 
-	u.out.Notice(locale.Tl("use_reset_notice_reset", "Stopped using your project runtime"))
+	u.out.Print(output.Prepare(
+		locale.Tl("use_reset_notice_reset", "Stopped using your project runtime"),
+		&struct{}{},
+	))
```
Naatan, line 62: Should just mark as does not support.

Naatan (changes requested):
Many of these feel to me as commands that just don't support JSON, where the issue isn't that the empty JSON isn't standardized, but that we address a JSON request in the first place.

The only benefit I can see to having them support JSON is to get JSON formatted errors, but that is a use-case onto itself that we can apply to every single command and feels out of scope for this story.

internal/runners/secrets/secrets.go
mitchell-as: If there are no secrets, the array will be empty, and the JSON output will be `null` instead of `[]`. Not sure exactly why. I'm using the `[i]` notation here to avoid multiple calls to `append()` because we already know the length.

mitchell-as:
Funny, I was thinking the same thing about marking just about everything as does not support JSON, but then I thought perhaps we should not require 3rd party tools to know exactly which commands do and do not support JSON output. `state use reset` and `state secrets set` were examples that came to mind. A `{}` return signals success (other than exit value), so 3rd party tools might just want to look for the former, not the latter.

Obviously there are some commands that will never have JSON support like `state uninstall`, so 3rd party tools shouldn't make that mistake, but for commands like `state use reset` and `state secrets set`, I didn't think there was any harm in returning `{}`. Regardless, I made your requested changes.

internal/runners/secrets/secrets.go
Naatan: Cool, thanks for the explanation!

Naatan:
> Funny, I was thinking the same thing about marking just about everything as does not support JSON, but then I thought perhaps we should not require 3rd party tools to know exactly which commands do and do not support JSON output. 

Yeah, I think long term this should still be a design goal for us, I see the recent improvements as us creating a not-quite-blank-but-close-enough slate to work off of.

> Obviously there are some commands that will never have JSON support like `state uninstall`, so 3rd party tools shouldn't make that mistake, but for commands like `state use reset` and `state secrets set`, I didn't think there was any harm in returning `{}`. Regardless, I made your requested changes.

You're right it's not much harm, but I feel like it sets a double standard. Either we support JSON or we don't. Supporting it just to communicate `{}` feels like we're kidding ourselves and our users. Note I think many of the commands might actually be able to communicate some useful JSON, but that's a story for another day..

cmd/state/internal/cmdtree/invite.go
```diff
@@ -40,5 +40,5 @@ func newInviteCommand(prime *primer.Values) *captain.Command {
 		func(ccmd *captain.Command, args []string) error {
 			return inviteRunner.Run(&params, args)
 		},
-	).SetGroup(PlatformGroup).SetUnstable(true).SetHasVariableArguments()
+	).SetGroup(PlatformGroup).SetUnstable(true).SetHasVariableArguments().SetDoesNotSupportStructuredOutput()
```
Naatan, line 43:
```suggestion
	).
		SetGroup(PlatformGroup).
		SetUnstable(true).
		SetHasVariableArguments().
		SetDoesNotSupportStructuredOutput()
```

Total nit.

Naatan (approved):

Naatan (approved):

PR #2600, opened 2023-06-22
Add `state publish` tests

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1888" title="DX-1888" target="_blank">TaskDX-1888</a>  We have integration tests for `state upload` 
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/api/graphql/request/publish.go
+++ b/pkg/platform/api/graphql/request/publish.go
@@ -114,8 +114,7 @@ func (p PublishVariables) MarshalYaml(includeExample bool) ([]byte, error) {
 }
 
 func (p *PublishVariables) UnmarshalYaml(b []byte) error {
-	err := yaml.Unmarshal(b, p)
-	return err
+	return yaml.Unmarshal(b, p)
 }
 
 var exampleAuthor = ExampleAuthorVariables{[]PublishVariableAuthor{{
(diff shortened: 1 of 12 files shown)


test/integration/publish_int_test.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+
+			cp := ts.SpawnWithOpts(
+				e2e.WithArgs(append([]string{"publish"}, args...)...),
+			)
+
+			// Send custom input via --editor
+			if tt.input.editorValue != nil {
+				cp.Expect("Press enter when done editing")
+				rx := regexp.MustCompile(`file:\s*?(.*?)\.\s`)
```
mitchell-as, line 218: I believe convention is to put this in the top-level scope as a constant so we catch it at compile time, not runtime. However, this is in an integration test, so I'll leave it up to you.

pkg/platform/api/graphql/request/publish.go
```diff
@@ -113,8 +113,9 @@ func (p PublishVariables) MarshalYaml(includeExample bool) ([]byte, error) {
 	return v, nil
 }
 
-func (p PublishVariables) UnmarshalYaml(b []byte) error {
-	return yaml.Unmarshal(b, &p)
+func (p *PublishVariables) UnmarshalYaml(b []byte) error {
+	err := yaml.Unmarshal(b, p)
+	return err
```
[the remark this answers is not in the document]
Naatan: Good eye. I made this change while debugging to inspect the value, and forgot to revert.

test/integration/publish_int_test.go
Naatan: Yeah with integration tests I tend to take more liberties, but fair point! Will address.

mitchell-as (approved):

PR #2601, opened 2023-06-23
Refactor pointer package to instead use generics

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1665" title="DX-1665" target="_blank">TaskDX-1665</a>  Our "p" (pointer) package uses generics
  </td></table>
  <br />


internal/analytics/client/sync/client.go
```diff
@@ -192,10 +192,10 @@ func (a *Client) EventWithLabel(category string, action, label string, dims ...*
 	}
 
 	if a.auth != nil && a.auth.UserID() != nil {
-		a.customDimensions.UserID = p.StrP(string(*a.auth.UserID()))
+		a.customDimensions.UserID = ptr.To(string(*a.auth.UserID()))
```
daved, line 195: Maybe...
```suggestion
		a.customDimensions.UserID = ptr.Clone(a.auth.UserID())
```

daved (changes requested):

internal/analytics/client/sync/client.go
daved: The returned type is a *strfmt.UUID, so it will need to be something like:
```
a.customDimensions.UserID = ptr.To(a.auth.UserID().String())
```

daved (changes requested):
Comment was left in existing thread.

daved (approved):

PR #2602, opened 2023-06-23
Return high-level localized error if there was a low-level runtime setup error.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1881" title="DX-1881" target="_blank">BugDX-1881</a>  Missing error l10n while setting up runtime
  </td></table>
  <br />
 

At least one error in the chain needs to be localized, so I picked the upper-most error that wraps lower-level errors like the "abnormal closure" error originally reported.

I tested by unplugging my network cable. It didn't give the same error, but I at least got my localized error:

```
█ Something Went Wrong

 x Could not install dependencies
 x Error setting up runtime
 x Could not resolve recipe for project mitchell-as/python#71a77bda-f2c3-4c4d-9dc7-f0b6d59d701e
 x Unknown error resolving order
```

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -266,7 +266,7 @@ func (s *Setup) updateArtifacts() ([]artifact.ArtifactID, error) {
 		return err
 	})
 	if err != nil {
-		return artifacts, errs.Wrap(err, "Error setting up runtime")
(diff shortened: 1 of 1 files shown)


daved (approved):

daved:
Approved, but... is there enough info here for the user?

PR #2603, opened 2023-06-23
Add stage cmd

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1857" title="DX-1857" target="_blank">StoryDX-1857</a>  We have a `state stage` command which creates a commit from the current buildscript
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/buildscript/buildscript.go
+++ b/internal/runbits/buildscript/buildscript.go
@@ -42,12 +42,12 @@ func Sync(proj *project.Project, commitID *strfmt.UUID, out output.Outputer, aut
 	logging.Debug("Synchronizing local build script using commit %s", commitID)
 	script, err := buildscript.NewScriptFromProjectDir(proj.Dir())
 	if err != nil && !buildscript.IsDoesNotExistError(err) {
-		return false, errs.Wrap(err, "Could not get local build script")
+		return staged, errs.Wrap(err, "Could not get local build script")
 	}
 
 	expr, err := getBuildExpression(proj, commitID, auth)
 	if err != nil {
-		return false, errs.Wrap(err, "Could not get remote build expr for provided commit")
+		return staged, errs.Wrap(err, "Could not get remote build expr for provided commit")
 	}
 
 	// Note: merging and/or conflict resolution will happen in another ticket (DX-1912).
@@ -56,23 +56,23 @@ func Sync(proj *project.Project, commitID *strfmt.UUID, out output.Outputer, aut
 	if script != nil && commitID == nil {
 		logging.Debug("Checking for changes")
 		if script.Equals(expr) {
-			return false, nil // nothing to do
+			return staged, nil // nothing to do
 		}
 		logging.Debug("Merging changes")
 		bytes, err := json.Marshal(script)
 		if err != nil {
-			return false, errs.Wrap(err, "Unable to marshal local build script to JSON")
+			return staged, errs.Wrap(err, "Unable to marshal local build script to JSON")
 		}
 		expr, err = bpModel.NewBuildExpression(bytes)
 		if err != nil {
-			return false, errs.Wrap(err, "Unable to translate local build script to build expression")
+			return staged, errs.Wrap(err, "Unable to translate local build script to build expression")
 		}
 
 		out.Notice(locale.Tl("buildscript_update", "Updating project to reflect build script changes..."))
 
 		localCommitID, err := localcommit.Get(proj.Dir())
 		if err != nil {
-			return false, errs.Wrap(err, "Unable to get local commit ID")
+			return staged, errs.Wrap(err, "Unable to get local commit ID")
 		}
 
 		bp := model.NewBuildPlannerModel(auth)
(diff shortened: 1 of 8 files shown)


internal/runners/stage/stage.go
```diff
@@ -0,0 +1,78 @@
...
+		out:       p.Output(),
+		proj:      p.Project(),
+		auth:      p.Auth(),
+		analytics: p.Analytics(),
+		svcModel:  p.SvcModel(),
+	}
+}
+
+func (s *Stage) Run() error {
+	if err := buildscript.Sync(s.proj, nil, s.out, s.auth); err != nil {
```
Naatan, line 44: s.proj is going to need a nil check first.

Naatan (changes requested):

internal/runbits/buildscript/buildscript.go
```diff
@@ -83,21 +83,22 @@ func Sync(proj *project.Project, commitID *strfmt.UUID, out output.Outputer, aut
...
-			return errs.Wrap(err, "Could not update project to reflect build script changes.")
+			return false, errs.Wrap(err, "Could not update project to reflect build script changes.")
 		}
+		staged = true
 		commitID = &stagedCommitID
 
 		expr, err = getBuildExpression(proj, commitID, auth) // timestamps might be different
 		if err != nil {
-			return errs.Wrap(err, "Could not get remote build expr for staged commit")
+			return false, errs.Wrap(err, "Could not get remote build expr for staged commit")
```
mitchell-as, line 93: Technically it was staged at this point, but there was an error fetching the resulting expression. Would it be bad to change this `false` to `staged`?

internal/runbits/buildscript/buildscript.go
```diff
@@ -83,21 +83,22 @@ func Sync(proj *project.Project, commitID *strfmt.UUID, out output.Outputer, aut
...
 		}
 	}
 
 	if err := buildscript.UpdateOrCreate(proj.Dir(), expr); err != nil {
-		return errs.Wrap(err, "Could not update local build script.")
+		return false, errs.Wrap(err, "Could not update local build script.")
```
mitchell-as, line 98: Same question, would it be bad to change this to `staged`?

mitchell-as (approved):
This looks good, though there is a questionable return value.

internal/runbits/buildscript/buildscript.go
daved: Good question, I'm not exactly sure. If the user ends up in a state where that staging is irrelevant/useless, then I'd argue that false is appropriate. If there is some utility to the staged commit, then I'd argue true. If we land on true while an error has occurred, then I think the caller will need to communicate more clearly to the user what is going on.

mitchell-as: To be safe, I would return `staged` here (and perhaps everywhere for clarity). It will be up to the caller to handle the distinction.

mitchell-as (changes requested):
To be safe, I would return staged here (and perhaps everywhere for clarity). It will be up to the caller to handle the distinction.

daved:
So then, the question becomes, what do I do as the caller to understand the state that the project is in based on the staged/error values? Is the stagedCommit id used in subsequent calls? Is it fine for it to dangle and/or be recreated repeatedly if the user runs `state stage` multiple times with failure? My lack of understanding of the overall design is limiting me at the moment.

mitchell-as:
I don't understand it completely either. Perhaps we can head down that route when/if we get there. Maybe @MDrakos or @Naatan can weigh in here.

internal/runbits/buildscript/buildscript.go
Naatan: I think false is appropriate. It does feel pointless to succeed on something we know won't work. If there is indeed a use-case for this let's hear about it before we address it.

Naatan: That said; maybe rename `staged` to `success` or something? It does seem that the name is a little off given it's not actually communicating whether it was staged (not JUST whether it was staged, anyway).

internal/runners/stage/stage.go
```diff
@@ -58,19 +63,29 @@ func (s *Stage) Run() error {
...
+		"No change to the buildscript was found.",
+	)
+	if changesStaged {
+		message = locale.Tl(
 			"refresh_project_statement",
 			"", s.proj.NamespaceString(), s.proj.Dir(), execDir,
-		),
+		)
+	}
+	s.out.Print(output.Prepare(
```
Naatan, line 77: It's pointless to print out the full struct when no changes were made imo.

Naatan (changes requested):

Naatan (approved):

daved:
The added test fails on:
Linux and Windows at `Could not meet expectation: 'Runtime updated'` with `Could not synchronize the buildscript`
Mac at `Could not meet expectation: 'No change'` with `Could not synchronize the buildscript`

mitchell-as (approved):

Naatan (approved):

PR #2604, opened 2023-06-23
Replace captain interceptor with "dumb" but straight forward mechanic

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1768" title="DX-1768" target="_blank">TaskDX-1768</a>  Simplify cmdtree intercept functionality
  </td></table>
  <br />
 

@daved I imagine you won't like this, I'm sorry; I tried using the pattern you had in a way that didn't lead to head scratching and could not find one. 

The problem as I see it is not the pattern itself, but the context in which it is applied. Usually, I think you will see this pattern in a fairly focussed use-case, whereas in our case there is so much already going on with the shimming happening between cobra and captain, that this additional complexity just makes it that much harder to grok what the code is doing.

I've opted instead for a very "dumb" mechanic, which to me seems perfectly fine for our use-cases.


daved (changes requested):
Now there are two stacks. This means no nesting of functions which has some ramifications.

The stack setup was:
```
cmd.AppendInterceptChain(printdur.New().Intercept, more.New().Intercept, extra.New().Intercept)
```
and is now:
```
p := printdur.New()
m := more.New()
e := extra.New()
cmd.OnExecStart(p, m, e)
cmd.OnExecStop(e, m, p)
```

The implementation was:
```
func (p *PrintDur) InterceptExec(next captain.ExecuteFunc) captain.ExecuteFunc {
	return func(cmd *captain.Command, args []string) error {
		start := time.Now()
		if err := next(cmd, args); err != nil {
			return err
		}
		fmt.Println(time.Since(start))
		return nil
	}
}
```
and is now:

```
type PrintDur struct {
	start time.Time
}

func (p *PrintDur) OnExecStart(cmd *captain.Command, _ []string) error {
	p.start := time.Now()
	return nil
}

func (c *CmdCall) OnExecStop(cmd *captain.Command, _ []string) error {
	if p.start.IsZero() {
		return errors.New("PrintDur OnExecStart was not called")
	}
	fmt.Println(time.Since(p.start))
	return nil
}
```
The new design somewhat goes against the natural flow of the call stack such that the relationship between the pre and post behaviors being added is disconnected and must be reconnected through a structure (rather than staying within one function scope) that is fully dependent on how it was setup by external callers (e.g. `p.start` is not surely set). Similarly, we cannot modify the cmd or args values since each pre, post, and endpoint function is being called in its own branch. This will also make future use of context.Context clunky. That is, each On{Start,Stop} function would need to return the context.Context type that was passed in along with the error. Also note that, unless the args to OnExecStop are reversed from OnExecStart, the function scopes will not "unwind" as a normal stack would.

PR #2605, opened 2023-06-26
Tag and report runtime setup errors to analytics.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1830" title="DX-1830" target="_blank">TaskDX-1830</a>  As a PM I can identify the failure reason of sourcing runtimes
  </td></table>
  <br />
 

Repurpose "runtime" category into "runtime-debug".

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -189,23 +189,27 @@ func (r *Runtime) recordCompletion(err error) {
 	// download error to be cause by an input error.
 	case locale.IsInputError(err):
 		errorType = "input"
-	// Progress errors should come before any other non-input error.
-	case errs.Matches(err, &setup.ProgressReportError{}) || errs.Matches(err, &buildlog.ProgressReportError{}):
-		errorType = "progress"
 	case errs.Matches(err, &model.SolverError{}):
 		errorType = "solve"
 	case errs.Matches(err, &setup.BuildError{}) || errs.Matches(err, &buildlog.BuildError{}):
 		errorType = "build"
 	case errs.Matches(err, &setup.ArtifactSetupErrors{}):
-		errorType = "install" // initial guess
 		if setupErrors := (&setup.ArtifactSetupErrors{}); errors.As(err, &setupErrors) {
 			for _, err := range setupErrors.Errors() {
-				if errs.Matches(err, &setup.ArtifactDownloadError{}) {
+				switch {
+				case errs.Matches(err, &setup.ArtifactDownloadError{}):
 					errorType = "download"
 					break // it only takes one download failure to report the runtime failure as due to download error
+				case errs.Matches(err, &setup.ArtifactInstallError{}):
+					errorType = "install"
+					// Note: do not break because there could be download errors, and those take precedence
 				}
 			}
 		}
+	// Progress/event handler errors should come last because they can wrap one of the above errors,
+	// and those errors actually caused the failure, not these.
+	case errs.Matches(err, &setup.ProgressReportError{}) || errs.Matches(err, &buildlog.EventHandlerError{}):
+		errorType = "progress"
 	}
 
 	r.analytics.Event(anaConsts.CatRuntime, action, &dimensions.Values{
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


pkg/platform/runtime/runtime.go
```diff
@@ -171,16 +172,50 @@ func (r *Runtime) recordCompletion(err error) {
...
+	switch {
+	case locale.IsInputError(err):
+		errorType = "input"
+	case errs.Matches(err, &setup.ProgressReportError{}): // should come before non-input errors
+		errorType = "progress"
+	case errs.Matches(err, &model.SolverError{}):
+		errorType = "solve"
+	case errs.Matches(err, &setup.BuildError{}):
+		errorType = "build"
+	case errs.Matches(err, &setup.ArtifactSetupErrors{}):
```
Naatan, line 202: The type you created is called `ArtifactInstallError`. I think SetupErrors is more appropriate though. `ArtifactInstallError` is assuming too many things are install errors. For example I think a download error would also be an install error by the current implementation.

pkg/platform/runtime/setup/setup.go
```diff
@@ -923,3 +939,11 @@ func (s *Setup) fetchAndInstallArtifactsFromDir(installFunc artifactInstaller) (
...
 	return installedArtifacts, <-aggregatedErr
 }
+
+func (s *Setup) handleEvent(ev events.Eventer) error {
+	err := s.eventHandler.Handle(ev)
+	if err != nil {
+		return &ProgressReportError{errs.Wrap(err, "Error handling event: %v", errs.JoinMessage(err))}
+	}
+	return nil
+}
```
Naatan, line 949: Clever! I like it. Keeps the progressReporter free of this concern.

Naatan (changes requested):

pkg/platform/runtime/runtime.go
mitchell-as: Right, all errors that happen during `fetchAndInstallArtifacts()` are aggregated into a single `ArtifactInstallError` object. You can only figure out we had an `ArtifactDownloadError` or `ArtifactInstallError` by iterating through the list of errors in that original object. Yes, an artifact download error is considered an install error, but I'm tagging it as a download error and identifying it in the subsequent loop here.

pkg/platform/runtime/setup/buildlog/buildlog.go
```diff
@@ -48,6 +48,16 @@ type Events interface {
 	Heartbeat(time.Time)
 }
 
+// BuildError designates a build log build error.
+type BuildError struct {
+	*locale.LocalizedError
+}
+
+// ProgressReportError designates an error in the event handler for reporting progress.
+type ProgressReportError struct {
```
Naatan, line 57: Can we name this "EventHandlerError"? The reason being the buildlog package has no concept or care of "progress reporting", all it cares about is invoking a handler.

pkg/platform/runtime/runtime.go
```diff
@@ -177,27 +178,23 @@ func (r *Runtime) recordCompletion(err error) {
...
 
-	var errorType string
+	errorType := "unknown"
 	switch {
+	// IsInputError should always be first because it is technically possible for something like a
+	// download error to be cause by an input error.
 	case locale.IsInputError(err):
 		errorType = "input"
-	case errs.Matches(err, &setup.ProgressReportError{}): // should come before non-input errors
+	// Progress errors should come before any other non-input error.
```
Naatan, line 192: Please explain why (in the comment).

pkg/platform/runtime/runtime.go
Naatan: ok, but my point was that the `ArtifactInstallError` type doesn't seem to be serving any purpose? It's only ever constructed here:

https://github.com/ActiveState/cli/pull/2605/files#diff-c415bd62e9cd64f30d585a034cd17ab1f283c67287d63a1448787a2a81efc043R215

But it's never inspected, so I'm not seeing what it is doing? 

Also note where it is constructed it is too aggressively doing so as we cannot assert at that point whether the failure was in fact an install failure as oppose to a download failure.

Naatan (changes requested):

mitchell-as:
Okay, sorry, I was wrong and misspoke about download errors being considered install errors. They are not. When an `ArtifactInstallError` is created, it is only as a result of the call to `installFunc()`, which is invoked to install a downloaded artifact (that is, after it was successfully downloaded). I don't believe it's possible to have an error be tagged as `ArtifactInstallError` and `ArtifactDownloadError` at the same time.

That said, I was being overly broad in labeling `ArtifactSetupErrors` as being either download or install errors, which is not accurate. I have changed the code to explicitly look for both download and install errors, and pick the errorType based on the search. I've also annotated the logic with comments which are hopefully helpful.

pkg/platform/runtime/runtime.go
mitchell-as: Good that you had me do that. These should actually come last :/ Sorry about that.

mitchell-as:
By the way, if you'd like to go over this on Zoom in order to avoid all the back and forth, we can.

Naatan (approved):

PR #2607, opened 2023-06-26
Made `state publish` meta file an optional flag instead of an optional argument

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1937" title="DX-1937" target="_blank">StoryDX-1937</a>  Better UX for uploading ingredient with meta flags
  </td></table>
  <br />


Naatan (changes requested):
Looks good! But please also update publish_int_test.go.

Naatan (approved):

mitchell-as:
Thanks for the reminder. This is another reason why working on feature branches is hard: can't run tests if the branch won't build/run by default.

PR #2611, opened 2023-06-28
Use inventory API's NormalizeNames endpoint before searching for a package to install.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1932" title="DX-1932" target="_blank">StoryDX-1932</a>  We normalize package names provided to `state install`
  </td></table>
  <br />
 

While there wasn't an issue for mismatched name case, it was an issue for mixed '-' and '_' for Python.

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -181,7 +181,7 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 			return locale.WrapInputError(err, "package_ingredient_alternatives", "", requirementName, strings.Join(suggestions, "\n"))
 		}
 
-		origRequirementName, requirementName = requirementName, normalized
+		requirementName = normalized
 
 		pg.Stop(locale.T("progress_found"))
 		pg = nil
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -157,10 +157,16 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 		requirementVersion = ""
 	}
 
+	origRequirementName := requirementName
 	if validatePkg {
 		pg = output.StartSpinner(out, locale.Tl("progress_search", "", requirementName), constants.TerminalAnimationInterval)
 
-		packages, err := model.SearchIngredientsStrict(ns, requirementName, false, false)
+		normalized, err := model.FetchNormalizedName(ns, requirementName)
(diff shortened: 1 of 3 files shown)


internal/runbits/requirements/requirements.go
```diff
@@ -157,10 +157,16 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
...
 	if validatePkg {
 		pg = output.StartSpinner(out, locale.Tl("progress_search", "", requirementName), constants.TerminalAnimationInterval)
 
-		packages, err := model.SearchIngredientsStrict(ns, requirementName, false, false)
+		normalized, err := model.FetchNormalizedName(ns, requirementName)
+		if err != nil {
+			multilog.Error("Failed to normalize '%s': %v", requirementName, err)
+		}
+
+		packages, err := model.SearchIngredientsStrict(ns, normalized, false, false) // ideally case-sensitive would be true (PB-4371)
```
mitchell-as, line 169: I will create a JIRA ticket to address this once this PR is merged so I have a concrete link to this line.

internal/runbits/requirements/requirements.go
```diff
@@ -174,10 +180,8 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 			}
 			return locale.WrapInputError(err, "package_ingredient_alternatives", "", requirementName, strings.Join(suggestions, "\n"))
 		}
-		if name := packages[0].Ingredient.Name; name != nil && requirementName != *name {
-			logging.Debug("Requirement to install's letter case differs from Platform's ('%s' != '%s')", requirementName, *name)
-			requirementName = *name // match case
-		}
+
+		origRequirementName, requirementName = requirementName, normalized
```
daved, line 184: Why does origRequirementName need to be assigned again?

daved (changes requested):

internal/runbits/requirements/requirements.go
mitchell-as: We need to keep track of what the user requested to be installed, and what was actually installed. That way we can notify the user that something was silently changed on them.

daved: The value doesn't seem to be different than when set at line 160.

daved (changes requested):

mitchell-as:
Yes, you are correct. Sorry about that.

daved (approved):

PR #2612, opened 2023-06-28
Initial implementation of `state projects move`

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1938" title="DX-1938" target="_blank">StoryDX-1938</a>  As a user I can change the owner of a project
  </td></table>
  <br />
 

This is a super-sized PR because running `state run generate-api-client` and `state run generate-inventory-client` brought in a lot of swagger changes.


internal/runners/projects/move.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	m.out.Notice(locale.Tl("move_success",
+		"Project [NOTICE]{{.V0}}[/RESET] successfully moved to the [NOTICE]{{.V1}}[/RESET] organization",
+		params.Namespace.Project, params.NewOwner))
+
+	return nil
+}
+
+func (m *Move) updateLocalCheckouts(params *MoveParams) error {
```
MDrakos, line 76: Is there any appetite for unifying this and the edit `updateLocalCheckouts` function somewhere like the `projectfile` package?

MDrakos (approved):
Looks good to me! Consider my question more of a nit or potential minor improvement.

internal/runners/projects/move.go
mitchell-as: I was hoping to do so, but the logic in the loops is slightly different, and the input parameters are different. Trying to shoehorn them into some sort of union didn't seem worth it and would probably cause more complexity than it's worth.

PR #2613, opened 2023-06-29
Allow partial version numbers for `state init`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1927" title="DX-1927" target="_blank">TaskDX-1927</a>  state init makes it painless to specify partial version numbers
  </td></table>
  <br />
 

Also remove `python3` and `python2` distinction.

### DIFF
--- a/internal/language/language.go
+++ b/internal/language/language.go
@@ -120,16 +120,12 @@ func MakeByName(name string) Language {
 
 // MakeByNameAndVersion will retrieve a language by a given name and version.
 func MakeByNameAndVersion(name, version string) (Language, error) {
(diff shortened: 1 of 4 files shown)


daved (changes requested):
There seems to be missing test cases. For example:
- `state install org/project --language python` installs Python 3.11.4
- `state install org/project --language python@3` installs Python 3.11.4
- `state install org/project --language python@3.10` installs Python 3.10.12

mitchell-as:
The ticket notes those are manual tests (mainly for QA, but I ran them locally to verify everything works as expected.)

daved (approved):

PR #2614, opened 2023-06-29
The BuildExpression parser uses the same types as the BuildPlanner

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1941" title="DX-1941" target="_blank">StoryDX-1941</a>  The Build Expression parser uses the same types as the Build Planner parser
  </td></table>
  <br />
 

This changeset is largely based on this file: https://github.com/ActiveState/cli/pull/2582/files#diff-5d76865da8b2a218196c43428aa29c557e43a72824821266c752b83468cc3b8e. Follow-up story to unify these types filed here: https://activestatef.atlassian.net/browse/DX-1961


pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -56,9 +69,66 @@ type Requirement struct {
...
+	Name      string
+	Arguments []*Value
+}
+
+type In struct {
+	FuncCall *Ap
+	Name     *string
+}
+
+type context []string
```
mitchell-as, line 110: Perhaps place a documentation comment here? I won't know what this is or what it's for until I look for its use.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -97,122 +167,254 @@ type BuildExpression struct {
...
-		err = bx.addRequirement(requirement)
-	case OperationRemoved:
-		err = bx.removeRequirement(requirement)
-	case OperationUpdated:
-		err = bx.updateRequirement(requirement)
-	default:
-		return errs.New("Unsupported operation")
+func newLet(ctx context, m map[string]interface{}) (*Let, error) {
+	ctx.push(ctxLet)
+	defer ctx.pop()
```
mitchell-as, line 203: Technically you're ignoring an error if the stack is empty. While I don't see how this could fail in this function, improper stack management in other functions may cause an error. Perhaps an error check is warranted.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -97,122 +167,254 @@ type BuildExpression struct {
...
+		return nil, errs.Wrap(err, "Could not parse 'let' key")
 	}
 
-	(*bx.solveNode)[AtTimeKey] = formatted
+	return &Let{Assignments: *assignments, In: in}, nil
+}
+
+func isAp(ctx context, value map[string]interface{}) bool {
+	ctx.push(ctxIsAp)
+	defer ctx.pop()
```
mitchell-as, line 228: Again, perhaps check for errors in this and all subsequent cases.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -97,122 +167,254 @@ type BuildExpression struct {
...
 
-	return nil
+	_, hasIn := value["in"]
+	if hasIn && !ctx.contains(ctxAssignments) {
```
mitchell-as, line 231: Is this check dependent on order? i.e. is there a certain key processing order that would cause an invalid result? I don't know enough about build expressions to determine this, but the fact that you're dependent on the stack for the check leads me to think that pushing/popping at unexpected times may lead to unexpected results.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -97,122 +167,254 @@ type BuildExpression struct {
...
+	switch v := inValue.(type) {
+	case map[string]interface{}:
+		f, err := newAp(ctx, v)
+		if err != nil {
+			return nil, errs.Wrap(err, "'in' object is not a function call")
+		}
+		in.FuncCall = f
+
+	case string:
+		in.Name = p.StrP(strings.TrimPrefix(v, "$"))
```
mitchell-as, line 371: Out of curiosity, what is '$' for? I couldn't figure out what '$' was doing, so I was stripping it/adding it as necessary when processing build scripts without actually knowing if what I was doing was correct.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -232,37 +434,72 @@ func (bx BuildExpression) MarshalJSON() ([]byte, error) {
...
 
-		err = validateRequirements(node)
-		if err != nil {
-			return nil, errs.Wrap(err, "Requirements in BuildExpression are invalid")
+	return reqs
+}
+
+func getVersionRequirements(v *[]*Value) []VersionRequirement {
+	var reqs []VersionRequirement
+	for _, r := range *v {
```
mitchell-as, line 485: Can `v` be nil? If so, it'll panic.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -278,49 +515,233 @@ func getRequirements(solveNode map[string]interface{}) ([]Requirement, error) {
...
-//	    }
-//	  }
-//	}
-func getFuncNode(expression map[string]interface{}, funcName string) (map[string]interface{}, error) {
-	for k, v := range expression {
-		node, ok := v.(map[string]interface{})
-		if !ok {
+func (e *BuildExpression) addRequirement(requirement Requirement) error {
+	obj := []*Var{
+		{Name: RequirementNameKey, Value: &Value{Str: p.StrP(requirement.Name)}},
```
mitchell-as, line 559: Does Go auto-convert this struct to a pointer-to-struct? I've been doing:

```
[]*Var{
  &Var{Name: name, Value: value},
}
```

all this time. If what you're doing is fine, then I can save myself some typing!

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -278,49 +515,233 @@ func getRequirements(solveNode map[string]interface{}) ([]Requirement, error) {
...
+		{Name: RequirementNamespaceKey, Value: &Value{Str: p.StrP(requirement.Namespace)}},
+	}
+
+	if requirement.VersionRequirement != nil {
+		for _, r := range requirement.VersionRequirement {
+			obj = append(obj, &Var{Name: RequirementVersionRequirementsKey, Value: &Value{List: &[]*Value{
+				{Object: &[]*Var{
+					{Name: RequirementComparatorKey, Value: &Value{Str: p.StrP(r[RequirementComparatorKey])}},
+					{Name: RequirementVersionKey, Value: &Value{Str: p.StrP(r[RequirementVersionKey])}},
```
mitchell-as, line 568: Note: it might be worth merging the v40 branch into this one sometime soon because the `p` package and its API has changed significantly. This is completely subjective though.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -278,49 +515,233 @@ func getRequirements(solveNode map[string]interface{}) ([]Requirement, error) {
...
+				}}},
+			}})
+		}
+	}
+
+	requirementsNode := append(e.getRequirementsNode(), &Value{Object: &obj})
+
+	for _, arg := range e.getSolveNode().Arguments {
```
mitchell-as, line 576: `getSolveNode()` returns a pointer. If it can be `nil`, it'll panic here and in subsequent instances of this construct. Can we guarantee a non-nil pointer will be returned? If so, maybe append a comment like `// guaranteed to exist`. Or, better yet, create a new function `getSolveNodeArguments()` that returns a concrete list to range over (empty if the solve node or its arguments are `nil`).

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -278,49 +515,233 @@ func getRequirements(solveNode map[string]interface{}) ([]Requirement, error) {
...
+	case v.Object != nil:
+		m := make(map[string]interface{})
+		for _, assignment := range *v.Object {
+			m[assignment.Name] = assignment.Value
+		}
+		return json.Marshal(m)
+	case v.Ident != nil:
+		return json.Marshal(v.Ident)
+	}
+	return json.Marshal([]*Value{}) // participle does not create v.List if it's empty
```
mitchell-as, line 721: Maybe update the comment to not mention participle, as it's not used here.

mitchell-as (changes requested):
Some things to address, but this looks pretty good to me.

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: Yes, you're right however I don't think the `pop` function needs to return an error or a string. We really only want to inspect what's in the stack with `contains`

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: I _believe_ that build expressions only support function calls as a result of an assignment, so something like `runtime = solve()` is what we're checking for. You're right that it would fail if there were unexpected pushes and pops but I don't think that's possible. I will double-check that my assumption is correct though.

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: It's to reference a variable. When `runtime` is assigned it's not there but for any reference, it should be.

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: Unlikely but still possible.

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: Yep, it's part of how composite literals work in Go. Some more information [here](https://go.dev/ref/spec#Composite_literals) if you're interested. There is an example near the end of that section that pertains to this usage.

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: I like the `getSolveNodeArguments()` idea!

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: Good point, I noticed that as well. I think that should be part of the integration story.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -107,19 +107,20 @@ type In struct {
...
 
 func (s *context) push(str string) {
 	*s = append(*s, str)
 }
 
-func (s *context) pop() (string, error) {
+func (s *context) pop() {
 	if len(*s) == 0 {
-		return "", errs.New("stack is empty")
+		return
```
mitchell-as, line 120: Do you think we should have a `multilog.Error()` here? If we're doing some stack mismanagement, we'd probably want to know about it.

mitchell-as (approved):
Approved with one thing to think about.

MDrakos:
@Naatan This code is a slightly modified version of Mitchell's build expression parser which has already been reviewed by both you and David. I just want to get your eyes on my changes, especially this `context` type that I've introduced [here](https://github.com/ActiveState/cli/pull/2614/files#diff-8c58a61719419697e1e9a73761423bba4b49182d89b8a94868d8a4384acb2ae7R113). It's fairly simple and works for our use-case but there are some pitfalls. Maybe you have a better idea of how we can record the context?

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -24,12 +28,22 @@ const (
...
+	RequirementVersionRequirementsKey = "version_requirements"
+	RequirementVersionKey             = "version"
+	RequirementComparatorKey          = "comparator"
+
+	ctxLet         = "let"
+	ctxIn          = "in"
+	ctxAp          = "ap"
+	ctxValue       = "value"
+	ctxAssignments = "assignments"
+	ctxIsAp        = "isAp"
```
Naatan, line 46: I feel like the logic for walking the AST should not be part of the model package. The model package should be about marshalling/unmarshalling the data at a high level.

Compare it to for example the json or yaml packages. Both these packages are inevitably used from inside a model package, but they aren't themselves models.

So please keep the model/buildexpression package limited to high level boilerplate for packing and unpacking the data retrieved from API's, but find a more appropriate place for the new AST walking logic.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -56,9 +70,68 @@ type Requirement struct {
...
+}
+
+func (s *context) contains(str string) bool {
+	for _, v := range *s {
+		if v == str {
+			return true
+		}
+	}
+	return false
 }
```
Naatan, line 135: This feels like some convenience helper function rather than anything to do with build expressions. I'd rather we use the native types and write out the slightly more complex push/pop logic. This feels like we're trying to replicate the API of another language inside Golang.

Another option is to use the sliceutils and strutils packages. Though I think really the only thing you might need is strutils.GetString to replace the `contains()` function. The other two IMO don't need abstraction.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -97,122 +170,255 @@ type BuildExpression struct {
 //	  }
 //	}
 func NewBuildExpression(data []byte) (*BuildExpression, error) {
-	expression := make(map[string]interface{})
-	err := json.Unmarshal(data, &expression)
+	m := make(map[string]interface{})
```
Naatan, line 173: Please use words not letters. Letters are fine for very short pieces of code where the intend of the variable is very evident, I don't think this qualifies.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -278,49 +524,242 @@ func getRequirements(solveNode map[string]interface{}) ([]Requirement, error) {
...
+
+func (e *BuildExpression) addRequirement(requirement Requirement) error {
+	obj := []*Var{
+		{Name: RequirementNameKey, Value: &Value{Str: p.StrP(requirement.Name)}},
+		{Name: RequirementNamespaceKey, Value: &Value{Str: p.StrP(requirement.Namespace)}},
+	}
+
+	if requirement.VersionRequirement != nil {
+		for _, r := range requirement.VersionRequirement {
+			obj = append(obj, &Var{Name: RequirementVersionRequirementsKey, Value: &Value{List: &[]*Value{
```
Naatan, line 583: This feels like we're missing an abstraction layer. I don't think the model code should be this concerned with the markup of the build expression.

pkg/platform/api/buildplanner/model/buildexpression.go
```diff
@@ -278,49 +524,242 @@ func getRequirements(solveNode map[string]interface{}) ([]Requirement, error) {
...
+	case v.Ap != nil:
+		return json.Marshal(v.Ap)
+	case v.List != nil:
+		return json.Marshal(v.List)
+	case v.Str != nil:
+		return json.Marshal(strings.Trim(*v.Str, `"`))
+	case v.Null != nil:
+		return json.Marshal(nil)
+	case v.Assignment != nil:
+		return json.Marshal(v.Assignment)
```
Naatan, line 729: Can't these be addressed with an `OmitEmpty`?

Naatan (changes requested):

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: No, we're not unmarshaling to these types. We first unmarshal to a map and then the "parser" iterates through the map to populate these values.

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: Does moving the parsing logic to its own package address this? Otherwise, I may be missing the point of this comment.

pkg/platform/api/buildplanner/model/buildexpression.go
MDrakos: Decided to go with a new set of functions in `sliceutils`

internal/sliceutils/sliceutils.go
```diff
@@ -47,3 +50,25 @@ func intsContain(ns []int, v int) bool {
 func InsertStringAt(data []string, i int, v string) []string {
 	return append(data[:i], append([]string{v}, data[i:]...)...)
 }
+
+func Push[T any](data []T, v T) []T {
+	return append(data, v)
+}
```
Naatan, line 56: Sorry, but this one is just replacing a native language API with effectively an alias that's opinionated towards a different language API. There's no new logic here. Please drop.

internal/sliceutils/sliceutils.go
```diff
@@ -47,3 +50,25 @@ func intsContain(ns []int, v int) bool {
...
+
+func Pop[T any](data []T) (T, []T, error) {
+	var t T
+	if len(data) == 0 {
+		return t, nil, errs.New("Cannot pop from empty slice")
+	}
+
+	return data[len(data)-1], data[:len(data)-1], nil
```
Naatan, line 64: Nit: Seeing as the slice is modified in memory there doesn't seem to be any utility in having it return the slice (second arg).

Naatan (changes requested):
I think the buildexpression is probably still doing a bit too much, but for this iteration let's not get hung up on that. 

Just asking for changes on the new sliceutil functions.

internal/sliceutils/sliceutils.go
MDrakos: I believe that because the size of the underlying array is changing the slice isn't modified outside of the function. I was messing around trying to get it to work but it would require passing a pointer to the slice.

Naatan: huh, interesting.. I wasn't aware of that behaviour.

Naatan (approved):

PR #2617, opened 2023-07-05
Rename `state stage` to `state commit`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1948" title="DX-1948" target="_blank">TaskDX-1948</a>  Rename "stage" term to "commit"
  </td></table>
  <br />

### DIFF
--- a/internal/runners/commit/commit.go
+++ b/internal/runners/commit/commit.go
@@ -40,12 +40,12 @@ func New(p primeable) *Commit {
 	}
 }
 
-func (s *Commit) Run() error {
-	if s.proj == nil {
+func (c *Commit) Run() error {
+	if c.proj == nil {
 		return locale.NewInputError("err_no_project")
 	}
 
-	changesCommitted, err := buildscript.Sync(s.proj, nil, s.out, s.auth)
+	changesCommitted, err := buildscript.Sync(c.proj, nil, c.out, c.auth)
 	if err != nil {
 		return locale.WrapError(
 			err, "err_commit_sync_buildscript",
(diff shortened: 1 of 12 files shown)


test/integration/commit_int_test.go
```diff
@@ -27,7 +27,7 @@ func (suite *StageIntegrationTestSuite) TestCommitManualBuildScriptMod() {
 	cp := ts.SpawnWithOpts(
 		e2e.WithArgs(
 			"checkout",
-			"ActiveState-CLI/Stage-Test-A#7a1b416e-c17f-4d4a-9e27-cbad9e8f5655",
+			"ActiveState-CLI/Commit-Test-A#7a1b416e-c17f-4d4a-9e27-cbad9e8f5655",
```
mitchell-as, line 30: I've also renamed this Platform project.

internal/runners/commit/commit.go
```diff
@@ -40,34 +40,34 @@ func New(p primeable) *Stage {
 	}
 }
 
-func (s *Stage) Run() error {
+func (s *Commit) Run() error {
```
daved, line 43: Please rename the receiver to either `c` (for "commit") or `r` (for "runner" - far less preferable in my opinion).

daved (changes requested):

internal/runners/commit/commit.go
mitchell-as: Thanks for pointing this out.

daved (approved):

PR #2620, opened 2023-07-06
Postpone uninstalling of outdated artifacts until the last minute.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1933" title="DX-1933" target="_blank">StoryDX-1933</a>  State tool should install to a temp dir and then only move it to actual runtime dir when the entire installation completes
  </td></table>
  <br />
 

If outdated artifacts are deleted during the fetch and prepare stage, the runtime could become corrupted. Wait until just before installing updated artifacts.

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -151,6 +151,7 @@ type ArtifactSetuper interface {
 }
 
 type artifactInstaller func(artifact.ArtifactID, string, ArtifactSetuper) error
+type artifactUninstaller func() error
 
 // New returns a new Setup instance that can install a Runtime locally on the machine.
 func New(target Targeter, eventHandler events.Handler, auth *authentication.Auth, an analytics.Dispatcher) *Setup {
@@ -213,7 +214,7 @@ func (s *Setup) updateArtifacts() ([]artifact.ArtifactID, error) {
 	// Fetch and install each runtime artifact.
 	// Note: despite the name, we are "pre-installing" the artifacts to a temporary location.
 	// Once all artifacts are fetched, unpacked, and prepared, final installation occurs.
-	artifacts, err := s.fetchAndInstallArtifacts(func(a artifact.ArtifactID, archivePath string, as ArtifactSetuper) (rerr error) {
+	artifacts, uninstallFunc, err := s.fetchAndInstallArtifacts(func(a artifact.ArtifactID, archivePath string, as ArtifactSetuper) (rerr error) {
 		defer func() {
 			if rerr != nil {
 				rerr = &ArtifactInstallError{errs.Wrap(rerr, "Unable to install artifact")}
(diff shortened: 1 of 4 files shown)


pkg/platform/runtime/setup/setup.go
```diff
@@ -968,3 +976,20 @@ func (s *Setup) handleEvent(ev events.Eventer) error {
...
+		return locale.WrapError(err, "err_stored_artifacts")
+	}
+
+	err = setup.DeleteOutdatedArtifacts(changedArtifacts, storedArtifacts, alreadyInstalled)
+	if err != nil {
+		multilog.Error("Could not delete outdated artifacts: %v, falling back to removing everything", err)
+		err = os.RemoveAll(s.store.InstallPath())
+		if err != nil {
+			return errs.Wrap(err, "Failed to clean installation path")
+		}
```
Naatan, line 992:
```suggestion
		// This multilog is technically redundant and may be dropped after we can collect data on this error for a while as rollbar is not surfacing the returned error
		multilog.Error("Could not delete outdated artifacts: %s", errs.JoinMessage(err))
		return errs.Wrap(err, "Could not delete outdated artifacts")
```

This code feels really weird. We shouldn't be using such a fallback IMO. Looking at the code I don't see any reason why DeleteOutdatedArtifacts would fail whereas RemoveAll would succeed, it makes no sense. But looking at rollbar I can't find any records of "Failed to clean installation path".

There are a total of 80 occurrences of the "Could not delete outdated artifacts" error over 6 months. I feel like we're somehow loosing the "Failed to clean installation path" error on rollbar rather than that it's not occurring, so I'm inclined to say we should drop this RemoveAll and address the actual error, but seeing as we might be missing it I would suggest dropping the RemoveAll but keeping the multilog here and adding an `errs.JoinMessage(err)` so we have a better idea of why it failed. Odds are this is user input.

test/integration/runtime_int_test.go
```diff
@@ -102,7 +102,7 @@ func (suite *RuntimeIntegrationTestSuite) TestInterruptSetup() {
 		e2e.AppendEnv("ACTIVESTATE_CLI_DISABLE_RUNTIME=false",
 			"ACTIVESTATE_CLI_RUNTIME_SETUP_WAIT=true"),
 	)
-	time.Sleep(10 * time.Second)
+	time.Sleep(30 * time.Second)
```
Naatan, line 105: Is this related to the changes? I wouldn't expect these changes to affect installation time.

Naatan (changes requested):

test/integration/runtime_int_test.go
mitchell-as: I did not realize that the integration tests utilize a 70MB download. It was cached locally on my machine, so I did not account for download time. It is likely this test is only interrupting a download, rather than a pre-install. I want to give it a better chance of testing the right thing.

Naatan (changes requested):
Please include the comment I added in my suggestion:

https://github.com/ActiveState/cli/pull/2620#discussion_r1256103647

mitchell-as:
Whoops, sorry about that :(

Naatan (approved):

PR #2621, opened 2023-07-07
Ensure valid installation paths

MDrakos:


internal/installation/paths.go
```diff
@@ -38,6 +38,10 @@ func InstallRoot(path string) (string, error) {
 		return "", errs.Wrap(err, "Could not find install marker file in path")
 	}
 
+	if !isValidInstallPath(filepath.Dir(installFile)) {
```
mitchell-as, line 41: Will this affect our local `build/` roots? i.e. will our local build directories considered to be invalid with this change?

internal/installation/paths_windows.go
```diff
@@ -3,10 +3,23 @@ package installation
...
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/errs"
 )
 
 func installPathForBranch(branch string) (string, error) {
-	return filepath.Join(os.Getenv("USERPROFILE"), "AppData", "Local", "ActiveState", "StateTool", branch), nil
+	home := os.Getenv("USERPROFILE")
+	if dir := os.Getenv(constants.HomeEnvVarName); dir != "" {
+		home = dir
+	}
```
mitchell-as, line 15: Please refactor this to use the `internal/osutils/user` package's `HomeDir()` function to get the user's home directory.

mitchell-as (changes requested):

internal/installation/paths.go
MDrakos: No, we will hit the code that facilitates that use case before validating the install path.

internal/osutils/user/user_win.go
```diff
@@ -0,0 +1,23 @@
+//go:build windows
```
mitchell-as, line 1: This is not necessary. Go's `os.UserHomeDir()` accounts for Windows, MacOS, and Linux.

mitchell-as (changes requested):

mitchell-as (approved):

PR #2623, opened 2023-07-10
Initial support for auto-merge of build expressions into build scripts (local orders)

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1912" title="DX-1912" target="_blank">StoryDX-1912</a>  We automatically merge local order changes that are compatible
  </td></table>
  <br />
 

Some of this PR is fixes to handle build plan feature branch updates that I needed to pull in for testing purposes.


internal/runbits/buildscript/buildscript_test.go
```diff
@@ -33,10 +33,10 @@ in:
 
 	bytes, err := json.Marshal(script)
 	require.NoError(t, err)
-	expr, err := model.NewBuildExpression(bytes)
+	expr, err := buildexpression.New(bytes)
 	require.NoError(t, err)
 
-	(*script.Let.Assignments[0].Value.FuncCall.Arguments[0].Assignment.Value.List)[0].Str = p.StrP(`"77777"`)
+	(*script.Let.Assignments[0].Value.FuncCall.Arguments[0].Assignment.Value.List)[0].Str = ptr.To(`"77777"`)
```
Naatan, line 39: 😅

internal/runbits/buildscript/merge.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/ActiveState/cli/pkg/platform/api/mono/mono_models"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
+	"github.com/ActiveState/cli/pkg/platform/runtime/buildscript"
+	"github.com/ActiveState/cli/pkg/project"
+	"github.com/go-openapi/strfmt"
+)
+
+func Merge(proj *project.Project, remoteCommit strfmt.UUID, strategies *mono_models.MergeStrategies, auth *authentication.Auth) error {
```
Naatan, line 21: This feels like it's doing too much for a runbit. Can we move the merge specific logic into its own package (or maybe make it part of the buildexpression?). The runbit shouldn't contain business logic, just direct input and business logic.

I'll defer the rest of the review for the next iteration as I imagine this change would be significant.

Naatan (changes requested):

internal/runbits/buildscript/merge.go
mitchell-as: Sure. I moved it into a subpackage of buildexpression to avoid an import cycle.

pkg/platform/runtime/buildexpression/merge/merge.go
```diff
@@ -0,0 +1,132 @@
...
+			data, err := constraint.MarshalBinary()
+			if err != nil {
+				return nil, errs.Wrap(err, "Could not marshal requirement version constraints")
+			}
+			m := make(map[string]string)
+			err = json.Unmarshal(data, &m)
+			if err != nil {
+				return nil, errs.Wrap(err, "Could not unmarshal requirement version constraints")
+			}
+			versionRequirements = append(versionRequirements, m)
```
Naatan, line 59: This is a neat solution! But we may be able to keep it simple, the `version_constraint` field holds the already marshalled values, and according to dashboard we're safe to use this despite it being marked deprecated:

https://activestate.slack.com/archives/CKMUXF64Q/p1689281817627129

pkg/platform/runtime/buildexpression/merge/merge.go
```diff
@@ -0,0 +1,132 @@
...
+			return true
+		}
+		if m2, ok := v.(map[string]interface{}); ok {
+			if deleteRequirements(&m2) {
+				return true
+			}
+		}
+	}
+	return false
+}
```
Naatan, line 132: @MDrakos I think this won't work when we have installers or python wheels right? Can you share some insight?

pkg/platform/runtime/buildscript/buildexpression.go
```diff
@@ -206,6 +207,11 @@ func (s *Script) EqualsBuildExpression(otherJson []byte) bool {
...
 
-func (s *Script) Equals(other *model.BuildExpression) bool {
-	return s.EqualsBuildExpression([]byte(other.String()))
+func (s *Script) Equals(other *buildexpression.BuildExpression) bool {
+	data, err := json.Marshal(other)
+	if err != nil {
+		multilog.Error("Unable to marshal buildexpression to JSON: %v", err)
+		return false
+	}
+	return s.EqualsBuildExpression(data)
```
Naatan, line 216: Nit: not technically introduced in this PR but the naming of `Equals()` and `EqualsBuildExpression` feels inverted to me.

Naatan (changes requested):
This is looking really good, just some slight changes and a question for Mike to weigh in on.

internal/runbits/buildscript/merge.go
```diff
@@ -0,0 +1,58 @@
...
+	// Get the local and remote build expressions to merge.
+	bp := model.NewBuildPlannerModel(auth)
+	localCommit, err := localcommit.Get(proj.Dir())
+	if err != nil {
+		return errs.Wrap(err, "Unable to get local commit")
+	}
+	exprA, err := bp.GetBuildExpression(proj.Owner(), proj.Name(), localCommit.String())
+	if err != nil {
+		return errs.Wrap(err, "Unable to get buildexpression for local commit")
+	}
```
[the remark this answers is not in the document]
mitchell-as: The reason is that the build expression might contain data that is not in the build script. I don't know exactly what that is, but I wanted to err on the side of caution here.

internal/runbits/buildscript/merge.go
```diff
@@ -0,0 +1,58 @@
...
+// UUID, performing the given merge strategy (e.g. from model.MergeCommit).
+func Merge(proj *project.Project, remoteCommit strfmt.UUID, strategies *mono_models.MergeStrategies, auth *authentication.Auth) error {
+	// Verify we have a build script to merge.
+	script, err := buildscript.NewScriptFromProjectDir(proj.Dir())
+	if err != nil && !buildscript.IsDoesNotExistError(err) {
+		return errs.Wrap(err, "Could not get local build script")
+	}
+	if script == nil {
+		return nil // no build script to merge
+	}
```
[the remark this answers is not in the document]
mitchell-as: Maybe the user previously made a local modification, tried to pull, got a conflict that they did not want to resolve, and deleted the local build script in the hope that the remote one would be pulled in automatically (which it will eventually).

pkg/platform/runtime/buildexpression/merge/merge.go
```diff
@@ -0,0 +1,132 @@
...
+	return errs.Matches(err, &MergeConflictsError{})
+}
+
+func Merge(exprA *buildexpression.BuildExpression, exprB *buildexpression.BuildExpression, strategies *mono_models.MergeStrategies) (*buildexpression.BuildExpression, error) {
+	if !isAutoMergePossible(exprA, exprB) {
+		return nil, &AutoMergeNotPossibleError{errs.New("Unable to merge buildexpressions")}
+	}
+	if len(strategies.Conflicts) > 0 {
+		return nil, &MergeConflictsError{errs.New("Unable to merge buildexpressions due to conflicting requirements")}
+	}
```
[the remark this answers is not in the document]
mitchell-as: No, we may eventually support partial auto-merges (just not in this ticket). `IsAutoMergePossible()` returns false if something **other** than requirements is different (the contents of `strategies` is irrelevant), and in that case, we bail out no matter what.

pkg/platform/runtime/buildexpression/merge/merge.go
```diff
@@ -0,0 +1,132 @@
...
+		switch req.Operation {
+		case mono_models.CommitChangeEditableOperationAdded:
+			op = bpModel.OperationAdded
+		case mono_models.CommitChangeEditableOperationRemoved:
+			op = bpModel.OperationRemoved
+		case mono_models.CommitChangeEditableOperationUpdated:
+			op = bpModel.OperationUpdated
+		default:
+			return nil, errs.New("Unknown requirement operation: %s", op)
+		}
```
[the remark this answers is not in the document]
mitchell-as: I'm not sure about that. We've been given a mono API string/constant and we're trying to map it to a buildplanner API string/constant. If I tried to do the requested unmarshaling, wouldn't the buildplanner API have to know about the mono API? I thought marshaling/unmarshaling for types was restricted to their own enties/quantities, not outside ones.

pkg/platform/runtime/buildexpression/merge/merge.go
mitchell-as: Ah, sadly `bpModel.VersionRequirements` is a `map[string]string`, and `req.VersionConstraint` is just a string (e.g. "=1.2.3"). The build planner is expecting something like `{"comparator": "eq", "version": "1.2.3"}`, so extracting the data from `req.VersionConstraints` (which can be tricked into to a `map[string]string`) is easier than interpreting `req.VersionConstraint` and creating a map from it.

pkg/platform/runtime/buildscript/buildexpression.go
mitchell-as: The naming has always bugged me. I've used even more explicit naming.

internal/runbits/buildscript/merge.go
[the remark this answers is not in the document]
Naatan: I'd rather we handle this as an explicit use-case though. For now, I think it's appropriate for us to error out on this case.

I've been bit many times introducing "features" that weren't requested but I figured they were easy enough to facilitate while I was at it.

internal/runbits/buildscript/merge.go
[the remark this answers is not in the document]
Naatan: Actually that just convinced me that we should use the local buildscript instead. If there are differences then it will lead to a conflict, and the user can resolve those differences.

With the current implementation, it seems that if we encounter this scenario then we would silently blow away the users customizations.

As you say it shouldn't happen, but if we're going to err on the side of safety then it feels more safe to use the local buildscript.

pkg/platform/runtime/buildexpression/merge/merge.go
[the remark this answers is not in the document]
Naatan: Dealing with the API types and translating them to local types is precisely the intend of the model package. Although usually, it's about converting API types to local types and not vice-versa; there isn't any rule against that.

internal/runners/pull/pull.go
```diff
@@ -212,6 +214,45 @@ func (p *Pull) performMerge(strategies *mono_models.MergeStrategies, remoteCommi
...
+	// Attempt the merge.
+	mergedExpr, err := merge.Merge(exprA, exprB, strategies)
+	if err != nil {
+		err := buildscriptRunbits.GenerateAndWriteDiff(p.project, script, exprB)
+		if err != nil {
+			return locale.WrapError(err, "err_diff_build_script", "Unable to generate differences between local and remote build script")
+		}
+		return locale.NewInputError(
+			"err_build_script_merge",
+			"Unable to automatically merge build scripts. Please resolve conflicts manually and then run [ACTIONABLE]`state commit`[/RESET]")
```
Naatan, line 249: Please specify what file they need to resolve the conflicts in. I would make a code suggestion but I don't actually remember what we called it for the interim.

Naatan (changes requested):
Also, see responses to previous comments.

internal/runbits/buildscript/merge.go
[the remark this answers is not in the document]
mitchell-as: What about the case where the user has an existing project that does not have a build script/build plan because this is the first time they are pulling after build plans were introduced? In that case, every initial pull would be an error.

I'm just thinking there are legitimate reasons why a build script may not exist, and so the default should be to accept the remote/incoming one; no merge is needed.

mitchell-as:
I've applied the requested change. It's a bigger diff than expected because I took the opportunity to simplify conversions and comparisons between buildscripts and buildexpressions. Previously, buildexpression structs were constantly changing and not easily converted to/from JSON, so it was easier to work directly with buildexpression JSON. Now that buildexpressions are more solidified and have marshaler/unmarshaler JSON interfaces, we can now work with buildexpression structs directly more easily.

internal/runbits/buildscript/merge.go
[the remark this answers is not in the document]
Naatan: That's a great use-case to consider, but luckily one that we already have covered with:

https://github.com/ActiveState/cli/blob/a2e9a48e6581ceda1ed08eff601cdba015cf0a9d/pkg/platform/runtime/buildscript/file.go#L36-L36

This is called from the pull runner as well as other runners. So the scenario you're covering here still should not happen and we should make it an error condition as a result, imo.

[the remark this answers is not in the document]
Naatan: I just realized this is only called at the end. So your scenario could still play out. I'm filing a story to find a better solution here as I don't want our code to have to be aware of this scenario at every turn if we can help it.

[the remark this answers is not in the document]
mitchell-as: Okay, done.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -79,6 +80,20 @@ func (o Operation) String() string {
 	}
 }
 
+func (o *Operation) Unmarshal(v string) error {
```
Naatan, line 83: I was hoping this would be possible without the pointer, but I guess not with string values.

Considering this, maybe change this to a `NewOperation(v string) error`? It's awkward having to create a pointer first just to unmarshal.

Naatan (changes requested):
Love the marshal/unmarshal changes, much more readable!

Just two minor changes (check comment response here: https://github.com/ActiveState/cli/pull/2623#discussion_r1263923170)

pkg/platform/api/buildplanner/model/buildplan.go
mitchell-as: Notice that I didn't actually have to create a pointer:

```
		var op bpModel.Operation
		err := op.Unmarshal(req.Operation)
```

I ran it through a debugger and the underlying value changes as expected.

It's actually more awkward to create a pointer-to-int in a `NewOperation()` function than to work with an Operation instance in the example I gave.

Naatan: Huh ok, interesting. I was not aware this was possible. TIL!

The only problem my suggestion was meant to solve is having to create a pointer var, so if that's a non-issue we can disregard my suggestion.

Naatan (approved):

PR #2624, opened 2023-07-10
Add example runner

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1951" title="DX-1951" target="_blank">TaskDX-1951</a>  Introduce example runner that is only meant for devx developers to reference
  </td></table>
  <br />


daved:
@Naatan Would you like comments and/or tests?

Naatan (changes requested):
Unflagging myself for review since you said you were still working on this.

daved:
@Naatan There's an old saying in MVC, "Thin controllers, fat models.", and I think it makes a lot of sense if we can aspire to making Run funcs (which are our controllers/handlers) as thin as possible so that first reads of a runner are fast and clear. So, I broke down the `state hello info` command into sub-handlers. I've done this to be a bit provocative, but am happy to eliminate it from the example if you don't like it or prefer to defer that sort of rethink for a different story.

internal/runbits/hello_example.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"errors"
+
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+)
+
+func SayHello(out output.Outputer, to string) error {
+	if to == "" {
+		return errors.New("test")
```
Naatan, line 12: Shouldn't this be an input error?

internal/runners/hello/info_example.go
```diff
@@ -0,0 +1,135 @@
+package hello
```
Naatan, line 1: This runner code is too opinionated. The point of the example is to show basic standards. I think a lot of what is happening here isn't standards so much as how you individually would approach the problem. Which is totally fair for an individual runner you authored, but not something I want to portray as a standard.

Anyway, we don't need two example runners. One is enough. Suggest getting rid of this one.

internal/runners/hello/hello_example.go
```diff
@@ -0,0 +1,63 @@
...
+// New contains the scope in which an instance of Hello is constructed from an
+// implementation of primeable.
+func New(p primeable) *Hello {
+	return &Hello{
+		out: p.Output(),
+	}
+}
+
+// Run contains the scope in which the hello runner logic is executed.
+func (h *Hello) Run(params *RunParams) error {
```
Naatan, line 53: One thing we often trip over is having a project primeable but forgetting to nil check it. It would be nice if the example runner could handle this case.

cmd/state/internal/cmdtree/hello_example.go
```diff
@@ -0,0 +1,75 @@
+package cmdtree
```
Naatan, line 1: Please exclude the example stuff from being built on CI. We don't want to ship these, just have them serve as code examples when developing locally.

cmd/state/internal/cmdtree/hello_example.go
```diff
@@ -0,0 +1,75 @@
...
+
+func newHelloCommand(prime *primer.Values) *captain.Command {
+	runner := hello.New(prime)
+
+	params := hello.NewRunParams()
+
+	cmd := captain.NewCommand(
+		"hello",
+		locale.Tl("hello_cmd_title", "Saying hello"),
+		locale.Tl("hello_cmd_description", "An example command"),
```
Naatan, line 18:
```suggestion
		// The description is shown on --help output
		locale.Tl("hello_cmd_description", "An example command"),
```

cmd/state/internal/cmdtree/hello_example.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/ActiveState/cli/internal/runners/hello"
+)
+
+func newHelloCommand(prime *primer.Values) *captain.Command {
+	runner := hello.New(prime)
+
+	params := hello.NewRunParams()
+
+	cmd := captain.NewCommand(
+		"hello",
```
Naatan, line 16:
```suggestion
		// The command's name should not be localized as we want commands to behave consistently regardless of localization.
		"hello",
```

cmd/state/internal/cmdtree/hello_example.go
```diff
@@ -0,0 +1,75 @@
...
+		locale.Tl("hello_cmd_title", "Saying hello"),
+		locale.Tl("hello_cmd_description", "An example command"),
+		prime,
+		[]*captain.Flag{},
```
Naatan, line 20: Can you add an example flag and add a command similar to the command name on saying it should not be localized?

Naatan: Maybe a good opportunity to showcase custom flag types..

cmd/state/internal/cmdtree/hello_example.go
```diff
@@ -0,0 +1,75 @@
...
+				),
+				Value: &params.Named,
+			},
+		},
+		func(_ *captain.Command, _ []string) error {
+			return runner.Run(params)
+		},
+	)
+
+	cmd.SetGroup(UtilsGroup)
```
Naatan, line 36:
```suggestion
	// The group is used to group together commands in the --help output
	cmd.SetGroup(UtilsGroup)
```

cmd/state/internal/cmdtree/hello_example.go
```diff
@@ -0,0 +1,75 @@
...
+	// cmd.SetUnstable(true)
```
Naatan, line 37:
```suggestion
	// Any new command should be marked unstable for the first release it goes out in.
	// cmd.SetUnstable(true)
```

cmd/state/internal/cmdtree/hello_example.go
```diff
@@ -0,0 +1,75 @@
...
+	// cmd.SetHidden(true)
```
Naatan, line 38:
```suggestion
	// Certain commands like `state deploy` are there for backwards compatibility, but we don't want to show them in the --help output as they are not part of the happy path or our long term goals.
	// cmd.SetHidden(true)
```

internal/runners/hello/hello_example.go
```diff
@@ -0,0 +1,63 @@
...
+	// Reusable runner logic is contained within the runbits package.
```
Naatan, line 54:
```suggestion
	// Reusable runner logic is contained within the runbits package.
	// You should only use this if you intend to share logic between runners. Runners should NEVER invoke other runners.
```

internal/runners/hello/hello_example.go
```diff
@@ -0,0 +1,63 @@
...
+	if err := runbits.SayHello(h.out, params.Named); err != nil {
+		// Errors should nearly always be localized.
+		return locale.WrapError(
```
Naatan, line 57: Should be an input error (would be good to maybe handle both cases to point out the difference).

internal/runbits/hello_example.go
```diff
@@ -0,0 +1,16 @@
+package runbits
+
+import (
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+)
+
+func SayHello(out output.Outputer, name string) error {
+	if name == "" {
+		return locale.NewInputError("hello_err_no_name", "No name provided.")
```
Naatan, line 10:
```suggestion
		// Errors that are due to USER input should use `NewInputError` or `WrapInputError`
		return locale.NewInputError("hello_err_no_name", "No name provided.")
```

internal/runners/hello/hello_example.go
```diff
@@ -0,0 +1,136 @@
...
+func (h *Hello) Run(params *RunParams) error {
+	// Reusable runner logic is contained within the runbits package.
+	// You should only use this if you intend to share logic between
+	// runners. Runners should NEVER invoke other runners.
+	if err := runbits.SayHello(h.out, params.Name); err != nil {
+		// Errors should nearly always be localized.
+		return locale.WrapError(
+			err, "hello_cannot_say", "Cannot say hello.",
+		)
+	}
```
Naatan, line 71: Can we move this after the input validation? Since this is meant to demonstrate good standards.

internal/runners/hello/hello_example.go
```diff
@@ -0,0 +1,136 @@
...
+		err := locale.NewInputError(
+			"hello_info_err_no_project", "Not in a project directory.",
+		)
+
+		// It's useful to offer users reasonable tips on recourses.
+		return errs.AddTips(err, locale.Tl(
+			"hello_suggest_checkout",
+			"Try using [ACTIONABLE]`state checkout`[/RESET] first.",
+		))
+	}
```
Naatan, line 91: This should happen before any sort of output imo.

internal/runners/hello/simplenamespace_example.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Type implements the flagmarshaler interface.
+func (ns *SimpleNamespace) Type() string {
+	return "simple-namespace"
+}
+
+func (ns *SimpleNamespace) IsSet() bool {
+	return ns.isSet
+}
```
Naatan, line 61: This is overkill for demonstrating the mechanic. Please boil it way down to something super basic. We're just demoing the mechanic of custom flag types, we're not demoing an actual implementation of a flag. The whole thing should ideally be no more than like 30 lines of code, preferably even less.

Namespace was only one example; it can be anything so long as it demonstrates the mechanic.

Naatan (changes requested):

internal/runners/hello/hello_example.go
daved: The project was only necessary when the extra flag was set, but I agree with your sentiment and moved things around to do validation first (regardless of flags).

internal/runners/hello/hello_example.go
```diff
@@ -0,0 +1,139 @@
...
+	return &RunParams{
+		Name: "Friend",
+	}
+}
+
+// Hello defines the app-level dependencies that are accessible within the Run
+// function.
+type Hello struct {
+	out output.Outputer
+	pj  *project.Project
```
Naatan, line 49: Please rename to `project`; use full words. Sorry since this is an example of best practices I have to be extra pedantic.. :p

Naatan (changes requested):

Naatan (approved):

PR #2625, opened 2023-07-11
Do not require installation path validation when installing for the first time.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1986" title="DX-1986" target="_blank">BugDX-1986</a>  ST installation/upgrading is failing 
  </td></table>
  <br />

### DIFF
--- a/internal/installation/paths.go
+++ b/internal/installation/paths.go
@@ -22,12 +22,11 @@ const (
 )
 
 func DefaultInstallPath() (string, error) {
-	return InstallPathForBranch(constants.BranchName, true)
+	return InstallPathForBranch(constants.BranchName)
 }
 
 // InstallPathForBranch gets the installation path for the given branch.
-// validate should always be `true` unless the installer wants a path to install to.
-func InstallPathForBranch(branch string, validate bool) (string, error) {
+func InstallPathForBranch(branch string) (string, error) {
 	if v := os.Getenv(constants.InstallPathOverrideEnvVarName); v != "" {
 		return filepath.Clean(v), nil
 	}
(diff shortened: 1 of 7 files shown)


internal/installation/paths.go
```diff
@@ -22,14 +22,25 @@ const (
...
+
+	installPath, err := installPathForBranch(branch)
+	if err != nil {
+		return "", errs.Wrap(err, "Unable to determine install path for branch")
+	}
+	if validate && !isValidInstallPath(installPath) {
+		return "", errs.New("Invalid install path: %s", installPath)
+	}
+
+	return installPath, nil
```
Naatan, line 43: I don't think this should ever have `validate=true`. I think the bug that was introduced is that we are validating on methods that aren't retrieving the installation path of the installed tool, but rather are retrieving the intended installation path for new installs. I'm partly to blame for that for not anticipating this in the original stories AC's. I think the fix to this bug is to simply remove the validation from `InstallPathForBranch`. From some quick searches I can't find any scenario's where we use this function to mean "give me the installed state tool path" (nor should we).

Naatan (changes requested):

internal/installation/paths.go
mitchell-as: @Naatan  What about update? https://github.com/ActiveState/cli/pull/2625/files#diff-b444fc649cba95f05cd878b61c154267c36c3dc2330cc5555b9fbb4dad0826b7R72 Surely that needs a valid installation path. Should it be using a different function?

Do I need to change https://github.com/ActiveState/cli/pull/2625/files#diff-ef481bb41441e200b51ddc73ec680b82a0e4f174d53b7292587ac9463d383cfeR25 to not call `InstallPathForBranch()`? If I simply remove the validate param, then the original "fix" (Mike's, not mine) will be completely negated and this ticket will be back at square one.

Naatan: The use in update is kind of miss-leading. It is not updating anything in that codepath, instead it is installing a new state tool version under the given branch path which is not meant to necessarily be an existing install. I've wanted to remove it because it is effectively redundant with `install.sh -b` and not technically an "update", but Pete likes it.

I was suggesting you remove validation from `InstallPathForBranch()`, ie. both the param you added and the actual validation logic invoked.

mitchell-as: Okay, I've removed the param. Hopefully I understood what you were requesting. Let me know otherwise.

PR #2626, opened 2023-07-11
Update install archive

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1958" title="DX-1958" target="_blank">StoryDX-1958</a>  As a devops engineer I can install the state tool simply by extracting it from an archive
  </td></table>
  <br />


Naatan (changes requested):
Now we're adding it twice, once during unpacking and once during installation. Not clear if the unpacking one is actually working as a result, though. The tests are still covered by the original install of this file.

Naatan (approved):

PR #2629, opened 2023-07-14
Fix builds in progress

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1994" title="DX-1994" target="_blank">BugDX-1994</a>  Buildplanner implementation errors when build is in progress
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -531,16 +531,20 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
 		recipeID = buildResult.RecipeID
 	}
 
+	if !buildResult.BuildReady {
+		err = buildplan.AddBuildArtifacts(artifacts, buildResult.Build)
+		if err != nil {
+			return nil, nil, errs.Wrap(err, "Could not add build artifacts to build plan")
+		}
+	}
+
 	if err := s.eventHandler.Handle(events.Start{
 		RecipeID:      recipeID,
 		RequiresBuild: !buildResult.BuildReady,
 		ArtifactNames: artifactNames,
 		LogFilePath:   logFilePath,
 		ArtifactsToBuild: func() []artifact.ArtifactID {
-			if !buildResult.BuildReady {
-				return artifact.ArtifactIDsFromBuildPlanMap(artifacts) // This does not account for cached builds
-			}
-			return []artifact.ArtifactID{}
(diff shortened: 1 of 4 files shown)


pkg/platform/runtime/setup/setup.go
```diff
@@ -397,6 +397,13 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
 		}
 	}
 
+	if !buildResult.BuildReady {
+		err = buildplan.AddBuildArtifacts(artifacts, buildResult.Build)
+		if err != nil {
+			return nil, nil, errs.Wrap(err, "Could not add build artifacts to build plan")
+		}
```
Naatan, line 404: No we're adding it individually under the installFromBuildLog and installFromBuildResult methods. It seems like we should be able to do this before either of those as they both seem to take the same artifacts as input.

Point being there should probably only be one call to this function. And shouldn't the result of this be fed into https://github.com/ActiveState/cli/blob/78a7d9f5495dc6ff466a0ada71d780a175df8d3d/pkg/platform/runtime/setup/events/events.go#L31-L31 ?

Naatan (changes requested):

pkg/platform/runtime/setup/setup.go
MDrakos: Yes, you're right we only need to add the build artifacts once. By adding it earlier in the setup process we feed the results into the `ArtifactsToBuild` as well as `ArtifactsToDownload` and others that are populated by the start event.

Naatan: We're still calling it twice after this PR would land in its current state though, the second one being here:

https://github.com/ActiveState/cli/blob/78a7d9f5495dc6ff466a0ada71d780a175df8d3d/pkg/platform/runtime/setup/setup.go#L687-L687

Naatan (changes requested):

pkg/platform/runtime/setup/setup.go
MDrakos: That should be removed here: https://github.com/ActiveState/cli/pull/2629/files#diff-c415bd62e9cd64f30d585a034cd17ab1f283c67287d63a1448787a2a81efc043L687

Naatan: Ahh I see it now. My bad.

Naatan (approved):

PR #2630, opened 2023-07-14
Add the private project authentication disclaimer to revert commit errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1969" title="DX-1969" target="_blank">BugDX-1969</a>  REVERT: Cannot revert commit that is more than 1 commit behind the current
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -224,7 +224,7 @@ err_auth_username_check:
 err_auth_failed_unknown_cause:
   other: "Authentication failed due to an unknown cause, error received: {{.V0}}"
 err_auth_token:
-  other: "Failed to create authentication token, please try again or run [ACTIONABLE]`state clean config`[/RESET] if the issue persists."
+  other: "Failed to create authentication token, please try again or run '[ACTIONABLE]state clean config[/RESET]' if the issue persists."
 auth_err_password_prompt:
   other: The provided password is invalid
 confirm_password_account_creation:
@@ -921,7 +921,7 @@ err_deprecation:
 err_auth_required:
   other: Authentication is required. Please authenticate by running [ACTIONABLE]`state auth`[/RESET].
 err_auth_needinput:
-  other: You are logged out. In order to log in, either run this command in interactive mode, or provide your credentials via flags (see [ACTIONABLE]`state auth --help`[/RESET]).
+  other: You are logged out. In order to log in, either run this command in interactive mode, or provide your credentials via flags (see '[ACTIONABLE]state auth --help[/RESET]').
 auth_required_activate:
   other: You need to be authenticated to activate an ActiveState Platform project
 prompt_login_or_signup:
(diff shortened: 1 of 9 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1748,6 +1748,11 @@ err_user_network_solution:
     If your issue persists consider reporting it on our forums at {{.V0}}.
 err_revert_refresh:
   other: Could not get revert commit to check if changes were indeed made
+err_revert_commit:
+  other: |
+    Could not revert{{.V0}} commit: {{.V1}}
+    Please ensure that the local project is synchronized with the platform and that the given commit ID belongs to the current project.
+    If this is a private project, you may need to authenticate with [ACTIONABLE]`state auth`[/RESET]
```
daved, line 1755: Consider using AddTips. In my opinion, error output is simply a list of what was tried and/or what went wrong, and AddTips is for suggestions. This makes me wonder if we have more error messages like this.

daved (changes requested):

internal/locale/locales/en-us.yaml
mitchell-as: We do, because I initially used tips, but dropped them in favor of existing messages.

I changed those messages to use tips too.

internal/locale/locales/en-us.yaml
```diff
@@ -917,9 +919,9 @@ err_deprecation:
 
     Please update by running [ACTIONABLE]`state update`[/RESET].
 err_auth_required:
-  other: Authentication is required, please authenticate by running 'state auth'
+  other: Authentication is required. Please authenticate by running [ACTIONABLE]`state auth`[/RESET].
 err_auth_needinput:
-  other: You are logged out, in order to log in either run this command in interactive mode or provide your credentials via flags (see '[ACTIONABLE]state auth --help[/RESET]').
+  other: You are logged out. In order to log in, either run this command in interactive mode, or provide your credentials via flags (see '[ACTIONABLE]state auth --help[/RESET]').
```
daved, line 924: Please swap the single quotes to back tics and move them inside the actionable/reset tags.

daved (changes requested):
Just one little thing.

internal/locale/locales/en-us.yaml
mitchell-as: Okay. I changed other instances I found.

daved (approved):
I'm not sure @Naatan would be content with affecting so many other localizations not having to do with this story, but I'm OK with this.

Naatan (changes requested):
Thanks for tagging me. Please descope the locale changes from this PR. This looks like a useful change but I'm concerned with the use of tidle's rather than single quotes as those get evaluated by bash scripts and I'm pretty sure this is a trap we fell into and why we resorted to single quotes.

But it seems unrelated to this story so better to just drop it from this PR rather than iterate on changes not intended for this PR.

Naatan (approved):

PR #2631, opened 2023-07-14
BuildPlanner implementation can modify platforms

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1975" title="DX-1975" target="_blank">StoryDX-1975</a>  The BuildPlanner implementation can modify platforms
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildexpression/buildexpression.go
+++ b/pkg/platform/runtime/buildexpression/buildexpression.go
@@ -540,7 +540,7 @@ func (e *BuildExpression) getPlatformsNode() *[]*Value {
 }
 
 // Update updates the BuildExpression's requirements based on the operation and requirement.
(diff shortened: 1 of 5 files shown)


pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -659,6 +679,59 @@ func (e *BuildExpression) updateRequirement(requirement model.Requirement) error
...
+		return errs.New("Unsupported operation")
+	}
+	if err != nil {
+		return errs.Wrap(err, "Could not update BuildExpression's platform")
+	}
+
+	err = e.updateTimestamp(timestamp)
+	if err != nil {
+		return errs.Wrap(err, "Could not update BuildExpression's timestamp")
+	}
```
Naatan, line 699: This doesn't feel like it should be part of UpdatePlatform. It should have its own call.

Naatan (changes requested):

Naatan (approved):

PR #2637, opened 2023-07-18
Treat BuildPlanner errors are input errors

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2000" title="DX-2000" target="_blank">TaskDX-2000</a>  We treat buildplan errors as input errors
  </td></table>
  <br />
 

This PR adds back the `Wrapped` field to the `BuildPlanError` type. When we report errors we unwrap the error and check to see if any of the underlying errors are input errors, if they are we do not report.

This is only added when we encounter a planning error as other errors before this occur when a commit or project could not be found.

### DIFF
--- a/pkg/platform/model/buildplanner.go
+++ b/pkg/platform/model/buildplanner.go
@@ -305,7 +305,6 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
 			}
 		}
 		return "", &bpModel.BuildPlannerError{
-			Wrapped:          locale.NewInputError("err_buildplanner"),
 			ValidationErrors: errs,
 			IsTransient:      isTransient,
 		}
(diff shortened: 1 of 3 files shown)


Naatan (changes requested):
@MDrakos does this mean that any buildplan error will always just say "The Platform failed to plan the build for this project."?

The intend of the story is to not have it report to rollbar. Not to change the error message. We should still be able to give an error message that is actionable.

MDrakos:
@Naatan good point. I think this latest change addresses your concerns.

pkg/platform/model/buildplanner.go
```diff
@@ -305,6 +305,7 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
 			}
 		}
 		return "", &bpModel.BuildPlannerError{
+			Wrapped:          locale.NewInputError("err_buildplanner"),
```
Naatan, line 308: Looks like you can drop this?

Naatan (changes requested):

Naatan (approved):

PR #2641, opened 2023-07-19
Add debug logging to graphQL responses

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2024" title="DX-2024" target="_blank">StoryDX-2024</a>  Add extra debug logging to build planner calls
  </td></table>
  <br />

### DIFF
--- a/internal/gqlclient/gqlclient.go
+++ b/internal/gqlclient/gqlclient.go
@@ -109,11 +109,13 @@ func (c *Client) RunWithContext(ctx context.Context, request Request, response i
 	if err != nil {
 		return NewRequestError(err, request)
 	}
-	responseData, err := json.MarshalIndent(response, "", "  ")
-	if err != nil {
-		return errs.Wrap(err, "failed to marshal response")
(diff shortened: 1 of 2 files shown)


internal/gqlclient/gqlclient.go
```diff
@@ -107,6 +109,11 @@ func (c *Client) RunWithContext(ctx context.Context, request Request, response i
 	if err != nil {
 		return NewRequestError(err, request)
 	}
+	responseData, err := json.MarshalIndent(response, "", "  ")
+	if err != nil {
+		return errs.Wrap(err, "failed to marshal response")
+	}
+	logging.Debug("gqlclient: response: %s", responseData)
```
Naatan, line 116: Would be good if this used the `ACTIVESTATE_DEBUG_SERVICE_REQUESTS` and only logged it when set. Otherwise, this'll get a bit verbose and we will end up missing valuable data in rollbar logs as they only support a limited log size.

Naatan (changes requested):

Naatan (approved):

PR #2644, opened 2023-07-20
Ensure package updates are performed

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2017" title="DX-2017" target="_blank">BugDX-2017</a>  No `local` commit created and no package list updated after successful package update
  </td></table>
  <br />
 

The issue was fixed in https://activestatef.atlassian.net/browse/DX-2020

This is just adding a test covering the slightly different repro.


MDrakos (dismissed):
LGTM but there is a unit test failure that will need to be addressed.

Naatan:
> LGTM but there is a unit test failure that will need to be addressed.

Well that seems reasonable for a PR which in no way shape or form touches unit tests... 🤨

PR #2646, opened 2023-07-20
Fix builds in progress failing on file transformations

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2016" title="DX-2016" target="_blank">BugDX-2016</a>  `Could not apply necessary file transformations after unpacking.` during package install
  </td></table>
  <br />
 

Fixing one issue brought up another, so there are several things happening here:

- We're no longer tracking build progress of build-time artifacts
  - Followup filed here: https://activestatef.atlassian.net/browse/PB-4432
- I had to revert DX-1996 as artifacts that haven't built yet are reporting an empty URL
- I had to add workarounds for the buildlogstreamer reporting on artifacts we don't care about, notably:
  - Build progress reporting will now ignore and debug log unexpected artifact build events, rather than error out.
  - The buildlogstreamer will now send a build success event once all artifacts have been satisfied, rather than when IT sends the build success events, because IT only sends it once all artifacts are done, which includes ones we don't care about.


pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -45,6 +47,17 @@ func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
...
+	// Eliminate noop artifacts
+	for _, prebuiltArtf := range build.Artifacts {
+		if prebuiltArtf.NodeID != "" && prebuiltArtf.Status != "" &&
+			prebuiltArtf.Status == model.ArtifactSucceeded &&
+			strings.HasPrefix(prebuiltArtf.URL, "s3://as-builds/noop/") {
+			if _, exists := res[prebuiltArtf.NodeID]; exists {
+				delete(res, prebuiltArtf.NodeID)
+			}
+		}
+	}
```
Naatan, line 59: Previously we were only doing this for download/install artifacts. But it made sense to extend this to all. There's no point in looking at these at all.

MDrakos: Artifact URLs from the buildplanner should not have an `s3` scheme, they should all be `https` so I don't think this will have an effect.

pkg/platform/runtime/setup/setup.go
```diff
@@ -382,46 +382,25 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
 	}
 
 	// Compute and handle the change summary
-	var artifacts artifact.Map
+	// runtimeAndBuildtimeArtifacts records all artifacts that will need to be build in order to obtain the runtime.
```
MDrakos, line 385:
```suggestion
	// runtimeAndBuildtimeArtifacts records all artifacts that will need to be built in order to obtain the runtime.
```

pkg/platform/runtime/setup/setup.go
```diff
@@ -382,46 +382,25 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
...
+	// Disabled due to PB-4432.
```
MDrakos, line 386: Do we want to create a follow-up story in DevX now?

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -298,7 +309,12 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
 // adds the artifact's dependencies to the map. This is useful for when we are
 // using the BuildLogStreamer as it operates on the older recipeID and will include
 // more artifacts than what we originally calculated in the runtime closure.
-func AddBuildArtifacts(artifactMap artifact.Map, build *model.Build) error {
+func AddBuildArtifacts(artifactMap artifact.Map, build *model.Build) (artifact.Map, error) {
```
MDrakos, line 312: You've removed the only call to this function in `setup.go`. Do we even need this anymore?

MDrakos (changes requested):

MDrakos (dismissed):

PR #2647, opened 2023-07-20
Add more context to requirement removal error message

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2021" title="DX-2021" target="_blank">BugDX-2021</a>  General error provide instead of related to the failure when trying to uninstall non existent package
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildexpression/buildexpression.go
+++ b/pkg/platform/runtime/buildexpression/buildexpression.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
 	"github.com/ActiveState/cli/internal/sliceutils"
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
This fix is too specific to the error we received. The error is indicative of a larger issue, I'd like to see the PR for this story address that larger issue.

The larger issue being that we apparently do not localize API errors that we were previously localizing. I doubt we manually constructed the error you saw before.

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -614,7 +615,7 @@ func (e *BuildExpression) removeRequirement(requirement model.Requirement) error
 	}
 
 	if !found {
-		return errs.New("Could not find requirement")
+		return locale.NewInputError("err_remove_requirement_not_found", "Could not remove requirement with namespace: {{.V0}}, name: {{.V1}}", requirement.Namespace, requirement.Name)
```
Naatan, line 618:
```suggestion
		return locale.NewInputError("err_remove_requirement_not_found", "Could not remove requirement '[ACTIONABLE]{{.V0}}[/RESET]', because it does not exist.", requirement.Name)
```

Naatan (changes requested):

Naatan (approved):

PR #2651, opened 2023-07-21
Fix install integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2034" title="DX-2034" target="_blank">BugDX-2034</a>  Nightly failure: TestInstallFromLocalSource
  </td></table>
  <br />


Naatan:
@MDrakos just to hopefully save you some time - you seem to be introducing new code to fix what is a regression. We should find what broke this and undo that change. This shouldn't require new logic.

PR #2653, opened 2023-07-22
Save existing refresh token during dcag instead of creating new one

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1873" title="DX-1873" target="_blank">BugDX-1873</a>  state auth creates two apikeys
  </td></table>
  <br />

### DIFF
--- a/pkg/cmdlets/auth/login.go
+++ b/pkg/cmdlets/auth/login.go
@@ -247,10 +247,10 @@ func AuthenticateWithBrowser(out output.Outputer, auth *authentication.Auth, pro
 		out.Notice(locale.Tr("err_browser_open", *response.VerificationURIComplete))
 	}
 
-	var refreshToken string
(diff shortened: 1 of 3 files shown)


pkg/cmdlets/auth/login.go
```diff
@@ -240,9 +247,11 @@ func AuthenticateWithBrowser(out output.Outputer, auth *authentication.Auth, pro
 		out.Notice(locale.Tr("err_browser_open", *response.VerificationURIComplete))
 	}
 
+	var refreshToken string
```
Naatan, line 250: Please use our local terminology, I believe this is an API key? (update all references)

Naatan (changes requested):
Looks good! Just one minor change.

Naatan (approved):

PR #2654, opened 2023-07-22
Lock down hello cmd to our ci/testing

daved:


Naatan (changes requested):
Now we can't run or test it locally.

Naatan (changes requested):
Let's just make the command hidden (`captain.SetHidden()`) and leave it at that. It's not really a big deal that we're shipping this. I mean even with the current code we are shipping it, it's just obfuscated.

I propose:

- Mark it hidden
- Mark it unstable
- Rename the command itself to `_hello` (avoid any collision with future commands)
- Have it print out "This command is for example use only".

Naatan (approved):

PR #2656, opened 2023-07-25
Update `state publish` to new namespace format

Naatan:


pkg/platform/model/vcs.go
```diff
@@ -82,10 +82,8 @@ const (
 	// NamespaceCamelFlagsMatch is the namespace used for passing camel flags
 	NamespaceCamelFlagsMatch = `^camel-flags$`
 
-	// NamespaceSharedMatch is the namespace used for shared requirements (usually runtime libraries)
-	NamespaceSharedMatch = `^shared$`
-
-	NamespaceOrgSharedMatch = `^\w+\/shared(?$|\/\w+$)`
+	// NamespaceOrgMatch is the namespace used for org specific requirements
+	NamespaceOrgMatch = `^org\/(\w+)$`
```
MDrakos, line 86: Can org names have dashes or any other punctuation?

MDrakos (approved):
LGTM just maybe verify the org question.

pkg/platform/model/vcs.go
Naatan: I don't know, but we probably shouldn't care either way. I'll drop the suffix matching.

MDrakos (approved):

PR #2657, opened 2023-07-25
Ensure RemediableSolverErrors are presented to users

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2037" title="DX-2037" target="_blank">BugDX-2037</a>  Package install attempt with NON-existent version did not generate proper errors
  </td></table>
  <br />
 

Adds errors in the format below
image

### DIFF
--- a/pkg/platform/api/buildplanner/model/buildplan.go
+++ b/pkg/platform/api/buildplanner/model/buildplan.go
@@ -61,6 +61,10 @@ const (
 	XActiveStateArtifactMimeType = "application/x-activestate-artifacts"
 	XCamelInstallerMimeType      = "application/x-camel-installer"
 	XGozipInstallerMimeType      = "application/x-gozip-installer"
+
+	// Error types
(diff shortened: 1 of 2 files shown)


pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -61,6 +61,10 @@ const (
 	XActiveStateArtifactMimeType = "application/x-activestate-artifacts"
 	XCamelInstallerMimeType      = "application/x-camel-installer"
 	XGozipInstallerMimeType      = "application/x-gozip-installer"
+
+	// Error types
+	RemediableSolveErrorType = "RemediableSolveError"
+	GenericSolveErrorType    = "GenericSolveError"
```
Naatan, line 67: Doesn't look like `GenericSolveErrorType` is used?

Naatan (changes requested):

Naatan (approved):

PR #2659, opened 2023-07-26
`state init` and `state install` supports wildcard notation for language and package versions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2011" title="DX-2011" target="_blank">StoryDX-2011</a>  We can install requirements with wildcard version matching
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -225,7 +225,7 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 		return errs.Wrap(err, "Could not resolve requirement name and version")
 	}
 
-	requirements, err := versionStringToRequirements(version)
+	requirements, err := model.VersionStringToRequirements(version)
 	if err != nil {
 		return errs.Wrap(err, "Could not process version string into requirements")
 	}
@@ -410,20 +410,4 @@ func initializeProject() (*project.Project, error) {
 	}
 
 	return project.FromPath(target)
-}
-
-func versionStringToRequirements(version string) ([]bpModel.VersionRequirement, error) {
-	constraints, err := model.VersionStringToConstraints(version)
-	if err != nil {
-		return nil, errs.Wrap(err, "Unable to process version string into constraints")
-	}
-
-	requirements := make([]bpModel.VersionRequirement, len(constraints))
-	for i, constraint := range constraints {
-		requirements[i] = bpModel.VersionRequirement{
-			bpModel.VersionRequirementComparatorKey: constraint.Comparator,
-			bpModel.VersionRequirementVersionKey:    constraint.Version,
-		}
-	}
-	return requirements, nil
 }
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -225,12 +225,17 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 		return errs.Wrap(err, "Could not resolve requirement name and version")
 	}
 
+	requirements, err := versionStringToRequirements(version)
+	if err != nil {
+		return errs.Wrap(err, "Could not process version string into requirements")
+	}
+
 	params := model.StageCommitParams{
 		Owner:                pj.Owner(),
 		Project:              pj.Name(),
 		ParentCommit:         string(parentCommitID),
 		RequirementName:      name,
-		RequirementVersion:   version,
+		RequirementVersion:   requirements,
 		RequirementNamespace: ns,
 		Operation:            operation,
 		TimeStamp:            *latest,
@@ -405,4 +410,20 @@ func initializeProject() (*project.Project, error) {
 	}
 
 	return project.FromPath(target)
+}
+
+func versionStringToRequirements(version string) ([]bpModel.VersionRequirement, error) {
+	constraints, err := model.VersionStringToConstraints(version)
+	if err != nil {
+		return nil, errs.Wrap(err, "Unable to process version string into constraints")
+	}
+
+	requirements := make([]bpModel.VersionRequirement, len(constraints))
+	for i, constraint := range constraints {
+		requirements[i] = bpModel.VersionRequirement{
+			bpModel.VersionRequirementComparatorKey: constraint.Comparator,
+			bpModel.VersionRequirementVersionKey:    constraint.Version,
+		}
+	}
+	return requirements, nil
 }
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -139,7 +139,7 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 		}
 	}
 
-	version, err := deriveVersion(lang, languageVersion)
+	version, err := deriveVersion(getKnownVersionsFromPlatform, lang, languageVersion)
 	if err != nil {
 		if inferred || !locale.IsInputError(err) {
 			return locale.WrapError(err, "err_init_lang", "", languageName, languageVersion)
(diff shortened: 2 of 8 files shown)


internal/runners/initialize/init.go
```diff
@@ -258,6 +258,8 @@ func verifyLangAndVersion(lang language.Language, version string) error {
 		return locale.NewInputError("err_init_language_not_found", "The selected language cannot be found")
 	}
 
+	version = strings.Replace(strings.Replace(version, ".x", "", 1), ".X", "", 1) // strip any wildcard
```
mitchell-as, line 261: We're just verifying the wildcarded version exists in the first place. e.g. 3.10.x will check that 3.10 is a valid version prefix.

pkg/platform/model/vcs.go
```diff
@@ -526,29 +512,6 @@ func DeleteBranch(branchID strfmt.UUID) error {
 	return nil
 }
 
-// CommitPackage commits a package to an existing parent commit
```
mitchell-as, line 529: This function is unused.

pkg/platform/model/vcs.go
```diff
@@ -691,23 +676,6 @@ func (cs indexedCommits) countBetween(first, last string) (int, error) {
 	return ct, nil
 }
 
-// CommitRequirement commits a single requirement to the platform
```
mitchell-as, line 694: This function is unused.

pkg/platform/model/vcs.go
```diff
@@ -643,6 +600,34 @@ func CommitInitial(hostPlatform string, langName, langVersion string) (strfmt.UU
 	return res.Payload.CommitID, nil
 }
 
+func versionStringToConstraints(version string) ([]*mono_models.Constraint, error) {
+	constraints := []*mono_models.Constraint{}
```
mitchell-as, line 604: Note that exact versions (no wildcards) will be supported in a separate PR that addresses DX-2010.

internal/runners/initialize/init.go
```diff
@@ -239,60 +238,83 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
...
 	if err != nil {
-		multilog.Error("Failed to fetch supported languages (using hardcoded default version): %s", errs.JoinMessage(err))
-		return lang.RecommendedVersion()
+		return "", errs.Wrap(err, "Unable to get known versions for language %s", lang.Requirement())
 	}
 
-	for _, l := range langs {
-		if lang.String() == l.Name || (lang == language.Python3 && l.Name == language.Python3.Requirement()) {
-			return l.DefaultVersion
+	prefix := strings.Replace(strings.Replace(version, ".x", "", 1), ".X", "", 1) // strip any wildcard
```
mitchell-as, line 291: We're just verifying the wildcarded version exists in the first place. e.g. 3.10.x will check that 3.10 is a valid version prefix.

daved (changes requested):

daved (approved):

internal/runbits/requirements/requirements.go
```diff
@@ -406,3 +411,19 @@ func initializeProject() (*project.Project, error) {
 
 	return project.FromPath(target)
 }
+
+func versionStringToRequirements(version string) ([]bpModel.VersionRequirement, error) {
+	constraints, err := model.VersionStringToConstraints(version)
+	if err != nil {
+		return nil, errs.Wrap(err, "Unable to process version string into constraints")
+	}
```
Naatan, line 419: Can we make this only act if we find that the value isn't just decimal numbers? Most of the time we shouldn't need to reach out to this API.

internal/runbits/requirements/requirements.go
```diff
@@ -406,3 +411,19 @@ func initializeProject() (*project.Project, error) {
...
+
+	requirements := make([]bpModel.VersionRequirement, len(constraints))
+	for i, constraint := range constraints {
+		requirements[i] = bpModel.VersionRequirement{
+			bpModel.VersionRequirementComparatorKey: constraint.Comparator,
+			bpModel.VersionRequirementVersionKey:    constraint.Version,
+		}
+	}
+	return requirements, nil
```
Naatan, line 428: I don't like that we're modeling this data outside of the model package. It feels like this should be owned by the model. We should use and enrich the params that model.StageCommit uses.

pkg/platform/model/vcs.go
```diff
@@ -643,6 +600,38 @@ func CommitInitial(hostPlatform string, langName, langVersion string) (strfmt.UU
 	return res.Payload.CommitID, nil
 }
 
+func isWildcardVersion(version string) bool {
+	return strings.Index(version, "x") >= 0 || strings.Index(version, "X") >= 0
```
Naatan, line 604: For `init` you require `x` to be after a dot. Here you don't. I think requiring the dot probably makes more sense, suggests we keep it consistent.

pkg/platform/model/vcs.go
```diff
@@ -643,6 +600,38 @@ func CommitInitial(hostPlatform string, langName, langVersion string) (strfmt.UU
...
+			return nil, locale.NewInputError("err_version_wildcard_start", "A version number cannot start with a wildcard")
+		}
+		constraints = append(constraints, &mono_models.Constraint{Comparator: "gte", Version: strings.Join(parts[:i], ".")})
+		previousPart, err := strconv.Atoi(parts[i-1])
+		if err != nil {
+			return nil, locale.WrapInputError(err, "err_version_number_expected", "Version parts are expected to be numeric")
+		}
+		parts[i-1] = strconv.Itoa(previousPart + 1)
+		constraints = append(constraints, &mono_models.Constraint{Comparator: "lt", Version: strings.Join(parts[:i], ".")})
+	}
```
Naatan, line 631: This does not feel like it should be using mono model structs since we're not sending this to the mono API anymore. Probably makes more sense in the buildplan model?

Naatan (changes requested):

internal/runbits/requirements/requirements.go
mitchell-as: This is done in a separate PR: https://github.com/ActiveState/cli/pull/2663/files#diff-d9612958a129acac4c7a4c03e241b6b912c286635b188a24975afe794fcc94f5R605-R618

pkg/platform/model/vcs.go
mitchell-as: We are using the mono API for `state init` (vcs.go's `CommitInitial()` function). In order to avoid duplication, I've decided to have the string -> constraints list translation happen here in vcs.go. BuildPlanner makes use of this, but translates from mono API constraints to build planner constraints.

Let me know if you'd like me to reverse this. The `state init` work came first, so it has precedence in my mind.

pkg/platform/model/vcs.go
```diff
@@ -600,34 +601,18 @@ func CommitInitial(hostPlatform string, langName, langVersion string) (strfmt.UU
...
 }
 
-func isWildcardVersion(version string) bool {
-	return strings.Index(version, "x") >= 0 || strings.Index(version, "X") >= 0
-}
-
-func VersionStringToConstraints(version string) ([]*mono_models.Constraint, error) {
-	if !isWildcardVersion(version) {
-		return []*mono_models.Constraint{{Comparator: "eq", Version: version}}, nil // exact version
+func versionStringToConstraints(version string) ([]*mono_models.Constraint, error) {
```
Naatan, line 604: Not asking for a change, but I just realized I was wrong in the first review in that we don't use the mono model. Filed a followup here: https://activestatef.atlassian.net/browse/DX-2064

Naatan (approved):

PR #2660, opened 2023-07-26
Allow installing overtop previous state tool as long as the branch matches

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2047" title="DX-2047" target="_blank">StoryDX-2047</a>  If you run the installer and have a pre-existing installation it should not require --force if you are installing from the same branch.
  </td></table>
  <br />


mitchell-as (approved):
Looks good to me. Nicely done.

PR #2661, opened 2023-07-26
Increase logging to help us diagnose API token issues

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2039" title="DX-2039" target="_blank">BugDX-2039</a>  Export failed just after auth (Windows11)
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/authentication/auth.go
+++ b/pkg/platform/authentication/auth.go
@@ -117,7 +117,7 @@ func (s *Auth) Sync() error {
 	defer profile.Measure("auth:Sync", time.Now())
 
 	if token := s.AvailableAPIToken(); token != "" {
-		logging.Debug("Authenticating with stored API token: %s..", token[0:2])
+		logging.Debug("Authenticating with stored API token: %s..", desensitizeToken(token))
 		if err := s.Authenticate(); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/platform/authentication/auth.go
```diff
@@ -116,8 +116,8 @@ func (s *Auth) SyncRequired() bool {
 func (s *Auth) Sync() error {
 	defer profile.Measure("auth:Sync", time.Now())
 
-	if s.AvailableAPIToken() != "" {
-		logging.Debug("Authenticating with stored API token")
+	if token := s.AvailableAPIToken(); token != "" {
+		logging.Debug("Authenticating with stored API token: %s..", token[0:2])
```
Naatan, line 120: Obviously don't want to leak sensitive info, hence why I'm just logging the first 2 characters which seems harmless enough.

mitchell-as: Maybe create a function `desensitizeToken()` or something that returns `token[0:2]`? That way you don't have to repeat the magic numbers in more than one place. If we need a different method, then it only needs to be changed in one place.

mitchell-as (changes requested):

mitchell-as (approved):

PR #2663, opened 2023-07-27
Support constraint strings for `state init` and `state install`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2010" title="DX-2010" target="_blank">StoryDX-2010</a>  As a user I can explicitly request either full or partial language versions
  </td></table>
  <br />
 

These are strings like ">=3.10,<3.11" that the Dashboard shows.


pkg/platform/api/reqsimport/reqsimport.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/pkg/platform/api"
-	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/api/mono/mono_models"
```
mitchell-as, line 14: Changing this to avoid an import cycle when importing this file from the model package.

pkg/platform/model/vcs.go
```diff
@@ -600,13 +602,37 @@ func CommitInitial(hostPlatform string, langName, langVersion string) (strfmt.UU
 	return res.Payload.CommitID, nil
 }
 
+var versionRe = regexp.MustCompile(`^\d+(\.\d+)+$`)
```
daved, line 605: Consider making the repeat non-capturing, and limited in count: `^\d+(?:\.\d+){0,2}$`
Also, how will other variations play into this? Like `3.4.5rc2` or `1.2.3x4`.
Maybe something more specific to the wildcards:
```
func isExactVersion(v string) bool {
	return !strings.HasSuffix(v, ".x") && !strings.Contains(v, ".x.")
}
```

pkg/platform/model/vcs.go
```diff
@@ -600,13 +602,37 @@ func CommitInitial(hostPlatform string, langName, langVersion string) (strfmt.UU
...
+		dummyLanguage := language.Python3.Requirement()
+		changeset, err := reqsimport.Init().Changeset([]byte("name "+version), dummyLanguage)
+		if err != nil {
+			return nil, errs.Wrap(err, "Unable to translate version string into requirement list")
+		}
+		constraints := []*mono_models.Constraint{}
+		for _, change := range changeset {
+			constraints = append(constraints, change.VersionConstraints...)
+		}
+		return constraints, nil
```
daved, line 635: This sections makes me concerned for cases like `1.2.3rc2` or some oddball like `3.2.1x4`.
```
switch {
case isExactVersion(version):
	// ...
case isWildcardVersion(version):
	// ...
case hasVersionComparator(version):
	// ...
default:
	// the user likely provided a mix of wildcard and version comparator forms - input error
}
```
```
func isExactVersion(v string) bool {
	return !isWildcardVersion(v) && !hasVersionComparator(v)
}

func hasVersionComparator(v string) bool {
	return strings.ContainsAny(v, ">=<,")
}

func isWildcardVersion(v string) bool {
	return strings.HasSuffix(v, ".x") || strings.Contains(v, ".x.")
}
```

daved (changes requested):

pkg/platform/model/vcs.go
mitchell-as: I have never seen a platform version that is not of the form x.y.z. Also, the ACs said to not perform any sort of processing/interpretation of the version string given. I only have the `isExactVersion()` there to avoid a platform call for performance reasons. I am perfectly content having the platform do all the work in interpreting a version, and erroring out if it is invalid.

daved: I think I see what you're saying. My thought is that isExactVersion could be a bit more accurate to its name. @Naatan, is it inappropriate to consider values like 1.2.3rc4 in this case?

Naatan: I'm not sure. @swagile1 what's the contract on ingredient versions?

For that matter; could we have versions that have `.x` in them?

pkg/platform/model/vcs_test.go
```diff
@@ -152,7 +152,7 @@ func (suite *VCSTestSuite) TestVersionStringToConstraints() {
 	}
 
 	for _, tt := range tests {
-		got, err := VersionStringToConstraints(tt.version)
+		got, err := versionStringToConstraints(tt.version)
```
mitchell-as, line 155: I got confused between this PR and #2659. This fix was destined for that PR and should have been re-merged into this PR so that it would not show up. Oh well.

pkg/platform/model/vcs.go
swagile1: We have Python versions of 2.7.17.10, for example. It seems unlikely we'll have unstable versions generally recommended in the foreseeable future though, so I suggest we stick with dot-separated integers for now and worry about names attached to release numbers later.

For what it's worth, `parseversion` from [langtools](https://github.com/ActiveState/langtools/) is probably the closest we have to a standard or contract on ingredient versions.

Naatan: @daved @mitchell-as so based on Shane's response I think the current changes are fine, no need to be more smarter /s.

daved (dismissed):

internal/runners/initialize/init.go
```diff
@@ -287,7 +287,10 @@ func deriveVersion(getKnownVersions knownVersionsFunc, lang language.Language, v
 		return "", errs.Wrap(err, "Unable to get known versions for language %s", lang.Requirement())
 	}
 
-	prefix := strings.Replace(strings.Replace(version, ".x", "", 1), ".X", "", 1) // strip any wildcard
+	prefix, _, _ := strings.Cut(version, ",") // only keep first part of e.g. ">=3.10,<3.11"
```
Naatan, line 290: I think we should drop all of this language verification logic we have here. With complex version numbers this gets way too involved for the client. We should just defer this to the buildplanner.

So if the user specified a version that can't be solved then that'll resolve in an error while sourcing the runtime. We should make sure that in this scenario the buildplanner gives an actionable error to the user though, and if not we need to file an enhancement with extensibility.

pkg/platform/model/buildplanner.go
```diff
@@ -361,16 +364,45 @@ func processBuildPlannerError(bpErr error, fallbackMessage string) error {
...
+		}}, nil
+	}
+
+	if !isWildcardVersion(version) {
+		// Ask the Platform to translate a string like ">=1.2,<1.3" into a list of requirements.
+		// Note that:
+		// - The given requirement name does not matter; it is not looked up.
+		// - The given language only needs to exist. It has no bearing on the translation and is not
+		//   present in the requirements read and returned.
+		dummyLanguage := language.Python3.Requirement()
```
Naatan, line 391: Can we actually provide this the name "dummy" or it has to be Python?

Can you make sure with Dashboard-squad that this logic is safe?

Naatan (changes requested):

pkg/platform/model/buildplanner.go
mitchell-as: We need to provide a valid language name. I tried "dummy" and it didn't work.

I confirmed with DS that this is safe. Their response was: "That service is unaware of any real world packages, so your example should be ok.  I can't promise that it won't change in the future to be smarter/interact with our ingredients DB to validate pcakage names, but I doubt that will happen, and if it did, it would likely fall under a "bad request" type response which you are going to handle any way."

internal/runners/initialize/init.go
mitchell-as: I don't think we can drop this basic logic because we need to figure out if a given version number is a partial one or not. For example: `state init org/project --language python@3.10` without any version checking would end up installing python 3.10.0 (not 3.10.12 or whatever the latest or recommended point release is; the solver appears to just tack on a ".0" to the end of what we gave it). The logic currently in place would detect that "3.10" is not a full version number and would append ".x" to the end, resulting in "3.10.x", which gives the latest point release.

mitchell-as:
New test failures are due to a new Platform change that disallows passwords to contain usernames.

internal/runners/initialize/init.go
Naatan: That is a problem for the API to address though, not us. And keep in mind that most users won't specify a version, this is already an advanced case. I don't want to have special solving logic around to address API limitations as it'll just lead to the API's never addressing those limitations and us having special workaround logic that needs to be maintained.

Naatan (changes requested):
I still wanna get rid of the logic raised here:

https://github.com/ActiveState/cli/pull/2663#discussion_r1283725782

We're simply doing too much as a client at the moment. I don't disagree with the desired user flow, but this is the wrong place for us to address it.

Naatan (dismissed):

mitchell-as:
An outdated unit test was failing, masked by existing integration test failures :( Please re-review.

Naatan (dismissed):

PR #2667, opened 2023-07-28
Report buildplan failures to DWH

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2060" title="DX-2060" target="_blank">TaskDX-2060</a>  We send runtime-debug data on buildplan failures
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -21,6 +21,7 @@ import (
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
+	bpModel "github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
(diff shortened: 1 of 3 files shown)


pkg/platform/runtime/runtime.go
```diff
@@ -193,6 +194,8 @@ func (r *Runtime) recordCompletion(err error) {
 		errorType = "solve"
 	case errs.Matches(err, &setup.BuildError{}) || errs.Matches(err, &buildlog.BuildError{}):
 		errorType = "build"
+	case errs.Matches(err, bpModel.BuildPlannerError{}):
```
MDrakos, line 197: Does this need to be a pointer?

MDrakos (approved):

pkg/platform/runtime/runtime.go
Naatan: Nice catch! Technically I believe it should work regardless, but better safe than sorry.

MDrakos (approved):

PR #2668, opened 2023-07-28
Create buildscripts at request time if they do not exist.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1997" title="DX-1997" target="_blank">TaskDX-1997</a>  As a devx engineer I don't have to consider old state tool versions whenever working with local orders
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildscript/file.go
+++ b/pkg/platform/runtime/buildscript/file.go
@@ -9,37 +9,57 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/pkg/localcommit"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
+	"github.com/ActiveState/cli/pkg/projectfile"
 )
 
-type DoesNotExistError struct{ error }
-
-func IsDoesNotExistError(err error) bool {
-	return errs.Matches(err, &DoesNotExistError{})
+func NewScriptFromProjectDir(dir string, auth *authentication.Auth) (*Script, error) {
+	return newScriptFromFile(filepath.Join(dir, constants.BuildScriptFileName), auth)
 }
 
-func NewScriptFromProjectDir(dir string) (*Script, error) {
-	return newScriptFromFile(filepath.Join(dir, constants.BuildScriptFileName))
-}
-
-func newScriptFromFile(path string) (*Script, error) {
-	data, err := fileutils.ReadFile(path)
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return nil, &DoesNotExistError{errs.New("Build script '%s' does not exist", path)}
-		}
+func newScriptFromFile(path string, auth *authentication.Auth) (*Script, error) {
+	if data, err := fileutils.ReadFile(path); err == nil {
+		return NewScript(data)
+	} else if !errors.Is(err, os.ErrNotExist) {
 		return nil, errs.Wrap(err, "Could not read build script")
 	}
-	return NewScript(data)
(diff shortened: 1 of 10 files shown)


pkg/platform/runtime/buildscript/file.go
```diff
@@ -9,37 +9,57 @@ import (
...
+		return NewScript(data)
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return nil, errs.Wrap(err, "Could not read build script")
+	}
 
-func newScriptFromFile(path string) (*Script, error) {
-	data, err := fileutils.ReadFile(path)
+	logging.Debug("Build script does not exist. Creating one.")
+	dir := filepath.Dir(path)
+	pjf, err := projectfile.FromPath(dir)
```
Naatan, line 32: In every scenario this is called you are passing the dir from the project package. Then you are constructing the projectfile based on this dir. Seems more productive if we can just pass the actual project instance.

pkg/platform/runtime/buildscript/file.go
```diff
@@ -9,37 +9,57 @@ import (
...
+	expr, err := buildplanner.GetBuildExpression(pjf.Owner(), pjf.Name(), commitId.String())
+	if err != nil {
+		return nil, errs.Wrap(err, "Unable to get the remote build expression")
+	}
+	script, err := NewScriptFromBuildExpression(expr)
+	if err != nil {
+		return nil, errs.Wrap(err, "Unable to convert build expression to build script")
+	}
+	err = fileutils.WriteFile(path, []byte(script.String()))
+	if err != nil {
```
Naatan, line 50:
```suggestion
	if err := fileutils.WriteFile(path, []byte(script.String())); err != nil {
```

Avoids reusing the err var and its type (which in turn avoids issues with nil assertions).

pkg/platform/runtime/buildscript/file.go
```diff
@@ -9,37 +9,57 @@ import (
...
+		return nil, errs.Wrap(err, "Unable to write build script")
+	}
+	return script, nil
 }
 
-func UpdateOrCreate(dir string, newExpr *buildexpression.BuildExpression) error {
+func UpdateOrCreate(dir string, newExpr *buildexpression.BuildExpression, auth *authentication.Auth) error {
```
Naatan, line 56: Can we name this `Update` now? I know technically it can still create, but that's implicit logic owned by the library going forward.

Naatan (changes requested):

pkg/platform/runtime/buildscript/file.go
```diff
@@ -13,32 +13,37 @@ import (
...
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
-	"github.com/ActiveState/cli/pkg/projectfile"
+	"github.com/ActiveState/cli/pkg/project"
 )
 
-func NewScriptFromProjectDir(dir string, auth *authentication.Auth) (*Script, error) {
-	return newScriptFromFile(filepath.Join(dir, constants.BuildScriptFileName), auth)
+type targeter interface { // note: cannot import runtime/setup.Targeter due to import cycle
+	ProjectDir() string
```
Naatan, line 20: If you add an alias for `Dir()` called `ProjectDir()` to the `Project` package then this interface can satisfy both the runtime target instance as well as the project instance, allowing you to consolidate the constructors and remove the direct dependency on the project package.

Naatan (changes requested):

pkg/project/project.go
```diff
@@ -210,6 +210,11 @@ func (p *Project) Dir() string {
 	return filepath.Dir(p.projectfile.Path())
 }
 
+// ProjectDir is an alias for Dir() to implement the runtime setup.Targeter interface.
```
Naatan, line 213:
```suggestion
// ProjectDir is an alias for Dir() to satisfy interfaces that may also target the setup.Targeter interface.
```

Naatan (changes requested):

Naatan (approved):

PR #2669, opened 2023-07-28
Add Parsing error support

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2029" title="DX-2029" target="_blank">StoryDX-2029</a>  Use new parsing errors from Build Planner
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/buildplanner.go
+++ b/pkg/platform/model/buildplanner.go
@@ -287,7 +287,7 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
 	}
 
 	if resp.Error != nil {
-		return "", locale.NewInputError("Failed to stage commit, API returned message: {{.V0}}", resp.Error.Message)
+		return "", locale.NewError("Failed to stage commit, API returned message: {{.V0}}", resp.Error.Message)
 	}
 
 	if resp.ParseError != nil {
--- a/pkg/platform/model/buildplanner.go
+++ b/pkg/platform/model/buildplanner.go
@@ -287,13 +287,13 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
 	}
 
 	if resp.Error != nil {
-		return "", errs.New("Failed to stage commit: " + resp.Error.Message)
+		return "", locale.NewInputError("Failed to stage commit, API returned message: {{.V0}}", resp.Error.Message)
 	}
 
 	if resp.ParseError != nil {
 		return "", locale.NewInputError(
 			"err_stage_commit_parse",
-			"The platform failed to parse the build expression, recieved the following message: {{.V0}}@{{.V1}}",
+			"The platform failed to parse the build expression, recieved the following message: {{.V0}}\n Path: {{.V1}}",
 			resp.ParseError.Message,
 			resp.ParseError.Path,
 		)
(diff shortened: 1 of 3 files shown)


pkg/platform/model/buildplanner.go
```diff
@@ -286,6 +286,19 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
 		return "", processBuildPlannerError(err, "failed to stage commit")
 	}
 
+	if resp.Error != nil {
+		return "", errs.New("Failed to stage commit: " + resp.Error.Message)
```
Naatan, line 290: Suggest we localize this, as it may have useful information for the user?

pkg/platform/model/buildplanner.go
```diff
@@ -286,6 +286,19 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
...
+	}
+
+	if resp.ParseError != nil {
+		return "", locale.NewInputError(
+			"err_stage_commit_parse",
+			"The platform failed to parse the build expression, recieved the following message: {{.V0}}@{{.V1}}",
```
Naatan, line 296: It feels a little weird to `@` these two fields. That seems to make a strong assumption about the format of the values returned.

Naatan (changes requested):

pkg/platform/model/buildplanner.go
```diff
@@ -287,13 +287,13 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
 	}
 
 	if resp.Error != nil {
-		return "", errs.New("Failed to stage commit: " + resp.Error.Message)
+		return "", locale.NewInputError("Failed to stage commit, API returned message: {{.V0}}", resp.Error.Message)
```
Naatan, line 290:
```suggestion
		return "", locale.NewError("Failed to stage commit, API returned message: {{.V0}}", resp.Error.Message)
```

Unless we know for certain that this is an input error?

pkg/platform/model/buildplanner.go
```diff
@@ -287,13 +287,13 @@ func (bp *BuildPlanner) StageCommit(params StageCommitParams) (strfmt.UUID, erro
...
 	}
 
 	if resp.ParseError != nil {
 		return "", locale.NewInputError(
 			"err_stage_commit_parse",
-			"The platform failed to parse the build expression, recieved the following message: {{.V0}}@{{.V1}}",
+			"The platform failed to parse the build expression, recieved the following message: {{.V0}}\n Path: {{.V1}}",
```
Naatan, line 296:
```suggestion
			"The platform failed to parse the build expression, received the following message: {{.V0}}. Path: {{.V1}}",
```

Naatan (changes requested):

pkg/platform/model/buildplanner.go
MDrakos: I recall we didn't want to flood rollbar with buildplanner errors and you followed up with PB to ensure they are being logged by them. That was my reasoning for changing it but I'm happy to make it a regular error if that's not the case.

Naatan: That's right, but this was handled in one place through the buildplan error type, which I assumed is wrapped around this. I don't want to have this special handling of errors implemented in different ways throughout the package.

Naatan (changes requested):

Naatan (approved):

PR #2671, opened 2023-07-28
Fix install e2e tests 

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2061" title="DX-2061" target="_blank">BugDX-2061</a>  Nightly failure on installer tests
  </td></table>
  <br />


Naatan (changes requested):
This is doing way too much for a bug fix. Not necessarily against accepting these changes, but please provide some summary of what happened here as it's obviously doing more than fixing a bug.

daved:
For sure. Debugging this was made difficult for me due to it needing to run in CI to expose malfunctions. Further, the way that the scripts were being relied on by the asy scripts and tests meant that there was some hidden knowledge about directory locations . I did my best to simplify the default values (and flag handling, since that was quite noisy) so that things are solid now and easier to read/maintain going forward.

internal/fileutils/fileutils.go
```diff
@@ -17,21 +17,22 @@ import (
...
-	"github.com/gofrs/flock"
-	"github.com/thoas/go-funk"
 )
 
 // nullByte represents the null-terminator byte
 const nullByte byte = 0
 
 // FileMode is the mode used for created files
-const FileMode = 0644
+const FileMode = 0o644
```
daved, line 35: This and similar changes are due to the current recommendation that octal notation be more explicit. My linter kept changing this, and I got tired of trying to fix or avoid it.

internal/fileutils/fileutils.go
```diff
@@ -265,6 +268,11 @@ func CopyFile(src, target string) error {
 	if err != nil {
 		return errs.Wrap(err, "out.Close failed")
 	}
+
```
daved, line 271: I found that, in two cases, I was recreating logic that was here simply because the file permissions were not be copied along with the files themselves.

internal/fileutils/fileutils.go
```diff
@@ -265,6 +268,11 @@ func CopyFile(src, target string) error {
...
+	if err := os.Chmod(out.Name(), inInfo.Mode().Perm()); err != nil {
+		return errs.Wrap(err, "chmod failed")
+	}
```
daved, line 274: I found that, in two cases, I was recreating logic that was here simply because the file permissions were not be copied along with the files themselves.

internal/fileutils/fileutils.go
```diff
@@ -373,8 +381,7 @@ func PrependToFile(filepath string, data []byte) error {
 // AmendFile amends data to a file, supports append, or prepend
 func AmendFile(filePath string, data []byte, flag AmendOptions) error {
 	switch flag {
-	case
-		AmendByAppend, AmendByPrepend:
+	case AmendByAppend, AmendByPrepend:
```
daved, line 384: This and the change at line 52 were also formatting changes that I stopped trying to "unfix".

scripts/ci/update-generator/main.go
```diff
@@ -140,32 +128,33 @@ func createInstaller(outputPath, channel, platform string) error {
...
+	)
+
+	flag.StringVar(
+		&platform, "platform", platform,
+		"Target platform in the form OS-ARCH. Defaults to running os/arch or the combination of the environment variables GOOS and GOARCH if both are set.",
+	)
+	flag.StringVar(&inDir, "i", inDir, "Override directory to gather payload from.")
+	flag.StringVar(&outDir, "o", outDir, "Override directory to output archive to.")
+	flag.StringVar(&branch, "b", branch, "Override target branch. (Branch to receive update.)")
+	flag.StringVar(&version, "v", version, "Override version number for this update.")
```
daved, line 146: During debugging, I got lost following what might be going wrong with the global+init setup of flags. The above design matches that of payload-generator, and it's less noisy, so I hope it's acceptable.

activestate.generators.yaml
```diff
@@ -94,25 +94,34 @@ scripts:
...
+      TEST_INSTALLER=${TEST_STAGED_DIR}/$constants.BUILD_INSTALLER_TARGET
       go build -o $TEST_STATE_TOOL ./internal/testhelpers/installation/fakeversioncmd
       cp -a $TEST_STATE_TOOL $TEST_STATE_SVC
       cp -a $TEST_STATE_TOOL $TEST_STATE_EXEC
       cp -a $BUILD_TARGET_DIR/$constants.BUILD_INSTALLER_TARGET $TEST_INSTALLER
 
-      go run scripts/ci/update-generator/main.go -v ${TEST_VERSION} -o ./build/test-update ./build/test/state-install
+      $scripts.generate-payload -v ${TEST_VERSION} -o ${TEST_PREPARED_DIR} -i ${TEST_STAGED_DIR}
+
+      go run scripts/ci/update-generator/main.go -v ${TEST_VERSION} -o ${TEST_ARCHIVE_DIR} -i ${TEST_PREPARED_DIR}
```
daved, line 118: The flag handling in payload-generator, and the flag setting here (and below) are the real crux of this PR.

scripts/ci/payload-generator/main.go
```diff
@@ -15,6 +15,9 @@ import (
 )
 
 var (
+	defaultInputDir  = filepath.Join(environment.GetRootPathUnsafe(), "build")
+	defaultOutputDir = filepath.Join(defaultInputDir, "payload", "state-install")
```
Naatan, line 19: I'd rather not bury it this deep. It should just be `build/<dir>`. I'm guessing this is for the generate-test-update logic? Suggest using `build/test-payload` for that, and `build/payload` here.

activestate.generators.yaml
Naatan, line 118: I feel like this is a payload generator on top of a payload generator. The fact that we are now generating this payload with a script should mean that this code should've become much simpler rather than more complex. There shouldn't need to be an interim stage-components dir. It seems the only thing that this is doing that's custom is inject `fakeversioncmd`, so I would suggest running `generate-payload` as normal and then simply injecting the custom cmd at the end.

scripts/ci/payload-generator/main.go
```diff
@@ -36,40 +39,40 @@ func main() {
 
 func run() error {
 	var (
+		inDir   = defaultInputDir
+		outDir  = defaultOutputDir
 		branch  = constants.BranchName
 		version = constants.Version
 	)
 
+	flag.StringVar(&inDir, "i", inDir, "Override directory to gather payload components from.")
```
Naatan, line 48: Once we address my other comment there should be no need anymore for this flag. A big point of the payload-generator is that this logic is centralized. By giving it a custom input path we're just creating more generators in front of generators.

Naatan (changes requested):

scripts/ci/payload-generator/main.go
daved: In general, the archive is expected to be in a directory called "state-install". So, we can shorten this to build/state-install, or the update generator can copy the files from whatever directory we want into a temp directory "{tmp}/state-install", so the final archive is correct.

activestate.generators.yaml
daved: I moved the fake cmd copying into the payload generator as discussed in stand. It works nicely for generate-test-update, and has no effect on generate-update. Tests are passing and generate-test-update is now slimmer.

Naatan (changes requested):
Sorry I'm still not a fan of this approach. I still don't see why we can't keep the payload generator dumb and simple and just have `generate-test-update` splice in the fake exec with a manual copy after the payload was generated. This is very specific to `generate-test-update`.

The only reason payload-generator exists is so we can link up the types. It is not meant to do anything other than copy files over to opinionated locations. The complexity being introduced here just isn't warranted.

daved:
@Naatan Ok, I got something simpler in a certain way. I'm not exactly sure it works (the state-{goos}-{goarch}{version}.tar.gz archives seem correct, the state-installer.tar.gz archive only has the installer). I think we should not care so long as CI is working for now and file a story. We could even not file or defer the filed story since this is for local testing and it's likely we'd all use CI for now.

scripts/ci/payload-generator/main.go
Naatan, line 48: I think we can drop this now?

scripts/ci/update-generator/main.go
```diff
@@ -140,32 +128,33 @@ func createInstaller(outputPath, channel, platform string) error {
...
+		outDir   = defaultOutputDir
+		branch   = constants.BranchName
+		version  = constants.Version
+	)
+
+	flag.StringVar(
+		&platform, "platform", platform,
+		"Target platform in the form OS-ARCH. Defaults to running os/arch or the combination of the environment variables GOOS and GOARCH if both are set.",
+	)
+	flag.StringVar(&inDir, "i", inDir, "Override directory to gather payload from.")
```
Naatan, line 143: Same here

Naatan (changes requested):
Note I added some integration tests to run.. fingers crossed

scripts/ci/payload-generator/main.go
daved: @Naatan I think this still could use your guidance.

scripts/ci/payload-generator/main.go
daved: I've removed this one.

scripts/ci/update-generator/main.go
daved: I think we should keep this one available since the output location of the payload-generator script can be set, and this input flag would then necessarily need to be set.

scripts/ci/payload-generator/main.go
Naatan: Where is this requirement of it having to be in a directory named "state-install" coming from? It sounds like whatever is creating this requirement needs to be updated.

Naatan (approved):

PR #2676, opened 2023-08-01
Implement edit tests for state publish and fix related issues

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1944" title="DX-1944" target="_blank">StoryDX-1944</a>  We have integration tests for `state publish --edit`
  </td></table>
  <br />

### DIFF
--- a/internal/gqlclient/gqlclient.go
+++ b/internal/gqlclient/gqlclient.go
@@ -153,7 +153,7 @@ func (c *Client) RunWithContext(ctx context.Context, request Request, response i
 	name := strutils.Summarize(request.Query(), 25)
 	defer profile.Measure(fmt.Sprintf("gqlclient:RunWithContext:(%s)", name), time.Now())
 
-	if fileRequest, ok := request.(RequestWithFiles); ok && len(fileRequest.Files()) > 0 {
+	if fileRequest, ok := request.(RequestWithFiles); ok {
 		return c.runWithFiles(ctx, fileRequest, response)
 	}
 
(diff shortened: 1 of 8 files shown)


internal/gqlclient/gqlclient.go
```diff
@@ -153,7 +153,7 @@ func (c *Client) RunWithContext(ctx context.Context, request Request, response i
 	name := strutils.Summarize(request.Query(), 25)
```
Naatan, line 153: Changes in this file are to facilitate us calling the API without any files to upload.

Because the API was specifically architected around file uploads we can't just use the normal graphql route as the API simply isn't expecting the request to occur in that format.

internal/gqlclient/gqlclient.go
```diff
@@ -304,20 +305,46 @@ func (c *Client) runWithFiles(ctx context.Context, gqlReq RequestWithFiles, resp
...
+	var res *http.Response
+	resErrChan := make(chan error)
+	go func() {
+		var err error
+		res, err = http.DefaultClient.Do(req)
+		resErrChan <- err
+	}()
+
+	// Due to the streaming uploads the request error can happen both before and after the http request itself, hence
+	// the creative select case you see before you.
```
Naatan, line 318: Without this the request would hang forever if there are input issues, as it would hang waiting for a response that would never come.

internal/runners/packages/search.go
```diff
@@ -153,8 +156,12 @@ func formatSearchResults(packages []*model.IngredientAndVersion) *searchOutput {
 	}
 
 	for i, pack := range packages {
+		name := filterNilStr(pack.Ingredient.Name)
+		if showNamespace {
+			name = fmt.Sprintf("%s/%s", *pack.Ingredient.PrimaryNamespace, name)
+		}
```
Naatan, line 162: If you search for a custom namespace it will respond with namespaces as part of the names. This helped me debug, but also felt like a useful enhancement.

test/integration/publish_int_test.go
```diff
@@ -24,7 +27,7 @@ func (suite *PublishIntegrationTestSuite) TestPublish() {
 	suite.OnlyRunForTags(tagsuite.Publish)
 
 	// For development convenience, should not be committed without commenting out..
-	// os.Setenv(constants.APIHostEnvVarName, "staging.activestate.build")
+	// os.Setenv(constants.APIHostEnvVarName, "pr11496.activestate.build")
```
mitchell-as, line 30: Do you still want to keep this?

mitchell-as (approved):
Looks fine to me. Nicely done :)

test/integration/publish_int_test.go
Naatan: Yeah just for convenience for now. We'll likely be able to drop this once we have a testing strategy for staging.

PR #2677, opened 2023-08-01
Update build in progress reporting

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2033" title="DX-2033" target="_blank">StoryDX-2033</a>  We give accurate build progress
  </td></table>
  <br />


pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+				if _, ok := deps[id]; !ok {
+					continue
+				}
+				uniqueDeps = append(uniqueDeps, id)
+			}
+
+			info, err := getSourceInfo(a.GeneratedBy, lookup)
+			if err != nil {
+				return nil, errs.Wrap(err, "Could not resolve source information")
+			}
```
Naatan, line 357: Please add comments explaining what this is doing.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+		}
+
+		_, ok := result[strfmt.UUID(a.NodeID)]
+		if !ok && a.Status != model.ArtifactNotSubmitted {
+			deps := make(map[strfmt.UUID]struct{})
+			for _, depID := range a.RuntimeDependencies {
+				artifact, ok := lookup[depID].(*model.Artifact)
+				if ok {
+					if artifact.MimeType == model.XActiveStateBuilderMimeType {
+						logging.Debug("Skipping active state builder artifact: %s", depID)
```
Naatan, line 331: You're logging this case but not the one on line 321?

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+		if a.MimeType == model.XActiveStateBuilderMimeType {
+			continue
+		}
+
+		_, ok := result[strfmt.UUID(a.NodeID)]
+		if !ok && a.Status != model.ArtifactNotSubmitted {
+			deps := make(map[strfmt.UUID]struct{})
+			for _, depID := range a.RuntimeDependencies {
+				artifact, ok := lookup[depID].(*model.Artifact)
+				if ok {
```
Naatan, line 329: So if we can't find it in our lookup tables we include it anyway? Doesn't that suggest we'd be including artifacts that aren't linked up?

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+			if err != nil {
+				return nil, errs.Wrap(err, "Could not resolve source information")
+			}
+
+			result[strfmt.UUID(a.NodeID)] = artifact.Artifact{
+				ArtifactID:       strfmt.UUID(a.NodeID),
+				Name:             info.Name,
+				Namespace:        info.Namespace,
+				Version:          &info.Version,
+				RequestedByOrder: true,
```
Naatan, line 364: This is almost certainly false, but can we inherit that info rather than construct it ourselves?

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+				GeneratedBy:      a.GeneratedBy,
+				Dependencies:     uniqueDeps,
+			}
+		}
+	}
+
+	return result, nil
+}
+
+func generateBuildtimeDependencies(depdendencyID strfmt.UUID, lookup map[strfmt.UUID]interface{}, result map[strfmt.UUID]struct{}) (map[strfmt.UUID]struct{}, error) {
```
Naatan, line 374: Please add lots of comments

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+				for id := range recursiveDeps {
+					deps[id] = struct{}{}
+				}
+			}
+
+			var uniqueDeps []strfmt.UUID
+			for id := range deps {
+				if _, ok := deps[id]; !ok {
+					continue
+				}
```
Naatan, line 350: This conditional seems redundant, how would this ever not find the entry?

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+						continue
+					}
+				}
+
+				deps[depID] = struct{}{}
+				recursiveDeps, err := generateBuildtimeDependencies(depID, lookup, deps)
+				if err != nil {
+					return nil, errs.Wrap(err, "Could not resolve runtime dependencies for artifact: %s", depID)
+				}
+				for id := range recursiveDeps {
```
Naatan, line 341: Does this work when `generateBuildTimeDepenencies` returns `nil, nil`?

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+		result[depID] = struct{}{}
+		_, err := generateBuildtimeDependencies(depID, lookup, result)
+		if err != nil {
+			return nil, errs.Wrap(err, "Could not build map for runtime dependencies of artifact %s", artifact.NodeID)
+		}
+	}
+
+	step, ok := lookup[artifact.GeneratedBy].(*model.Step)
+	if !ok {
+		_, ok := lookup[artifact.GeneratedBy].(*model.Source)
```
Naatan, line 400: In this scenario if `ok==true` then `step` will be nil and line 409 would panic.

pkg/platform/runtime/setup/setup.go
```diff
@@ -502,6 +503,13 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
 				artifactsToInstall = append(artifactsToInstall, a.ArtifactID)
 			}
 		}
+
+		// We also caclulate the artifacts to be built which includes more than the runtime artifacts.
+		// This is used to determine if we need to show the "build in progress" screen.
+		buildtimeArtifacts, err = buildplan.BuildtimeArtifacts(buildResult.Build)
+		if err != nil {
+			return nil, nil, errs.Wrap(err, "Could not get buildtime artifacts")
+		}
```
Naatan, line 512: If the build is ready then we don't need to wait for a build and including build-time artifacts in the progress reporting is moot as it would instantly complete anyway. It only makes sense to do this if wen can identify them when a build is ongoing.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: That's what the function docstring is doing.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: The lookup is _only_ used to validate the mime type if it's an artifact. Otherwise, we just use the dependency ID.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: Yes, you're right this should be false. We could construct this by checking for this node ID in the runtime closure but we would either have to pass that into this function or recalculate it here. It looks like this field is only used when calculating the changeset... do we know if what kind of impact this could have?

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -291,3 +291,132 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+	}
+	for _, source := range build.Sources {
+		lookup[source.NodeID] = source
+	}
+
+	for _, a := range build.Artifacts {
+		// Since we are using the BuildLogStreamer, we need to add all of the
+		// artifacts that have been submitted to be built except for the artifacts
+		// of the builder mime type as the BuildLogStreamer does not send events
+		// for builder artifacts.
```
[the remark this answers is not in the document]
MDrakos: Yeah fair enough. Added an argument to help is determine if we want to include builder artifacts or not.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: Let me know if the docstring needs more detail.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: Whoops, yes the map should already be unique.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: By work do you mean not error? If so, then yes it will work.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: If `ok` is true we don't make it to line 409, we return `nil` before that.

pkg/platform/runtime/setup/setup.go
MDrakos: Is that not what's happening here? Above on line 490 we ask if the build is ready. In this else block `BuildReady` was false so assume the build is in progress and ask for the buildtime artifacts for progress reporting.

pkg/platform/runtime/buildplan/buildplan.go
Naatan: I meant the individual code. I didn't want to flag every piece of code, just saying that this whole chunk has a lot of logic and I'd like to be able to ready over it quickly by reading a short comment before each "block" explaining what this code is doing so that I can just read the code with context rather than read the code and trying to parse context.

pkg/platform/runtime/buildplan/buildplan.go
Naatan: I'm guessing right now it wouldn't have any impact because the "requested by order" stuff is not used anymore until buildplans can give us this information.

pkg/platform/runtime/buildplan/buildplan.go
Naatan: Yeah, not sure how go handles iterating over a `nil`.. I would've expected a panic.

pkg/platform/runtime/setup/setup.go
Naatan: D'oh.. github's diff was hiding the `} else {`..

Naatan (changes requested):
I think we're mostly there, only comment I'd still like to see addressed is this one:

https://github.com/ActiveState/cli/pull/2677#discussion_r1282493739

Even if we understand buildplans through and through, it's a lot to keep stored in your head. 6 months from now these comments will be super helpful in reminding us. 

My problem with the highlighted code as-is is that I'd first have to backward engineer it in my head and then I can read it with intend and meaning. I'd like to eliminate the backward engineering part with comments.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: Let me test it in the go playground just to be safe :P

MDrakos: Yeah, shouldn't panic: https://go.dev/play/p/jkOK0owZEcI

MDrakos:
> Even if we understand buildplans through and through, it's a lot to keep stored in your head. 6 months from now these comments will be super helpful in reminding us.
> 
> My problem with the highlighted code as-is is that I'd first have to backward engineer it in my head and then I can read it with intend and meaning. I'd like to eliminate the backward engineering part with comments.

That makes sense. I think it's clear to me because it's all I've been doing for the last few weeks :P

Added a bunch of comments. Happy to add more if there is something that isn't clear.

Naatan (approved):
Looks good!

PR #2678, opened 2023-08-03
Retrieve update info from svc

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1984" title="DX-1984" target="_blank">StoryDX-1984</a>  Update checks are moved to the state-svc
  </td></table>
  <br />


daved:
Recreating due to branch name concerns.

PR #2682, opened 2023-08-04
Retrieve update info from svc

daved:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1984" title="DX-1984" target="_blank">StoryDX-1984</a>  Update checks are moved to the state-svc
  </td></table>
  <br />


daved:
Because the checker now lives in the service, and the updater remains in the tool, some of the data/logic from the checker needed to be, instead, associated strictly with the updater. Related, the checker was interpreting the received update info and returning nil if it should not update, but that frustrated its integration with the service comms since nil has a special meaning for cached data.

The checker now always returns and caches the available information (non-current versions are not cached). The updater now always gets data from the checker and interprets whether it should update or not. To do that, handlers that rely on these types/functions will check a method on Update rather than checking the update info being nil or not.

Lastly, there is a change to an e2e test that, I think, implies something was not working correctly in the past. I'm not sure why it was passing previously. I may simply be too close to the code at the moment to spot why.

daved:
Also, which branch should this PR now target?

internal/updater/updater.go
```diff
@@ -23,54 +24,96 @@ import (
...
+	Channel string
+	Version string
+}
+
+func NewOriginDefault() *Origin {
+	return &Origin{
+		Channel: constants.BranchName,
+		Version: constants.Version,
+	}
+}
```
Naatan, line 48: What purpose does this have? Far as I can tell Origin will ALWAYS be what we have in constants. State-svc isn't meant to address requests from any state tools other than the one it shipped with.

Naatan (changes requested):

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -115,30 +116,64 @@ func (r *Resolver) Version(ctx context.Context) (*graph.Version, error) {
...
-		logging.Debug("No update info in cache")
-		return nil, nil
+	var (
+		update *updater.Update
+		ok     bool
+		err    error
+	)
+
+	switch {
+	case channel == constants.BranchName && version == "":
```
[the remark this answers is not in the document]
daved: constants.BranchName with empty version fits the use case of updating the current instance of the state tool to whatever is the latest version (a typical "available update" request when a command is run). If a version or alternate branch name is provided, we cannot reasonably (at this time) cache that data, so the request must be made directly.

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -115,30 +116,64 @@ func (r *Resolver) Version(ctx context.Context) (*graph.Version, error) {
...
+			update, err = updateFromChecker(r.cfg, r.an, channel, version)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			logging.Debug("Update info pulled from poller cache")
+		}
+
+	default:
+		logging.Debug("Update info requested for specific branch/version")
```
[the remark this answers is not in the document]
daved: This message is in contrast to the "standard" update request that should be cached.

internal/runners/update/lock.go
```diff
@@ -128,18 +136,21 @@ func confirmLock(prom prompt.Prompter) error {
...
-	info, err := updater.NewDefaultChecker(cfg, an).CheckFor(channel, version)
+
+	upd, err := svc.CheckUpdate(context.Background(), channel, version)
 	if err != nil {
 		return "", locale.WrapInputError(err, "err_update_fetch", "Could not retrieve update information, please verify that '{{.V0}}' is a valid channel.", channel)
 	}
 
-	if info == nil { // if info is empty, we are at the current version
+	avUpdate := updater.NewAvailableUpdate(upd.Channel, upd.Version, upd.Platform, upd.Path, upd.Sha256, "")
+	update := updater.NewUpdateCurrent(an, avUpdate)
```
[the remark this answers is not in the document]
daved: AvailableUpdate is the update information structure. Update is a context sharing structure that manages the actual mechanism of updating. Unmarshaling the AvailableUpdate data, when it was within the Update type itself, made working with it odd here: https://github.com/ActiveState/cli/pull/2682/files/ee23161abe170173ac49c0cb9911e9ead8789856#diff-90c7d941ebf622dffde2bb127e649d6fdb358b31bac15f3d9496c9305a895e5dR139-R150
By "odd" I mean that the type having data unmarshaled to it doesn't need analytics, but the update process does.

internal/updater/updater.go
daved: So far, this is only for testing. Along with renaming the update construction functions, I added a comment to the one that ends up be used by the package level test.

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -116,64 +116,55 @@ func (r *Resolver) Version(ctx context.Context) (*graph.Version, error) {
...
 	)
 
 	switch {
-	case channel == constants.BranchName && version == "":
-		update, ok = r.updatePoller.ValueFromCache().(*updater.Update)
-		if !ok || update == nil {
+	case desiredChannel == constants.BranchName && desiredVersion == "":
+		avUpdate, ok = r.updatePoller.ValueFromCache().(*updater.AvailableUpdate)
+		if !ok || avUpdate == nil {
+			avUpdate = &updater.AvailableUpdate{}
```
Naatan, line 140: This doesn't seem to do anything since you're returning on line 142.

cmd/state-svc/internal/resolver/resolver.go
```diff
@@ -116,64 +116,55 @@ func (r *Resolver) Version(ctx context.Context) (*graph.Version, error) {
...
+		avUpdate *updater.AvailableUpdate
+		ok       bool
+		err      error
 	)
 
 	switch {
-	case channel == constants.BranchName && version == "":
-		update, ok = r.updatePoller.ValueFromCache().(*updater.Update)
-		if !ok || update == nil {
+	case desiredChannel == constants.BranchName && desiredVersion == "":
```
Naatan, line 137: Nit: This switch case seems pointless considering there's only really one conditional.

internal/updater/updater.go
```diff
@@ -72,36 +73,54 @@ func NewAvailableUpdate(channel, version, platform, path, sha256, tag string) *A
 	}
 }
 
+func NewAvailableUpdateFromGraph(au *graph.AvailableUpdate) *AvailableUpdate {
+	if au == nil {
+		return &AvailableUpdate{}
+	}
+	return NewAvailableUpdate(au.Channel, au.Version, au.Platform, au.Path, au.Sha256, "")
+}
```
Naatan, line 81: Nit: Why add a dependency on `graph` here? Seems whatever is calling this should just be calling `NewAvailableUpdate()`.

internal/updater/updater.go
```diff
@@ -72,36 +73,54 @@ func NewAvailableUpdate(channel, version, platform, path, sha256, tag string) *A
...
-	return NewUpdate(an, NewOriginDefault(), avUpdate)
+func NewUpdate(an analytics.Dispatcher, avUpdate *AvailableUpdate) *Update {
+	return NewUpdateByOrigin(an, NewOriginDefault(), avUpdate)
 }
 
-func (u *Update) ShouldSkip() bool {
-	return os.Getenv(constants.ForceUpdateEnvVarName) != "true" &&
-		u.AvUpdate.Channel == u.Origin.Channel &&
-		u.AvUpdate.Version == u.Origin.Version
+func (u *Update) NotNeeded() bool {
```
Naatan, line 120: Nit: Suggest naming this `IsValid()` as well. "Not" functions tend to lead to double negatives.

test/integration/update_int_test.go
```diff
@@ -329,3 +347,42 @@ func (suite *UpdateIntegrationTestSuite) TestUpdateToCurrent() {
...
+		for _, substr := range substrs {
+			if strings.Contains(txt, substr) {
+				return
+			}
+			printf("Expected to find (iter %d):\n%q\nReceived:\n%q", i, substr, txt)
+		}
+
+		time.Sleep(pause)
+	}
+}
```
Naatan, line 388: This feels over-engineered for what is needed. It could just be a simple for loop with a break.

```
found := false
for x :=0; x<5; x++ {
		cp := ts.SpawnCmdWithOpts(stateExePath, spawnOpts...)
		cp.Wait(time.Second * 10)
		if strings.Contains(cp.Snapshot(), "Updating State Tool") {
			found = true
			break
		}
}
suite.Require().True(found, "Expecting to find 'Updating State Tool'")
``

That is roughly about as much code as you currently have to invoke the logic introduced here.

Code like this (repeatCheck) creates two problems, which is a price that may in some situations be worth paying:

- We need to maintain it
- Reading test code I now have to jump to this implementation to understand what the test is doing, where a simple for loop would've been self-explanatory

File: internal/runners/update/update.go [hunk 10]
```diff
@@ -45,17 +49,20 @@ func New(prime primeable) *Update {
...
 }
 
 func (u *Update) Run(params *Params) error {
 	// Check for available update
-	checker := updater.NewDefaultChecker(u.cfg, u.an)
-	up, err := checker.CheckFor(params.Channel, "")
+	upd, err := u.svc.CheckUpdate(context.Background(), params.Channel, "")
 	if err != nil {
-		return locale.WrapError(err, "err_update_check", "Could not check for updates.")
+		return locale.WrapInputError(err, "err_update_fetch", "Could not retrieve update information, please verify that '{{.V0}}' is a valid channel.", params.Channel)
```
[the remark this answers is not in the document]
Naatan: This comment was not addressed.

internal/runners/update/lock.go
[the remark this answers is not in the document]
Naatan: I don't think I agree with the rationale, but I also don't care that much. Could you just update the type `Update` to instead be called `UpdateInstaller` or something? Reading through the code the distinction between the two isn't very evident.

Naatan (changes requested):
Please also see responses to previous comments.

cmd/state-svc/internal/resolver/resolver.go
daved: I tend to prefer switch/case/default over if/else since, imo, it's more readable.

internal/updater/updater.go
daved: This is added as a convenience for callers so that they don't have to extract the gql response fields themselves as in line 80. It also provides simple nil handling.

internal/updater/updater.go
```diff
@@ -88,7 +88,7 @@ func (u *AvailableUpdate) Equals(origin *Origin) bool {
 	return u.Channel == origin.Channel && u.Version == origin.Version
 }
 
-type Update struct {
+type UpdateInstall struct {
```
Naatan, line 91:
```suggestion
type UpdateInstaller struct {
```

`UpdateInstall` suggests to me that it holds information about an update, which is exactly the confusion this rename was meant to address.

internal/runners/update/update.go
```diff
@@ -57,14 +57,19 @@ func (u *Update) Run(params *Params) error {
...
+			"Could not retrieve update information.",
+		), locale.Tl(
+			"err_tip_update_fetch", "Try again, and/or try restarting State Service",
+		))
 	}
 
-	update := updater.NewUpdate(u.an, updater.NewAvailableUpdateFromGraph(upd))
-	if update.NotNeeded() {
+	update := updater.NewUpdateInstall(u.an, updater.NewAvailableUpdateFromGraph(upd))
+	if !update.IsUseful() {
```
Naatan, line 69: I'm sorry to keep nitpicking, but this doesn't tell me anything. Please name this something that explains what it does. Eg. `ShouldInstall()`.

Imo `IsValid()` is fine too. I'm guessing you're avoiding it because you attached a specific meaning of "Does this struct hold valid values" to it, but it could also just mean "Is this valid for use".

Naatan (changes requested):
Nearly there, just want to get these names right, as I am stumbling over this just reviewing, so I'll stumble over it 6 months from now while trying to understand this code again.

daved:
Story filed for MacOS CI error: https://activestatef.atlassian.net/jira/software/c/projects/DX/issues/DX-2164

Naatan (approved):

MDrakos (approved):

daved:
I wonder why the symptoms seemed to increase for my PR.

PR #2684, opened 2023-08-04
Integration tests that create users should use a better password.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2078" title="DX-2078" target="_blank">BugDX-2078</a>  Nightly failure: Signup-related tests
  </td></table>
  <br />
 

Passwords can no longer contain usernames.


mitchell-as:
The existing test failure here is outside the scope of this ticket. All of the signup tests are passing now, which is all that matters in this ticket.

Naatan (approved):

PR #2685, opened 2023-08-07
Updated rollbar client library.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1983" title="DX-1983" target="_blank">TaskDX-1983</a>  Get rollbar IP logging to actually work
  </td></table>
  <br />
 

I looked into the history of our rollbar upgrades, and I only found a reference to downgrading the C# version of rollbar, which appeared to pull in tons of crap. I couldn't find anything similar to Go, and judging by the resuling `go.mod`/`go.sum`, it didn't pull in anything extra. I think we're okay.

### DIFF
--- a/internal/rollbar/rollbar.go
+++ b/internal/rollbar/rollbar.go
@@ -82,7 +82,6 @@ func SetupRollbar(token string) {
 	rollbar.SetCodeVersion(constants.Version)
 	rollbar.SetServerRoot("github.com/ActiveState/cli")
 	rollbar.SetLogger(&rollbar.SilentClientLogger{})
-	rollbar.SetCaptureIp(rollbar.CaptureIpFull)
 
 	// We can't use runtime.GOOS for the official platform field because rollbar sees that as a server-only platform
 	// (which we don't have credentials for). So we're faking it with a custom field untill rollbar gets their act together.
--- a/go.mod
+++ b/go.mod
@@ -49,7 +49,7 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/phayes/permbits v0.0.0-20190108233746-1efae4548023
 	github.com/posener/wstest v0.0.0-20180216222922-04b166ca0bf1
-	github.com/rollbar/rollbar-go v1.1.0
+	github.com/rollbar/rollbar-go v1.4.5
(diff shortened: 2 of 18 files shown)


internal/rollbar/rollbar.go
```diff
@@ -82,7 +82,6 @@ func SetupRollbar(token string) {
 	rollbar.SetCodeVersion(constants.Version)
 	rollbar.SetServerRoot("github.com/ActiveState/cli")
 	rollbar.SetLogger(&rollbar.SilentClientLogger{})
-	rollbar.SetCaptureIp(rollbar.CaptureIpFull)
```
mitchell-as, line 85: Rollbar already logs full IPs by default. No need for this setting.

Naatan:
@mitchell-as did you verify or find a smoking gun suggesting updating rollbar fixes the issue? The story is about fixing IP logging, which updating this lib is one "potential" solution for.

mitchell-as:
@Naatan 
- https://github.com/ActiveState/cli/pull/2685/files#diff-b99eeec970b7a387b5293a746a4d4b6c02f47bcb69a4efbbc3cb804ff7e44180R36-R39
- https://github.com/ActiveState/cli/pull/2685/files#diff-b99eeec970b7a387b5293a746a4d4b6c02f47bcb69a4efbbc3cb804ff7e44180R87-R111
- https://github.com/ActiveState/cli/pull/2685/files#diff-b99eeec970b7a387b5293a746a4d4b6c02f47bcb69a4efbbc3cb804ff7e44180R102

That last line was not being sent upstream to rollbar with our previous client library. I think there's something server-side that logs the IP. We don't actually send an IP up there.

Naatan:
@mitchell-as would you just mind verifying that this indeed fixes the IP logging? Just add a `multilog.Error()` call to a test file, trigger it and then check on rollbar for your entry.

mitchell-as:
The rollbar client does not have the ability to get the user's IP address, and the rollbar server does not log an incoming IP address. We are responsible for obtaining and sending the user's IP to rollbar, possible via the body's `data["request"]["user_ip"]` field. This field does not appear to be documented anywhere, so it might be magic.

PR #2691, opened 2023-08-09
Fix parsing errors

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2089" title="DX-2089" target="_blank">BugDX-2089</a>  build expression: Panic: interface conversion: interface {} is float64, not map[string]interface
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildexpression/buildexpression.go
+++ b/pkg/platform/runtime/buildexpression/buildexpression.go
@@ -275,6 +275,9 @@ func newAp(path []string, m map[string]interface{}) (*Ap, error) {
 	var name string
 	var argsInterface interface{}
 	for key, value := range m {
+		// The value of an application will always be a map[string]interface{}
+		// that represents the application's arguments. In a BuildExpression,
+		// an applications arguments are always key-value pairs.
 		valueMap, ok := value.(map[string]interface{})
 		if !ok {
 			continue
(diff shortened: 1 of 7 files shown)


pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -269,7 +275,11 @@ func newAp(path []string, m map[string]interface{}) (*Ap, error) {
 	var name string
 	var argsInterface interface{}
 	for key, value := range m {
-		if isAp(path, value.(map[string]interface{})) {
+		valueMap, ok := value.(map[string]interface{})
+		if !ok {
+			continue
```
Naatan, line 280: Can we add a comment explaining why this can be discarded?

Naatan (changes requested):

pkg/platform/runtime/buildexpression/buildexpression.go
Naatan: Sorry, I don't feel like this comment is addressing why we can discard it, it's only addressing why we expect this format. Why is encountering an unexpected format not an error condition?

Naatan (changes requested):

pkg/platform/runtime/buildexpression/buildexpression.go
MDrakos: You're correct this should be an error condition. Taking a closer look at the code it appears that it was incorrect. I've made some more updates.

Naatan (approved):

PR #2694, opened 2023-08-10
Executors should not setup environment (again) in an activated state.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2068" title="DX-2068" target="_blank">BugDX-2068</a>  State tool executors gobble up environment variables
  </td></table>
  <br />

### DIFF
--- a/cmd/state-exec/meta.go
+++ b/cmd/state-exec/meta.go
@@ -6,6 +6,9 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/hash"
+	"github.com/ActiveState/cli/internal/installation/storage"
 	"github.com/ActiveState/cli/pkg/platform/runtime/executors/execmeta"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/state-exec/meta.go
```diff
@@ -34,10 +37,24 @@ func newExecutorMeta(execPath string) (*executorMeta, error) {
...
+		projectCacheDir := filepath.Join(storage.CachePath(), hash.ShortHash(projectDir))
+		if projectCacheDir == execDir {
+			transformEnv = false // runtime environment variables already set
+		}
+	}
+
+	env := os.Environ()
+	if transformEnv {
+		env = transformedEnv(env, meta.Env)
+	}
```
daved, line 53: nit: Consider moving the conditional into a new function (e.g. "shouldTransformEnv"), then calling as `env := os.Environ(); if shouldTransformEnv() {; env = transformedEnv(env, meta.Env);}")`.

daved (approved):

PR #2701, opened 2023-08-11
Support number values in buildscripts.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2106" title="DX-2106" target="_blank">BugDX-2106</a>  Nightly failure: TestPullIntegrationTestSuite/TestPull_Merge
  </td></table>
  <br />
 

A recent change to fix buildexpression parser errors (supporting ints/floats) bubbled up to buildscripts, so now they have to support ints/floats. Whenever we get around to unifying their datastructures, issues like this should go away. For now, we have to manually intercede.

### DIFF
--- a/pkg/platform/runtime/buildscript/buildscript.go
+++ b/pkg/platform/runtime/buildscript/buildscript.go
@@ -3,6 +3,7 @@ package buildscript
 import (
 	"bytes"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/ActiveState/cli/internal/constants"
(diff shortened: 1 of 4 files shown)


pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -109,6 +111,9 @@ func (v *Value) String() string {
 	case v.Str != nil:
 		return *v.Str
 
+	case v.Number != nil:
+		return strconv.FormatFloat(*v.Number, 'G', -1, 64)
```
MDrakos, line 115: Could you please add a comment explaining the format values? Just to help avoid having to look up what these mean when we revisit this code.

MDrakos (changes requested):

MDrakos (approved):

PR #2702, opened 2023-08-11
Match the org/project case for checkouts, and org case for init when writing activestate.yaml files.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2076" title="DX-2076" target="_blank">BugDX-2076</a>  `state checkout` and `state init` should use the casing from the API rather than from user input
  </td></table>
  <br />

### DIFF
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -150,16 +150,21 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 
 	// Match the case of the organization.
 	// Otherwise the incorrect case will be written to the project file.
-	owner := params.Namespace.Owner
+	var owner string
 	orgs, err := model.FetchOrganizations()
 	if err != nil {
 		return errs.Wrap(err, "Unable to get the user's writable orgs")
 	}
 	for _, org := range orgs {
-		if strings.EqualFold(org.DisplayName, owner) {
+		if strings.EqualFold(org.DisplayName, params.Namespace.Owner) {
 			owner = org.DisplayName
 			break
 		}
+	}
+	if owner == "" {
+		return locale.NewInputError("err_invalid_org",
+			"The organization '{{.V0}}' either does not exist, or you do not have permissions to create a project in it",
+			params.Namespace.Owner)
 	}
 	namespace := project.Namespaced{Owner: owner, Project: params.Namespace.Project}
 
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -163,7 +163,7 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 	}
 	if owner == "" {
 		return locale.NewInputError("err_invalid_org",
-			"The organization '{{.V0}}' either does not exist, or you do not have permissions to create a project in it",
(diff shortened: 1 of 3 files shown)


internal/runners/initialize/init.go
```diff
@@ -148,9 +148,24 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
...
 
+	// Match the case of the organization.
+	// Otherwise the incorrect case will be written to the project file.
+	owner := params.Namespace.Owner
+	orgs, err := model.FetchOrganizations()
+	if err != nil {
+		return errs.Wrap(err, "Unable to get the user's writable orgs")
+	}
+	for _, org := range orgs {
+		if strings.EqualFold(org.DisplayName, owner) {
```
Naatan, line 159: We may as well error out if we can't find a matching org. Otherwise, we'll error out with the API call further down anyway, but the user will have an activestate.yaml which can't be used.

pkg/cmdlets/checkout/checkout.go
```diff
@@ -103,12 +103,23 @@ func (r *Checkout) Run(ns *project.Namespaced, branchName, cachePath, targetPath
...
 	}
 
+	// Match the case of the organization.
+	// Otherwise the incorrect case will be written to the project file.
+	owners, err := model.FetchOrganizationsByIDs([]strfmt.UUID{pj.OrganizationID})
+	if err != nil {
+		return "", errs.Wrap(err, "Unable to get the project's org")
+	}
+	if len(owners) == 0 {
+		return "", errs.New("Unable to get the project's org")
```
Naatan, line 113: This needs localization, as this will likely get quite a few hits (eg. misstyped org name).

It should also be an input error.

Naatan (changes requested):

pkg/cmdlets/checkout/checkout.go
mitchell-as: If the user mistypes an org name, an error will be triggered all the way back on line 62 (`model.FetchProjectByName()` failed). If we get to this point, the org exists, it's just that the lookup by UUID failed for whatever reason (I cannot see that happening). Regardless, I will localize this error to be safe.

internal/runners/initialize/init.go
```diff
@@ -150,17 +150,22 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
...
 	for _, org := range orgs {
-		if strings.EqualFold(org.DisplayName, owner) {
+		if strings.EqualFold(org.DisplayName, params.Namespace.Owner) {
 			owner = org.DisplayName
 			break
 		}
 	}
+	if owner == "" {
+		return locale.NewInputError("err_invalid_org",
+			"The organization '{{.V0}}' either does not exist, or you do not have permissions to create a project in it",
```
Naatan, line 166:
```suggestion
			"The organization '[ACTIONABLE]{{.V0}}[/RESET]' either does not exist, or you do not have permissions to create a project in it.",
```

Naatan (changes requested):

Naatan (approved):

PR #2703, opened 2023-08-11
Merge v0.41 into v0.42.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2076" title="DX-2076" target="_blank">BugDX-2076</a>  `state checkout` and `state init` should use the casing from the API rather than from user input
  </td></table>
  <br />
 

There was a merge conflict with projectfile's no longer having commit IDs in them thanks to the localcommit file.

### DIFF
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -153,14 +153,29 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


mitchell-as:
@Naatan please approve and force-merge this. I cannot do it. Some PRs I am allowed to force-merge, and some I am not allowed to. This is one I am not allowed to.

PR #2704, opened 2023-08-11
Send device ID as user IP to rollbar.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1983" title="DX-1983" target="_blank">TaskDX-1983</a>  Get rollbar IP logging to actually work
  </td></table>
  <br />
 

We cannot get IP, but this is good enough.

### DIFF
--- a/internal/rollbar/rollbar.go
+++ b/internal/rollbar/rollbar.go
@@ -82,7 +82,6 @@ func SetupRollbar(token string) {
 	rollbar.SetCodeVersion(constants.Version)
 	rollbar.SetServerRoot("github.com/ActiveState/cli")
 	rollbar.SetLogger(&rollbar.SilentClientLogger{})
-	rollbar.SetCaptureIp(rollbar.CaptureIpFull)
 
 	// We can't use runtime.GOOS for the official platform field because rollbar sees that as a server-only platform
 	// (which we don't have credentials for). So we're faking it with a custom field untill rollbar gets their act together.
@@ -91,6 +90,12 @@ func SetupRollbar(token string) {
 		// We're not a server, so don't send server info (could contain sensitive info, like hostname)
 		data["server"] = map[string]interface{}{}
 		data["platform_os"] = runtime.GOOS
+		data["request"] = map[string]string{
+			// Rollbar specially interprets the body["request"]["user_ip"] key, but it does not have to
+			// actually be an IP address. Use device_id for now. We may want to use real IP later if it
+			// is easily accessible (Rollbar does not provide it for us with non-http.Request errors).
+			"user_ip": uniqid.Text(),
+		}
 	})
 
 	source, err := storage.InstallSource()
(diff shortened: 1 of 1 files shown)


internal/rollbar/rollbar.go
```diff
@@ -82,7 +82,6 @@ func SetupRollbar(token string) {
 	rollbar.SetCodeVersion(constants.Version)
 	rollbar.SetServerRoot("github.com/ActiveState/cli")
 	rollbar.SetLogger(&rollbar.SilentClientLogger{})
-	rollbar.SetCaptureIp(rollbar.CaptureIpFull)
```
mitchell-as, line 85: This is the default configuration option anyway, so this is not needed.

internal/rollbar/rollbar.go
```diff
@@ -91,6 +90,12 @@ func SetupRollbar(token string) {
 		// We're not a server, so don't send server info (could contain sensitive info, like hostname)
 		data["server"] = map[string]interface{}{}
 		data["platform_os"] = runtime.GOOS
+		data["request"] = map[string]string{
+			// Rollbar specially interprets the body["request"]["user_ip"] key, but it does not have to
+			// actually be an IP address. Use device_id for now. We may want to use real IP later if it
+			// is easily accessible (Rollbar does not provide it for us with non-http.Request errors).
+			"user_ip": uniqid.Text(),
+		}
```
Naatan, line 98: I'm guessing `data["request"]` isn't normally set at this time? Can we just add a check to instantiate it if it's not set, and then set the user_ip.

I'm concerned there might be scenario's where we would override other request parameters.

Naatan (changes requested):

internal/rollbar/rollbar.go
mitchell-as: Correct. It's only set for an error involving an `http.Request`. I've added the check just in case though.

internal/rollbar/rollbar.go
```diff
@@ -91,6 +90,15 @@ func SetupRollbar(token string) {
...
 		data["platform_os"] = runtime.GOOS
+		if _, exists := data["request"]; !exists {
+			data["request"] = map[string]string{}
+		}
+		if request, ok := data["request"].(map[string]string); ok {
+			// Rollbar specially interprets the body["request"]["user_ip"] key, but it does not have to
+			// actually be an IP address. Use device_id for now. We may want to use real IP later if it
+			// is easily accessible (Rollbar does not provide it for us with non-http.Request errors).
+			request["user_ip"] = uniqid.Text()
+		}
```
Naatan, line 101:
```suggestion
		// Rollbar specially interprets the body["request"]["user_ip"] key, but it does not have to
		// actually be an IP address. Use device_id for now. We may want to use real IP later if it
		// is easily accessible (Rollbar does not provide it for us with non-http.Request errors).
		data["request"]["user_ip"] = uniqid.Text()
```

Doesn't seem like the check serves a purpose anymore due to the line 93 addition.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2704/files#r1291800212

internal/rollbar/rollbar.go
mitchell-as: The check is required. It's a compile error otherwise because data is a `map[string]interface{}`.

Naatan (approved):

PR #2705, opened 2023-08-14
Use proper org case for integration tests.

mitchell-as:
The checkout integration tests have a test case for handling different org/project cases.


MDrakos (approved):
Is this not behavior we want to support? If it is then we may need to follow-up with whatever broke this.

mitchell-as:
The failing tests are basically outdated and written wrong. Our `state checkout` integration tests handle what we want with regards to mismatched case.

PR #2710, opened 2023-08-16
Update shell rcfiles to indicate the user is already in an activated state.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2045" title="DX-2045" target="_blank">StoryDX-2045</a>  As a user I can expect to always see a clear indication that I am in a virtual environment
  </td></table>
  <br />
 

The PS1 prompt is only changed for the immediate subshell. Nested subshells do not inherit it, so it was not clear the user is still in an activated state.

cmd.exe does not need this because its PS1 equivalent is unmodified when opening another shell within a shell. I've also added a comment to the JIRA ticket explaining why I don't want to satisfy the Windows AC.

### DIFF
--- a/internal/assets/contents/shells/fishrc_append.fish
+++ b/internal/assets/contents/shells/fishrc_append.fish
@@ -6,8 +6,7 @@ set -xg {{$K}} "{{$V}}:$PATH"
 set -xg {{$K}} "{{$V}}"
 {{- end}}
 {{- end}}
-if test ! -z "${{.ActivatedEnv}}"
-  and test -f "${{.ActivatedEnv}}/{{.ConfigFile}}"
+if test ! -z "${{.ActivatedEnv}}"; test -f "${{.ActivatedEnv}}/{{.ConfigFile}}"
   echo "State Tool is operating on project ${{.ActivatedNamespaceEnv}}, located at ${{.ActivatedEnv}}"
 end
 # {{.Stop}}
--- a/internal/assets/contents/shells/fishrc_append.fish
+++ b/internal/assets/contents/shells/fishrc_append.fish
@@ -6,4 +6,8 @@ set -xg {{$K}} "{{$V}}:$PATH"
 set -xg {{$K}} "{{$V}}"
 {{- end}}
 {{- end}}
+if test ! -z "${{.ActivatedEnv}}"
+  and test -f "${{.ActivatedEnv}}/{{.ConfigFile}}"
+  echo "State Tool is operating on project ${{.ActivatedNamespaceEnv}}, located at ${{.ActivatedEnv}}"
+end
 # {{.Stop}}
--- a/internal/subshell/sscommon/rcfile_test.go
+++ b/internal/subshell/sscommon/rcfile_test.go
@@ -3,6 +3,7 @@ package sscommon
 import (
 	"fmt"
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -43,35 +44,50 @@ func TestWriteRcFile(t *testing.T) {
 		path           string
 		env            map[string]string
 	}
+
+	zsh := fmt.Sprintf(
+		`export PATH="foo:$PATH"
+if [[ ! -z "$%s" && -f "$%s/%s" ]]; then
+  echo "State Tool is operating on project $%s, located at $%s"
+fi`,
+		constants.ActivatedStateEnvVarName,
+		constants.ActivatedStateEnvVarName,
+		constants.ConfigFileName,
+		constants.ActivatedStateNamespaceEnvVarName,
+		constants.ActivatedStateEnvVarName)
+	if runtime.GOOS == "windows" {
+		zsh = strings.ReplaceAll(zsh, "\n", "\r\n")
+	}
+
 	tests := []struct {
 		name         string
 		args         args
-		want error
+		want         error
 		wantContents string
 	}{
 		{
 			"Write RC to empty file",
 			args{
-				"fishrc_append.fish",
+				"zshrc_append.sh",
 				fakeFileWithContents("", "", ""),
 				map[string]string{
 					"PATH": "foo",
 				},
 			},
 			nil,
-			fakeContents("", `set -xg PATH "foo:$PATH"`, ""),
+			fakeContents("", zsh, ""),
 		},
 		{
 			"Write RC update",
 			args{
-				"fishrc_append.fish",
+				"zshrc_append.sh",
 				fakeFileWithContents("before", "SOMETHING ELSE", "after"),
 				map[string]string{
 					"PATH": "foo",
 				},
 			},
 			nil,
-			fakeContents(strings.Join([]string{"before", "after"}, fileutils.LineEnd), `set -xg PATH "foo:$PATH"`, ""),
+			fakeContents(strings.Join([]string{"before", "after"}, fileutils.LineEnd), zsh, ""),
 		},
 	}
 	for _, tt := range tests {
--- a/internal/subshell/sscommon/rcfile_test.go
+++ b/internal/subshell/sscommon/rcfile_test.go
@@ -45,18 +45,18 @@ func TestWriteRcFile(t *testing.T) {
 		env            map[string]string
 	}
 
-	zsh := fmt.Sprintf(
-		`export PATH="foo:$PATH"
-if [[ ! -z "$%s" && -f "$%s/%s" ]]; then
+	fish := fmt.Sprintf(
+		`set -xg PATH "foo:$PATH"
+if test ! -z "$%s"; test -f "$%s/%s"
   echo "State Tool is operating on project $%s, located at $%s"
-fi`,
+end`,
 		constants.ActivatedStateEnvVarName,
 		constants.ActivatedStateEnvVarName,
 		constants.ConfigFileName,
 		constants.ActivatedStateNamespaceEnvVarName,
 		constants.ActivatedStateEnvVarName)
 	if runtime.GOOS == "windows" {
-		zsh = strings.ReplaceAll(zsh, "\n", "\r\n")
+		fish = strings.ReplaceAll(fish, "\n", "\r\n")
 	}
 
 	tests := []struct {
@@ -68,26 +68,26 @@ fi`,
 		{
 			"Write RC to empty file",
 			args{
-				"zshrc_append.sh",
+				"fishrc_append.fish",
 				fakeFileWithContents("", "", ""),
 				map[string]string{
 					"PATH": "foo",
 				},
 			},
 			nil,
-			fakeContents("", zsh, ""),
+			fakeContents("", fish, ""),
 		},
 		{
 			"Write RC update",
 			args{
-				"zshrc_append.sh",
+				"fishrc_append.fish",
 				fakeFileWithContents("before", "SOMETHING ELSE", "after"),
 				map[string]string{
 					"PATH": "foo",
 				},
 			},
 			nil,
-			fakeContents(strings.Join([]string{"before", "after"}, fileutils.LineEnd), zsh, ""),
+			fakeContents(strings.Join([]string{"before", "after"}, fileutils.LineEnd), fish, ""),
 		},
 	}
 	for _, tt := range tests {
--- a/test/integration/shell_int_test.go
+++ b/test/integration/shell_int_test.go
@@ -7,10 +7,10 @@ import (
 	"runtime"
 	"testing"
 
-	"github.com/ActiveState/cli/internal/condition"
 	"github.com/ActiveState/cli/internal/config"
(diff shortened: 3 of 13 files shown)


internal/assets/contents/shells/fishrc_append.fish
```diff
@@ -6,4 +6,8 @@ set -xg {{$K}} "{{$V}}:$PATH"
 set -xg {{$K}} "{{$V}}"
 {{- end}}
 {{- end}}
-# {{.Stop}}
\ No newline at end of file
+if test ! -z "${{.ActivatedEnv}}"
+  and test -f "${{.ActivatedEnv}}/{{.ConfigFile}}"
```
Naatan, line 10:
```suggestion
if test ! -z "${{.ActivatedEnv}}" and test -f "${{.ActivatedEnv}}/{{.ConfigFile}}"
```

I was tripping over this for a sec as it wasn't obvious where the condition ended. Assuming this isn't some fish shell oddity? Man I don't like it when languages rely on line breaks for delimiters :p

internal/subshell/sscommon/rcfile_test.go
```diff
@@ -43,35 +44,50 @@ func TestWriteRcFile(t *testing.T) {
...
-				"fishrc_append.fish",
+				"zshrc_append.sh",
 				fakeFileWithContents("before", "SOMETHING ELSE", "after"),
 				map[string]string{
 					"PATH": "foo",
 				},
 			},
 			nil,
-			fakeContents(strings.Join([]string{"before", "after"}, fileutils.LineEnd), `set -xg PATH "foo:$PATH"`, ""),
+			fakeContents(strings.Join([]string{"before", "after"}, fileutils.LineEnd), zsh, ""),
```
Naatan, line 90: I don't think it makes sense to change the original tests here, they were testing something different, it seems. 

You could certainly add some tests here, but this feels like it's better covered with an integration test where we assert that the modifications were made rather than that they exactly match some expected output letter for letter, as that could become painful to maintain.

test/integration/shell_int_test.go
```diff
@@ -276,6 +278,53 @@ func (suite *ShellIntegrationTestSuite) SetupRCFile(ts *e2e.Session) {
...
+	suite.Assert().NotContains(cp.TrimmedSnapshot(), "State Tool is operating on project")
+
+	binary := ss.Binary() // platform-specific shell (zsh on macOS, bash on Linux, etc.)
+	// Cannot run bare binary because it will not load the rcFile in ts.Dirs.HomeDir.
+	// Instead, tell the shell to load rcFile. This is not needed in a non-test environment.
+	switch ss.Shell() {
+	case bash.Name:
+		binary = fmt.Sprintf("%s --rcfile %s", binary, rcFile)
+	case zsh.Name:
+		binary = fmt.Sprintf("ZDOTDIR=%s %s", filepath.Dir(rcFile), binary)
```
Naatan, line 319: We shouldn't be specifying an RC file here. The use-case we're testing is the user running a new subshell (eg. just running `bash`), so this test should do that.

Naatan (changes requested):

test/integration/shell_int_test.go
mitchell-as: There's a comment 6 lines above explaining why I need it for the test: https://github.com/ActiveState/cli/pull/2710/files/3cfa3a53d6ba9ccb67ab832f4208c24e9df8c332#diff-5a8d1281306bbab65982eb86bcf4706c963f88bb9821aeb35bebacd03d2ea8b6R313-R314

Running a bare `zsh` sources the machine's *~/.zshrc*, which is unmodified because our tests use a test-specific user home directory.

internal/subshell/sscommon/rcfile_test.go
mitchell-as: I'm not sure I agree since we need to treat modifications to user RC files VERY carefully. The last thing we want to do is have a superficial check that verifies our boundaries were written between a line or two we expect to see, but we accidentally deleted a line or reformatted something that breaks the user's RC file. Doing absolute checks is safer, and is worth the maintenance I think.

internal/assets/contents/shells/fishrc_append.fish
mitchell-as: I used this reference: https://fishshell.com/docs/current/cmds/if.html#example

However, there's a `;` delimiter that acts like `and` so I'll use that.

test/integration/shell_int_test.go
Naatan: That seems like a limitation in our test suite though. I just tried `HOME=/foo zsh` and it tried to source its rc file from `/foo`. I'd say we should do that for our test suites, but considering that may be too big a change for this PR I suggest just doing it for these tests.

test/integration/shell_int_test.go
```diff
@@ -288,13 +291,19 @@ func (suite *ShellIntegrationTestSuite) TestNestedShellNotification() {
 
 	cfg, err := config.New()
 	suite.Require().NoError(err)
+	if runtime.GOOS == "darwin" && condition.OnCI() {
+		cfg.Set(subshell.ConfigKeyShell, "zsh") // GitHub actions runs on bash, so override with zsh
```
Naatan, line 295: I'd rather we use the `SHELL=zsh` environment variable, as it's more future-proof.

test/integration/shell_int_test.go
```diff
@@ -288,13 +291,19 @@ func (suite *ShellIntegrationTestSuite) TestNestedShellNotification() {
...
+	}
 
+	os.Setenv(constants.HomeEnvVarName, ts.Dirs.HomeDir)
+	defer func() { os.Unsetenv(constants.HomeEnvVarName) }()
```
Naatan, line 299: This is not safe for other tests running in parallel. You can set it with AppendEnv though.

test/integration/shell_int_test.go
```diff
@@ -288,13 +291,19 @@ func (suite *ShellIntegrationTestSuite) TestNestedShellNotification() {
...
 	ss := subshell.New(cfg)
 	err = subshell.ConfigureAvailableShells(ss, cfg, nil, sscommon.InstallID, true) // mimic installer
 	suite.Require().NoError(err)
```
Naatan, line 302: Why is this necessary for these tests?

Naatan (changes requested):

test/integration/shell_int_test.go
mitchell-as: This is needed to put the .zshrc and .bashrc script in the test userhome directory. If I don't have this, the test will fail with ".zshrc not found".

test/integration/shell_int_test.go
mitchell-as: I've tried to minimize time spent in a modified env state. It's needed at the function-level for `ss.RcFile()` and spawn-process-level for `user.HomeDir()`.

test/integration/shell_int_test.go
Naatan: Can we instead just "touch" those files? This feels like we're throwing too much at it. Especially in tests the variables should be as minimal as possible.

mitchell-as: No. I need the content in the *_append.* files. Otherwise spawning a nested subshell will not print the notice.

test/integration/shell_int_test.go
Naatan: Ok, please find an alternative to using `ss.RcFile()` that doesn't require `os.Setenv()`. This is just a test so it's ok if we're hard coding some stuff here so long as it's not prone to breaking all the time.

As for the process-level; `AppendEnv()` should be able to handle that.

Minimizing the window doesn't change that it can impact other tests in ways that will be really hard to identify after the fact.

Naatan (changes requested):
See comment responses.

test/integration/shell_int_test.go
```diff
@@ -276,6 +280,64 @@ func (suite *ShellIntegrationTestSuite) SetupRCFile(ts *e2e.Session) {
...
+	cfg, err := config.New()
+	suite.Require().NoError(err)
+
+	// Setup subshell. Since our CI tests are running in bash, override with zsh on macOS.
+	var ss subshell.SubShell
+	env := []string{"ACTIVESTATE_CLI_DISABLE_RUNTIME=false"}
+	if runtime.GOOS != "darwin" || !condition.OnCI() {
+		ss = subshell.New(cfg)
+	} else {
+		cfg.Set(subshell.ConfigKeyShell, "zsh")
```
mitchell-as, line 300: Cannot set SHELL env here via `os.Setenv()` because that might interfere with parallel tests.

test/integration/shell_int_test.go
mitchell-as: We actually cannot do this because `os.Setenv("SHELL")` will affect parallel tests.

test/integration/shell_int_test.go
mitchell-as: I've hardcoded creating and populating shellrc files like the installer would.

test/integration/shell_int_test.go
```diff
@@ -292,27 +291,33 @@ func (suite *ShellIntegrationTestSuite) TestNestedShellNotification() {
...
 	suite.Require().NoError(err)
 
+	// Setup subshell. Since our CI tests are running in bash, override with zsh on macOS.
 	var ss subshell.SubShell
 	env := []string{"ACTIVESTATE_CLI_DISABLE_RUNTIME=false"}
 	if runtime.GOOS != "darwin" || !condition.OnCI() {
 		ss = subshell.New(cfg)
 	} else {
-		os.Setenv("SHELL", "zsh") // GitHub actions runs on bash, so override with zsh
+		cfg.Set(subshell.ConfigKeyShell, "zsh")
```
Naatan, line 300: Please get rid of the subshell package in these tests. We can just make assumptions based on OS. eg. Mac = zsh, Linux = bash, Windows = cmd.

The subshell package adds too much logic to these tests imo. I'd rather keep this dead simple and hardcoded so the tests leave little to the imagination and we don't accidentally break them without realizing it when eg. updating the subshell package.

test/integration/shell_int_test.go
```diff
@@ -292,27 +291,33 @@ func (suite *ShellIntegrationTestSuite) TestNestedShellNotification() {
...
+	case "zsh":
+		rcFile = filepath.Join(ts.Dirs.HomeDir, ".zshrc")
+		rcFileAppend = "zshrc_append.sh"
+	default:
+		suite.Fail("Unknown subshell: %s", ss.Shell())
+	}
 	suite.Require().Equal(filepath.Dir(rcFile), ts.Dirs.HomeDir, "rc file not in test suite homedir")
-	suite.Require().FileExists(rcFile)
+	fileutils.TouchFileUnlessExists(rcFile)
+	sscommon.WriteRcFile(rcFileAppend, rcFile, sscommon.InstallID, nil)
```
Naatan, line 320: I don't follow why we need this? The `TouchFile` above seems to already be addressing the needs of this test. And without passing any env information this wouldn't actually be writing any templated data. The InstallID is also irrelevant because you're using the append template, which doesn't use this.

Pretty sure you can drop this and your test would still run as expected.

Naatan (changes requested):

test/integration/shell_int_test.go
mitchell-as: Forgive me. I was under the impression "integration tests" were meant to test as much of the existing system in-place as possible, so breaking changes introduced in the subshell packages might be caught here.

mitchell-as:
I expect more churn.

test/integration/shell_int_test.go
Naatan: Ahhh that's the part I was missing. I was looking at my local zshrc_append file and wondering what you were looking to set up -.-

Naatan (changes requested):
Sorry; I was not fully up to speed on what the test was trying to do. I now understand you need to set up the bashrc file with the new modifications you have introduced in this PR. This would normally happen as part of the installation, which it makes sense you want to avoid as part of this test.

Now that I understand this; it does seem like calling `ConfigureAvailableShells()` is the most appropriate solution, as that matches what the installer does. I apologize as this was how you had done it at first; I sent us on a wild goose-chase and I'll own that. 

I think long term we should actually do an "install" for a test like this, but at the moment all our install/update testing is already so fragile that it doesn't make sense to add to that problem.

Naatan (approved):
Let's go with these changes. I forgot about the Setenv call; that is definitely something worth avoiding. Reading through the code it doesn't look too complicated to make subshell take a "home dir" argument in its constructor or something along those lines, but that is going to lead to refactorings which does feel like a bit of a scope creep. The story you filed sufficiently addresses the concern already, let's just address it there.

Thanks for bearing with me.

PR #2711, opened 2023-08-17
Executors should only setup PATH again in activated state.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2068" title="DX-2068" target="_blank">TaskDX-2068</a>  State tool "local" executors gobble up environment variables
  </td></table>
  <br />
 

The other environment variables have already been set.

The first attempt at fixing this assumed that `PATH` modifications by `state shell` were complete. However, they were not. Executors also modify `PATH` based on meta information, so that absolutely needs to be done every time an executor is run. Any other environment variables are already set properly, so they can be ignored.

### DIFF
--- a/cmd/state-exec/meta.go
+++ b/cmd/state-exec/meta.go
@@ -37,24 +37,21 @@ func newExecutorMeta(execPath string) (*executorMeta, error) {
 		return nil, fmt.Errorf("cannot get matching bin by path: %w", err)
 	}
 
-	transformEnv := true
+	onlyTransformPath := false
 	if projectDir := os.Getenv(constants.ActivatedStateEnvVarName); projectDir != "" {
 		// Note: cannot import runtime/target for ProjectDirToTargetDir() because that would bloat
 		// up the exe from <4MB to >15MB.
 		projectCacheDir := filepath.Join(storage.CachePath(), hash.ShortHash(projectDir))
-		if projectCacheDir == execDir {
-			transformEnv = false // runtime environment variables already set
+		if projectCacheDir == filepath.Dir(execDir) {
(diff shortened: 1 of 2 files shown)


daved (approved):

cmd/state-exec/meta.go
```diff
@@ -37,24 +37,21 @@ func newExecutorMeta(execPath string) (*executorMeta, error) {
...
 
-	transformEnv := true
+	onlyTransformPath := false
 	if projectDir := os.Getenv(constants.ActivatedStateEnvVarName); projectDir != "" {
 		// Note: cannot import runtime/target for ProjectDirToTargetDir() because that would bloat
 		// up the exe from <4MB to >15MB.
 		projectCacheDir := filepath.Join(storage.CachePath(), hash.ShortHash(projectDir))
-		if projectCacheDir == execDir {
-			transformEnv = false // runtime environment variables already set
+		if projectCacheDir == filepath.Dir(execDir) {
```
Naatan, line 45: I realize this wasn't introduced in this PR, but it just occurred to me that this isn't going to work if someone checked out with a custom path.

Additionally, we can't assume the two sides of the path were constructed the same way. We get around this in other places with this function:

https://github.com/ActiveState/cli/blob/d6bc46198b35e47cbe7958e5b935482f09d17008/internal/fileutils/fileutils.go#L1165-L1165

Naatan (changes requested):
Not actually requesting changes. I think we should close this PR and revert the work done here so far, because as you can see from my other comment this still has significant issues. I think we need to take this back to the drawing board first as it feels like we're playing whack a mole without a strategy (that's not on you - that's just the reality of our initial plan falling apart).

PR #2719, opened 2023-08-21
Support user-facing BuildPlanner errors

MDrakos:
While working on this PR, a better way to handle the various error types coming from the BuildPlanner surfaced. This PR updates our error handling in this regard.


pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -215,59 +189,93 @@ func (b *BuildPlanByProject) CommitID() (strfmt.UUID, error) {
...
+	case HeadOnBranchMovedErrorType:
+		return locale.NewInputError("err_buildplanner_head_on_branch_moved", "Head on branch has moved, recieved message: {{.V0}}", commit.Error.Message)
+	case NoChangeSinceLastCommitErrorType:
+		return locale.NewInputError("err_buildplanner_no_change_since_last_commit", "No change since last commit, recieved message: {{.V0}}", commit.Error.Message)
+	default:
+		return errs.New(fallbackMessage)
+	}
+}
+
+func ProcessBuildError(build *Build, fallbackMessage string) error {
```
Naatan, line 274: The other Process functions do a nil check first, this one doesn't. Is that intentional?

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -282,31 +290,29 @@ func (b *BuildPlanByCommit) Build() (*Build, error) {
...
-		if b.Commit.Error.Message != "" {
-			return "", errs.New("BuildPlanByCommit.CommitID: Could not get commit ID, API returned commit error message: %s", b.Commit.Message)
-		}
-		return "", errs.New("BuildPlanByCommit.CommitID: Could not retrieve commit")
+	if project.Type == NotFoundErrorType {
+		return locale.NewInputError("err_buildplanner_project_not_found", "Unable to find project, recieved message: {{.V0}}", project.Message)
 	}
 
-	return b.Commit.CommitID, nil
+	return errs.New(fallbackMessage)
```
Naatan, line 311: We don't want to silence this one (input error)?

Naatan (changes requested):

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -215,59 +189,93 @@ func (b *BuildPlanByProject) CommitID() (strfmt.UUID, error) {
...
+		errorType == NoChangeSinceLastCommitErrorType ||
+		errorType == HeadOnBranchMovedErrorType ||
+		errorType == ForbiddenErrorType ||
+		errorType == RemediableSolveErrorType ||
+		errorType == PlanningErrorType
+}
+
+func ProcessCommitError(commit *Commit, fallbackMessage string) error {
+	if commit.Error == nil {
+		return errs.New(fallbackMessage)
```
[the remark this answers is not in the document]
MDrakos: No, if this happens we've encountered an error type but the message wasn't set which would be an invalid according to the build planner's schema.

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: If we have to use the fallback message we've encountered an error type that we aren't handling so we should update these functions.

Naatan (approved):

PR #2724, opened 2023-08-22
Ensure build action is set

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2129" title="DX-2129" target="_blank">BugDX-2129</a>  runtime-debug: build action is not sent
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/buildplanner.go
+++ b/pkg/platform/model/buildplanner.go
@@ -17,7 +17,6 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
 	bpModel "github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
 	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/request"
-	"github.com/ActiveState/cli/pkg/platform/api/headchef"
 	"github.com/ActiveState/cli/pkg/platform/api/headchef/headchef_models"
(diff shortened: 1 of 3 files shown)


pkg/platform/model/buildplanner.go
```diff
@@ -53,7 +52,7 @@ type BuildResult struct {
 	CommitID            strfmt.UUID
 	Build               *bpModel.Build
 	BuildStatusResponse *headchef_models.V1BuildStatusResponse
-	BuildStatus         headchef.BuildStatusEnum
+	BuildStatus         string
 	BuildReady          bool
```
mitchell-as, line 56: I wonder if `BuildReady` is necessary now that we can just test that `BuildStatus == bpModel.Complete`.

mitchell-as, line 56: I wonder if `BuildReady` is necessary now that we can just test that `BuildStatus == bpModel.Complete`.

mitchell-as (approved):
Looks good to me. I did add a comment that may or may not be addressed.

pkg/platform/model/buildplanner.go
MDrakos: Yes, it likely can be replaced however I will leave that to be part of the upcoming runtime refactor.

PR #2725, opened 2023-08-22
Delete a remotely created project if runtime setup fails.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2093" title="DX-2093" target="_blank">BugDX-2093</a>  init project with `--language python@3.9` cause actual creation of the project in the platform
  </td></table>
  <br />

### DIFF
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -234,6 +234,11 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 
 	err = runbits.RefreshRuntime(r.auth, r.out, r.analytics, proj, commitID, true, target.TriggerInit, r.svcModel)
 	if err != nil {
+		logging.Debug("Deleting remotely created project due to runtime setup error")
+		err2 := model.DeleteProject(namespace.Owner, namespace.Project, r.auth)
+		if err2 != nil {
+			multilog.Error("Error deleting remotely created project after runtime setup error: %v", errs.JoinMessage(err2))
+		}
 		return locale.WrapError(err, "err_init_refresh", "Could not setup runtime after init")
 	}
 
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -238,6 +238,7 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 		err2 := model.DeleteProject(namespace.Owner, namespace.Project, r.auth)
 		if err2 != nil {
 			multilog.Error("Error deleting remotely created project after runtime setup error: %v", errs.JoinMessage(err2))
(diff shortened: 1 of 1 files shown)


mitchell-as:
@daved ping

daved:
@Naatan Should the deletion occur when there is any error after platform project creation?

internal/runners/initialize/init.go
```diff
@@ -234,6 +234,11 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 
 	err = runbits.RefreshRuntime(r.auth, r.out, r.analytics, proj, commitID, true, target.TriggerInit, r.svcModel)
 	if err != nil {
+		logging.Debug("Deleting remotely created project due to runtime setup error")
+		err2 := model.DeleteProject(namespace.Owner, namespace.Project, r.auth)
+		if err2 != nil {
+			multilog.Error("Error deleting remotely created project after runtime setup error: %v", errs.JoinMessage(err2))
```
Naatan, line 240: This should be returned to the user. It means they're in a broken state and they should be made aware of it.

Naatan:
> @Naatan Should the deletion occur when there is any error after platform project creation?

The AC says "when the runtime fails to source"; so no not any error, just runtime errors. Seems the code as is addressed the AC, although it should still communicate failure of deletion to the user.

daved (approved):

daved:
@Naatan The reason I asked is because I thought the underlying purpose was to ensure that the remote and local states were effectively equal. Deleting only during runtime handling seems to leave a window open for the disparity to occur.

Naatan:
@daved What other conditions do you think we should consider? The ACs are pretty clear. I'm not against expanding the scope if we find other cases, but this discussion feels very open-ended at the moment. Are you seeing a particular issue, or are you just putting out feelers?

daved:
@Naatan The errors that can happen after a project is first created on the platform are from:
- branch, err := model.DefaultBranchForProject(platformProject)
- err = model.UpdateProjectBranchCommitWithModel(platformProject, branch.Label, commitID)
- err = runbits.RefreshRuntime(r.auth, r.out, r.analytics, proj, commitID, true, target.TriggerInit, r.svcModel)
This PR will remove the created project from the platform for the last behavior only, but there are two steps prior that could leave the user in an undesirable workflow.

Naatan:
I see.. I don't think it's worth expanding the check to those necessarily. While those are API's that can of course give an error, the error condition would only happen in unexpected scenario's, in which case I'd rather not go deleting projects.

For the runtime scenario, we know what this error scenario means, so acting on it with intend is perfectly reasonable.

PR #2727, opened 2023-08-23
Detecting the current project should fall back on the activated one within a shell.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2109" title="DX-2109" target="_blank">BugDX-2109</a>  Despite activated project `state` commands still error `To run this command you need to be in an existing project.`
  </td></table>
  <br />
 

If the working directory does not contain a project, fall back on the activated project (if there is one).

### DIFF
--- a/internal/virtualenvironment/virtualenvironment.go
+++ b/internal/virtualenvironment/virtualenvironment.go
@@ -42,14 +42,12 @@ func (v *VirtualEnvironment) GetEnv(inherit bool, useExecutors bool, projectDir,
 	}
 
 	if projectDir != "" {
-		configFile := filepath.Join(projectDir, constants.ConfigFileName)
-
 		envMap[constants.ActivatedStateEnvVarName] = projectDir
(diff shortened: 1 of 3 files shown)


mitchell-as:
The original bug report is inaccurate: running `state pkg` within a shell for the default project (but outside that project) works just fine. It's only when you run `state pkg` within a shell without a default project (and outside that shell's project) do you get an error message. I've posted a comment in that ticket detailing as such.

I'm fine abandoning the ticket on those grounds and closing this PR if the side effect I noted in the PR description is a showstopper.

Naatan (changes requested):
> A side effect of this is that you cannot run the state tool for any project outside the current one when you are in an activated state/shell.

This is unfortunately a dealbreaker. We're trying to fix a bug, not make functional changes. And being able to use other projects while in an activated state is a use-case we want to address.

mitchell-as:
Okay, I've removed the side effect and updated the PR description accordingly.

PR #2728, opened 2023-08-23
Don't retry DNS errors

MDrakos:


internal/retryhttp/client.go
```diff
@@ -201,6 +201,11 @@ func retryPolicy(ctx context.Context, resp *http.Response, err error) (bool, err
 			}
 		}
 
+		var dnsError *net.DNSError
+		if errors.As(err, &dnsError) {
+			return false, locale.NewError("err_network", "Network error, please ensure you have a stable internet connection")
```
Naatan, line 206: https://github.com/ActiveState/cli/blob/9396c9d1d96d61ae25ac3c47adcfa2ac358a8f31/internal/locale/locales/en-us.yaml#L1753-L1753

Naatan:
https://stackoverflow.com/questions/76296690/http-client-faster-timeout-when-no-network

It doesn't seem like there are any good solutions to this problem. Let's bring the ticket back to triage so we can discuss.

PR #2735, opened 2023-08-25
Initial support for showing last runtime use date and time via `state projects`

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1962" title="DX-1962" target="_blank">StoryDX-1962</a>  As a user getting a overage message i would like to be able to see where the local runtimes are
  </td></table>
  <br />


cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -98,6 +105,39 @@ func (w *Watcher) check() {
...
+		for _, checkout := range checkouts {
+			proj, err := project.FromPath(checkout)
+			if err != nil {
+				multilog.Error("Unable to get project %s from checkout: %v", checkout, err)
+				continue
+			}
+			projectTarget := target.NewProjectTarget(proj, nil, "")
+			execDir := setup.ExecDir(projectTarget.Dir())
+			logging.Debug("Looking at project %s located at %s whose executables are in %s", namespace, checkout, execDir)
+			if !strings.HasPrefix(e.Exec, execDir) {
```
Naatan, line 124: We'll not to do something more sophisticated here to handle different representations of the same path. See https://github.com/ActiveState/cli/blob/79cb7134123e7836b1c9fd52111dd274aea956e4/internal/fileutils/fileutils.go#L1157-L1157 for example.

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -98,6 +105,39 @@ func (w *Watcher) check() {
...
+	localProjects := projectfile.GetProjectMapping(w.cfg)
+	for namespace, checkouts := range localProjects {
+		for _, checkout := range checkouts {
+			proj, err := project.FromPath(checkout)
+			if err != nil {
+				multilog.Error("Unable to get project %s from checkout: %v", checkout, err)
+				continue
+			}
+			projectTarget := target.NewProjectTarget(proj, nil, "")
+			execDir := setup.ExecDir(projectTarget.Dir())
```
Naatan, line 122:
```suggestion
			execDir := setup.ExecDir(proj.Dir())
```

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -98,6 +105,39 @@ func (w *Watcher) check() {
...
+			logging.Debug("Looking at project %s located at %s whose executables are in %s", namespace, checkout, execDir)
+			if !strings.HasPrefix(e.Exec, execDir) {
+				logging.Debug("Executable is not in this runtime")
+				continue
+			}
+			logging.Debug("Executable is in this runtime")
+			err = w.cfg.GetThenSet(
```
Naatan, line 129: This error isn't handled.

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -98,6 +105,39 @@ func (w *Watcher) check() {
...
+				LastUsedCfgKey,
+				func(v interface{}) (interface{}, error) {
+					lastUsed := cast.ToStringMap(v)
+					lastUsed[execDir] = time.Now().Format(time.RFC3339)
+					logging.Debug("New 'last used' time for %s is %v", execDir, lastUsed[execDir])
```
Naatan, line 134:
```suggestion
```

This could end up spamming our log files.

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -98,6 +105,39 @@ func (w *Watcher) check() {
...
+}
+
+func (w *Watcher) UpdateLastUsed(e entry) {
+	logging.Debug("Updating last usage of project that contains %s", e.Exec)
+	localProjects := projectfile.GetProjectMapping(w.cfg)
+	for namespace, checkouts := range localProjects {
+		for _, checkout := range checkouts {
+			proj, err := project.FromPath(checkout)
+			if err != nil {
+				multilog.Error("Unable to get project %s from checkout: %v", checkout, err)
```
Naatan, line 118:
```suggestion
```

This would almost certainly have to be some form of permission error that we can't help with. `GetProjectMapping()` cleans up the projects before returning them, so you should only get valid ones to begin with.

I'm concerned since this is happening in a loop that runs periodically that it could end up spamming our rollbar. Which shouldn't be an issue but sadly it is due to rollbar quotas.

internal/runners/projects/projects.go
```diff
@@ -21,9 +26,12 @@ type projectWithOrg struct {
 	Organization   string   `json:"organization"`
 	LocalCheckouts []string `json:"local_checkouts,omitempty"`
 	Executables    []string `json:"executables,omitempty"`
+	LastUsed       []string `json:"last_used,omitempty"`
 }
 
-func newProjectWithOrg(name, org string, checkouts []string) projectWithOrg {
+const runtimeInUseCutoff = 90 * time.Second // heartbeats are usually sent once per minute
```
Naatan, line 32: Please have this share a central value with the watcher, ie.

https://github.com/ActiveState/cli/blob/79cb7134123e7836b1c9fd52111dd274aea956e4/cmd/state-svc/internal/rtwatcher/watcher.go#L21-L21

We can probably just move that to constants.

internal/graph/generated.go
```diff
@@ -20,6 +20,10 @@ type AvailableUpdate struct {
 	Sha256   string `json:"sha256"`
 }
 
+type CheckRuntimeLastUsedResponse struct {
+	Times map[string]interface{} `json:"times"`
```
Naatan, line 24: Please have this either return a string value map, or ideally even a datetime if graphql supports it.

If this is too challenging to handle on the protocol level we could dial it up to the model. But we shouldn't be passing interface{} values around if we can help it.

internal/runners/projects/projects.go
```diff
@@ -34,6 +42,19 @@ func newProjectWithOrg(name, org string, checkouts []string) projectWithOrg {
 			multilog.Error("Unable to get project %s from checkout: %v", checkout, err)
 		}
 		p.Executables = append(p.Executables, execDir)
+		lastUsed := locale.Tl("projects_last_use_unknown", "unknown")
+		if v, exists := lastUsedTimes[execDir]; exists {
```
Naatan, line 46: This'll need more sophisticated path matching as well.

internal/runners/projects/projects.go
```diff
@@ -34,6 +42,19 @@ func newProjectWithOrg(name, org string, checkouts []string) projectWithOrg {
...
+			if t, err := time.Parse(time.RFC3339, v.(string)); err == nil {
+				if time.Since(t) <= runtimeInUseCutoff && os.Getenv(constants.RuntimeInUseNoCutoffTimeEnvVarName) == "" {
```
Naatan, line 48: This logic should live in the svc imo. Maybe have it respond with two fields: last used time, and whether it is currently in use.

internal/runners/projects/projects.go
```diff
@@ -52,31 +73,36 @@ func (o *projectsOutput) MarshalOutput(f output.Format) interface{} {
...
 		executables := []string{}
+		lastUseds := []string{}
 		for i, checkout := range v.LocalCheckouts {
 			if len(v.Executables) <= i {
 				continue // remote project listings do not have executables
 			}
 			execDir := v.Executables[i]
-			if execDir != "" {
+			lastUsed := v.LastUsed[i]
+			if execDir != "" || lastUsed != "" {
```
Naatan, line 90: It seems like execDir being empty and lastUsed being non-empty is an impossible scenario. Is this supposed to be `&&`? If so we're still missing the `execDir!='' && lastUsed==''` scenario.

Naatan (changes requested):

cmd/state-svc/internal/rtwatcher/watcher.go
mitchell-as: Nope, that will return the activestate.yaml directory with "exec" tacked on (I tried). I need the runtime exec directory.

internal/runners/projects/projects.go
mitchell-as: That's a good idea. It cleans up the graphql too.

internal/graph/generated.go
mitchell-as: graphql's "Map" scalar is a `map[string]interface{}` so there was no easy way around it from what I could see. Anyway, I've changed the schema for the better.

internal/runners/projects/projects.go
mitchell-as: You're right that the conditional is now superfluous.

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -98,6 +106,53 @@ func (w *Watcher) check() {
...
+func (w *Watcher) UpdateLastUsed(e entry) {
+	if stateExec, err := installation.StateExec(); err == nil {
+		if isStateExec, err := fileutils.PathsEqual(e.Exec, stateExec); err == nil && isStateExec {
+			return // cannot infer which runtime is being used
+		} else if err != nil {
+			multilog.Error("Could not determine if %s is the State Tool executable: %s", e.Exec, errs.JoinMessage(err))
+		}
+	} else {
+		multilog.Error("Could not determine the state tool executable: %s", errs.JoinMessage(err))
+	}
```
mitchell-as, line 121: This still needs work to try and infer runtime from e.g. `state exec`.

mitchell-as: This might be tricky because all I have to work with is e.Dims, which has a namespace, but there might be multiple checkouts for that namespace.

mitchell-as:
Putting this up for review again with this caveat: https://github.com/ActiveState/cli/pull/2735/files#r1306199010. Let me know if you have any ideas @Naatan. Otherwise I'll give it a go on Monday.

Naatan (changes requested):
I think we need to take this back to the drawing board. We have the information we need, it's just not being sent in a way that we can use it, and I don't want to implement a flawed solution. There's no pressure on us to get this done so let's take whatever time we need to design this solution properly. Maybe we do a tech design brief first.

It feels to me that we need to send additional meta information along with the heartbeats. Eg. the local path of the project.

Let's bring this story back to triage so we can discuss how best to address this.

PR #2739, opened 2023-08-30
Temporary PR for visualizing branch changes

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1954" title="DX-1954" target="_blank">TaskDX-1954</a>  Improve wait up/down and enrich content of svc wait up/down error message
  </td></table>
  <br />

PR #2740, opened 2023-08-30
Enrich service wait up/down errors with more debug info.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1954" title="DX-1954" target="_blank">TaskDX-1954</a>  Improve wait up/down and enrich content of svc wait up/down error message
  </td></table>
  <br />
 

This should help with rollbar reports.

This is based on work done in the https://github.com/ActiveState/cli/commits/green/upg_svc_wait.DX-1824 branch: https://github.com/ActiveState/cli/pull/2739/files

I omitted the wait up/down pacing (https://github.com/ActiveState/cli/pull/2739/commits/c1920bb1d130ef07c8eab04774ca68a7f33ad2bf) because that's additional complication for little gain from what little I understood of it. Maintenance going forward may be a burden.

### DIFF
--- a/cmd/state-svc/autostart/autostart.go
+++ b/cmd/state-svc/autostart/autostart.go
@@ -13,7 +13,7 @@ import (
 var Options = autostart.Options{
 	Name:           constants.SvcAppName,
 	LaunchFileName: constants.SvcLaunchFileName,
-	Args:           []string{"start", "--startup"},
+	Args:           []string{"start", "--autostart"},
 }
 
 func RegisterConfigListener(cfg *config.Instance) error {
--- a/cmd/state-svc/autostart/autostart.go
+++ b/cmd/state-svc/autostart/autostart.go
@@ -13,7 +13,7 @@ import (
 var Options = autostart.Options{
 	Name:           constants.SvcAppName,
 	LaunchFileName: constants.SvcLaunchFileName,
-	Args:           []string{"start"},
+	Args:           []string{"start", "--startup"},
 }
 
 func RegisterConfigListener(cfg *config.Instance) error {
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -138,7 +138,7 @@ func run(cfg *config.Instance) error {
 				logging.Debug("Running CmdStart")
 				argText := "svc-start:cli"
 				if autostart {
-					argText += "-autostart"
+					argText = "svc-start:auto"
 				}
 				return runStart(out, argText)
 			},
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -122,16 +122,25 @@ func run(cfg *config.Instance) error {
 	)
 
 	var foregroundArgText string
+	var startup bool
 
 	cmd.AddChildren(
 		captain.NewCommand(
 			cmdStart,
 			"",
 			"Start the ActiveState Service (Background)",
-			p, nil, nil,
+			p,
+			[]*captain.Flag{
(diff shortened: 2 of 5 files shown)


daved (commented):
@Naatan For this story, I agree that we should step back from the wait up/down pacing (new story filed here: https://activestatef.atlassian.net/browse/DX-2176), but I'm curious to bring up the third point in the AC:
> Consider and possibly include having some flag or arg passed in to the service start call at system startup so that we can better understand the context of the reported errors.

I feel fairly strongly that this would be an exceedingly useful data point. However, we didn't discuss a design. If you agree, should we create a separate story or update the current one?

Naatan:
Could you guys please add your concerns with the ACs to the jira ticket and move it back to triage.

We should not be omitting AC's without conversation with either myself or Pete.

I have opinions, but need more context. Probably easier if we talk about this during stand.

cmd/state-svc/main.go
```diff
@@ -122,16 +122,25 @@ func run(cfg *config.Instance) error {
...
 
 	cmd.AddChildren(
 		captain.NewCommand(
 			cmdStart,
 			"",
 			"Start the ActiveState Service (Background)",
-			p, nil, nil,
+			p,
+			[]*captain.Flag{
+				{Name: "startup", Value: &startup}, // differentiate between autostart and cli invocation
```
daved, line 134: I'm not sure if we have curated the help output for the service, but I suggest setting Hidden to true.

cmd/state-svc/main.go
```diff
@@ -122,16 +122,25 @@ func run(cfg *config.Instance) error {
...
+			},
+			nil,
 			func(ccmd *captain.Command, args []string) error {
 				logging.Debug("Running CmdStart")
-				return runStart(out, "svc-start:cli")
+				argText := "svc-start:cli"
+				if startup {
+					argText = "svc-start:startup"
```
daved, line 141:
```suggestion
					argText += "-atstart"
```

daved (changes requested):

cmd/state-svc/autostart/autostart.go
```diff
@@ -13,7 +13,7 @@ import (
 var Options = autostart.Options{
 	Name:           constants.SvcAppName,
 	LaunchFileName: constants.SvcLaunchFileName,
-	Args:           []string{"start"},
+	Args:           []string{"start", "--startup"},
```
daved, line 16:
```suggestion
	Args:           []string{"start", "--autostart"},
```

cmd/state-svc/main.go
daved: I suggest renaming "startup" to autostart.

daved (changes requested):

daved (approved):

cmd/state-svc/main.go
```diff
@@ -122,16 +122,25 @@ func run(cfg *config.Instance) error {
...
+			[]*captain.Flag{
+				{Name: "autostart", Value: &autostart, Hidden: true}, // differentiate between autostart and cli invocation
+			},
+			nil,
 			func(ccmd *captain.Command, args []string) error {
 				logging.Debug("Running CmdStart")
-				return runStart(out, "svc-start:cli")
+				argText := "svc-start:cli"
+				if autostart {
+					argText += "-autostart"
```
Naatan, line 141:
```suggestion
					argText += "svc-start:auto"
```

Naatan (changes requested):

Naatan (approved):

PR #2741, opened 2023-09-01
User-facing errors in example runner

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2151" title="DX-2151" target="_blank">StoryDX-2151</a>  User facing errors for example runner
  </td></table>
  <br />

### DIFF
--- a/internal/errs/errs.go
+++ b/internal/errs/errs.go
@@ -55,6 +55,14 @@ func (e *userFacingError) UserError() string {
 	return e.message
 }
 
+func (e *userFacingError) AddTips(tips ...string) {
+	e.tips = append(e.tips, tips...)
+}
+
+func (e *userFacingError) ErrorTips() []string {
+	return e.tips
+}
+
 // PackedErrors represents a collection of errors that aren't necessarily related to each other
 // note that rtutils replicates this functionality to avoid import cycles
 type PackedErrors struct {
@@ -137,14 +145,6 @@ func WrapUserFacingError(wrapTarget error, message string, args ...interface{}) 
 		fmt.Sprintf(message, args...),
 		nil,
 	}
-}
-
-func (e *userFacingError) AddTips(tips ...string) {
-	e.tips = append(e.tips, tips...)
-}
-
-func (e *userFacingError) ErrorTips() []string {
-	return e.tips
 }
 
 // Pack creates a new error that packs the given errors together, allowing for multiple errors to be returned
--- a/internal/errs/errs.go
+++ b/internal/errs/errs.go
@@ -36,33 +36,6 @@ type TransientError interface {
 	IsTransient()
 }
 
-type UserFacingError interface {
-	error
-	UserError() string
-}
-
-type userFacingError struct {
-	wrapped error
-	message string
-	tips    []string
-}
-
-func (e *userFacingError) Error() string {
-	return JoinMessage(Wrap(e.wrapped, e.message))
-}
-
-func (e *userFacingError) UserError() string {
-	return e.message
-}
-
-func (e *userFacingError) AddTips(tips ...string) {
-	e.tips = append(e.tips, tips...)
-}
-
-func (e *userFacingError) ErrorTips() []string {
-	return e.tips
-}
-
 // PackedErrors represents a collection of errors that aren't necessarily related to each other
 // note that rtutils replicates this functionality to avoid import cycles
 type PackedErrors struct {
@@ -129,22 +102,6 @@ func New(message string, args ...interface{}) *WrapperError {
 func Wrap(wrapTarget error, message string, args ...interface{}) *WrapperError {
 	msg := fmt.Sprintf(message, args...)
 	return newError(msg, wrapTarget)
-}
-
-func NewUserFacingError(message string, args ...interface{}) *userFacingError {
-	return &userFacingError{
-		nil,
-		fmt.Sprintf(message, args...),
-		nil,
-	}
-}
-
-func WrapUserFacingError(wrapTarget error, message string, args ...interface{}) *userFacingError {
-	return &userFacingError{
-		wrapTarget,
-		fmt.Sprintf(message, args...),
-		nil,
-	}
 }
 
 // Pack creates a new error that packs the given errors together, allowing for multiple errors to be returned
@@ -300,15 +257,4 @@ func Unpack(err error) []error {
 		}
 	}
 	return result
-}
-
-// IsUserFacing identifies whether this error was curated for end-users and
-// returns that error. This is NOT the same as an error that is localized.
-func IsUserFacing(err error) (error, bool) {
-	var userFacingError UserFacingError
-	if errors.As(err, &userFacingError) {
-		return userFacingError, true
-	}
-
-	return err, false
 }
--- a/internal/errs/errs.go
+++ b/internal/errs/errs.go
@@ -36,8 +36,17 @@ type TransientError interface {
 	IsTransient()
 }
 
-type UserFacingError interface {
-	UserFacingError() error
+type UserFacingError struct {
+	wrapped error
+	message string
+}
+
+func (e UserFacingError) Error() string {
+	return JoinMessage(Wrap(e.wrapped, e.message))
+}
+
+func (e UserFacingError) UserError() string {
+	return e.message
 }
 
 // PackedErrors represents a collection of errors that aren't necessarily related to each other
@@ -106,6 +115,13 @@ func New(message string, args ...interface{}) *WrapperError {
 func Wrap(wrapTarget error, message string, args ...interface{}) *WrapperError {
 	msg := fmt.Sprintf(message, args...)
 	return newError(msg, wrapTarget)
+}
+
+func WrapUserFacingError(wrapTarget error, message string, args ...interface{}) *UserFacingError {
+	return &UserFacingError{
+		wrapTarget,
+		fmt.Sprintf(message, args...),
+	}
 }
 
 // Pack creates a new error that packs the given errors together, allowing for multiple errors to be returned
@@ -263,12 +279,12 @@ func Unpack(err error) []error {
 	return result
 }
 
-func UserFacing(err error) (error, bool) {
-	errs := Unpack(err)
-	for _, err := range errs {
-		if uerr, ok := err.(UserFacingError); ok {
-			return uerr.UserFacingError(), true
-		}
+// IsUserFacing identifies whether this error was curated for end-users and
+// returns that error. This is NOT the same as an error that is localized.
+func IsUserFacing(err error) (error, bool) {
+	var userFacingError *UserFacingError
+	if errors.As(err, &userFacingError) {
+		return userFacingError, true
 	}
 
 	return err, false
--- a/internal/errs/errs.go
+++ b/internal/errs/errs.go
@@ -36,6 +36,10 @@ type TransientError interface {
 	IsTransient()
 }
 
+type UserFacingError interface {
+	UserFacingError() error
+}
+
 // PackedErrors represents a collection of errors that aren't necessarily related to each other
 // note that rtutils replicates this functionality to avoid import cycles
 type PackedErrors struct {
@@ -257,4 +261,15 @@ func Unpack(err error) []error {
 		}
 	}
 	return result
+}
+
+func UserFacing(err error) (error, bool) {
+	errs := Unpack(err)
+	for _, err := range errs {
+		if uerr, ok := err.(UserFacingError); ok {
+			return uerr.UserFacingError(), true
+		}
+	}
+
+	return err, false
 }
--- a/internal/errs/errs.go
+++ b/internal/errs/errs.go
@@ -44,6 +44,7 @@ type UserFacingError interface {
 type userFacingError struct {
 	wrapped error
 	message string
+	tips    []string
 }
 
 func (e *userFacingError) Error() string {
@@ -122,11 +123,28 @@ func Wrap(wrapTarget error, message string, args ...interface{}) *WrapperError {
 	return newError(msg, wrapTarget)
 }
 
+func NewUserFacingError(message string, args ...interface{}) *userFacingError {
+	return &userFacingError{
+		nil,
+		fmt.Sprintf(message, args...),
+		nil,
+	}
+}
+
 func WrapUserFacingError(wrapTarget error, message string, args ...interface{}) *userFacingError {
 	return &userFacingError{
 		wrapTarget,
 		fmt.Sprintf(message, args...),
+		nil,
 	}
+}
+
+func (e *userFacingError) AddTips(tips ...string) {
+	e.tips = append(e.tips, tips...)
+}
+
+func (e *userFacingError) ErrorTips() []string {
+	return e.tips
 }
 
 // Pack creates a new error that packs the given errors together, allowing for multiple errors to be returned
--- a/internal/errs/errs.go
+++ b/internal/errs/errs.go
@@ -36,16 +36,21 @@ type TransientError interface {
 	IsTransient()
 }
 
-type UserFacingError struct {
+type UserFacingError interface {
+	error
+	UserError() string
+}
+
+type userFacingError struct {
 	wrapped error
 	message string
 }
 
-func (e UserFacingError) Error() string {
+func (e *userFacingError) Error() string {
 	return JoinMessage(Wrap(e.wrapped, e.message))
 }
 
-func (e UserFacingError) UserError() string {
+func (e *userFacingError) UserError() string {
 	return e.message
 }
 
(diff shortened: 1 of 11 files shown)


pkg/cmdlets/errors/errors.go
```diff
@@ -93,8 +93,6 @@ func ParseUserFacing(err error) (int, error) {
 		return 0, nil
 	}
 
-	_, hasMarshaller := err.(output.Marshaller)
```
Naatan, line 96: This change is breaking the original intend of the function. Just way predated the new design and isn't looking to address the same use-case. What you effectively found here is that the term "User Facing" was already in use, and we need to either resolve the conflict of overlapping terms, or ensure that they are compatible.

Looking through the code it seems `hasMarshaller` will mainly identify whether we already have an OutputError on our hands, so we don't redundantly wrap it. While I don't see any scenario's in which this could be true, do we want to risk breaking this behaviour for this PR?

Please take some time to understand what OutputError{} is doing and how the new concept of user facing errors plays into this.

Naatan (changes requested):

internal/runners/hello/hello_example.go
```diff
@@ -58,8 +66,40 @@ func New(p primeable) *Hello {
...
+				// If this is not a specific error we are looking for, do nothing.
+				continue
+			}
+		}
+
+		return locale.WrapError(err, "hello_cannot_say", "Cannot say hello.")
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
MDrakos: I prefer the approach that is currently in the PR. While having a function that does post-processing + `defer` fits nicely in a mental model in practice I find that when reading the code it requires more jumping around, or at least remembering that you have to jump to the defer function after reading the runner logic. This isn't a problem with the current code as you read the `Run` function, then the function that it calls, and return to the caller.

Post-processing with a `defer` also requires a named return value which is different from the _vast_ majority of the functions in our codebase and also adds another layer of complexity when reading the code.

This is all subjective, as is your suggestion, and in the end, it's your call.

pkg/cmdlets/errors/errors.go
MDrakos: From what I can grok the `OutputError` type adds extra post-processing and the handling in this function ensures that errors returned all have a similar UX. I've wrapped the user-facing error in this `OutputError` type as well to ensure it's consistent with all of the other errors we are presenting to users.

I've also added back the `hasMarshaller` check as you were right, we don't want to wrap the error again in the `OutputError` type.

internal/runners/hello/hello_example.go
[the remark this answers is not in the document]
Naatan: The pattern of the named return is not uncommon. It is used frequently in our codebase.

I'm open to other solutions, but between the current implementation and what I am suggesting I'd prefer my suggestion. Giving each runner a redundant `Run()` function that you have to interpret to understand does not feel like the better solution to me, but you're right, that is subjective. But, given this is meant as a reference for our codebase rather than a one-off piece of code I'm gonna be nitpicky and insist on either my solution or an alternate yet to be determined one.

internal/errs/errs.go
```diff
@@ -36,6 +36,19 @@ type TransientError interface {
 	IsTransient()
 }
 
+type UserFacingError struct {
+	wrapped error
+	message string
+}
+
+func (e UserFacingError) Error() string {
+	return JoinMessage(Wrap(e.wrapped, e.message))
```
Naatan, line 45: Joining errors is the responsibility of the code that interprets them.

I suggest instead just using "User Facing Error: <message>" so it's clear from reading the standard error stack what we're dealing with.

internal/errs/errs.go
```diff
@@ -36,6 +36,19 @@ type TransientError interface {
 	IsTransient()
 }
 
+type UserFacingError struct {
+	wrapped error
+	message string
+}
+
+func (e UserFacingError) Error() string {
```
Naatan, line 44:
```suggestion
func (e *UserFacingError) Error() string {
```

internal/errs/errs.go
```diff
@@ -36,6 +36,19 @@ type TransientError interface {
...
+	return JoinMessage(Wrap(e.wrapped, e.message))
+}
+
+func (e UserFacingError) UserError() string {
```
Naatan, line 48:
```suggestion
func (e *UserFacingError) UserError() string {
```

internal/errs/errs.go
```diff
@@ -258,3 +278,14 @@ func Unpack(err error) []error {
 	}
 	return result
 }
+
+// IsUserFacing identifies whether this error was curated for end-users and
+// returns that error. This is NOT the same as an error that is localized.
+func IsUserFacing(err error) (error, bool) {
+	var userFacingError *UserFacingError
```
Naatan, line 285: Please use an interface instead. We don't want to require the UserFacingError type itself, the type is just for convenience. It should be possible for us to construct our own user facing error type.

internal/runbits/hello_example.go
```diff
@@ -5,10 +5,14 @@ import (
...
 
+type NoNameProvidedError struct {
+	*locale.LocalizedError
+}
+
 func SayHello(out output.Outputer, name string) error {
 	if name == "" {
 		// Errors that are due to USER input should use `NewInputError` or `WrapInputError`
-		return locale.NewInputError("hello_err_no_name", "No name provided.")
+		return &NoNameProvidedError{locale.NewInputError("hello_err_no_name", "No name provided.")}
```
Naatan, line 15: It just occurred to me that we should not be localizing any errors in the example runner or example runbits (except for the user facing error itself).

Because our ideal is that once all runners speak "User facing error", we no longer have localized errors beyond the user facing ones.

internal/runners/hello/hello_example.go
```diff
@@ -24,6 +24,16 @@ type primeable interface {
...
 
+// HelloUserFacingError is an error from this runner that is intended to be
+// shown to the user. It is a wrapper around a localized error.
+type HelloUserFacingError struct {
+	*locale.LocalizedError
+}
+
+func (e *HelloUserFacingError) UserFacingError() error {
+	return e
+}
```
Naatan, line 35: Shouldn't need this anymore.

pkg/cmdlets/errors/errors.go
```diff
@@ -98,11 +98,21 @@ func ParseUserFacing(err error) (int, error) {
 	// unwrap exit code before we remove un-localized wrapped errors from err variable
 	code := errs.ParseExitCode(err)
 
+	// If there is a user facing error in the error stack we want to ensure
+	// that is it forwarded to the user.
+	uerr, ok := errs.IsUserFacing(err)
+	if ok {
+		logging.Debug("Returning user facing error, error stack: \n%s", errs.JoinMessage(err))
+		return code, &OutputError{uerr}
```
Naatan, line 106: I don't think OutputError produces the error string from `UserFacingError()`? 

Please make sure you run this code to ensure it produces the error message you expect.

Naatan (changes requested):

pkg/cmdlets/errors/errors.go
MDrakos: It does here: https://github.com/ActiveState/cli/blob/master/pkg/cmdlets/errors/errors.go. I've confirmed this from running the code.

internal/errs/errs.go
MDrakos: Is it possible to make this a code suggestion? I don't understand what you're asking for.

internal/runbits/hello_example.go
MDrakos: I've made the change in this PR though this has required me to introduce a centralized errors file. We will likely need to iterate on this or make the change in a different story/PR.

pkg/cmdlets/errors/errors.go
Naatan: I don't understand, you linked to code on master which definitely would not be handling this new concept. Wrong link?

internal/errs/errs.go
```diff
@@ -258,3 +301,14 @@ func Unpack(err error) []error {
...
+
+// IsUserFacing identifies whether this error was curated for end-users and
+// returns that error. This is NOT the same as an error that is localized.
+func IsUserFacing(err error) (error, bool) {
+	var userFacingError UserFacingError
+	if errors.As(err, &userFacingError) {
+		return userFacingError, true
+	}
+
+	return err, false
```
Naatan, line 313:
```suggestion
func IsUserFacing(err error) bool {
	var userFacingError UserFacingError
	if errors.As(err, &userFacingError) {
		return true
	}

	return false
```

If you want to cast the error type you should just use `errors.As()` directly.

internal/runners/hello/hello_example.go
```diff
@@ -58,29 +56,48 @@ func New(p primeable) *Hello {
 	}
 }
 
+// processError contains the scope in which errors are processed. This is
+// useful for ensuring that errors are wrapped in a user-facing error and
+// localized.
+func processError(err *error) {
+	if err != nil {
```
Naatan, line 63: Just return early `if err == nil`, so this whole block doesn't need to be indented.

internal/errs/errs.go
```diff
@@ -104,6 +131,22 @@ func Wrap(wrapTarget error, message string, args ...interface{}) *WrapperError {
...
 
+func NewUserFacingError(message string, args ...interface{}) *userFacingError {
+	return &userFacingError{
+		nil,
+		fmt.Sprintf(message, args...),
+		nil,
+	}
+}
+
+func WrapUserFacingError(wrapTarget error, message string, args ...interface{}) *userFacingError {
```
Naatan, line 142: Seeing as user facing errors should always be localized we should just take a single message arg, no further args. Because splicing the message together is the job of the localization lib.

internal/runners/hello/hello_example.go
```diff
@@ -58,29 +56,48 @@ func New(p primeable) *Hello {
...
+			// the unpacked error that we are inspecting.
+			*err = errs.WrapUserFacingError(*err, locale.Tl("hello_err_no_name", "Cannot say hello because no name was provided."))
+		case errs.Matches(*err, &errs.ErrNoProject{}):
+			werr := errs.WrapUserFacingError(*err, locale.Tl("hello_err_no_project", "Cannot say hello because you are not in a project directory."))
+
+			// It's useful to offer users reasonable tips on recourses.
+			*err = errs.AddTips(werr, locale.Tl(
+				"hello_suggest_checkout",
+				"Try using [ACTIONABLE]`state checkout`[/RESET] first.",
+			))
```
Naatan, line 77: This feels like it would be a common thing we'd want to do. How about we make the last arg for `WrapUserFacingError()` take tips?

internal/runners/hello/hello_example.go
```diff
@@ -58,29 +56,48 @@ func New(p primeable) *Hello {
...
+		}
+
+		*err = errs.Wrap(*err, "Cannot say hello")
```
Naatan, line 80: Doesn't seem worth wrapping this case.

Naatan (changes requested):
Suggest we try to do some pair programming tomorrow to resolve this PR faster.

internal/errs/errs.go
MDrakos: Fair enough, though for our current use case, this change makes the function pointless. I'm just going to remove it and use `errors.As()` where we were calling it previously

internal/runners/hello/hello_example.go
```diff
@@ -58,29 +57,47 @@ func New(p primeable) *Hello {
...
+		// Errors that we are looking for should be wrapped in a user-facing error.
+		// Ensure we wrap the top-level error returned from the runner and not
+		// the unpacked error that we are inspecting.
+		*err = errs.WrapUserFacingError(*err, locale.Tl("hello_err_no_name", "Cannot say hello because no name was provided."))
+	case errs.Matches(*err, &errors.ErrNoProject{}):
+		// It's useful to offer users reasonable tips on recourses.
+		*err = errs.WrapUserFacingError(
+			*err,
+			locale.Tl("hello_err_no_project", "Cannot say hello because you are not in a project directory."),
+			locale.Tl("hello_suggest_checkout", "Try using [ACTIONABLE]`state checkout`[/RESET] first."),
```
Naatan, line 79:
```suggestion
			errs.WithTips(
			locale.Tl("hello_suggest_checkout", "Try using [ACTIONABLE]`state checkout`[/RESET] first."),
			)
```

pkg/cmdlets/errors/errors.go
```diff
@@ -98,11 +98,21 @@ func ParseUserFacing(err error) (int, error) {
 	// unwrap exit code before we remove un-localized wrapped errors from err variable
 	code := errs.ParseExitCode(err)
 
+	// If there is a user facing error in the error stack we want to ensure
+	// that is it forwarded to the user.
+	var userFacingError errs.UserFacingError
+	if errors.As(err, &userFacingError) {
+		logging.Debug("Returning user facing error, error stack: \n%s", errs.JoinMessage(err))
+		return code, &OutputError{userFacingError}
+	}
```
Naatan, line 107: Should be after the IsSilent line below.

Naatan (changes requested):

internal/errs/errs.go
```diff
@@ -36,6 +36,33 @@ type TransientError interface {
...
+}
+
+type userFacingError struct {
+	wrapped error
+	message string
+	tips    []string
+}
+
+func (e *userFacingError) Error() string {
+	return JoinMessage(Wrap(e.wrapped, e.message))
```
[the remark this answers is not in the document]
MDrakos: This leads to what feels like an awkward UX with the "User Facing Error: " string before the actual message. Do we want to keep this or remove it?

```
❯ ~/work/cli/build/state _hello
█ Saying hello (Beta)

This command is for example use only
█ Something Went Wrong

 x User Facing Error: Cannot say hello because you are not in a project directory

█ Need More Help?

 • Try using `state checkout` first
```

pkg/cmdlets/errors/errors.go
MDrakos: Whoops, yes wrong link.

I've run the code and confirmed the output from `UserError()` is there. I believe it's done via the output mediator here: https://github.com/ActiveState/cli/blob/DX-2151/internal/output/mediator.go#L54

internal/errs/userfacing.go
```diff
@@ -1,10 +1,14 @@
 package errs
 
+import "errors"
+
 type UserFacingError interface {
 	error
 	UserError() string
 }
 
+type ErrOpts func(err *userFacingError) *userFacingError
```
Naatan, line 10:
```suggestion
type ErrOpt func(err *userFacingError) *userFacingError
```

The type is only covering a single option.

pkg/cmdlets/errors/errors.go
```diff
@@ -103,6 +103,16 @@ func ParseUserFacing(err error) (int, error) {
 		return code, nil
 	}
 
+	// If there is a user facing error in the error stack we want to ensure
+	// that is it forwarded to the user.
+	var userFacingError errs.UserFacingError
+	if errors.As(err, &userFacingError) {
+		logging.Debug("Returning user facing error, error stack: \n%s", errs.JoinMessage(err))
+		return code, &OutputError{userFacingError}
```
Naatan, line 111: OutputError doesn't know to look for UserError().

Naatan (changes requested):

pkg/cmdlets/errors/errors.go
MDrakos: Ahh okay, now I understand why I was seeing the user-facing error output as `OutputError` will eventually call the `Error()` method, but we want it to call `UserError()`. Thanks for catching that!

pkg/cmdlets/errors/errors.go
```diff
@@ -39,18 +39,23 @@ func (o *OutputError) MarshalOutput(f output.Format) interface{} {
...
-	}
-	for _, errv := range rerrs {
-		message := trimError(locale.ErrorMessage(errv))
-		if f == output.PlainFormatName {
-			outLines = append(outLines, fmt.Sprintf(" [NOTICE][ERROR]x[/RESET] %s", message))
-		} else {
-			outLines = append(outLines, message)
+	var userFacingError errs.UserFacingError
+	if errors.As(o.error, &userFacingError) {
+		outLines = append(outLines, fmt.Sprintf(" [NOTICE][ERROR]x[/RESET] %s", userFacingError.UserError()))
```
Naatan, line 44: You'll want the same conditional here as on line 54.

Naatan (changes requested):

Naatan (approved):

internal/runbits/hello_example.go
```diff
@@ -1,14 +1,19 @@
...
 	"github.com/ActiveState/cli/internal/output"
 )
 
+type NoNameProvidedError struct {
+	*errs.WrapperError
+}
+
 func SayHello(out output.Outputer, name string) error {
 	if name == "" {
 		// Errors that are due to USER input should use `NewInputError` or `WrapInputError`
```
mitchell-as, line 15: This comment is now incorrect and causing confusion.

mitchell-as, line 15: This comment is now incorrect and causing confusion @Naatan @MDrakos

PR #2742, opened 2023-09-06
Add support for attachStagedCommit

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2133" title="DX-2133" target="_blank">StoryDX-2133</a>  Use new AttachStagedCommit BuildPlanner mutation for `state push`
  </td></table>
  <br />
 

The BuildPlanner currently doesn't support attaching a commit that doesn't have a parent. This means that in some cases we are still using the old VCS API. I've filed a story with PB to address this.

### DIFF
--- a/internal/runners/push/push.go
+++ b/internal/runners/push/push.go
@@ -105,7 +105,7 @@ func (r *Push) Run(params PushParams) error {
 	}
 
 	// Prompt for namespace IF:
-	// - No namespace could be detect so far
+	// - No namespace could be detected so far
 	// - We want to create a copy of the current namespace, and no custom namespace was provided
 	if !targetNamespace.IsValid() || (intend&pushFromNoPermission > 0 && !params.Namespace.IsValid()) {
 		var err error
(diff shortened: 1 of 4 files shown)


pkg/platform/api/buildplanner/request/attachStagedCommit.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+type buildPlanByAttachStagedCommit struct {
+	vars map[string]interface{}
+}
+
+func (b *buildPlanByAttachStagedCommit) Query() string {
+	return `
+mutation ($organization: String!, $project: String!, $parentCommit: ID!, $stagedCommit: ID!, $branchRef: String!) {
+  attachStagedCommit(input: {organization: $organization, project: $project, parentCommitId: $parentCommit, stagedCommitId: $stagedCommit, branchRef: $branchRef}) {
```
mdrohmann, line 20: Scott suggested to rename this to `mergeCommit` actually (https://activestatef.atlassian.net/browse/PB-4375?focusedCommentId=43427), but it sounds like you also need an API to attach a staged commit to a branch (and possibly even create that branch/project?)

I wonder if that is actually more of a `createProject` use-case... Where you create a project with an existing commit.

mdrohmann: Also:  You'll need this for the next state tool release, correct?

MDrakos: `createProject` might be a useful mutation to add... we have this story: https://activestatef.atlassian.net/browse/DX-2173 which could spawn more use-cases for alternate mutations beyond the ones we have now.

internal/runners/push/push.go
```diff
@@ -209,12 +209,24 @@ func (r *Push) Run(params PushParams) error {
...
+	if branch.CommitID != nil {
+		bp := model.NewBuildPlannerModel(r.auth)
+		logging.Debug("Attaching staged commit via build planner")
+		err = bp.AttachStagedCommit(targetNamespace.Owner, targetNamespace.Project, branch.CommitID.String(), commitID.String(), branch.Label)
+		if err != nil {
+			return locale.WrapError(err, "err_push_attach_staged_commit", "Failed to attach staged commit to project.")
+		}
+	} else {
+		logging.Debug("Updating project branch commit via VCS")
+		err = model.UpdateProjectBranchCommitWithModel(targetPjm, branch.Label, commitID)
```
Naatan, line 223: `WithModel` seems redundant?

internal/runners/push/push.go
```diff
@@ -209,12 +209,24 @@ func (r *Push) Run(params PushParams) error {
...
-			return locale.WrapError(err, "push_project_branch_commit_err", "Failed to update new project {{.V0}} to current commitID.", targetNamespace.String())
+	// Currently, the build planner doesn't support attaching a staged commit
+	// without a parent commit ID
+	if branch.CommitID != nil {
+		bp := model.NewBuildPlannerModel(r.auth)
+		logging.Debug("Attaching staged commit via build planner")
+		err = bp.AttachStagedCommit(targetNamespace.Owner, targetNamespace.Project, branch.CommitID.String(), commitID.String(), branch.Label)
+		if err != nil {
+			return locale.WrapError(err, "err_push_attach_staged_commit", "Failed to attach staged commit to project.")
+		}
```
Naatan, line 220: I'm trying to understand what this is doing. If the branch has a commit attached to it, then we can't use `UpdateProjectBranch..`? That seems inverted at least in terms of my understanding reading this code. If that is the logic we want please add comments explaining why.

Additionally, we're not supplying a branch here, so what's getting updated?

internal/runners/push/push.go
```diff
@@ -209,12 +209,24 @@ func (r *Push) Run(params PushParams) error {
 	}
 
 	// Update the project at the given commit id.
-	err = model.UpdateProjectBranchCommitWithModel(targetPjm, branch.Label, commitID)
-	if err != nil {
-		if errs.Matches(err, &model.ErrUpdateBranchAuth{}) {
-			return locale.WrapInputError(err, "push_project_branch_no_permission", "You do not have permission to push to {{.V0}}.", targetNamespace.String())
-		} else {
-			return locale.WrapError(err, "push_project_branch_commit_err", "Failed to update new project {{.V0}} to current commitID.", targetNamespace.String())
+	// Currently, the build planner doesn't support attaching a staged commit
```
Naatan, line 212: This is a blocker for this story. Please file something with extensibility to address. There's no value in us adopting this new endpoint if it's not eliminating the old one.

Naatan (changes requested):

PR #2743, opened 2023-09-11
Process PlanningError build plan types as errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2183" title="DX-2183" target="_blank">BugDX-2183</a>  Failed Build error is not marked as input error
  </td></table>
  <br />
 

Not all of these types have non-nil Error fields.


MDrakos (approved):
This change makes sense but have we confirmed it addresses the issue in the story?

mitchell-as:
Yes, I stepped through it with a debugger both before and after. The fix is legit.

PR #2747, opened 2023-09-11
Update failure analytics events should set target version, not version.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2144" title="DX-2144" target="_blank">BugDX-2144</a>  Update events incorrectly setting the target_version
  </td></table>
  <br />


Naatan (changes requested):
Looks good, but you missed two; lines 112 and 119.

mitchell-as:
Thanks for pointing those out.

Naatan (approved):

PR #2749, opened 2023-09-11
`state auth signup` sends the user to the signup page instead of the login page.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2148" title="DX-2148" target="_blank">StoryDX-2148</a>  As a user when I use the state auth signup command I want it to automatically open the account creation page instead of the login page
  </td></table>
  <br />
 

The idea is to share as much of the authentication with browser logic is possible, but account for the differences in URL.

### DIFF
--- a/pkg/cmdlets/auth/login.go
+++ b/pkg/cmdlets/auth/login.go
@@ -1,9 +1,7 @@
 package auth
 
 import (
-	"fmt"
 	"net/url"
-	"strings"
 	"time"
 
 	"github.com/ActiveState/cli/internal/constants"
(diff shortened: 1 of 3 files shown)


daved (dismissed):

pkg/cmdlets/auth/login.go
```diff
@@ -226,22 +228,57 @@ func promptToken(credentials *mono_models.Credentials, out output.Outputer, prom
...
+		signupURL, err := url.Parse(constants.PlatformSignupURL)
+		if err != nil {
+			return errs.Wrap(err, "constants.PlatformSignupURL is not valid")
+		}
+		query := signupURL.Query()
+		query.Add("nextRoute", parsedURL.RequestURI())
+		signupURL.RawQuery = query.Encode()
+
+		verificationURL = signupURL.String()
+	}
```
MDrakos, line 272: Would this be more well placed in a function of its own? One function for authenticating with the browser and another for signup.

MDrakos (approved):
Looks good to me, just one question/potential improvement.

pkg/cmdlets/auth/login.go
mitchell-as: No, because they both use the same flow (request an auth token, call the browser, listen for response). The only difference between flows is the URL, hence the "if" you're commenting on. I didn't want to duplicate any code.

PR #2750, opened 2023-09-12
Add source to analytics events.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2145" title="DX-2145" target="_blank">TaskDX-2145</a>  We identify State Tool traffic as State Tool traffic
  </td></table>
  <br />
 

This allows for differentiation between state tool, offline installer, executor, etc.


cmd/state-installer/cmd.go
```diff
@@ -37,9 +38,6 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
-const AnalyticsCat = "installer"
-const AnalyticsFunnelCat = "installer-funnel"
```
mitchell-as, line 41: Moved into analytics/constants.go

internal/svcctl/comm.go
```diff
@@ -138,7 +138,7 @@ func HeartbeatHandler(cfg *config.Instance, resolver Resolver, analyticsReporter
 			}
 
 			logging.Debug("Firing runtime usage events for %s", metaData.Namespace)
-			analyticsReporter.Event(constants.CatRuntimeUsage, constants.ActRuntimeAttempt, dims)
+			analyticsReporter.Event(constants.CatRuntimeUsage, constants.ActRuntimeAttempt, constants.SrcExecutor, dims)
```
mitchell-as, line 141: It looks to me like all heartbeats come from executors, so that's why the source is this and not State Tool.

test/integration/analytics_int_test.go
```diff
@@ -179,7 +179,7 @@ func (suite *AnalyticsIntegrationTestSuite) TestActivateEvents() {
 
 func countEvents(events []reporters.TestLogEntry, category, action string) int {
 	filteredEvents := funk.Filter(events, func(e reporters.TestLogEntry) bool {
-		return e.Category == category && e.Action == action
+		return e.Category == category && e.Action == action && e.Source != ""
```
mitchell-as, line 182: Just ensure we're setting the source, as it can vary between State Tool and Executor.

Naatan (changes requested):
I think we should set this at construction time. We would never want to modify this on a per-event basis.

Also note with the svc that it handles two types of analytics:

- First, it has its own analytics that it sets up here:
	- https://github.com/ActiveState/cli/blob/4087dc2011eb58bd54c369380549da8e8aa6fa1b/cmd/state-svc/main.go#L101-L101 
	- This one should have the state-svc as the source
- It also facilitates analytics for the state tool itself, this one is constructed here:
	- https://github.com/ActiveState/cli/blob/4087dc2011eb58bd54c369380549da8e8aa6fa1b/cmd/state-svc/internal/resolver/resolver.go#L76-L76
	- This one should have the state tool as the source.

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -97,7 +97,7 @@ func (w *Watcher) check() {
 
 func (w *Watcher) RecordUsage(e entry) {
 	logging.Debug("Recording usage of %s (%d)", e.Exec, e.PID)
-	w.an.Event(anaConst.CatRuntimeUsage, anaConst.ActRuntimeHeartbeat, e.Dims)
+	w.an.EventWithSource(anaConst.CatRuntimeUsage, anaConst.ActRuntimeHeartbeat, anaConst.SrcExecutor, e.Dims)
```
mitchell-as, line 100: It looks to me like all heartbeats come from executors, so that's why the source is this and not State Tool.

mitchell-as:
Good call with putting source in the analytics client. We still need some wiggle room to differentiate between state tool and executors (particularly in the runtime area), so state-svc cannot assume `anForClient` is always coming from State Tool, and instead forwards on behalf of the incoming source (currently State Tool or Executor).

If you don't care to differentiate, let me know and I'll drop all that special handling. Or if you have another idea, let me know.

cmd/state-svc/internal/rtwatcher/watcher.go
Naatan: Good call! I forgot about that caveat.

pkg/platform/runtime/analytics/analytics.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/osutils"
+)
+
+var isExecutor bool
+
+func init() {
+	if osutils.Executable() == constants.StateExecutorCmd {
+		isExecutor = true
+	}
```
Naatan, line 16: This won't work because we rename executors to eg. python.exe.

But more importantly, executors do not import the analytics package. Instead they contact the svc here:

https://github.com/ActiveState/cli/blob/7d126eb0eda866520ebfc5ce0e6ac8fe13aa03da/cmd/state-exec/main.go#L87-L87

Which is handled in the svc here:

https://github.com/ActiveState/cli/blob/7d126eb0eda866520ebfc5ce0e6ac8fe13aa03da/internal/svcctl/comm.go#L84-L84

So I think you can spoof the source here:

https://github.com/ActiveState/cli/blob/7d126eb0eda866520ebfc5ce0e6ac8fe13aa03da/internal/svcctl/comm.go#L141-L141

Naatan (changes requested):
Looks great! In addition to the one case I flagged for executor handling, I think we should also have our analytics integration tests specifically check for it.

IIRC the analytics tests are already triggering state tool events as well as executor events, so with the data we're already generating we could assert that they have the expected source set.

pkg/platform/runtime/analytics/analytics.go
mitchell-as: Thanks for the note about executor names. I meant to use `runtime/executors.IsExecutor()` but it looks like that change got lost in the churn.

As for executors and analytics, I did make that change here: https://github.com/ActiveState/cli/pull/2750/files#diff-f69f873b04661f16e266f4298e50a2f0ee7040fc6810650aad57d7e08d407b96R141. Perhaps your IDE didn't show it.

pkg/platform/runtime/analytics/analytics.go
```diff
@@ -4,15 +4,15 @@ import (
...
 	"github.com/ActiveState/cli/internal/osutils"
+	"github.com/ActiveState/cli/pkg/platform/runtime/executors"
 )
 
 var isExecutor bool
 
 func init() {
-	if osutils.Executable() == constants.StateExecutorCmd {
-		isExecutor = true
+	if isExec, err := executors.IsExecutor(osutils.Executable()); err == nil {
```
Naatan, line 14: As I said executors do not use the analytics package directly, so I don't think we need this?

test/integration/analytics_int_test.go
```diff
@@ -82,16 +82,16 @@ func (suite *AnalyticsIntegrationTestSuite) TestActivateEvents() {
...
 			cp.Snapshot(), ts.DebugLogs()))
 
 	// Runtime:success events
-	suite.assertNEvents(events, 1, anaConst.CatRuntime, anaConst.ActRuntimeSuccess,
+	suite.assertNEvents(events, 1, anaConst.CatRuntime, anaConst.ActRuntimeSuccess, anaConst.SrcStateTool,
 		fmt.Sprintf("output:\n%s\n%s",
 			cp.Snapshot(), ts.DebugLogs()))
 
-	heartbeatInitialCount := countEvents(events, anaConst.CatRuntimeUsage, anaConst.ActRuntimeHeartbeat)
+	heartbeatInitialCount := countEvents(events, anaConst.CatRuntimeUsage, anaConst.ActRuntimeHeartbeat, anaConst.SrcExecutor)
```
Naatan, line 94: Can we also assert here that we got both executor and state tool heartbeats?

ie. the `state activate` should've triggered heartbeats from the state tool, whereas the executor would've triggered it from the executor.

Naatan (changes requested):

test/integration/analytics_int_test.go
mitchell-as: Okay, so my comment here is not valid, right? https://github.com/ActiveState/cli/pull/2750#discussion_r1324799079

pkg/platform/runtime/analytics/analytics.go
Naatan: That's right, but this code is checking the current executable. There is never a scenario in which an executor invokes the analytics code (whether sync or async) directly. Instead they talk directly to the svc without the analytics library. The svc then uses its analytics library to report the event.

In other words; I don't believe there is a scenario in which this conditional evaluates to true.

test/integration/analytics_int_test.go
Naatan: Sorry that's my bad. Your comment there is indeed incorrect, I should've known that but I guess I wasn't thinking it through.

pkg/platform/runtime/analytics/analytics.go
mitchell-as: Okay, you're right. I went through a debugger and that code does not appear to be hit. I've removed it.

Naatan (changes requested):
Looks good! I think you missed this comment though: https://github.com/ActiveState/cli/pull/2750#discussion_r1326363917

mitchell-as:
Sorry about that. Too many double negatives. I got confused and thought the opposite of what you were thinking :S

cmd/state-svc/internal/rtwatcher/watcher.go
```diff
@@ -97,7 +99,11 @@ func (w *Watcher) check() {
 
 func (w *Watcher) RecordUsage(e entry) {
 	logging.Debug("Recording usage of %s (%d)", e.Exec, e.PID)
-	w.an.EventWithSource(anaConst.CatRuntimeUsage, anaConst.ActRuntimeHeartbeat, anaConst.SrcExecutor, e.Dims)
+	source := anaConst.SrcExecutor
+	if filepath.Base(e.Exec) == constants.StateCmd+osutils.ExeExt {
+		source = anaConst.SrcStateTool
+	}
+	w.an.EventWithSource(anaConst.CatRuntimeUsage, anaConst.ActRuntimeHeartbeat, source, e.Dims)
```
Naatan, line 106: This feels too error prone to me. Could we instead add the source as an argument for the ReportRuntimeUsage resolver on the svc?

The only place executors call this is here:

https://github.com/ActiveState/cli/blob/be9579908d5177b979384030e17906609b3645df/internal/svcctl/comm.go#L142-L142

Any other place would be state tool. I'd be fine defaulting to state tool if the source is empty.

Naatan (changes requested):

mitchell-as:
The change got a bit involved since there's graphql involved, so let me know if you had something else in mind.

Naatan (approved):
Looks great!

PR #2751, opened 2023-09-12
Fix JSON messages in history

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2195" title="DX-2195" target="_blank">BugDX-2195</a>  Running `state history` in dashboard-created Windows project shows JSON in "changes" section
  </td></table>
  <br />
 

This change isn't ideal but it will work for now until we can get the history in the format of a build expression. Asked the Dashboard squad how they are handling this. Here is their changeset: https://github.com/ActiveState/TheHomeRepot/pull/11922/files.

Here is an example of the updated output:

```
█ Viewing Project History

Operating on project mitchell-as/docker, located at C:\Users\Mike\work\testing\docker.

Here are the most recent changes made to this project.

  Commit     a518738b-402b-404b-805f-97f9276f2cdf
  Author     mitchell-as
  Date       01 Sep 23 16:41 UTC
  Message    Not provided.
  Changes     - + Docker/Installer packagers
              - + python 3.10.12
              - + Platform
```

### DIFF
--- a/pkg/cmdlets/commit/commit.go
+++ b/pkg/cmdlets/commit/commit.go
@@ -105,6 +105,13 @@ func FormatChanges(commit *mono_models.Commit) ([]string, []*requirementChange) 
 			versionConstraints = ""
 		}
 
+		// This is a temporary fix until we start getting history in the form of build expressions
+		if model.NamespaceMatch(change.Namespace, model.NamespaceBuildFlagsMatch) &&
+			(strings.Contains(change.Requirement, "docker") || strings.Contains(change.Requirement, "installer")) {
+			requirement = locale.T("namespace_label_packager")
+			versionConstraints = ""
+		}
+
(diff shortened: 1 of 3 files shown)


pkg/cmdlets/commit/commit.go
```diff
@@ -105,6 +105,13 @@ func FormatChanges(commit *mono_models.Commit) ([]string, []*requirementChange)
 			versionConstraints = ""
 		}
 
+		// This is a temporary fix until we start getting history in the form of build expressions
```
mitchell-as, line 108: Is there a ticket filed for this? We should probably reference here so we don't lose track of this comment.

mitchell-as (changes requested):
This looks fine to me, but we should add some more tracking info.

mitchell-as (approved):

PR #2754, opened 2023-09-15
Address merge conflict between v41 and v42.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2145" title="DX-2145" target="_blank">TaskDX-2145</a>  We identify State Tool traffic as State Tool traffic
  </td></table>
  <br />
 

The conflict was in setup.go with some import reordering.


mitchell-as:
@Naatan please force-merge, as I cannot do so.

PR #2761, opened 2023-09-18
Update runtime use tracking

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2162" title="DX-2162" target="_blank">TaskDX-2162</a>  We track ALL runtime use as runtime use
  </td></table>
  <br />
 

Decided to go with the simpler blacklist approach. This means we will have to be a little more careful calling `IndicatesUsage()`

### DIFF
--- a/pkg/platform/runtime/target/target.go
+++ b/pkg/platform/runtime/target/target.go
@@ -54,7 +54,8 @@ func NewExecTrigger(cmd string) Trigger {
 }
 
 func (t Trigger) IndicatesUsage() bool {
-	return !strings.EqualFold(string(t), string(TriggerRefresh))
+	// All triggers should indicate runtime use except for refreshing executors
+	return !strings.EqualFold(string(t), string(TriggerResetExec))
 }
 
 type ProjectTarget struct {
--- a/pkg/platform/runtime/target/target.go
+++ b/pkg/platform/runtime/target/target.go
@@ -14,7 +14,6 @@ import (
 	"github.com/ActiveState/cli/pkg/localcommit"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/go-openapi/strfmt"
-	"github.com/thoas/go-funk"
 )
 
 type Trigger string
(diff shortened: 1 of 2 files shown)


pkg/platform/runtime/target/target.go
```diff
@@ -50,40 +49,12 @@ const (
...
 func (t Trigger) IndicatesUsage() bool {
-	if funk.Contains(usageTriggers, t) {
-		return true
-	}
-	return t.IsExecTrigger() && funk.Contains(usageTriggers, TriggerExec)
-}
-
-func (t Trigger) IsExecTrigger() bool {
-	return strings.HasPrefix(string(t), string(TriggerExec)+": ")
+	return !strings.EqualFold(string(t), string(TriggerRefresh))
```
mitchell-as, line 57: It is not clear that this is a blacklist. Also, I thought we wanted to consider `state refresh` as a runtime trigger (bullet 3 in the JIRA ticket). Furthermore, AC 2 wants to ensure we don't trigger use on state tool update. I'm not positive this is addressed here.

pkg/platform/runtime/target/target_test.go
```diff
@@ -15,7 +15,7 @@ func TestTrigger_IndicatesUsage(t *testing.T) {
 		},
 		{
 			"Reset exec does not count as usage",
-			TriggerResetExec,
+			TriggerRefresh,
```
mitchell-as, line 18: I think we do want to count this as runtime use. Regardless of what this ends up being, please update the documentation string in the line above.

mitchell-as (changes requested):
I'm not sure all of the ACs have been met. Just because this ticket has been "in progress" for a prolonged period of time shouldn't mean we rush it through. There are valid reasons for delay, so let's spend the necessary time to get it right :)

pkg/platform/runtime/target/target.go
MDrakos: Whoops, I used the wrong trigger. I think updating it should address the issues you've raised but please let me know if I've missed something.

mitchell-as (approved):
I approve, provided that AC 2 is met. Based on comments and changes, it appears that `state update` only triggers `TriggerResetExec`, which is ignored here.

MDrakos:
> I approve, provided that AC 2 is met. Based on comments and changes, it appears that `state update` only triggers `TriggerResetExec`, which is ignored here.

That's correct, `state update` triggers `state _prepare` which will reset executors.

PR #2762, opened 2023-09-18
Remove `state auth signup --prompt`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2198" title="DX-2198" target="_blank">StoryDX-2198</a>  We no longer allow signup via a prompt
  </td></table>
  <br />
 

Only signups via the browser are supported.


mitchell-as:
The auth integration test failure doesn't appear to be with these changes. It looks like only the most recent error is being shown to the user. I think it's related to https://activestatef.atlassian.net/browse/DX-2205

pkg/cmdlets/auth/login.go
```diff
@@ -58,8 +58,8 @@ func AuthenticateWithInput(
 			if !uniqueUsername(credentials) {
 				return errs.Wrap(err, "uniqueUsername failed")
 			}
-			if err := promptSignup(credentials, out, prompt, auth); err != nil {
-				return errs.Wrap(err, "promptSignup failed")
+			if err := SignupWithBrowser(out, auth, prompt); err != nil {
+				return errs.Wrap(err, "SignupWithBrowser failed")
```
Naatan, line 62: Should just error out if we reach this. Launching the browser without context while in a prompt flow will just cause confusion.

Naatan (changes requested):

Naatan (approved):

PR #2764, opened 2023-09-19
Fix error messages being incomplete

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2205" title="DX-2205" target="_blank">BugDX-2205</a>  SHELL: Missing message in case user try `state shell` to the unchecked project.
  </td></table>
  <br />

### DIFF
--- a/internal/locale/errors.go
+++ b/internal/locale/errors.go
@@ -12,6 +12,8 @@ import (
 	"github.com/ActiveState/cli/internal/rtutils"
 )
 
+var _ ErrorLocalizer = &LocalizedError{}
+
 // LocalizedError is an error that has the concept of user facing (localized) errors as well as whether an error is due
 // to user input or not
 type LocalizedError struct {
@@ -27,8 +29,8 @@ func (e *LocalizedError) Error() string {
 	return e.localized
 }
 
-// UserError is the user facing error message, it's the same as Error() but identifies it as being user facing
-func (e *LocalizedError) LocalizedError() string {
+// LocaleError is the user facing error message, it's the same as Error() but identifies it as being user facing
+func (e *LocalizedError) LocaleError() string {
 	return e.localized
 }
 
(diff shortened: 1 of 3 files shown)


pkg/cmdlets/errors/errors.go
```diff
@@ -113,14 +113,6 @@ func ParseUserFacing(err error) (int, error) {
 		return code, nil
 	}
 
-	// If there is a user facing error in the error stack we want to ensure
-	// that is it forwarded to the user.
-	var userFacingError errs.UserFacingError
-	if errors.As(err, &userFacingError) {
-		logging.Debug("Returning user facing error, error stack: \n%s", errs.JoinMessage(err))
-		return code, &OutputError{userFacingError}
-	}
```
Naatan, line 122: This wasn't achieving anything, either way we wrap it in OutputError with the `error` type.

internal/locale/errors.go
```diff
@@ -27,8 +29,8 @@ func (e *LocalizedError) Error() string {
 	return e.localized
 }
 
-// UserError is the user facing error message, it's the same as Error() but identifies it as being user facing
-func (e *LocalizedError) LocalizedError() string {
+// LocaleError is the user facing error message, it's the same as Error() but identifies it as being user facing
+func (e *LocalizedError) LocaleError() string {
```
Naatan, line 33: I was scratching my head for a while on this PR, debugging it nothing seemed wrong. Seeing no other options I figured I'd try just renaming the symbol as that's all that was done in the other PR that affected the behaviour, and in fact that did fix the issue.

I have NO idea why this fixes it though :\ Or rather, why it breaks with the name being `LocalizedError`.

MDrakos (approved):

Naatan:
After more digging and with a tip from @mitchell-as found the issue was down to the method naming. Filed a bug with Go here - https://github.com/golang/go/issues/62735.

PR #2765, opened 2023-09-19
Added integration test covering projects with docker and offline installers.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2201" title="DX-2201" target="_blank">TaskDX-2201</a>  We have integration tests covering projects with docker/installers
  </td></table>
  <br />


mitchell-as:
The Windows failure is logged by DX-2141 (Windows `package` integration tests time out on PRs)

test/integration/package_int_test.go
```diff
@@ -670,6 +670,33 @@ func (suite *PackageIntegrationTestSuite) TestRuby() {
...
 }
 
+// TestProjectWithOfflineInstallerAndDocker just makes sure we can checkout and install/uninstall
+// packages for projects with offline installers and docker runtimes.
+func (suite *PackageIntegrationTestSuite) TestProjectWithOfflineInstallerAndDocker() {
+	suite.OnlyRunForTags(tagsuite.Package)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	ts.LoginAsPersistentUser()
```
MDrakos, line 680: Is this necessary? The project is public and this only slows down the test.

MDrakos (changes requested):

test/integration/package_int_test.go
mitchell-as: Yes it is. I had a slack conversation in the buildplanner channel about this yesterday: https://activestate.slack.com/archives/C03T15HMBMG/p1695155288925919. Originally I wasn't logged in. Once I was, the problem went away.

MDrakos: Ahh, I see... it makes sense since installers are an enterprise feature you would have to be logged in and a part of the org to make changes. Could you add a quick comment that explains this? I could see someone (me) looking at this test later and removing that line because it "shouldn't" be necessary

MDrakos (changes requested):
See comment

MDrakos (approved):

PR #2766, opened 2023-09-19
Fixed commit date display in `state history`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2196" title="DX-2196" target="_blank">BugDX-2196</a>  Commit date field is incorrect in `state history` output
  </td></table>
  <br />

### DIFF
--- a/pkg/cmdlets/commit/commit.go
+++ b/pkg/cmdlets/commit/commit.go
@@ -74,8 +74,7 @@ func newCommitOutput(commit *mono_models.Commit, orgs []gmodel.Organization, isL
 		StructuredChanges: structuredChanges,
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cmdlets/commit/commit.go
```diff
@@ -74,8 +74,7 @@ func newCommitOutput(commit *mono_models.Commit, orgs []gmodel.Organization, isL
 		StructuredChanges: structuredChanges,
 	}
 
-	commitOutput.Date = commit.AtTime.String()
```
mitchell-as, line 77: This is immediately changed on line 82 (81 in the new diff), so that's why I removed it here.

MDrakos (approved):

PR #2767, opened 2023-09-19
 No need to see if there were any buildexpression changes after the fact

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2014" title="DX-2014" target="_blank">TaskDX-2014</a>  We don't rely on GetRevertCommit for detecting order changes
  </td></table>
  <br />
 

StageCommit will error if there are no changes.

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -261,11 +261,21 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
 		if err != nil {
 			return errs.Wrap(err, "Unable to get local commit")
 		}
-		revertCommit, err := model.GetRevertCommit(localCommitID, commitID)
+
+		parentExpr, err := bp.GetBuildExpression(pj.Owner(), pj.Name(), localCommitID.String())
 		if err != nil {
-			return locale.WrapError(err, "err_revert_refresh")
+			return errs.Wrap(err, "Could not get remote build expr to see if changes were made")
 		}
-		exprChanged = len(revertCommit.Changeset) > 0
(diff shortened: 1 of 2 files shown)


mitchell-as:
It's certainly possible that build expression comparison will be changed/overhauled by DX-1691 (Unify BuildPlanner and Local Order Build Expression types).

internal/runbits/requirements/requirements.go
```diff
@@ -261,11 +261,21 @@ func (r *RequirementOperation) ExecuteRequirementOperation(requirementName, requ
...
+			return errs.Wrap(err, "Could not transform remote build expr to build script to see if changes were made")
+		}
+
+		expr, err := bp.GetBuildExpression(pj.Owner(), pj.Name(), commitID.String())
 		if err != nil {
-			return locale.WrapError(err, "err_revert_refresh")
+			return errs.Wrap(err, "Could not get new build expr to see if changes were made")
 		}
-		exprChanged = len(revertCommit.Changeset) > 0
+		exprChanged = !parentScript.EqualsBuildExpression(expr)
```
MDrakos, line 278: `stageCommit` _should_ throw an error if we submit an expression that has no changes so I wonder if this calculation is even necessary now. Could you verify and remove this code if that is the case?

MDrakos (changes requested):

internal/runbits/requirements/requirements.go
mitchell-as: Yes, you are correct! I've removed this unnecessary code :)

MDrakos (approved):

PR #2768, opened 2023-09-20
Turn off terminal echo while progressbar is present.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2050" title="DX-2050" target="_blank">BugDX-2050</a>  Key presses during artifact download causes stuff terminal UI elements
  </td></table>
  <br />
 

I am still able to Ctrl+C and cancel, and the echo is restored.

### DIFF
--- a/internal/osutils/termecho/termecho.go
+++ b/internal/osutils/termecho/termecho.go
@@ -1,0 +1,9 @@
+package termecho
+
+func Off() error {
+	return toggle(false)
+}
+
+func On() error {
+	return toggle(true)
+}
--- a/internal/osutils/termecho/termecho_darwin.go
+++ b/internal/osutils/termecho/termecho_darwin.go
@@ -1,0 +1,6 @@
+package termecho
+
+import "golang.org/x/sys/unix"
+
+const ioctlReadTermios = unix.TIOCGETA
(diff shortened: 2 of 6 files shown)


Naatan (changes requested):
This is really cool, definitely something I'd like to incorporate! However, this is for a bug, not for a story or task. Bugs should lead to fixes, not new enhancements.

It seems likely this cannot be fixed without an enhancement, so we should bring this back to triage to determine how better to address this.

I'd see this ultimately landing as part of the subshell package, or at least have it use the subshell package to determine the shell. We cannot rely on the SHELL environment variable alone.

mitchell-as:
I agree that it would be nice to somehow detect the current shell on Windows. On Linux and macOS it does not matter -- the POSIX APIs are the same and are independent of the running shell. However, I would like to point out that we tend to associated Git Bash with Bash on Windows, and that's a non-interactive shell. We probably won't be able to turn off echoing there, since it's not a true terminal with termios functionality. This may be the best we can do on Windows.

Anyway, punting back to triage as suggested.

mitchell-as:
Implemented in DX-2214 as a story.

mitchell-as:
Implemented in DX-2214 as a task.

PR #2771, opened 2023-09-25
Update changelog

MDrakos:


changelog.md
```diff
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* Users can now check out a project without cloning the associated git
+  repository.
+
+### Fixed
+
+* Several error message improvements for errors arising from unauthenticated actions.
+* Fixed failed runtime installations due to parsing errors.
+* No longer including unneccessary artifacts in runtime installations.
+
+### Changed
```
Naatan, line 34: Please put Changed after Added (for consistency).

changelog.md
```diff
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### 0.41.0
+
+### Added
+
+* `state init` is now a stable command, meaning you no longer need to opt-in to
+  unstable commands to use it.
+* Signing up for a new account now opens the account creation page in your
+  browser.
```
Naatan, line 16:
```suggestion
* Signing up for a new account now opens the account creation page in your
  browser. Instead of bringing you to the sign in page.
```

changelog.md
```diff
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* `state shell` can now detect currently active subshells preventing nested
+  shells from being created.
+* The State Tool will now print a message when starting a subshell, indicating
+  the project that is being used.
+* Wildcard version matching is now supported for `state install`.
```
Naatan, line 21: Nit: Would be good to give an example.

changelog.md
```diff
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* Full or partial version matching is now supported for language versions when
```
Naatan, line 22: Nit: Probably good to combine this with the previous one (ie. state install and state init now support wildcard versioning).

changelog.md
Naatan: This should be in Changed imo. It always opened your browser, the URL is just different now (brings you straight to signup).

changelog.md
```diff
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  using `state init`.
+* Added messaging on the potentially disruptive nature of editing or moving a project.
+* Users can now check out a project without cloning the associated git
```
Naatan, line 25: Nit: Would be good to give an example.

changelog.md
```diff
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+### Fixed
+
+* Several error message improvements for errors arising from unauthenticated actions.
+* Fixed failed runtime installations due to parsing errors.
+* No longer including unneccessary artifacts in runtime installations.
+
+### Changed
+
+* The default Ruby version is now 3.2.2
+* Errors encountered while sourcing a runtime now have more informative error
```
Naatan, line 37: Imo this and the one on line 31 should be in the same section.

changelog.md
```diff
@@ -6,6 +6,40 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  using `state init`.
+* Added messaging on the potentially disruptive nature of editing or moving a project.
+* Users can now check out a project without cloning the associated git
+  repository.
+
+### Fixed
+
+* Several error message improvements for errors arising from unauthenticated actions.
+* Fixed failed runtime installations due to parsing errors.
+* No longer including unneccessary artifacts in runtime installations.
```
Naatan, line 32: These don't sound very user facing. Try to address what was fixed for the user. Not what was changed under the hood to facilitate the fix.

Naatan (changes requested):

changelog.md
```diff
@@ -6,6 +6,43 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  browser. Instead of bringing you to the sign in page.
+* `state shell` can now detect currently active subshells preventing nested
+  shells from being created.
+* The State Tool will now print a message when starting a subshell, indicating
+  the project that is being used.
+* Wildcard and partial version matching is now supported for `state install` 
+  and for language versions with `state init`.
+  For example: 
+  * `state install pytest@2.x`
+  * `state init "python@>3.7,<3.8"`
```
Naatan, line 25: I don't think this one is new? Only the wildcards. You can already do the larger than / smaller than on the currently released version.

changelog.md
```diff
@@ -6,6 +6,43 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* Added messaging on the potentially disruptive nature of editing or moving a project.
+* Users can now check out a project without cloning the associated git
+  repository.
+  For example: `state checkout <orgname/project> --no-clone`
```
Naatan, line 29:
```suggestion
  repository. For example: `state checkout <orgname/project> --no-clone`
```

changelog.md
```diff
@@ -6,6 +6,43 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  unstable commands to use it.
+* Signing up for a new account now opens the account creation page in your
+  browser. Instead of bringing you to the sign in page.
+* `state shell` can now detect currently active subshells preventing nested
+  shells from being created.
+* The State Tool will now print a message when starting a subshell, indicating
+  the project that is being used.
+* Wildcard and partial version matching is now supported for `state install` 
+  and for language versions with `state init`.
+  For example:
```
Naatan, line 23: Do you intend for this to be rendered on a new line? It's not. Same for the one on line 29.

Naatan (changes requested):

changelog.md
```diff
@@ -6,6 +6,43 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### 0.41.0
+
+### Added
+
+* `state init` is now a stable command, meaning you no longer need to opt-in to
+  unstable commands to use it.
+* Signing up for a new account now opens the account creation page in your
```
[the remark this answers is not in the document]
MDrakos: I applied your suggestion as part of a batch of changes. I've updated the message again incase that's what you were looking for.

changelog.md
MDrakos: This example is from this story: https://activestatef.atlassian.net/browse/DX-2010. It looks like there was something preventing us from doing this before?

changelog.md
MDrakos: Whoops, yes I did. Thanks for catching that.

MDrakos: Though it looks from your suggestion that you'd prefer it be on one line so I've made that change. Also removed the second example with the wildcards.

changelog.md
Naatan: It had to do with what partial versions got turned into. But fully qualified versions like the second example were already working.

The change is basically that you now have to use the wildcard notation if you don't want to specify a full version or use the larger than/smaller than notation.

changelog.md
Naatan: Oh I didn't mean to suggest that change, I was trying to figure out what this should look like an accidentally submitted it.

Naatan (approved):

PR #2772, opened 2023-09-26
Add ability to install artifacts in the build time closure

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2175" title="DX-2175" target="_blank">StoryDX-2175</a>  As a user, I can install my runtime with build-time dependencies
  </td></table>
  <br />


pkg/platform/runtime/setup/setup.go
```diff
@@ -481,23 +508,17 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
...
 
 	artifactsToInstall := []artifact.ArtifactID{}
-	// buildtimeArtifacts := runtimeArtifacts
-	if buildResult.BuildReady {
-		// If the build is already done we can just look at the downloadable artifacts as they will be a fully accurate
-		// prediction of what we will be installing.
-		for _, a := range downloadablePrebuiltResults {
-			if _, alreadyInstalled := alreadyInstalled[a.ArtifactID]; !alreadyInstalled {
-				artifactsToInstall = append(artifactsToInstall, a.ArtifactID)
-			}
```
Naatan, line 491: I don't understand why this conditional could be dropped? If the build isn't ready then the prebuild results cannot be relied upon.

test/integration/checkout_int_test.go
```diff
@@ -244,6 +244,26 @@ func (suite *CheckoutIntegrationTestSuite) TestCheckoutCaseInsensitive() {
...
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("checkout", "ActiveState-CLI/small-python"),
+		e2e.OptAppendEnv(constants.InstallBuildDependencies+"=true"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
+	)
+	// Expect the number of build deps to be 27 which is more than the number of runtime deps.
+	cp.Expect("27")
```
Naatan, line 263: Can we check for a specific dependency instead? eg. check if some file exists after it is done. This feels error-prone and hard to maintain should we make changed to the build later on.

Naatan (changes requested):

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -36,7 +36,7 @@ const (
 
 	// Tag types
 	TagSource     = "src"
-	TagDependency = "dep"
+	TagDependency = "deps"
```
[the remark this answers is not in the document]
MDrakos: Yes, we weren't correctly identifying the step dependencies before.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -268,15 +271,36 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+	}
 
+	// Get the platform ID for the current platform
+	platformID, err := platformModel.FilterCurrentPlatform(platformModel.HostPlatform, bpPlatforms)
+	if err != nil {
+		return nil, locale.WrapError(err, "err_filter_current_platform")
+	}
+
+	// Filter the build terminals to only include the current platform
+	var filteredTerminals []*model.NamedTarget
```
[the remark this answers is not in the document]
MDrakos: Yes, there can be.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -268,15 +271,36 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
+	var bpPlatforms []strfmt.UUID
+	for _, t := range build.Terminals {
+		if t.Tag == model.TagOrphan {
+			continue
+		}
+		bpPlatforms = append(bpPlatforms, strfmt.UUID(strings.TrimPrefix(t.Tag, "platform:")))
+	}
 
+	// Get the platform ID for the current platform
+	platformID, err := platformModel.FilterCurrentPlatform(platformModel.HostPlatform, bpPlatforms)
```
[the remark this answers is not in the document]
MDrakos: I realize now that we were filtering the platform IDs when we fetched the build plan, which feels incorrect to me as fetching doesn't imply that the response will be modified. I've removed the filtering there and move it to a helper function that is called when we process the build plan for both runtime and buildtime closures.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -287,131 +311,150 @@ func BuildtimeArtifacts(build *model.Build, includeBuilders bool) (artifact.Map,
...
-				return nil, errs.Wrap(err, "Could not build map for step dependencies of artifact %s", artifact.NodeID)
+				return nil, errs.Wrap(err, "Could not build map for step dependencies of artifact %s", currentArtifact.NodeID)
 			}
 		}
 	}
 
-	return result, nil
+	var uniqueDeps []strfmt.UUID
+	for id := range deps {
+		if _, ok := deps[id]; !ok {
```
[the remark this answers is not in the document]
MDrakos: Whoops, `uniqueDeps` is a slice, and `deps` is a map so the check is definitely redundant.

test/integration/checkout_int_test.go
MDrakos: Removed this test as we don't want to support installing the build time closure.

pkg/platform/runtime/setup/setup.go
MDrakos: Added back.

pkg/platform/runtime/setup/setup.go
```diff
@@ -382,19 +382,40 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
...
 		}
 	}
 
-	setup, err := s.selectSetupImplementation(buildResult.BuildEngine, runtimeArtifacts)
+	var setup Setuper
+	if !buildResult.BuildReady {
+		setup, err = s.selectSetupImplementation(buildResult.BuildEngine, runtimeAndBuildtimeArtifacts)
+	} else {
+		setup, err = s.selectSetupImplementation(buildResult.BuildEngine, runtimeArtifacts)
+	}
```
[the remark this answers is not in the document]
MDrakos: It does not, however, because the setup implementation is responsible for resolving artifact names if we do not pass it the correct artifact map we will get a bunch of `unknown` artifact progress bars.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -287,131 +311,150 @@ func BuildtimeArtifacts(build *model.Build, includeBuilders bool) (artifact.Map,
...
-	step, ok := lookup[artifact.GeneratedBy].(*model.Step)
+	// Here we iterate through the inputs of the step that generated the
+	// artifact, specifically the inputs that are tagged as dependencies.
+	// We recursively add all of the dependencies of the step intputs to
+	// the result map. This is the buildtime closure calculation.
+	step, ok := lookup[currentArtifact.GeneratedBy].(*model.Step)
 	if !ok {
-		_, ok := lookup[artifact.GeneratedBy].(*model.Source)
+		// Artifact was not generated by a step or a source, meaning that
+		// the buildplan is likely malformed.
```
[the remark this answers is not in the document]
MDrakos: Good point, I don't believe this is.

pkg/platform/api/buildplanner/model/buildplan.go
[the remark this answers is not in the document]
Naatan: That's on me, I should have been clearer :p Does this fix something unrelated to this PR? And if so, what is the use-case that is fixed?

pkg/platform/runtime/buildplan/buildplan.go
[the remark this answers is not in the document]
Naatan: Sure, but what I am asking for is to separate out the identification of the correct platform and feed that back into the flow for sourcing the runtime. The key point being that the outcome for deciding which platform to use comes from a central place. Right now there's two distinct pieces of code that each need to be updated should we make changes to how the platform is selected.

pkg/platform/runtime/setup/setup.go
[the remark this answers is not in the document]
Naatan: Ok, that seems like a problem worth addressing, but let's address it explicitly rather than implicitly. This is breaking the contract I worked hard to set up because of all the issues we ran into by being so implicit.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -48,6 +56,33 @@ func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
 	return res, nil
 }
 
+func filterTerminals(build *model.Build) ([]*model.NamedTarget, error) {
```
Naatan, line 59: Add comments please. Reading the code this seems to be about filtering only the terminal relevant to our platform, might be good to name it something more indicative like `filterPlatformTerminals()`. Current naming is very generic.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -268,15 +302,18 @@ func NewNamedMapFromBuildPlan(build *model.Build) (artifact.NamedMap, error) {
...
-// BuildtimeArtifacts iterates through all artifacts in a given build and
+// NewBuildtimeMap iterates through all artifacts in a given build and
 // adds the artifact's dependencies to a map. This is different from the
 // runtime dependency calculation as it includes ALL of the input artifacts of the
 // step that generated each artifact. The includeBuilders argument determines whether
 // or not to include builder artifacts in the final result.
-func BuildtimeArtifacts(build *model.Build, includeBuilders bool) (artifact.Map, error) {
-	result := make(artifact.Map)
-	lookup := make(map[strfmt.UUID]interface{})
+func NewBuildtimeMap(build *model.Build) (artifact.Map, error) {
```
Naatan, line 310: This is my second run reading through this code and I'll apologize in advance for not raising this the first time around;

It feels a little unproductive for us to iterate and convert the build map twice through two distinct code paths to calculate the artifact map. Once for runtime artifacts, and then once for build time artifacts. In each case the logic is very similar.

Would it be better if we did a single calculation of the artifact map, and the artifact map itself simply has a property like `IsBuildTimeOnly` which can be further filtered on?

That'd keep the performance nimble but also reduces the amount of code to maintain and update.

Naatan (changes requested):

pkg/platform/runtime/buildplan/buildplan.go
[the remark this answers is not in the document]
Naatan: Note this comment here would also indirectly resolve this issue:

https://github.com/ActiveState/cli/pull/2772#discussion_r1348972686

pkg/platform/api/buildplanner/model/buildplan.go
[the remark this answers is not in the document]
MDrakos: No, it's related to this PR. Without thins change, we aren't able to correctly identify step dependencies.

pkg/platform/runtime/setup/setup.go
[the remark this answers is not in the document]
MDrakos: Fair enough. I've move the name resolution out to it's own type that is responsible for just that.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: I was able to use one function with a boolean switch. Unfortunately, we can't filter the build plan down from the runtime closure to the buildtime closure because the runtime closure is a subset of the buildtime one. This would mean that we would have to _always_ calculate the buildtime closure and then filter down to the runtime one.

Instead, I've opted to just use one function to calculate both and use a boolean switch. This does mean that there are instances where we may have to iterate through the build plan multiple times to calculate both closures.

Naatan: I don't fully understand this limitation. We wouldn't be filtering down from either runtime or buildtime closure. We would be iterating through the buildplan and generating both based (if requested). The point being there is only a single iteration that produces both a build time and runtime closure.

pkg/platform/runtime/setup/implementations/camel/resolver.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+type Resolver struct{}
+
+func NewResolver() *Resolver {
+	return &Resolver{}
+}
+
+func (r *Resolver) ResolveArtifactName(_ artifact.ArtifactID) string {
+	return locale.Tl("camel_bundle_name", "bundle")
```
Naatan, line 15:
```suggestion
	return locale.Tl("camel_bundle_name", "legacy bundle")
```

pkg/platform/runtime/setup/setup.go
```diff
@@ -401,19 +404,47 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
...
+	if strings.EqualFold(os.Getenv(constants.InstallBuildDependencies), "true") {
+		logging.Debug("Installing build dependencies")
+		includeBuildtimeClosure = true
+		if runtimeAndBuildtimeArtifacts == nil {
+			runtimeAndBuildtimeArtifacts, err = buildplan.NewMapFromBuildPlan(buildResult.Build, includeBuildtimeClosure)
+			if err != nil {
+				return nil, nil, errs.Wrap(err, "Failed to create artifact map from build plan")
+			}
+		}
+		runtimeArtifacts = runtimeAndBuildtimeArtifacts
```
Naatan, line 429: This is raising a red flag for me. I'd rather we be explicit. I would suggest capturing the runtime and build time closures separately, and combining them for use-cases where that may be needed. Note I am not suggesting you calculate them separately, but you can do a single calculation that captures both.

pkg/platform/runtime/setup/setup.go
```diff
@@ -401,19 +404,47 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
...
 			return nil, nil, errs.Wrap(err, "Failed to create artifact map from build plan")
 		}
 	}
 
-	setup, err := s.selectSetupImplementation(buildResult.BuildEngine, runtimeArtifacts)
+	var resolver ArtifactResolver
+	if !buildResult.BuildReady {
+		resolver, err = selectArtifactResolver(buildResult.BuildEngine, runtimeAndBuildtimeArtifacts)
+	} else {
+		resolver, err = selectArtifactResolver(buildResult.BuildEngine, runtimeArtifacts)
```
Naatan, line 441: This seems to be handling implicit logic. I'd prefer we rewrite/rename things to make it explicit, or if that's not possible, add a comment.

Naatan (changes requested):
It's looking better but I still feel like there's a lot of implicit logic. Granted that is a painpoint of the runtime package already, but I'd rather not add to it as efforts of late have been to reduce it.

That said I realize this PR is dragging on due to reviews, I'm open to finding interim solutions (eg. lots of commenting to surface the implicit logic).

Naatan:
Also see my response here - https://github.com/ActiveState/cli/pull/2772#discussion_r1364187054

pkg/platform/runtime/setup/setup.go
```diff
@@ -482,6 +463,7 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
 		s.analytics.Event(anaConsts.CatRuntimeDebug, anaConsts.ActRuntimeBuild, dimensions)
 	}
 
+	includeBuildtimeClosure := strings.EqualFold(os.Getenv(constants.InstallBuildDependencies), "true") || !buildResult.BuildReady
```
Naatan, line 466: I understand we would want to report progress on build time artifacts if the build is not ready. But I can't tell from reading this whether this is saying "include it for progress reporting" or "include it for everything". Worth renaming the var to make it explicit.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -12,6 +12,78 @@ import (
...
 
+type ArtifactListing struct {
+	build            *model.Build
+	runtimeClosure   artifact.Map
+	buildtimeClosure artifact.Map
+	artifactIDs      []artifact.ArtifactID
+}
+
+func NewArtifactListing(build *model.Build) *ArtifactListing {
+	return &ArtifactListing{build: build}
```
Naatan, line 23: Would it be possible to calculate the closures and ID's during construction or that's too early since you don't have this context yet?

Naatan (changes requested):
I feel like this starts to address my concerns but leaves the actual business logic part where this new "ArtifactListing" is invoked still very implicit, which is where the hotspot for confusion is.

I'd suggest trying to update any symbol names inside the fetchAndInstall logic so that they explicitly state what they contain. You may find that through doing this you cannot name them all explicitly, and any for which this is true would need to be refactored or ensured through comments that the implicit nature is evident.

Particularly it's not clear to me reading through this code when we are dealing with buildtime closure for the purpose of progress reporting and buildtime closure for the purpose of installation. These are not one and the same but reading through it I am left with the impression that they are sometimes treated as such. Whether or not that's true isn't really the main issue, the main issue is that I can't easily tell.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -12,6 +12,78 @@ import (
...
+	if al.runtimeClosure != nil {
+		for _, artifact := range al.runtimeClosure {
+			al.artifactIDs = append(al.artifactIDs, artifact.ArtifactID)
+		}
+		return al.artifactIDs, nil
+	}
+
+	// Favor the buildtime closure over the runtime closure as it will
+	// include more artifact IDs
+	buildTimeClosure, err := al.BuildtimeClosure()
```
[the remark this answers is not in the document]
MDrakos: I'm confident, but not certain. Best to leave that up to the caller.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: Yeah, we can do that. In some cases, depending on what methods the caller uses we may calculate both but that shouldn't happen.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -1,19 +1,107 @@
...
+
+	return al.artifactIDs, nil
+}
+
 // NewMapFromBuildPlan creates an artifact map from a build plan. It creates a
 // lookup table and calls the recursive function buildMap to build up the
 // artifact map by traversing the build plan from the terminal targets through
 // all of the runtime dependencies for each of the artifacts in the DAG.
-func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
+
```
mitchell-as, line 99: This newline breaks up the documentation comment. Was that intentional? If not, please add a `//` prefix to this empty line. Also, the name has changed to `newMapFromBuildPlan`, so please update the comment on line 95.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -38,16 +131,53 @@ func NewMapFromBuildPlan(build *model.Build) (artifact.Map, error) {
...
+		for _, id := range terminalTargetIDs {
+			if err := buildBuildtimeClosureMap(id, lookup, res); err != nil {
+				return nil, errs.Wrap(err, "Could not build map for terminal %s", id)
+			}
+		}
+	} else {
+		for _, id := range terminalTargetIDs {
+			if err := buildRuntimeClosureMap(id, lookup, res); err != nil {
+				return nil, errs.Wrap(err, "Could not build map for terminal %s", id)
+			}
```
mitchell-as, line 144: Nit: you could probably assign a `buildMap` variable to be the `buildBuildtimeClosureMap` or `buildRuntimeClosureMap` runction based on an `if calculateBuildtimeClosure` statement and just have a single loop that calls `buildMap(...)`.

pkg/platform/runtime/setup/implementations/alternative/resolver.go
```diff
@@ -0,0 +1,21 @@
...
+
+func NewResolver(artifactsForNameResolving artifact.Map) *Resolver {
+	return &Resolver{artifactsForNameResolving: artifactsForNameResolving}
+}
+
+func (r *Resolver) ResolveArtifactName(id artifact.ArtifactID) string {
+	if artf, ok := r.artifactsForNameResolving[id]; ok {
+		return artf.Name
+	}
+	return locale.Tl("alternative_unknown_pkg_name", "unknown")
```
mitchell-as, line 20: This locale key is duplicating the one here: https://github.com/ActiveState/cli/pull/2772/files#diff-612cef18aaba9eaeff0a01319b8d2afad73dc05cb80015adce68d8c2bb2d851aR142. Please add it to en-us.yaml and use `locale.T()` to reference it here and there (https://github.com/ActiveState/cli/pull/2772/files#diff-612cef18aaba9eaeff0a01319b8d2afad73dc05cb80015adce68d8c2bb2d851aR142)

test/integration/checkout_int_test.go
```diff
@@ -262,6 +262,28 @@ func (suite *CheckoutIntegrationTestSuite) TestCheckoutCaseInsensitive() {
...
+
+	if runtime.GOOS != "linux" {
+		suite.T().Skip("Skipping buildtime closure test on non-linux platform")
+	}
+
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("checkout", "ActiveState-CLI/small-python"),
```
mitchell-as, line 276: We should probably checkout with a UUID. Otherwise this may fail as the small-python project is updated.

mitchell-as (changes requested):
Looks good to me overall. I just identified a few minor things to address. I will defer to your judgement that you've addressed Nathan's requested changes.

pkg/platform/runtime/buildplan/buildplan.go
MDrakos: Yeah, that was intentional as the two blocks are discussing different concepts related to the function.

test/integration/checkout_int_test.go
MDrakos: Good point!

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -131,17 +130,16 @@ func newMapFromBuildPlan(build *model.Build, calculateBuildtimeClosure bool) (ar
...
-			}
-		}
+		buildMap = buildBuildtimeClosureMap
 	} else {
-		for _, id := range terminalTargetIDs {
-			if err := buildRuntimeClosureMap(id, lookup, res); err != nil {
-				return nil, errs.Wrap(err, "Could not build map for terminal %s", id)
-			}
+		buildMap = buildRuntimeClosureMap
+	}
```
mitchell-as, line 138: I think you can simply use

```
buildMap := buildRuntimeClosureMap
if calculateBuildtimeClosure {
    buildMap = buildBuildtimeClosureMap
}
```
 and save some typing :)

mitchell-as (approved):

PR #2774, opened 2023-09-27
Build scripts should use buildexpression types.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1961" title="DX-1961" target="_blank">StoryDX-1961</a>  Unify BuildPlanner and Local Order Build Expression types
  </td></table>
  <br />
 

Initially build scripts and build expressions were developed in parallel. Now they should use common types.


pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -592,10 +592,6 @@ func (e *BuildExpression) getSolveNode() (*Ap, error) {
 			continue
 		}
 
-		if a.Assignment.Name == "" || a.Assignment.Name != "runtime" {
```
mitchell-as, line 595: I removed this code because it was causing a failure with this build script: https://github.com/ActiveState/cli/blob/878706429d6b1aa137258db5955faf8741bb7f23/pkg/platform/runtime/buildscript/buildscript_test.go#L80-L104

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -625,10 +620,6 @@ func recurseLets(let *Let) (*Ap, error) {
 			continue
 		}
 
-		if a.Name == "" || a.Name != "runtime" {
```
mitchell-as, line 628: Removed this check for the same reason above.

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -302,6 +302,10 @@ func newValue(path []string, valueInterface interface{}) (*Value, error) {
 	case float64:
 		value.Float = ptr.To(v)
 
+	case nil:
```
mitchell-as, line 305: This avoids the unnecessary logging on line 310. nil is an acceptable value.

mitchell-as:
Test failures are due to nightly failures from what I can tell. Logs seem to indicate success as well as local tests.

mitchell-as:
@MDrakos ping. I'd like to land this by the end of the day today.

pkg/platform/runtime/buildscript/buildscript_test.go
```diff
@@ -217,15 +240,16 @@ func TestExample(t *testing.T) {
...
 }
 
 func TestString(t *testing.T) {
 	script, err := NewScript([]byte(
 		`let:
     runtime = solve(
-        requirements=[{name="language/python"}],
-        platforms=["12345", "67890"]
+        platforms=["12345", "67890"],
+        requirements=[{name="language/python"}]
```
MDrakos, line 252: Was this order messing with the tests? If so then we need to allow for arbitrary order of solve arguments.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -2,18 +2,24 @@ package buildscript
...
 )
 
+// Script's tagged fields will be initially filled in by Participle.
+// Expr will be constructed later and is this script's buildexpression.
 type Script struct {
-	Let *Let `parser:"'let' ':' @@"`
-	In  *In  `parser:"'in' ':' @@"`
+	Let  *Let `parser:"'let' ':' @@"`
+	In   *In  `parser:"'in' ':' @@"`
+	Expr *buildexpression.BuildExpression
```
MDrakos, line 22: It's not immediately clear to me why we're now embedding the expression within the script but this feels a bit off. Ideally, these two implementations would be separate as well as the code that does the translations.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -153,3 +174,26 @@ func (f *FuncCall) String() string {
...
+		multilog.Error("Unable to marshal this buildscript to JSON: %v", err)
+		return false
+	}
+	otherJson, err := json.Marshal(expr)
+	if err != nil {
+		multilog.Error("Unable to marshal other buildscript to JSON: %v", err)
+		return false
+	}
+	return string(myJson) == string(otherJson)
+}
```
MDrakos, line 199: It appears that these functions are unused.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -62,44 +68,59 @@ func NewScript(data []byte) (*Script, error) {
...
+		return nil, errs.Wrap(err, "Could not construct build expression")
+	}
+	script.Expr = expr
+
 	return script, nil
 }
 
+func NewScriptFromBuildExpression(expr *buildexpression.BuildExpression) (*Script, error) {
+	return &Script{Expr: expr}, nil
+}
```
MDrakos, line 87: Shouldn't this also create the Script? Is there a potential that a user calling a script method could get an error because the struct only has an expression assigned?

pkg/platform/runtime/buildexpression/buildexpression.go
MDrakos: Does the function now just return the first solve node it finds? You are correct that we don't want `runtime = solve()` I believe it's there because we need a way to identify _one_ solve node that we can make modifications to. I recall this was how PB suggested we do it, but that could be incorrect or may have changed. Would you be able to follow up with them?

MDrakos (changes requested):

pkg/platform/runtime/buildscript/buildscript_test.go
mitchell-as: This is why I filed DX-2221 (Implement a canonical/stable sort for buildscript output.) It doesn't have anything to do with build expressions, just with the order that the build scripts are written in.

pkg/platform/runtime/buildscript/buildscript.go
mitchell-as: Originally I tried just that. However, I ended up writing a translator between participle-produced Go structs and build expression Go structs. That is very brittle and any changes to the build expression structs would need to be addressed in this package too -- it's one of the problems this ticket is trying to resolve.

One solution is to put participle tags into the build expressions and construct build scripts with that, but I figured build expressions should remain distinct from build scripts.

I found it's easiest to work directly with build expressions and JSON.

pkg/platform/runtime/buildscript/buildscript.go
mitchell-as: In this PR yes, but in the code base they are used in a few places.

pkg/platform/runtime/buildscript/buildscript.go
mitchell-as: Not yet. Any merges that need to take place will be done before writing to disk. That's how it was before, so I don't believe I've changed anything.

mitchell-as:
I will follow up with PB about your question. Until I hear back, I've responded to your initial feedback. Thanks! Let me know if you think I need to change something or add documentation, etc.

pkg/platform/runtime/buildexpression/buildexpression.go
mitchell-as: After talking to PB, multiple solve nodes are not supported, and will not be supported until the VCS migration is complete. I've reverted this change, disabled the failing test, and filed https://activestatef.atlassian.net/browse/DX-2238 to re-enable that test.

pkg/platform/runtime/buildscript/buildscript_test.go
```diff
@@ -73,10 +88,12 @@ in:
 			},
 		},
 		&In{Name: ptr.To("runtime")},
+		expr,
 	}, script)
 }
 
-func TestComplex(t *testing.T) {
+// Re-enable when multiple solves are supported in DX-2238.
+func NoTestComplex(t *testing.T) {
```
MDrakos, line 96: Sorry to be nitpicky but can we do this with a `t.Skip()` for now? It just makes it more explicit imo.

MDrakos (changes requested):
Looking good, just one very minor thing.

pkg/platform/runtime/buildscript/buildscript_test.go
mitchell-as: Good call!

MDrakos (approved):

PR #2776, opened 2023-09-28
User facing errors for State Checkout

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2008" title="DX-2008" target="_blank">StoryDX-2008</a>  User facing errors for state checkout
  </td></table>
  <br />


internal/errs/userfacing.go
```diff
@@ -27,19 +28,24 @@ func (e *userFacingError) ErrorTips() []string {
 	return e.tips
 }
 
+func (e *userFacingError) InputError() bool {
```
MDrakos, line 31: If I recall correctly from my previous PR we wanted to avoid user facing errors from colliding with existing error implementations. This is an implementation of the `ErrorInput` interface from the locale package.

internal/locale/locales/en-us.yaml
```diff
@@ -334,9 +334,7 @@ err_api_forbidden:
 err_api_org_not_found:
   other: Unable to find requested Organization
 err_api_project_not_found:
-  other: The requested project [NOTICE]{{.V0}}[/RESET] does not exist under [NOTICE]{{.V1}}[/RESET]. Please ensure the owner is correct and that the project has been created.
-err_api_project_not_found_unauthenticated:
-  other: The requested project [NOTICE]{{.V0}}/{{.V1}}[/RESET] could not be found
+  other: The requested project [NOTICE]{{.V1}}[/RESET] does not exist under [NOTICE]{{.V0}}[/RESET]. Please ensure the owner is correct and that the project has been created.
```
MDrakos, line 337: nit: Why are the variables here out of order?

internal/runbits/runtime/rationalize.go
```diff
@@ -0,0 +1,51 @@
...
+		*rerr = errs.Pack(*rerr, errs.New("project is nil"))
+
+	// Could not find a platform that matches on the given branch, so suggest alternate branches if ones exist
+	case isUpdateErr && errors.As(*rerr, &errNoMatchingPlatform):
+		branches, err := model.BranchNamesForProjectFiltered(proj.Owner(), proj.Name(), proj.BranchName())
+		if err == nil && len(branches) > 1 {
+			// Suggest alternate branches
+			*rerr = errs.NewUserFacingError(locale.Tr(
+				"err_alternate_branches",
+				errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch,
```
MDrakos, line 34: nit: While I see the value in adding this suggestion it also adds complexity where we could just provide a more generic, but still helpful, message.

internal/runbits/rationalize/types.go
```diff
@@ -0,0 +1,9 @@
+package rationalize
+
+// Inner is just an alias because otherwise external use of this struct would not be able to construct the error
+// property, and we want to keep the boilerplate minimal.
+type Inner error
+
+type ErrNoProject struct {
+	Inner
+}
```
MDrakos, line 9: Having centralized errors in a package called rationalize is a bit confusing. Is this because the error is used by the rationalized functions?

MDrakos (changes requested):
Overall this looks really good. Just a few points.

internal/errs/userfacing.go
Naatan: This isn't a collision. The problem I ran into is wanting to communicate an input failure on the user facing error side. So when I asserted the type I could set the user facing error, but also assert that it is due to user input.

If we use WrapInputError or something like that we add even more errors to the stack, and the API would be pretty awkward. This solves both those problems, you can now set a user facing error as an input error with `NewUserFacing(msg, SetInput(true))`

internal/runbits/runtime/rationalize.go
Naatan: This PR is just hoisting the existing error up to a user facing level. It's not seeking to make changes to the actual error.

internal/runbits/rationalize/types.go
Naatan: Rationalize is the term I want to use for this, so when you read "Rationalize" your mind should immediately make the connection.

"Centralized errors" isn't a term we ever used for this, it's just a generic way of thinking about this problem space. I want to be explicit about it.

Btw happy to explore other terms if you don't like "Rationalize". The intend is that it's both sensible and explicit.

internal/locale/locales/en-us.yaml
Naatan: To keep the order consistent in the code itself, see lines 43 and 46 here:

https://github.com/ActiveState/cli/blob/d657e19c76b2d44150c02d7327361ba21591ed00/pkg/platform/model/projects.go#L43-L46

internal/runbits/rationalize/types.go
MDrakos: I'd prefer not to get caught up in the naming. This makes enough sense to me.

MDrakos (approved):

PR #2781, opened 2023-10-03
Fix some update analytics.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2128" title="DX-2128" target="_blank">BugDX-2128</a>  Fix update metrics not firing
  </td></table>
  <br />
 

The install success event was not being fired, so I added that.

However,
- The download event is fired here via autoUpdate() -> InstallBlocking() -> prepareInstall(): https://github.com/ActiveState/cli/blob/30a336cb4185db84569a7fbab6dadee5c4ec4719/internal/updater/updater.go#L152
- The relaunch event is fired here via autoUpdate(): https://github.com/ActiveState/cli/blob/30a336cb4185db84569a7fbab6dadee5c4ec4719/cmd/state/autoupdate.go#L115-L117

What can prevent these events from firing successfully are:
- The "should-autoupdate" "disabled-config" event is fired: https://github.com/ActiveState/cli/blob/30a336cb4185db84569a7fbab6dadee5c4ec4719/cmd/state/autoupdate.go#L61-L67
- The "install" "failure" event is fired (e.g. insufficient permissions, update in progress, etc.): https://github.com/ActiveState/cli/blob/30a336cb4185db84569a7fbab6dadee5c4ec4719/cmd/state/autoupdate.go#L74-L95
- The "relaunch" "failure" event is fired: https://github.com/ActiveState/cli/blob/30a336cb4185db84569a7fbab6dadee5c4ec4719/cmd/state/autoupdate.go#L100-L113

I don't have permission to view the tableau report, but this is what I gleaned looking at the code.

### DIFF
--- a/cmd/state/autoupdate.go
+++ b/cmd/state/autoupdate.go
@@ -73,30 +73,14 @@ func autoUpdate(svc *model.SvcModel, args []string, cfg *config.Instance, an ana
 
 	err = up.InstallBlocking("")
 	if err != nil {
+		if errs.Matches(err, &updater.ErrorInProgress{}) {
+			return false, nil // ignore
+		}
 		if os.IsPermission(err) {
-			an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateInstall, anaConst.UpdateLabelFailed, &dimensions.Values{
-				TargetVersion: ptr.To(avUpdate.Version),
-				Error:         ptr.To("Could not update the state tool due to insufficient permissions."),
-			})
 			return false, locale.WrapInputError(err, locale.Tl("auto_update_permission_err", "", constants.DocumentationURL, errs.JoinMessage(err)))
 		}
-		if errs.Matches(err, &updater.ErrorInProgress{}) {
-			an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateInstall, anaConst.UpdateLabelFailed, &dimensions.Values{
-				TargetVersion: ptr.To(avUpdate.Version),
-				Error:         ptr.To(anaConst.UpdateErrorInProgress),
-			})
-			return false, nil
-		}
-		an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateInstall, anaConst.UpdateLabelFailed, &dimensions.Values{
-			TargetVersion: ptr.To(avUpdate.Version),
-			Error:         ptr.To(anaConst.UpdateErrorInstallFailed),
-		})
 		return false, locale.WrapError(err, locale.T("auto_update_failed"))
 	}
-
-	an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateInstall, anaConst.UpdateLabelSuccess, &dimensions.Values{
-		TargetVersion: ptr.To(avUpdate.Version),
-	})
 
 	out.Notice(locale.Tr("auto_update_relaunch"))
 	out.Notice("") // Ensure output doesn't stick to our messaging
--- a/cmd/state/autoupdate.go
+++ b/cmd/state/autoupdate.go
@@ -94,6 +94,10 @@ func autoUpdate(svc *model.SvcModel, args []string, cfg *config.Instance, an ana
 		return false, locale.WrapError(err, locale.T("auto_update_failed"))
 	}
 
+	an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateInstall, anaConst.UpdateLabelSuccess, &dimensions.Values{
+		TargetVersion: ptr.To(avUpdate.Version),
+	})
+
 	out.Notice(locale.Tr("auto_update_relaunch"))
 	out.Notice("") // Ensure output doesn't stick to our messaging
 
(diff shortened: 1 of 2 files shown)


mitchell-as:
The relaunch event fires here: https://github.com/ActiveState/cli/blob/30a336cb4185db84569a7fbab6dadee5c4ec4719/cmd/state/autoupdate.go#L115-L117

cmd/state/autoupdate.go
```diff
@@ -94,6 +94,10 @@ func autoUpdate(svc *model.SvcModel, args []string, cfg *config.Instance, an ana
 		return false, locale.WrapError(err, locale.T("auto_update_failed"))
 	}
 
+	an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateInstall, anaConst.UpdateLabelSuccess, &dimensions.Values{
+		TargetVersion: ptr.To(avUpdate.Version),
+	})
```
Naatan, line 99: The install action should only be invoked by the installer. You are signaling that the autoupdate was successful here, but the event is intended for the installer being successful, regardless of autoupdate.

internal/updater/updater.go
```diff
@@ -268,5 +268,5 @@ func (u *UpdateInstaller) analyticsEvent(action, label, version, msg string) {
 		dims.Error = ptr.To(msg)
 	}
 
-	u.an.EventWithLabel(anaConst.CatUpdates, anaConst.ActUpdateDownload, label, dims)
```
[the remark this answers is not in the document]
Naatan: Great catch! This sounds like it could likely be the cause of the issue I was seeing.

Naatan (changes requested):

cmd/state/autoupdate.go
mitchell-as: Okay, I've moved all install/success failure analytics into the installer itself.

internal/updater/updater.go
```diff
@@ -173,9 +173,24 @@ func (u *UpdateInstaller) prepareInstall(installTargetPath string, args []string
 	return installerPath, args, nil
 }
 
-func (u *UpdateInstaller) InstallBlocking(installTargetPath string, args ...string) error {
+func (u *UpdateInstaller) InstallBlocking(installTargetPath string, args ...string) (rerr error) {
```
Naatan, line 176: `InstallBlocking` only addresses the `state update` flow IIRC. For eg. `install.sh` it uses `InstallWithProgress`, so either we need the same changes there or we need to handle these at a higher level.

Naatan (changes requested):

internal/updater/updater.go
mitchell-as: @Naatan I have been operating under that assumption that we're lacking in update events for install, download, etc., **not** bare install events. The document link in the ticket links to the "Updates" section.

Naatan: @mitchell-as that was the original reason for enriching a lof of these events, yeah. But that doesn't change the fact that the events in question are for the installer. We need to be able to rely on these events behaving as we expect them to. Having them only fire on certain use-cases would violate that expectation.

mitchell-as:
Based on our zoom call, it seems like this PR is okay as-is.

internal/updater/updater.go
```diff
@@ -259,14 +276,14 @@ func (u *UpdateInstaller) InstallWithProgress(installTargetPath string, progress
 	return proc, nil
 }
 
-func (u *UpdateInstaller) analyticsEvent(action, label, version, msg string) {
+func (u *UpdateInstaller) analyticsEvent(action, label, msg string) {
 	dims := &dimensions.Values{
-		TargetVersion: ptr.To(version),
+		TargetVersion: ptr.To(u.AvailableUpdate.Version),
```
Naatan, line 281: This probably needs a nil check. eg. the `ActUpdateCheck` event won't have an available update yet.

Naatan (changes requested):

internal/updater/updater.go
mitchell-as: `ActUpdateCheck` happens in the `checker.go` file as individual events: https://github.com/ActiveState/cli/blob/mitchell/dx-2128/internal/updater/checker.go#L122 and https://github.com/ActiveState/cli/blob/mitchell/dx-2128/internal/updater/checker.go#L139. This code happens in an `UpdateInstaller` method: https://github.com/ActiveState/cli/blob/mitchell/dx-2128/internal/updater/updater.go#L279, whose `AvailableUpdate` member is never `nil` from what I can see. Furthermore, this change is just a simplification, as all calls to `analyticsEvent()` were already dereferencing `u.AvailableUpdate`, e.g. https://github.com/ActiveState/cli/pull/2781/files/5f4ebc6c75ce6f701be009e7a111f26969cba933#diff-07f86aa363d0be00f9de8b4134d96188f3923b9557bc3428c65c8541027ea68dL135

Naatan: That's fair, but all the same for future compatibility I'd rather we nil check this just to be safe. The function this is implemented in could in the future be used for events that may not have the AvailableUpdate set.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/2781#discussion_r1353481923

Naatan (approved):

PR #2782, opened 2023-10-03
Clarify why we normalize terminal line endings to '\n'.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2169" title="DX-2169" target="_blank">BugDX-2169</a>  Figure out why termtest is using Windows line endings
  </td></table>
  <br />
 

~~When I commented out the code as instructed in the ticket, I got test passes on macOS and Linux (was not able to reproduce the original failure), but failures on Windows.~~

~~It is quite possible that the original, multi-line expectation string in TestBranch_List was silently using CR+LF line endings. (That is, the test was originally written and committed on Windows with CR+LF line endings.) Switching to a quoted string with newline (LF) escapes silently fixed the problem on macOS and Linux, but still requires a fix for Windows, which uses CR+LF.~~

Long story short: terminal emulators are emulating old TTYs, which emit CR+LF for newlines. You can read more [here](https://unix.stackexchange.com/questions/343324/why-in-the-output-of-script-1-the-newline-is-cr-lf-dos-style) and [here](https://superuser.com/questions/714078/wrong-newline-character-over-serial-port-cr-instead-of-lf).

### DIFF
--- a/internal/testhelpers/e2e/session.go
+++ b/internal/testhelpers/e2e/session.go
@@ -268,6 +268,7 @@ func (s *Session) SpawnCmdWithOpts(exe string, optSetters ...SpawnOptSetter) *Sp
 	// raw terminal output will contain "\r\n". Since our multi-line expectation messages often use
 	// '\n', normalize line endings to that for convenience, regardless of platform ('\n' for Linux
 	// and macOS, "\r\n" for Windows).
+	// More info: https://superuser.com/a/1774370
 	spawnOpts.TermtestOpts = append(spawnOpts.TermtestOpts,
 		termtest.OptNormalizedLineEnds(true),
 	)
--- a/internal/testhelpers/e2e/session.go
+++ b/internal/testhelpers/e2e/session.go
@@ -263,9 +263,9 @@ func (s *Session) SpawnCmdWithOpts(exe string, optSetters ...SpawnOptSetter) *Sp
 		termtest.OptRows(30), // Needs to be able to accommodate most JSON output
 	)
 
-	// Work around issue where multiline values sometimes have the wrong line endings
-	// See for example TestBranch_List
-	// https://activestatef.atlassian.net/browse/DX-2169
+	// Multi-line expectation strings (e.g. TestBranch_List), as well as macOS and Linux, use '\n' to
+	// represent newlines. However, Windows uses '\r\n' to represent newlines. Normalize all line
+	// endings to be just '\n' or else the expectation will fail on Windows.
 	spawnOpts.TermtestOpts = append(spawnOpts.TermtestOpts,
 		termtest.OptNormalizedLineEnds(true),
 	)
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
It's not that simple I think. When I ran into this I was getting \r\n line endings on macOS, not Windows. I'm not sure why, that's why I worked around it and filed the bug.

mitchell-as:
I understand, but the linked code in the ticket no longer exists: https://github.com/ActiveState/cli/blob/3d8ed79520c715a5c1f1a642541e7c78e8524aeb/internal/testhelpers/e2e/session.go#L260-L260. It has been changed in the meantime. I tried to reproduce on a Mac (as noted earlier) and it worked. It also worked on CI for both Linux and macOS. It only failed on Windows due to Windows using \r\n and the test expecting \n.

Naatan (changes requested):
Please try to reproduce with the linked code. I understand the code has changed, but you should still be able to disable the current line ending sanitization mechanic, and you can bring back the original test for the purposes of this bug. The changes were made to work around this very bug, so we can't abandon it because it's "no longer happening".

The current solution may be fine, but I didn't have time to dig into the core issue when I was originally working on this, hence the workaround. If we can determine that this workaround is in fact an appropriate solution then that works for me, but it seems we haven't concluded that that's the case.

mitchell-as:
Looks like I was wrong. After restoring the old code, I reproduced the issue and did some research. Long story short: terminal emulators are emulating old TTYs, which emit CR+LF for newlines. You can read more [here](https://unix.stackexchange.com/questions/343324/why-in-the-output-of-script-1-the-newline-is-cr-lf-dos-style) and [here](https://superuser.com/questions/714078/wrong-newline-character-over-serial-port-cr-instead-of-lf)

internal/testhelpers/e2e/session.go
```diff
@@ -263,9 +263,11 @@ func (s *Session) SpawnCmdWithOpts(exe string, optSetters ...SpawnOptSetter) *Sp
...
 	)
 
-	// Work around issue where multiline values sometimes have the wrong line endings
-	// See for example TestBranch_List
-	// https://activestatef.atlassian.net/browse/DX-2169
+	// TTYs output newlines in two steps: '\r' (CR) to move the caret to the beginning of the line,
+	// and '\n' (LF) to move the caret one line down. Terminal emulators do the same thing, so the
+	// raw terminal output will contain "\r\n". Since our multi-line expectation messages often use
+	// '\n', normalize line endings to that for convenience, regardless of platform ('\n' for Linux
+	// and macOS, "\r\n" for Windows).
```
Naatan, line 270:
```suggestion
	// TTYs output newlines in two steps: '\r' (CR) to move the caret to the beginning of the line,
	// and '\n' (LF) to move the caret one line down. Terminal emulators do the same thing, so the
	// raw terminal output will contain "\r\n". Since our multi-line expectation messages often use
	// '\n', normalize line endings to that for convenience, regardless of platform ('\n' for Linux
	// and macOS, "\r\n" for Windows).
	// More info: https://superuser.com/a/1774370
```

Naatan (changes requested):
Nice digging! Just wanna add the link you found since it has really useful context.

Naatan (approved):

PR #2785, opened 2023-10-03
Installer should check for an xml file by comparing the first few bytes instead of depending on `file`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2212" title="DX-2212" target="_blank">StoryDX-2212</a>  We should not depend on the `file` utility to install the State Tool
  </td></table>
  <br />


mitchell-as:
Test failures are due to timeouts in runtime setup, not the install.sh command.

Naatan (approved):

PR #2786, opened 2023-10-03
Revert: Update shell rcfiles to indicate the user is already in an activated state (v42).

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2239" title="DX-2239" target="_blank">StoryDX-2239</a>  Remove virtual environment indication
  </td></table>
  <br />
 

It's broken :(

### DIFF
--- a/internal/assets/contents/shells/bashrc_append.sh
+++ b/internal/assets/contents/shells/bashrc_append.sh
@@ -6,7 +6,4 @@ export {{$K}}="{{$V}}:$PATH"
 export {{$K}}="{{$V}}"
 {{- end}}
 {{- end}}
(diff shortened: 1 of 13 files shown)


mitchell-as:
#2787 will fail to propagate to v42 due to a conflict in shell_int_test.go. The resolution is to remove the TestNestedShellNotification() function, but leave the existing TestRuby() function. See https://github.com/ActiveState/cli/pull/2786.

mitchell-as:
Test failures are due to https://github.com/ActiveState/cli/pull/2788, which is not related to this merge.

PR #2790, opened 2023-10-04
Solve node does not require runtime variable

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2242" title="DX-2242" target="_blank">BugDX-2242</a>  Unable to parse new build exprs using solve and a build function
  </td></table>
  <br />


Naatan (commented):
Flagged Martin for review instead as he'll understand the implications of this change.

mdrohmann (approved):
Yes, that looks correct to me.  Just look for `* = solve` or `* = solve_legacy`, where `*` could be any variable name, not just `runtime`.

:+1:

PR #2792, opened 2023-10-04
Add integration tests for project with complex package version requirements

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2241" title="DX-2241" target="_blank">BugDX-2241</a>  UNINSTALL: ALL ST VERSIONS install/uninstall packages fail due build <nil>
  </td></table>
  <br />


MDrakos:
Failing Windows integration tests are not related to this PR change.

Naatan (changes requested):
This looks good, but my one gripe is I can't tell from the test itself what's being tested.

I'd recommend either adding comments or updating the tests themselves so it's clear what is being tested for.

Point being I don't know what I would do if this test started failing and/or needed updates.

MDrakos:
I tried adding an import step with the requirements data that I used to set up the project but I kept running into timeout issues due to the project building. Instead, I've added a comment that hopefully helps understand what the test is doing. We could still run into timeout issues but it doesn't appear to be happening with the current project and test configuration.

The other windows failures are unrelated to the PR.

Naatan (changes requested):
I don't think this is testing what we want to test. We want to test that we can eg. `state import` a project with the given version constraints. Testing a project that was previously created isn't valid because it is the creation that is what we want to test.

MDrakos:
What we want to test is that we can make modifications to a build expression that has complex version requirements, submit that, and get the expected response. The test I added mirrors how the bug surfaced. That said, the test you propose is also something worth having so I've added it.

Note that for the new test, I'm not verifying the version numbers because of the bug I filed earlier today: https://activestatef.atlassian.net/browse/DX-2272

test/integration/package_int_test.go
```diff
@@ -701,6 +701,86 @@ func (suite *PackageIntegrationTestSuite) TestProjectWithOfflineInstallerAndDock
...
+	cp.ExpectExitCode(0)
+
+	// Test version numbers once the following story is resolved:
+	// https://activestatef.atlassian.net/browse/DX-2272
+	cp = ts.Spawn("packages")
+	cp.Expect("Mopidy-Dirble")
+	cp.Expect("coverage")
+	cp.Expect("docopt")
+	cp.Expect("requests")
+	cp.Expect("urllib3")
```
Naatan, line 740: Since you can't use `state packages` and since verifying the version numbers is the point of this tests maybe instead we can use `state history`?

I don't see any point in adding the test as is because again it is not testing what it is set out to test.

test/integration/package_int_test.go
```diff
@@ -701,6 +701,86 @@ func (suite *PackageIntegrationTestSuite) TestProjectWithOfflineInstallerAndDock
...
+
+	cp = ts.Spawn("uninstall", "Mopidy-Dirble")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("uninstall", "requests")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("uninstall", "urllib3")
+	cp.ExpectExitCode(0)
+}
```
Naatan, line 782: With your latest explanation I think I am caught up. This test should not exist. We are testing an API now, rather than the state tool. That's not our job, we have a hard enough time keeping our tests under control with just covering the State Tool itself.

The new test you added has value but does seem unrelated to the story at hand. Perhaps we should abandon the ticket and file a new one for the `import` test.

Naatan (changes requested):

MDrakos:
Closing in favor of: https://activestatef.atlassian.net/browse/DX-2276

PR #2794, opened 2023-10-05
Fix failing integration tests due to expect mismatch

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2247" title="DX-2247" target="_blank">BugDX-2247</a>  Nightly failures due to outdated expect string
  </td></table>
  <br />


MDrakos:
The test failures are not related to the tests fixed in this PR. Some of them are due to projects building on the platform and other reasons

Naatan (approved):

PR #2797, opened 2023-10-06
Increase error logging on runtime panics

Naatan:


MDrakos (approved):
Very odd... but yes this additional data should help us. I believe the panic in the test was from the `EnvDef`. It might also be useful to dump that data on panic as well.

Naatan:
> Very odd... but yes this additional data should help us. I believe the panic in the test was from the `EnvDef`. It might also be useful to dump that data on panic as well.

Good idea. It indeed appears to be on EnvDef, but I can't find any reason why it would be nil.

MDrakos (approved):

PR #2800, opened 2023-10-06
Drop vscode tests as we no longer publish the extension

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2253" title="DX-2253" target="_blank">BugDX-2253</a>  TestPushIntegrationTestSuite/TestInitAndPush_VSCode Nightly Failure
  </td></table>
  <br />
 

While user facing errors surfaced this test failure, it was not what actually caused the failure. The actual cause of the failure is that `state init` now also does an implicit push, and the test in question was assuming it did not. As a result when it ran `state push` it got an error saying there was nothing to push.

There's no point fixing these tests though. They are testing something that we no longer support, so I dropped them instead.

I have no idea why this test failure didn't pop up sooner, the implicit push on `state init` has been around for quite a while now.


mitchell-as:
If the goal is to remove legacy VSCode support, then please also:

1. Remove https://github.com/ActiveState/cli/blob/8e2b6d23a93e4f2a18ea172e72f68d4668cbd845/cmd/state/internal/cmdtree/cmdtree.go#L282-L287
2. Address https://github.com/ActiveState/cli/blob/c260bb5c3eb7bbf44f19b5bd189b42f0b67f0198/cmd/state/main.go#L111-L120
3. Remove https://github.com/ActiveState/cli/blob/8e2b6d23a93e4f2a18ea172e72f68d4668cbd845/internal/testhelpers/tagsuite/tagsuite.go#L72

mitchell-as (changes requested):
If the goal is to remove legacy VSCode support, then please also:

1. Remove https://github.com/ActiveState/cli/blob/8e2b6d23a93e4f2a18ea172e72f68d4668cbd845/cmd/state/internal/cmdtree/cmdtree.go#L282-L287
2. Address https://github.com/ActiveState/cli/blob/c260bb5c3eb7bbf44f19b5bd189b42f0b67f0198/cmd/state/main.go#L111-L120
3. Remove https://github.com/ActiveState/cli/blob/8e2b6d23a93e4f2a18ea172e72f68d4668cbd845/internal/testhelpers/tagsuite/tagsuite.go#L72

Naatan:
@mitchell-as thanks for spotting! Since that is testing a flag that's still implemented I don't feel comfortable dropping it for this PR, but I filed this followup story:

https://activestatef.atlassian.net/browse/DX-2262

I've dropped the vscode tag.

mitchell-as (approved):

PR #2803, opened 2023-10-09
Allow bare version numbers to be passed to install scripts.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1979" title="DX-1979" target="_blank">StoryDX-1979</a>  As a user I can install a specific State Tool version without knowing its SHA
  </td></table>
  <br />
 

I made copies of S3 release folders from v33 to v40 with just the bare version names, and updated the `info.json` files for each arch to point to the latest bare version folder.

I changed the paths `info.json` files in the 0.40.1 directory to use the bare directory name, but left the paths alone for other releases because (1) it's a lot of work to manually update those json files and reset permissions to be world-readable, and (2) we're keeping the SHA folders around for compatibility. My local testing has shown that everthing seems to work for both older versions and newer versions.

Finally, I double-checked this PR's deployment bucket in S3 and the folder just has the version number (not SHA) and its `info.json` files point to the correct folder and archives per architecture.

### DIFF
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -114,15 +114,18 @@ function error([string] $msg)
 }
 
 if (!$script:VERSION) {
-    # Determine the latest version to fetch.
+    # If the user did not specify a version, formulate a query to fetch the JSON info of the latest
+    # version, including where it is.
     $jsonURL = "$script:BASEINFOURL/?channel=$script:CHANNEL&platform=windows&source=install"
 } elseif (!($script:VERSION | Select-String -Pattern "-SHA" -SimpleMatch)) {
-    # Determine the full version SHA to fetch.
+    # If the user specified a partial version (i.e. no SHA), formulate a query to fetch the JSON
+    # info of that version's latest SHA, including where it is.
     $jsonURL = "$script:BASEINFOURL/?channel=$script:CHANNEL&platform=windows&source=install&target-version=$script:VERSION"
 }
 
 if ($jsonURL) {
-  # Parse info.
+    # If the user specified no version or a partial version we need to use the json URL to get the
+    # actual installer URL.
     try {
         $infoJson = ConvertFrom-Json -InputObject (download $jsonURL)
     } catch [System.Exception] {
@@ -140,6 +143,8 @@ if ($jsonURL) {
     $checksum = $infoJson.Sha256
     $relUrl = $infoJson.Path
 } else {
+    # If the user specified a full version, strip the SHA to get the folder name of the installer
+    # URL. Then we can construct the installer URL.
     $versionNoSHA = $script:VERSION -replace "-SHA.*", ""
     $relUrl = "$script:CHANNEL/$versionNoSHA/windows-amd64/state-windows-amd64-$script:VERSION.zip"
 }
--- a/installers/install.ps1
+++ b/installers/install.ps1
@@ -114,14 +114,34 @@ function error([string] $msg)
 }
 
 if (!$script:VERSION) {
-  # Determine the latest version to fetch and parse info.
-  $jsonURL = "$script:BASEINFOURL/?channel=$script:CHANNEL&platform=windows&source=install"
-  $infoJson = ConvertFrom-Json -InputObject (download $jsonURL)
-  $version = $infoJson.Version
-  $checksum = $infoJson.Sha256
-  $relUrl = $infoJson.Path
+    # Determine the latest version to fetch.
+    $jsonURL = "$script:BASEINFOURL/?channel=$script:CHANNEL&platform=windows&source=install"
+} elseif (!($script:VERSION | Select-String -Pattern "-SHA" -SimpleMatch)) {
+    # Determine the full version SHA to fetch.
+    $jsonURL = "$script:BASEINFOURL/?channel=$script:CHANNEL&platform=windows&source=install&target-version=$script:VERSION"
+}
+
+if ($jsonURL) {
+  # Parse info.
+    try {
+        $infoJson = ConvertFrom-Json -InputObject (download $jsonURL)
+    } catch [System.Exception] {
+    }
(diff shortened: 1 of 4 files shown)


mitchell-as:
Linux failure is due to https://activestatef.atlassian.net/browse/DX-2261, Windows failure is due to timeout, and macOS failure is due to some weirdness around having no command output: https://github.com/ActiveState/cli/actions/runs/6461317571/job/17540739169?pr=2803#step:27:597. It passes locally though.

test/integration/install_scripts_int_test.go
```diff
@@ -185,8 +186,10 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall_VersionDoesNotExist
 	} else {
 		cp = ts.SpawnCmdWithOpts("powershell.exe", e2e.OptArgs(args...), e2e.OptAppendEnv("SHELL="))
 	}
-	cp.Expect("Could not download")
-	cp.Expect("does-not-exist")
```
mitchell-as, line 189: This line removal was intentional, because not all download failures are due to version errors (invalid branch could be specified for example).

Naatan (changes requested):
I'm not sure I understand what the changes to the install.sh and install.ps1 are for. It seems you implemented support to allow both passing a version specifier with, and without the SHA? 

I can see from the ACs that this comes into play for non-release branches, where we want to facilitate specifying the SHA as for PRs those are relevant. But that's something that should be entirely addressed in the update-generator. The install.sh should just take the version the user gave verbatim, and if it can't find the info.json then they specified a wrong version.

So I'm not sure why there would be any changes to the install.sh or install.ps1 for this PR. But I would expect to see some channel awareness in the update-generator to handle the difference in utilizing the SHA when not on release branches.

mitchell-as:
> I'm not sure I understand what the changes to the install.sh and install.ps1 are for. It seems you implemented support to allow both passing a version specifier with, and without the SHA?

Yes. Prior to this PR, we had this: https://github.com/ActiveState/cli/pull/2803/files#diff-ee220484bf5d245fa7e0f17cf3faf1f00b5b78f13fc7355e365e952bd6ce511aL129
If you give `./install.sh -v 0.40.0`, it will expand to something like "release/0.40.0/windows-amd64/state-windows-amd64-0.40.0.zip", which is wrong since there is no SHA. We need to lookup the SHA in `info.json` for a bare version number.

While I was at it, I figured we may want to support `-v 0.40.0-SHA0000000` as well for the sake of compatibility. If we do not, I can remove it. We will still need most of the existing changes for `info.json` lookup purposes.

> So I'm not sure why there would be any changes to the install.sh or install.ps1 for this PR. But I would expect to see some channel awareness in the update-generator to handle the difference in utilizing the SHA when not on release branches.

The key takeaway is that for a bare version number, we need an `info.json` lookup to get the SHA. That requires install.sh and install.ps1 changes.

I hope this makes sense. If not, we can talk on zoom.

installers/install.ps1
```diff
@@ -114,14 +114,34 @@ function error([string] $msg)
...
-  $checksum = $infoJson.Sha256
-  $relUrl = $infoJson.Path
+    # Determine the latest version to fetch.
+    $jsonURL = "$script:BASEINFOURL/?channel=$script:CHANNEL&platform=windows&source=install"
+} elseif (!($script:VERSION | Select-String -Pattern "-SHA" -SimpleMatch)) {
+    # Determine the full version SHA to fetch.
+    $jsonURL = "$script:BASEINFOURL/?channel=$script:CHANNEL&platform=windows&source=install&target-version=$script:VERSION"
+}
+
+if ($jsonURL) {
```
Naatan, line 124:
```suggestion
// If the user specified no version or a partial version we need to use the json URL to get the actual installer URL.
if ($jsonURL) {
```

installers/install.ps1
```diff
@@ -114,14 +114,34 @@ function error([string] $msg)
...
+      Write-Error $_.Exception.Message
+      exit 1
+    }
+    $version = $infoJson.Version
+    $checksum = $infoJson.Sha256
+    $relUrl = $infoJson.Path
 } else {
-  $relUrl = "$script:CHANNEL/$script:VERSION/windows-amd64/state-windows-amd64-$script:VERSION.zip"
+    $versionNoSHA = $script:VERSION -replace "-SHA.*", ""
+    $relUrl = "$script:CHANNEL/$versionNoSHA/windows-amd64/state-windows-amd64-$script:VERSION.zip"
```
Naatan, line 144: This made me realize we don't verify the checksum in this scenario, but that's not a bug introduced in this PR so filed another story: https://activestatef.atlassian.net/browse/DX-2266

Naatan (changes requested):

installers/install.sh
```diff
@@ -112,8 +112,18 @@ mkdir -p "$INSTALLERTMPDIR"
...
   # Determine the latest version to fetch.
-  STATEURL="$BASE_INFO_URL?channel=$CHANNEL&source=install&platform=$OS"
-  $FETCH $INSTALLERTMPDIR/info.json $STATEURL || exit 1
+  JSONURL="$BASE_INFO_URL?channel=$CHANNEL&source=install&platform=$OS"
+elif [ -z "`echo $VERSION | grep -o '\-SHA'`" ]; then
+  # Determine the full version SHA to fetch.
+  JSONURL="$BASE_INFO_URL?channel=$CHANNEL&source=install&platform=$OS&target-version=$VERSION"
+fi
+
+if [ ! -z "$JSONURL" ]; then
```
Naatan, line 121: Could you add comments on the conditionals to explain what they are? I find these take me a while to interpret.

mitchell-as: Good idea.

Naatan (approved):

PR #2809, opened 2023-10-10
Implement analytics for executor exit code.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2166" title="DX-2166" target="_blank">TaskDX-2166</a>  We track executor exit codes
  </td></table>
  <br />

### DIFF
--- a/internal/analytics/client/sync/client.go
+++ b/internal/analytics/client/sync/client.go
@@ -212,9 +212,13 @@ func (a *Client) EventWithSourceAndLabel(category, action, source, label string,
 	}
 
 	a.customDimensions.Sequence = ptr.To(a.sequence)
-	a.sequence++
 
 	actualDims := mergeDimensions(a.customDimensions, dims...)
+
+	if a.sequence == *actualDims.Sequence {
+		// Increment the sequence number unless dims overrides it (e.g. heartbeats use -1).
+		a.sequence++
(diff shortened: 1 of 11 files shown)


internal/analytics/client/sync/client.go
```diff
@@ -212,10 +212,14 @@ func (a *Client) EventWithSourceAndLabel(category, action, source, label string,
...
 
 	a.customDimensions.Sequence = ptr.To(a.sequence)
-	a.sequence++
 
 	actualDims := mergeDimensions(a.customDimensions, dims...)
 
+	if a.sequence == *actualDims.Sequence {
+		// Increment the sequence number unless dims overrides it (e.g. heartbeats use -1).
+		a.sequence++
+	}
```
mitchell-as, line 221: This is needed to pass https://github.com/ActiveState/cli/blob/3f28923089bf7c2ad43b0ab806631d64162684e3/test/integration/analytics_int_test.go#L208. Heartbeat events with a `-1` sequence were needlessly triggering an increment of `a.sequence` here, resulting in seemingly non-sequential events.

Naatan: Nice catch!

Naatan (approved):
Looks great!

PR #2812, opened 2023-10-11
Update update integration test to not expect SHA in folder name.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2267" title="DX-2267" target="_blank">BugDX-2267</a>  Nightly failure: TestUpdateGenIntegrationTestSuite/TestUpdateBits
  </td></table>
  <br />
 

The SHA is only in the file name.

### DIFF
--- a/scripts/ci/update-generator/main.go
+++ b/scripts/ci/update-generator/main.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"strings"
 
 	"github.com/mholt/archiver"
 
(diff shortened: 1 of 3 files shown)


mitchell-as:
Windows failure is due to timeout, not this PR.

mitchell-as:
Thanks for the tip. I found out we already have `constants.VersionNumber`, so I'm using that instead of stripping SHA from `constants.Version`. I've double-checked on S3 and the update generator/deployer is still using bare version numbers.

Also, the Linux test failure is not due to this PR and is being addressed in another ticket.

Naatan (approved):

PR #2814, opened 2023-10-12
Support a master State Tool versions list

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1960" title="DX-1960" target="_blank">StoryDX-1960</a>  As a devops engineer I can find links for State Tool install archives and their checksums
  </td></table>
  <br />
 

The master version list is located here: https://state-tool.s3.amazonaws.com/update/state/versions.json

I tested this on CI temporarily allowing for PR releases to deploy and it worked. The version list was updated here: https://github.com/ActiveState/cli/actions/runs/6499308092/job/17652710058?pr=2814#step:12:30 and uploaded to S3 here: https://github.com/ActiveState/cli/actions/runs/6499308092/job/17652710058?pr=2814#step:13:53. I double-checked the file on S3 and it was indeed updated and the links valid. Since then I've reverted to "release" and "beta" channel versions starting at v0.40.0.

I've excluded "master" channel from this list because it would be unnecessary pollution in my opinion. Let me know if I should include them. Only "beta", "release", and "LTS" channels will show up in the new master list.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -415,6 +415,10 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
 
+      - # === Install State Tool ===
+        name: Install State Tool
+        uses: ActiveState/setup-state-tool@v1
+
       - # === Setup ===
         name: Setup
         shell: bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -401,7 +401,6 @@ jobs:
       SHELL: bash
       GITHUB_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     timeout-minutes: 10
-    if: contains(fromJSON('["refs/heads/master", "refs/heads/beta", "refs/heads/release", "refs/heads/LTS"]'), github.ref) || startsWith(github.event.pull_request.head.ref, 'version/')
 
     # === Deploy Steps ===
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -465,6 +465,11 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
           mask-aws-account-id: true
 
+      - # === Generate updated master versions.json if necessary ===
+        name: Generate version list
+        shell: bash
+        run: state run generate-versions-list
+
       - # === Deploy ===
         name: Deploy
         shell: bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -401,6 +401,7 @@ jobs:
       SHELL: bash
       GITHUB_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     timeout-minutes: 10
+    if: contains(fromJSON('["refs/heads/master", "refs/heads/beta", "refs/heads/release", "refs/heads/LTS"]'), github.ref) || startsWith(github.event.pull_request.head.ref, 'version/')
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -416,6 +416,10 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
 
+      - # === Install State Tool ===
+        name: Install State Tool
+        uses: ActiveState/setup-state-tool@v1
```
Naatan, line 421: Since the workflow for generating the version list file is pretty dependent on CI I'd say let's move the scripts from the activestate.yaml into the build.yml, so we don't need to install the State Tool and thus slow down deployments.

activestate.generators.yaml
```diff
@@ -120,3 +120,22 @@ scripts:
...
+  - name: generate-test-versions-list
+    language: bash
+    description: Generates test versions.json
+    value: |
+      set -e
+      $constants.SET_ENV
+
+      $scripts.generate-test-update
+
+      go run scripts/ci/update-version-list/main.go ./build/test-update
```
Naatan, line 141: I think we can drop this considering it's not used, and if you want to test the script yourself you can just run it directly (ie. this isn't solving some big testing hurdle if I understand correctly).

scripts/ci/update-version-list/main.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/ActiveState/cli/internal/httputil"
+)
+
+// Where the master version file lives on S3.
+const S3PrefixURL = "https://state-tool.s3.amazonaws.com/"
+const S3Bucket = "update/state/"
+const VersionsJson = "versions.json"
+
+// Valid channels to update the master version file with.
+var ValidChannels = []string{"beta", "release", "LTS"}
```
Naatan, line 24: You could use the constants here:

https://github.com/ActiveState/cli/blob/a6674251904f1bbbc8ee3fd8798ff8532117dceb/internal/constants/constants.go#L231-L231

Also suggest dropping LTS as it never went into use.

scripts/ci/update-version-list/main.go
```diff
@@ -0,0 +1,97 @@
...
+		if !funk.Contains(ValidChannels, channel) {
+			continue
+		}
+		fmt.Printf("Found %s\n", file.RelativePath())
+		bytes, err := fileutils.ReadFile(file.Path())
+		if err != nil {
+			log.Fatalf("Unable to read file: %s", err.Error())
+		}
+		info := map[string]string{}
+		err = json.Unmarshal(bytes, &info)
```
Naatan, line 72: We should link the types used to generate and use the info so we don't accidentally break things in the future. Also to eliminate unmarshalling to a key/value map; I'd like to rely on Go's type system.

Naatan (changes requested):

.github/workflows/build.yml
mitchell-as: State tool is required for deployments. It's not just for the new feature. If you look at the build.yml file, you'll see a lot of other uses of ST (e.g. `state run install-deps`, `state run preprocess -v`, `state run deploy-updates`, etc.)

I added this block because it was missed in #2784.

Naatan: Gotcha, ok no worries that works then. I thought we were introducing it just for this story.

Naatan (approved):

PR #2816, opened 2023-10-13
Fix failing integration test

MDrakos:


MDrakos:
Note that we will not run into this issue with our regular integration tests because they copy binaries from the build directory and execute those directly. The issue surfaced in this test because we are running `state` which is resolving to the state tool that was installed by the GH action earlier.

test/integration/install_scripts_int_test.go
```diff
@@ -50,6 +50,10 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
 			ts := e2e.New(suite.T(), false)
 			defer ts.Close()
 
+			// The home dir is set already for the test session, but we need to
+			// set it for the test so helper functions can find it.
+			suite.T().Setenv(constants.HomeEnvVarName, ts.Dirs.HomeDir)
```
Naatan, line 55: Nice find, but we should probably do this as part of `e2e.New()`. Otherwise we'll likely keep running into this.

test/integration/install_scripts_int_test.go
```diff
@@ -110,28 +113,43 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
 
 			if tt.Activate != "" || tt.ActivateByCommand != "" {
 				cp.Expect("Creating a Virtual Environment")
-				cp.Expect("Quick Start", termtest.OptExpectTimeout(time.Second*60))
+				cp.Expect("Quick Start", termtest.OptExpectTimeout(time.Minute*2))
```
Naatan, line 116:
```suggestion
				cp.Expect("Quick Start", e2e.RuntimeSourcingTimeoutOpt)
```

test/integration/install_scripts_int_test.go
```diff
@@ -110,28 +113,43 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
...
 
-			cp.SendLine("state --version")
+			// We get the default install path and use that to directly invoke
+			// the state tool. This is to avoid inadvertently using the state
+			// tool that is already on the PATH.
+			installPath, err := installation.InstallPathForBranch(constants.BranchName)
+			suite.NoError(err)
+
+			binPath := filepath.Join(installPath, "bin")
+			statePath := filepath.Join(binPath, "state"+osutils.ExeExt)
```
Naatan, line 131: e2e should handle this, and if it's not it'll need to be fixed there. You should be able to use `ts.Exe`.

But that is too much here. This shell should have no awareness of the CI State Tool. It sounds like we're bleeding the state tool PATH from the CI environment into our test environment. We should focus on fixing that because there's no way we'll always remember this caveat.

Naatan (changes requested):

Naatan (changes requested):
This looks good, but as we said this is an interim workaround. I'd like to see comments calling this out as a temp workaround with a link to the jira story so we don't lose track of the core issue.

Naatan (dismissed):

test/integration/install_scripts_int_test.go
```diff
@@ -146,8 +146,10 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
 			suite.NoError(err)
 			suite.FileExists(stateExec)
 
+
 			suite.assertBinDirContents(binPath)
 			suite.assertCorrectVersion(ts, binPath, tt.Version, tt.Channel)
+      suite.assertAnalytics(ts)
```
Naatan, line 152:
```suggestion
			suite.assertAnalytics(ts)
```

Naatan (changes requested):

MDrakos:
For some reason I thought v42 was already out so changed the base to v43... I've now changed it back to v42 🤦‍♂️

Naatan (approved):

PR #2817, opened 2023-10-13
Remote installer should send a sessionToken to analytics.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2237" title="DX-2237" target="_blank">StoryDX-2237</a>  The remote installer supports a session token
  </td></table>
  <br />

### DIFF
--- a/cmd/state-remote-installer/main.go
+++ b/cmd/state-remote-installer/main.go
@@ -95,6 +95,12 @@ func main() {
 		return
 	}
 
+	// Store sessionToken to config
+	err = cfg.Set(anaConst.CfgSessionToken, "remote_"+constants.RemoteInstallerVersion)
+	if err != nil {
+		logging.Error("Unable to set session token: " + errs.JoinMessage(err))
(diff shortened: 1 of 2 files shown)


cmd/state-remote-installer/main.go
```diff
@@ -95,6 +95,12 @@ func main() {
 		return
 	}
 
+	// Store sessionToken to config
+	err = cfg.Set(anaConst.CfgSessionToken, "remote_"+constants.RemoteInstallerVersion)
```
mitchell-as, line 99: Note: this will override any subsequent `ACTIVESTATE_SESSION_TOKEN` environment variables set by `install.sh` due to https://github.com/ActiveState/cli/blob/c6a5565c1eb184571bca3833bc8ccac7aac81eb8/cmd/state-installer/installer.go#L53-L57, but that should be addressed in DX-2228.

MDrakos (approved):

PR #2823, opened 2023-10-16
Fixed sessionToken installer analytics.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2228" title="DX-2228" target="_blank">BugDX-2228</a>  Investigate missing session ID in state tool analytics
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -131,7 +131,7 @@ func main() {
 		}
 		err := cfg.Set(anaConst.CfgSessionToken, sessionToken)
 		if err != nil {
-			logging.Error("Unable to set session token: " + errs.JoinMessage(err))
+			multilog.Error("Unable to set session token: " + errs.JoinMessage(err))
 		}
 		break
 	}
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -123,6 +123,19 @@ func main() {
 	logging.Debug("Original Args: %v", os.Args)
 	logging.Debug("Processed Args: %v", processedArgs)
 
+	// Store sessionToken to config
+	for _, envVar := range []string{constants.OverrideSessionTokenEnvVarName, constants.SessionTokenEnvVarName} {
+		sessionToken, ok := os.LookupEnv(envVar)
+		if !ok {
+			continue
+		}
+		err := cfg.Set(anaConst.CfgSessionToken, sessionToken)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


mitchell-as:
Test failures are not due to this PR. They are due to existing, known issues and the usual timeouts.

cmd/state-installer/cmd.go
```diff
@@ -123,6 +123,19 @@ func main() {
...
 
+	// Store sessionToken to config
+	for _, envVar := range []string{constants.OverrideSessionTokenEnvVarName, constants.SessionTokenEnvVarName} {
+		sessionToken, ok := os.LookupEnv(envVar)
+		if !ok {
+			continue
+		}
+		err := cfg.Set(anaConst.CfgSessionToken, sessionToken)
+		if err != nil {
+			logging.Error("Unable to set session token: " + errs.JoinMessage(err))
```
MDrakos, line 134: We _may_ want to report this to rollbar so we can more easily diagnose if the issue of a missing session token comes up again.

test/integration/install_scripts_int_test.go
```diff
@@ -260,6 +264,21 @@ func (suite *InstallScriptsIntegrationTestSuite) assertCorrectVersion(ts *e2e.Se
...
 }
 
+func (suite *InstallScriptsIntegrationTestSuite) assertAnalytics(ts *e2e.Session) {
+	// Verify analytics reported the correct sessionToken.
+	sessionTokenFound := false
+	events := parseAnalyticsEvents(suite, ts)
+	suite.Require().NotEmpty(events)
+	for _, event := range events {
+		if event.Category == anaConst.CatInstallerFunnel && event.Dimensions != nil {
+			suite.Assert().NotEmpty(*event.Dimensions.SessionToken)
```
MDrakos, line 274: nit: We _could_ assert the equality of the token value, not just is presence, as the comment above suggests.

MDrakos (approved):
Looks good to me, just a couple of nits.

cmd/state-installer/cmd.go
mitchell-as: Yes, that's a good idea.

test/integration/install_scripts_int_test.go
mitchell-as: I've updated the comment. I am guilty of copy-pasting the code from another test :/ The important thing is that it's not empty.

MDrakos (approved):

PR #2829, opened 2023-10-18
Support old project format with commitID in activestate.yaml.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2263" title="DX-2263" target="_blank">StoryDX-2263</a>  As a State Tool user, I should be able to use my project with the old commit ID format.
  </td></table>
  <br />
 

Instead of automatically migrating projects, prompt the user. Projects in the old format are read-only.


mitchell-as:
Test failures are not due to this PR, but are known, existing issues and the usual timeouts.

cmd/state/main.go
```diff
@@ -219,6 +220,10 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	project.RegisterExpander("mixin", project.NewMixin(auth).Expander)
 	project.RegisterExpander("secrets", project.NewSecretPromptingExpander(secretsapi.Get(), prompter, cfg, auth))
 
+	if pj != nil {
+		projectmigration.PromptToMigrateIfNecessary(pj, prompter, out)
+	}
```
Naatan, line 225: I'd rather see this addressed when we try to access the local commit. Because otherwise we'll be doing this check on commands that don't even use the yaml file. And it could be problematic for commands such as `state _prepare`.

internal/locale/locales/en-us.yaml
```diff
@@ -2080,5 +2080,10 @@ err_local_commit_file:
     To view your projects commits run '[ACTIONABLE]state history[/RESET]'.
     
     Please avoid deleting or editing this file directly.
+projectmigration_confirm:
+  other: |
+    Your project is currently read-only until you migrate to the new project format.
+    After you migrate, older versions of the State Tool will not be able to use your project until they update.
+    Would you like to perform the migration now?
```
Naatan, line 2087: @rawktron please review

internal/runbits/projectmigration/projectmigration.go
```diff
@@ -0,0 +1,68 @@
...
+
+type promptable interface {
+	Confirm(title, message string, defaultChoice *bool) (bool, error)
+}
+
+// PromptToMigrateIfNecessary checks if the given project has been migrated to the new project
+// format. If not, the user is prompted to do so. This should be called in main() as soon as a
+// project file is parsed.
+func PromptToMigrateIfNecessary(proj projecter, prompt promptable, out output.Outputer) error {
+	projectDir := filepath.Dir(proj.Path())
```
Naatan, line 29: It has a `Dir()` method on the project struct.

internal/runbits/projectmigration/projectmigration.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	defaultChoice := false
+	if migrate, err := prompt.Confirm("", locale.T("projectmigration_confirm"), &defaultChoice); err == nil && !migrate {
+		if out.Config().Interactive {
+			out.Notice(locale.Tl("projectmigration_declined", "Migration declined for now"))
+		}
+		return nil
+	} else if err != nil {
+		return locale.WrapError(err, "err_projectmigration_confirm", "Could not confirm migration choice")
```
Naatan, line 41: We shouldn't be locale wrapping anymore, unless it's a vague error but we think it contains valuable information. This can probably just be an errs.Wrap().

pkg/localcommit/localcommit.go
```diff
@@ -64,6 +69,18 @@ func Get(projectDir string) (strfmt.UUID, error) {
...
+// the old format (activestate.yaml).
+func GetCompatible(proj projecter) (strfmt.UUID, error) {
+	if commitID, err := Get(proj.Dir()); err == nil {
+		return commitID, nil
+	} else if IsFileDoesNotExistError(err) {
+		return strfmt.UUID(proj.LegacyCommitID()), nil
+	} else {
+		return "", err
+	}
+}
```
Naatan, line 82: I don't think this belongs in the `localcommit` package. Its only concern should be the local commit file.

That said I realize the ship has kinda sailed with `AddToGitignore` also being implemented here. We should probably address that, but not as part of this PR. Still, we shouldn't add to the problem.

test/integration/project_migration_int_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/ActiveState/cli/internal/testhelpers/e2e"
+	"github.com/ActiveState/cli/internal/testhelpers/tagsuite"
+	"github.com/stretchr/testify/suite"
+)
+
+type ProjectMigrationIntegrationTestSuite struct {
+	tagsuite.Suite
+}
+
+func (suite *ProjectMigrationIntegrationTestSuite) TestPromptMigration() {
```
Naatan, line 19: Great test!

Naatan (changes requested):

pkg/localcommit/localcommit.go
mitchell-as: I wasn't sure where this was supposed to go, so I threw it into a new `internal/runbits/commitid` package. I'm sure that's wrong, so please let me know what you prefer. Some clients will want `localcommit.Get()` because they require such a thing. Other clients will want a way to get a legacy commit ID or a new localcommit ID, so for now it's `commitid.GetCompatible()`.

internal/runbits/projectmigration/projectmigration.go
```diff
@@ -22,31 +23,41 @@ type promptable interface {
...
-	if _, err := localcommit.Get(projectDir); err == nil || !localcommit.IsFileDoesNotExistError(err) {
-		return err
+var prompt promptable
+var out output.Outputer
+var declined bool
+
+func Register(prompt_ promptable, out_ output.Outputer) {
+	prompt = prompt_
+	out = out_
+}
```
Naatan, line 33: This is an antipattern for how we do this type of thing in our codebase. Either we should pass around a pointed struct that contains this context, or we always require these variables to be passed when invoking PromptAndMigrate.

That said I can see why you went this route as it feels really wasteful to add all this boilerplate for effectively a legacy use-case.

I feel like we need a pattern for handling legacy transitions like this, but that's probably too much to handle within the scope of this story.

Perhaps for this story we should clearly call out its legacy nature. I'm thinking add comments here clearly explaining that this is an antipattern and it's only ok because this is for a legacy use-case. I'm also thinking we should clearly mark packages meant for legacy use-cases as such, maybe by putting them a dir deeper under a parent dir called "legacy"? So you'd have `runbits/legacy/projectmigration`.

I'm not a big fan of any of these solutions, but I don't see a solution that doesn't have downsides. Main things I want to achieve here is maintainable code that doesn't start infecting other code, which I think is also where you were coming from. Happy to iterate with other ideas.

pkg/project/project.go
```diff
@@ -247,9 +248,15 @@ func (p *Project) Cache() string { return p.projectfile.Cache }
 
 // Namespace returns project namespace
 func (p *Project) Namespace() *Namespaced {
-	commitID, err := localcommit.GetCompatible(p)
+	commitID, err := localcommit.Get(p.Dir())
```
Naatan, line 251: It just occurred to me we should not be exposing the commit ID through here at all. Could you see if it's easy to drop this from here or if not file a followup story? Effectively the project package should have zero awareness of localcommit, only of legacy commit.

internal/runbits/commitid/commitid.go
```diff
@@ -0,0 +1,33 @@
+package commitid
```
Naatan, line 1: Since this invoked the prompter I think it makes sense you put it in as a runbit, though the naming of the package and method is triggering my OCD a little.

Open to other naming, but how about `commitmediator.Get()`?

Naatan (changes requested):

internal/runbits/projectmigration/projectmigration.go
mitchell-as: The need for this arose because we were auto-migrating the project as soon as it was read, which did not allow for any prompt/output parameter passing (i.e. `projectfile.FromPath(pjPath)` followed by `project.New(pjf, out)`), especially since analytics was not available yet (prompt needs this, and analytics needs to know the project; circular dependency).

However, after a couple of iterations, we're making prompting a lazy action, so we can pass around prompter and output interfaces as normal. It's quite verbose, but I've made the change.

From what I understand, there's no need for a `legacy/` parent package.

internal/runbits/commitid/commitid.go
mitchell-as: I had no clue what name to use, so I'm glad for your suggestion.

pkg/project/project.go
mitchell-as: Sure. After looking through the codebase, I cannot find a legitimate use for it anymore.

pkg/projectfile/projectfile_test.go
```diff
@@ -468,25 +467,3 @@ languages:
...
-	assert.Equal(t, "Owner", proj.Owner())
-	assert.Equal(t, "Name", proj.Name())
-	assert.Equal(t, "main", proj.BranchName())
-
-	commitIdFile := filepath.Join(tempDir, constants.ProjectConfigDirName, constants.CommitIdFileName)
-	require.FileExists(t, commitIdFile)
-	assert.Equal(t, commitID, string(fileutils.ReadFileUnsafe(commitIdFile)))
-
-	assert.NotContains(t, string(fileutils.ReadFileUnsafe(asy)), commitID)
-}
```
mitchell-as, line 492: Migration is no longer automatic. The new integration test covers this.

Naatan (changes requested):
I'm sorry, but as I expressed during stand I wanted to avoid adding this much boilerplate for the purpose of a legacy transition mechanic. This is adding a ton of complexity and code that will have to be maintained and any new code will also have to satisfy this pattern, all to facilitate migration logic for when grabbing a commit ID.

I'm open to other solutions aside from the one I suggested earlier, but the design goal here is not to have the legacy handling code introduce tons of boilerplate or add complexity whenever we need to access a commit.

This feels like the type of problem that we can't have a non-compromising solution for, at least now without heavy investment, which we want to avoid for this story.

mitchell-as:
Based on your feedback, you left the door open to interpretation. Either avoid the anti-pattern and pass things around (which is what we *should* be doing), or call it out and make legacy packages, etc. You acknowledged that there is no solution that does not have downsides. Since your first bit of feedback was "this is an anti-pattern, we should be passing things around", I figured that was your preference. If I had known you didn't want parameter passing, I wouldn't have done it.

internal/locale/locales/en-us.yaml
rawktron: What's the full message? We have to give them some context for this. Is there another part of this message?

rawktron: It should be something like "State Tool has introduced a new project format to provide enhanced functionality. Your project is currently using the old format and will remain read-only until you migrate to the new project format.

WARNING: After you migrate, older versions of the State Tool will not be able to use your project.

Would you like to perform the migration now?"

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -0,0 +1,79 @@
...
+		return false, nil
+	} else if err != nil {
+		return false, errs.Wrap(err, "Could not confirm migration choice")
+	}
+
+	if err := localcommit.Set(proj.Dir(), proj.LegacyCommitID()); err != nil {
+		return false, errs.Wrap(err, "Could not create local commit file")
+	}
+
+	if fileutils.DirExists(filepath.Join(proj.Dir(), ".git")) {
```
Naatan, line 59: Realize this probably wasn't introduced here but we should go up the directory tree to check for this also, since people may not checkout their activestate.yaml inside the same dir as their git repo (eg. TheHomeRepot).

You can use:

https://github.com/ActiveState/cli/blob/35d1e55db5e9cf76e5e78f63dc6ae461060b1f3e/internal/fileutils/fileutils.go#L422

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -0,0 +1,79 @@
...
+		err := localcommit.AddToGitIgnore(proj.Dir())
+		if err != nil {
+			multilog.Error("Unable to add local commit file to .gitignore: %v", err)
```
Naatan, line 62: This would almost certainly be due to something client-side like permission errors I think? We should probably not log this to rollbar. Or maybe add a check to only log if it's not a permission or IO error.

pkg/project/project.go
```diff
@@ -242,11 +247,8 @@ func (p *Project) Cache() string { return p.projectfile.Cache }
 
 // Namespace returns project namespace
 func (p *Project) Namespace() *Namespaced {
-	commitID, err := localcommit.Get(p.Dir())
-	if err != nil && !localcommit.IsFileDoesNotExistError(err) {
-		multilog.Error("Unable to get local commit: %v", errs.JoinMessage(err))
-	}
-	return &Namespaced{p.projectfile.Owner(), p.projectfile.Name(), &commitID, false}
+	commitId := strfmt.UUID(p.projectfile.LegacyCommitID())
+	return &Namespaced{p.projectfile.Owner(), p.projectfile.Name(), &commitId, false}
```
Naatan, line 251: Is the legacy commit ID still required as part of the namespace? Suggest dropping it altogether from the namespace.

Naatan (changes requested):

internal/runbits/legacy/projectmigration/projectmigration.go
mitchell-as: Sadly, that's files only, not directories. Still, I'll do something to walk up the tree.

internal/runbits/legacy/projectmigration/projectmigration.go
mitchell-as: I wasn't able to find a generic ErrIO variable, but I did find ErrPermission, so we'll go with that.

pkg/project/project.go
mitchell-as: Having an eye towards https://activestatef.atlassian.net/jira/software/c/projects/DX/issues/DX-2289, I'd say we probably shouldn't drop it yet.

mitchell-as: Never mind. We abandoned during stand, so I removed it. Let's see what breaks.

Naatan (approved):
Looks good! Thanks for sticking with it.

PR #2830, opened 2023-10-18
Remove TestStateTrayRemoval

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2260" title="DX-2260" target="_blank">TaskDX-2260</a>  Nightly Failure: Update TestStateTrayRemoval to be an installer test instead
  </td></table>
  <br />
 

Turns out it was failing on an install of state tool v0.35. We were testing an update flow of sorts which needed an older state tool. Obviously not worth addressing going forward.


mitchell-as (approved):
Cool!

PR #2834, opened 2023-10-19
Don't add language templating to yaml if runtime is disabled

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2290" title="DX-2290" target="_blank">BugDX-2290</a>  Nightly failure on TestActivate_NamespaceWins 
  </td></table>
  <br />
 

Fixes CI failure due to python2 not existing

### DIFF
--- a/pkg/projectfile/projectfile.go
+++ b/pkg/projectfile/projectfile.go
@@ -923,14 +923,6 @@ type CreateParams struct {
 	Cache      string
 }
 
-// TestOnlyCreateWithProjectURL a new activestate.yaml with default content
-func TestOnlyCreateWithProjectURL(projectURL, path string) (*Project, error) {
-	return createCustom(&CreateParams{
-		ProjectURL: projectURL,
-		Directory:  path,
-	}, language.Python3)
-}
-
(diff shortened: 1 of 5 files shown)


pkg/projectfile/projectfile.go
```diff
@@ -923,14 +923,6 @@ type CreateParams struct {
 	Cache      string
 }
 
-// TestOnlyCreateWithProjectURL a new activestate.yaml with default content
-func TestOnlyCreateWithProjectURL(projectURL, path string) (*Project, error) {
-	return createCustom(&CreateParams{
-		ProjectURL: projectURL,
-		Directory:  path,
-	}, language.Python3)
-}
```
Naatan, line 932: Apologies for sneaking this into my PR, it's not directly related to the story but I ran into it for one of the failing tests and wanted to do a quick cleanup. If it distracts from the review I'll open a separate PR.

This code shouldn't have been inside a non-test file.

mitchell-as: That's fine. I'm in favor of fixing things as you find them. It's faster and you stay in the zone.

mitchell-as (approved):

PR #2842, opened 2023-10-23
Lack of having a commit ID for push/pull is also an error.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2286" title="DX-2286" target="_blank">TaskDX-2286</a>  Audit and possibly remove instances where lack of a commit ID file is not an error
  </td></table>
  <br />
 

This will not be merged before DX-2224 is.

### DIFF
--- a/test/integration/push_int_test.go
+++ b/test/integration/push_int_test.go
@@ -264,24 +264,6 @@ func (suite *PushIntegrationTestSuite) TestPush_NoChanges() {
 	}
 }
 
-func (suite *PushIntegrationTestSuite) TestPush_NoCommit() {
-	suite.OnlyRunForTags(tagsuite.Push)
-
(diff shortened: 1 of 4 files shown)


test/integration/push_int_test.go
```diff
@@ -264,24 +264,6 @@ func (suite *PushIntegrationTestSuite) TestPush_NoChanges() {
 	}
 }
 
-func (suite *PushIntegrationTestSuite) TestPush_NoCommit() {
```
MDrakos, line 267: Isn't this still testing a valid error case? We could just update the test to reflect the changes in this PR.

MDrakos (changes requested):

MDrakos (approved):

PR #2844, opened 2023-10-24
Initial usage of buildplanner for `state init`.

mitchell-as:


mitchell-as:
I'm not really sure what I'm doing here, so feel free to correct me or suggest what I can do better.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -355,6 +355,14 @@ type StageCommitResult struct {
 	Commit *Commit `json:"stageCommit"`
 }
 
+type projectCreated struct {
+	Commit *Commit `json:"commit"`
+}
+
+type CreateProjectResult struct {
+	ProjectCreated *projectCreated `json:"createProject"`
+}
```
MDrakos, line 364: This type needs to also embed the error types that the API can return. Once they are modeled, the buildplaner function for creating a project needs to inspect this result and handle the potential errors.

The commit type has some embedded errors which you might have to mirror here. https://github.com/ActiveState/cli/blob/master/pkg/platform/api/buildplanner/model/buildplan.go#L371

There are also some convenience functions that you can use and/or expand to handle the new error cases: https://github.com/ActiveState/cli/blob/master/pkg/platform/api/buildplanner/model/buildplan.go#L371

Finally, you will likely need a function to process the `ProjectCreatedError` specifically, like this one: https://github.com/ActiveState/cli/blob/master/pkg/platform/api/buildplanner/model/buildplan.go#L274

I realize this is a fair amount of information to take in and handle so let me know if I can help clear anything up.

pkg/platform/model/buildplanner.go
```diff
@@ -349,6 +350,54 @@ func (bp *BuildPlanner) GetBuildExpression(owner, project, commitID string) (*bu
...
+						"camel_flags": [],
+						"platforms": ["%s"],
+						"requirements": [],
+						"solver_version": null
+					}
+				},
+				"in": "$runtime"
+			}
+		}
+	`, time.Now().Format(time.RFC3339), platformID)))
```
MDrakos, line 374: Could this be a convenience function in the `buildexpression` package? This seems like it should live there.

pkg/platform/model/buildplanner.go
```diff
@@ -349,6 +350,54 @@ func (bp *BuildPlanner) GetBuildExpression(owner, project, commitID string) (*bu
...
+	})
+
+	request := request.CreateProject(owner, project, private, expr, locale.T("commit_message_add_initial"))
+	resp := &bpModel.CreateProjectResult{}
+	err = bp.client.Run(request, resp)
+	if err != nil {
+		return "", processBuildPlannerError(err, "Failed to create project")
+	}
+	if resp.ProjectCreated == nil {
+		return "", errs.New("ProjectCreated is nil")
```
MDrakos, line 395: This function will need error handling to this: https://github.com/ActiveState/cli/blob/master/pkg/platform/api/buildplanner/model/buildplan.go#L245

MDrakos (changes requested):
Honestly, this looks really good so far. The error handling will need to be updated to work with this new mutation. It's not obvious how to do that and the implicit nature of graphQL + Go unmarshalling doesn't make it easier. I've left some examples that hopefully help get you started. Let me know if I can help clear anything up.

mitchell-as:
I started implementing user-facing errors for buildplanner errors raised during project creation. I created a ticket for finishing user-facing errors for the rest of the initialize runner's package here: https://activestatef.atlassian.net/browse/DX-2300

pkg/platform/api/buildplanner/model/buildplan.go
mitchell-as: That's fine, but the `Type` and `Error` info needs to live in the `projectCreated` struct for some reason. If I put it in `CreateProjectResult`, I don't get any error information.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -333,6 +333,19 @@ func ProcessProjectError(project *Project, fallbackMessage string) error {
...
+func (p *ProjectCreatedError) Error() string { return "ProjectCreatedError" }
+
+func ProcessProjectCreatedError(pcErr *projectCreated, fallbackMessage string) error {
+	switch pcErr.Type {
+	case AlreadyExistsErrorType, ForbiddenErrorType:
+		return &ProjectCreatedError{pcErr.Type}
+	default:
+		return errs.New(fallbackMessage)
+	}
+}
```
MDrakos, line 347: For most of the error types returned from the buildplanner we want to ensure that the `message` field is being forwarded to the user. This is true for all of them, like a `forbidden` error but for parsing and validation errors we should forward the message. I don't think this will happen with the current implementation.

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -171,6 +171,37 @@ func New(data []byte) (*BuildExpression, error) {
 	return expr, nil
 }
 
+// NewFromPlatformIDAndRequirements creates a minimal buildexpression from the given platform UUID
+// and requirement list.
+func NewFromPlatformIDAndRequirements(platformID string, requirements []model.Requirement) (*BuildExpression, error) {
```
MDrakos, line 176: This is really specific to the current use case. Do you think we could make this more generic? It could return a blank build expression and then we leave it up to the caller to update any requirements and timestamps.

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -171,6 +171,37 @@ func New(data []byte) (*BuildExpression, error) {
...
+						"camel_flags": [],
+						"platforms": ["%s"],
+						"requirements": [],
+						"solver_version": null
+					}
+				},
+				"in": "$runtime"
+			}
+		}
+	`, time.Now().Format(time.RFC3339), platformID)))
```
MDrakos, line 195: We need to source the timestamp from the inventory API when we make updates like this. If we push it too far there could be unforeseen issues. If you look at other calls to `StageCommit` we are using this function to update the timestamp: https://github.com/ActiveState/cli/blob/master/pkg/platform/model/inventory.go#L404

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: Ahh okay, that has to do with how the response is structured. Our types need to reflect the JSON structure so we have to be careful with how it all comes together.

MDrakos (changes requested):
Almost there! My comments are primarily based on the knowledge that I picked up from working on the Build Planner code previously, so I wouldn't expect you to have known about them.

mitchell-as:
I had to do a bit of cleanup. Let me know if I missed anything.

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -333,17 +333,18 @@ func ProcessProjectError(project *Project, fallbackMessage string) error {
...
-func (p *ProjectCreatedError) Error() string { return "ProjectCreatedError" }
+func (p *ProjectCreatedError) Error() string { return p.Message }
 
 func ProcessProjectCreatedError(pcErr *projectCreated, fallbackMessage string) error {
-	switch pcErr.Type {
-	case AlreadyExistsErrorType, ForbiddenErrorType:
-		return &ProjectCreatedError{pcErr.Type}
-	default:
-		return errs.New(fallbackMessage)
+	if pcErr.Type != "" {
```
MDrakos, line 344: Can we add a comment here and reference the user-facing errors story? We should be handling the specific error types but I know that's out of scope currently.

MDrakos (dismissed):
Looks good! Just one nit.

MDrakos (approved):

PR #2849, opened 2023-10-26
state-svc reports its log file as an absolute, not relative path.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2301" title="DX-2301" target="_blank">BugDX-2301</a>  `state-svc status` log path is corrupted
  </td></table>
  <br />

### DIFF
--- a/internal/logging/defaults.go
+++ b/internal/logging/defaults.go
@@ -72,6 +72,9 @@ func FilePath() string {
 }
 
 func FilePathFor(filename string) string {
+	if filepath.IsAbs(filename) {
+		return filename
+	}
 	return filepath.Join(datadir, "logs", filename)
 }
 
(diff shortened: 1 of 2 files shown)


internal/logging/defaults.go
```diff
@@ -72,6 +72,9 @@ func FilePath() string {
 }
 
 func FilePathFor(filename string) string {
+	if filepath.IsAbs(filename) {
```
MDrakos, line 75: If the path is absolute it could still not be the correct log file path. We may want to look into the `GetLogFileName` of the `comm` package because it appears that it's not returning just the log file name but the entire path as well.

MDrakos (changes requested):

internal/logging/defaults.go
mitchell-as: Going up the chain it looks like the service is instantiated with a full path instead of just a name.

MDrakos (approved):

PR #2862, opened 2023-11-01
Added commitID for activation tests that immediately run `state pull` to update.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2308" title="DX-2308" target="_blank">BugDX-2308</a>  Nightly failure: Activation errors
  </td></table>
  <br />
 

The commitID doesn't matter, but it ought to be specified. Projects without commitIDs don't exist in the real world as far as I can tell.

### DIFF
--- a/test/integration/activate_int_test.go
+++ b/test/integration/activate_int_test.go
@@ -423,6 +423,7 @@ version: %s
 `, constants.BranchName, constants.Version))
 
 	ts.PrepareActiveStateYAML(content)
+	ts.PrepareCommitIdFile("59404293-e5a9-4fd0-8843-77cd4761b5b5")
 
 	// Pull to ensure we have an up to date config file
 	cp := ts.Spawn("pull")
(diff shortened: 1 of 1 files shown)


mitchell-as:
The failure here is intermittent and not related to this PR.

test/integration/activate_int_test.go
```diff
@@ -423,6 +423,7 @@ version: %s
 `, constants.BranchName, constants.Version))
 
 	ts.PrepareActiveStateYAML(content)
+	ts.PrepareCommitIdFile("59404293-e5a9-4fd0-8843-77cd4761b5b5")
```
MDrakos, line 426: Shouldn't the commit ID file be disabled for v43? Sorry if I'm missing something here...

mitchell-as: That function body will have a temporary shim to handle legacy vs. commit ID file. At this time since that PR is not merged, we're still using commit ID files.

MDrakos: Does this change now that the PR is removed?

MDrakos (approved):

test/integration/activate_int_test.go
mitchell-as: The function ends up modifying the activestate.yaml to the `project: org/project#commitId` format instead of creating a commit file. Whenever we end up putting localcommits and buildscripts back, that function will go back to writing commit files instead of modifying activestate.yaml.

PR #2865, opened 2023-11-02
Add error tips to structured/JSON output.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2284" title="DX-2284" target="_blank">StoryDX-2284</a>  When `-o json` used the `If this is a private project you may need to authenticate with 'state auth'` is missing
  </td></table>
  <br />
 

For example, this suggests to authenticate prior to checking out a private project.

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -251,7 +251,9 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 		if childCmd != nil {
 			cmdName = childCmd.JoinedSubCommandNames() + " "
 		}
-		err = errs.AddTips(err, locale.Tl("err_tip_run_help", "Run → [ACTIONABLE]`state {{.V0}}--help`[/RESET] for general help", cmdName))
+		if !out.Type().IsStructured() {
+			err = errs.AddTips(err, locale.Tl("err_tip_run_help", "Run → [ACTIONABLE]`state {{.V0}}--help`[/RESET] for general help", cmdName))
+		}
 		cmdletErrors.ReportError(err, cmds.Command(), an)
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/cmdlets/errors/errors.go
```diff
@@ -92,6 +82,20 @@ func (o *OutputError) MarshalOutput(f output.Format) interface{} {
...
+	errorTips := []string{}
+	for _, err := range errs.Unpack(err) {
+		if v, ok := err.(ErrorTips); ok {
+			for _, tip := range v.ErrorTips() {
+				if !funk.Contains(errorTips, tip) {
+					errorTips = append(errorTips, tip)
+				}
+			}
+		}
+	}
```
MDrakos, line 95: nit: Maybe we can invert some logic to reduce the nesting here?

cmd/state/main.go
```diff
@@ -251,7 +251,9 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 		if childCmd != nil {
 			cmdName = childCmd.JoinedSubCommandNames() + " "
 		}
-		err = errs.AddTips(err, locale.Tl("err_tip_run_help", "Run → [ACTIONABLE]`state {{.V0}}--help`[/RESET] for general help", cmdName))
+		if !out.Type().IsStructured() {
+			err = errs.AddTips(err, locale.Tl("err_tip_run_help", "Run → [ACTIONABLE]`state {{.V0}}--help`[/RESET] for general help", cmdName))
+		}
```
MDrakos, line 256: Is there any reasonable way we can avoid the output check? Is this check going to be required for adding other error tips?

MDrakos (changes requested):

cmd/state/main.go
mitchell-as: We only have to do it for this one because we put this tip before all others. Any other tips added (like a link to the forums) comes later and does not need a check like this because `getErrorTips()` will not have those tips yet.

MDrakos: Awesome, thank you for the clarification!

MDrakos (approved):

PR #2867, opened 2023-11-02
Use httpretry client for update fetcher.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2132" title="DX-2132" target="_blank">TaskDX-2132</a>  Our installer / updater clearly communicates network errors
  </td></table>
  <br />
 

Take advantage of network error handling.

### DIFF
--- a/internal/updater/fetcher.go
+++ b/internal/updater/fetcher.go
@@ -3,33 +3,41 @@ package updater
 import (
 	"crypto/sha256"
 	"fmt"
+	"io/ioutil"
 
 	"github.com/ActiveState/cli/internal/analytics"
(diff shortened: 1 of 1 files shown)


MDrakos (changes requested):
This looks like a good change for the updater. Can you also please verify that the installer will surface network errors?

mitchell-as:
Yes, if I unplug my network cable prior to running the installer (after it's been downloaded), it appears to run just fine. (It doesn't appear to need internet access). If I unplug my network cable prior to running install.sh (to fetch the installer), I'm getting OS error messages that hint at a networking issue.

Is this sufficient?

MDrakos (approved):

PR #2870, opened 2023-11-07
Make DNS errors input errors to inform the user something is wrong with their network.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2269" title="DX-2269" target="_blank">BugDX-2269</a>  state init fails to setup runtime
  </td></table>
  <br />

### DIFF
--- a/internal/retryhttp/client.go
+++ b/internal/retryhttp/client.go
@@ -124,7 +124,7 @@ func normalizeResponse(res *http.Response, err error) (*http.Response, error) {
 
 	var dnsError *net.DNSError
 	if errors.As(err, &dnsError) {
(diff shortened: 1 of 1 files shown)


MDrakos (commented):
My understanding is that input errors only affect the reporting to rollbar, not the user. Am I wrong?

mitchell-as:
You are correct. I thought for some reason that input errors would be surfaced to the user (taking priority over `errs.Wrap()` and `locale.WrapError()` errors. After some experimentation I found this was not the case, so I made some comments to the original ticket and punted it back to triage.

Thanks for catching this.

PR #2871, opened 2023-11-08
Show indication the user is still in a `state shell` when starting a new shell within it.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2240" title="DX-2240" target="_blank">StoryDX-2240</a>  Reimplement - As a user I can expect to always see a clear indication that I am in a virtual environment
  </td></table>
  <br />

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -93,7 +93,7 @@ const ActivatedStateEnvVarName = "ACTIVESTATE_ACTIVATED"
 // ActivatedStateIDEnvVarName is the name of the environment variable that is set when in an activated state, its value will be a unique id identifying a specific instance of an activated state
 const ActivatedStateIDEnvVarName = "ACTIVESTATE_ACTIVATED_ID"
 
-// ActivatedStateProjectEnvVarName is the name of the environment variable that specifies the activated state's org/project namespace.
+// ActivatedStateNamespaceEnvVarName is the name of the environment variable that specifies the activated state's org/project namespace.
 const ActivatedStateNamespaceEnvVarName = "ACTIVESTATE_ACTIVATED_NAMESPACE"
 
 // ForwardedStateEnvVarName is the name of the environment variable that is set when in an activated state, its value will be the path of the project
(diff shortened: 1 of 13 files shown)


internal/constants/constants.go
```diff
@@ -93,6 +93,9 @@ const ActivatedStateEnvVarName = "ACTIVESTATE_ACTIVATED"
 // ActivatedStateIDEnvVarName is the name of the environment variable that is set when in an activated state, its value will be a unique id identifying a specific instance of an activated state
 const ActivatedStateIDEnvVarName = "ACTIVESTATE_ACTIVATED_ID"
 
+// ActivatedStateProjectEnvVarName is the name of the environment variable that specifies the activated state's org/project namespace.
```
MDrakos, line 96:
```suggestion
// ActivatedStateNamespaceEnvVarName is the name of the environment variable that specifies the activated state's org/project namespace.
```

internal/runners/shell/shell.go
```diff
@@ -87,7 +90,9 @@ func (u *Shell) Run(params *Params) error {
 	}
 
 	if process.IsActivated(u.config) {
-		return locale.NewInputError("err_shell_already_active", "", proj.NamespaceString(), proj.Dir())
+		activatedProjectNamespace := os.Getenv(constants.ActivatedStateNamespaceEnvVarName)
+		activatedProjectDir := os.Getenv(constants.ActivatedStateEnvVarName)
+		return locale.NewInputError("err_shell_already_active", "", activatedProjectNamespace, activatedProjectDir)
```
MDrakos, line 95: nit: do we want to have a default message in the case that one or both of these env vars are not set? While we do set them on activation the user still has the ability to change their environment.

MDrakos (changes requested):
Looks good to me, just one very minor change.

internal/runners/shell/shell.go
mitchell-as: I'm not sure I understand your question correctly, but I don't believe we guard against any user-setting/unsetting of `ACTIVESTATE*` env variables that we make use of/depend on. If these are set, we assume it was because of us. If they are not, we assume it was because of us.

MDrakos: In the case that these aren't set the message the user will get will be incomplete. My point was that we could guard against that by having a default message. This might be preferable to a message like:

```
You are already in a shell for ' ', located at ' '.
```

It's a niche case, which is why this is a nit, but it is also fairly easy for us to provide a default message in the case that these variables are not set.

MDrakos (approved):

internal/runners/shell/shell.go
mitchell-as: I see what you mean. Technically, that is out of scope for this PR since the error message already existed. Since Mark has not run into this in his test plan, I suspect there isn't currently a need for this. Feel free to open a task in JIRA to discuss on triage though.

PR #2875, opened 2023-11-09
Update error handling on revert (and pull)

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2319" title="DX-2319" target="_blank">BugDX-2319</a>  Error during `revert` command to a local commit and not auth execution
  </td></table>
  <br />
 

After discussing with Pete we will go forward with the improved error handling and follow up with the desired behavior once PB has had an opportunity to address it.

### DIFF
--- a/internal/runners/pull/rationalize.go
+++ b/internal/runners/pull/rationalize.go
@@ -26,6 +26,23 @@ func rationalizeError(err *error) {
 				),
 				errs.SetInput(),
 			)
+		case model.NotFoundErrorType, model.ForbiddenErrorType:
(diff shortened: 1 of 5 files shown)


mitchell-as (approved):
I approve on the condition that you run the "revert" integration tests and they pass. Yesterday the "activate" integration test failed after the error message changed and the test was not updated.

MDrakos:
> I approve on the condition that you run the "revert" integration tests and they pass. Yesterday the "activate" integration test failed after the error message changed and the test was not updated.

Good call!

PR #2876, opened 2023-11-09
Recognize headless errors on requirement operations

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2224" title="DX-2224" target="_blank">TaskDX-2224</a>  Remove headless commits
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/rationalize/types.go
+++ b/internal/runbits/rationalize/types.go
@@ -17,6 +17,8 @@ var ErrNotAuthenticated = errors.New("not authenticated")
 
 var ErrActionAborted = errors.New("aborted by user")
 
+var ErrHeadless = errors.New("headless")
+
 type ErrAPI struct {
 	Wrapped error
 	Code    int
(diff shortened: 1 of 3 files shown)


MDrakos:
Test failures are not due to the changes introduced in this PR

internal/runbits/rationalize/types.go
```diff
@@ -17,6 +17,8 @@ var ErrNotAuthenticated = errors.New("not authenticated")
 
 var ErrActionAborted = errors.New("aborted by user")
 
+var ErrHeadless = errors.New("headless")
```
mitchell-as, line 20: I view errors in this file as "global" errors. Is this going to be used outside the internal/runbits/requirements package? If not, we should put it in that package and remove it here.

mitchell-as (changes requested):

internal/runbits/rationalize/types.go
mitchell-as: Never mind, I think this will be used in user-facing errors for requirement operations like install/uninstall.

mitchell-as (approved):
Changing to approved after further thought. This looks good to me.

PR #2877, opened 2023-11-10
Initial implementation of user-facing errors for `state install` and `state uninstall`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2232" title="DX-2232" target="_blank">StoryDX-2232</a>  As a user I can expect user facing errors when I install/uninstall packages
  </td></table>
  <br />
 

I don't really know what I'm doing. I tried to identify what could go wrong during install/uninstall, and tried some various install/uninstall scenarios to see what happened:

- Not authenticated install/uninstall
- Install with non-existent package name
- Install package that already exists
- Uninstall with non-existent package

They should all return user-facing errors now.

### DIFF
--- a/internal/runners/packages/rationalize.go
+++ b/internal/runners/packages/rationalize.go
@@ -35,6 +35,12 @@ func rationalizeError(auth *authentication.Auth, err *error) {
 				errs.SetInput(),
 				errs.SetTips(locale.T("tip_private_project_auth")),
 			)
+		case bpModel.ForbiddenErrorType:
+			*err = errs.WrapUserFacing(*err,
+				locale.Tl("err_packages_forbidden", "Could not make runtime changes because you do not have permission to do so."),
+				errs.SetInput(),
+				errs.SetTips(locale.T("tip_private_project_auth")),
+			)
 		case bpModel.HeadOnBranchMovedErrorType:
 			*err = errs.WrapUserFacing(*err,
 				locale.T("err_buildplanner_head_on_branch_moved"),
@@ -42,7 +48,12 @@ func rationalizeError(auth *authentication.Auth, err *error) {
 			)
 		case bpModel.NoChangeSinceLastCommitErrorType:
 			*err = errs.WrapUserFacing(*err,
-				locale.Tl("err_packages_exists", "That package is already installed"),
+				locale.Tl("err_packages_exists", "That package is already installed."),
+				errs.SetInput(),
+			)
+		default:
+			*err = errs.WrapUserFacing(*err,
+				locale.Tl("err_packages_buildplanner_error", "Could not make runtime changes due to the following error: {{.V0}}", commitError.Message),
 				errs.SetInput(),
 			)
 		}
--- a/internal/runners/packages/rationalize.go
+++ b/internal/runners/packages/rationalize.go
@@ -1,0 +1,57 @@
+package packages
+
+import (
+	"errors"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/runbits/rationalize"
+	bpModel "github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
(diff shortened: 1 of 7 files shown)


pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -302,22 +302,43 @@ func IsErrorResponse(errorType string) bool {
 		errorType == ComitHasNoParentErrorType
 }
 
+type CommitError struct {
+	Type                   string
+	Message                string
+	*locale.LocalizedError // for legacy, non-user-facing error usages
```
mitchell-as, line 308: Since not all consumers are using user-facing errors, keep using the localized errors from the buildplanner for those consumers. The user-facing error rationalizer in this PR is making use of `Type`.

mitchell-as:

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: I believe that for this to be used this type needs an `Error()` function that returns the localized error message. Unless there is another way to surface this that I am missing.

MDrakos (changes requested):
This is a great start and my comments are relatively minor. I think the user-facing error initiative is one that we will add to and evolve. This is great groundwork for us to add more errors to the package operation flow over time.

pkg/platform/api/buildplanner/model/buildplan.go
mitchell-as: I believe it inherits `locale.LocalizedError`'s `Error()` function. Otherwise there would be a compiler error that you cannot use a `CommitError` struct as an `error` type.

internal/runners/packages/rationalize.go
```diff
@@ -0,0 +1,57 @@
...
+	case errors.Is(*err, rationalize.ErrNoProject):
+		*err = errs.WrapUserFacing(*err,
+			locale.T("err_no_project"),
+			errs.SetInput(),
+		)
+
+	// Error staging a commit during install.
+	case errors.As(*err, &commitError):
+		switch commitError.Type {
+		case bpModel.NotFoundErrorType:
```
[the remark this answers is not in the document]
mitchell-as: Not quite because of the message, but I've added another case with a slightly different message.

mitchell-as:
Again, any test failures are not due to this PR.

internal/runners/packages/rationalize.go
```diff
@@ -0,0 +1,68 @@
...
+				errs.SetTips(locale.T("tip_private_project_auth")),
+			)
+		case bpModel.HeadOnBranchMovedErrorType:
+			*err = errs.WrapUserFacing(*err,
+				locale.T("err_buildplanner_head_on_branch_moved"),
+				errs.SetInput(),
+			)
+		case bpModel.NoChangeSinceLastCommitErrorType:
+			*err = errs.WrapUserFacing(*err,
+				locale.Tl("err_packages_exists", "That package is already installed."),
```
MDrakos, line 51: I think this message will have to be more generic as it could also arise when running `state uninstall`.

pkg/platform/api/buildplanner/model/buildplan.go
MDrakos: Ahh yes, you're right!

MDrakos (changes requested):
Looking good, just one minor change request.

internal/runners/packages/rationalize.go
mitchell-as: May I ask how? If the user specifies a requirement that is not already installed, they'll get a `requirementNotFoundErr`, which is handled below.

mitchell-as:
https://github.com/ActiveState/cli/pull/2877#discussion_r1393416013

internal/runners/packages/rationalize.go
MDrakos: Ahh I see, I missed that part. Thanks for pointing that out.

MDrakos (approved):

PR #2882, opened 2023-11-13
Disable terminal echo while State Tool is running.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2214" title="DX-2214" target="_blank">TaskDX-2214</a>  As a user I can't interfere with State Tool by pressing keys during read-only actions
  </td></table>
  <br />
 

The exception is for prompts and subshells (which temporarily re-enable echoing).

Note: the survey package does this automatically, so no changes to the prompt package are required.


internal/subshell/bash/bash.go
```diff
@@ -219,3 +220,17 @@ func (v *SubShell) IsAvailable() bool {
 	}
 	return fileutils.FileExists(rcFile)
 }
+
+func (v *SubShell) TurnOffEcho() {
+	if runtime.GOOS == "windows" {
+		return // not supported
```
mitchell-as, line 226: Windows does not have access to the termios functions needed to disable echo in emulated shells like Git Bash (which is not interactive anyway).

internal/subshell/bash/bash.go
```diff
@@ -219,3 +220,17 @@ func (v *SubShell) IsAvailable() bool {
...
+	}
+	termecho.Off()
+}
+
+func (v *SubShell) TurnOnEcho() {
+	if runtime.GOOS == "windows" {
+		return // not supported
+	}
+	termecho.Off()
```
MDrakos, line 235: Shouldn't these, and the other `TurnOnEcho` functions be calling `termecho.On()`?

cmd/state/main.go
```diff
@@ -207,6 +207,13 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 
 	// Set up conditional, which accesses a lot of primer data
 	sshell := subshell.New(cfg)
+	if isInteractive {
+		// Disable terminal echo while State Tool is running.
+		// Other than in prompts and subshells (which temporarily re-enable echo), user typing should
+		// not interfere with output (e.g. runtime progress bars).
+		sshell.TurnOffEcho()
+		defer sshell.TurnOnEcho()
```
MDrakos, line 215: Does this also disable ctrl+c and other similar inputs? I don't believe that's something we want to disallow.

MDrakos: Also, is it possible that if the state tool errors mid-execution, we could leave the user's shell in a state where they cannot echo?

MDrakos (changes requested):
Something that's not obvious in this PR is how the echo is re-enabled for prompts... how is that handled?

mitchell-as:
As mentioned in the PR description, the survey library automatically handles the flipping on/off of terminal echo. I don't know exactly how it does it, but during testing it did.

internal/subshell/bash/bash.go
mitchell-as: Oh, how embarrassing. Thanks for catching this.

cmd/state/main.go
mitchell-as: You didn't see this, but in my original PR (https://github.com/ActiveState/cli/pull/2768), I noted that echo is restored after ^C, so no problem there. The `defer` does its job.

If there is a hard crash, echo will probably be lost, but this is the case for any terminal program that messes with termios, so the only resolution is for the user to exit this terminal and start up a new one.

MDrakos: It looks like the `defer` will run even on a panic as long as it's on the function call stack so we should be good even in the event of a crash as this defer is placed ahead of the actual command logic.

MDrakos (approved):

PR #2884, opened 2023-11-15
Added "preserve.prompt" config option for not changing PS1 in activated environments.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2152" title="DX-2152" target="_blank">StoryDX-2152</a>  Users Can Disable State Tool Prompt Edits
  </td></table>
  <br />


MDrakos (commented):
This looks great to me. My only contention is the actual config option itself. I feel like having `ps1` in the config key name _could_ be a bit too specific/technical. I'd like to get @rawktron's opinion on the key.

As it stands the user would run `state config set preserve.ps1 true`

mitchell-as:
I feel like anyone who actually cares about this setting would know what ps1 is :) I agree that @rawktron should approve this though.

rawktron:
> I feel like anyone who actually cares about this setting would know what ps1 is :) I agree that @rawktron should approve this though.

What about `preserve.prompt` which is slightly more descriptive in terms of what it does?

mitchell-as:
I've updated the user-facing config key to use `preserve.prompt`, but kept the internal PS1 references in the code because they make sense in that context.

MDrakos (approved):

PR #2885, opened 2023-11-20
Raise an error when updating a runtime currently in use.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2091" title="DX-2091" target="_blank">StoryDX-2091</a>  We ensure the runtime is not in use before updating it
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/setup.go
+++ b/pkg/platform/runtime/setup/setup.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/multilog"
+	"github.com/ActiveState/cli/internal/osutils"
 	"github.com/ActiveState/cli/internal/proxyreader"
 	"github.com/ActiveState/cli/internal/rollbar"
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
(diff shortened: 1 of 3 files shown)


mitchell-as:

pkg/platform/runtime/setup/setup.go
```diff
@@ -220,6 +235,35 @@ func (s *Setup) Update() (rerr error) {
 	return nil
 }
 
+func (s *Setup) getProcessesInUse() map[string]int32 {
```
MDrakos, line 238: I think it makes more sense for this function to be part of the `osutils` package and we can supply it the directory to compare. Also we could have os-specific versions of these functions to reduce the if statements here.

MDrakos (changes requested):

pkg/platform/runtime/setup/setup.go
mitchell-as: Okay, I moved it into the `osutils` package. However, when I tried splitting it up into OS-specific functions, I ended up with too much duplication. I left it as a single function.

MDrakos (approved):
Looks good, is there an easy way for us to introduce an integration test for this? If it's too much it can be a follow-up story. Just a nice to have right now.

mitchell-as:
https://activestatef.atlassian.net/browse/DX-2331

PR #2886, opened 2023-11-20
Initial implementation of `state export log`.

mitchell-as:


internal/runners/export/log.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// Filter down the list based on the given prefix.
+	filteredLogFiles := []*logFile{}
+	regex, err := regexp.Compile(params.Prefix + `-\d+-(\d+)\.log`)
+	if err != nil {
+		return ErrInvalidLogPrefix
+	}
+	for _, file := range logFiles {
+		if regex.MatchString(file) {
+			timestamp, err := strconv.Atoi(regex.FindStringSubmatch(file)[1])
```
MDrakos, line 59: Does the `regex.MatchString` guarantee that we will find a second group? If not the indexing of the slice here is a potential panic.

MDrakos (changes requested):
Looks great, just one question.

internal/runners/export/log.go
mitchell-as: Yes, if there is a match, the capture from the pattern guarantees a valid index.

MDrakos (approved):

PR #2887, opened 2023-11-21
Consistently use single quotes in actionable commands in user-facing messages.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2006" title="DX-2006" target="_blank">TaskDX-2006</a>  Replace backticks/tildes in locale messages with single quotes
  </td></table>
  <br />
 

Shells can interpret backticks in special circumstances, so use quotes to be safe.

### DIFF
--- a/internal/runbits/git/git.go
+++ b/internal/runbits/git/git.go
@@ -65,7 +65,7 @@ func (r *Repo) CloneProject(owner, name, path string, out output.Outputer, an an
 		err = locale.WrapError(err, "err_clone_repo", "Could not clone repository with URL: {{.V0}}, error received: {{.V1}}.", *project.RepoURL, err.Error())
 		tipMsg := locale.Tl(
 			"err_tip_git_ssh-add",
-			"If you are using an SSH key please ensure it's configured by running `[ACTIONABLE]ssh-add <path-to-key>[/RESET]`.",
+			"If you are using an SSH key please ensure it's configured by running '[ACTIONABLE]ssh-add <path-to-key>[/RESET]'.",
 		)
 		return errs.AddTips(err, tipMsg)
 	}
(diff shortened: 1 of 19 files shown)


internal/runbits/git/git.go
```diff
@@ -65,7 +65,7 @@ func (r *Repo) CloneProject(owner, name, path string, out output.Outputer, an an
 		err = locale.WrapError(err, "err_clone_repo", "Could not clone repository with URL: {{.V0}}, error received: {{.V1}}.", *project.RepoURL, err.Error())
 		tipMsg := locale.Tl(
 			"err_tip_git_ssh-add",
-			"If you are using an SSH key please ensure it's configured by running `[ACTIONABLE]ssh-add <path-to-key>[/RESET]`.",
+			"If you are using an SSH key please ensure it's configured by running '[ACTIONABLE]ssh-add <path-to-key>[/RESET]'.",
```
MDrakos, line 68: This localization has the quotes before the tag whereas some others have it after the tag and around the command. Normally, I wouldn't be this nitpicky but since this PR is about consistency we should make sure they follow the same pattern. Additionally, is there any difference to the output where the quotes are placed? I assume not, but we should verify just to be sure.

MDrakos (changes requested):

internal/runbits/git/git.go
mitchell-as: You are right. I got confused reading my own ticket, thinking that single quotes either inside or outside the tags are okay (i.e. leave them be position-wise, just make sure they're single quotes).

However I agree that we should be consistent. The quotes should be outside the tags so that only the actionable text is highlighted. It makes it more clear what the user should type (no confusion on whether or not to type/include quotes, haha).

MDrakos (approved):

PR #2899, opened 2023-11-27
Platform signup URL should be constructed based on any host override.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2326" title="DX-2326" target="_blank">BugDX-2326</a>  `state auth signup` does not honour `ACTIVESTATE_API_HOST` like `state auth` does
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/auth/login.go
+++ b/internal/runbits/auth/login.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"net/url"
+	"os"
 	"time"
 
 	"github.com/ActiveState/cli/internal/constants"
@@ -222,9 +223,14 @@ func authenticateWithBrowser(out output.Outputer, auth *authentication.Auth, pro
 			return errs.Wrap(err, "Verification URL is not valid")
 		}
 
-		signupURL, err := url.Parse(constants.PlatformSignupURL)
+		apiHost := constants.DefaultAPIHost
+		if hostOverride := os.Getenv(constants.APIHostEnvVarName); hostOverride != "" {
+			apiHost = hostOverride
+		}
+		fullURL := "https://" + apiHost + constants.PlatformSignupPath
+		signupURL, err := url.Parse(fullURL)
 		if err != nil {
-			return errs.Wrap(err, "constants.PlatformSignupURL is not valid")
+			return errs.Wrap(err, "Platform signup URL '%s' is not valid", fullURL)
 		}
 		query := signupURL.Query()
 		query.Add("nextRoute", parsedURL.RequestURI())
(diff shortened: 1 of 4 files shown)


internal/runners/tutorial/tutorial.go
```diff
@@ -158,9 +158,10 @@ func (t *Tutorial) authFlow() error {
 		}
 	case signUpBrowser:
 		t.analytics.EventWithLabel(anaConsts.CatTutorial, "authentication-action", "sign-up-browser")
-		err := open.Run(constants.PlatformSignupURL)
+		signupURL := "https://" + constants.DefaultAPIHost + constants.PlatformSignupPath
+		err := open.Run(signupURL)
```
mitchell-as, line 162: I honestly don't think there's any utility to making the tutorial signup URL respect a host override because new users use `state tutorial`, and they will *never* need a host override.

internal/runbits/auth/login.go
```diff
@@ -222,9 +223,14 @@ func authenticateWithBrowser(out output.Outputer, auth *authentication.Auth, pro
 			return errs.Wrap(err, "Verification URL is not valid")
 		}
 
-		signupURL, err := url.Parse(constants.PlatformSignupURL)
+		apiHost := constants.DefaultAPIHost
+		if hostOverride := os.Getenv(constants.APIHostEnvVarName); hostOverride != "" {
+			apiHost = hostOverride
+		}
+		fullURL := "https://" + apiHost + constants.PlatformSignupPath
+		signupURL, err := url.Parse(fullURL)
```
MDrakos, line 231: I think this is a good opportunity to add a function to the `api` package like `GetPlatformURL`. That way we can try to centralize usage of the host override env var to one package, rather than potentially sprinkle it throughout the codebase.

internal/runners/tutorial/tutorial.go
MDrakos: External users yes, but internal users _may_ want to test changes using the state tool. It would be a similar scenario that spawned this story.

MDrakos (changes requested):

internal/runners/tutorial/tutorial.go
mitchell-as: Fair enough. I've added a new method to the `api` package. Let me know if the naming can be improved.

pkg/platform/api/settings.go
```diff
@@ -136,3 +136,17 @@ func getProjectHost(service Service) *string {
 
 	return &url.Host
 }
+
+// GetURL returns a generic Platform URL for the given path.
+// This is for retrieving non-service URLs (e.g. signup URL).
+func GetURL(path string) *url.URL {
```
MDrakos, line 142: nit: Maybe `GetPlatformURL`?

MDrakos (approved):

MDrakos (approved):

PR #2900, opened 2023-11-27
Fixed version number on v43 branch.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2329" title="DX-2329" target="_blank">BugDX-2329</a>  Wrong build name (0.42...) in the `version/0-43-0-RC1` branch
  </td></table>
  <br />
 

I checked S3 and the correct folder (0.43.0) is populated (not 0.42.0).

We are going to have to try and prevent this one from propagating up when it's merged, or else it'll overwrite master's v44 version.txt.


MDrakos (approved):
Good catch. Yeah we will have to be careful with v43 ticket to ensure this doesn't change.

PR #2904, opened 2023-11-28
Do not query for build when staging a commit

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2324" title="DX-2324" target="_blank">TaskDX-2324</a>  We no longer ask for the build when staging a commit
  </td></table>
  <br />
 

It turns out that while we are querying for the build when we stage a commit we are not using it so the changeset is fairly small.


MDrakos:
Test failures are just timeouts waiting for the build to finish.

mitchell-as (approved):

PR #2910, opened 2023-11-29
Add integration test for "runtime in use" error.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2331" title="DX-2331" target="_blank">TaskDX-2331</a>  Devise an integration test for asserting an error updating a runtime when it is in use
  </td></table>
  <br />


MDrakos (approved):
Nice!

PR #2911, opened 2023-11-29
Better formatting for multi-line error messages

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2278" title="DX-2278" target="_blank">TaskDX-2278</a>  We have a better way of showing multi line errors
  </td></table>
  <br />
 

After trying to find a solution in the `output` package, it turns out this can be more easily, and simply, accomplished by modifying how the error messages are marshalled.

### DIFF
--- a/internal/runbits/errors/errors.go
+++ b/internal/runbits/errors/errors.go
@@ -45,7 +45,7 @@ func (o *OutputError) MarshalOutput(f output.Format) interface{} {
 	if errors.As(o.error, &userFacingError) {
 		message := userFacingError.UserError()
 		if f == output.PlainFormatName {
-			outLines = append(outLines, fmt.Sprintf(" [NOTICE][ERROR]x[/RESET] %s", message))
+			outLines = append(outLines, formatMessage(message)...)
 		} else {
 			outLines = append(outLines, message)
 		}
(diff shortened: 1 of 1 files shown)


internal/runbits/errors/errors.go
```diff
@@ -82,6 +82,20 @@ func (o *OutputError) MarshalOutput(f output.Format) interface{} {
 	return strings.Join(outLines, "\n")
 }
 
+func formatMessage(message string) []string {
```
mitchell-as, line 85: Please add a brief comment about what this function does. The name is not specific enough.

mitchell-as (changes requested):
This is clever. We do some word wrapping, so do we care if wrapped lines are not indented? i.e. error messages with long lines will wrap and not indent, while multi-line error messages will indent.

Now that we are starting on user-facing errors with a single error being shown in most cases, we may want to consider dropping the " x " thing and just show the error message unindented and unformatted like in many terminal programs.

MDrakos:
> This is clever. We do some word wrapping, so do we care if wrapped lines are not indented? i.e. error messages with long lines will wrap and not indent, while multi-line error messages will indent.
> 
> Now that we are starting on user-facing errors with a single error being shown in most cases, we may want to consider dropping the " x " thing and just show the error message unindented and unformatted like in many terminal programs.

Yeah, I agree that the word wrapping should also be indented for consistency but I'm inclined to make that a separate follow-up story as it affects more than just the error output. It will also have to make the `colorize` package aware of indentation which I believe will not be straightforward. I'll spend some time looking into it and file a follow-up if necessary.

I also agree that removing the `x` is something we should consider and will file a story for us to discuss at our next stand.

mitchell-as (approved):

PR #2914, opened 2023-11-30
Add CI Linting step

MDrakos:


MDrakos:
Going to close this as I can't get it to trigger on the formatting change and it's proving to be more effort than it's worth.

PR #2916, opened 2023-12-01
Avoid redundant "State Tool is operating on project..." in tcsh.

mitchell-as:


MDrakos (approved):
Wow, quite the workaround but great work with the research on this one.

PR #2918, opened 2023-12-04
Use real installation for uninstall integration tests.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2296" title="DX-2296" target="_blank">TaskDX-2296</a>  Uninstalls correctly assert environment changes
  </td></table>
  <br />

### DIFF
--- a/test/integration/install_scripts_int_test.go
+++ b/test/integration/install_scripts_int_test.go
@@ -86,27 +86,25 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
 				argsWithActive = append(argsWithActive, "-c", cmd)
 			}
 
+			// Make the directory to install to.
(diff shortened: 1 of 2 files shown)


test/integration/install_scripts_int_test.go
```diff
@@ -86,27 +86,25 @@ func (suite *InstallScriptsIntegrationTestSuite) TestInstall() {
 				argsWithActive = append(argsWithActive, "-c", cmd)
 			}
 
+			// Make the directory to install to.
```
mitchell-as, line 89: The uninstall integration test invokes an actual installer instead of mocking an install, and that invocation is based on this code with some improvements. I copied over those improvements here.

MDrakos (approved):
Nice work!

PR #2921, opened 2023-12-05
Clean up GitHub Actions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2339" title="DX-2339" target="_blank">TaskDX-2339</a>  Audit use of `state run preprocess` in github actions
  </td></table>
  <br />
 

This task turned into a general GitHub Actions cleanup. It really needed to be done.

### DIFF
--- a/.github/workflows/propagate.yml
+++ b/.github/workflows/propagate.yml
@@ -13,14 +13,7 @@ jobs:
   propagate:
     name: Propagate to affected version branches
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        go-version:
-          - 1.20.x
     env:
-      ACTIVESTATE_CI: true
-      ACTIVESTATE_CLI_DISABLE_RUNTIME: true
-      SHELL: bash
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       JIRA_USERNAME: ${{ secrets.JIRA_EMAIL }}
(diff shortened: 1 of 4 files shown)


activestate.yaml
```diff
@@ -46,14 +46,6 @@ constants:
     if: ne .OS.Name "Windows"
     value: .sh
 scripts:
-  - name: install-deps
-    language: bash
-    if: ne .Shell "cmd"
-    value: |
-      if ! type "gozip" &> /dev/null; then
```
mitchell-as, line 53: This was originally needed for the offline-installer, but that was moved to its own repo, so this is no longer needed.

.github/workflows/propagate.yml
```diff
@@ -13,14 +13,7 @@ jobs:
   propagate:
     name: Propagate to affected version branches
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        go-version:
-          - 1.20.x
     env:
-      ACTIVESTATE_CI: true
```
MDrakos, line 21: It makes sense to remove this one here but I think we should include it on workflows that use the state tool as consumers of our analytics could be using this now or may want to in the future.

MDrakos: Just searched and it looks like it's all there. Sorry, when I saw it removed I didn't look in the other files.

MDrakos (approved):

PR #2922, opened 2023-12-05
Clean up locale file by removing unused entries.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2348" title="DX-2348" target="_blank">TaskDX-2348</a>  Remove unused locale entries
  </td></table>
  <br />
 

The logic behind which keys to remove is outlined in a comment in the ticket.


MDrakos (approved):
Wow nice!

PR #2929, opened 2023-12-05
Make package filter case-insensitive.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-401" title="DX-401" target="_blank">BugDX-401</a>  Filtering packages in a project is case-sensitive
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/list.go
+++ b/internal/runners/packages/list.go
@@ -162,7 +162,7 @@ func newFilteredRequirementsTable(requirements []*gqlModel.Requirement, filter s
 
 	rows := make([]packageRow, 0, len(requirements))
 	for _, req := range requirements {
-		if !strings.Contains(req.Requirement, filter) {
+		if !strings.Contains(strings.ToLower(req.Requirement), strings.ToLower(filter)) {
 			continue
 		}
 


mitchell-as:

internal/runners/packages/list.go
```diff
@@ -162,7 +162,7 @@ func newFilteredRequirementsTable(requirements []*gqlModel.Requirement, filter s
 
 	rows := make([]packageRow, 0, len(requirements))
 	for _, req := range requirements {
-		if !strings.Contains(req.Requirement, filter) {
+		if !strings.Contains(strings.ToLower(req.Requirement), strings.ToLower(filter)) {
```
MDrakos, line 165: I think this is a good case for `strings.EqualFold`

MDrakos (changes requested):

internal/runners/packages/list.go
mitchell-as: I think we want to have a substring search though. If you look at the ticket you will see the user expected that searching for 'flask' and 'Flask' would result in matches for any package with the 'flask-' or 'Flask-' prefix.

MDrakos: Ahh I see, okay makes sense 👍

MDrakos (approved):

PR #2933, opened 2023-12-06
Propagate `locale.Tl()` to `locale.Tr()` changes.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2352" title="DX-2352" target="_blank">TaskDX-2352</a>  Propagate DX-1157
  </td></table>
  <br />
 

They failed to propagate to master because their ticket has an "Any" fix version we could not get rid of.


mitchell-as:
This will need a force merge because I don't have permission.

PR #2937, opened 2023-12-07
Removed unused binaries.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2347" title="DX-2347" target="_blank">TaskDX-2347</a>  Remove packr and upx from .github/deps/<os>/bin/
  </td></table>
  <br />


MDrakos (approved):
👍

PR #2941, opened 2023-12-12
Sandbox test environment

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2285" title="DX-2285" target="_blank">TaskDX-2285</a>  Our integration test environment is completely sandboxed
  </td></table>
  <br />


.github/workflows/build.yml
```diff
@@ -258,7 +264,7 @@ jobs:
 
       - # === Deploy for Integration Tests # NEVER run this against production branches. This is meant for PR deployments. ===
         name: Deploy for Integration Tests # NEVER run this against production branches. This is meant for PR deployments.
-        if: "!contains(fromJSON('[\"refs/heads/beta\", \"refs/heads/release\", \"refs/heads/LTS\"]'), github.ref)"
+        if: '!contains(fromJSON(''["refs/heads/beta", "refs/heads/release", "refs/heads/LTS"]''), github.ref)'
```
mitchell-as, line 267: Is this really allowed? It has me worried, because it doesn't look syntactically valid.

internal/testhelpers/e2e/env_windows.go
```diff
@@ -0,0 +1,37 @@
+//go:build windows
```
mitchell-as, line 1: Have you tried running e2e tests locally with these changes? The variables defined here may be too finely-tuned for GitHub Actions CI.

mitchell-as (changes requested):
This looks quite good overall, with my only real concern being that it might be too finely-tuned for running on GitHub Actions CI that it might not run on our systems locally (at least for Windows).

internal/testhelpers/e2e/env_windows.go
MDrakos: Yes, they ran fine on my local machine. We also can disable the sandbox with an env var if it's a problem. I expect there to be some issues with this and we can change it to opt-in as well as fine-tune the sandboxing if/when issues pop up.

MDrakos: Oh and as for the env vars on Windows they appear to be fairly standard, except for the msys one which should not only be prepended when on CI.

mitchell-as (approved):
Excellent work. I know this was a major pain, but kudos for sticking with it!

MDrakos:
Test failures should be unrelated to this PR

mitchell-as (approved):

PR #2945, opened 2023-12-12
Added comprehensive test for determining the current project.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2360" title="DX-2360" target="_blank">StoryDX-2360</a>  We have a test that checks our project sourcing priorities
  </td></table>
  <br />
 

I also upgraded the existing unit test for this.

### DIFF
--- a/pkg/projectfile/projectfile.go
+++ b/pkg/projectfile/projectfile.go
@@ -758,6 +758,10 @@ func (p *Project) SetBranch(branch string) error {
 }
 
 // GetProjectFilePath returns the path to the project activestate.yaml
+// It considers projects in the following order:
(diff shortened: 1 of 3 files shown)


MDrakos (changes requested):
While I think the test you've setup is great it makes more sense as an integration test. When we're setting env vars in the test because certain commands do it signals to me that the behavior should be covered by an integration test. Right now, if we were to inadvertently change the behavior of `state shell` this unit test wouldn't fail, whereas if this was an integration test we would catch it.

MDrakos (approved):
Looks good to me! Feel free to remove the unit test if you like.

PR #2948, opened 2023-12-14
Added `state revert HEAD` to revert latest commit.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-136" title="DX-136" target="_blank">StoryDX-136</a>  `state revert HEAD` reverts the HEAD commit
  </td></table>
  <br />

### DIFF
--- a/internal/runners/revert/revert.go
+++ b/internal/runners/revert/revert.go
@@ -53,32 +53,39 @@ func New(prime primeable) *Revert {
 	}
 }
 
+const headCommitID = "HEAD"
+
 func (r *Revert) Run(params *Params) (rerr error) {
 	defer rationalizeError(&rerr)
 
 	if r.project == nil {
 		return locale.NewInputError("err_no_project")
 	}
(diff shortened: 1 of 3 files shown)


internal/runners/revert/revert.go
```diff
@@ -53,32 +53,39 @@ func New(prime primeable) *Revert {
...
 func (r *Revert) Run(params *Params) (rerr error) {
 	defer rationalizeError(&rerr)
 
 	if r.project == nil {
 		return locale.NewInputError("err_no_project")
 	}
-	if !strfmt.IsUUID(params.CommitID) {
+
+	commitID := params.CommitID
+	if !strfmt.IsUUID(commitID) && commitID != headCommitID {
```
MDrakos, line 66: I think here and the other equality check with the constant should be case insensitive. I think `strings.EqualFold` will get you there.

MDrakos (changes requested):

MDrakos (approved):

PR #2951, opened 2023-12-15
`state publish`

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2334" title="DX-2334" target="_blank">TaskDX-2334</a>  Reapply change set for `state publish`
  </td></table>
  <br />
 

Try again.

This was the problem: https://github.com/ActiveState/cli/pull/2951/commits/049d1aa238eb8821d780a75b8dad25cd5167cd75. I'm honestly not sure why the tests were passing before vs. after this change. Perhaps something on the platform side changed with regards to typing. `strfmt.DateTime` appears to be equivalent to `time.Time` from Go's perspective, but perhaps serialization changes in some way. Anyway, I'm seeing some recent, unrelated test failures that are due to platform changes, indicating some flux that could be causing trouble here. I'll file tickets to investigate those.


mitchell-as:

MDrakos (approved):
Nice find!

mitchell-as:
@MDrakos please re-review. I found another set of integration test failures that is fixed with: https://github.com/ActiveState/cli/pull/2951/commits/021913e2f50b1585d29fc2d15977158da10bdaff

MDrakos (approved):
I'm good to merge but I did add a comment that is worth considering.

mitchell-as:
Okay, I've added comments.

MDrakos (approved):
Works for me!

PR #2953, opened 2023-12-15
Fix nightly integration test failure

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2369" title="DX-2369" target="_blank">BugDX-2369</a>  Nightly failure: Artifact failed to build
  </td></table>
  <br />


mitchell-as (approved):
Wow, so simple!

PR #2959, opened 2023-12-19
Requirement operations do not set commit ID on build failure

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2343" title="DX-2343" target="_blank">TaskDX-2343</a>  On a requirement change, we update the commit ID after successfully sourcing the runtime
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/refresh.go
+++ b/internal/runbits/refresh.go
@@ -29,9 +29,14 @@ func RefreshRuntime(
 	//if err != nil {
 	//	return locale.WrapError(err, "err_update_build_script")
 	//}
-	target := target.NewProjectTarget(proj, nil, trigger)
+	var t *target.ProjectTarget
+	if proj != nil && *proj.Namespace().CommitID != commitID {
+		t = target.NewProjectTarget(proj, &commitID, trigger)
+	} else {
+		t = target.NewProjectTarget(proj, nil, trigger)
+	}
 	isCached := true
-	rt, err := runtime.New(target, an, svcm, auth)
+	rt, err := runtime.New(t, an, svcm, auth)
 	if err != nil {
 		if runtime.IsNeedsUpdateError(err) {
 			isCached = false
--- a/internal/runbits/refresh.go
+++ b/internal/runbits/refresh.go
@@ -1,8 +1,6 @@
 package runbits
 
 import (
-	"strings"
-
 	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/output"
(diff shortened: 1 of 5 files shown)


internal/runbits/requirements/requirements.go
```diff
@@ -297,10 +297,6 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 	//	return errs.Wrap(err, "Could not get remote build expr")
 	//}
 
-	if err := commitmediator.Set(r.Project, commitID.String()); err != nil {
-		return locale.WrapError(err, "err_package_update_commit_id")
-	}
-
 	// Note: a commit ID file needs to exist at this point.
```
mitchell-as, line 300: I think we need to move this commented out chunk of code to below the new location of `commitmediator.Set()`. As the top comment says, a commitID file needs to exist at this point.

MDrakos:
Test failures do not appear to be related to this PR

internal/runbits/refresh.go
```diff
@@ -29,9 +29,14 @@ func RefreshRuntime(
 	//if err != nil {
 	//	return locale.WrapError(err, "err_update_build_script")
 	//}
-	target := target.NewProjectTarget(proj, nil, trigger)
+	var t *target.ProjectTarget
```
mitchell-as, line 32: Can we simplify this to something like `target := target.NewProjectTarget(proj, commitIDFromProject(proj), trigger)` where `commitIDFromProject()` returns a `*strfmt.UUID`? I know the name probably doesn't make sense, but perhaps there is one that's better.

Single-character variable names are a bit icky and the duplicated `target.NewProjectTarget()` with one different argument does not seem ideal.

internal/runbits/requirements/requirements.go
```diff
@@ -297,6 +297,12 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 	//	return errs.Wrap(err, "Could not get remote build expr")
 	//}
 
+	// refresh or install runtime
+	err = runbits.RefreshRuntime(r.Auth, r.Output, r.Analytics, r.Project, commitID, true, trigger, r.SvcModel)
+	if err != nil {
+		return handleRefreshError(err, r.Project, parentCommitID)
```
mitchell-as, line 303: If this is not a build error, the commit ID will be updated, but the buildscript will not be updated on line 310 below (commented for now, but it will eventually be uncommented when local orders are restored).

Perhaps we need to break out lines 306-315 into a separate function and call that function on line 306 and inside `handleRefreshError()`.

mitchell-as (changes requested):

mitchell-as (approved):
Looks great!

PR #2963, opened 2023-12-19
`state update` should clean up after itself.

mitchell-as:


internal/updater/updater.go
```diff
@@ -230,6 +232,19 @@ func (u *UpdateInstaller) InstallBlocking(installTargetPath string, args ...stri
...
+	updateDir := filepath.Dir(filepath.Dir(installerPath))
+	logging.Debug("Cleaning up temporary update directory: %s", updateDir)
+	if strings.HasPrefix(filepath.Base(updateDir), "state-update") {
+		err = os.RemoveAll(updateDir)
+		if err != nil {
+			multilog.Error("Unable to remove update directory '%s': %v", updateDir, err)
+		}
+	} else {
+		multilog.Error("Update directory structure has changed without updating it here, too. "+
+			"installerPath: %s\nupdateDir: %s\nExpected a 'state-update' prefix for the latter", installerPath, updateDir)
```
MDrakos, line 245: I can see us changing the structure of updates and getting a bunch of these errors on Rollbar from older versions. I think the error above is enough to report.

MDrakos (changes requested):

MDrakos (approved):

PR #2965, opened 2023-12-19
Added `state languages search`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-75" title="DX-75" target="_blank">StoryDX-75</a>  We have a way to discover languages
  </td></table>
  <br />

### DIFF
--- a/internal/language/language.go
+++ b/internal/language/language.go
@@ -120,7 +120,7 @@ func MakeByName(name string) Language {
 }
 
 // MakeByNameAndVersion will retrieve a language by a given name and version.
-func MakeByNameAndVersion(name, version string) (Language, error) {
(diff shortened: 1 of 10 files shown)


internal/language/language.go
```diff
@@ -120,7 +120,7 @@ func MakeByName(name string) Language {
 }
 
 // MakeByNameAndVersion will retrieve a language by a given name and version.
-func MakeByNameAndVersion(name, version string) (Language, error) {
```
mitchell-as, line 123: We were never returning an error here, so just remove it to simplify the API.

MDrakos (approved):

PR #2967, opened 2023-12-20
Do not show EOL platforms as available options in `state platforms search`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-952" title="DX-952" target="_blank">BugDX-952</a>  CLI - PLATFORMS: `state platforms search` provides platforms that are not supported anymore.
  </td></table>
  <br />


MDrakos (approved):
Nice! A bit surprised the API has this info.

PR #2968, opened 2023-12-20
Allow a "reconfigure" installation.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1189" title="DX-1189" target="_blank">StoryDX-1189</a>  Installer should reinstall if current install is not configured
  </td></table>
  <br />
 

State Tool might be installed, but not on PATH. If that's the case, allow installation to proceed so PATH is reconfigured.

Only error out if State Tool is installed and properly configured on PATH.


MDrakos (approved):
👍

PR #2969, opened 2023-12-20
Make parentCommit required in StageCommit endpoint.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2375" title="DX-2375" target="_blank">BugDX-2375</a>  StageCommit mutation request should assume that a parentCommitId is required.
  </td></table>
  <br />


mitchell-as:
Test failures appear unrelated to this PR. They're the usual timeouts.

MDrakos (approved):
Huh, I thought a change like this would invalidate the query as it no longer matches the schema... but the integration tests prove me wrong. Good to know.

PR #2970, opened 2023-12-20
Allow glibc version on checkout

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-774" title="DX-774" target="_blank">StoryDX-774</a>  When I state activate a project with multiple versions of a OS, I want to be able specify the OS version
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/runtime/rationalize.go
+++ b/internal/runbits/runtime/rationalize.go
@@ -49,17 +49,12 @@ func rationalizeError(auth *authentication.Auth, proj *project.Project, rerr *er
 				errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch,
 				proj.BranchName(), strings.Join(branches, "\n - ")))
 		} else {
-			if errNoMatchingPlatform.LibcVersion != "" {
-				*rerr = errs.NewUserFacing(
-					locale.Tr("err_no_platform_data_remains", errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch),
-					errs.SetInput(),
-					errs.SetTips(locale.Tr("err_user_libc_solution", api.GetPlatformURL(fmt.Sprintf("%s/%s", proj.NamespaceString(), "customize")).String())),
-				)
-			} else {
-				*rerr = errs.NewUserFacing(locale.Tr(
-					"err_no_platform_data_remains",
-					errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch))
-			}
+			libcErr := errNoMatchingPlatform.LibcVersion != ""
+			*rerr = errs.NewUserFacing(
+				locale.Tr("err_no_platform_data_remains", errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch),
+				errs.SetIf(libcErr, errs.SetInput()),
+				errs.SetIf(libcErr, errs.SetTips(locale.Tr("err_user_libc_solution", api.GetPlatformURL(fmt.Sprintf("%s/%s", proj.NamespaceString(), "customize")).String()))),
+			)
 		}
 
 	// If there was an artifact download error, say so, rather than reporting a generic "could not
--- a/internal/runbits/runtime/rationalize.go
+++ b/internal/runbits/runtime/rationalize.go
@@ -2,12 +2,14 @@ package runtime
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/multilog"
+	"github.com/ActiveState/cli/pkg/platform/api"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/runtime/setup"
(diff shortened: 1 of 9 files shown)


internal/runbits/runtime/rationalize.go
```diff
@@ -47,9 +49,17 @@ func rationalizeError(auth *authentication.Auth, proj *project.Project, rerr *er
 				errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch,
 				proj.BranchName(), strings.Join(branches, "\n - ")))
 		} else {
-			*rerr = errs.NewUserFacing(locale.Tr(
-				"err_no_platform_data_remains",
-				errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch))
+			if errNoMatchingPlatform.LibcVersion != "" {
```
mitchell-as, line 52: Nathan created `errs.SetIf()` for this this kind of `if` statement: https://github.com/ActiveState/cli/blob/301fe307423cc054daf44f06eea26e95c3c75bf8/internal/runbits/checkout/rationalize.go#L26. Can we do something like:

```
libCError := errNoMatchingPlatform.LibcVersion != ""
*rerr = errs.NewUserFacing(
	locale.Tr("err_no_platform_data_remains", errNoMatchingPlatform.HostPlatform, errNoMatchingPlatform.HostArch),
	errs.SetIf(libCError, errs.SetInput()),
	errs.SetIf(libCError, errs.SetTips(locale.Tr("err_user_libc_solution", api.GetPlatformURL(fmt.Sprintf("%s/%s", proj.NamespaceString(), "customize")).String()))),
)
```
?

internal/runners/checkout/checkout.go
```diff
@@ -83,6 +88,33 @@ func (u *Checkout) Run(params *Params) (rerr error) {
 		return locale.WrapError(err, "err_project_frompath")
 	}
 
+	if params.LibcVersion != "" {
```
mitchell-as, line 91: Nit: It may be worth putting this all in a separate function and calling it from here to keep the function smaller.

pkg/platform/model/inventory.go
```diff
@@ -271,7 +282,21 @@ func filterPlatformIDs(hostPlatform, hostArch string, platformIDs []strfmt.UUID)
 		return fallback, nil
 	}
 
-	return nil, &ErrNoMatchingPlatform{hostPlatform, hostArch}
+	return nil, &ErrNoMatchingPlatform{hostPlatform, hostArch, libcVersion}
+}
+
+func fetchLibcVersion() string {
+	var libcVersion string
```
mitchell-as, line 289: What do you think about simplifying this to:

```
if runtime.GOOS != "linux" {
    return ""
}

libcInfo, err := sysinfo.Libc()
...
return libcInfo.Version()
```
?

internal/runners/checkout/checkout.go
```diff
@@ -83,6 +88,33 @@ func (u *Checkout) Run(params *Params) (rerr error) {
...
+		if err != nil {
+			return locale.WrapError(err, "err_libc_version_invalid", "libc version must be in the form of major.minor")
+		}
+
+		minor, err := strconv.Atoi(parts[1])
+		if err != nil {
+			return locale.WrapError(err, "err_libc_version_invalid", "libc version must be in the form of major.minor")
+		}
+
+		sysinfo.SetLibcInfo(&sysinfo.LibcInfo{
```
mitchell-as, line 111: Cool, I didn't know this was a thing!

mitchell-as (changes requested):
This looks good. I've added a few code structure suggestions to consider.

internal/runbits/runtime/rationalize.go
MDrakos: Very nice! Didn't know about this and improves the readability quite a bit!

pkg/platform/model/inventory.go
MDrakos: I always forget about that pattern when it's my own code :P

internal/runners/checkout/checkout.go
```diff
@@ -125,3 +135,36 @@ func (u *Checkout) Run(params *Params) (rerr error) {
...
 	return nil
 }
+
+func setLibcVersion(libcVersion string) error {
+	if libcVersion == "" {
+		return nil
+	}
+
+	if rt.GOOS != "linux" {
+		return locale.NewError("err_libc_version_not_supported", "libc version is only supported on linux")
```
mitchell-as, line 145: Sorry if I missed this the first time, but aren't all the errors in this function supposed to be input errors?

pkg/platform/model/inventory.go
```diff
@@ -271,7 +282,24 @@ func filterPlatformIDs(hostPlatform, hostArch string, platformIDs []strfmt.UUID)
...
+	}
+
+	var libcVersion string
+
+	libcInfo, err := sysinfo.Libc()
+	if err != nil {
+		logging.Error("Failed to fetch libc info: %v", err)
+		return ""
+	}
+	libcVersion = libcInfo.Version()
```
mitchell-as, line 300: nit: I think you can remove the `var libcVersion string` above and use `:=` here.

mitchell-as (changes requested):
I think I missed something in the previous review. Sorry about that.

internal/runners/checkout/checkout.go
MDrakos: Yes, you're right. Thanks for catching that!

MDrakos:
I missed an integration test failure that called out a flaw in my approach. I've updated the `sysinfo` API to avoid the failure.

mitchell-as (approved):

PR #2973, opened 2023-12-21
Disable ruby tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2383" title="DX-2383" target="_blank">BugDX-2383</a>  Nightly failure: TestPackageIntegrationTestSuite/TestRuby
  </td></table>
  <br />

### DIFF
--- a/test/integration/package_int_test.go
+++ b/test/integration/package_int_test.go
@@ -594,6 +594,9 @@ func (suite *PackageIntegrationTestSuite) TestUpdate() {
 }
 
 func (suite *PackageIntegrationTestSuite) TestRuby() {
+	// TODO: Re-enable this test when Ruby is supported on the Platform
+	// https://activestatef.atlassian.net/browse/DX-2384
+	suite.T().Skip("Ruby is currently not building on the Platform")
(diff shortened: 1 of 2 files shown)


test/integration/package_int_test.go
```diff
@@ -594,6 +594,9 @@ func (suite *PackageIntegrationTestSuite) TestUpdate() {
 }
 
 func (suite *PackageIntegrationTestSuite) TestRuby() {
+	// TODO: Re-enable this test when Ruby is supported on the Platform
+	// https://activestatef.atlassian.net/browse/DX-2384
+	suite.T().Skip("Ruby is currently not building on the Platform")
```
mitchell-as, line 599: Nit: technically it works on Linux. It's only Windows that's causing a failure (since macOS is skipped).

mitchell-as (approved):

test/integration/package_int_test.go
MDrakos: Good point.

mitchell-as (approved):

PR #2977, opened 2023-12-28
Error out if the commit of the namespace to activate doesn't match the checked out project.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1392" title="DX-1392" target="_blank">BugDX-1392</a>  ACTIVATE: Inexistent commit_id not identified and command executed with no errors.
  </td></table>
  <br />
 

This is similar to `state use` and `state shell` with mismatched commit IDs.

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1294,6 +1294,8 @@ err_use_default_project_does_not_exist:
   other: "Cannot find your project. Please either check it out again with [ACTIONABLE]state checkout[/RESET] or run [ACTIONABLE]state use reset[/RESET] to stop using it."
 err_shell_commit_id_mismatch:
   other: "Cannot start a shell/prompt for the given commit ID. Please use '[ACTIONABLE]state checkout[/RESET]' to check out a specific commit ID and then try again."
+err_activate_commit_id_mismatch:
+  other: "Cannot activate a project with the given commit ID. Please use '[ACTIONABLE]state checkout[/RESET]' to check out a specific commit ID and then try again."
 err_shell_cannot_load_project:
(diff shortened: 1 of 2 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1294,6 +1294,8 @@ err_use_default_project_does_not_exist:
   other: "Cannot find your project. Please either check it out again with [ACTIONABLE]state checkout[/RESET] or run [ACTIONABLE]state use reset[/RESET] to stop using it."
 err_shell_commit_id_mismatch:
   other: "Cannot start a shell/prompt for the given commit ID. Please use '[ACTIONABLE]state checkout[/RESET]' to check out a specific commit ID and then try again."
+err_activate_commit_id_mismatch:
+  other: "Cannot activate a project with the given commit ID. Please use '[ACTIONABLE]state checkout[/RESET]' to check out a specific commit ID and then try again."
```
MDrakos, line 1298: Nit:
```suggestion
  other: "Cannot activate a project with the given commit ID. Please verify the commit ID {{.V0}} is correct and then try again."
```

I believe the only way this can happen is if the commit ID was passed in via the command.

MDrakos (approved):

internal/locale/locales/en-us.yaml
mitchell-as: That is true, but if the commitID provided on the command line is valid (but not checked out), then the suggested error message will cause confusion ("I have verified the commit ID. It exists in the history and on the dashboard, but `state` won't except it. Boo").

This message matches the message formats for `state shell` and `state use`, which also accept optional commitIDs via command line.

PR #2979, opened 2024-01-02
Use "channel" instead of "branch" when referring to State Tool versions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-207" title="DX-207" target="_blank">BugDX-207</a>  State tool consistent terminology (Branch VS Channel)
  </td></table>
  <br />
 

This tends to cause confusion because Platform projects have branches.

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -242,7 +242,7 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 
 		if childCmd.Name() != "update" && pj != nil && pj.IsLocked() {
 			if (pj.Version() != "" && pj.Version() != constants.Version) ||
-				(pj.VersionBranch() != "" && pj.VersionBranch() != constants.ChannelName) {
+				(pj.Channel() != "" && pj.Channel() != constants.ChannelName) {
 				return errs.AddTips(
 					locale.NewInputError("lock_version_mismatch", "", pj.Source().Lock, constants.ChannelName, constants.Version),
 					locale.Tr("lock_update_legacy_version", constants.DocumentationURLLocking),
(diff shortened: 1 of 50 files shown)


mitchell-as:
Initial critical and installer tests passed. I've now kicked off "Test: all", but I'm confident this is ready for review.

mitchell-as:
Failures are the usual timeouts and temporary Platform errors.

cmd/state/main.go
```diff
@@ -242,9 +242,9 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 
 		if childCmd.Name() != "update" && pj != nil && pj.IsLocked() {
 			if (pj.Version() != "" && pj.Version() != constants.Version) ||
-				(pj.VersionBranch() != "" && pj.VersionBranch() != constants.BranchName) {
+				(pj.VersionBranch() != "" && pj.VersionBranch() != constants.ChannelName) {
```
MDrakos, line 245: Should the project methods here also be renamed?

MDrakos (dismissed):
Minor nit comment but otherwise looks fine to me.

cmd/state/main.go
mitchell-as: No. You got caught by the purpose of this ticket: to eliminate confusion between channels (state tool version branches) and branches (Platform things). The project methods refer to Platform project branches, so they should not be changed.

mitchell-as: I was going to say no, but you're right. I thought all projectfile branch methods were related to Platform project branches, but apparently there's a way to lock projectfiles to State Tool channels. I learned something new.

MDrakos (approved):

PR #2980, opened 2024-01-02
Ensure log file handle is closed on exit.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-394" title="DX-394" target="_blank">BugDX-394</a>  Global log file is not properly closed
  </td></table>
  <br />


MDrakos (approved):
This _shouldn't_ cause any issues but you may want to run some integration tests against this PR just to be safe.

PR #2982, opened 2024-01-02
Make identifier distinct in error message.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2381" title="DX-2381" target="_blank">BugDX-2381</a>  SWITCH: Error messages during `state switch` to non-existent branch/commit not consistent with emphasized by a quote of the parameter's names.
  </td></table>
  <br />
 

I chose a `:` over quotes because our error messages tend to prefer this over strings that end in quotes in Go (e.g. an apostrophe and then a double-quote).

### DIFF
--- a/internal/runners/swtch/switch.go
+++ b/internal/runners/swtch/switch.go
@@ -92,7 +92,7 @@ func (s *Switch) Run(params SwitchParams) error {
 
 	identifier, err := resolveIdentifier(project, params.Identifier)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


MDrakos (changes requested):
I think to make this consistent with other error messages related to branch names it should be quoted. Will quoting this clash with other potential error messages?

MDrakos (approved):

PR #2983, opened 2024-01-02
Delete failed checkouts due to libc errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2386" title="DX-2386" target="_blank">BugDX-2386</a>  CHECKOUT: In case `--runtime-env-libc-version` flag used and fail it still create a folder
  </td></table>
  <br />


MDrakos (approved):
image

PR #2984, opened 2024-01-02
Generate versioninfo for some binaries

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-748" title="DX-748" target="_blank">BugDX-748</a>  CLI - PROC: Missing metadata for "state-svc" and "state" processes
  </td></table>
  <br />

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
 name: Build-Test-Deploy
 
 # === Triggers ===
-"on":
+'on':
   push:
     branches:
       - master
@@ -20,7 +20,7 @@ name: Build-Test-Deploy
 # === Workflow Permissions ===
 permissions:
   id-token: write # This is required for requesting the JWT
-  contents: read  # This is required for actions/checkout
+  contents: read # This is required for actions/checkout
 
 # === Workflow-level environment variables ===
 env:
@@ -29,7 +29,6 @@ env:
 
 # === JOBS ===
 jobs:
-
   # === OS Specific Job (runs on each OS) ===
   os_specific:
     name: ${{ matrix.sys.os }}
@@ -39,9 +38,9 @@ jobs:
         go-version:
           - 1.20.x
         sys:
-          - {os: ubuntu-20.04}
-          - {os: macos-11, shell: zsh}
-          - {os: windows-2019}
+          - { os: ubuntu-20.04 }
+          - { os: macos-11, shell: zsh }
+          - { os: windows-2019 }
       fail-fast: false
     runs-on: ${{ matrix.sys.os }}
     env:
@@ -105,6 +104,7 @@ jobs:
         shell: pwsh
         run: |
           echo "${PSScriptRoot}/.github/deps/${{ runner.os }}/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          state run install-deps-ci
 
       - # == Setup macOS ==
         name: Setup (macOS)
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -105,6 +104,7 @@ jobs:
         shell: pwsh
         run: |
           echo "${PSScriptRoot}/.github/deps/${{ runner.os }}/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+          state run install-deps-ci
```
mitchell-as, line 107: Can we make this the only change in the file? The change on line 261 makes me nervous.

activestate.yaml
```diff
@@ -67,25 +67,44 @@ scripts:
         echo "actionlint was not found on your PATH, installing .."
         go install github.com/rhysd/actionlint/cmd/actionlint@v1.6.23
       fi
+      if { [[ "$GOOS" == "windows" ]] || [[ "$OS" == "Windows_NT" ]]; } && ! type "goversioninfo" &> /dev/null; then
+        echo "actionlint was not found on your PATH, installing .."
```
mitchell-as, line 71: This echo is incorrect.

activestate.yaml
```diff
@@ -67,25 +67,44 @@ scripts:
...
     language: bash
     if: ne .OS.Name "Linux"
     description: Install OS specific deps
     value: ""
+  - name: install-deps-ci
+    language: bash
+    if: ne .Shell "cmd"
+    value: |
+      if { [[ "$GOOS" == "windows" ]] || [[ "$OS" == "Windows_NT" ]]; } && ! type "goversioninfo" &> /dev/null; then
+        echo "actionlint was not found on your PATH, installing .."
```
mitchell-as, line 84: This echo is also incorrect.

activestate.yaml
```diff
@@ -67,25 +67,44 @@ scripts:
...
+        go run scripts/versioninfo-generator/main.go version.txt cmd/state-svc/versioninfo.json "State Service"
+      fi
   - name: build
     language: bash
     description: Builds the project with the host OS as the target OS.
     value: |
       set -e
       $constants.SET_ENV
 
+      pushd cmd/state > /dev/null
```
mitchell-as, line 105: Please add a comment that this is to generate syso from versioninfo.txt (if that's indeed what it does).

activestate.yaml
```diff
@@ -104,6 +123,9 @@ scripts:
       set -e
       $constants.SET_ENV
 
+      pushd cmd/state-svc > /dev/null
```
mitchell-as, line 126: Please add a similar comment.

internal/assets/contents/versioninfo.json
```diff
@@ -0,0 +1,41 @@
...
+    "FileType": "01",
+    "FileSubType": "00"
+  },
+  "StringFileInfo": {
+    "Comments": "",
+    "CompanyName": "ActiveState Software Inc.",
+    "FileDescription": "",
+    "FileVersion": "",
+    "InternalName": "",
+    "LegalCopyright": "© ActiveState Software, Inc. 2024",
```
mitchell-as, line 25: Are we going to have to remember to change this every year? If so, please change the generator to modify this field. Perhaps omit the year here and append it there.

scripts/versioninfo-generator/main.go
```diff
@@ -0,0 +1,159 @@
...
+func run() error {
+	if len(os.Args) != 4 {
+		return errs.New("Usage: versioninfo-generator <version.txt> <versioninfo.json>")
+	}
+
+	versionFile := os.Args[1]
+	versionInfoFilePath := os.Args[2]
+	productName := os.Args[3]
+
+	// Read version.txt
```
mitchell-as, line 98: I really like the comments in this function!

mitchell-as (changes requested):
This is cool! There are a few minor things to address before merging.

.github/workflows/build.yml
MDrakos: Bah, that's my damn YAML formatter.

mitchell-as (approved):
Looks great!

PR #2990, opened 2024-01-05
The CI environment sandbox needs to pass through `ACTIVESTATE_CI`.

mitchell-as:
Otherwise analytics will be skewed.


mitchell-as:
Wrong branch name.

PR #2993, opened 2024-01-05
Fix failing integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2432" title="DX-2432" target="_blank">BugDX-2432</a>  Nightly failure: TestShellIntegrationTestSuite
  </td></table>
  <br />
 

This also fixes DX-2429 and DX-2431

Since the sandboxed test environment doesn't have a default python on `PATH` we have to enable the runtime setup. It's likely that these tests weren't erroring out before because of the CI runner environment.

### DIFF
--- a/test/integration/run_int_test.go
+++ b/test/integration/run_int_test.go
@@ -232,9 +232,11 @@ func (suite *RunIntegrationTestSuite) TestRun_Unauthenticated() {
 
 	suite.createProjectFile(ts, 2)
 
-	cp := ts.SpawnWithOpts(e2e.OptArgs("activate"))
-	cp.Expect("Skipping runtime setup")
-	cp.Expect("Activated")
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("activate"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
+	)
+	cp.Expect("Activated", e2e.RuntimeSourcingTimeoutOpt)
 	cp.ExpectInput(termtest.OptExpectTimeout(10 * time.Second))
 
 	cp.SendLine(fmt.Sprintf("%s run testMultipleLanguages", cp.Executable()))
--- a/test/integration/run_int_test.go
+++ b/test/integration/run_int_test.go
@@ -236,7 +236,7 @@ func (suite *RunIntegrationTestSuite) TestRun_Unauthenticated() {
 		e2e.OptArgs("activate"),
 		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
 	)
-	cp.Expect("Activated")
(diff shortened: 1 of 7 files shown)


Naatan (changes requested):
We disable the runtime because the test doesn't need them and is otherwise needlessly slowed down. If these tests were failing because of a missing runtime we should look into why they needed them, because for most of these I don't think they should actually need one.

MDrakos:
When we checkout or activate a runtime one of the events is to print out a message. This is done by invoking the `python` (or other language) executable. If we do not setup the runtime then this executable isn't available and the checkout will return with a non-zero exit code. The sandboxed environment does not have a system python on `PATH` which is likely why these tests were passed before and are not now.

Naatan (changes requested):
Ok but the point still stands that these tests aren't testing the runtime and we want them to be performant. So whatever solution we come up with should still satisfy those conditions.

Two potential solutions perhaps:

- We could disable this message when this env var is set
- We could allow tests to explicitly pass in the system python for these kinds of cases

MDrakos:
Our integration tests are in a weird place between integration tests and end-to-end tests. The spirit of sandboxing was to give us a test environment that more accurately represents a system where the state tool would be run.

I would argue that disabling the runtime for the majority of the tests isn't what we want to do. We are sacrificing accuracy for convenience though this is more in line with an integration test rather than an end-to-end test. Further, by adding an event that uses the runtime, every test that activates or checks out a project indirectly tests the runtime.

That's just my two cents and in the end, I would rather continue down the road we are on. Just food for thought. I've gone with your first suggestion of disabling the events via an env var which is set by default in the test environment.

Naatan:
For tracking; filed followup for the performance issue here: https://activestatef.atlassian.net/browse/DX-2435

test/integration/run_int_test.go
```diff
@@ -232,8 +232,10 @@ func (suite *RunIntegrationTestSuite) TestRun_Unauthenticated() {
 
 	suite.createProjectFile(ts, 2)
 
-	cp := ts.SpawnWithOpts(e2e.OptArgs("activate"))
-	cp.Expect("Skipping runtime setup")
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("activate"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
+	)
 	cp.Expect("Activated")
```
Naatan, line 239:
```suggestion
	cp.Expect("Activated", e2e.RuntimeSourcingTimeoutOpt)
```

Naatan (changes requested):

Naatan (approved):

PR #2994, opened 2024-01-06
Fix edit integration test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2430" title="DX-2430" target="_blank">BugDX-2430</a>  Nightly failure: TestEditIntegrationTestSuite/Test{Edit,Edit_UpdateCorrectPlatform}
  </td></table>
  <br />
 

It turns out that this test uses the `go` binary so we have to do some inspection of the environment CI is running on and use that in the test.

### DIFF
--- a/internal/testhelpers/e2e/env.go
+++ b/internal/testhelpers/e2e/env.go
@@ -3,8 +3,10 @@ package e2e
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/ActiveState/cli/internal/constants"
@@ -23,6 +25,10 @@ func sandboxedTestEnvironment(t *testing.T, dirs *Dirs, updatePath bool, extraEn
 	if value := os.Getenv(constants.ActiveStateCIEnvVarName); value != "" {
 		env = append(env, fmt.Sprintf("%s=%s", constants.ActiveStateCIEnvVarName, value))
 	}
+
+	// add go binary to PATH
+	goBinary := goBinaryPath()
+	basePath = fmt.Sprintf("%s%s%s", basePath, string(os.PathListSeparator), filepath.Dir(goBinary))
 
 	env = append(env, []string{
 		constants.ConfigEnvVarName + "=" + dirs.Config,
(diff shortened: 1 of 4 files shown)


test/integration/edit_int_test.go
```diff
@@ -43,14 +44,28 @@ scripts:
...
+		// The LOCALAPPDATA env var is used by the go binary to locate the cache dir
+		env = append(env, "LOCALAPPDATA="+ts.Dirs.TempDir)
 	}
+
+	// Locate the go binary as it will not be on PATH in the sandboxed environment
+	cmd := exec.Command(locator, "go")
+	output, err := cmd.Output()
+	suite.Require().NoError(err, "go binary not found in path: %s", string(output))
+	goBinary := string(output)
+	goBinary = strings.TrimSpace(string(goBinary))
```
mitchell-as, line 62: I can see other tests needing this, so perhaps we should move this into the testhelpers/e2e package? You might have to update the default session env var list though. It might get messy.

mitchell-as (changes requested):

internal/testhelpers/e2e/env.go
```diff
@@ -92,5 +98,20 @@ func prepareHomeDir(dir string) error {
...
 
+func goBinaryPath(t *testing.T) string {
+	locator := "which"
+	if runtime.GOOS == "windows" {
+		locator = "where"
+	}
+	cmd := exec.Command(locator, "go")
+	output, err := cmd.Output()
+	if err != nil {
+		t.Log("Could not find go binary")
```
mitchell-as, line 111: Is it worth logging the error message here?

internal/testhelpers/e2e/env_windows.go
```diff
@@ -35,6 +35,7 @@ func platformSpecificEnv(dirs *Dirs) []string {
 		// Other environment variables are commonly set by CI systems, but this one is not.
 		// This is requried for some tests in order to get the correct powershell output.
 		fmt.Sprintf("PSModulePath=%s", os.Getenv("PSModulePath")),
+		fmt.Sprintf("LOCALAPPDATA=%s", dirs.TempDir),
```
mitchell-as, line 38: Can we annotate why this env var is needed too?

mitchell-as (changes requested):
Almost there. This looks much better!

MDrakos:
Closing this and will bring it in with the new sandboxing PR

PR #3004, opened 2024-01-09
No need to have the update checker because we do autoupdate on `state` invocation.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2391" title="DX-2391" target="_blank">BugDX-2391</a>  Same message printed twice when update available
  </td></table>
  <br />
 

Autoupdate was removed in https://github.com/ActiveState/cli/pull/1360 (May 2021) in favor of update checking, and then re-added in https://github.com/ActiveState/cli/pull/1527 (Sept 2021), but update checking was left in place.

This allows for update notices twice in certain circumstances (`autoupdate` config option is `false` and you are running one of: `state {activate,checkout,run,use,--version}`).

Since we run autoupdate upon `state` invocation, no matter the command, there's no need for manual update checking anymore, and the user will only ever see one notice at a time per command.

### DIFF
--- a/test/integration/update_int_test.go
+++ b/test/integration/update_int_test.go
@@ -57,6 +57,7 @@ func (suite *UpdateIntegrationTestSuite) env(disableUpdates, forceUpdate bool) [
 	}
 
 	if forceUpdate {
+		env = append(env, constants.TestAutoUpdateEnvVarName+"=true")
 		env = append(env, constants.ForceUpdateEnvVarName+"=true")
 	}
 
@@ -102,6 +103,11 @@ func (suite *UpdateIntegrationTestSuite) TestUpdateAvailable() {
 
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
+
+	cfg, err := config.NewCustom(ts.Dirs.Config, singlethread.New(), true)
+	suite.Require().NoError(err)
+	defer cfg.Close()
+	cfg.Set(constants.AutoUpdateConfigKey, "false")
 
 	search, found := "Update Available", false
 	for i := 0; i < 4; i++ {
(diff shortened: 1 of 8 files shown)


test/integration/update_int_test.go
```diff
@@ -103,6 +104,11 @@ func (suite *UpdateIntegrationTestSuite) TestUpdateAvailable() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
+	cfg, err := config.NewCustom(ts.Dirs.Config, singlethread.New(), true)
+	suite.Require().NoError(err)
+	defer cfg.Close()
+	cfg.Set(constants.AutoUpdateConfigKey, "false")
```
mitchell-as, line 110: Since autoupdate always happens, we have a config key to avoid going through with it. When that config key is enabled, we just output a notice that there's an update available. So this block effectively tests both that there's an update available, and that the config key prevents the update from happening.

mitchell-as:
Note: the lone macOS test failure is unrelated, and probably sporadic.

Naatan (changes requested):
Auto updates can be disabled, so the notifier can still play its part. How about we enable it only if auto updates are disabled?

https://github.com/ActiveState/cli/blob/c33086f331a6f8ab958ae33eb01c4e709ccd78a8/cmd/state/autoupdate.go#L37-L37

mitchell-as:
Even if autoupdates are disabled, the autoupdate mechanic will still check and notify the user that there are autoupdates: https://github.com/ActiveState/cli/blob/c33086f331a6f8ab958ae33eb01c4e709ccd78a8/cmd/state/autoupdate.go#L60-L66. The update checker is redundant.

Naatan (dismissed):

PR #3005, opened 2024-01-09
Revert sandboxing

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2442" title="DX-2442" target="_blank">TaskDX-2442</a>  Consider rolling back CI integration test sandbox changes and into a new, comprehensive PR
  </td></table>
  <br />


MDrakos:
It looks like there are some legitimate test failures. I don't see how they could be caused by this PR but I'm open to fixing them here or getting this in and fixing them through our normal process.

mitchell-as (approved):
Let's get this in, and then you can unabandon any new nightly failures that are occurring here and investigate.

PR #3014, opened 2024-01-12
Buildexpressions now have a default "atTime".

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2425" title="DX-2425" target="_blank">TaskDX-2425</a>  Update timestamp handling when staging commits
  </td></table>
  <br />
 

No need to fetch and set timestamps.

### DIFF
--- a/pkg/platform/runtime/buildexpression/testdata/buildexpression.json
+++ b/pkg/platform/runtime/buildexpression/testdata/buildexpression.json
@@ -2,7 +2,7 @@
   "let": {
     "runtime": {
       "solve_legacy": {
-        "at_time": "2023-04-27T17:30:05.999000Z",
+        "at_time": "$at_time",
         "build_flags": [],
         "camel_flags": [],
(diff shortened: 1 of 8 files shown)


pkg/platform/runtime/buildexpression/testdata/buildexpression.json
```diff
@@ -2,7 +2,7 @@
   "let": {
     "runtime": {
       "solve_legacy": {
-        "at_time": "2023-04-27T17:30:05.999000Z",
+        "at_time": "$at_time",
```
mitchell-as, line 5: I'm not changing all test cases because I think this field can be either `$at_time` or a valid timestamp, and we want to be okay with both cases.

mitchell-as:
The lone test failure is unrelated to this PR.

MDrakos (approved):
Looks good, nice work!

PR #3015, opened 2024-01-12
Use config for glibc version

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2443" title="DX-2443" target="_blank">StoryDX-2443</a>  As a user I can specify the preferred glibc version for my machine
  </td></table>
  <br />

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -443,6 +443,9 @@ const ReportErrorsConfig = "report.errors"
 // ReportAnalyticsConfig is the config key used to determine if we will send analytics reports
 const ReportAnalyticsConfig = "report.analytics"
 
+// PreferredGlibcVersionConfig is the config key used to determine the preferred glibc version
+const PreferredGlibcVersionConfig = "runimte.preferred.glibc"
+
 // SvcAppName is the name we give our state-svc application
 const SvcAppName = "State Service"
 
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -444,7 +444,7 @@ const ReportErrorsConfig = "report.errors"
 const ReportAnalyticsConfig = "report.analytics"
 
 // PreferredGlibcVersionConfig is the config key used to determine the preferred glibc version
-const PreferredGlibcVersionConfig = "runimte.preferred.glibc"
+const PreferredGlibcVersionConfig = "runtime.preferred.glibc"
 
 // SvcAppName is the name we give our state-svc application
 const SvcAppName = "State Service"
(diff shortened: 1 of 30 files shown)


mitchell-as (changes requested):
JIRA says this should be targeted at v43, but the PR is based on v44 and pulling a lot of unnecessary stuff in.

MDrakos:
Whoops. Target & Verify hasn't been working for me for a while so I've just been doing this manually.

internal/constants/constants.go
```diff
@@ -443,6 +443,9 @@ const ReportErrorsConfig = "report.errors"
 // ReportAnalyticsConfig is the config key used to determine if we will send analytics reports
 const ReportAnalyticsConfig = "report.analytics"
 
+// PreferredGlibcVersionConfig is the config key used to determine the preferred glibc version
+const PreferredGlibcVersionConfig = "runimte.preferred.glibc"
```
mitchell-as, line 447: There's a typo in this config key name.

pkg/platform/model/inventory.go
```diff
@@ -284,17 +293,17 @@ func filterPlatformIDs(hostPlatform, hostArch string, platformIDs []strfmt.UUID)
...
+func fetchLibcVersion() (string, error) {
 	if runtime.GOOS != "linux" {
-		return ""
+		return "", nil
 	}
 
-	libcInfo := sysinfo.GetRequestedLibcInfo()
-	if libcInfo == nil {
-		return ""
+	cfg, err := config.New()
```
mitchell-as, line 301: Is this really okay? Or do we need to be passing the config instance from main all the way to here?

mitchell-as (changes requested):
This looks mostly fine. I'm guessing this config is system-wide, so a user cannot have one project with one preferred libc version, and another project with the default libc version. Is that something we don't care about supporting?

pkg/platform/model/inventory.go
MDrakos: Updated to pass the config instance down from main.

MDrakos:
> I'm guessing this config is system-wide, so a user cannot have one project with one preferred libc version, and another project with the default libc version. Is that something we don't care about supporting?

For now, we've opted to go with a "per-machine" config. See comment here: https://activestatef.atlassian.net/browse/DX-2389?focusedCommentId=49153

MDrakos:
Closing in favor of https://github.com/ActiveState/cli/pull/3019

PR #3017, opened 2024-01-12
Initial ability to list package versions installed instead of just "Auto".

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-942" title="DX-942" target="_blank">StoryDX-942</a>  `state packages` should show what's installed, not what's ordered
  </td></table>
  <br />
 

Before:

```
>state packages
█ Listing Packages

Operating on project ActiveState/cli, located at C:\Users\me\cli.

  Name        Version
───────────────────────
  libsass     Auto
  pexpect     4.4.0
  psutil      Auto
  requests    2.25.0
```

After:

```
>build\state.exe packages
█ Listing Packages

Operating on project ActiveState/cli, located at C:\Users\me\cli.

Skipping runtime setup because it was disabled by an environment variable
  Name        Version
─────────────────────────────
  libsass     Auto → 0.21.0
  pexpect     4.4.0
  psutil      Auto → 5.9.0
  requests    2.25.0
```

PR #3018, opened 2024-01-12
Added runtime method to determine resolved package versions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2408" title="DX-2408" target="_blank">StoryDX-2408</a>  We have a runtime mechanic for resolving requested versions to actual versions
  </td></table>
  <br />
 

Demo in action: #3017 

We don't need to do any normalization at the moment because any inputs would be coming from Platform API calls (e.g. in the demo above, names come from requirements). If we want to do any normalization (e.g. from user input), we already have inventory.go's `model.FetchNormalizedName()`, which can be used in conjunction with this function if need be.

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -26,6 +26,7 @@ import (
 	bpModel "github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildscript"
 	"github.com/ActiveState/cli/pkg/platform/runtime/envdef"
 	"github.com/ActiveState/cli/pkg/platform/runtime/setup"
@@ -36,13 +37,14 @@ import (
 )
 
 type Runtime struct {
-	disabled  bool
-	target    setup.Targeter
-	store     *store.Store
-	analytics analytics.Dispatcher
-	svcm      *model.SvcModel
-	auth      *authentication.Auth
-	completed bool
+	disabled          bool
+	target            setup.Targeter
+	store             *store.Store
+	analytics         analytics.Dispatcher
+	svcm              *model.SvcModel
+	auth              *authentication.Auth
+	completed         bool
+	resolvedArtifacts []artifact.Artifact
 }
 
 // NeedsUpdateError is an error returned when the runtime is not completely installed yet.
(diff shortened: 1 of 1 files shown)


mitchell-as:
I'm not 100% sure this satisfies the ACs, but I'm putting it up for review for feedback on what I'm missing.

Naatan (changes requested):
This is useful for the `state install` use-case bit for cases like `state packages` I think we should be able to more proactively provide a mapping of packages and their resolved versions. Two things I'd like to see this PR include:

1. The runtime mechanic should already know which packages were requested. So from an external point of view I should just be able to call something like `ResolvedVersions()` which returns a map of all the normalized package names and their resolved version.
  * I realize normalization might be a bit too foreign to the runtime mechanic, so possibly we can drop that AC and handle that at a different level.
2. We should not be re-initializing the buildplan for each resolution. I think we already parse it whenever the runtime is read? If not we can probably add that logic somewhere (in an on-demand type of way).

mitchell-as:
Okay, I changed the method to be `ResolvedArtifacts()`, which among other things, contains the resolved version number for packages. The returned name is the normalized version, so callers are responsible for normalizing names prior to comparisons. I've updated #3017 to utilize the new method (https://github.com/ActiveState/cli/pull/3017/files#diff-67826c0e717c1b1a1a3072382df29bb908a524e39bc69d3742619b79b306c943R205-R211)

pkg/platform/runtime/runtime.go
```diff
@@ -356,3 +358,29 @@ func (r *Runtime) ExecutableDirs() (envdef.ExecutablePaths, error) {
...
+	if r.resolvedArtifacts == nil {
+		runtimeStore := r.store
+		if runtimeStore == nil {
+			runtimeStore = store.New(r.target.Dir())
+		}
+
+		plan, err := runtimeStore.BuildPlan()
+		if err != nil {
+			return nil, errs.Wrap(err, "Unable to fetch build plan")
+		}
```
Naatan, line 372: This would still mean the buildplan gets parsed twice if it was used for other purposes (eg. runtime was updated). Given that this is a costly operation we may want to introduce some caching here too? Suggest checking with Mike, he may have some ideas/opinions.

This is based on my assumption that parsing buildplans still has a performance impact that is noticeable by users. If I'm wrong on that I'll drop this concern.

Naatan (changes requested):

pkg/platform/runtime/runtime.go
mitchell-as: After a discussion on Slack, Mike said "As far as calling this method goes I don’t think it’s a big performance issue. This is returning the build structure which is created by parsing the build plan, so we aren’t parsing again to get it."

The function is just unmarshaling json into a Go struct representing an already parsed build plan.

Naatan (approved):

PR #3020, opened 2024-01-15
Show resolved language, package, bundle, etc. versions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1965" title="DX-1965" target="_blank">StoryDX-1965</a>  As a user I can see both the requested and the resulting version of requirements
  </td></table>
  <br />
 

Before:

```
>state packages
█ Listing Packages

Operating on project ActiveState/cli, located at C:\Users\me\cli.

  Name        Version
───────────────────────
  libsass     Auto
  pexpect     4.4.0
  psutil      Auto
  requests    2.25.0
```

After:

```
>build\state.exe packages
█ Listing Packages

Operating on project ActiveState/cli, located at C:\Users\me\cli.

Skipping runtime setup because it was disabled by an environment variable
  Name        Version
─────────────────────────────
  libsass     Auto → 0.21.0
  pexpect     4.4.0
  psutil      Auto → 5.9.0
  requests    2.25.0
```

Also:

```
>state init org/project --language python@3.10

[...]

>state languages
█ Listing Languages (Beta)

  Name      Version
───────────────────────────
  Python    3.10 → 3.10.0
```

Note: `state platforms` already displays resolved versions. Also, platforms are not artifacts, so we cannot obtain version numbers from the build plan.

### DIFF
--- a/internal/runners/languages/languages.go
+++ b/internal/runners/languages/languages.go
@@ -3,18 +3,26 @@ package languages
 import (
 	"strings"
 
+	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/runbits/commitmediator"
+	"github.com/ActiveState/cli/internal/runbits/runtime"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/target"
 	"github.com/ActiveState/cli/pkg/project"
 )
 
 // Languages manages the listing execution context.
 type Languages struct {
-	out     output.Outputer
-	project *project.Project
+	out       output.Outputer
+	project   *project.Project
+	analytics analytics.Dispatcher
+	svcModel  *model.SvcModel
+	auth      *authentication.Auth
 }
 
 // NewLanguages prepares a list execution context for use.
@@ -22,6 +30,9 @@ func NewLanguages(prime primeable) *Languages {
 	return &Languages{
 		prime.Output(),
 		prime.Project(),
+		prime.Analytics(),
+		prime.SvcModel(),
+		prime.Auth(),
 	}
 }
 
@@ -51,7 +62,18 @@ func (l *Languages) Run() error {
 		return locale.WrapError(err, "err_fetching_languages", "Cannot obtain languages")
 	}
 
+	rt, err := runtime.NewFromProject(l.project, target.TriggerLanguage, l.analytics, l.svcModel, l.out, l.auth)
+	if err != nil {
+		multilog.Error("Unable to initialize runtime for version resolution: %v", errs.JoinMessage(err))
+	}
+
 	for i := range langs {
+		if resolvedVersion, err := rt.ResolveArtifactVersion(model.NewNamespaceLanguage(), langs[i].Name); err == nil {
+			langs[i].Version = locale.Tr("constraint_auto_resolved", resolvedVersion)
+		} else {
+			multilog.Error("Unable to resolve language version: %v", errs.JoinMessage(err))
+			langs[i].Version = locale.T("constraint_auto")
+		}
 		langs[i].Name = strings.Title(langs[i].Name)
 	}
 
--- a/internal/runners/languages/languages.go
+++ b/internal/runners/languages/languages.go
@@ -67,12 +67,21 @@ func (l *Languages) Run() error {
 		multilog.Error("Unable to initialize runtime for version resolution: %v", errs.JoinMessage(err))
 	}
 
+	artifacts, err := rt.ResolvedArtifacts()
+	if err != nil {
+		multilog.Error("Unable to retrieve runtime resolved artifact list: %v", errs.JoinMessage(err))
+	}
+	ns := model.NewNamespaceLanguage()
+
 	for i := range langs {
-		if resolvedVersion, err := rt.ResolveArtifactVersion(model.NewNamespaceLanguage(), langs[i].Name); err == nil {
-			langs[i].Version = locale.Tr("constraint_auto_resolved", resolvedVersion)
-		} else {
-			multilog.Error("Unable to resolve language version: %v", errs.JoinMessage(err))
+		if langs[i].Version == "" {
 			langs[i].Version = locale.T("constraint_auto")
+			for _, a := range artifacts {
+				if a.Namespace == ns.String() && a.Name == langs[i].Name {
+					langs[i].Version = locale.Tr("constraint_auto_resolved", *a.Version)
+					break
+				}
+			}
 		}
 		langs[i].Name = strings.Title(langs[i].Name)
 	}
--- a/internal/runners/languages/languages.go
+++ b/internal/runners/languages/languages.go
@@ -42,6 +42,12 @@ type languagePlainOutput struct {
 	Version string `locale:"version"`
 }
 
+type languageOutput struct {
+	Name            string `json:"name"`
+	Version         string `json:"version"`
+	ResolvedVersion string `json:"resolved_version"`
+}
+
 // Run executes the list behavior.
 func (l *Languages) Run() error {
 	if l.project == nil {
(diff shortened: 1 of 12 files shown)


internal/runners/languages/languages.go
```diff
@@ -51,8 +68,41 @@ func (l *Languages) Run() error {
 		return locale.WrapError(err, "err_fetching_languages", "Cannot obtain languages")
 	}
 
+	// Fetch resolved artifacts list for showing full version numbers.
+	// Note: any errors here are not fatal, and only some of them should be reported to rollbar.
```
Naatan, line 72: I'm not sure I agree with this. Sure we could just show the languages without resolved version information, but we changed our stance on implicit errors on this quite a while ago to be explicit. ie. error out so we can fix the issue once and for all.

I won't mark each logged error in the review, but assume most if not all of them should be changed to error returns for this review iteration.

internal/runners/languages/languages.go
```diff
@@ -51,8 +68,41 @@ func (l *Languages) Run() error {
...
+	}
+	ns := model.NewNamespaceLanguage()
+
 	for i := range langs {
-		langs[i].Name = strings.Title(langs[i].Name)
+		name := langs[i].Name
+		version := langs[i].Version
+		constraints := langs[i].VersionConstraints()
+
+		langs[i].Name = strings.Title(name)
```
Naatan, line 89: This would break structured output I think.

pkg/platform/model/checkpoints.go
```diff
@@ -78,15 +61,9 @@ func FetchLanguagesForCommit(commitID strfmt.UUID) ([]Language, error) {
...
-			}
-			if lang.Version == "" {
-				version, err := getVersionFromConstraints(requirement.VersionConstraints)
-				if err != nil {
-					return nil, errs.Wrap(err, "Unable to determine commit's language version based on constraints")
-				}
-				lang.Version = version
+				Name:               requirement.Requirement,
+				Version:            requirement.VersionConstraint,
+				versionConstraints: requirement.VersionConstraints,
```
Naatan, line 66: You can use `VersionConstraint` here. We should not be playing with the structured version constraints on the client as it makes us too opinionated and prone to backward incompatible changes.

I know it says `VersionConstraint` is deprecated; but this entire API will be replaced long before `VersionConstraint` will be fully dropped.

pkg/platform/model/buildplanner.go
```diff
@@ -544,9 +546,46 @@ func VersionStringToRequirements(version string) ([]bpModel.VersionRequirement,
...
+		case bpModel.ComparatorLT:
+			parts[i] = fmt.Sprintf("<%s", version)
+		case bpModel.ComparatorLTE:
+			parts[i] = fmt.Sprintf("<=%s", version)
+		case bpModel.ComparatorNE:
+			parts[i] = fmt.Sprintf("!%s", version)
+		}
+	}
+	return strings.Join(parts, ",")
+}
```
Naatan, line 591: You should be able to drop this by using `VersionConstraint` (string) as mentioned in my other comment. I'd like to avoid encoding/decoding constraints on the client.

internal/runners/packages/list.go
```diff
@@ -154,12 +189,23 @@ type packageRow struct {
 	Version string `json:"version" locale:"package_version,Version"`
 }
 
-func newFilteredRequirementsTable(requirements []*gqlModel.Requirement, filter string, nstype model.NamespaceType) []packageRow {
+func newFilteredRequirementsTable(requirements []*gqlModel.Requirement, filter string, nstype model.NamespaceType, rt *runtime.Runtime, ns *model.Namespace) []packageRow {
```
Naatan, line 192: Nit: Passing in the resolved artifacts instead feels cleaner.

internal/runners/packages/list.go
```diff
@@ -154,12 +189,23 @@ type packageRow struct {
 	Version string `json:"version" locale:"package_version,Version"`
```
Naatan, line 189: Would be good if we can include the resolved version in the structured output as well.

internal/runners/packages/list.go
```diff
@@ -172,7 +218,18 @@ func newFilteredRequirementsTable(requirements []*gqlModel.Requirement, filter s
...
 		if versionConstraint == "" {
-			versionConstraint = "Auto"
+			versionConstraint = locale.T("constraint_auto")
+			if len(req.VersionConstraints) > 0 {
+				reqs := model.MonoModelConstraintsToRequirements(&req.VersionConstraints)
+				versionConstraint = model.RequirementsToVersionString(reqs)
+			}
+
+			for _, a := range artifacts {
+				if a.Namespace == ns.String() && a.Name == req.Requirement {
```
Naatan, line 228: Should also check if versionConstrained doesn't already equal `*a.Version` like you did with languages.

pkg/platform/runtime/store/store.go
```diff
@@ -255,7 +255,12 @@ func (s *Store) InstallPath() string {
 	return s.installPath
 }
 
+var ErrNoBuildPlanFile = errs.New("no build plan file")
+
 func (s *Store) BuildPlanRaw() ([]byte, error) {
+	if !fileutils.FileExists(s.buildPlanFile()) {
+		return nil, ErrNoBuildPlanFile
```
Naatan, line 262: The previous error would contain an IO error, which this does not. I'm not sure if that matters (ie. do we asser this anywhere?), but please verify that this isn't breaking anything.

Naatan (changes requested):

internal/runners/languages/languages.go
mitchell-as: I think I know what you mean, but this is how it was previously.

pkg/platform/model/checkpoints.go
mitchell-as: I wish we could use `requirement.VersionConstraint` (that's what I tried originally), but if you `state init org/namespace --language python@3.10`, `VersionConstraint` will be 3.10 instead (which was requested) of 3.10.0. `VersionConstraints` contains 3.10.0 (which was resolved), so that's why the extra processing is needed.

internal/runners/packages/list.go
mitchell-as: Good call.

internal/runners/packages/list.go
mitchell-as: That's a good point. We can display the "Auto -> 1.1.1" for printed output and keep separate "Auto" and "1.1.1" fields for structured output.

pkg/platform/runtime/store/store.go
mitchell-as: It's not. The only other callers end up logging the errors, not interpreting them in any other way.

internal/runners/languages/languages.go
Naatan: I think we should just print what the API returns. 

I didn't realize this is what we were doing before though. Feel free to disregard, what I care about is that structured output still works and includes the language name.

pkg/platform/model/checkpoints.go
Naatan: I'm not sure I follow. Both `VersionConstraint` and `VersionConstraints` will contain the requested version, not the resulting version. I guess we have some logic that translates it to a full semver perhaps? But either way it's still the requested version, which in this case just happens to be a full version. Point is this cannot be used to determine the resolved version, so it being incomplete shouldn't matter.

internal/runners/languages/languages.go
```diff
@@ -1,30 +1,47 @@
...
 		prime.Output(),
 		prime.Project(),
+		prime.Analytics(),
+		prime.SvcModel(),
+		prime.Auth(),
+		prime.Config(),
 	}
 }
 
+type languagePlainOutput struct {
```
mitchell-as, line 40: Note: cannot piggy-back off of `Language` struct because its `Version` and `Resolved` version fields must remain separate (e.g. "Auto" and "1.1.0") while this struct's `Version` field would be something like "Auto → 1.1.0".

internal/runners/languages/languages.go
```diff
@@ -51,10 +68,42 @@ func (l *Languages) Run() error {
 		return locale.WrapError(err, "err_fetching_languages", "Cannot obtain languages")
 	}
 
+	// Fetch resolved artifacts list for showing full version numbers.
+	rt, err := runtime.NewFromProject(l.project, target.TriggerLanguage, l.analytics, l.svcModel, l.out, l.auth, l.cfg)
+	if err != nil {
+		return locale.WrapError(err, "err_languages_runtime", "Could not initialize runtime")
```
mitchell-as, line 74: I know we're supposed to be shying away from `locale.*Error` functions in favor of user-facing errors, but this command doesn't yet have user-facing errors and implementing them here is too much for this ticket. Please let me know if there's a better way of handling this.

internal/runners/languages/languages.go
mitchell-as: I've made the change to print only what the Platform returns. In this case, users will see "python" for both plain and structured output. This is a change because they used to see "Python" for both.

pkg/platform/model/buildplanner.go
mitchell-as: I've removed this as requested.

mitchell-as:
I expect more churn.

pkg/platform/model/checkpoints.go
```diff
@@ -27,8 +27,9 @@ type Checkpoint []*mono_models.Checkpoint
 
 // Language represents a language requirement
 type Language struct {
-	Name    string `json:"name"`
-	Version string `json:"version"`
+	Name            string `json:"name"`
+	Version         string `json:"version"`
+	ResolvedVersion string `json:"resolved_version,omitempty"`
```
Naatan, line 32: This doesn't belong here imo. ResolvedVersion is a concern only of the language runner as written, so this type should exist only there. You could inherit this struct though.

internal/runners/packages/list.go
```diff
@@ -72,12 +99,77 @@ func (l *List) Run(params ListRunParams, nstype model.NamespaceType) error {
...
+		}
+		requirementsPlainOutput = append(requirementsPlainOutput, requirementPlainOutput{
+			Name:    req.Requirement,
+			Version: plainVersion,
+		})
+
+		requirementsOutput = append(requirementsOutput, requirement{
+			Name:            req.Requirement,
+			Version:         version,
+			ResolvedVersion: resolvedVersion,
```
Naatan, line 163: Because we only set resolvedVersion if it does not equal the requested version it means that the structured output does not provide the `resolved_version` in this scenario. I think for structured output we should always set it, regardless of whether it matches the requested version. I think `state languages` will have the same issue.

Naatan (changes requested):

pkg/platform/model/checkpoints.go
mitchell-as: Okay, I've removed it.

internal/runners/packages/list.go
mitchell-as: Okay, structured output now always gives the resolved version, both for `state languages` and `state packages`.

Naatan (approved):

PR #3022, opened 2024-01-15
Update state import to use the buildplanner

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2444" title="DX-2444" target="_blank">StoryDX-2444</a>  `state import` uses the BuildPlanner
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/import.go
+++ b/internal/runners/packages/import.go
@@ -14,12 +14,13 @@ import (
 	"github.com/ActiveState/cli/internal/runbits"
 	"github.com/ActiveState/cli/internal/runbits/commitmediator"
 	"github.com/ActiveState/cli/pkg/platform/api"
+	bpModel "github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
 	"github.com/ActiveState/cli/pkg/platform/api/reqsimport"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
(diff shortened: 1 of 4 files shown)


internal/runners/packages/import.go
```diff
@@ -147,20 +168,34 @@ func fetchImportChangeset(cp ChangesetProvider, file string, lang string) (model
...
 
-func commitChangeset(project *project.Project, msg string, changeset model.Changeset) (strfmt.UUID, error) {
-	localCommitID, err := commitmediator.Get(project)
-	if err != nil {
-		return "", errs.Wrap(err, "Unable to get local commit")
-	}
-	commitID, err := model.CommitChangeset(localCommitID, msg, changeset)
-	if err != nil {
-		return "", errs.AddTips(locale.WrapError(err, "err_packages_removed"),
-			locale.T("commit_failed_push_tip"),
```
mitchell-as, line 158: Please also remove this now-unused locale entry from en-us.yaml.

internal/runners/packages/import.go
```diff
@@ -147,20 +168,34 @@ func fetchImportChangeset(cp ChangesetProvider, file string, lang string) (model
...
-			locale.T("commit_failed_pull_tip"))
```
mitchell-as, line 159: Please also remove this from en-us.yaml.

mitchell-as (changes requested):
This looks good. Just a couple minor things to address.

internal/runners/packages/import.go
MDrakos: Thanks for catching these!

mitchell-as (approved):

PR #3027, opened 2024-01-16
Add vulnerabilities service client

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2447" title="DX-2447" target="_blank">TaskDX-2447</a>  Add service to fetch vulnerabilities
  </td></table>
  <br />
 

This is largely the initial changeset that @mitchell-as created with a few minor tweaks. 

I've removed the actual call to the service as there are other stories that will for that. The original call is here: https://github.com/ActiveState/cli/pull/3027/commits/f8a3cec0f3174c74da5f756f6998965726e15db3#diff-a6406350474d337b91add9f6183a8c40d3819a0852c0ac7689ec22b4f79d1103R168.

We _may_ have to add to the models for things we need in other stories, but that should be fairly easy from looking at the schema. It may require some trial and error.

Lastly, this call (and maybe all calls to this service) _require_ authentication.

### DIFF
--- a/internal/runners/packages/search.go
+++ b/internal/runners/packages/search.go
@@ -153,6 +153,19 @@ type searchOutput []searchPackageRow
 func formatSearchResults(packages []*model.IngredientAndVersion, showNamespace bool) *searchOutput {
 	rows := make(searchOutput, len(packages))
 
+	ingredients := make([]*model.VulnerabilityIngredient, len(packages))
+	for i, pack := range packages {
+		ingredients[i] = &model.VulnerabilityIngredient{
+			Namespace: *pack.Ingredient.PrimaryNamespace,
+			Name:      *pack.Ingredient.Name,
+			Version:   pack.Version,
+		}
+	}
+	_, err := model.FetchVulnerabilitiesForIngredients(ingredients)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


internal/runners/packages/search.go
```diff
@@ -150,9 +153,18 @@ type searchPackageRow struct {
...
 	rows := make(searchOutput, len(packages))
 
+	ingredients := make([]*model.VulnerabilityIngredient, len(packages))
+	for i, pack := range packages {
+		ingredients[i] = &model.VulnerabilityIngredient{
+			Namespace: *pack.Ingredient.PrimaryNamespace,
+			Name:      *pack.Ingredient.Name,
+			Version:   pack.Version,
+		}
+	}
```
Naatan, line 166: Guessing you were using this to test?

Naatan (changes requested):
Looks good! Just seems some debugging code made it in.

internal/runners/packages/search.go
MDrakos: Whoops, yeah that was for debugging.

Naatan (approved):

PR #3028, opened 2024-01-16
Fix failing integration test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2446" title="DX-2446" target="_blank">BugDX-2446</a>  Nightly failure: TestUpdateIntegrationTestSuite/TestLockUnlock
  </td></table>
  <br />
 

I spent some time trying to determine why this started failing a few days ago. I'm still not sure but regardless the error in the log is expected and should be ignored.


mitchell-as (changes requested):
Please add a comment why the error is expected.

Normally this is implicit when we `cp.ExpectNotExitCode(0)` or `cp.Expect("error message")`, but since none of that is present here, we need to document why we expect the log to have an error.

mitchell-as (approved):

PR #3032, opened 2024-01-17
Set timestamp to default on import

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2457" title="DX-2457" target="_blank">TaskDX-2457</a>  Import runner updates the existing build expression to use the `$at_time` value.
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildexpression/buildexpression.go
+++ b/pkg/platform/runtime/buildexpression/buildexpression.go
@@ -882,8 +882,8 @@ func (e *BuildExpression) SetDefaultTimestamp() error {
 			continue
 		}
 
-		if arg.Assignment.Name == "at_time" {
-			arg.Assignment.Value.Str = ptr.To("$at_time")
+		if arg.Assignment.Name == AtTimeKey {
+			arg.Assignment.Value.Str = ptr.To(fmt.Sprintf("$%s", AtTimeKey))
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -871,6 +871,25 @@ func (e *BuildExpression) removePlatform(platformID strfmt.UUID) error {
...
+	if err != nil {
+		return errs.Wrap(err, "Could not get solve node")
+	}
+
+	for _, arg := range solveNode.Arguments {
+		if arg.Assignment == nil {
+			continue
+		}
+
+		if arg.Assignment.Name == "at_time" {
```
mitchell-as, line 885: Please use the `AtTimeKey` constant here.

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -871,6 +871,25 @@ func (e *BuildExpression) removePlatform(platformID strfmt.UUID) error {
...
+			arg.Assignment.Value.Str = ptr.To("$at_time")
```
mitchell-as, line 886: I'm not sure if we should be using a constant here too, but it may be worth considering.

mitchell-as (changes requested):

mitchell-as (approved):

PR #3034, opened 2024-01-18
Set default timestamp on package operations

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2451" title="DX-2451" target="_blank">BugDX-2451</a>  Nightly failure:  TestPackageIntegrationTestSuite/TestUpdate
  </td></table>
  <br />

### DIFF
--- a/internal/testhelpers/e2e/session.go
+++ b/internal/testhelpers/e2e/session.go
@@ -69,9 +69,8 @@ var (
 	PersistentPassword string
 	PersistentToken    string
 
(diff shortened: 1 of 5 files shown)


MDrakos:
The lone test failure is for `TestPackageIntegrationTestSuite/TestResolved` which is not actually in the v43 codebase... it's part of v44 and master so I'm unsure why it's being run in this PR but I cannot add extra timeouts to fix it.

mitchell-as (approved):

PR #3047, opened 2024-01-22
Various fixes to `state publish`, mainly intended to facilitate revisions

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1885" title="DX-1885" target="_blank">TaskDX-1885</a>  The publish API automatically creates revisions as needed
  </td></table>
  <br />

### DIFF
--- a/internal/runners/publish/publish.go
+++ b/internal/runners/publish/publish.go
@@ -19,10 +19,12 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/api"
 	graphModel "github.com/ActiveState/cli/pkg/platform/api/graphql/model"
 	"github.com/ActiveState/cli/pkg/platform/api/graphql/request"
+	"github.com/ActiveState/cli/pkg/platform/api/inventory/inventory_models"
 	auth "github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/ActiveState/graphql"
+	"github.com/go-openapi/strfmt"
 	"github.com/skratchdot/open-golang/open"
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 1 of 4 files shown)


internal/runners/publish/publish.go
```diff
@@ -63,12 +65,19 @@ func New(prime primeable) *Runner {
...
 	client.SetTokenProvider(prime.Auth())
 	client.EnableDebugLog()
 	return &Runner{auth: prime.Auth(), out: prime.Output(), prompt: prime.Prompt(), project: prime.Project(), client: client}
 }
 
+type ParentIngredient struct {
+	IngredientID        strfmt.UUID
+	IngredientVersionID strfmt.UUID
+	Version             string
+	// dependencies
```
MDrakos, line 77: Not sure how helpful this comment is... unless this was just meant to create some space in the struct definition.

internal/runners/publish/publish.go
```diff
@@ -118,33 +127,54 @@ func (r *Runner) Run(params *Params) error {
...
+		if err != nil {
+			return locale.WrapInputError(err, "err_uploadingredient_getversion", "Could not grab ingredient by version")
+		}
+		ingredient = &ParentIngredient{*i.IngredientID, *i.IngredientVersionID, *i.Version, i.Dependencies}
+		isRevision = true
+	}
+
+	if ingredient == nil {
+		// Attempt to find the existing ingredient, if we didn't already get it from the version specific call above
+		ts := time.Now()
```
MDrakos, line 145: Do we want to use the latest platform timestamp here? Will using a timestamp that is potentially ahead of the platform cause issues?

MDrakos (changes requested):
Looks good to me, just some minor stuff.

internal/runners/publish/publish.go
Naatan: It shouldn't cause issues. It's essentially telling the platform "you can use ingredients as new as this date". So theoretically even a date 100 years in the future should work.

The platform timestamp wouldn't be appropriate here since that would predate the timestamp that a user published their own ingredient.

MDrakos (approved):

PR #3050, opened 2024-01-23
List additional dependencies being installed.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1974" title="DX-1974" target="_blank">StoryDX-1974</a>  As a user I can see what packages were brought in for my install request
  </td></table>
  <br />
 

Samples:
```
>..\..\cli\build\state.exe install scipy
█ Installing Package

Operating on project ActiveState-CLI/small-python, located at C:\Users\me\tmp\small-python.

 • Searching for scipy in the ActiveState Catalog ✔ Found
 • Creating commit ✔ Done
█ Updating Runtime

Changes to your runtime may require some dependencies to be rebuilt.

Setting Up Runtime
Resolving Dependencies ✔ Done

Installing scipy@1.10.1 includes 7 direct dependencies.
  ├─ ifort-distributables@2021.2.0.2871
  ├─ libmkl@2020.4
  ├─ mkl-win@2020.4.311
  ├─ numpy@1.26.3
  ├─ openssl@1.11.0.20 → openssl@3.2.0 (updated)
  ├─ sqlite3@3.41.2 → sqlite3@3.44.2 (updated)
  └─ zlib@1.2.13 → zlib@1.3 (updated)
```

```
>..\..\cli\build\state.exe install numpy
█ Installing Package

Operating on project ActiveState-CLI/small-python, located at C:\Users\me\tmp\small-python.

 • Searching for numpy in the ActiveState Catalog ✔ Found
 • Creating commit ✔ Done
█ Updating Runtime

Changes to your runtime may require some dependencies to be rebuilt.

Setting Up Runtime
Resolving Dependencies ✔ Done

Installing numpy@1.26.3 includes 7 direct dependencies, and a total of 12 direct and indirect dependencies.
  ├─ cython@0.29.36 (10 dependencies)
  ├─ flit-core@3.9.0 (4 dependencies)
  ├─ libmkl@2020.4 (1 dependencies)
  ├─ python-module-build-support@1.00 (12 dependencies)
  ├─ setuptools@62.6.0 (9 dependencies)
  ├─ toml@0.10.2 (9 dependencies)
  └─ wheel@0.41.2 (9 dependencies)
```


mitchell-as:
The lone test failure is a timeout.

mitchell-as:
Please see my comment on why the message in the ACs cannot be used as written: https://activestatef.atlassian.net/browse/DX-1974?focusedCommentId=49422

internal/runbits/requirements/requirements.go
```diff
@@ -270,6 +271,17 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 	pg.Stop(locale.T("progress_success"))
 	pg = nil
 
+	// If applicable, output a list of additional dependencies (if any) being installed along with the
+	// requested package or bundle.
+	if operation == bpModel.OperationAdded &&
```
Naatan, line 276: Updates should also be considered. Updating a package can also bring in new dependencies.

internal/runbits/requirements/requirements.go
```diff
@@ -488,3 +500,71 @@ func packageCommitMessage(op bpModel.Operation, name, version string) string {
...
 }
+
+// outputAdditionalRequirements computes and lists the additional dependencies being installed for the given package name.
+// This should only be called if a package or bundle is being added. Otherwise, the results may be nonsensical.
+func (r *RequirementOperation) outputAdditionalRequirements(parentCommitId, commitId strfmt.UUID, packageName string) (rerr error) {
+	pg := output.StartSpinner(r.Output, locale.T("progress_dependencies"), constants.TerminalAnimationInterval)
+	bp := model.NewBuildPlannerModel(r.Auth)
+
+	// Fetch old build plan with sources to compare against.
+	oldBuildPlan, err := bp.FetchBuildResult(parentCommitId, r.Project.Owner(), r.Project.Name())
```
Naatan, line 511: We should not be reaching out to the API for these buildplans, because those API's are known to be -slow-.

We have both the before and after buildplan stored locally, though there's probably an order of operations problem that'll need to be resolved. Point is though; this PR should not alter the performance of installing packages in any significant way.

internal/runbits/requirements/requirements.go
```diff
@@ -488,3 +500,71 @@ func packageCommitMessage(op bpModel.Operation, name, version string) string {
...
+		"   Installing [NOTICE]{{.V0}}@{{.V1}}[/RESET] will result in adding {{.V2}} additional dependencies.",
+		packageName, packageVersion, strconv.Itoa(len(additions)),
+	))
+	for i, req := range additions {
+		prefix := "   ├─"
+		if i == len(additions)-1 {
+			prefix = "   └─"
+		}
+		r.Output.Notice(fmt.Sprintf("  [DISABLED]%s[/RESET] %s@%s", prefix, req.Name, req.Version))
+	}
```
Naatan, line 567: I had not considered how significantly our UI has changed since this was first introduced. To address this I've created a new mock for this story:

https://docs.google.com/document/d/144wprjlt7Ta6jrFtAXG6_J9G6dEu1eYN4fIi4SVOjB0/edit

That's on me, but still I would like to handle the UI changes as part of this story. I've already updated the AC's.

Naatan (changes requested):

internal/runbits/requirements/requirements.go
mitchell-as: I've updated to use the cached build plan.

I am aware of the desire to avoid API calls and was cognizant of introducing performance issues. In my testing, I discovered that prior to these changes, when the runtime update gets to the "Resolving dependencies" stage, it's pretty slow (the buildplanner has to perform the solve, etc.). After these changes, this slowness is shifted prior to showing what dependencies are being added, but the subsequent runtime update stage is very fast (the buildplanner has already solved and is returning a cached result). Basically, we are shifting the solver slowness from at runtime update to showing additional dependencies.

This seems acceptable to me. The alternative would be to somehow modify the runtime to be able to receive information that a particular package was added, and if so, the runtime would to do this computation. That seems like a lot of unnecessary work in order to avoid a single API call for a cached buildplanner result.

internal/runbits/requirements/requirements.go
mitchell-as: Okay, I've used your wording, but again, I cannot compute direct dependencies vs. indirect dependencies as described in a link in a previous comment.

internal/runbits/requirements/dependencies.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+// maxListLength is the maximum number of direct dependencies to show before adding a "more..."
+// sentinel.
+const maxListLength = 10
+
+// showUpdatedPackages specifies whether or not to include updated dependencies in the direct
+// dependencies list, and whether or not to include updated dependencies when calculating indirect
+// dependency numbers.
+const showUpdatedPackages = false
```
mitchell-as, line 28: If this was `true`, a sample listing would look like this:

```
>..\..\cli\build\state.exe install requests
█ Installing Package

Operating on project ActiveState-CLI/small-python, located at C:\Users\me\tmp\small-python.

 • Searching for requests in the ActiveState Catalog ✔ Found
 • Creating commit ✔ Done
 • Resolving dependencies ✔ Done

Installing requests@2.31.0 includes 7 direct dependencies, and a total of 20 direct and indirect dependencies.
  ├─ certifi@2023.11.17 (3 dependencies)
  ├─ charset-normalizer@3.3.2 (3 dependencies)
  ├─ idna@3.6 (3 dependencies)
  ├─ openssl@1.11.0.20 → openssl@3.2.0 (Updated)
  ├─ sqlite3@3.41.2 → sqlite3@3.44.2 (1 dependencies) (Updated)
  ├─ urllib3@2.1.0 (3 dependencies)
  └─ zlib@1.2.13 → zlib@1.3 (Updated)
```

mitchell-as:
Okay, I've managed to compute direct and indirect dependencies. See the PR summary for an updated view of what this might look like. We also have the ability to list changed packages with a constant toggle depending on what route we want to take: https://github.com/ActiveState/cli/pull/3050/files#r1467078277

internal/runbits/requirements/dependencies.go
mitchell-as: Okay, I've enabled it.

internal/runbits/requirements/dependencies.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+
+		subdependencies := ""
+		if numSubs := numDependencies[dep.ArtifactID]; numSubs > 0 {
+			subdependencies = fmt.Sprintf(" ([ACTIONABLE]%s[/RESET] dependencies)", strconv.Itoa(numSubs)) // intentional leading space
+		}
+
+		item := fmt.Sprintf("[ACTIONABLE]%s@%s[/RESET]%s", dep.Name, version, subdependencies) // intentional omission of space before last %s
+		if oldVersion, exists := oldRequirements[fmt.Sprintf("%s/%s", dep.Namespace, dep.Name)]; exists && version != "" && oldVersion != version {
+			item = fmt.Sprintf("[ACTIONABLE]%s@%s[/RESET] → %s (%s)", dep.Name, oldVersion, item, locale.Tl("updated", "Updated"))
```
Naatan, line 161:
```suggestion
			item = fmt.Sprintf("[ACTIONABLE]%s@%s[/RESET] → %s (%s)", dep.Name, oldVersion, item, locale.Tl("updated", "updated"))
```

Consistent casing with "dependencies".

internal/runbits/requirements/dependencies.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+	"github.com/ActiveState/cli/pkg/platform/runtime/buildplan"
+	"github.com/ActiveState/cli/pkg/platform/runtime/store"
+	"github.com/ActiveState/cli/pkg/platform/runtime/target"
+)
+
+// maxListLength is the maximum number of direct dependencies to show before adding a "more..."
+// sentinel.
+const maxListLength = 10
```
Naatan, line 23: Playing with this PR and seeing the output, I'm going back and forth on whether to limit the list. The issue I see is that even with `state info` you would not be able to retroactively see what dependencies were brought in with a `state install` that you just ran. 

I'm not able to come up with a solution right here on the spot that doesn't have caveats. Let's discuss during stand how we might want to handle this.

Naatan (changes requested):
Looking really good, great work! I think we're almost there except for some minor tweaks and organization.

internal/runbits/requirements/dependencies.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return errs.Wrap(err, "Unable to fetch new build plan to compare with")
+	}
+
+	// Find the resolved version of the newly added or updated package.
+	packageVersion := locale.T("constraint_auto")
+	for _, source := range newBuildResult.Build.Sources {
+		if source.Name == packageName &&
+			(model.NamespaceMatch(source.Namespace, model.NamespaceBundlesMatch) ||
+				model.NamespaceMatch(source.Namespace, model.NamespacePackageMatch)) {
```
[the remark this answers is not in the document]
mitchell-as: Yeah we can, it just requires an additional function argument.

internal/runbits/requirements/dependencies.go
```diff
@@ -0,0 +1,217 @@
...
+
+// filterDependencies removes from the given dependency list any dependencies seen in the given
+// existing dependency map.
+// When `keepUpdated` is true, dependencies that changed version (i.e. they were updated updated)
+// are not filtered out.
+func filterDependencies(dependencies []*artifact.Artifact, existingArtifacts *map[string]string, keepUpdated bool) []*artifact.Artifact {
+	added := make([]*artifact.Artifact, 0)
+	for _, dep := range dependencies {
+		key := fmt.Sprintf("%s/%s", dep.Namespace, dep.Name)
+		oldVersion, exists := (*existingArtifacts)[key]
```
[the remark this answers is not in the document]
mitchell-as: I want to pass around pointers instead of `map[string]string` values. Or are `map[string]string` passed around by reference?

internal/runbits/requirements/dependencies.go
mitchell-as: Based on discussion during stand, I've removed the list truncation. All dependencies are shown.

mitchell-as:
I managed to find a way to infer that a runtime update was triggered by adding a single package. This avoids having to devise a method of communicating between the runbits/requirements package and the pkg/platform/runtime package. I've repurposed the unused runbits/changesummary package to display dependency information. I did have to plumb through an outputer interface to the runtime, though.

internal/output/plain.go
```diff
@@ -101,13 +101,13 @@ func (f *Plain) write(writer io.Writer, value interface{}) {
...
 func (f *Plain) writeNow(writer io.Writer, value string) {
-	_, err := colorize.Colorize(wordWrap(value), writer, !f.cfg.Colored)
+	_, err := colorize.Colorize(WordWrap(value), writer, !f.cfg.Colored)
 	if err != nil {
 		logging.ErrorNoStacktrace("Writing colored output failed: %v", err)
 	}
 }
 
-func wordWrap(text string) string {
+func WordWrap(text string) string {
```
mitchell-as, line 110: I need to expose this for a unit test that fails on CI due to a smaller terminal width than I have locally. (My local test passes.)

internal/runbits/changesummary/changesummary.go
```diff
@@ -22,49 +26,132 @@ func New(out output.Outputer) *ChangeSummary {
...
+
+	cs.out.Notice("") // blank line
+}
+
+// isDependency iterates over all artifacts and their dependencies in the given changeset, and
+// returns whether or not the given artifact is a dependency of any of those artifacts or
+// dependencies.
+func isDependency(a artifact.ArtifactID, changeset artifact.ArtifactChangeset, artifacts artifact.Map) bool {
+	for _, artifactId := range changeset.Added {
+		if artifactId == a {
```
Naatan, line 132: I think this would always match line 132? On line 37 you iterate through this same slice, and call this function with the artifact ID in that slice.

I might be missing something though, still having my morning coffee.

pkg/platform/runtime/buildplan/dependencies.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	return deps, nil
+}
+
+// DependencyTreeFor returns for the given artifact ID a map of dependencies to a list of unique
+// subdependencies. The returned map excludes any dependencies in the given filter, unless there is
+// a version mismatch and `keepUpdated` is true (this signifies a dependency's version was
+// updated).
+func DependencyTreeFor(a artifact.ArtifactID, artifacts artifact.Map, filter artifact.Map, keepUpdated bool) map[artifact.ArtifactID][]artifact.ArtifactID {
```
Naatan, line 119: Nit: Calling this a tree feels misleading since the second level is essentially flattening the rest of the tree.

Perhaps for future use this could actually construct the whole tree, but on each level give a count summary of what's underneath. 

Anyway, strong nit; happy to accept as is.

Naatan (changes requested):
I don't know why it's not working for me:

```
~/Projects/cli DX-2423*                                                                                                                 09:25:06 am
❯ state --version
ActiveState CLI by ActiveState Software Inc.
License BSD 3
Version 0.44.0-SHA28e6c0c
Revision 28e6c0ceaf7093a2388d95a419bf1c2f95140431
Channel mitchell/dx-1974
Built Mon Jan 29 2024 20:28:51 +0000 UTC


~/Projects/cli DX-2423*                                                                                                                 09:25:16 am
❯ state install argparse
█ Installing Package

Operating on project ActiveState/cli, located at /Users/nathanrijksen/Projects/cli.

 • Searching for argparse in the ActiveState Catalog ✔ Found
 • Creating commit ✔ Done
█ Installing Runtime

Installing your runtime and dependencies.
Setting Up Runtime
```

mitchell-as:
> I don't know why it's not working for me:

Argparse does not have many dependencies, and if those dependencies are already up to date, you won't see anything. Try something more complex like scipy or numpy.

internal/runbits/changesummary/changesummary.go
mitchell-as: When `state install <pkg>` brings in a few additional packages, each of those packages shows up in `changeset.Added`. I need to iterate through each of those packages, and then determine if a given package is named as a dependency of one of the added (or updated) packages.

It's a complex dance, but this is necessary.

pkg/platform/runtime/buildplan/dependencies.go
mitchell-as: I've renamed the function. If you still don't like it, please suggest a better one :)

internal/runbits/changesummary/changesummary.go
Naatan: But my point is that on line 37 you iterate through artifact ID's on the same slice you are iterating on artifact ID's here. So naturally this conditional on line 132 would match, and it would just "continue" and never actually get to the rest of the code in this function?

Naatan (changes requested):
Looking at the output, it's a bit weird that we print the build log, then the dependency tree, and then the build progress:

```
Setting Up Runtime
Resolving Dependencies ✔ Done
Build Log /Users/nathanrijksen/Library/Application
Support/activestate/cli-mitchell/dx-1974/logs/build-2a10adbe-7ba3-4755-bbaf-1ec39910e5c4-20240130131101.log

Installing numpy@1.26.3 includes 7 direct dependencies, and a total of 33 direct and indirect dependencies.
  ├─ cython@0.29.36 (33 dependencies)
  ├─ flit-core@3.9.0 (32 dependencies)
  ├─ libmkl@2020.4 (1 dependencies)
  ├─ packaging@23.2 (32 dependencies)
  ├─ python-module-build-support@1.00 (28 dependencies)
  ├─ setuptools@62.6.0 (25 dependencies)
  └─ wheel@0.38.4 (24 dependencies)

Building                       /             0/48 [--------------------------------------]   0 %
```

I think we should print the build log directly before the building step.

internal/runbits/changesummary/changesummary.go
mitchell-as: No, it would still check against every other dependency in `changeset.Added` and then later `changeset.Updated`.

I hope we're not talking past each other here.

Assume you've added dependencies [foo, bar, baz], and updated dependency [quux]. foo depends on bar, and bar depends on baz and quux.

Line 37 iterates over [foo, bar, baz], and for each one calls `isDependency()`. For foo, the question is: is foo a dependency of any packages in the changeset? Well the only way to find out is to iterate over every package in the changeset: [foo, bar, baz] first, then [quux]. Naturally 132 will match the first time, so we don't bother checking to see if foo is a dependency of foo. The next two iterations will show that neither bar nor baz has foo as a dependency. Then we move on to quux and determine that foo is not a dependency of quux either. Thus, foo has been independently added and is a candidate for `state install`.

Next comes bar during the line 37 iteration. `isDependency()` is called. On the first iteration we see that bar is a dependency of foo, so go ahead and return. If this wasn't the case, the second iteration would match line 132 and skip checking if bar is a dependency of bar and continue on to baz and then quux.

etc.

Sorry if I'm stating the obvious here, but I cannot see the problem you're raising.

Naatan: Thanks for sticking with me. You are indeed stating the obvious, sometimes my brain gets stuck in a loop itself, quite the irony. 🙏

Naatan (approved):
Great work @mitchell-as, I feel like this has a lot of value, and it paves the way for other similar features!

PR #3051, opened 2024-01-23
Update `state info` output

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2399" title="DX-2399" target="_blank">StoryDX-2399</a>  Clean up the `state info` UI
  </td></table>
  <br />

### DIFF
--- a/internal/colorize/colorize.go
+++ b/internal/colorize/colorize.go
@@ -17,7 +17,9 @@ var colorRx *regexp.Regexp
 func init() {
 	defer profile.Measure("colorize:init", time.Now())
 	var err error
-	colorRx, err = regexp.Compile(`\[(HEADING|NOTICE|SUCCESS|ERROR|WARNING|CAUTION|ALERT|DISABLED|ACTIONABLE|/RESET)!?\]`)
+	colorRx, err = regexp.Compile(
+		`\[(HEADING|NOTICE|SUCCESS|ERROR|WARNING|DISABLED|ACTIONABLE|CYAN|GREEN|RED|ORANGE|YELLOW|MAGENTA|/RESET)!?\]`,
+	)
 	if err != nil {
 		panic(fmt.Sprintf("Could not compile regex: %v", err))
 	}
@@ -31,8 +33,12 @@ type ColorTheme interface {
 	Warning(writer io.Writer)
 	Disabled(writer io.Writer)
 	Actionable(writer io.Writer)
-	Caution(writer io.Writer)
-	Alert(writer io.Writer)
+	Cyan(writer io.Writer)
+	Green(writer io.Writer)
+	Red(writer io.Writer)
+	Orange(writer io.Writer)
+	Yellow(writer io.Writer)
+	Magenta(writer io.Writer)
 	Reset(writer io.Writer)
 }
 
@@ -75,14 +81,34 @@ func (dct defaultColorTheme) Actionable(writer io.Writer) {
 	colorstyle.New(writer).SetStyle(colorstyle.Cyan, true)
 }
 
-// Caution switches to orange foreground
-func (dct defaultColorTheme) Caution(writer io.Writer) {
-	colorstyle.New(writer).SetStyle(colorstyle.Orange, true)
+// Blue switches to blue foreground
+func (dct defaultColorTheme) Cyan(writer io.Writer) {
+	colorstyle.New(writer).SetStyle(colorstyle.Cyan, false)
 }
 
-// Alert switches to magenta foreground
-func (dct defaultColorTheme) Alert(writer io.Writer) {
-	colorstyle.New(writer).SetStyle(colorstyle.Magenta, true)
+// Green switches to green foreground
+func (dct defaultColorTheme) Green(writer io.Writer) {
+	colorstyle.New(writer).SetStyle(colorstyle.Green, false)
+}
+
+// Red switches to red foreground
+func (dct defaultColorTheme) Red(writer io.Writer) {
+	colorstyle.New(writer).SetStyle(colorstyle.Red, false)
+}
+
+// Orange switches to orange foreground
+func (dct defaultColorTheme) Orange(writer io.Writer) {
+	colorstyle.New(writer).SetStyle(colorstyle.Orange, false)
+}
+
+// Yellow switches to yellow foreground
+func (dct defaultColorTheme) Yellow(writer io.Writer) {
+	colorstyle.New(writer).SetStyle(colorstyle.Yellow, false)
+}
+
+// Magenta switches to magenta foreground
+func (dct defaultColorTheme) Magenta(writer io.Writer) {
+	colorstyle.New(writer).SetStyle(colorstyle.Magenta, false)
 }
 
 // Reset re-sets all color settings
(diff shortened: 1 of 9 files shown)


MDrakos:
Had to make some changes to the `colorize` and `output` packages in order to get the output close to the mock.

internal/runners/packages/info.go
```diff
@@ -113,67 +136,145 @@ func specificIngredientVersion(ingredientID *strfmt.UUID, version string) (*inve
...
+				authorsOutput = append(authorsOutput, *author.Name)
+			}
+		}
+		res.Authors = fmt.Sprintf("[ACTIONABLE]%s[/RESET]", strings.Join(authorsOutput, ", "))
 	}
 
-	for _, author := range authors {
-		if author.Name != nil {
-			res.Authors = append(res.Authors, *author.Name)
+	if len(so.Vulnerabilities) > 0 {
```
Naatan, line 199: I can't comment the whole block cause it's so big it requires scrolling.. but this whole block feels like it is doing a lot of interpretation that we could probably do in the model. So that this code only deals with representing the info, not interpreting.

That'll come in handy also with the other stories that show this type of information.

Another idea might be to turn this into a runbit. Since again showing the quick cve summary is something we'll want to do in other commands.

internal/colorize/colorize.go
```diff
@@ -140,6 +152,10 @@ func colorize(ct ColorTheme, writer io.Writer, arg string) {
 		ct.Disabled(writer)
 	case `ACTIONABLE`:
 		ct.Actionable(writer)
+	case `CAUTION`:
+		ct.Caution(writer)
+	case `ALERT`:
+		ct.Alert(writer)
```
Naatan, line 158: It might be worth simply adding things like `[RED]`. So far we've avoided that since we weren't opinionated to that level, but with CVE information it seems appropriate to call out the color.

Naatan (changes requested):
This is looking great already, excited to see this live! Some additional comments here in summary since it doesn't apply to any one specific bit of code:


- Drop the "█ Displaying Package Information" title, it's redundant
- There are is a lot of bolded text which is not bolded in the mock. This is probably cause we're reusing things like [ACTIONABLE].. we should probably introduce more color tags to address these.
- I've realized I missed a few things in the mock and made changes to address, to summarize:
	- License field should also be colored
	- Added some help text under the "this package has vulnerabilities" block to make them actionable. 
	- I've added an "if not authenticated" mock, since we need auth for CVE info.

internal/runners/packages/info.go
MDrakos: Decided to go with processing this information in the model package. This code block is still fairly large but it's using the already organized information to build up the output.

pkg/platform/model/vulnerabilities.go
```diff
@@ -1,16 +1,35 @@
...
+	Medium   []string
+	Low      []string
+}
 
-func FetchVulnerabilitiesForIngredients(auth *authentication.Auth, ingredients []*VulnerabilityIngredient) ([]model.VulnerableIngredientsFilter, error) {
+func (v Vulnerabilites) Length() int {
+	return len(v.Critical) + len(v.High) + len(v.Medium) + len(v.Low)
+}
+
+func FetchVulnerabilitiesForIngredients(auth *authentication.Auth, ingredients []*request.Ingredient) ([]VulnerabilityIngredient, error) {
```
Naatan, line 32: Nit: Fetch to me says it's going to make an API call.

Naatan (changes requested):
Looking good! But I wanted to test the UI locally and getting a panic:

```
 state info tensorflow
An unexpected error occurred.
Error: runtime error: invalid memory address or nil pointer dereference
Stack trace: goroutine 51 [running]:
runtime/debug.Stack()
	/Users/runner/hostedtoolcache/go/1.20.12/x64/src/runtime/debug/stack.go:24 +0x65
github.com/ActiveState/cli/internal/sighandler.(*sigHandler).WaitForFunc.func1.1()
	/Users/runner/work/cli/cli/internal/sighandler/awaiting.go:48 +0x2d
panic({0x1d1fbc0, 0x29cab90})
	/Users/runner/hostedtoolcache/go/1.20.12/x64/src/runtime/panic.go:884 +0x213
github.com/ActiveState/cli/internal/runners/packages.newInfoResult({0xc0003225f0, 0xc0009a8c60, {0xc000154c00, 0x1, 0x4}, {0xc0007ca640, 0xf, 0x13}, {0xc00019aa80, 0xe, ...}})
	/Users/runner/work/cli/cli/internal/runners/packages/info.go:209 +0x6de
github.com/ActiveState/cli/internal/runners/packages.(*infoOutput).MarshalOutput(0xc0003fa070, {0x1d23620?, 0xc0003fa070?})
	/Users/runner/work/cli/cli/internal/runners/packages/info.go:286 +0x98
github.com/ActiveState/cli/internal/output.mediatorValue({0x1d23620, 0xc0003fa070}, {0x1ea11fb, 0x5})
	/Users/runner/work/cli/cli/internal/output/mediator.go:62 +0x15e
github.com/ActiveState/cli/internal/output.(*Mediator).Print(0xc00023dfc0, {0x1d23620?, 0xc0003fa070?})
	/Users/runner/work/cli/cli/internal/output/mediator.go:31 +0x3d
github.com/ActiveState/cli/internal/runners/packages.(*Info).Run(0xc00013ee80, {{{0x0, 0x0}, {0x207152c3b, 0xa}, {0x0, 0x0}}, {{0x0, 0x0}, 0x0}, ...}, ...)
	/Users/runner/work/cli/cli/internal/runners/packages/info.go:111 +0x8ee
github.com/ActiveState/cli/cmd/state/internal/cmdtree.newInfoCommand.func1(0x0?, {0xc00095ab40?, 0x0?, 0x0?})
	/Users/runner/work/cli/cli/cmd/state/internal/cmdtree/packages.go:205 +0x85
github.com/ActiveState/cli/internal/captain.(*Command).cobraExecHandler.func2()
	/Users/runner/work/cli/cli/internal/captain/command.go:703 +0x1b3
github.com/ActiveState/cli/internal/sighandler.(*sigHandler).WaitForFunc.func1()
	/Users/runner/work/cli/cli/internal/sighandler/awaiting.go:50 +0x6f
created by github.com/ActiveState/cli/internal/sighandler.(*sigHandler).WaitForFunc
	/Users/runner/work/cli/cli/internal/sighandler/awaiting.go:47 +0xca
```

pkg/platform/model/vulnerabilities.go
MDrakos: It is, just not in the changeset: https://github.com/ActiveState/cli/pull/3051/files#diff-1dec3c13927fe0d61f8457d8c7951d7a60d655d643f1aeff2a37a59bd0b41031R46

MDrakos:
Fixed the panic and tested with `tensorflow` myself. When there are _a lot_ of CVEs the output with the CVE IDs can be a bit awkward. I experimented with a custom table output but it also wasn't great. Not sure if we want to address that here.

Naatan (approved):
Looks good!

PR #3052, opened 2024-01-24
When listing requested versions, consider version constraints list.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2464" title="DX-2464" target="_blank">BugDX-2464</a>  install `django@4.1.x` where expected resolve should be `4.1.13` but under `state pkg` shown `Auto → 4.1.13`
  </td></table>
  <br />
 

The version constraint string will be empty if it's a complex constraint.

Sample:

```
% ../../cli/build/state packages         
█ Listing Packages

Operating on project ActiveState-CLI/small-python, located at 
/Users/me/tmp/small-python.

  Name      Version              
─────────────────────────────────
  django    >=4.1,<4.2 → 4.1.13
```

### DIFF
--- a/pkg/platform/model/checkpoints.go
+++ b/pkg/platform/model/checkpoints.go
@@ -1,7 +1,6 @@
 package model
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/ActiveState/cli/pkg/platform/api/mono/mono_models"
@@ -13,7 +12,6 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/pkg/platform/api/graphql"
 	gqlModel "github.com/ActiveState/cli/pkg/platform/api/graphql/model"
 	"github.com/ActiveState/cli/pkg/platform/api/graphql/request"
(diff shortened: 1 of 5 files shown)


mitchell-as:
The lone test failure is one of the usual timeouts.

pkg/platform/model/checkpoints.go
```diff
@@ -100,6 +102,35 @@ func GqlReqsToMonoCheckpoint(requirements []*gqlModel.Requirement) []*mono_model
 	return result
 }
 
+func GqlReqVersionConstraintsString(requirement *gqlModel.Requirement) string {
```
mitchell-as, line 105: This is similar to https://github.com/ActiveState/cli/blob/07489ff538ad996075fe0ac08fc256312cd3288c/pkg/platform/model/inventory.go#L478, but I could not reconcile these into a single function because their underlying types are different.

Naatan: I was concerned about adding this to the client due to the maintenance overhead this would create. In the end that was the compromise made, but having it in the client twice feels like it makes that problem worse.

Could you make something to standardize the inputs so the logic for this isn't duplicated?

eg. either refactor the function to take its own types, and have some model function that converts the two different requirement types to the desired one, or use an interface. Either way you'd need two separate ways to standardize the requirement types.

Naatan (changes requested):

mitchell-as:
Test failures are due to the temporary inability to stage commits, which has nothing to do with this PR.

Naatan (approved):

PR #3055, opened 2024-01-24
Replace uses of `os.Exit()` outside main with `panic()`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2468" title="DX-2468" target="_blank">BugDX-2468</a>  Running `state projects remote` when not auth, exit state prematurely causing printed command not echo in the prompt
  </td></table>
  <br />
 

main recovers from panics and reports them to Rollbar.

Also require authentication prior to running `state projects remote`, which triggered this PR.

Any calls to `os.Exit()` will ignore the deferred method in main that turns back on terminal echoing. (It is supposed to be turned off during State Tool execution.) Linux was susceptible to having terminal echo left off after a premature exit.

### DIFF
--- a/internal/runners/projects/remote.go
+++ b/internal/runners/projects/remote.go
@@ -18,6 +18,10 @@ import (
 func (r *Projects) RunRemote(params *Params) error {
 	projectfile.CleanProjectMapping(r.config)
 
+	if !r.auth.Authenticated() {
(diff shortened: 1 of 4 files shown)


mitchell-as:
This needs to be targeted at v43. Closing in favor of a new PR.

PR #3056, opened 2024-01-24
`state projects remote` checks for authentication

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2468" title="DX-2468" target="_blank">BugDX-2468</a>  Running `state projects remote` when not auth, exit state prematurely causing printed command not echo in the prompt
  </td></table>
  <br />
 

The lack of authentication results in `auth.Client()` calling `os.Exit()`, which ignores the deferred method in main that turns back on terminal echoing. (It is supposed to be turned off during State Tool execution.) Linux was susceptible to having terminal echo left off after a premature exit.

### DIFF
--- a/internal/runners/projects/remote.go
+++ b/internal/runners/projects/remote.go
@@ -18,6 +18,10 @@ import (
 func (r *Projects) RunRemote(params *Params) error {
 	projectfile.CleanProjectMapping(r.config)
 
+	if !r.auth.Authenticated() {
+		return locale.NewInputError("err_api_not_authenticated")
+	}
+
 	remoteProjects, err := r.newRemoteProjectsOutput(params.Local)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


Naatan:
@MDrakos Yeah I agree, we should not be using panics this way (or for the most part; at all).

I think we have a pattern for this type of problem already, you can return an error that embeds the exit code. It is interpreted here:

https://github.com/ActiveState/cli/blob/master/internal/errs/exitcode.go#L32-L32

mitchell-as:
Very well, I've reverted the panic change and just left the authentication check for `state projects remote`. I filed a new ticket for avoiding cases like this in the future: https://activestatef.atlassian.net/browse/DX-2489

MDrakos (approved):

PR #3060, opened 2024-01-26
CVE Information for `state install`

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2393" title="DX-2393" target="_blank">StoryDX-2393</a>  CVE information for state install
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -42,6 +42,11 @@ func init() {
 	configMediator.RegisterOption(constants.SecurityPromptLevelConfig, configMediator.String, configMediator.EmptyEvent, configMediator.EmptyEvent)
 }
 
+const (
+	promptDefault      = true
+	promptDefaultLevel = vulnModel.SeverityCritical
+)
+
 type PackageVersion struct {
 	captain.NameVersionValue
 }
@@ -420,13 +425,13 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
 
 func (r *RequirementOperation) shouldPromptForSecurity(vulnerabilities *model.VulnerabilityIngredient) bool {
 	if !r.Config.IsSet(constants.SecurityPromptConfig) {
-		if err := r.Config.Set(constants.SecurityPromptConfig, true); err != nil {
+		if err := r.Config.Set(constants.SecurityPromptConfig, promptDefault); err != nil {
 			multilog.Error("Failed to set security prompt config: %v", err)
 		}
 	}
 
 	if !r.Config.IsSet(constants.SecurityPromptLevelConfig) {
-		if err := r.Config.Set(constants.SecurityPromptLevelConfig, vulnModel.SeverityCritical); err != nil {
+		if err := r.Config.Set(constants.SecurityPromptLevelConfig, promptDefaultLevel); err != nil {
 			multilog.Error("Failed to set security prompt level config: %v", err)
 		}
 	}
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -263,7 +263,7 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 					severityBreakdown = append(severityBreakdown, fmt.Sprintf("[MAGENTA]%d low[/RESET]", len(vulnerabilities.Vulnerabilities.Low)))
 				}
 
-				out.Print("    " + locale.Tr("warning_vulnerable", strconv.Itoa(vulnerabilities.Vulnerabilities.Length()), strings.Join(severityBreakdown, ", ")))
+				out.Print("    " + strings.TrimSpace(locale.Tr("warning_vulnerable", strconv.Itoa(vulnerabilities.Vulnerabilities.Length()), strings.Join(severityBreakdown, ", "))))
 			}
 		} else {
 			pg.Stop(locale.T("progress_safe"))
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -26,6 +26,7 @@ import (
 	"github.com/ActiveState/cli/internal/runbits/rationalize"
 	bpModel "github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
 	medmodel "github.com/ActiveState/cli/pkg/platform/api/mediator/model"
+	"github.com/ActiveState/cli/pkg/platform/api/vulnerabilities/request"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildscript"
@@ -210,26 +211,40 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 	}
 
 	if r.Auth.Authenticated() && operation == bpModel.OperationAdded && ns.Type() == model.NamespacePackage {
-		pg = output.StartSpinner(out, locale.Tr("progress_pkg_nolang", requirementName), constants.TerminalAnimationInterval)
+		pg = output.StartSpinner(out, locale.Tr("progress_cve_search", requirementName), constants.TerminalAnimationInterval)
 
-		vulnerabilities, err := model.FetchVulnerabilitiesForIngredients(r.Auth, []*model.VulnerabilityIngredient{
-			{
-				Namespace: ns.String(),
-				Name:      requirementName,
-				Version:   requirementVersion,
-			},
+		vulnerabilities, err := model.FetchVulnerabilitiesForIngredient(r.Auth, &request.Ingredient{
+			Namespace: ns.String(),
+			Name:      requirementName,
+			Version:   requirementVersion,
 		})
 		if err != nil {
 			return errs.Wrap(err, "Failed to retrieve vulnerabilities")
 		}
 
-		if len(vulnerabilities) > 0 {
-			pg.Stop(locale.T("progress_vulnerable"))
-			// TODO: Iterate over vulnerabilities and tally up the counts for each severity
-			for _, vuln := range vulnerabilities {
+		var safe bool
+		if vulnerabilities == nil || (vulnerabilities != nil && vulnerabilities.Vulnerabilities.Length() == 0) {
+			safe = true
+		}
 
-			}
-			out.Print(locale.Tr("warning_vulnerable", strconv.Itoa(len(vulnerabilities))))
+		// There should only be one ingredient in the list
+		var severityBreakdown []string
+		if len(vulnerabilities.Vulnerabilities.Critical) > 0 {
+			severityBreakdown = append(severityBreakdown, fmt.Sprintf("[RED]%d Critical[/RESET]", len(vulnerabilities.Vulnerabilities.Critical)))
+		}
+		if len(vulnerabilities.Vulnerabilities.High) > 0 {
+			severityBreakdown = append(severityBreakdown, fmt.Sprintf("[ORANGE]%d high[/RESET]", len(vulnerabilities.Vulnerabilities.High)))
+		}
+		if len(vulnerabilities.Vulnerabilities.Medium) > 0 {
+			severityBreakdown = append(severityBreakdown, fmt.Sprintf("[YELLOW]%d medium[/RESET]", len(vulnerabilities.Vulnerabilities.Medium)))
+		}
+		if len(vulnerabilities.Vulnerabilities.Low) > 0 {
+			severityBreakdown = append(severityBreakdown, fmt.Sprintf("[MAGENTA]%d low[/RESET]", len(vulnerabilities.Vulnerabilities.Low)))
+		}
+
+		if !safe {
+			pg.Stop(locale.T("progress_unsafe"))
+			out.Print("    " + locale.Tr("warning_vulnerable", strconv.Itoa(vulnerabilities.Vulnerabilities.Length()), strings.Join(severityBreakdown, ", ")))
 		} else {
 			pg.Stop(locale.T("progress_safe"))
 		}
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -209,6 +209,34 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 		pg = nil
 	}
 
+	if r.Auth.Authenticated() && operation == bpModel.OperationAdded && ns.Type() == model.NamespacePackage {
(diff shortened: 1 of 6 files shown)


internal/runbits/requirements/requirements.go
```diff
@@ -209,6 +217,61 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+					severityBreakdown = append(severityBreakdown, fmt.Sprintf("[ORANGE]%d high[/RESET]", len(vulnerabilities.Vulnerabilities.High)))
+				}
+				if len(vulnerabilities.Vulnerabilities.Medium) > 0 {
+					severityBreakdown = append(severityBreakdown, fmt.Sprintf("[YELLOW]%d medium[/RESET]", len(vulnerabilities.Vulnerabilities.Medium)))
+				}
+				if len(vulnerabilities.Vulnerabilities.Low) > 0 {
+					severityBreakdown = append(severityBreakdown, fmt.Sprintf("[MAGENTA]%d low[/RESET]", len(vulnerabilities.Vulnerabilities.Low)))
+				}
+
+				out.Print("    " + strings.TrimSpace(locale.Tr("warning_vulnerable", strconv.Itoa(vulnerabilities.Vulnerabilities.Length()), strings.Join(severityBreakdown, ", "))))
```
MDrakos, line 266: I am still unsure why there is a new line at the end of this string. I spent some time trying to figure it out. For now, I'm trimming it off.

internal/runbits/requirements/requirements.go
```diff
@@ -209,6 +222,61 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 		pg = nil
 	}
 
+	if r.Auth.Authenticated() && operation == bpModel.OperationAdded && ns.Type() == model.NamespacePackage {
```
mitchell-as, line 225: What about bundles that pull in individual packages? Are they checked too?

internal/runbits/requirements/requirements.go
```diff
@@ -209,6 +222,61 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+		pg = output.StartSpinner(out, locale.Tr("progress_cve_search", requirementName), constants.TerminalAnimationInterval)
+
+		vulnerabilities, err := model.FetchVulnerabilitiesForIngredient(r.Auth, &request.Ingredient{
+			Namespace: ns.String(),
+			Name:      requirementName,
+			Version:   requirementVersion,
+		})
+		if err != nil {
+			return errs.Wrap(err, "Failed to retrieve vulnerabilities")
```
mitchell-as, line 234: `pg` is not stopped on error here.

internal/runbits/requirements/requirements.go
```diff
@@ -209,6 +222,61 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+		}
+
+		var safe bool
+		if vulnerabilities == nil || (vulnerabilities != nil && vulnerabilities.Vulnerabilities.Length() == 0) {
```
mitchell-as, line 238: You can simplify this to just `vulnerabilities == nil || vulnerabilities.Vulnerabilities.Length() == 0`.

internal/runbits/requirements/requirements.go
mitchell-as: It's because multi-line messages in the locale file have a newline at the end by default. I've been bitten by this before too.

internal/runbits/requirements/requirements.go
```diff
@@ -209,6 +222,61 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+					severityBreakdown = append(severityBreakdown, fmt.Sprintf("[MAGENTA]%d low[/RESET]", len(vulnerabilities.Vulnerabilities.Low)))
+				}
+
+				out.Print("    " + strings.TrimSpace(locale.Tr("warning_vulnerable", strconv.Itoa(vulnerabilities.Vulnerabilities.Length()), strings.Join(severityBreakdown, ", "))))
+			}
+		} else {
+			pg.Stop(locale.T("progress_safe"))
+		}
+
+		pg = nil
```
mitchell-as, line 277: Did you consider splitting this into its own function instead of continuing to add to this already-ginormous function? :)

internal/runbits/requirements/requirements.go
```diff
@@ -355,6 +423,85 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
 	return nil
 }
 
+func (r *RequirementOperation) shouldPromptForSecurity(vulnerabilities *model.VulnerabilityIngredient) bool {
```
mitchell-as, line 426: I like that this is its own function!

internal/runbits/requirements/requirements.go
```diff
@@ -355,6 +423,85 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
...
+			multilog.Error("Failed to set security prompt level config: %v", err)
+		}
+	}
+
+	prompt := r.Config.GetBool(constants.SecurityPromptConfig)
+	promptLevel := r.Config.GetString(constants.SecurityPromptLevelConfig)
+
+	switch promptLevel {
+	case vulnModel.SeverityCritical:
+		return prompt && len(vulnerabilities.Vulnerabilities.Critical) > 0
```
mitchell-as, line 444: There is no guarantee that `vulnerabilities` is not nil. Line 238 (https://github.com/ActiveState/cli/pull/3060/files#diff-e6c3f1c43aa6786ae16024b493aff78e55e41cf15d5cec53d76242974a4c30d0R238) leaves `safe` as `false` if `vulnerabilities` is `nil`, and that will call this function with a `nil` argument.

internal/runbits/requirements/requirements.go
```diff
@@ -355,6 +423,85 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
...
+				len(vulnerabilities.Vulnerabilities.Medium) > 0)
+	case vulnModel.SeverityLow:
+		return prompt &&
+			(len(vulnerabilities.Vulnerabilities.Critical) > 0 ||
+				len(vulnerabilities.Vulnerabilities.High) > 0 ||
+				len(vulnerabilities.Vulnerabilities.Medium) > 0 ||
+				len(vulnerabilities.Vulnerabilities.Low) > 0)
+	}
+
+	return r.Config.GetBool(constants.SecurityPromptConfig)
```
mitchell-as, line 462: I feel like if this is `false`, then we shouldn't be doing any of the work above? Maybe check if this is `false` at the top and return `false` immediately if so. Then you can remove `prompt &&` from each switch case. Then return `false` down here I think.

internal/runbits/requirements/requirements.go
mitchell-as, line 426: Did you consider using the argument type of `model.VulnerabilityIngredient.Vulnerabilities` here? That way you don't have to keep typing `vulnerabilities.Vulnerabilities` everywhere. (I don't think any other field is being used.)

internal/runbits/requirements/requirements.go
```diff
@@ -355,6 +423,85 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
...
+}
+
+func (r *RequirementOperation) promptForSecurity(out output.Outputer, vulnerabilities *model.VulnerabilityIngredient) (bool, error) {
```
mitchell-as, line 465: Did you consider using the argument type of `model.VulnerabilityIngredient.Vulnerabilities` here? That way you don't have to keep typing `vulnerabilities.Vulnerabilities` everywhere. (I don't think any other field is being used.)

internal/runbits/requirements/requirements.go
```diff
@@ -209,6 +222,61 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+		if vulnerabilities == nil || (vulnerabilities != nil && vulnerabilities.Vulnerabilities.Length() == 0) {
+			safe = true
+		}
+
+		if !safe {
+			pg.Stop(locale.T("progress_unsafe"))
+			pg = nil
+
+			if r.shouldPromptForSecurity(vulnerabilities) {
+				out.Print("")
```
mitchell-as, line 247: Please don't use `out.Print()` with bare strings, because that causes trouble with structured output. You either want to use `out.Notice()`, or `out.Print(output.Prepare())` with a structure for structured output. I believe this is just extra information for the user, so `output.Notice()` should be sufficient.

internal/runbits/requirements/requirements.go
```diff
@@ -355,6 +423,85 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
...
+				len(vulnerabilities.Vulnerabilities.High) > 0 ||
+				len(vulnerabilities.Vulnerabilities.Medium) > 0 ||
+				len(vulnerabilities.Vulnerabilities.Low) > 0)
+	}
+
+	return r.Config.GetBool(constants.SecurityPromptConfig)
+}
+
+func (r *RequirementOperation) promptForSecurity(out output.Outputer, vulnerabilities *model.VulnerabilityIngredient) (bool, error) {
+	out.Print(locale.Tr("warning_vulnerable_simple", strconv.Itoa(vulnerabilities.Vulnerabilities.Length())))
```
mitchell-as, line 466: Again, please don't use `out.Print()` with strings. I think you can safely replace all of your calls with `out.Notice()` for this ticket.

internal/runbits/requirements/requirements.go
```diff
@@ -355,6 +423,85 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
...
+		lowOutput := fmt.Sprintf("[MAGENTA]%d Low: [/RESET]", len(vulnerabilities.Vulnerabilities.Low))
+		lowOutput += fmt.Sprintf("[CYAN]%s[/RESET]", strings.Join(vulnerabilities.Vulnerabilities.Low, ", "))
+		pkgVersionVulns = append(pkgVersionVulns, lowOutput)
+	}
+
+	out.Print(pkgVersionVulns)
+	out.Print("")
+	out.Print(locale.T("more_info_vulnerabilities"))
+
+	confirm, err := r.Prompt.Confirm("", locale.Tr("prompt_continue_pkg_operation"), ptr.To(false))
```
mitchell-as, line 497: How does this work with non-interactive mode? At first glance, it looks like it will always fail (i.e. `state install vulnerable-package -n` will not work. I could be wrong though. I'm in a few layers too deep, heh.

internal/testhelpers/e2e/session.go
```diff
@@ -182,6 +182,13 @@ func new(t *testing.T, retainDirs, updatePath bool, extraEnv ...string) *Session
 	err = fileutils.Touch(filepath.Join(dirs.Base, installation.InstallDirMarker))
 	require.NoError(session.T, err)
 
+	cfg, err := config.New()
+	require.NoError(session.T, err)
+
+	if err := cfg.Set(constants.SecurityPromptConfig, false); err != nil {
```
mitchell-as, line 188: I just want to double-check that this is an okay place/way to configure something like this. I'm used to seeing environment variables used to define behavior like this, like in `sandboxedTestEnvironment()`.

mitchell-as (changes requested):
This is a good start! I've left some feedback about some things to address and some things to think about.

internal/runbits/requirements/requirements.go
MDrakos: Thanks for pointing this out!

internal/runbits/requirements/requirements.go
MDrakos: Yeah, this and your other suggestion are great. Reduces the stuttering.

internal/runbits/requirements/requirements.go
MDrakos: I usually forget about structure output so this comment was helpful. 

PS. Don't look at the `info` runner :P

This may be something we want to audit and fix in our runners.

internal/runbits/requirements/requirements.go
MDrakos: You're right it will not install a package if you have prompting enabled and there are CVEs that are "higher" or equal to your configured preference.

internal/testhelpers/e2e/session.go
MDrakos: We could introduce an environment variable to configure this for testing purposes though that feels redundant with the config. I don't believe there should be any issues setting it this way as each config is contained to a single test, maybe a slight performance hit.

internal/runbits/requirements/requirements.go
MDrakos: I don't believe bundles were considered here... I'm not sure how much we are pushing their use. Looking at the dashboard you have to go to the project configuration page to see the CVEs for each package in the bundle rather than a report for the bundle itself. 

I'm also pretty sure that bundles are just for Perl so I'm not sure how much we want to complicate this by dealing with bundles.

internal/runbits/requirements/requirements.go
MDrakos: There is a defer at the top of this function that should handle that.

internal/runbits/requirements/requirements.go
MDrakos: I did not, only because this is following the pattern that already exists in the function. I know we've talked about refactoring this function before and there should be a story for it somewhere. I think that's a more appropriate place to start breaking this flow up.

internal/runbits/requirements/requirements.go
MDrakos: Good catch. We can skip some potentially unnecessary processing.

internal/runbits/requirements/requirements.go
MDrakos: In testing this, `Notice` does not produce the desired output when passing it a slice. Should I use the `output.Prepare` function here? I'm unfamiliar with how that works.

MDrakos: Whoops, sorry I was looking at the wrong line. When I pass a slice on line 494 I'm using `Print` but `Notice` is fine here.

internal/runbits/requirements/requirements.go
mitchell-as: So then a follow up question: is this the intended behavior? Should `-n` be a bypass, or should we have a `--force` flag?

mitchell-as (dismissed):
This looks fine now, but I've left a question on how we want to handle `-n` vs. adding a `--force` option with the new prompting mechanic. Depending on the answer, an additional change may be necessary.

internal/runbits/requirements/requirements.go
MDrakos: We discussed this during stand and decided to add an error message with a tip in non-interactive mode. Should cover this use-case.

mitchell-as (approved):

PR #3062, opened 2024-01-29
Implement 'state builds'

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2423" title="DX-2423" target="_blank">StoryDX-2423</a>  As a user I can inspect which builds were produced for my runtime
  </td></table>
  <br />


internal/runners/builds/builds.go
```diff
@@ -0,0 +1,202 @@
...
+func (b *Builds) Run(params *Params) (rerr error) {
+	defer rationalizeError(&rerr)
+
+	if b.project == nil {
+		return rationalize.ErrNoProject
+	}
+
+	// We don't use the runtime returned here, because builds needs more advanced runtime info, but we still want to call
+	// it in case our runtime isn't up-to date.
+	_, err := runtime.NewFromProject(b.project, target.TriggerBuilds, b.analytics, b.svcModel, b.out, b.auth, b.config)
```
MDrakos, line 101: Couldn't we use the `refresh` runbit for this?

internal/runners/builds/builds.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return locale.WrapInputError(err, "err_refresh_runtime_new", "Could not update runtime for this project.")
+	}
+
+	runtimeStore := store.New(target.NewProjectTarget(b.project, nil, target.TriggerBuilds).Dir())
+	bp, err := runtimeStore.BuildPlan()
```
MDrakos, line 107: The `store` is part of the initialized runtime. The runtime type has methods for resolved artifacts, we could add a method for to give us the build plan and use the runtime that we aren't assigning above.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -275,20 +302,49 @@ func buildRuntimeClosureMap(baseID strfmt.UUID, lookup map[strfmt.UUID]interface
...
+			Namespace:        info.Namespace,
+			Version:          &info.Version,
+			RequestedByOrder: true,
+			GeneratedBy:      currentArtifact.GeneratedBy,
+			Dependencies:     uniqueDeps,
+			URL:              currentArtifact.URL,
+			MimeType:         currentArtifact.MimeType,
+		}
+	} else {
+		// For whatever reason we have artifacts whose displayName are prefixed with their node ID, so strip it
```
MDrakos, line 323: The `displayName` property was added later, hence why it's not used here. I believe they went with a slightly odd naming convention. You might want to verify that the assumptions you're making below are in line with what is outlined here: https://activestatef.atlassian.net/browse/PB-3681

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -312,12 +368,14 @@ type SourceInfo struct {
 //
 //	Artifact (GeneratedBy) -> Step (Input) -> Source
 func getSourceInfo(sourceID strfmt.UUID, lookup map[strfmt.UUID]interface{}) (SourceInfo, error) {
-	source, ok := lookup[sourceID].(*model.Source)
+	node := lookup[sourceID]
```
MDrakos, line 371: If you add another `ok` check here we could skip the following logic if the `sourceID` is not a key in the `lookup` map.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -102,12 +105,31 @@ func (al *ArtifactListing) ArtifactIDs(buildtimeClosure bool) ([]artifact.Artifa
...
+
+type TerminalArtifactMap map[string]artifact.Map
+
 // Setting calculateBuildtimeClosure as true calculates the artifact map with the buildtime
 // dependencies. This is different from the runtime dependency calculation as it
 // includes ALL of the input artifacts of the step that generated each artifact.
-func newMapFromBuildPlan(build *model.Build, calculateBuildtimeClosure bool, cfg platformModel.Configurable) (artifact.Map, error) {
-	res := make(artifact.Map)
-
+func NewMapFromBuildPlan(build *model.Build, calculateBuildtimeClosure bool, filterStateToolArtifacts bool, filterTerminal *model.NamedTarget) (TerminalArtifactMap, error) {
```
MDrakos, line 132: nit: We should update this docstring and it should start with the function name.

I know you didn't write the docstring. I did so that's on me, but we should still improve it here if we can.

pkg/platform/runtime/buildplan/buildplan.go
```diff
@@ -102,12 +105,31 @@ func (al *ArtifactListing) ArtifactIDs(buildtimeClosure bool) ([]artifact.Artifa
 // artifact map by traversing the build plan from the terminal targets through
 // all of the runtime dependencies for each of the artifacts in the DAG.
 
+func newFilteredMapFromBuildPlan(build *model.Build, calculateBuildtimeClosure bool, cfg platformModel.Configurable) (artifact.Map, error) {
```
MDrakos, line 108: nit: the function name at the start of this docstring does not match the new function name.

MDrakos (changes requested):
This looks really good! A few comments to follow up on and consider.

internal/runners/builds/builds.go
```diff
@@ -0,0 +1,202 @@
...
+		*err = errs.WrapUserFacing(*err,
+			locale.Tr("err_no_project"),
+			errs.SetInput())
+	}
+}
+
+func (b *Builds) Run(params *Params) (rerr error) {
+	defer rationalizeError(&rerr)
+
+	if b.project == nil {
```
MDrakos, line 95: While working on the `state builds dl` command I've come to realize that requiring the project to be present on the system is a bit of an odd UX choice. We can source all of this information remotely. 

If a user wanted to download the windows installer they would first have to download and install the entire project to get the ID they are interested in and then download the actual artifact they are interested in.

Naatan: That's a good point! Wanna file a followup? I'd rather land this as an improvement than add to the scope at this stage.

internal/runners/builds/builds.go
Naatan: I looked at that, but unfortunately the refresh runbit gives UI even when there is nothing to refresh. I don't want this command to show this UI when it's not needed.

pkg/platform/runtime/buildplan/buildplan.go
Naatan: Thanks, I've commented on that ticket. It doesn't seem that it was implemented the way that ticket describes.

internal/runners/builds/builds.go
MDrakos: Sure, filed a follow-up here: https://activestatef.atlassian.net/browse/DX-2513.

I will continue on with `state builds dl` using a locally checked out project and it can be iterated on in the above story.

MDrakos (approved):
Nice work!

PR #3064, opened 2024-01-30
Document required fields when editing the YAML metafile prior to publishing.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2496" title="DX-2496" target="_blank">BugDX-2496</a>  Mandatory fields have no examples in the YAML
  </td></table>
  <br />
 

It will look something like this:

```
# Edit the following file and confirm in your terminal when done
# The 'namespace' field should be in "org/project" format.
# The 'version' field should follow semantic versioning and match the version in the filename (if any).
name: publish-1.0.1.zip
namespace: org/ActiveState
version: 1.0.1
description: Not Provided

## Optional -- Example Author:
# authors:
#     - name: John Doe
#       email: johndoe@domain.tld
#       websites:
#         - https://example.com
# 
## Optional -- Example Dependencies:
# dependencies:
#     - dependency:
#         name: example-linux-specific-ingredient
#         namespace: shared
#         versionRequirements: '>= 1.0.0'
#       conditions:
#         - name: linux
#           namespace: kernel
#           versionRequirements: '>= 0'
# 
```

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1492,6 +1492,11 @@ err_uploadingredient_edit_description_not_supported:
 uploadingredient_editor_opening:
   other: |
     Opening editor to edit ingredient meta information. Alternatively you may manually edit the following file: [ACTIONABLE]{{.V0}}[/RESET].
+uploadingredient_editor_header_comment:
+  other: |
+    # Edit the following file and confirm in your terminal when done
+    # The 'namespace' field should be in "org/project" format.
+    # The 'version' field should follow semantic versioning and match the version in the filename (if any).
 uploadingredient_success:
   other: |
     Successfully published as:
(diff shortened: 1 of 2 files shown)


mitchell-as:
@rawktron Would you please look at the sample YAML file in this PR description and see if it's what you were expecting? If so, you can approve the PR. Otherwise, let me know what else you'd like to see.

rawktron (changes requested):
So, there's a couple things I think we should do and some related to other tickets so might be overlap.

1. The comments for the fields should be in-line with them, so like

# The namespace field should be in the format org/folder. Org can simply be your username or any organization you're a member of.
namespace: org/project

etc.

2. We should include the feature field as an optional field

mitchell-as:
The ticket has been marked as "Done", but I don't remember why. Closing this PR.

PR #3066, opened 2024-01-30
Implement new migration flow for local orders

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2506" title="DX-2506" target="_blank">StoryDX-2506</a>  Local order has a backward compatible upgrade mechanic
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/legacy/projectmigration/projectmigration.go
+++ b/internal/runbits/legacy/projectmigration/projectmigration.go
@@ -119,8 +119,6 @@ func CreateMigrateScript(proj projecter) error {
 			Value: string(scriptValue),
 		},
 		projectfile.ScriptFields{
-			// No localized since this is a legacy migration mechanic and we'll remove this long before we'll introduce
-			// another language.
 			Description: locale.T("projectmigration_script_description"),
 			Standalone:  true,
 			Language:    scriptLanguage,
--- a/internal/runbits/legacy/projectmigration/projectmigration.go
+++ b/internal/runbits/legacy/projectmigration/projectmigration.go
@@ -1,10 +1,14 @@
 package projectmigration
 
 import (
+	"bytes"
+	_ "embed"
 	"errors"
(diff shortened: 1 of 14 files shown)


pkg/projectfile/projectfile.go
```diff
@@ -564,6 +564,39 @@ func (p *Project) LegacyCommitID() string {
...
+	}
+
+	p.parsedURL.LegacyCommitID = ""
+	p.Project = pf.String()
+	return nil
+}
+
+// SetLegacyCommit sets the commit id within the current project file. This is done
+// in-place so that line order is preserved.
+// If headless is true, the project is defined by a commit-id only
```
mitchell-as, line 584: Nit: we don't have headless commits anymore, right? So this comment is superfluous.

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -78,16 +78,110 @@ func PromptAndMigrate(proj projecter) (bool, error) {
...
+	}
+
+	script := projectfile.Script{
+		projectfile.NameVal{
+			Name:  "migrate-to-buildscripts",
+			Value: string(scriptValue),
+		},
+		projectfile.ScriptFields{
+			// No localized since this is a legacy migration mechanic and we'll remove this long before we'll introduce
+			// another language.
```
mitchell-as, line 123: I'm not following this comment. It looks like there is localization?

mitchell-as (approved):
Looks good to me. Minor nits to think about and possibly address.

internal/runbits/legacy/projectmigration/projectmigration.go
Naatan: Forgot to drop it after I moved it to locale cause it was getting noisy.

mitchell-as (approved):

PR #3067, opened 2024-01-30
Attempt to extract a usable name from the filename given to `state publish`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2500" title="DX-2500" target="_blank">BugDX-2500</a>  default name for package includes tar.gz
  </td></table>
  <br />
 

Samples:

```
>build\state.exe publish c:\Users\me\tmp\publish-1.0.2.zip
█ Publish Ingredient (Beta)

Publish following ingredient?
name: publish
namespace: org/ActiveState
version: 1.0.2
description: Not Provided

 (Y/n)
```

```
>build\state.exe publish c:\Users\me\tmp\publish-json2-1.0.2.zip
█ Publish Ingredient (Beta)

Publish following ingredient?
name: publish-json2
namespace: org/ActiveState
version: 1.0.2
description: Not Provided

 (Y/n)
```

```
>build\state.exe publish c:\Users\me\tmp\publish.zip
█ Publish Ingredient (Beta)

Publish following ingredient?
name: publish
namespace: org/ActiveState
version: 0.0.1
description: Not Provided

 (Y/n)
```

### DIFF
--- a/internal/runners/publish/publish.go
+++ b/internal/runners/publish/publish.go
@@ -78,6 +78,8 @@ type ParentIngredient struct {
 	Dependencies        []inventory_models.Dependency `json:"dependencies"`
 }
 
+var nameRegexp = regexp.MustCompile(`\w+([_-]\w+)*`)
+
 func (r *Runner) Run(params *Params) error {
 	if !r.auth.Authenticated() {
 		return locale.NewInputError("err_auth_required")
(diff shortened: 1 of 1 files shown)


internal/runners/publish/publish.go
```diff
@@ -78,6 +78,8 @@ type ParentIngredient struct {
 	Dependencies        []inventory_models.Dependency `json:"dependencies"`
 }
 
+var nameRegexp = regexp.MustCompile(`\w+([_-]\w+)*`)
```
Naatan, line 81: This feels too aggressive, it means that if I upload `hello-world.zip` it will default to the name `hello`.

Naatan (changes requested):

internal/runners/publish/publish.go
mitchell-as: Actually, it works as you'd expect:

```
>build\state.exe publish c:\Users\me\tmp\hello-world.zip
█ Publish Ingredient (Beta)

Publish following ingredient?
name: hello-world
namespace: org/ActiveState
version: 0.0.1
description: Not Provided

 (Y/n)
```

Naatan: Oh I see, you're using the full match, so anything not covered by this regex is considered superfluous. I thought the group was what you were stripping.

Naatan (approved):

PR #3072, opened 2024-01-31
Implement `state builds dl`

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2397" title="DX-2397" target="_blank">StoryDX-2397</a>  As a user I can download individual artifacts
  </td></table>
  <br />


internal/runners/builds/download.go
```diff
@@ -0,0 +1,175 @@
...
+
+	d.out.Notice(locale.Tl("msg_download_success", "Downloaded {{.V0}} to {{.V1}}", artifact.ArtifactID.String(), params.OutputDir))
+	return nil
+}
+
+func (d *Download) downloadArtifact(pg events.Handler, artifact *artifact.Artifact, targetDir string) (rerr error) {
+	defer func() {
+		evs := []events.Eventer{
+			events.ArtifactDownloadSuccess{artifact.ArtifactID},
+			events.Success{},
```
MDrakos, line 118: For some reason after a successful download the artifact bar is completed but the overall download bar is not. I thought this event would handle it but it doesn't. Any insights here would be welcome or I can continue to debug what's going on.

cmd/state/internal/cmdtree/builds.go
```diff
@@ -34,3 +34,38 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
...
+				Name:        "ID",
+				Description: locale.Tl("builds_download_arg_id", "The ID of the artifact to download"),
+				Value:       &params.BuildID,
+				Required:    true,
+			},
+			{
+				Name:        "target",
+				Description: locale.Tl("builds_download_arg_target", "The target directory to download the artifact to"),
+				Value:       &params.OutputDir,
+				Required:    true,
```
Naatan, line 59: This should be optional, if not provided it can download to the current working directory.

cmd/state/internal/cmdtree/builds.go
```diff
@@ -34,3 +34,38 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
...
+			},
+		},
+		func(_ *captain.Command, _ []string) error {
+			return runner.Run(params)
+		},
+	)
+	cmd.SetGroup(ProjectUsageGroup)
+	cmd.SetSupportsStructuredOutput()
+	cmd.SetUnstable(true)
+	cmd.DeprioritizeInHelpListing()
```
Naatan, line 69:
```suggestion
	cmd.SetSupportsStructuredOutput()
	cmd.SetUnstable(true)
```

This is a subcommand of `state builds`, so this isn't necessary / could cause issues.

internal/runners/builds/download.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (d *Download) Run(params *DownloadParams) (rerr error) {
+	defer rationalizeError(&rerr)
+
+	if d.project == nil {
+		return rationalize.ErrNoProject
+	}
+
+	pg := runbits.NewRuntimeProgressIndicator(d.out)
+	defer rtutils.Closer(pg.Close, &rerr)
```
Naatan, line 66: This is for runtime progress, not for downloads of individual files. You should be able to use the progress library directly, we use "github.com/vbauerster/mpb/v7".

I would not expect to see any use of the runtime progress handler in this PR. That just so happens to be the only thing that uses "GetWithProgress", but that doesn't mean "GetWithProgress" is married to it.

Naatan (changes requested):
I did not finish the review since I think the current review will already bring about substantial changes.

internal/runners/builds/download.go
```diff
@@ -0,0 +1,176 @@
...
+		},
+		ReportIncrementCb: func(inc int) error {
+			downloadBar.IncrBy(inc)
+			return nil
+		},
+	})
+	if err != nil {
+		return errs.Wrap(err, "Download %s failed", artifactURL.String())
+	}
+
```
Naatan, line 167: You'll need to wait for the progressbar to be done, the reporting is done in a subroutine, and right now your function ends before it has completed:

```
state builds dl AA0F839B
█ Download build artifacts (Beta)

bzip2                          |    8.4KB/117.6KB [==>-----------------------------------]   7 %
Downloaded bzip2 to /Users/nathanrijksen/Projects/cli/artifact.tar.gz
```

Naatan (changes requested):

internal/runners/builds/builds.go
```diff
@@ -76,17 +75,6 @@ func New(p primeable) *Builds {
...
-func rationalizeError(err *error) {
-	switch {
-	case err == nil:
-		return
-	case errors.Is(*err, rationalize.ErrNoProject):
-		*err = errs.WrapUserFacing(*err,
-			locale.Tr("err_no_project"),
-			errs.SetInput())
-	}
-}
```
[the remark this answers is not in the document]
MDrakos: I like this approach and this is good context to have. I think it's better that we police this kind of stuff more strictly than we have in the past so we don't end up regressing on these new initiatives. 

I've moved this error to `rationalizeCommonError` but that's leaving the more specific rationalize functions fairly bare. This still enforced the pattern though.

internal/runners/builds/download.go
```diff
@@ -0,0 +1,176 @@
...
+	if err := d.downloadArtifact(artifact, targetDir); err != nil {
+		return errs.Wrap(err, "Could not download artifact %s", artifact.ArtifactID.String())
+	}
+
+	return nil
+}
+
+func (d *Download) downloadArtifact(artifact *artifact.Artifact, targetDir string) (rerr error) {
```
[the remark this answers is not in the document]
MDrakos: I implemented this only to find that the spinner does not produce any output on non-interactive: https://github.com/ActiveState/cli/blob/master/internal/output/progress.go#L36. Confirmed with Mitchell on slack that we only do progress on plain output as we want to ensure any output is valid JSON.

internal/runners/builds/download.go
MDrakos: Added it and it's gross

internal/runners/builds/download.go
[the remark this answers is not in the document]
Naatan: Non-interactive is not the same as structured. Plain output can be either interactive or non-interactive. For example CI would be non-interactive, but would still need some form of progress indication otherwise most CI systems would consider the task "hung" after a certain amount of time.

test/integration/builds_int_test.go
```diff
@@ -115,11 +115,11 @@ func (suite *BuildsIntegrationTestSuite) TestBuilds_Download() {
 	cp.ExpectExitCode(0, e2e.RuntimeSourcingTimeoutOpt)
 
 	cp = ts.SpawnWithOpts(
-		e2e.OptArgs("builds", "dl", "16291D17", "."),
+		e2e.OptArgs("builds", "dl", "1f924587", "."),
 	)
-	cp.Expect("Downloaded", e2e.RuntimeSourcingTimeoutOpt)
+	cp.Expect("Downloaded sqlite3", e2e.RuntimeSourcingTimeoutOpt)
```
Naatan, line 120: sqlite3 is not tiny :p 

image

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3072#discussion_r1480622918

test/integration/builds_int_test.go
MDrakos: It's 2.3 MB! Found a smaller one. In testing this out I also found a different approach rather than aborting the download bar.

internal/runners/builds/download.go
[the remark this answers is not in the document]
MDrakos: That makes sense. Do we want to update the spinner in this PR? I feel like it needs its own story as it could have further reaching implications.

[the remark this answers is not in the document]
Naatan: No I wouldn't want to update it for this PR, but if you only invoke it when in non-interactive mode then it should still give us the desired behavior.

Naatan (changes requested):
Looks good, but I'd still like to resolve the non-interactive mode:

https://github.com/ActiveState/cli/pull/3072#discussion_r1480643311

Naatan (approved):
Looking good!

PR #3077, opened 2024-02-01
Greatly simplify local order migration mechanic

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2521" title="DX-2521" target="_blank">StoryDX-2521</a>  Local orders uses a simple unintrusive migration mechanic
  </td></table>
  <br />


internal/constraints/constraints.go
```diff
@@ -82,8 +82,7 @@ type projectable interface {
 	Path() string
 	Dir() string
 	URL() string
-	LegacyCommitID() string // for commitmediator.Get
-	StripLegacyCommitID() error
+	LegacyCommitID() string
```
mitchell-as, line 85: Nit: I'd change the comment to be `// for localcommit.Get` since it's not clear why this interface needs this method.

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -1,185 +1,84 @@
...
-	}
+func New(out output.Outputer, pj projecter) *migrator {
+	return &migrator{out: out, proj: pj}
+}
 
-	if os.Getenv(constants.DisableProjectMigrationPrompt) == "true" {
+// setupProject will ensure that the stored project matches the path that was requested. In most cases this should match,
+// and setting up a new project instance will be rare.
+func (m *migrator) setupProject(pjpath string) error {
+	if !ptr.IsNil(m.proj) && m.proj.Dir() == pjpath {
```
mitchell-as, line 36: Nit: we could probably just test `m.proj != nil` and then drop the `ptr` package import.

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -1,185 +1,84 @@
...
-	if err := proj.StripLegacyCommitID(); err != nil {
-		return errs.Wrap(err, "Could not strip legacy commit ID")
-	}
-
-	out.Notice(locale.Tl("projectmigration_success", "Your project was successfully migrated"))
-
-	return nil
+	var err error
+	m.proj, err = project.FromPath(pjpath)
+	return err
```
mitchell-as, line 41: Do we want to wrap the error for stacktrace purposes? Or is it obvious enough where/how to track it down from a message/traceback?

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -1,185 +1,84 @@
...
-	scriptLanguage := "bash"
-	switch runtime.GOOS {
-	case "darwin":
-		scriptValue = migrateMacScript
-	case "windows":
-		scriptValue = migrateWinScript
-		scriptLanguage = "powershell"
+// Migrate returns the given project's commit ID in either the new format (commit file), or the old
+// format (activestate.yaml).
+// If you require the commit file, use localcommit.Get().
```
mitchell-as, line 46: This line is leftover from a previous comment and no longer applies. I would probably update the entire comment anyway, since as it is written, it implies you might call this function for the express purpose of obtaining a commit ID, rather than performing a migration step.

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -1,185 +1,84 @@
...
+func (m *migrator) Migrate(pjpath string) (strfmt.UUID, error) {
+	logging.Debug("Migrating project to new localcommit format: %s", pjpath)
+	if err := m.setupProject(pjpath); err != nil {
+		return "", err
```
mitchell-as, line 50: Do we want to wrap here? I never know when we need to do it, and when we can omit it.

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -1,185 +1,84 @@
...
+	if !strfmt.IsUUID(m.proj.LegacyCommitID()) {
+		return "", locale.NewInputError("err_commit_id_invalid", m.proj.LegacyCommitID())
 	}
 
-	if err := fileutils.WriteFile(proj.Source().Path(), asB); err != nil {
-		return errs.Wrap(err, "Could not write activestate.yaml")
-	}
-
-	return nil
+	return strfmt.UUID(m.proj.LegacyCommitID()), nil
```
mitchell-as, line 70: Nit: using a single variable would avoid referencing the full `m.proj.LegacyCommitID()` more than once.

internal/runbits/legacy/projectmigration/projectmigration.go
```diff
@@ -1,185 +1,84 @@
...
 }
 
-// Only show once per state tool invocation
-var warned = false
-
-func Warn(proj projecter) error {
-	if warned {
-		return nil
+func (m *migrator) Set(pjpath string, commitID string) error {
```
mitchell-as, line 73: `migrator.Set()` by name is not very clear. Maybe rename to `migrator.SetLegacyCommitID()`?

test/integration/branch_int_test.go
```diff
@@ -21,7 +21,7 @@ func (suite *BranchIntegrationTestSuite) TestBranch_List() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
-	ts.PrepareProject("ActiveState-CLI/Branches", "")
+	ts.PrepareProject("ActiveState-CLI/Branches", "35af7414-b44b-4fd7-aa93-2ecad337ed2b")
```
mitchell-as, line 24: May I ask why this is needed? There are other uses of `ts.PrepareProject()` with empty commit IDs. The empty ID may have been intentional for the test.

mitchell-as (changes requested):
This is looking much cleaner! There are only some minor things to address before we can merge this.

internal/runbits/legacy/projectmigration/projectmigration.go
Naatan: That's what I had before, but it doesn't work because it is an interface. So `m.proj == nil` will assert if the interface is nil, it does not assert that the underlying type that satisfies the interface is nil.

mitchell-as: Oh thanks, I learned something new!

internal/runbits/legacy/projectmigration/projectmigration.go
Naatan: I feel like this is redundant. The whole package is to migrate from a legacy version, don't need the methods to repeat that.

I called the Get() function Migrate() because it does more than just Get the value. But for Set that doesn't apply.

pkg/project/project.go
```diff
@@ -214,15 +214,11 @@ func (p *Project) ProjectDir() string {
...
+// LegacyCommitID is for use by legacy mechanics ONLY
 func (p *Project) LegacyCommitID() string {
 	return p.projectfile.LegacyCommitID()
 }
 
-func (p *Project) StripLegacyCommitID() error {
-	return p.projectfile.StripLegacyCommitID()
-}
-
 func (p *Project) SetLegacyCommit(commitID string) error {
```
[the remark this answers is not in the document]
Naatan: Will do for consistency, though to be honest we should not need that comment with a name like this :p

test/integration/branch_int_test.go
Naatan: In this case the migration mechanic was breaking on it since there was nothing to migrate. And a project without a commit ID is malformed. If a test doesn't NEED it then I guess it's not an issue, but in this case it does.

internal/runbits/legacy/projectmigration/projectmigration.go
Naatan: Yeah we should. I was being a bit lazy cause this is for handling legacy code, but you were right to flag me on it. 🙏

internal/constraints/constraints.go
Naatan: I dropped it because it's not necessary. All we need to know is that it's legacy. What code it's used by is what tooling is for. Case in point; all comments like this were essentially invalidated by this PR, and tooling doesn't update comments..

mitchell-as (approved):

PR #3078, opened 2024-02-02
Merge en-us.yaml to address merge conflict.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2499" title="DX-2499" target="_blank">BugDX-2499</a>  state publish doesn't display the name of the published package in its output
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1538,8 +1538,10 @@ uploadingredient_editor_opening:
 uploadingredient_success:
   other: |
     Successfully published as:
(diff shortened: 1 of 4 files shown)


mitchell-as:
@Naatan this will need to be force-merged into v44 and then up to master since I do not have permission, and Target & Verify will fail. There was a locale file conflict when propagating v43 rc2 to v44 branch (https://github.com/ActiveState/cli/actions/runs/7758207588/job/21159431336). This PR resolves that conflict.

PR #3080, opened 2024-02-05
Tests should always supply a commit ID

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2527" title="DX-2527" target="_blank">BugDX-2527</a>  Nightly failure: lots of things that need commitIDs
  </td></table>
  <br />

### DIFF
--- a/test/integration/branch_int_test.go
+++ b/test/integration/branch_int_test.go
@@ -29,7 +29,7 @@ func (suite *BranchIntegrationTestSuite) TestBranch_List() {
 	cp.Expect("To switch to another branch,")
 	cp.ExpectExitCode(0)
 
-	ts.PrepareProject("ActiveState-CLI/small-python", "00000000-0000-0000-0000-000000000000")
+	ts.PrepareProject("ActiveState-CLI/small-python", e2e.CommitIDNotChecked)
 	cp = ts.Spawn("branch")
 	cp.Expect("main")
 	suite.Assert().NotContains(cp.Snapshot(), "To switch to another branch,") // only shows when multiple branches are listed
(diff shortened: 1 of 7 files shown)


test/integration/branch_int_test.go
```diff
@@ -29,7 +29,7 @@ func (suite *BranchIntegrationTestSuite) TestBranch_List() {
 	cp.Expect("To switch to another branch,")
 	cp.ExpectExitCode(0)
 
-	ts.PrepareProject("ActiveState-CLI/small-python", "")
+	ts.PrepareProject("ActiveState-CLI/small-python", "00000000-0000-0000-0000-000000000000")
```
mitchell-as, line 32: Is it worth allowing this to be empty and default to this UUID in `PrepareProject()`? Technically, this UUID does not exist for this project, so leaving it empty from a visual perspective makes sense to me.

test/integration/export_int_test.go
```diff
@@ -65,7 +65,7 @@ func (suite *ExportIntegrationTestSuite) TestExport_ConfigDir() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
-	ts.PrepareProject("cli-integration-tests/Export", "")
+	ts.PrepareProject("cli-integration-tests/Export", "00000000-0000-0000-0000-000000000000")
```
mitchell-as, line 68: If we're not leaving this blank, than maybe we should just always use a valid UUID. It's weird that some tests use a valid UUID and others use this junk one.

mitchell-as (approved):
I'll approve this, but I'm skeptical that explicitly using a junk UUID is a good solution.

test/integration/export_int_test.go
Naatan: That's fair. I definitely wanted to go the explicit route.

mitchell-as (approved):
Perfect, thanks!

PR #3081, opened 2024-02-05
Normalize name given in `state info`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2491" title="DX-2491" target="_blank">StoryDX-2491</a>  `state info` uses normalized names
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/info.go
+++ b/internal/runners/packages/info.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
+	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
 	"github.com/ActiveState/cli/pkg/platform/api/inventory/inventory_models"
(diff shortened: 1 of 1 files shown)


internal/runners/packages/info.go
```diff
@@ -63,7 +64,12 @@ func (i *Info) Run(params InfoRunParams, nstype model.NamespaceType) error {
 		ns = ptr.To(model.NewNamespacePkgOrBundle(language, nstype))
 	}
 
-	packages, err := model.SearchIngredientsStrict(ns.String(), params.Package.Name, true, true, params.Timestamp.Time)
+	normalized, err := model.FetchNormalizedName(*ns, params.Package.Name)
+	if err != nil {
+		multilog.Error("Failed to normalize '%s': %v", params.Package.Name, err)
+	}
+
+	packages, err := model.SearchIngredientsStrict(ns.String(), normalized, false, false, params.Timestamp.Time) // ideally case-sensitive would be true (PB-4371)
```
MDrakos, line 72: Since we're not returning an error when we attempt to fetch the normalized name we will need to check if the name is empty and pass the original. Otherwise, on error we are potentially just passing an empty string. If that's not how `FetchNormalizedName` works can you add a comment?

MDrakos (changes requested):

internal/runners/packages/info.go
mitchell-as: You are correct that the default on error should be the user-provided name. Thanks for catching this.

MDrakos (approved):

PR #3082, opened 2024-02-05
Do not autoupdate while a project is locked.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2490" title="DX-2490" target="_blank">TaskDX-2490</a>  `state update lock` should also disable auto updates
  </td></table>
  <br />

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -231,13 +231,11 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	cmds.OnExecStop(msger.OnExecStop)
 
 	if childCmd != nil && !childCmd.SkipChecks() && !out.Type().IsStructured() {
-		if pj == nil || !pj.IsLocked() {
-			// Auto update to latest state tool version
-			if updated, err := autoUpdate(svcmodel, args, cfg, an, out); err == nil && updated {
-				return nil // command will be run by updated exe
-			} else if err != nil {
-				multilog.Error("Failed to autoupdate: %v", err)
-			}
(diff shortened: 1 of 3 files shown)


Naatan (changes requested):
This should be addressed by `state update lock` toggling the `autoupdate` config var.

https://github.com/ActiveState/cli/blob/3d2b9f229323765b5e9c9b93c7179f613fbb2126/cmd/state/autoupdate.go#L37-L37

mitchell-as:
I'm not sure how I feel about that, given that we're changing a universal configuration option for a project-specific setting. If you were to unlock the project, is it supposed to re-enable the auto update option? What if I've explicitly set `state config set autoupdate false` prior to lock/unlock?

Naatan:
> I'm not sure how I feel about that, given that we're changing a universal configuration option for a project-specific setting. If you were to unlock the project, is it supposed to re-enable the auto update option? What if I've explicitly set `state config set autoupdate false` prior to lock/unlock?

I share your reservations, but Pete made the call after I shared them with him. 

@rawktron fyi. 

I think it might be good for us to review the use-cases here again and find a good solution for them, even if it means breaking the current command (which this is kinda already doing).

Anyway, this isn't a blocker for landing this ticket. As I said Pete and I already discussed, don't want to keep going back on decisions made, just tagging Pete here since you also raised the concern @mitchell-as.

Naatan (changes requested):
Looks good, but needs error checks on `cfg.Set()`.

Naatan (approved):

PR #3083, opened 2024-02-05
Include original webclient ID in auth and signup URLs.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2529" title="DX-2529" target="_blank">StoryDX-2529</a>  Pass webclient_id to signup and sign-in pages
  </td></table>
  <br />
 

The login URL will look like this:

```
https://platform.activestate.com/authorize/device?user-code=<uuid>&webclient_id=<id>
```

and the signup URL will look like this:

```
https://platform.activestate.com/create-account?nextRoute=%2Fauthorize%2Fdevice%3Fuser-code%3D<uuid>&webclient_id=<id>
```

Note that `webclient_id` for the signup page is part of the request to `create-account`, not a follow up query to the `nextRoute` page.


Naatan (changes requested):
I think this is trying to merge v0.44 into v0.43?

mitchell-as:
Whoops, I didn't see this was supposed to be for v43. Sorry about that :(

PR #3084, opened 2024-02-05
Quote state executable with spaces in name for `state shell`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2487" title="DX-2487" target="_blank">BugDX-2487</a>  `state shell` messes with whitespace in PATH
  </td></table>
  <br />
 

Without the quotes, the first part before the space would be incorrectly considered a command.

### DIFF
--- a/internal/subshell/sscommon/rcfile.go
+++ b/internal/subshell/sscommon/rcfile.go
@@ -348,7 +348,11 @@ func SetupProjectRcFile(prj *project.Project, templateName, ext string, env map[
 	pathList, ok := env["PATH"]
 	inPathList, err := fileutils.PathInList(listSep, pathList, currExecAbsDir)
 	if !ok || !inPathList {
-		rcData["ExecAlias"] = currExec // alias {ExecName}={ExecAlias}
+		safeExec := currExec
+		if strings.Index(currExec, " ") != -1 {
+			safeExec = fmt.Sprintf(`"%s"`, currExec) // quote for alias
+		}
+		rcData["ExecAlias"] = safeExec // alias {ExecName}={ExecAlias}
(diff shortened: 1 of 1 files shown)


internal/subshell/sscommon/rcfile.go
```diff
@@ -348,7 +348,11 @@ func SetupProjectRcFile(prj *project.Project, templateName, ext string, env map[
 	pathList, ok := env["PATH"]
 	inPathList, err := fileutils.PathInList(listSep, pathList, currExecAbsDir)
 	if !ok || !inPathList {
-		rcData["ExecAlias"] = currExec // alias {ExecName}={ExecAlias}
+		safeExec := currExec
+		if strings.Index(currExec, " ") != -1 {
+			safeExec = fmt.Sprintf(`"%s"`, currExec) // quote for alias
+		}
+		rcData["ExecAlias"] = safeExec // alias {ExecName}={ExecAlias}
```
MDrakos, line 355: I believe that `strings.Index` returns `-1` if the substrings _isn't_ found. So this logic seems backwards.

You could replace the function call here with `strings.ContainsAny(curreExec, " ")` or with a simple regex like `\s`

MDrakos (changes requested):

internal/subshell/sscommon/rcfile.go
mitchell-as: I did a double-take as well. In this case, if there is a space found, we want to quote the executable (e.g. `"/path to state.exe"`). Otherwise, it can remain unquoted.

MDrakos (approved):

PR #3085, opened 2024-02-06
Revert runtime in use checking.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2472" title="DX-2472" target="_blank">BugDX-2472</a>  `state checkout` has awkward pause
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1259,6 +1259,8 @@ setup_runtime:
   other: "Setting Up Runtime"
 progress_solve:
   other: "Resolving Dependencies"
+progress_runtime_in_use_check:
+  other: Verifying runtime executables are not currently in use
 additional_dependencies:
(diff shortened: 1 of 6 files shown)


mitchell-as:
Test failure is not due to this PR.

Naatan:
@mitchell-as this issue only started happening for me around the time we started work on v0.44. Whereas this functionality you're updating here hasn't been touched in quite a while (afaik?). Are you sure this isn't a red herring?

My thought was that this slowdown is likely caused by the change in buildplan sourcing needed for dependency calculation.

mitchell-as:
@Naatan No, I confirmed that it's caused by the runtime in use checking introduced recently. It's a macOS-only issue and the telltale sign is Terminal.app changing the tab label to `lsof` during the pause.

mitchell-as:
Closing in favor of #3088

PR #3086, opened 2024-02-06
Include original webclient ID in auth and signup URLs.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2529" title="DX-2529" target="_blank">StoryDX-2529</a>  Pass webclient_id to signup and sign-in pages
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/auth/login.go
+++ b/internal/runbits/auth/login.go
@@ -4,6 +4,7 @@ import (
 	"net/url"
 	"time"
 
+	anaConst "github.com/ActiveState/cli/internal/analytics/constants"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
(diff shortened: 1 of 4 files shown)


internal/runbits/auth/login.go
```diff
@@ -230,6 +231,9 @@ func authenticateWithBrowser(out output.Outputer, auth *authentication.Auth, pro
 
 		verificationURL = signupURL.String()
 	}
+	if webclientId := cfg.GetString(anaConst.CfgSessionToken); webclientId != "" {
+		verificationURL += "&webclient_id=" + webclientId
```
Naatan, line 235: We should use `url.Parse()` here and ensure this is properly being attached.

Naatan (changes requested):

Naatan (approved):

PR #3087, opened 2024-02-06
Revert localcommit mechanic, but preserve package for further refactors

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2534" title="DX-2534" target="_blank">TaskDX-2534</a>  Revert local commit mechanic
  </td></table>
  <br />


internal/runners/initialize/init.go
```diff
@@ -246,13 +241,6 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 	if err := localcommit.Set(proj.Dir(), commitID.String()); err != nil {
 		return errs.Wrap(err, "Unable to create local commit file")
 	}
-	if emptyDir || fileutils.DirExists(filepath.Join(path, ".git")) {
-		err := localcommit.AddToGitIgnore(path)
-		if err != nil {
-			r.out.Notice(locale.Tr("notice_commit_id_gitignore", constants.ProjectConfigDirName, constants.CommitIdFileName))
```
mitchell-as, line 252: Please remove this entry in the en-us.yaml locale file.

internal/runbits/requirements/rationalize.go
```diff
@@ -21,12 +19,6 @@ func (r *RequirementOperation) rationalizeError(err *error) {
 	case err == nil:
 		return
 
-	// Local commit file not found or malformed
-	case errors.As(*err, &localCommitFileErr):
-		*err = errs.WrapUserFacing(*err,
-			locale.Tr("err_local_commit_file", localCommitFileErr.File),
```
mitchell-as, line 27: Please remove the en-us.yaml locale entry for this.

pkg/localcommit/localcommit.go
```diff
@@ -1,151 +1,53 @@
...
-
-var handler migrationHandler = nil
-
-func RegisterMigrator(h migrationHandler) {
-	handler = h
-}
+// proj holds the project instance most recently accessed, if any.
+// Using globals in this way is an anti-pattern, but because the commit mechanic is going through a lot of changes
+// we're currently handling it this way to help further refactors. Once we've landed the go-forward mechanic we should
+// remove this anti-pattern.
```
mitchell-as, line 13: You may want to add a JIRA ticket to this comment so we don't lose track of it.

mitchell-as (changes requested):
This largely looks complete. There are failing unit tests though. Please run "Test: all" prior to submitting for re-review, so we can proactively avoid any nightly failures caused by this large change.

test/integration/switch_int_test.go
```diff
@@ -19,7 +18,7 @@ type SwitchIntegrationTestSuite struct {
 
 func (suite *SwitchIntegrationTestSuite) TestSwitch_Branch() {
 	suite.OnlyRunForTags(tagsuite.Switch)
-	ts := e2e.New(suite.T(), false)
+	ts := e2e.New(suite.T(), true)
```
mitchell-as, line 21: Is this an intentional change?

internal/testhelpers/e2e/session.go
```diff
@@ -332,6 +332,14 @@ func (s *Session) PrepareCommitIdFile(commitID string) {
 	require.NoError(s.T, pjfile.SetLegacyCommit(commitID))
 }
 
+// CommitID is used to grab the current commit ID for the project in our working directory. This is distinct
+// from localcommit.Get() and pjfile.LegacyCommitID() in that it always produces a fresh result from the on-disk file.
```
mitchell-as, line 336: Does this mean we shouldn't be using `localcommit.Get()` or `pjfile.LegacyCommitID()` from integration tests, that we should only be using this? It might be good to add that to this comment. Otherwise, it's not clear to me why we should be using this function.

mitchell-as (dismissed):
I will approve this and assume you successfully ran Test: all (I'm noticing a timeout for the most recent set of tests).

There are a couple things you may want to consider before merging this, though. I've added some comments.

mitchell-as (approved):

PR #3093, opened 2024-02-07
Fix `state publish` failing on new versions

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2543" title="DX-2543" target="_blank">BugDX-2543</a>  state publish fails if given new version number
  </td></table>
  <br />

### DIFF
--- a/internal/installation/appinfo.go
+++ b/internal/installation/appinfo.go
@@ -44,12 +44,11 @@ func newExecFromDir(baseDir string, exec executableType) (string, error) {
 		path = filepath.Dir(osutils.Executable())
 	}
 
-	// Work around dlv debugger giving an unexpected executable path
-	if !condition.BuiltViaCI() && len(os.Args) > 1 && strings.Contains(os.Args[0], "__debug_bin") {
+	// Work around dlv and goland debugger giving an unexpected executable path
+	if !condition.BuiltViaCI() && len(os.Args) > 1 &&
+		(strings.Contains(os.Args[0], "__debug_bin") || strings.Contains(filepath.ToSlash(os.Args[0]), "GoLand/___")) {
 		rootPath := filepath.Clean(environment.GetRootPathUnsafe())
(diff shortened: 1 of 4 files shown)


internal/installation/appinfo.go
```diff
@@ -44,12 +44,11 @@ func newExecFromDir(baseDir string, exec executableType) (string, error) {
...
-	// Work around dlv debugger giving an unexpected executable path
-	if !condition.BuiltViaCI() && len(os.Args) > 1 && strings.Contains(os.Args[0], "__debug_bin") {
+	// Work around dlv and goland debugger giving an unexpected executable path
+	if !condition.BuiltViaCI() && len(os.Args) > 1 &&
+		(strings.Contains(os.Args[0], "__debug_bin") || strings.Contains(filepath.ToSlash(os.Args[0]), "GoLand/___")) {
 		rootPath := filepath.Clean(environment.GetRootPathUnsafe())
-		if rootPath == filepath.Clean(path) {
-			path = filepath.Join(path, "build")
-		}
+		path = filepath.Join(rootPath, "build")
```
Naatan, line 51: Ran into this while debugging. Because of `BuiltViaCI()` this feels safe enough to include.

internal/runners/publish/publish.go
```diff
@@ -143,12 +143,16 @@ func (r *Runner) Run(params *Params) error {
 	isRevision := false
 	if params.Version != "" {
 		// Attempt to get the version if it already exists, it not existing is not an error though
-		i, err := model.GetIngredientByNameAndVersion(params.Namespace, params.Name, params.Version)
+		i, err := model.GetIngredientByNameAndVersion(reqVars.Namespace, reqVars.Name, params.Version)
```
Naatan, line 146: Note this also fixes another bug I found; this was still using the namespace and name from the user input, which might be empty.

PR #3095, opened 2024-02-07
Show dependency list on checkout.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2530" title="DX-2530" target="_blank">StoryDX-2530</a>  Show dependency tree on checkout
  </td></table>
  <br />
 

Sample:

```
>..\cli\build\state.exe checkout org/project sm2
Checking out project: org/project

█ Installing Runtime

Setting Up Runtime
Resolving Dependencies ✔ Done

Setting up the following dependencies:
├─ python@3.10.13 (8 dependencies)
├─ requests@2.31.0 (31 dependencies)
└─ scipy@1.10.1 (25 dependencies)
```

On the Dashboard:

image


mitchell-as:

internal/locale/locales/en-us.yaml
```diff
@@ -1439,6 +1439,14 @@ shell_project_statement:
...
     This shell is operating inside a virtual environment.
     For editors and other tooling use the executables at: [ACTIONABLE]{{.V2}}[/RESET].
+checking_out:
+  other: |
+    Checking out project: [NOTICE]{{.V0}}[/RESET]
+installing_runtime_title:
+  other: Installing Runtime
+installing_runtime_and_dependencies:
+  other: |
+    Installing your runtime and dependencies.
```
Naatan, line 1449: Looking at the mock again, this feels redundant. Let's drop it (I've updated the mock).

internal/runbits/dependencies/summary.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		subdependencies := ""
+		if numSubs := len(dependencies[dep.ArtifactID]); numSubs > 0 {
+			subdependencies = fmt.Sprintf("([ACTIONABLE]%s[/RESET] dependencies)", strconv.Itoa(numSubs))
+		}
+
+		item := fmt.Sprintf("[ACTIONABLE]%s@%s[/RESET] %s", dep.Name, version, subdependencies)
+
+		out.Notice(fmt.Sprintf("  [DISABLED]%s[/RESET] %s", prefix, item))
```
Naatan, line 53:
```suggestion
		out.Notice(fmt.Sprintf("[DISABLED]%s[/RESET] %s", prefix, item))
```

internal/runbits/dependencies/summary.go
```diff
@@ -0,0 +1,57 @@
...
+// OutputSummary lists the given runtime dependencies being installed along with their
+// subdependencies (if any).
+func OutputSummary(out output.Outputer, directDependencies []artifact.ArtifactID, artifacts artifact.Map) {
+	if len(directDependencies) == 0 {
+		return
+	}
+
+	dependencies := make(map[artifact.ArtifactID][]artifact.ArtifactID)
+	for _, artifactId := range directDependencies {
+		subdependencies := buildplan.RecursiveDependenciesFor(artifactId, artifacts)
```
Naatan, line 23: Currently, this shows top level dependencies, whereas the story calls for "direct requirements of the project". I worded that very poorly though.. what I meant was it should show the requirements the user specified in their commit.

Naatan (changes requested):

internal/runbits/dependencies/summary.go
mitchell-as: Okay. I've updated this and updated the sample output in the PR summary.

pkg/platform/runtime/setup/setup.go
```diff
@@ -550,9 +550,21 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
 		return nil, nil, errs.Wrap(err, "Failed to compute artifacts to build")
 	}
 
-	// Output a change summary if applicable.
-	if len(oldBuildPlanArtifacts) > 0 {
-		changesummary.New(s.out).ChangeSummary(changedArtifacts, artifactsToBuild, oldBuildPlanArtifacts)
+	// Output a dependency summary if applicable.
+	if !fileutils.DirExists(s.target.Dir()) {
```
Naatan, line 554: There's a potential order of operations issue here for future updates. The dir existing has implicit behaviour tied to it, which we may break without realizing in future updates. 

I think at this point we should have access to the trigger, can we identify checkout from there and then run this logic? 

I think either we need to make this logic more explicit, or we need to have a test covering it to ensure we don't break it.

Naatan (changes requested):

pkg/platform/runtime/setup/setup.go
mitchell-as: Ahhhh, thanks! Checking the trigger is exactly what I wanted. I definitely did not like the directory existing check, but I couldn't figure out how else to figure out if this was a fresh checkout or not.

Naatan (approved):
Looking good!

PR #3099, opened 2024-02-11
Clean up `state search` UI

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2400" title="DX-2400" target="_blank">StoryDX-2400</a>  Clean up the `state search` UI
  </td></table>
  <br />
 

This uses the `viewport` bubble to handle rendering and scrolling. Example [here](https://github.com/charmbracelet/bubbletea/blob/master/examples/pager/main.go) for reference.


internal/runners/packages/search.go
```diff
@@ -98,113 +149,34 @@ func targetedLanguage(languageOpt string, proj *project.Project) (string, error)
...
-
-		if row.versions > 1 {
-			olderVersions := row.versions - 1
-			newRow.OlderVersions = locale.Tl("search_older_versions", "+ {{.V0}} older versions", strconv.Itoa(olderVersions))
-		}
-		mergedRows = append(mergedRows, newRow)
+	vulnMap := make(map[string][]*model.VulnerabilityIngredient)
+	for _, v := range vulns {
+		key := ingredientVulnKey(v.PrimaryNamespace, v.Name, v.Version)
+		vulnMap[key] = append(vulnMap[key], v)
```
Naatan, line 170: We should move this to the model.

internal/runners/packages/search.go
```diff
@@ -71,7 +83,46 @@ func (s *Search) Run(params SearchRunParams, nstype model.NamespaceType) error {
...
+	}
+	v.content = table.Content()
+	v.packageNames = table.packageNames
+
+	if s.out.Type().IsStructured() {
+		s.out.Print(table)
+		return nil
+	}
+
+	if _, err := p.Run(); err != nil {
```
Naatan, line 123: Not sure where to tag this; but this doesn't work with non-interactive (ie. `-n`).

internal/runners/packages/searchTable.go
```diff
@@ -0,0 +1,222 @@
...
+			result.Website = pkg.Ingredient.Website.String()
+		}
+		if pkg.LatestVersion.LicenseExpression != nil {
+			result.License = *pkg.LatestVersion.LicenseExpression
+		}
+
+		var versions []string
+		for i, v := range pkg.Versions {
+			if i > 5 {
+				versions = append(versions, fmt.Sprintf("... (%d more)", len(pkg.Versions)-5))
```
Naatan, line 82: Not localized

internal/runners/packages/searchTable.go
```diff
@@ -0,0 +1,222 @@
...
+			if high > 0 {
+				result.Vulnerabilities["High"] = high
+			}
+			if medium > 0 {
+				result.Vulnerabilities["Medium"] = medium
+			}
+			if low > 0 {
+				result.Vulnerabilities["Low"] = low
+			}
+		}
```
Naatan, line 120: This feels like it should live in the model, maybe where `vulns[<key>]` is being calculated.

internal/runners/packages/searchTable.go
```diff
@@ -0,0 +1,222 @@
...
+		if pkg.Ingredient.Description != nil {
+			result.Description = *pkg.Ingredient.Description
+		}
+		if pkg.Ingredient.Website != "" {
+			result.Website = pkg.Ingredient.Website.String()
+		}
+		if pkg.LatestVersion.LicenseExpression != nil {
+			result.License = *pkg.LatestVersion.LicenseExpression
+		}
```
Naatan, line 77: Check out `ptr.From()`.

internal/runners/packages/searchTable.go
```diff
@@ -0,0 +1,222 @@
...
+
+	wrapped := valueStyle.Render(value)
+
+	// The rendered line ends up being a bit too long, so we need to reduce the
+	// width that we are working with to ensure that the wrapped value fits
+	indentedValue := strings.ReplaceAll(wrapped, "\n", "\n"+strings.Repeat(" ", len(paddedKey)-8))
+
+	formattedRow := fmt.Sprintf("%s%s", paddedKey, indentedValue)
+	return rowStyle.Render(formattedRow) + "\n"
+}
```
Naatan, line 222: This is a bit hard to follow, could you document the function, maybe show an example?

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,134 @@
...
+	content      string
+	remaining    int
+	packageNames []string
+	ready        bool
+	err          error
+	viewport     viewport.Model
+}
+
+func NewView() (*view, error) {
+	width, height, err := term.GetSize(int(os.Stdout.Fd()))
```
Naatan, line 35: Don't use `os` directly, instead use `output.Config().OutWriter`.

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,134 @@
...
+				if strings.Contains(l, "Name") {
+					if v.remaining <= 0 {
+						v.remaining = 0
+					} else {
+						v.remaining--
+					}
+				}
+			}
+			return v, nil
+		}
```
Naatan, line 82: It seems viewport also supports page up and page down, which this does not check for. Once you start using an alternate mode of scrolling not covered here the remaining number is out of sync. Wonder if there is a scroll event we can listen for?

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,134 @@
...
+		v.remaining = 0
+	}
+}
+
+func (v *view) footerView() string {
+	var footerText string
+	if v.remaining != 0 {
+		footerText += locale.Tl("search_more_matches", "... {{.V0}} more matches, press Down to scroll", strconv.Itoa(v.remaining))
+	}
+	footerText += fmt.Sprintf("\n%s'%s'", styleBold.Render(locale.Tl("search_more_info", "For more info run")), styleActionable.Render(locale.Tl("search_more_info_command", " state info <name>")))
```
Naatan, line 132:
```suggestion
	footerText += fmt.Sprintf("\n\n%s'%s'", styleBold.Render(locale.Tl("search_more_info", "For more info run")), styleActionable.Render(locale.Tl("search_more_info_command", " state info <name>")))
```

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,134 @@
...
+	v.remaining = len(v.packageNames) - currentEntryIndex - 1
+	if v.remaining < 0 {
+		v.remaining = 0
+	}
+}
+
+func (v *view) footerView() string {
+	var footerText string
+	if v.remaining != 0 {
+		footerText += locale.Tl("search_more_matches", "... {{.V0}} more matches, press Down to scroll", strconv.Itoa(v.remaining))
```
Naatan, line 130:
```suggestion
		footerText += locale.Tl("search_more_matches", "... {{.V0}} more matches, use arrow and page keys to scroll. Press Q to quit.", strconv.Itoa(v.remaining))
```

Naatan, line 130: Can we make this support the mouse scroll button?

Naatan (changes requested):

internal/runners/packages/searchView.go
MDrakos: There does not appear to be a scroll event. The viewport has a scroll percent function that will return the remaining percentage left. We might want to consider that if we're going to implement alternate scrolling methods. We could also look to some sort of index calculation. The mock just had `down` to scroll so that's all that is here. Do we want to take this further or file a follow up?

internal/runners/packages/searchView.go
MDrakos: Yes, I was able to get this working.

internal/runners/packages/searchView.go
MDrakos: This is expecting a file descriptor. The `OutWriter` is an `io.Writer` interface.

internal/runners/packages/searchTable.go
MDrakos: Tried to add a few more comments. I agree it's not easy to follow. If this batch doesn't help I can try to revisit it again.

internal/runners/packages/search.go
```diff
@@ -71,7 +83,46 @@ func (s *Search) Run(params SearchRunParams, nstype model.NamespaceType) error {
...
+
+	table, err := createSearchTable(v.width, v.height, processedPackages, vulns)
+	if err != nil {
+		return errs.Wrap(err, "Could not create search table")
+	}
+	v.content = table.Content()
+	v.packageNames = table.packageNames
+
+	if s.out.Type().IsStructured() {
+		s.out.Print(table)
```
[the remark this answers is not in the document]
MDrakos: I've moved the table view calculation to the actual view type. We are still using this table type to calculate the structured search results and the view does further processing for the plain output. Otherwise we dump the table from the search results.

internal/runners/packages/search.go
```diff
@@ -71,7 +83,46 @@ func (s *Search) Run(params SearchRunParams, nstype model.NamespaceType) error {
...
+			return errs.Wrap(err, "Could not fetch vulnerabilities")
+		}
+	}
+
+	table, err := createSearchTable(v.width, v.height, processedPackages, vulns)
+	if err != nil {
+		return errs.Wrap(err, "Could not create search table")
+	}
+	v.content = table.Content()
+	v.packageNames = table.packageNames
```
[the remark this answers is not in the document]
MDrakos: I've updated this a bit so that the view is more encapsulated. We are still using this type to process the search results but any view processing for the viewport is now done elsewhere.

internal/runners/packages/searchView.go
Naatan: We should solve it as part of this ticket. I think showing a percentage indication would be fine too, it's just about telling the user that there is something to scroll to since there is no scrollbar.

internal/runners/packages/searchView.go
Naatan: Can we add the `Fd()` method on `Outwriter`? I'm not confident that we'll end up using this long term but having everything agree on a same source of truth will make refactoring this later easier.

MDrakos: Added a method to the output `Config` type.

internal/runners/packages/searchView.go
MDrakos: Managed to get this and mouse scroll working with an index.

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,304 @@
...
+		v.remaining = 0
+	}
+}
+
+func (v *view) footerView() string {
+	var footerText string
+	if v.remaining != 0 {
+		footerText += locale.Tl("search_more_matches", "... {{.V0}} more matches, use arrow and page keys to scroll. Press Q to quit.", strconv.Itoa(v.remaining))
+	}
+	footerText += fmt.Sprintf("\n\n%s'%s'", styleBold.Render(locale.Tl("search_more_info", "For more info run")), styleActionable.Render(locale.Tl("search_more_info_command", " state info <name>")))
```
Naatan, line 244:
```suggestion
	footerText += fmt.Sprintf("\n\n%s '%s'", styleBold.Render(locale.Tl("search_more_info", "For more info run")), styleActionable.Render(locale.Tl("search_more_info_command", "state info <name>")))
```

internal/runners/packages/search.go
```diff
@@ -71,7 +76,34 @@ func (s *Search) Run(params SearchRunParams, nstype model.NamespaceType) error {
...
+		}
+	}
+
+	results, err := createSearchResults(packages, vulns)
+	if err != nil {
+		return errs.Wrap(err, "Could not create search table")
+	}
+
+	if s.out.Type().IsStructured() || !s.out.Config().Interactive {
+		s.out.Print(results)
```
Naatan, line 93: For non-interactive, it seems to restrict the first column size too aggressively:

```
   Name           tensor-algebra-dont-use
   Description    bla
   License        MIT
   Versions        - 0.0.1
   Vulnerabili    <nil>
  ties
```

Notice the Vulnerabilities title being wrapped.

Also note it holds the value `<nil>`, probably needs an omitEmpty.

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,304 @@
...
+	return nil
+}
+
+func (v *view) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	var (
+		cmd  tea.Cmd
+		cmds []tea.Cmd
+	)
+
+	switch msg := msg.(type) {
```
Naatan, line 87: It looks like scrolling is already handled through the viewport lib:

https://github.com/charmbracelet/bubbles/blob/master/viewport/viewport.go#L287

Is there a reason we can't make use of that?

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,304 @@
...
+	if direction == "up" {
+		// LineUp returns the new lines. In order to update the remaining count
+		// we need to iterate over the visible lines and find the remaining count
+		// for the last visible package name.
+		_ = v.viewport.LineUp(amount)
+		lines := strings.Split(v.viewport.View(), "\n")
+		for _, l := range lines {
+			if strings.Contains(l, "Name") {
+				v.remaining = len(v.searchResults.packageNames) - v.index[l]
+			}
```
Naatan, line 258: This is very hacky and prone to bugs. I'd rather we just show the percentage if the alternative is comparing strings.

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,304 @@
...
+			v.viewport.SetContent(v.processContent())
+			v.initialRemaining()
+			v.ready = true
+		} else {
+			v.width = msg.Width
+			v.height = msg.Height
+		}
+	case errMsg:
+		v.err = msg
+		return v, nil
```
Naatan, line 129: How might this happen?

internal/runners/packages/styles.go
```diff
@@ -0,0 +1,18 @@
...
+	colorYellow     = lipgloss.Color("4")
+	colorMagenta    = lipgloss.Color("5")
+	colorCyan       = lipgloss.Color("6")
+	styleCyan       = lipgloss.NewStyle().Foreground(colorCyan)
+	styleRed        = lipgloss.NewStyle().Foreground(colorRed)
+	styleOrange     = lipgloss.NewStyle().Foreground(colorOrange)
+	styleYellow     = lipgloss.NewStyle().Foreground(colorYellow)
+	styleMagenta    = lipgloss.NewStyle().Foreground(colorMagenta)
+	styleBold       = lipgloss.NewStyle().Bold(true)
+	styleActionable = lipgloss.NewStyle().Bold(true).Foreground(colorCyan)
```
Naatan, line 17: Can we move these definitions to the colorize package? That'll make it easier to keep these in sync. Perhaps we can even link them together.

pkg/platform/model/vulnerabilities.go
```diff
@@ -29,6 +29,15 @@ func (v Vulnerabilites) Length() int {
 	return len(v.Critical) + len(v.High) + len(v.Medium) + len(v.Low)
 }
 
+func (v *Vulnerabilites) Count() map[string]int {
+	return map[string]int{
+		"Critical": len(v.Critical),
+		"High":     len(v.High),
+		"Medium":   len(v.Medium),
+		"Low":      len(v.Low),
```
Naatan, line 37: I think these keys might need localization? Least the caps makes it seem these are intended for displaying.

Naatan (changes requested):
I can't select text in my terminal when viewing search results. There's a weird symbol on my cursor that I think indicates it has some sort of lock. Possibly this is something the viewport imposes, but we might be able to configure it?

internal/runners/packages/search.go
MDrakos: The AC states that for non-interactive we just dump everything, hence the lack of processing.

internal/runners/packages/searchView.go
MDrakos: Took some time to try and use this. The `Update` method does not implement the correct interface so it's not a simple replacement. We would likely have to do some wrapping. I didn't realize the viewport had these methods as I was following the example that I linked in my initial comment.

internal/runners/packages/search.go
Naatan: Just dump as in don't worry about scrolling. It does not mean we do not care about the output.

internal/runners/packages/searchView.go
MDrakos: Ahh good catch. This is a leftover from my initial attempts to get the view working and is no longer needed.

pkg/platform/model/vulnerabilities.go
MDrakos: Not localization but definitely can use constants which have already been created.

internal/runners/packages/search.go
MDrakos: I was able to fix the `omitEmpty` however the spacing seems to be in the output library. Can I propose that we rename this field to `CVEs` for now and file a follow up to address what's going on with the table output?

internal/runners/packages/searchView.go
MDrakos: While I like how this works, I'm also weary about this approach. I've updated it to use the `ScrollPercent()` function. The localization is my best guess so I expect you will want to update that.

MDrakos:
>I can't select text in my terminal when viewing search results. There's a weird symbol on my cursor that I think indicates it has some sort of lock. Possibly this is something the viewport imposes, but we might be able to configure it?

Unfortunately, it doesn't appear that this is currently supported: https://github.com/charmbracelet/bubbletea/issues/162

mitchell-as:
Try holding down Shift and click-dragging with the mouse. At least in Linux this should work with terminals whose current applications are handling mouse events.

internal/runners/packages/searchView.go
Naatan: Can't you just call the viewport's `Update()` method from your `Update()` method? ie. handle whatever you need to handle, then defer the rest to the viewport package.

Naatan:
> > I can't select text in my terminal when viewing search results. There's a weird symbol on my cursor that I think indicates it has some sort of lock. Possibly this is something the viewport imposes, but we might be able to configure it?
> 
> Unfortunately, it doesn't appear that this is currently supported: [charmbracelet/bubbletea#162](https://github.com/charmbracelet/bubbletea/issues/162)

Bah.. ok let's revert mouse scrolling then. It's more important that you can select text than that you can scroll with your mouse. And `git log` apparently has the same limitation.

@mitchell-as shift+click didn't do anything for me in iTerm :\

internal/runners/packages/searchresult.go
```diff
@@ -0,0 +1,71 @@
...
+
+func createSearchResults(packages []*model.IngredientAndVersion, vulns []*model.VulnerabilityIngredient) (*structuredSearchResults, error) {
+	var results []*searchResult
+	var packageNames []string
+	for _, pkg := range packages {
+		result := &searchResult{}
+		result.Name = ptr.From(pkg.Ingredient.Name, "")
+		result.Description = ptr.From(pkg.Ingredient.Description, "")
+		result.Website = pkg.Ingredient.Website.String()
+		result.License = ptr.From(pkg.LatestVersion.LicenseExpression, "")
```
Naatan, line 34: Author information is missing.

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		if pkg.Description != "" {
+			doc.WriteString(formatRow(colorize.StyleBold.Render(keyDescription), pkg.Description, maxKeyLength, v.width))
+		}
+		if pkg.Website != "" {
+			doc.WriteString(formatRow(colorize.StyleBold.Render(keyWebsite), colorize.StyleCyan.Render(pkg.Website), maxKeyLength, v.width))
+		}
+		if pkg.License != "" {
+			doc.WriteString(formatRow(colorize.StyleBold.Render(keyLicense), pkg.License, maxKeyLength, v.width))
+		}
```
Naatan, line 157: Note most of these values should be cyan (see mock).

I've also updated the mock to make the name actionable, so bold and cyan, since it makes sense that we want the name to stand out more than any of the other info.

internal/runners/packages/searchView.go
```diff
@@ -0,0 +1,245 @@
...
+		renderedKey := colorize.StyleBold.Render(key)
+		if len(renderedKey) > maxKeyLength {
+			maxKeyLength = len(renderedKey) + 2
+		}
+	}
+
+	doc := strings.Builder{}
+	for _, pkg := range v.searchResults.Results {
+		if pkg.Name != "" {
+			doc.WriteString(formatRow(colorize.StyleBold.Render(keyName), pkg.Name, maxKeyLength, v.width))
```
Naatan, line 147: I've updated the mock to make the keys greyed out (ie. the disabled style).

Playing with the live results the emphasis is too much on the keys at the moment. Making them grey should help with de-emphasising the keys and focussing the attention on the actual results. This'll also help with making the name stand out.

Naatan (changes requested):

internal/runners/packages/searchView.go
MDrakos: Took another look at this and no, it's not that simple. Again, the `Update()` function return a struct that doesn't satisfy the `tea.Model` interface. This means that we can't just return the values that the `Update()` function returns.

I didn't spend _a lot_ of time trying to get this working but I managed to by replacing the underlying viewport. Unfortunately, this means we lose the header and footer spacing that we want to setup so it looks a bit off.

Looking at bit more at the code you linked it's not immediately evident what benefit we get from using this `Update` function. It would be handling the scrolling keys for us but that's about it.

If there is some other benefit or if this is something that we need before this can be merged I can try to get it working but it will take some time.

internal/runners/packages/searchresult.go
MDrakos: I should have noted this somewhere but we discussed during stand that retrieving the author information requires us to make a separate API call for each package in the search results. We deemed that this would slow down the command too much so opted to leave the authors out.

internal/runners/packages/search.go
MDrakos: Filed here: https://activestatef.atlassian.net/browse/DX-2578

Naatan (changes requested):
Nearly there, two remaining issues:

1. Vulnerabilities is not colored correctly, and has inconsistent indentation:

image

2. Authors is still missing.

MDrakos:
Missed the vulnerabilities key on the last round of changes. Should be updated now as well as the vulnerability colours.

As for the authors see this comment: https://github.com/ActiveState/cli/pull/3099#discussion_r1491656889

internal/runners/packages/searchresult.go
Naatan: Thanks, forgot about that. I filed DX-2581 and a blocker with XT to hopefully get this resolved.

Naatan (approved):
Great work!

PR #3101, opened 2024-02-12
Added `--namespace` and `--commit` flags for `state builds` and subcommands.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2513" title="DX-2513" target="_blank">StoryDX-2513</a>  `state builds` and subcommands work with remote projects
  </td></table>
  <br />

### DIFF
--- a/cmd/state/internal/cmdtree/builds.go
+++ b/cmd/state/internal/cmdtree/builds.go
@@ -5,11 +5,12 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runners/builds"
+	"github.com/ActiveState/cli/pkg/project"
 )
 
 func newBuildsCommand(prime *primer.Values) *captain.Command {
 	runner := builds.New(prime)
-	params := &builds.Params{}
+	params := &builds.Params{Namespace: &project.Namespaced{}}
 
 	cmd := captain.NewCommand(
 		"builds",
@@ -19,8 +20,18 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
 		[]*captain.Flag{
 			{
 				Name:        "all",
-				Description: "List all builds, including individual package artifacts",
+				Description: locale.Tl("builds_flags_all_description", "List all builds, including individual package artifacts"),
 				Value:       &params.All,
+			},
+			{
+				Name:        "namespace",
+				Description: locale.Tl("builds_flags_namespace_description", "The namespace of the project to inspect builds for"),
+				Value:       params.Namespace,
+			},
+			{
+				Name:        "commit",
+				Description: locale.Tl("builds_flags_commit_description", "The commit ID to inspect builds for"),
+				Value:       &params.CommitID,
 			},
 		},
 		[]*captain.Argument{},
@@ -37,14 +48,25 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
 
 func newBuildsDownloadCommand(prime *primer.Values) *captain.Command {
 	runner := builds.NewDownload(prime)
-	params := &builds.DownloadParams{}
+	params := &builds.DownloadParams{Namespace: &project.Namespaced{}}
 
 	cmd := captain.NewCommand(
 		"dl",
 		locale.Tl("builds_download_title", "Download build artifacts"),
 		locale.Tl("builds_download_description", "Download build artifacts for a given build"),
 		prime,
-		[]*captain.Flag{},
+		[]*captain.Flag{
+			{
+				Name:        "namespace",
+				Description: locale.Tl("builds_download_flags_namespace_description", "The namespace of the project to download artifacts from"),
+				Value:       params.Namespace,
+			},
+			{
+				Name:        "commit",
+				Description: locale.Tl("builds_download_flags_commit_description", "The commit ID to download artifacts from"),
+				Value:       &params.CommitID,
+			},
+		},
 		[]*captain.Argument{
 			{
 				Name:        "ID",
--- a/internal/runners/builds/builds.go
+++ b/internal/runners/builds/builds.go
@@ -14,8 +14,9 @@ import (
 	"github.com/ActiveState/cli/internal/runbits/rationalize"
 	"github.com/ActiveState/cli/internal/runbits/runtime"
 	bpModel "github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
-	auth "github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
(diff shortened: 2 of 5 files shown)


cmd/state/internal/cmdtree/builds.go
```diff
@@ -5,11 +5,12 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runners/builds"
+	"github.com/ActiveState/cli/pkg/project"
 )
 
 func newBuildsCommand(prime *primer.Values) *captain.Command {
 	runner := builds.New(prime)
-	params := &builds.Params{}
+	params := &builds.Params{Namespace: &project.Namespaced{}}
```
Naatan, line 13:
```suggestion
	params := &builds.Params{}
```

This feels like an anti-pattern. Either it's nil or it's a real namespace, not a namespace struct with no values.

cmd/state/internal/cmdtree/builds.go
```diff
@@ -19,9 +20,19 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
...
 			{
 				Name:        "all",
-				Description: "List all builds, including individual package artifacts",
+				Description: locale.Tl("builds_flags_all_description", "List all builds, including individual package artifacts"),
 				Value:       &params.All,
 			},
+			{
+				Name:        "namespace",
+				Description: locale.Tl("builds_flags_namespace_description", "The namespace of the project to inspect builds for"),
+				Value:       params.Namespace,
```
Naatan, line 29:
```suggestion
				Value:       &params.Namespace,
```

I guess this is why line 13 had the empty struct. But you shouldn't need that.

cmd/state/internal/cmdtree/builds.go
```diff
@@ -19,9 +20,19 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
 		[]*captain.Flag{
 			{
 				Name:        "all",
-				Description: "List all builds, including individual package artifacts",
+				Description: locale.Tl("builds_flags_all_description", "List all builds, including individual package artifacts"),
```
Naatan, line 23: 🙏

cmd/state/internal/cmdtree/builds.go
```diff
@@ -42,14 +53,25 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
 
 func newBuildsDownloadCommand(prime *primer.Values) *captain.Command {
 	runner := builds.NewDownload(prime)
-	params := &builds.DownloadParams{}
+	params := &builds.DownloadParams{Namespace: &project.Namespaced{}}
```
Naatan, line 56:
```suggestion
	params := &builds.DownloadParams{}
```

cmd/state/internal/cmdtree/builds.go
```diff
@@ -42,14 +53,25 @@ func newBuildsCommand(prime *primer.Values) *captain.Command {
...
 		"dl",
 		locale.Tl("builds_download_title", "Download build artifacts"),
 		locale.Tl("builds_download_description", "Download build artifacts for a given build"),
 		prime,
-		[]*captain.Flag{},
+		[]*captain.Flag{
+			{
+				Name:        "namespace",
+				Description: locale.Tl("builds_download_flags_namespace_description", "The namespace of the project to download artifacts from"),
+				Value:       params.Namespace,
```
Naatan, line 67:
```suggestion
				Value:       &params.Namespace,
```

internal/runners/builds/builds.go
```diff
@@ -192,3 +187,72 @@ func (b *Builds) outputPlain(out *StructuredOutput, fullID bool) error {
...
+		commitID = *commitUUID
+	}
+
+	// Return the artifact map for the given commitID of the given project.
+	bp := model.NewBuildPlannerModel(auth)
+	buildPlan, err := bp.FetchBuildResult(commitID, namespace.Owner, namespace.Project)
+	if err != nil {
+		return nil, errs.Wrap(err, "Failed to fetch build plan")
+	}
+	return buildplan.NewMapFromBuildPlan(buildPlan.Build, false, false, nil)
```
Naatan, line 257: I see 3 behaviours here, but I think it's hard to grok that from this code. Can we make this a switch case where each case holds the behaviour?

So

- Source from project
- Source from namespace
- Source from commit

You'd end up with some redundancies with regard to fetching the build result, but that's already true, and I don't mind redundancy if it improves readability and thus maintainability.

internal/runners/builds/builds.go
```diff
@@ -192,3 +187,72 @@ func (b *Builds) outputPlain(out *StructuredOutput, fullID bool) error {
 	b.out.Print("\nTo download builds run '[ACTIONABLE]state builds dl <ID>[/RESET]'.")
 	return nil
 }
+
+// getTerminalArtifactMap returns a project's terminal artifact map, depending on the given
+// arguments. By default, the map for the current project is returned, but a map for a given
+// commitID for the current project can be returned, as can the map for a remote project
+// (and optional commitID).
+func getTerminalArtifactMap(
```
Naatan, line 195: Let's add a spinner to this function

https://github.com/ActiveState/cli/blob/8bb4da4efcc6dab1b8bff05ffa69ae64162befd9/internal/output/progress.go#L28-L28

Since this often seems to take a few seconds, or even minutes if the build is complex, and otherwise it just hangs.

Naatan (changes requested):

cmd/state/internal/cmdtree/builds.go
mitchell-as: Yep, that's exactly what I did first. However, if you supply a namespace, State Tool errors out due to https://github.com/ActiveState/cli/blob/cf6492853a1dfc5b10112b11ac192e3bdf273822/pkg/project/namespace.go#L45-L47

As a result, we need this. We do it for other commands too.

Naatan: Ahh I didn't realize we were doing this. Still something we should stop doing but fair enough; not in scope for this story.

internal/runners/builds/rationalize.go
```diff
@@ -9,10 +9,17 @@ import (
...
 	switch {
 	case errors.Is(*err, rationalize.ErrNoProject):
 		*err = errs.WrapUserFacing(*err,
 			locale.Tr("err_no_project"),
 			errs.SetInput())
+
+	case errors.As(*err, &invalidCommitIdErr):
+		*err = errs.WrapUserFacing(
+			*err, locale.Tr("err_commit_id_invalid", invalidCommitIdErr.id),
+			errs.SetInput())
```
Naatan, line 23: Nice, glad you caught it cause I didn't 😅

Naatan (approved):
Looks good!

PR #3104, opened 2024-02-12
`state init` uses defaults for namespace

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2494" title="DX-2494" target="_blank">StoryDX-2494</a>  state init has reasonable defaults for org/project and does not require them
  </td></table>
  <br />


internal/runners/initialize/init.go
```diff
@@ -161,23 +186,12 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
...
+	resolvedOwner, err = r.getOwner(paramOwner)
 	if err != nil {
-		return errs.Wrap(err, "Unable to get the user's writable orgs")
-	}
-	for _, org := range orgs {
-		if strings.EqualFold(org.URLname, params.Namespace.Owner) {
-			owner = org.URLname
-			break
-		}
+		return locale.WrapError(err, "err_init_owner", "Could not determine owner")
```
mitchell-as, line 191: Please use `errs.Wrap()`. The rationalizer should be handling this and reporting a user-facing error, and I think it does for `errNoOwner` at least.

pkg/projectfile/projectfile.go
```diff
@@ -1446,3 +1447,10 @@ func CleanProjectMapping(cfg ConfigGetter) {
 		logging.Debug("Could not clean project mapping in config, error: %v", err)
 	}
 }
+
+func SetRecentlyUsedNamespace(cfg ConfigGetter, namespace string) {
+	err := cfg.Set(constants.LastUsedNamespacePrefname, namespace)
+	if err != nil {
+		logging.Debug("Could not set recently used namespace in config, error: %v", err)
```
mitchell-as, line 1454: I think we normally use `multilog.Error()` for these sorts of things, but I don't know if that's needed here.

internal/constants/constants.go
```diff
@@ -429,6 +429,9 @@ const ForumsURL = "https://community.activestate.com/c/state-tool/"
 // GlobalDefaultPrefname is the pref that holds the path to the globally defaulted project
 const GlobalDefaultPrefname = "projects.active.path"
 
+// LastUsedNamespacePrefname is the pref that holds the last used org for `state init`
```
mitchell-as, line 432: According to the comment chain in the JIRA ticket, "last used" is not just `state init`, but for any read/write operation using the State Tool. I don't know if that's covered in this PR, but this comment is not accurate.

internal/runners/initialize/init.go
```diff
@@ -97,8 +98,32 @@ func inferLanguage(config projectfile.ConfigGetter) (string, string, bool) {
...
+			// If the namespace was invalid but an argument was passed, we
+			// assume it's a project name and not an owner.
+			logging.Error("Could not parse namespace: %v", err)
+			paramProjectName = params.Namespace
+		} else {
+			paramOwner = ns.Owner
+			paramProjectName = ns.Project
+
+		}
+	}
```
mitchell-as, line 122: I feel like cmdtree/init.go's `Run()` function should be interpreting parsed parameters and correctly filling in `RunParams` so that the runner doesn't have to do this kind of interpretation.

internal/runners/initialize/init.go
```diff
@@ -161,23 +186,12 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 		}
 	}
 
-	// Match the case of the organization.
-	// Otherwise the incorrect case will be written to the project file.
```
mitchell-as, line 165: Can we please keep this comment so that we don't lose the context? Otherwise `getOwner()` looks a bit convoluted with the loops.

mitchell-as (changes requested):
This looks mostly complete, except for the one AC where we let the user know we've done some extra work if they didn't specify a namespace:

```
We’ve created ‘[ACTIONABLE]org/name[/RESET]’ to track changes for this environment. You can change the name at any point by running ‘[ACTIONABLE]state projects edit --name[/RESET]’.
```

pkg/projectfile/projectfile.go
MDrakos: Opted to not use it here as we are now allowing this argument to be either a valid namespace or just a project name.

internal/constants/constants.go
MDrakos: The value is being set as part of the `StoreProjectMapping` function which is called every time a project file is parsed. You can see in the integration test that I checkout a project first to set the last-used value.

internal/runners/initialize/init.go
```diff
@@ -321,11 +321,14 @@ func (i *Initialize) getOwner(desiredOwner string) (string, error) {
 	if err != nil {
 		return "", errs.Wrap(err, "Unable to get the user's writable orgs")
 	}
+	fmt.Println("Desired Owner: ", desiredOwner)
 
 	// Prefer the desired owner if it's valid
 	if desiredOwner != "" {
 		for _, org := range orgs {
+			fmt.Println("Org URL Name: ", org.URLname)
```
mitchell-as, line 329: More debug stuff.

test/integration/init_int_test.go
```diff
@@ -201,7 +202,7 @@ func (suite *InitIntegrationTestSuite) TestInit_InferredOrg() {
 	ts.IgnoreLogErrors()
 
 	org := "ActiveState-CLI"
-	projectName := "test-project"
+	projectName := fmt.Sprintf("test-project-%s", model.HostPlatform)
```
mitchell-as, line 205: Good catch!

mitchell-as (changes requested):
Almost there :)

internal/runners/initialize/init.go
```diff
@@ -97,8 +100,24 @@ func inferLanguage(config projectfile.ConfigGetter) (string, string, bool) {
...
+		paramOwner          string
+		paramProjectName    string
+		resolvedOwner       string
+		resolvedProjectName string
+	)
+	if params.ParsedNS.IsValid() {
+		paramOwner = params.ParsedNS.Owner
+		paramProjectName = params.ParsedNS.Project
+	} else {
+		paramProjectName = params.Namespace
```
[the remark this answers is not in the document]
MDrakos: It already is, I just didn't update the variable assignment here.

internal/runners/initialize/init.go
```diff
@@ -300,3 +313,56 @@ func deriveVersion(lang language.Language, version string) (string, error) {
...
 }
+
+func (i *Initialize) getOwner(desiredOwner string) (string, error) {
+	// Match the case of the organization.
+	// Otherwise the incorrect case will be written to the project file.
+	orgs, err := model.FetchOrganizations()
+	if err != nil {
+		return "", errs.Wrap(err, "Unable to get the user's writable orgs")
+	}
+	fmt.Println("Desired Owner: ", desiredOwner)
```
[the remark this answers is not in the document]
MDrakos: Bah

.github/workflows/build.yml
```diff
@@ -1,7 +1,7 @@
 name: Build-Test-Deploy
 
 # === Triggers ===
-"on":
+'on':
```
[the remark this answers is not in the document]
MDrakos: british-cop-screaming-screaming

internal/constants/constants.go
MDrakos: That's fair. Updated the comment to hopefully be more clear.

mitchell-as (approved):
Well done!

PR #3105, opened 2024-02-13
Prefer latest commit time over platform timestamp when installing packages.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2569" title="DX-2569" target="_blank">StoryDX-2569</a>  `state install --ts` is not preserved 
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/install.go
+++ b/internal/runners/packages/install.go
@@ -1,8 +1,6 @@
 package packages
 
 import (
-	"time"
-
 	"github.com/ActiveState/cli/internal/captain"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/logging"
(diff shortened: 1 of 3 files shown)


mitchell-as:
Test failures are not due to this PR. They are due to outdated an JSON format that is fixed in v44. (See #3106)

internal/runners/packages/install.go
```diff
@@ -39,13 +43,35 @@ func (a *Install) Run(params InstallRunParams, nsType model.NamespaceType) (rerr
...
+		if err != nil {
+			return errs.Wrap(err, "Unable to fetch latest Platform timestamp")
+		}
+
+		commitID, err := commitmediator.Get(a.prime.Project())
+		if err != nil {
+			return errs.Wrap(err, "Unable to get commit ID")
+		}
+
+		_, atTime, err := model.FetchCheckpointForCommit(commitID)
```
MDrakos, line 58: Nit: a model function like `FetchTimeStampForCommit` might be nice. It could return the `time.Time` value of the timestamp for us.

MDrakos (dismissed):

MDrakos (approved):

PR #3109, opened 2024-02-14
Rationalize project not found errors in `state builds --namespace`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2513" title="DX-2513" target="_blank">StoryDX-2513</a>  `state builds` and subcommands work with remote projects
  </td></table>
  <br />

### DIFF
--- a/internal/runners/builds/builds.go
+++ b/internal/runners/builds/builds.go
@@ -85,8 +85,17 @@ type errInvalidCommitId struct {
 	id string
 }
 
+func rationalizeBuildsError(err *error, auth *authentication.Auth) {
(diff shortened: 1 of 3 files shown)


MDrakos (changes requested):
The addition to `rationalizeCommonError` makes sense here however the individual rationalize functions where added specifically to police/enforce the architecture so I think they should be put back. See context here: https://github.com/ActiveState/cli/pull/3072#discussion_r1480246184

MDrakos (approved):

PR #3111, opened 2024-02-15
Log build expression in buildplaner requests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2483" title="DX-2483" target="_blank">TaskDX-2483</a>  We have an environment variable to expose the build expression
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/buildplanner.go
+++ b/pkg/platform/model/buildplanner.go
@@ -49,6 +49,10 @@ func (c *client) Run(req gqlclient.Request, resp interface{}) error {
 }
 
 func logRequestVariables(req gqlclient.Request) {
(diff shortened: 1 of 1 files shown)


mitchell-as (changes requested):
Forgive me if I'm missing something obvious, but wouldn't this always log stuff instead of just when an environment variable is enabled?

MDrakos:
> Forgive me if I'm missing something obvious, but wouldn't this always log stuff instead of just when an environment variable is enabled?

🤦‍♂️  I had this logic in an earlier iteration but for some reason removed it. Thanks for catching that.

mitchell-as (approved):

PR #3115, opened 2024-02-20
Fixed artifact caching despite previously cached artifacts not existing.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2545" title="DX-2545" target="_blank">BugDX-2545</a>  state tool should install from cache locally whenever possible, especially for init
  </td></table>
  <br />


Naatan:
@mitchell-as this fix seems to be around cleaning up, was this process somehow interrupting cache from being created / accessed?

mitchell-as:
@Naatan If the cleanup process failed, the just-downloaded artifact was never properly stored. Hence it was downloaded again as reported in the ticket.

mitchell-as:
@MDrakos ping

MDrakos (approved):

PR #3117, opened 2024-02-20
Show `state publish` API errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2563" title="DX-2563" target="_blank">BugDX-2563</a>  state publish not showing helpful error messages
  </td></table>
  <br />
 

This gives more context around "Could not publish ingredient".

Sample:

```
% build/state publish ~/tmp/hello-world5-1.0.1.zip --namespace <namespace> -n
█ Publish Ingredient (Beta)

Publishing ingredient...
 x Could not publish ingredient.
  
  Error received: inventory.add_namespace: namespace needs to have 'org/' 
prefix.

█ Need More Help?
 • Run → 'state publish --help' for general help.
```

### DIFF
--- a/internal/runners/publish/publish.go
+++ b/internal/runners/publish/publish.go
@@ -247,7 +247,7 @@ func (r *Runner) Run(params *Params) error {
 	result := graphModel.PublishResult{}
 
 	if err := r.client.Run(pr, &result); err != nil {
-		return locale.WrapError(err, "err_uploadingredient_publish", "Could not publish ingredient")
+		return locale.WrapInputError(err, "err_uploadingredient_publish", "", err.Error())
 	}
 
 	if result.Publish.Error != "" {
--- a/internal/runners/publish/publish.go
+++ b/internal/runners/publish/publish.go
@@ -247,7 +247,7 @@ func (r *Runner) Run(params *Params) error {
 	result := graphModel.PublishResult{}
 
 	if err := r.client.Run(pr, &result); err != nil {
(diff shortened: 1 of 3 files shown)


mitchell-as:
Note: at this time we only have a graphql error message to go off of. We don't have any swagger bindings to work with to introspect the error.

internal/runners/publish/publish.go
```diff
@@ -247,7 +247,7 @@ func (r *Runner) Run(params *Params) error {
 	result := graphModel.PublishResult{}
 
 	if err := r.client.Run(pr, &result); err != nil {
-		return locale.WrapError(err, "err_uploadingredient_publish", "Could not publish ingredient")
+		return locale.WrapInputError(err, "err_uploadingredient_publish", "", err.Error())
```
Naatan, line 250: We can't just assume all publish errors are input errors. But we should be able to identify this from the http status code returned.

Naatan (changes requested):

internal/runners/publish/publish.go
mitchell-as: That's fair. I tried manipulating our graphql query, but discovered that the `publish` endpoint is not returning an expected error type. It's just returning a list of errors. I've filed https://activestatef.atlassian.net/jira/software/c/projects/PB/issues/PB-4889 and https://activestatef.atlassian.net/browse/DX-2592 to address this. Until then, at least we'll have a more informative error message to present the user with.

Naatan (approved):

PR #3119, opened 2024-02-20
Support the revision flag for `state install`

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2589" title="DX-2589" target="_blank">StoryDX-2589</a>  As a user I can specify a revision when installing packages
  </td></table>
  <br />

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -610,6 +610,8 @@ package_search_flag_ns_description:
   other: The namespace to search within. This trumps the language flag.
 package_flag_ts_description:
   other: The timestamp at which you want to query. Can be either 'now' or RFC3339 formatted timestamp.
+package_flag_rev_description:
+  other: The revision you want to use, this ensures you get this exact revision and nothing else.
 package_search_flag_language_description:
   other: The language used to constrain search results
 package_search_flag_exact-term_description:
@@ -1493,5 +1495,6 @@ uploadingredient_success:
     Name: [ACTIONABLE]{{.V0}}[/RESET]
     Namespace: [ACTIONABLE]{{.V1}}[/RESET]
     Version: [ACTIONABLE]{{.V2}}[/RESET]
+    Revision: [ACTIONABLE]{{.V3}}[/RESET]
 
-    You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --ts now`[/RESET].
+    You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --r {{.V3}} --ts now`[/RESET].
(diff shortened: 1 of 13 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1493,5 +1495,6 @@ uploadingredient_success:
     Name: [ACTIONABLE]{{.V0}}[/RESET]
     Namespace: [ACTIONABLE]{{.V1}}[/RESET]
     Version: [ACTIONABLE]{{.V2}}[/RESET]
+    Revision: [ACTIONABLE]{{.V3}}[/RESET]
 
-    You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --ts now`[/RESET].
+    You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --r {{.V3}} --ts now`[/RESET].
```
MDrakos, line 1500:
```suggestion
    You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --r {{.V2}} --ts now`[/RESET].
```

It should be V2 right?

MDrakos (changes requested):
This all looks good to me. Just requesting changes as the integration tests haven't run and we should do that before merging.

internal/locale/locales/en-us.yaml
Naatan: No, V2 holds the version, which is used above this line (it's a multiline locale).

Naatan:
> This all looks good to me. Just requesting changes as the integration tests haven't run and we should do that before merging.

The run is over here: https://github.com/ActiveState/cli/actions/runs/7980174137?pr=3119

Github seems bugged in its reporting.. :\

MDrakos:
> > This all looks good to me. Just requesting changes as the integration tests haven't run and we should do that before merging.
> 
> The run is over here: https://github.com/ActiveState/cli/actions/runs/7980174137?pr=3119
> 
> Github seems bugged in its reporting.. :\

That Link is just the `Target & Verify` job. But I will leave it up to you.

MDrakos (approved):

Naatan:
> > > This all looks good to me. Just requesting changes as the integration tests haven't run and we should do that before merging.
> > 
> > 
> > The run is over here: [ActiveState/cli/actions/runs/7980174137?pr=3119](https://github.com/ActiveState/cli/actions/runs/7980174137?pr=3119)
> > Github seems bugged in its reporting.. :\
> 
> That Link is just the `Target & Verify` job. But I will leave it up to you.

Woops, copied the wrong one. Fwiw: https://github.com/ActiveState/cli/actions/runs/7980174129?pr=3119

PR #3120, opened 2024-02-20
Use a more informative name for downloaded artifacts if necessary.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2561" title="DX-2561" target="_blank">BugDX-2561</a>  `state builds dl <pkg artifact ID` rewriting the downloaded `artifact.tar.gz`
  </td></table>
  <br />
 

Sample:

```
% build/state builds dl 0EC791E7 
█ Download build artifacts (Beta)

Setting Up Runtime
Resolving Dependencies ✔ Done
✔ All dependencies have been installed and verified

zlib               570.0KB/570.0KB       
Downloaded zlib to /Users/me/cli/zlib-1.2.12.tar.gz
```

### DIFF
--- a/internal/runners/builds/download.go
+++ b/internal/runners/builds/download.go
@@ -116,7 +116,19 @@ func (d *Download) downloadArtifact(artifact *artifact.Artifact, targetDir strin
 		return errs.Wrap(err, "Could not parse artifact URL %s.", artifact.URL)
 	}
 
-	downloadPath := filepath.Join(targetDir, path.Base(artifactURL.Path))
+	// Determine an appropriate basename for the artifact.
+	// Most platform artifact URLs are just "artifact.tar.gz", so use "<name>-<version>.<ext>" format.
+	// Some URLs are more complex like "<name>-<hash>.<ext>", so just leave them alone.
+	basename := path.Base(artifactURL.Path)
+	var ext string
+	if pos := strings.Index(basename, "."); pos != -1 {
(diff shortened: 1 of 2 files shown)


internal/runners/builds/download.go
```diff
@@ -116,7 +116,19 @@ func (d *Download) downloadArtifact(artifact *artifact.Artifact, targetDir strin
 		return errs.Wrap(err, "Could not parse artifact URL %s.", artifact.URL)
 	}
 
-	downloadPath := filepath.Join(targetDir, path.Base(artifactURL.Path))
+	// Determine an appropriate basename for the artifact.
+	// Most platform artifact URLs are just "artifact.tar.gz", so use "<name>-<version>.<ext>" format.
+	// Some URLs are more complex like "<name>-<hash>.<ext>", so just leave them alone.
+	basename := path.Base(artifactURL.Path)
```
MDrakos, line 122: I might be missing something here, so forgive me if I am, but what does the URL have to do with the naming aside from the file extension? We resolved the artifact name by parsing the build plan so we shouldn't care what the URL is.

mitchell-as: Docker images have weird names like "Docker Image (more-descriptive-name-hash).tar.gz". However, the URL name only has "more-descriptive-name-hash.tar.gz", which is what we want.

MDrakos (approved):

PR #3122, opened 2024-02-20
Remove top-level "let" buildscript key and rename "in" key to be an assignment to "main".

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2585" title="DX-2585" target="_blank">StoryDX-2585</a>  As a new user I should not be confused by the use of "let" in buildscripts
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/buildscript/buildscript_test.go
+++ b/internal/runbits/buildscript/buildscript_test.go
@@ -15,22 +15,20 @@ import (
 
 func TestDiff(t *testing.T) {
 	script, err := buildscript.NewScript([]byte(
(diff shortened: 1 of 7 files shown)


MDrakos (approved):
This looks sane to me though it's hard to tell exactly what's going on. If I remember correctly there is a test that translates the script to an expression and back. As long as that is functioning as we expect then I'm good.

mitchell-as:
Yes, what you are describing happens in unit tests, which are passing.

PR #3128, opened 2024-02-21
Fixed premature exit if State Tool could not be downloaded.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2584" title="DX-2584" target="_blank">BugDX-2584</a>  Installer should error out if branch is not valid
  </td></table>
  <br />
 

This prevented an informative error message from being printed.

Sample:

```
% ./installers/install.sh -b oops
Could not download a State Tool installer for the given command line arguments
```

### DIFF
--- a/installers/install.sh
+++ b/installers/install.sh
@@ -128,7 +128,7 @@ else
 fi
 
 # Fetch version info.
-$FETCH $INSTALLERTMPDIR/info.json $JSONURL || exit 1
+$FETCH $INSTALLERTMPDIR/info.json $JSONURL
 if [ ! -z "`grep -o Invalid $INSTALLERTMPDIR/info.json`" ]; then
 	error "Could not download a State Tool installer for the given command line arguments"
 	exit 1
--- a/installers/install.sh
+++ b/installers/install.sh
@@ -129,7 +129,7 @@ fi
 
 # Fetch version info.
 $FETCH $INSTALLERTMPDIR/info.json $JSONURL
-if [ ! -z "`grep -o Invalid $INSTALLERTMPDIR/info.json`" ]; then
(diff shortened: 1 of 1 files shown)


installers/install.sh
```diff
@@ -128,7 +128,7 @@ else
 fi
 
 # Fetch version info.
-$FETCH $INSTALLERTMPDIR/info.json $JSONURL || exit 1
+$FETCH $INSTALLERTMPDIR/info.json $JSONURL
```
Naatan, line 131: We don't appear to use `set -e` for install.sh, so wouldn't this just continue execution if the fetch failed?

Naatan (changes requested):

installers/install.sh
mitchell-as: It would, but the very next line catches the failed download.

Naatan: It looks for the word "Invalid" in `info.json`, which seems like a very specific type of error.

I think we need to:

- Exit if $FETCH had a non-zero exit code
- Ensure the $FETCH error was relayed

It looks like we silence wget and curl, so maybe we can't relay the error without also exposing the user to the verbosity of those commands (worth verifying), but we should probably still at least check for a non-zero exit code.

Naatan (changes requested):

installers/install.sh
mitchell-as: I've added an explicit check for exit status. From my testing, I'm not getting any error output from the $FETCH command, with or without `set -e`.

By the way, the grep test would fail if the file does not exist (which would be the case if the fetch command failed), hence why I didn't feel the need for an explicit exit status check.

Naatan (approved):

PR #3129, opened 2024-02-21
Mark unrecognized language errors as user-facing.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2587" title="DX-2587" target="_blank">BugDX-2587</a>  INIT: CRT error in log for user input 
  </td></table>
  <br />

### DIFF
--- a/internal/runners/initialize/rationalize.go
+++ b/internal/runners/initialize/rationalize.go
@@ -2,9 +2,11 @@ package initialize
 
 import (
 	"errors"
+	"strings"
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/language"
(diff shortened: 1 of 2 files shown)


internal/runners/initialize/rationalize.go
```diff
@@ -45,6 +48,13 @@ func rationalizeError(owner, project string, rerr *error) {
 			errs.SetInput(),
 		)
 
+	case errors.As(*rerr, &unrecognizedLanguageErr):
```
MDrakos, line 51: Couldn't this function handle the error type from the language package?

mitchell-as: I thought the same, until I realized that the language package uses a function call to return a localized error, and not a specific error type.

mitchell-as:
@MDrakos ping

MDrakos (approved):

PR #3130, opened 2024-02-21
Update requirements format

MDrakos:


pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -91,6 +91,11 @@ func indent(s string) string {
 func (s *Script) String() string {
 	buf := strings.Builder{}
 	for _, assignment := range s.Expr.Let.Assignments {
+		if assignment.Name == "requirements" {
+			buf.WriteString(requirementsString(assignment))
+			buf.WriteString("\n")
+			continue
+		}
 		buf.WriteString(assignmentString(assignment))
 		buf.WriteString("\n")
```
mitchell-as, line 100: What would make me over-the-moon-happy would be to have something like this:

```
if assignment.Name == "requirements" {
  assignment = &Assignment{assignment.Name, transformRequirements(assignment.Value)}
}
buf.WriteString(assignmentString(assignment))
buf.WriteString("\n")
```

where `transformRequirements` iterates over the given `buildexpression.Value` (whose `List` field is non-nil) and returns a new `buildexpression.Value` whose `List` field is constructed of `buildexpression.Ap` values of the name "Req" and whose arguments are `buildexpression.Value`s with `Assignment` fields of the appropriate "name" and "version" buildscript fields. (Note: you will need to add a method to https://github.com/ActiveState/cli/blame/master/pkg/platform/model/version_constraints.go that converts a `[]bpModel.VersionRequirement` to a string using the existing `versionConstraintsToString()` and use that returned value for the "version" field; your existing code does not handle complex version constraints, while the previous buildscript format did.)

By doing so, we should be able to avoid the entire `requirementsString()` function and piggy-back off of the existing `valueString()` and `apString()` functions. (Note: you may need to modify `apString()` to not print newlines if the name is "Req"/constant.)

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -105,7 +110,75 @@ func (s *Script) String() string {
...
+		}
+	}
+
+	return buf.String()
+}
+
 func assignmentString(a *buildexpression.Var) string {
+	if a.Name == "requirements" {
+		return requirementsString(a)
+	}
```
mitchell-as, line 181: I believe this is unnecessary due to prior handling in the `(s *Script) String()` method.

mitchell-as (changes requested):
This is a good start. The primary thing to address here is the ability to convert to/from complex version constraints (e.g. "3.10" means `{"comparator": "eq", "version": "3.10"}` and ">2" means `{"comparator": "gt", "version": "2"}` and vice-versa.) Most of the code to do this exists and should be utilized.

Going back over my review comments, I realize that this is a pretty complex ticket. I'm willing to pick this up if you'd rather move on to something else.

pkg/platform/runtime/buildscript/json.go
```diff
@@ -75,6 +88,48 @@ func (f *FuncCall) MarshalJSON() ([]byte, error) {
...
+					{Object: &[]*Assignment{
+						{comparatorKey, &Value{Str: ptr.To("eq")}},
+						{versionKey, argument.Assignment.Value},
+					}},
+				}}
+			} else {
+				mArgs[argument.Assignment.Key] = argument.Assignment.Value
+			}
+		case argument.FuncCall != nil:
+			mArgs[argument.FuncCall.Name] = argument.FuncCall.Arguments
```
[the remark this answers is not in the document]
MDrakos: Good catch. I had this here when prototyping but it's not longer needed.

pkg/platform/runtime/buildscript/buildscript.go
MDrakos: Was able to get this to work. Thanks for the pointers. It's not the _exact_ syntax you provided but it's close.

pkg/platform/runtime/buildscript/buildscript.go
MDrakos: It is still needed. Confirmed with test coverage from the unit tests. However, it's not `requirementsString` anymore, instead it's replacing the assignment, like you suggested in your comment above.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -105,7 +111,74 @@ func (s *Script) String() string {
 	return buf.String()
 }
 
+func transformRequirements(reqs *buildexpression.Var) *buildexpression.Var {
+	newReqs := &buildexpression.Var{
+		Name: "requirements",
+		Value: &buildexpression.Value{
+			List: &[]*buildexpression.Value{},
+		},
+	}
```
mitchell-as, line 120: This is awesome. It's like a type declaration, but better.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -105,7 +111,74 @@ func (s *Script) String() string {
...
+	return newReqs
+}
+
+func transformReq(req *buildexpression.Value) *buildexpression.Value {
+	newReq := &buildexpression.Value{
+		Ap: &buildexpression.Ap{
+			Name:      reqFuncName,
+			Arguments: []*buildexpression.Value{},
+		},
+	}
```
mitchell-as, line 135: Amazing.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -105,7 +111,74 @@ func (s *Script) String() string {
...
+			}
+
+		case buildexpression.RequirementNamespaceKey:
+			if name != "" {
+				name = fmt.Sprintf("%s/%s", *arg.Value.Str, name)
+			} else {
+				name = *arg.Value.Str
+			}
+		case buildexpression.RequirementVersionRequirementsKey:
+			version = model.BuildExpressionRequirementsToString(arg)
```
mitchell-as, line 154: w00t

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -160,15 +233,38 @@ func valueString(v *buildexpression.Value) string {
 }
 
 func apString(f *buildexpression.Ap) string {
+	if f.Name == reqFuncName {
+		return apReqString(f)
+	}
+
 	buf := bytes.Buffer{}
 	buf.WriteString(fmt.Sprintf("%s(\n", f.Name))
```
mitchell-as, line 241: By keeping spaces, we only need to check for `f.Name == reqFuncName` and the omit the newlines in this method.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -160,15 +233,38 @@ func valueString(v *buildexpression.Value) string {
...
+func apReqString(f *buildexpression.Ap) string {
+	buf := bytes.Buffer{}
+	buf.WriteString(fmt.Sprintf("%s(", f.Name))
+	for i, argument := range f.Arguments {
+		buf.WriteString(strutils.RemoveSpaces(valueString(argument)))
+
+		if i+1 < len(f.Arguments) {
+			buf.WriteString(", ")
+		}
+	}
```
mitchell-as, line 266: As mentioned previously, this is a lot of extra effort to remove spaces. Let's just keep them and simplify the code for now.

pkg/platform/runtime/buildscript/json.go
```diff
@@ -75,6 +87,64 @@ func (f *FuncCall) MarshalJSON() ([]byte, error) {
...
+	for _, argument := range args {
+		switch {
+		case argument.Assignment != nil:
+			if argument.Assignment.Key == buildexpression.RequirementNameKey && argument.Assignment.Value.Str != nil {
+				name, namespace := separateNamespace(*argument.Assignment.Value.Str)
+				mArgs[buildexpression.RequirementNameKey] = name
+				mArgs[buildexpression.RequirementNamespaceKey] = namespace
+			} else if argument.Assignment.Key == buildexpression.RequirementVersionKey && argument.Assignment.Value.Str != nil {
+				value := strings.Trim(*argument.Assignment.Value.Str, `"`)
+				versionReqs, err := model.VersionStringToRequirements(value)
```
mitchell-as, line 106: w00t^2

mitchell-as (changes requested):
I was so over-the-moon happy that I wanted to immediately approve, but there are a couple more minor things to address. Amazing work though :)

pkg/platform/runtime/buildscript/buildscript.go
mitchell-as: That's a bit of a let-down, but thanks for checking!

mitchell-as (approved):
Hammer-stamped. Well done!

PR #3133, opened 2024-02-22
Use generic "value" placeholder for flag values.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2547" title="DX-2547" target="_blank">StoryDX-2547</a>  Revisit flag rendering for `state --help`
  </td></table>
  <br />
 

Sample:

```
% build/state help publish
Publish an Ingredient for private consumption.

Usage:
    state publish [flags] <filepath>

Arguments:
  <filepath>            A tar.gz or zip archive containing the source files of the ingredient.

Flags:
      --author <value>        Ingredient author, in the format of "[<name>] <email>". Can be set multiple times.
      --depend <value>        Ingredient that this ingredient depends on, format as <namespace>/<name>[@<version>]. Can be set multiple times.
      --description <value>   A short description summarizing what this ingredient is for.
      --edit                  Create a revision for an existing ingredient, matched by their name and namespace.
      --editor                Edit the ingredient information in your editor before uploading.
      --feature <value>       Feature that this ingredient provides, format as <namespace>/<name>[@<version>]. Can be set multiple times.
      --meta <value>          A yaml file expressing the ingredient meta information. Use --editor to review the file format.
      --name <value>          The name of the ingredient. Defaults to basename of filepath.
      --namespace <value>     The namespace of the ingredient. Defaults to org/<orgname>. Must start with 'org/<orgname>'.
      --version <value>       Version of the ingredient (preferably semver).

Global Flags:
  -h, --help              Help for this command
  -l, --locale <value>    Set the localisation
      --mono              Force monochrome text output
  -n, --non-interactive   Run the State Tool without any prompts
  -o, --output <value>    Set the output method, possible values: plain, simple, json, editor
  -v, --verbose           Verbose output
```

### DIFF
--- a/internal/assets/contents/usage.tpl
+++ b/internal/assets/contents/usage.tpl
@@ -32,12 +32,12 @@ Arguments:
 {{- if .Cobra.HasAvailableFlags}}
 
 Flags:
-{{.Cobra.LocalFlags.FlagUsages | trimTrailingWhitespaces}}
+{{.Cmd.LocalFlagUsages | trimTrailingWhitespaces}}
 {{- end}}
 {{- if .Cobra.HasAvailableInheritedFlags}}
 
 Global Flags:
-{{.Cobra.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}
+{{.Cmd.InheritedFlagUsages | trimTrailingWhitespaces}}
 {{- end}}
 
 {{- if .Cobra.HasHelpSubCommands}}
--- a/internal/captain/flag.go
+++ b/internal/captain/flag.go
@@ -1,6 +1,7 @@
 package captain
 
 import (
+	"bytes"
 	"fmt"
 	"reflect"
 	"strconv"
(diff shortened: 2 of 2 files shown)


mitchell-as:
I'm not sure how I feel about seeing the following flags next to each other:

```
      --depend value         Ingredient that this ingredient depends on, format as <namespace>/<name>[@<version>]. Can be set multiple times.
      --description string   A short description summarizing what this ingredient is for.
```

Perhaps instead of "value", we use "string", "int", or other basic type depending on context.

Naatan (changes requested):
This feels wrong to me, because the Type() is apparently used also for error reporting and possibly some other scenarios. 

Reading through the code, there is some special handling to identify the type value here:

https://github.com/ActiveState/cli/blob/db3e742d4b7bccf92debe1214c0c41df72e0120b/vendor/github.com/spf13/pflag/flag.go#L570-L570

And indeed, if I change this line:

```
					"Ingredient author, in the format of \"[<name>] <email>\". Can be set multiple times.",
```

To:

```
					"Ingredient author, in the format of `[<name>] <email>`. Can be set multiple times.",
```

The result is:

```
Flags:
      --author [<name>] <email>   Ingredient author, in the format of [<name>] <email>. Can be set multiple times.

```

But it's so implicit. Hard to even audit what commands are affected by this cause cobra decided to be all magical about it.

Honestly, given these limitations Cobra gives us I'd rather we just implement our own version of `FlagUsages`, ie.

https://github.com/ActiveState/cli/blob/db3e742d4b7bccf92debe1214c0c41df72e0120b/vendor/github.com/spf13/pflag/flag.go#L677-L677

We could just copy that function and make the adjustments we need. We're already doing this in so many places for cobra. Frankly we should probably just schedule work to get rid of cobra at this point, but that's a story for another day.

Naatan (approved):

PR #3135, opened 2024-02-22
Fix merge conflict from v43 to v44.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2598" title="DX-2598" target="_blank">BugDX-2598</a>  State scripts defined outside of activestate.yaml no longer work
  </td></table>
  <br />


mitchell-as:
@Naatan This needs to be force-merged into v44 and then up to master since the ticket targets v43.

PR #3138, opened 2024-02-23
Buildscript variable values should be converted to "$var" in build expressions and vice-versa.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2586" title="DX-2586" target="_blank">StoryDX-2586</a>  As a user I can expect variables to be defined consistently
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildscript/buildscript_test.go
+++ b/pkg/platform/runtime/buildscript/buildscript_test.go
@@ -272,6 +272,7 @@ func TestRoundTrip(t *testing.T) {
 func TestJson(t *testing.T) {
 	script, err := NewScript([]byte(
 		`runtime = solve(
+		at_time = at_time,
 		requirements=[
 			Req(name = "language/python")
 		],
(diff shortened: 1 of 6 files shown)


pkg/platform/runtime/buildscript/buildscript_test.go
```diff
@@ -287,10 +288,11 @@ main = runtime
     "let": {
       "runtime": {
         "solve": {
+          "at_time": "$at_time",
           "requirements": [
             {
               "name": "python",
-			  "namespace": "language"
+              "namespace": "language"
```
MDrakos, line 295: I know this isn't related to your PR but shouldn't this be updated to the new `Req` format?

MDrakos (approved):

pkg/platform/runtime/buildscript/buildscript_test.go
mitchell-as: No, this is the result of the translation from buildscript to JSON. The new Req format was on line 276.

PR #3139, opened 2024-02-23
Fix table spacing and `omitEmpty` option

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2578" title="DX-2578" target="_blank">BugDX-2578</a>  Incorrect vertical table spacing with long key names
  </td></table>
  <br />
 

As we discussed in stand, it might be worthwhile to have a separate vertical table representation rather than trying to have a vertical toggle in the standard table representation. We could do this within the current library or look into the `charm` family of libraries to see if we can get something 3rd party to work.


internal/output/plain.go
```diff
@@ -406,7 +408,12 @@ func verticalRows(hdrs []string, rows [][]string) [][]string {
 		for j, hcol := range hrow {
 			var header string
 			if j < len(hdrs) {
-				header = hdrs[j]
+				// Align headers with rows
+				if hcol == "" {
```
MDrakos, line 412: I was getting incorrect headers for certain rows in the table. For example, the `Versions` header would be present for the `Website`. This fixed that misalignment.

internal/output/plain.go
```diff
@@ -266,7 +266,7 @@ func sprintMap(value interface{}) (string, error) {
 
 	sort.Slice(result, func(i, j int) bool { return result[i] < result[j] })
 
-	return "\n" + strings.Join(result, "\n"), nil
+	return strings.Join(result, "\n"), nil
```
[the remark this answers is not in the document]
MDrakos: Okay, added back.

internal/output/plain.go
MDrakos: Yeah, it is. Went with a different approach in this recent set of changes.

MDrakos:
I expect we may want to iterate on this but I wanted to try a different approach where we bind the header to the row content. This fixes the issue where the headers and rows could get misaligned. The new function for printing the vertical table does have some repeated logic but I think this starts to take us in the direction we ultimately want to go where we have a separate path for vertical tables rather than trying to have our existing tables with a vertical option.

internal/output/plain.go
```diff
@@ -299,14 +298,14 @@ func sprintTable(vertical bool, slice []interface{}) (string, error) {
...
-				continue
-			}
-
 			if firstIteration && !funk.Contains(field.opts, string(SeparateLineOpt)) {
 				headers = append(headers, localizedField(field.l10n))
 			}
 
+			if funk.Contains(field.opts, string(OmitEmpty)) && (stringValue == "" || stringValue == nilText) {
+				stringValue = ""
+			}
```
Naatan, line 307: This feels inappropriate. OmitEmpty means "don't print if it's empty". This is doing the exact opposite.

The old implementation is definitely awkward, because if we don't omit the column for each row you're going to run into some weird results. But then further down we also have `shiftColsVal` which allows for columns to span based on the `shiftCols=` param. Point is, this may all be by design, and this change you're making is backward incompatible. So either we need to audit existing use of this code and ensure it is compatible with your change, or we don't make any backward incompatible changes.

Naatan: Note this is for the horizontal table mechanic, I suspect you can just revert this back to what it looked like before this PR?

internal/output/plain.go
```diff
@@ -322,22 +321,70 @@ func sprintTable(vertical bool, slice []interface{}) (string, error) {
...
+
+			stringValue, err := sprint(field.value)
+			if err != nil {
+				return "", err
+			}
+
+			if funk.Contains(field.opts, string(OmitEmpty)) && (stringValue == "" || stringValue == nilText) {
+				stringValue = ""
+			}
```
Naatan, line 367: Same as with the other one; this isn't omitting anything.

internal/table/table.go
```diff
@@ -103,6 +104,11 @@ func (t *Table) calculateWidth(maxTableWidth int) ([]int, int) {
 		colWidthsCombined += colWidths[n]
 	}
 
+	var verticalHeaderWidth int
+	if len(colWidths) > 0 && t.Vertical {
+		verticalHeaderWidth = colWidths[0]
+	}
```
Naatan, line 110: This is only used on `rescaleColumns` and is based on inputs that are also passed to that function. So I suggest moving this logic inside `rescaleColumns`.

But also; what is the rationale here? Comment explaining this would be good.

internal/table/table.go
```diff
@@ -138,18 +144,29 @@ func equalizeWidths(colWidths []int, percentage int) {
...
 		colWidths[n] = int(float64(colWidths[n]) * multiplier)
 	}
 
 	// Account for floats that got rounded
 	if len(colWidths) > 0 {
 		colWidths[len(colWidths)-1] += targetTotal - mathutils.Total(colWidths...)
 	}
+
+	// If vertical, respect the header width
+	if vertical && len(colWidths) > 0 && colWidths[0] < verticalHeaderWidth {
```
Naatan, line 163: It looks like `verticalHeaderWidth` is either the value of `colWidths[0]` or just 0. So either this conditional doesn't trigger, or it triggers if `colWidths[0] < 0`, which feels inappropriate.

Naatan (changes requested):

internal/output/plain.go
MDrakos: Whoops, that's leftover from my initial implementation. It's been reverted now.

internal/table/table.go
MDrakos: Added a comment, but left this here. There is a call further down to an `equalizeWidths` function that changes the column widths. Before that call the value of the first column is the original header width, which we want to respect when rescaling. I think it makes sense to preserve this, equalize all of the columns, and then ensure the header width is respected when rescaling, but I'm open to move this somewhere else or take a different approach.

internal/table/table.go
MDrakos: As in my other comment, `verticalHeaderWidth` is the value before equalization so it can differ from `colWidths[0]`. I'll add a comment here as well.

internal/table/table.go
Naatan: Ok. I think you missed the first part of my comment though:

> This is only used on rescaleColumns and is based on inputs that are also passed to that function. So I suggest moving this logic inside rescaleColumns.

Naatan (changes requested):
Looks good, but still think we can move this code:

https://github.com/ActiveState/cli/pull/3139#discussion_r1531151600

internal/table/table.go
MDrakos: Ah sorry, I didn't explain well exactly what's happening here. 

The call to `equalizeWidths` changes the actual values in the `colWidths` slice. So we have to grab the column width of the first entry in the slice at this point as this is the max width of the headers column. So `verticalHeaderWidth` can be a different value from `colWidths[0]` after the `equalizeWidths` call.

Naatan (approved):

PR #3143, opened 2024-02-26
Mark `state cve` project not found errors as input errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2579" title="DX-2579" target="_blank">BugDX-2579</a>  Rollbar: security []: Returning error: execute failed:     Failed to retrieve vulnerability information:         Failed to fetch vulnerability information for project invalid/invalid:
  </td></table>
  <br />
 

Also mark unauthenticated errors as input errors.

### DIFF
--- a/pkg/platform/model/cve.go
+++ b/pkg/platform/model/cve.go
@@ -22,7 +22,7 @@ func FetchProjectVulnerabilities(auth *authentication.Auth, org, project string)
 	// This should be removed by https://www.pivotaltracker.com/story/show/176508740
 	if !auth.Authenticated() {
 		return nil, errs.AddTips(
-			locale.NewInputError("cve_needs_authentication"),
+			locale.NewError("cve_needs_authentication"),
 			locale.T("auth_tip"),
 		)
 	}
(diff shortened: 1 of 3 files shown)


pkg/platform/model/cve.go
```diff
@@ -22,7 +22,7 @@ func FetchProjectVulnerabilities(auth *authentication.Auth, org, project string)
 	// This should be removed by https://www.pivotaltracker.com/story/show/176508740
 	if !auth.Authenticated() {
 		return nil, errs.AddTips(
-			locale.NewError("cve_needs_authentication", "You need to be authenticated in order to access vulnerability information about your project."),
+			locale.NewInputError("cve_needs_authentication"),
```
Naatan, line 25: It feels wrong to me to mark this as an input error in the model. This assumes that we invoked the model with user input, but that may not always be the case.

If this is within scope of the rollbar report I'd suggest communicating a type here and asserting it from the runner to then return an input error.

Naatan (changes requested):

pkg/platform/model/cve.go
mitchell-as: You're right about that. Also, the cve runner actually errors out if you're not authenticated to begin with, so the changes around lack of authentication were superfluous. I've changed the model to return an `ErrProjectNotFound` error and check for it in the runner in order to raise an input error though. Thanks for the suggestion.

Naatan (dismissed):

mitchell-as:
Requesting re-review for a failed test 🤦

PR #3144, opened 2024-02-26
Use functions for buildscript version requirements instead of formatted strings.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2601" title="DX-2601" target="_blank">StoryDX-2601</a>  As a user I can expect version requirements to be fully typed
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildscript/buildscript.go
+++ b/pkg/platform/runtime/buildscript/buildscript.go
@@ -11,7 +11,6 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
-	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
 	"github.com/alecthomas/participle/v2"
 )
@@ -56,6 +55,17 @@ type In struct {
 	FuncCall *FuncCall `parser:"@@"`
 	Name     *string   `parser:"| @Ident"`
 }
+
+var (
+	reqFuncName = "Req"
+	eqFuncName  = "Eq"
+	neFuncName  = "Ne"
+	gtFuncName  = "Gt"
+	gteFuncName = "Gte"
+	ltFuncName  = "Lt"
+	lteFuncName = "Lte"
+	andFuncName = "And"
+)
 
 func NewScript(data []byte) (*Script, error) {
 	parser, err := participle.Build[Script]()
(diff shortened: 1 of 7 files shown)


pkg/platform/runtime/buildscript/json.go
```diff
@@ -90,48 +83,74 @@ func (f *FuncCall) MarshalJSON() ([]byte, error) {
...
+					fallthrough
+				case neFuncName:
+					fallthrough
+				case gtFuncName:
+					fallthrough
+				case gteFuncName:
+					fallthrough
+				case ltFuncName:
+					fallthrough
+				case lteFuncName:
```
MDrakos, line 119: Couldn't we list all of these values for a single case rather than use `fallthrough` here? So it would be something like:

```
case eqFuncName, neFuncName, gtFuncName, etc:
   ...
```

pkg/platform/runtime/buildscript/json.go
```diff
@@ -90,48 +83,74 @@ func (f *FuncCall) MarshalJSON() ([]byte, error) {
...
+					req = append(req, &Assignment{buildexpression.RequirementVersionKey, &Value{Str: funcCall.Arguments[0].Str}})
+					requirements = append(requirements, &Value{Object: &req})
+				case andFuncName:
+					for _, a := range funcCall.Arguments {
+						if a.FuncCall == nil {
+							return errors.New(fmt.Sprintf("Illegal argument for version comparator '%s': function expected", name))
+						}
+						err := addRequirement(a.FuncCall)
+						if err != nil {
+							return err
```
MDrakos, line 134: There are a couple errors in this function that need to be wrapped. May be worth skimming the changeset again to ensure we're wrapping everything.

pkg/platform/runtime/buildscript/json.go
```diff
@@ -90,48 +83,74 @@ func (f *FuncCall) MarshalJSON() ([]byte, error) {
...
+		case assignment.Key == buildexpression.RequirementNameKey && assignment.Value.Str != nil:
+			name, namespace := separateNamespace(*assignment.Value.Str)
+			requirement[buildexpression.RequirementNameKey] = name
+			requirement[buildexpression.RequirementNamespaceKey] = namespace
+
+		// Marshal the version argument (e.g. version = <op>("<version>")) into
+		// {"version_requirements": [{"comparator": "<op>", "version": "<version>"}]}
+		case assignment.Key == buildexpression.RequirementVersionKey && assignment.Value.FuncCall != nil:
+			var requirements []*Value
+			var addRequirement func(*FuncCall) error // recursive function for adding to requirements list
```
MDrakos, line 106: This is a clever solution!

pkg/platform/runtime/buildscript/buildscript_test.go
```diff
@@ -166,7 +169,8 @@ const example = `runtime = solve(
 	platforms = ["96b7e6f2-bebf-564c-bc1c-f04482398f38", "96b7e6f2-bebf-564c-bc1c-f04482398f38"],
 	requirements = [
 		Req(name = "language/python"),
-		Req(name = "language/python/requests", version = "3.10.10")
+		Req(name = "language/python/requests", version = Eq("3.10.10")),
+		Req(name = "language/python/argparse", version = And(Gt("1.0"), Lt("2.0")))
```
MDrakos, line 173: Thank you for adding this to the test.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -241,7 +294,11 @@ func apString(f *buildexpression.Ap) string {
 		indent  = indent
 	)
 
-	if f.Name == reqFuncName {
+	if f.Name == reqFuncName ||
+		f.Name == eqFuncName || f.Name == neFuncName ||
+		f.Name == gtFuncName || f.Name == gteFuncName ||
+		f.Name == ltFuncName || f.Name == lteFuncName ||
+		f.Name == andFuncName {
```
MDrakos, line 301: nit: It might be nice to create a map, slice, or function that can encapsulate this for us. So this line becomes a simple map check, `funk.Contains` or function call.

MDrakos (changes requested):
The logic here all looks good to me. Just some minor stuff to clean up.

MDrakos (approved):
Great work!

PR #3145, opened 2024-02-29
Bump default Ruby version to 3.3.0.

wtchappell:


mitchell-as:
Closing in favor of https://github.com/ActiveState/TheHomeRepot/pull/13428 because that's where we keep track of recommended versions. The file you modified is a fallback that will not work on Linux or Windows. It only works with macOS because it wasn't considered to be supported on macOS, but the linked PR fixes that.

PR #3146, opened 2024-03-04
Fix download tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2597" title="DX-2597" target="_blank">BugDX-2597</a>  Nightly failure: TestBuildsIntegrationTestSuite/TestBuilds_Download*
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -3,7 +3,6 @@ package runtime
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -91,7 +90,7 @@ func New(target setup.Targeter, an analytics.Dispatcher, svcm *model.SvcModel, a
 	logging.Debug("Initializing runtime for: %s/%s@%s", target.Owner(), target.Name(), target.CommitUUID())
 
 	if strings.ToLower(os.Getenv(constants.DisableRuntime)) == "true" {
-		fmt.Fprintln(os.Stderr, locale.T("notice_runtime_disabled"))
+		out.Notice(locale.T("notice_runtime_disabled"))
 		return &Runtime{disabled: true, target: target, analytics: an}, nil
 	}
 	recordAttempt(an, target)
--- a/test/integration/builds_int_test.go
+++ b/test/integration/builds_int_test.go
@@ -174,8 +174,8 @@ func (suite *BuildsIntegrationTestSuite) TestBuilds_Download() {
 		buildID = "dbf05bf8-4b2e-5560-a329-b5b70bc7b0fa"
 	} else {
 		buildID = suite.extractBuildID(ts, "bzip2@1.0.8", "")
-		suite.Require().NotEmpty(buildID)
 	}
+	suite.Require().NotEmpty(buildID)
 
 	cp = ts.SpawnWithOpts(
 		e2e.OptArgs("builds", "dl", buildID, "."),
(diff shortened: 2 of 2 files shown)


test/integration/builds_int_test.go
```diff
@@ -150,8 +167,18 @@ func (suite *BuildsIntegrationTestSuite) TestBuilds_Download() {
 	)
 	cp.ExpectExitCode(0, e2e.RuntimeSourcingTimeoutOpt)
 
+	var buildID string
+	if runtime.GOOS == "windows" {
+		// On Windows we need the specific build ID as the terminal buffer is not
+		// large enough to display all the builds
+		buildID = "dbf05bf8-4b2e-5560-a329-b5b70bc7b0fa"
```
MDrakos, line 174: Unfortunate work around that means we may have to periodically update this test. If there is a way to do this that I'm not seeing then please let me know. Otherwise, I will file a follow up story to increase the output buffer on Windows.

pkg/platform/runtime/runtime.go
```diff
@@ -91,7 +90,7 @@ func New(target setup.Targeter, an analytics.Dispatcher, svcm *model.SvcModel, a
 	logging.Debug("Initializing runtime for: %s/%s@%s", target.Owner(), target.Name(), target.CommitUUID())
 
 	if strings.ToLower(os.Getenv(constants.DisableRuntime)) == "true" {
-		fmt.Fprintln(os.Stderr, locale.T("notice_runtime_disabled"))
+		out.Notice(locale.T("notice_runtime_disabled"))
```
mitchell-as, line 93: Good catch!

test/integration/builds_int_test.go
```diff
@@ -164,13 +191,63 @@ func (suite *BuildsIntegrationTestSuite) TestBuilds_Download_Remote() {
...
 
-	cp := ts.Spawn("builds", "dl", "a46a74e9", ".", "--namespace", "ActiveState-CLI/Python-With-Custom-Builds")
+	var buildID string
+	if runtime.GOOS == "windows" {
+		// On Windows we need the specific build ID as the terminal buffer is not
+		// large enough to display all the builds
+		buildID = "dbf05bf8-4b2e-5560-a329-b5b70bc7b0fa"
+	} else {
+		buildID = suite.extractBuildID(ts, "bzip2@1.0.8", "ActiveState-CLI/Python-With-Custom-Builds")
+		suite.Require().NotEmpty(buildID)
```
mitchell-as, line 201: Should this be outside the conditional to assert for Windows too?

test/integration/builds_int_test.go
mitchell-as: Can we `cp.Expect()` "bzip2@1.0.8", `cp.ExpectRe()` a UUID, extract that UUID from the output, and then use it? Something like https://github.com/ActiveState/cli/blame/master/cmd/state-svc/test/integration/svc_int_test.go#L56-L67.

I feel like marshaling stdout to JSON is brittle. The output format could change, there could be sporadic I/O issues on CI, etc.

mitchell-as (changes requested):

test/integration/builds_int_test.go
MDrakos: I prefer the JSON approach here as this test will no fail if the output format changes, which is something we should be looking out for. The sporadic I/O issues would also affect the regex approach.

mitchell-as (approved):

PR #3149, opened 2024-03-06
Support top-level "at_time" variable in buildscripts.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2539" title="DX-2539" target="_blank">TaskDX-2539</a>  at_time is embedded in buildscript through var
  </td></table>
  <br />
 

The idea behind this PR is:

1. Change `buildplanner.GetBuildExpression()` to substitute "$at_time" with the commit's `atTime`. This applies to all build expressions, not just those used for buildscripts.
2. The buildscript will preprocess a buildexpression, extract the timestamp from step 1, replace it with an `at_time` variable, and define this variable at the top of the buildscript.
3. When translating the buildscript back to a buildexpression, the `at_time` variable will be replaced with the timestamp in the top-level assignment. (If there is no top-level assignment, the `at_time` variable will fall back to `"$at_time"`, which the Platform interprets as its default timestamp.)

### DIFF
--- a/pkg/platform/runtime/buildexpression/buildexpression.go
+++ b/pkg/platform/runtime/buildexpression/buildexpression.go
@@ -930,9 +930,9 @@ func (e *BuildExpression) MaybeUpdateTimestamp(ts strfmt.DateTime) error {
 
 	// Normalize existing timestamp.
 	// Platform timestamps may differ from the strfmt.DateTime format. For example, Platform
-	// timestamps will have 6-digit millisecond precision, while strfmt.DateTime will only have
-	// three-digit precision. This will affect comparisons between buildexpressions (which is
-	// normally done byte-by-byte).
+	// timestamps will have microsecond precision, while strfmt.DateTime will only have millisecond
+	// precision. This will affect comparisons between buildexpressions (which is normally done
+	// byte-by-byte).
 	case !strings.HasPrefix(*atTimeNode.Str, "$"):
 		atTime, err := strfmt.ParseDateTime(*atTimeNode.Str)
 		if err != nil {
@@ -942,6 +942,14 @@ func (e *BuildExpression) MaybeUpdateTimestamp(ts strfmt.DateTime) error {
 	}
 
 	return nil
+}
+
+func (e *BuildExpression) Copy() (*BuildExpression, error) {
+	bytes, err := json.Marshal(e)
+	if err != nil {
+		return nil, errs.Wrap(err, "Failed to marshal build expression during copy")
+	}
+	return New(bytes)
 }
 
 func (e *BuildExpression) MarshalJSON() ([]byte, error) {
--- a/pkg/platform/runtime/buildexpression/buildexpression.go
+++ b/pkg/platform/runtime/buildexpression/buildexpression.go
@@ -898,7 +898,7 @@ func (e *BuildExpression) removePlatform(platformID strfmt.UUID) error {
 func (e *BuildExpression) SetDefaultTimestamp() (*strfmt.DateTime, error) {
 	atTimeNode, err := e.getSolveAtTimeValue()
 	if err != nil {
-		return nil, errs.Wrap(err, "Could not get at time node")
+		return nil, errs.Wrap(err, "Could not get %s node", AtTimeKey)
 	}
 	var atTimePtr *strfmt.DateTime
 	if atTimeNode.Str != nil && !strings.HasPrefix(*atTimeNode.Str, "$") {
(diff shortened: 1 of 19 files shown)


pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -119,9 +140,35 @@ func (s *Script) String() string {
...
+// For example, the solve node's "at_time" will not contain a variable reference if this script
+// has a top-level assignment for it.
+func (s *Script) prepareForUse() {
+	if s.atTime != nil {
+		err := s.Expr.MaybeUpdateTimestamp(*s.atTime)
+		if err != nil {
+			multilog.Error("Failed to possibly update at time: %v", err)
+		}
+	}
+}
```
Naatan, line 170: I'm having a really hard time grokking this. Hard for me to suggest how to improve since I'm not quite grasping what this is doing. But hopefully my telling you that this is difficult to understand as written is itself helpful.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -119,9 +140,35 @@ func (s *Script) String() string {
...
+
+// prepareForUse prepares this script's buildexpression for use in a buildplanner context.
+// For example, the solve node's "at_time" will not contain a variable reference if this script
+// has a top-level assignment for it.
+func (s *Script) prepareForUse() {
+	if s.atTime != nil {
+		err := s.Expr.MaybeUpdateTimestamp(*s.atTime)
+		if err != nil {
+			multilog.Error("Failed to possibly update at time: %v", err)
```
Naatan, line 167: We should return errors here. It's ok that this breaks things for the user, that's our signal that something needs fixing.

Naatan (changes requested):

pkg/platform/runtime/buildscript/buildscript.go
mitchell-as: Fair enough. I think at this point we just need to assume the buildscript's internal representation of a buildexpression is not meant to be used in buildplanner contexts. I've hidden `script.Expr` and added a `script.BuildExpression()` method that performs `$at_time` variable substitution in the returned expression so that it can be used by the buildplanner. I anticipate we'll be doing similar processing in the future. Now anyone who uses a buildscript needs to call `BuildExpression()` to get a buildplanner-friendly buildexpression, and `script.expr` can still be used internally for buildscript-friendly buildexpressions.

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -944,6 +944,14 @@ func (e *BuildExpression) MaybeUpdateTimestamp(ts strfmt.DateTime) error {
 	return nil
 }
 
+func (e *BuildExpression) Copy() (*BuildExpression, error) {
+	bytes, err := json.Marshal(e)
+	if err != nil {
+		return nil, errs.Wrap(err, "Failed to marshal build expression during copy")
+	}
+	return New(bytes)
```
Naatan, line 952: Note I believe you can do this more cheaply by dereferencing and creating a new reference, eg.

```
v := *e
return &v
```

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -135,38 +138,35 @@ func (s *Script) String() string {
...
-	}
-}
-
-// prepareForUse prepares this script's buildexpression for use in a buildplanner context.
+// BuildExpression returns a copy of this script's underlying buildexpression for use in a
+// buildplanner context.
 // For example, the solve node's "at_time" will not contain a variable reference if this script
 // has a top-level assignment for it.
-func (s *Script) prepareForUse() {
+func (s *Script) BuildExpression() (*buildexpression.BuildExpression, error) {
```
Naatan, line 154: An elegant solution to a dirty problem, I like it!

I wonder if we should name this something more explicit though, to communicate that it's not the "raw" build expression? eg. `ProcessedBuildExpression()`.

Naatan (approved):
Looks good to me! Two minor suggestions/questions but I'm ok landing it as is.

pkg/platform/runtime/buildexpression/buildexpression.go
mitchell-as: I'm pretty sure that's just copying the top-level BuildExpression struct, not all the structs that live inside it as pointers.

pkg/platform/runtime/buildscript/buildscript.go
mitchell-as: I don't think clients need to be aware that there's a distinction. buildscript's internals care, sure, but outside of buildscripts, I don't think anyone needs to know.

PR #3150, opened 2024-03-06
Add `state eval`

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2603" title="DX-2603" target="_blank">StoryDX-2603</a>  As a user I can invoke specific buildscript targets
  </td></table>
  <br />


pkg/platform/model/buildplanner.go
```diff
@@ -29,8 +29,9 @@ import (
 )
 
 const (
-	pollInterval = 1 * time.Second
-	pollTimeout  = 30 * time.Second
+	pollInterval       = 1 * time.Second
+	pollTimeout        = 30 * time.Second
+	buildStatusTimeout = 30 * time.Minute
```
MDrakos, line 34: Not sure about this timeout. Ideally, we would be using the buildlogstreamer.

internal/locale/locales/en-us.yaml
```diff
@@ -1546,5 +1546,7 @@ uploadingredient_success:
     You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --ts now`[/RESET].
 err_runtime_cache_invalid:
   other: Your runtime needs to be updated, please run '[ACTIONABLE]state refresh[/RESET]'.
+err_target_not_found:
+  other: Could not find build target '{{.V0}}'. Please check your buildscript and ensure the target exists.
```
Naatan, line 1550:
```suggestion
  other: Could not find buildscript target '[ACTIONABLE]{{.V0}}[/reset]'. Please check your buildscript and ensure the target exists.
```

internal/runners/eval/eval.go
```diff
@@ -0,0 +1,92 @@
...
+	var target string
+	for _, assignment := range script.Assignments {
+		if strings.EqualFold(assignment.Key, params.Target) {
+			target = assignment.Key
+		}
+	}
+
+	if target == "" {
+		return errTargetNotFound{target: params.Target}
+	}
```
Naatan, line 78: This should be the job of the API to validate. This feels overly opinionated to validate this client side, we should just ensure that we can communicate the API error that this would result in.

pkg/platform/model/buildplanner.go
Naatan:
```suggestion
	buildStatusTimeout = 24 * time.Hour
```

Let's set this much higher. We don't really know yet what other potential use-cases there are. And technically the client shouldn't care if you run it for several hours, long as we can communicate the API timing out.

pkg/platform/model/buildplanner.go
```diff
@@ -600,3 +601,98 @@ func VersionStringToRequirements(version string) ([]bpModel.VersionRequirement,
...
+		return errs.New("Build is nil")
+	}
+
+	if bpModel.IsErrorResponse(resp.Project.Commit.Build.Type) {
+		return bpModel.ProcessBuildError(resp.Project.Commit.Build, "Could not process error response from evaluate target")
+	}
+
+	if err := bp.pollBuildStatus(commitID); err != nil {
+		return errs.Wrap(err, "Failed to poll build status")
+	}
```
Naatan, line 639: This should be its own public function that the runner invokes imo. These model functions shouldn't be too opinionated.

pkg/platform/model/buildplanner.go
```diff
@@ -600,3 +601,98 @@ func VersionStringToRequirements(version string) ([]bpModel.VersionRequirement,
...
+
+				if artifact.Status != bpModel.ArtifactSucceeded {
+					completed = false
+					break
+				}
+			}
+
+			if completed {
+				return nil
+			}
```
Naatan, line 693: We should also have a fallback for `build.Status == bpModel.Completed` here imo. Once we hit that there is no point to keep polling, but the code as written could technically do this.

Naatan (changes requested):

pkg/platform/model/buildplanner.go
MDrakos: Unfortunately, the build plan status does not accurately reflect the overall build. There is a story to get this updated but for now the solution we decided on was to check each artifact status.

See my question [here](https://activestate.slack.com/archives/C03T15HMBMG/p1709743461056899) and the follow-up story [here](https://activestatef.atlassian.net/browse/AR-127)

Naatan: Yeah but I think that only speaks to the false negatives, not false positives, right?

In other words; it might not say the status is completed, even though it is. But if it says it is completed, it is for sure done?

internal/runners/eval/eval.go
MDrakos: The API error in question has messaging around build expressions, not build scripts. Do we want to forward this to the user? The error in question is a planning error which can have different message contents so I believe it would come down to string matching to determine the error type.

Here's the error I got: `Unable to find requested target 'thing' in build expression`

pkg/platform/model/buildplanner.go
MDrakos: Ah I see. Yes, you're correct. Looking at the docstring for the build plan statuses `COMPLETED` should work here.

internal/runners/eval/eval.go
MDrakos: I'm fine with the string matching approach if that's what we want to do for now. We can file a follow-up to get a more specific error for this case from the API.

Naatan: Just communicating the API error alone is enough for now. But yeah we should file something with them so we can programmatically assert this error and give it our own message.

Naatan (changes requested):
Looks good; just need to conclude this point:

https://github.com/ActiveState/cli/pull/3150/files/860ed376161c6a5770851c7f3618a4feb99c7fe6#diff-5925869b04b7c4d506496436d05dd1c785690baccb0a2c734e6c17f061e61dc7

internal/runners/eval/eval.go
MDrakos: Follow up filed here: https://activestatef.atlassian.net/browse/PB-4929

Naatan (approved):

PR #3151, opened 2024-03-06
Added ability to fetch "now" time from the Platform.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2595" title="DX-2595" target="_blank">TaskDX-2595</a>  `--ts now` should source time from platform
  </td></table>
  <br />
 

This only works on a PR deploy: https://github.com/ActiveState/TheHomeRepot/pull/13250

### DIFF
--- a/internal/captain/values.go
+++ b/internal/captain/values.go
@@ -8,7 +8,10 @@ import (
 
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/platform/model"
 )
 
 // NameVersionValue represents a flag that supports both a name and a version, the following formats are supported:
@@ -242,7 +245,12 @@ func (u *TimeValue) String() string {
 
 func (u *TimeValue) Set(v string) error {
 	if v == "now" {
-		u.Time = ptr.To(time.Now())
+		latest, err := model.FetchLatestRevisionTimeStamp(authentication.LegacyGet())
+		if err != nil {
+			multilog.Error("Unable to determine latest revision time: %v", err)
+			latest = time.Now()
+		}
+		u.Time = ptr.To(latest)
 	} else {
 		u.raw = v
 		tsv, err := time.Parse(time.RFC3339, v)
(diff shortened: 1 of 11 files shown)


internal/captain/values.go
```diff
@@ -242,7 +245,12 @@ func (u *TimeValue) String() string {
 
 func (u *TimeValue) Set(v string) error {
 	if v == "now" {
-		u.Time = ptr.To(time.Now())
+		latest, err := model.FetchLatestRevisionTimeStamp(authentication.LegacyGet())
```
mitchell-as, line 248: I know it's gross to use legacy functions, but we don't have access to prime as far as I can tell.

Naatan: That's not a precendent, that's legacy code that hasn't yet been refactored.

This is kind of a weird situation though. From the perspective of captain.TimeValue it seems presumptuous to equate "now" to a platform API call.

I think what we should do is update TimeValue so that it can communicate what the original input was, and then we update it to the appropriate timestamp in the runner.

Naatan (changes requested):

mitchell-as:
Very well. A side effect of the change is that `state info` and `state search` gain the ability to use the project's current commit time if it's later than the inventory timestamp. Previously only `state install` had this capability.

internal/captain/values.go
```diff
@@ -232,6 +231,7 @@ func (p *PackagesValue) Type() string {
 type TimeValue struct {
 	raw  string
 	Time *time.Time
+	Now  bool
```
Naatan, line 234: Nit: Should make this a method rather than a public property, so it's compatible with interfaces.

internal/runners/packages/install.go
```diff
@@ -42,36 +46,58 @@ func (a *Install) Run(params InstallRunParams, nsType model.NamespaceType) (rerr
...
+	)
+}
+
+// getTime returns a timestamp based on the given "--ts" value.
+// If "now" was given, returns "now" according to the platform.
+// If no timestamp was given but the current project's commit time is after the latest inventory
+// timestamp, returns that commit time.
+// Otherwise, returns the specified timestamp or nil (which falls back on the default Platform
+// timestamp for a given operation)
+func getTime(ts *captain.TimeValue, auth *authentication.Auth, proj *project.Project) (*time.Time, error) {
```
Naatan, line 72: Please move this into its own file. It's used by `info.go`, `install.go`, `search.go`. Each of those files are specific to their own command.

Naatan (changes requested):
Looks good! Just some minor cleanup.

Naatan (approved):

PR #3152, opened 2024-03-07
Disable buildscripts by default and make them an opt-in mechanic

Naatan:
<table>
<td>
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -332,7 +332,7 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 		RequirementName:      name,
 		RequirementVersion:   requirements,
 		RequirementNamespace: *ns,
-		RequirementRevision: requirementRevision,
+		RequirementRevision:  requirementRevision,
 		Operation:            operation,
 		TimeStamp:            ts,
 	}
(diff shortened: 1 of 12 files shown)


internal/runbits/requirements/requirements.go
```diff
@@ -415,16 +415,17 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
...
-		return errs.Wrap(err, "Could not get remote build expr")
-	}
+	if r.Config.GetBool(constants.OptinBuildscriptsConfig) {
+		bp := model.NewBuildPlannerModel(r.Auth)
+		expr, err := bp.GetBuildExpression(r.Project.Owner(), r.Project.Name(), commitID.String())
+		if err != nil {
+			return errs.Wrap(err, "Could not get remote build expr")
+		}
 
-	// Note: a commit ID file needs to exist at this point.
```
mitchell-as, line 424: I would like to keep this comment because it's important. This code block has shifted around the file over the last year.

mitchell-as (changes requested):
Looks good to me, but with one minor change.

internal/runbits/requirements/requirements.go
Naatan: It's no longer relevant; the commit ID file no longer exists. We went back on that whole thing.

mitchell-as (approved):

PR #3155, opened 2024-03-07
Fix buildscript should only be created if opted in

Naatan:
<table>
<td>
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/runtime.go
+++ b/pkg/platform/runtime/runtime.go
@@ -128,12 +128,6 @@ func (r *Runtime) validateCache() error {
 		return nil
 	}
 
-	// Check if local build script has changes that should be committed.
-	script, err := buildscript.NewScriptFromProject(r.target, r.auth)
-	if err != nil {
-		return errs.Wrap(err, "Unable to get local build script")
-	}
-
 	commitID := r.target.CommitUUID().String()
 	expr, err := r.store.GetAndValidateBuildExpression(commitID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/platform/runtime/runtime.go
```diff
@@ -152,8 +146,15 @@ func (r *Runtime) validateCache() error {
...
 	}
 
-	if r.cfg.GetBool(constants.OptinBuildscriptsConfig) && !script.EqualsBuildExpressionBytes([]byte(expr)) {
-		return &NeedsCommitError{errs.New("Runtime changes should be committed")}
+	// Check if local build script has changes that should be committed.
+	if r.cfg.GetBool(constants.OptinBuildscriptsConfig) {
+		script, err := buildscript.NewScriptFromProject(r.target, r.auth)
+		if err != nil {
+			return errs.Wrap(err, "Unable to get local build script")
+		}
```
mitchell-as, line 154: My initial thought was that it's creating a build script from the just-stored buildexpression and comparing that expr against itself (which will never fail). The original intent of this block being above was to ensure using the previous build expression, and not the new one.

I turned on the "Test: commit" label to try and catch this. It failed, but I don't know if it was because of this concern though. Regardless, that failure needs to be addressed first.

mitchell-as (changes requested):

Naatan:
Yeah the commit test failing is unrelated to this story. For some reason this integration test thinks we're in a unit test..

https://activestatef.atlassian.net/browse/DX-2618

Naatan:
That's fair. I'll pick up the bug so we can stay ahead of the issue. Maybe you're right and it was introduced by #3152 but given the symptom (wrong API host) it just seems unlikely to me.

PR #3157, opened 2024-03-07
Fix buildscript test

Naatan:
<table>
<td>
  </td></table>
  <br />

### DIFF
--- a/test/integration/commit_int_test.go
+++ b/test/integration/commit_int_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/ActiveState/cli/internal/constants"
-	"github.com/ActiveState/cli/internal/errs"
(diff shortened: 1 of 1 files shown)


test/integration/commit_int_test.go
```diff
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/ActiveState/cli/internal/constants"
-	"github.com/ActiveState/cli/internal/errs"
```
mitchell-as, line 9: Oops, this is still needed.

mitchell-as (changes requested):

test/integration/commit_int_test.go
Naatan: You had one job Goland! 

Well actually a lot of jobs, and I probably just confused the hell out of it. But it's easier to blame Goland..

PR #3158, opened 2024-03-08
Support dependency type

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2516" title="DX-2516" target="_blank">StoryDX-2516</a>  `state publish` supports specifying the type of a dependency
  </td></table>
  <br />
 

This PR is mostly a change set from Nathan with a few minor adjustments. The other part of this change is [here](https://github.com/ActiveState/TheHomeRepot/pull/13328) which is the backend change to allow dependency type specification.


test/integration/publish_int_test.go
```diff
@@ -378,29 +406,24 @@ authors:
 
 					cp.Expect("Y/n")
 
-					snapshot := cp.Snapshot()
```
MDrakos, line 381: I removed this because with the dependencies from the first test the snapshot was too small to capture all of the YAML output.

mitchell-as: Still, the test was useful, right? Can we not add some sort of `cp.Expect()` or `cp.ExpectRe()` call that would be equivalent or close to it?

mitchell-as (changes requested):
This looks fine, but I did have one question in the review.

test/integration/publish_int_test.go
MDrakos: Sure, it's just frail as the correct YAML formatting has to be contained within the snapshot. 

I've added the parsing back and a conditional to attempt parsing the snapshot or not. If there is an error parsing the YAML we will also fall back to the provided ingredient name, namespace, and version.

mitchell-as (approved):
Nice!

PR #3159, opened 2024-03-08
Ensure commits only ever write the buildscript to a commit

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2620" title="DX-2620" target="_blank">BugDX-2620</a>  `buildscript.as` changed to previous content every time `state commit` executed
  </td></table>
  <br />
 

Effectively this was running into the implicitness of Sync, which is tracked here:

https://activestatef.atlassian.net/browse/DX-2619

Sync is doing too much at the moment. Rather than try and fix that and all the other issues that might arise from doing so, I opted to instead implement the commit logic in the commit runner, which is frankly where it should live anyway.

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1546,3 +1546,5 @@ uploadingredient_success:
     You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --ts now`[/RESET].
 err_runtime_cache_invalid:
   other: Your runtime needs to be updated, please run '[ACTIONABLE]state refresh[/RESET]'.
+err_buildscript_notexist:
+  other: Your project does not have a buildscript.as file. Please refer to the documentation on how to create one.
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1546,5 +1546,5 @@ uploadingredient_success:
     You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --ts now`[/RESET].
 err_runtime_cache_invalid:
   other: Your runtime needs to be updated, please run '[ACTIONABLE]state refresh[/RESET]'.
-err_buildscript_notexist:
+err_buildscript_not_exist:
   other: Your project does not have a buildscript.as file. Please refer to the documentation on how to create one.
--- a/pkg/platform/runtime/buildscript/file.go
+++ b/pkg/platform/runtime/buildscript/file.go
@@ -22,40 +22,59 @@ type projecter interface {
 	Name() string
 }
 
-func NewScriptFromProject(proj projecter, auth *authentication.Auth) (*Script, error) {
-	return newScriptFromFile(filepath.Join(proj.ProjectDir(), constants.BuildScriptFileName), proj.Owner(), proj.Name(), auth)
+var ErrBuildscriptNotExist = errors.New("Build script does not exist")
+
+// ScriptFromProjectWithFallback will source the buildscript from the project, and create it if it does not exist.
+func ScriptFromProjectWithFallback(proj projecter, auth *authentication.Auth) (*Script, error) {
+	path := filepath.Join(proj.ProjectDir(), constants.BuildScriptFileName)
+
+	buildscript, err := ScriptFromFile(path)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return nil, errs.Wrap(err, "Could not read build script from file")
+		}
+
+		logging.Debug("Build script does not exist. Creating one.")
+		commitId, err := localcommit.Get(filepath.Dir(path))
+		if err != nil {
+			return nil, errs.Wrap(err, "Unable to get the local commit ID")
+		}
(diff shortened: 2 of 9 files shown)


pkg/platform/runtime/buildscript/file.go
```diff
@@ -22,40 +22,59 @@ type projecter interface {
...
+var ErrBuildscriptNotExist = errors.New("Build script does not exist")
 
-func newScriptFromFile(path, org, project string, auth *authentication.Auth) (*Script, error) {
-	if data, err := fileutils.ReadFile(path); err == nil {
-		return NewScript(data)
-	} else if !errors.Is(err, os.ErrNotExist) {
-		return nil, errs.Wrap(err, "Could not read build script")
-	}
+// ScriptFromProjectWithFallback will source the buildscript from the project, and create it if it does not exist.
+func ScriptFromProjectWithFallback(proj projecter, auth *authentication.Auth) (*Script, error) {
```
Naatan, line 28: Hopefully this naming makes it evident that this is doing more than the name would otherwise suggest.

internal/locale/locales/en-us.yaml
```diff
@@ -1546,3 +1546,5 @@ uploadingredient_success:
     You can install this package by running `[ACTIONABLE]state install {{.V1}}/{{.V0}} --ts now`[/RESET].
 err_runtime_cache_invalid:
   other: Your runtime needs to be updated, please run '[ACTIONABLE]state refresh[/RESET]'.
+err_buildscript_notexist:
```
mitchell-as, line 1549: nit: err_buidlscript_not_exist is more readable.

mitchell-as (changes requested):
Please re-tag me for review once it builds properly and the tests pass.

test/integration/pull_int_test.go
```diff
@@ -98,15 +98,15 @@ func (suite *PullIntegrationTestSuite) TestMergeBuildScript() {
 	proj, err := project.FromPath(ts.Dirs.Work)
 	suite.NoError(err, "Error loading project")
 
-	_, err = buildscript.NewScriptFromProject(proj, nil)
+	_, err = buildscript.ScriptFromProjectWithFallback(proj, nil)
```
mitchell-as, line 101: Shouldn't this be without the fallback to verify the buildscript was created? Otherwise, this call creates it and the next line will always succeed.

test/integration/pull_int_test.go
```diff
@@ -98,15 +98,15 @@ func (suite *PullIntegrationTestSuite) TestMergeBuildScript() {
...
 	suite.Require().NoError(err) // just verify it's a valid build script
 
 	cp = ts.Spawn("pull")
 	cp.Expect("Unable to automatically merge build scripts")
 	cp.ExpectNotExitCode(0)
 	ts.IgnoreLogErrors()
 
-	_, err = buildscript.NewScriptFromProject(proj, nil)
+	_, err = buildscript.ScriptFromProjectWithFallback(proj, nil)
```
mitchell-as, line 109: Same question as above.

pkg/platform/runtime/buildscript/file.go
```diff
@@ -22,40 +22,59 @@ type projecter interface {
...
+	path := filepath.Join(proj.ProjectDir(), constants.BuildScriptFileName)
+	return ScriptFromFile(path)
+}
+
+func ScriptFromFile(path string) (*Script, error) {
+	data, err := fileutils.ReadFile(path)
 	if err != nil {
-		return nil, errs.Wrap(err, "Unable to write build script")
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, errs.Pack(err, ErrBuildscriptNotExist)
```
mitchell-as, line 69: For my own information, why `errs.Pack()` instead of `errs.Wrap()`?

mitchell-as (changes requested):
Please also run "Test: pull" to verify your pull test is actually succeeding.

test/integration/pull_int_test.go
Naatan: That's probably true, but this PR isn't about fixing all the buildscript things, just the `state commit` one. This code is still functionally the same as before, the function just got renamed.

pkg/platform/runtime/buildscript/file.go
Naatan: Let's say I implement 

```
type ErrBuildscriptNotExist struct {
    error
}
```

I can't just instantiate that as `ErrBuildScriptNotExist{wrappedErr}`, because then the type of `wrappedErr` can no longer be interpreted. What I could do is add an `Unwrap()` method to  this type, which returned the wrapped error, but that's a lot of boilerplate.

So to save myself some boilerplate and still easily be able to communicate two types I use `errs.Pack()`. 

This does mean that `ErrBuildscriptNotExist` does not wrap its underlying error, instead it lives adjacent to it. To me the compromise is worth it. If we were to require types to always go to the extent of adding Unwrap logic they'd be painful to use and we'd try to avoid using them, which ultimately is worse than the caveats of `.Pack()`.

mitchell-as (approved):

PR #3161, opened 2024-03-08
Refactor `os.Exit()` out of non-main packages.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2489" title="DX-2489" target="_blank">TaskDX-2489</a>  Refactor out uses of `os.Exit()`
  </td></table>
  <br />
 

Refactoring `os.Exit()` out of the `authentication` package turned into a more generic refactor of packages that depended on `authentication.LegacyGet()`, either explicitly or implicitly.

It's a big diff, but there's the added benefit that a lot of code is no longer depending on legacy authentication functions.


mitchell-as:
The failing test is a one of the usual timeouts.

Naatan (approved):
Looks great! Thanks for going through the trouble.

PR #3162, opened 2024-03-11
Show the full name of the artifact being downloaded.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2591" title="DX-2591" target="_blank">BugDX-2591</a>  BUILDS: The name of the build cut-off
  </td></table>
  <br />
 

It often gets truncated when put next to the progressbar.


MDrakos (approved):
I seem to recall that we decided on this localization update in stand, is that correct? It doesn't appear to be reflected in the story so we may want to update it there just for record keeping purposes.

PR #3164, opened 2024-03-12
Define and handle default configuration values.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2549" title="DX-2549" target="_blank">TaskDX-2549</a>  Our config library handles default values
  </td></table>
  <br />

### DIFF
--- a/internal/config/instance.go
+++ b/internal/config/instance.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/installation/storage"
 	"github.com/ActiveState/cli/internal/logging"
+	mediator "github.com/ActiveState/cli/internal/mediators/config"
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/profile"
 	"github.com/ActiveState/cli/internal/rtutils/singlethread"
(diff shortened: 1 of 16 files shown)


mitchell-as:
The test failures are sporadic and do not appear to be connected to this PR.

internal/config/instance.go
```diff
@@ -191,6 +192,16 @@ func (i *Instance) Get(key string) interface{} {
 	return result
 }
 
+func (i *Instance) Get(key string) interface{} {
+	result := i.rawGet(key)
+	if result == nil {
```
MDrakos, line 197: nit: can we flip this logic and return early?

internal/mediators/config/registry.go
```diff
@@ -31,13 +32,22 @@ var registry = make(Registry)
...
+		return Option{key, String, "", EmptyEvent, EmptyEvent, false}
 	}
 	return rule
 }
 
-func RegisterOption(key string, t Type, get Event, set Event) {
-	registry[key] = Option{key, t, get, set, true}
+// Registers a config option.
+// Name, Type, and Default are required fields.
+func RegisterOption(option Option) {
```
MDrakos, line 42: I'm worried that policing the required fields via a docstring isn't enough. We should consider making the required options arguments to this function rather than just an `Option` argument.

mitchell-as: Okay, I've switched it back and created a separate function for registering with get/set events. Since all existing callers do not use get/set events, they are using the simplified version.

MDrakos (approved):
Looks good!

PR #3167, opened 2024-03-12
Show timestamp with `state publish` output.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2628" title="DX-2628" target="_blank">TaskDX-2628</a>  `state publish` prints a timestamp that can be used in the buildscript
  </td></table>
  <br />
 

Sample output:

```
% build/state publish <name>-<version>.zip --namespace <namespace>
█ Publish Ingredient

Publishing ingredient...
Successfully published as:
Name: <name>
Namespace: <namespace>
Version: <version>
Revision: 1
Timestamp: 2024-03-12T18:44:34.092Z

You can install this package by running `state install <namespace>/<name> --ts now`.
```


Naatan (approved):
Nice, glad we can just source it from the response.

PR #3173, opened 2024-03-13
Buildscript version functions should take name, value assignment arguments instead of just values.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2637" title="DX-2637" target="_blank">TaskDX-2637</a>  Buildscript version functions take named parameters
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/buildscript/buildscript.go
+++ b/pkg/platform/runtime/buildscript/buildscript.go
@@ -197,7 +197,7 @@ func transformRequirements(reqs *buildexpression.Var) *buildexpression.Var {
 //
 // into something like
 //
-//	Req(name = "<namespace>/<name>", version = <op>("<version>"))
+//	Req(name = "<namespace>/<name>", version = <op>(value = "<version>"))
 func transformRequirement(req *buildexpression.Value) *buildexpression.Value {
 	newReq := &buildexpression.Value{
 		Ap: &buildexpression.Ap{
@@ -249,7 +249,7 @@ func transformRequirement(req *buildexpression.Value) *buildexpression.Value {
 //
 // into something like
 //
-//	And(<op1>("<version1>"), <op2>("<version2>"))
+//	And(<op1>(value = "<version1>"), <op2>(value = "<version2>"))
 func transformVersion(requirements *buildexpression.Var) *buildexpression.Ap {
 	var aps []*buildexpression.Ap
 	for _, constraint := range *requirements.Value.List {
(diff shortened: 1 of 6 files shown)


pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -266,14 +268,16 @@ func transformVersion(requirements *buildexpression.Var) *buildexpression.Ap {
...
 
 	args := make([]*buildexpression.Value, len(aps))
 	for i, ap := range aps {
-		args[i] = &buildexpression.Value{Ap: ap}
+		args[i] = &buildexpression.Value{
+			Assignment: &buildexpression.Var{Name: "val", Value: &buildexpression.Value{Ap: ap}},
+		}
 	}
-	return &buildexpression.Ap{Name: andFuncName, Arguments: args} // e.g. And(Gt("1.0"), Lt("3.0"))
+	return &buildexpression.Ap{Name: andFuncName, Arguments: args} // e.g. And(val = Gt(val = "1.0"), val = Lt(val = "3.0"))
```
mitchell-as, line 280: I'm not 100% sure this is what we expect for the `And()` operator.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -266,14 +268,26 @@ func transformVersion(requirements *buildexpression.Var) *buildexpression.Ap {
...
+		return aps[0] // e.g. Eq(value = "1.0")
 	}
 
-	args := make([]*buildexpression.Value, len(aps))
-	for i, ap := range aps {
-		args[i] = &buildexpression.Value{Ap: ap}
+	// e.g. And(left = Gt(value = "1.0"), right = Lt(value = "3.0"))
+	// Iterate backwards over the requirements array and construct a binary tree of 'And()' functions.
+	// For example, given [Gt(value = "1.0"), Ne(value = "2.0"), Lt(value = "3.0")], produce:
+	//   And(left = Gt(value = "1.0"), right = And(left = Ne(value = "2.0"), right = Lt(value = "3.0")))
```
Naatan, line 277: @srobertson could you confirm this is what you want? If `And()` only takes 2 parameters, then this is all we can reasonably do. But I could also see an argument for making `And()` take 2+ parameters, in which case we'd also need to revisit the naming.

To summarize the comment, technically you can have a version constraint of:

```
>1.0,<3.0,!=2.0
```

Which I think ideally translates to something like:

```
And(value1 = Gt(value="1.0"), value2 = Lt("3.0"), value3 = Ne("3.0"))
```

But as we only have the left and right parameters, we instead end up with:

```
And(left = Gt(value="1.0"), right = And(left = Lt("3.0"), right = Ne("3.0"))
```

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -266,14 +268,26 @@ func transformVersion(requirements *buildexpression.Var) *buildexpression.Ap {
...
+			right = &buildexpression.Value{Ap: ap}
+		}
+		args := []*buildexpression.Value{
+			{Assignment: &buildexpression.Var{Name: "left", Value: &buildexpression.Value{Ap: aps[i]}}},
+			{Assignment: &buildexpression.Var{Name: "right", Value: right}},
+		}
+		ap = &buildexpression.Ap{Name: andFuncName, Arguments: args}
 	}
-	return &buildexpression.Ap{Name: andFuncName, Arguments: args} // e.g. And(Gt("1.0"), Lt("3.0"))
+	return ap
```
Naatan, line 290: Love it when a complex problem can be solved with such relatively simple code. Kudos!

Naatan (approved):

PR #3174, opened 2024-03-13
Produce buildscript parsing errors

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2639" title="DX-2639" target="_blank">TaskDX-2639</a>  We produce buildscript parsing errors
  </td></table>
  <br />
 

Unfortunately, parsing errors don't cover all of the cases we may have initially though. For example, if you provide a requirement like `Req(name = "language/python", version = "3.10.13")` This will not be a parsing error but an error when we try to construct the equivalent build expression. This produces a not so helpful error: `Invalid or unknown argument: &{version 0x1400093e400}`.

For the actual parsing error we get something nicer like:

```
❯ ~/work/cli/build/state commit
█ Commit Changes

 • Creating commit x Failed

 x Could not parse build script: buildscript.as:15:57: literal not terminated.

█ Need More Help?
 • Run → 'state commit --help' for general help.
```

For now, this PR is only hoisting up the parsing errors.

### DIFF
--- a/pkg/platform/runtime/buildscript/buildscript.go
+++ b/pkg/platform/runtime/buildscript/buildscript.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
@@ -79,12 +80,20 @@ func NewScript(data []byte) (*Script, error) {
 
 	script, err := parser.ParseBytes(constants.BuildScriptFileName, data)
 	if err != nil {
+		parseErrors := errs.Unpack(err)
+		if len(parseErrors) > 0 {
+			return nil, locale.NewInputError("err_parse_buildscript_bytes", "Could not parse build script: {{.V0}}", parseErrors[len(parseErrors)-1].Error())
+		}
 		return nil, errs.Wrap(err, "Could not parse build script")
 	}
 
 	// Construct the equivalent buildexpression.
 	bytes, err := json.Marshal(script)
 	if err != nil {
+		parseErrors := errs.Unpack(err)
+		if len(parseErrors) > 0 {
+			return nil, locale.NewInputError("err_marshall_buildscript", "Could not marshall build script: {{.V0}}", parseErrors[len(parseErrors)-1].Error())
+		}
 		return nil, errs.Wrap(err, "Could not marshal build script to build expression")
 	}
 	expr, err := buildexpression.New(bytes)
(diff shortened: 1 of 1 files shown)


pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -79,6 +80,10 @@ func NewScript(data []byte) (*Script, error) {
 
 	script, err := parser.ParseBytes(constants.BuildScriptFileName, data)
 	if err != nil {
+		parseErrors := errs.Unpack(err)
```
Naatan, line 83: This should use errors.As or something more determinate. The code as written assumes that the returned error is specified in a very specific way that is not enforced through the type system.

Naatan (changes requested):

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -79,7 +80,7 @@ func NewScript(data []byte) (*Script, error) {
 
 	script, err := parser.ParseBytes(constants.BuildScriptFileName, data)
 	if err != nil {
-		return nil, errs.Wrap(err, "Could not parse build script")
+		return nil, locale.WrapInputError(err, "err_parse_buildscript_bytes", "Could not parse build script: {{.V0}}", errs.JoinMessage(err))
```
MDrakos, line 83: Should we assume an input error here?

Naatan: Hmm good question. The parser does use our parsing logic right? So if an error happened it could be due to something we did and not the user?

mitchell-as: It depends. 99% of the time this will be a user error because they introduced a syntax error in the buildscript. We have unit tests that do round-trip testing, so if we're improperly writing the buildscript, then we should catch it before we get to this point. Technically it's possible we're writing a bad build script though if we have a complicated and/or unexpected buildexpression we're translating. It's pretty unlikely in my opinion. I'm okay with this being an input error.

MDrakos: Correct, we define the language schema and the parser does it's work based on that. So we could get errors if:
1. We made a mistake in the language schema
2. Some unexpected error from the parsing library
3. A buildscript is actual malformed

We would likely want to report to rollbar for the first two types of errors and the third is the one we disregard. It's will be hard to filter those out though.

I was able to find a `participle.Error` interface that I'm now using in the error check. This should allow us to be more confident that they are parsing errors. I'm wrapping the participle errors as input and the others as regular local errors for now.

Naatan: Awesome, thanks guys! I'll defer to your better judgment.

pkg/platform/runtime/buildscript/buildscript.go
```diff
@@ -79,7 +81,11 @@ func NewScript(data []byte) (*Script, error) {
 
 	script, err := parser.ParseBytes(constants.BuildScriptFileName, data)
 	if err != nil {
-		return nil, errs.Wrap(err, "Could not parse build script")
+		var parseError participle.Error
+		if errors.As(err, &parseError) {
+			return nil, locale.WrapInputError(err, "err_parse_buildscript_bytes", "Could not parse build script: {{.V0}}: {{.V1}}", parseError.Position().String(), parseError.Message())
+		}
+		return nil, locale.WrapError(err, "err_parse_buildscript_bytes", "Could not parse build script")
```
Naatan, line 88: Would it still be worth reporting the underlying error here?

Naatan (changes requested):

Naatan (approved):

PR #3175, opened 2024-03-15
`state commit` should set the `at_time` specified in the buildscript.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2642" title="DX-2642" target="_blank">BugDX-2642</a>  Buildscript to build expression at_time definition doesn't get turned into "$at_time"
  </td></table>
  <br />
 

Keep "$at_time" variables in buildexpressions.

Do not replace buildexpression "$at_time" with commit time. Instead, just set the commit time when staging commits, etc. This allows for using a buildscripts `Expr` field directly again.

This reverts most of the changes in [#3149](https://github.com/ActiveState/cli/issues/3149).

### DIFF
--- a/internal/runbits/buildscript/buildscript.go
+++ b/internal/runbits/buildscript/buildscript.go
@@ -21,6 +21,10 @@ import (
 	"github.com/sergi/go-diff/diffmatchpatch"
 )
 
+// getEquivalentBuildScript constructs and returns a buildscript that represents the
+// buildexpression and at_time of a fetched project commit.
+// This is used either to compare the local build script with a remote equivalent, or to update
+// the local build script with a remote equivalent.
 func getEquivalentBuildScript(proj *project.Project, customCommit *strfmt.UUID, auth *authentication.Auth) (*buildscript.Script, error) {
 	bp := model.NewBuildPlannerModel(auth)
 	commitID, err := localcommit.Get(proj.Dir())
--- a/internal/runbits/buildscript/buildscript.go
+++ b/internal/runbits/buildscript/buildscript.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
@@ -14,14 +15,13 @@ import (
 	"github.com/ActiveState/cli/pkg/localcommit"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
-	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildscript"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/go-openapi/strfmt"
 	"github.com/sergi/go-diff/diffmatchpatch"
 )
 
-func getBuildExpression(proj *project.Project, customCommit *strfmt.UUID, auth *authentication.Auth) (*buildexpression.BuildExpression, error) {
+func getEquivalentBuildScript(proj *project.Project, customCommit *strfmt.UUID, auth *authentication.Auth) (*buildscript.Script, error) {
 	bp := model.NewBuildPlannerModel(auth)
 	commitID, err := localcommit.Get(proj.Dir())
 	if err != nil {
@@ -30,7 +30,11 @@ func getBuildExpression(proj *project.Project, customCommit *strfmt.UUID, auth *
 	if customCommit != nil {
 		commitID = *customCommit
 	}
-	return bp.GetBuildExpression(commitID.String())
+	expr, atTime, err := bp.GetBuildExpressionAndTime(commitID.String())
+	if err != nil {
+		return nil, errs.Wrap(err, "Unable to get remote build expression and time")
+	}
+	return buildscript.NewFromCommit(atTime, expr)
 }
 
 // Sync synchronizes the local build script with the remote one.
@@ -44,16 +48,16 @@ func Sync(proj *project.Project, commitID *strfmt.UUID, out output.Outputer, aut
 		return false, errs.Wrap(err, "Could not get local build script")
 	}
 
-	expr, err := getBuildExpression(proj, commitID, auth)
+	remoteScript, err := getEquivalentBuildScript(proj, commitID, auth)
 	if err != nil {
-		return false, errs.Wrap(err, "Could not get remote build expr for provided commit")
+		return false, errs.Wrap(err, "Could not get remote build script-equivalent")
 	}
 
 	// If commitID is given, a mutation happened, so prefer the remote build expression.
 	// Otherwise, prefer local changes.
 	if script != nil && commitID == nil {
 		logging.Debug("Checking for changes")
-		if script.EqualsBuildExpression(expr) {
+		if script.Equals(remoteScript) {
 			return false, nil // nothing to do
 		}
 		logging.Debug("Merging changes")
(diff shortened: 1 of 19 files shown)


Naatan (changes requested):
This still isn't using the "$at_time" value in the build expression.

Take the following buildscript:

```
at_time = "2024-03-14T22:36:01Z"
publish_receipt = pypi_publisher(
        audience = "testpypi",
        pypi_project_name = "as-oidc-test",
        pypi_project_version = "0.1",
        pypi_uri = "test.pypi.org",
        src = wheels
)
runtime = state_tool_artifacts(
        build_flags = [
        ],
        src = wheels
)
sources = solve(
        at_time = at_time,
        platforms = [
                "7c998ec2-7491-4e75-be4d-8885800ef5f2"
        ],
        requirements = [
                Req(name = "jeremyp/language/python/as-oidc-test", version = Eq(value = "0.1")),
                Req(name = "language/python", version = Eq(value = "3.9.18"))
        ],
        solver_version = null
)
wheels = wheel_artifacts(
        src = sources
)

main = publish_receipt
```

When I commit this, it gets turned into the following build expression:

```
{
  "let": {
    "in": "$publish_receipt",
    "publish_receipt": {
      "pypi_publisher": {
        "audience": "testpypi",
        "pypi_project_name": "as-oidc-test",
        "pypi_project_version": "0.1",
        "pypi_uri": "test.pypi.org",
        "src": "$wheels"
      }
    },
    "runtime": {
      "state_tool_artifacts": {
        "build_flags": [],
        "src": "$wheels"
      }
    },
    "sources": {
      "solve": {
        "at_time": "2024-03-14T22:36:01.000Z",
        "platforms": [
          "7c998ec2-7491-4e75-be4d-8885800ef5f2"
        ],
        "requirements": [
          {
            "name": "as-oidc-test",
            "namespace": "jeremyp/language/python",
            "version_requirements": [
              {
                "comparator": "eq",
                "version": "0.1"
              }
            ]
          },
          {
            "name": "python",
            "namespace": "language",
            "version_requirements": [
              {
                "comparator": "eq",
                "version": "3.9.18"
              }
            ]
          }
        ],
        "solver_version": null
      }
    },
    "wheels": {
      "wheel_artifacts": {
        "src": "$sources"
      }
    }
  }
}
```

But this is wrong, as it is hard coding the timestamp. It is now in two places; the commit itself AND the build expression.

The at_time field should instead look like this:

```
"at_time": "$at_time",
```

Note that it should ONLY look like that if in the buildscript we used the `at_time` variable there. We do also want to support the user hard coding a timestamp for that field, but that wasn't the case here.

mitchell-as:
Putting this up for review now before I leave for the end of the day. It may need another tweak or two, but tests are passing at the moment.

internal/runners/commit/commit.go
```diff
@@ -93,32 +94,38 @@ func (c *Commit) Run() (rerr error) {
...
+		exprAtTime = &atTimeTime
 	}
 
 	stagedCommitID, err := bp.StageCommit(model.StageCommitParams{
 		Owner:        c.proj.Owner(),
 		Project:      c.proj.Name(),
 		ParentCommit: localCommitID.String(),
-		Expression:   exprBuildscript,
+		Expression:   script.Expr,
+		TimeStamp:    exprAtTime,
```
mitchell-as, line 128: This one line is the crux of the PR. Everything else is just filler 🤣

internal/runbits/buildscript/buildscript.go
```diff
@@ -14,14 +15,13 @@ import (
...
 	"github.com/ActiveState/cli/pkg/platform/model"
-	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
 	"github.com/ActiveState/cli/pkg/platform/runtime/buildscript"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/go-openapi/strfmt"
 	"github.com/sergi/go-diff/diffmatchpatch"
 )
 
-func getBuildExpression(proj *project.Project, customCommit *strfmt.UUID, auth *authentication.Auth) (*buildexpression.BuildExpression, error) {
+func getEquivalentBuildScript(proj *project.Project, customCommit *strfmt.UUID, auth *authentication.Auth) (*buildscript.Script, error) {
```
Naatan, line 24: Can we name this `GetRemoteBuildScript` or something? Or add comments explaining what it does.

pkg/platform/runtime/store/store.go
```diff
@@ -331,3 +335,20 @@ func (s *Store) StoreBuildExpression(expr *buildexpression.BuildExpression, comm
...
+		return nil, ErrNoBuildScriptFile
+	}
+	bytes, err := fileutils.ReadFile(s.buildScriptFile())
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not read buildscript file")
+	}
+	return buildscript.New(bytes)
+}
+
+func (s *Store) StoreBuildScript(script *buildscript.Script) error {
```
Naatan, line 352: I'm not finding a reference to this being called.

Naatan (changes requested):

internal/runbits/buildscript/buildscript.go
mitchell-as: I've added a comment.

pkg/platform/runtime/store/store.go
mitchell-as: Thanks for catching this oversight on my part.

pkg/platform/runtime/runtime.go
```diff
@@ -141,15 +141,32 @@ func (r *Runtime) validateCache() error {
...
+			if err != nil && !errs.Matches(err, buildscript.ErrBuildscriptNotExist) {
 				return errs.Wrap(err, "Unable to get local build script")
 			}
-			if !script.Equals(cachedScript) {
+			if script != nil && !script.Equals(cachedScript) {
 				return &NeedsCommitError{errs.New("Runtime changes should be committed")}
 			}
-		} else if err != store.ErrNoBuildScriptFile {
+
+		case err == store.ErrNoBuildScriptFile:
```
Naatan, line 155:
```suggestion
		case errors.Is(err, store.ErrNoBuildScriptFile):
```

`errors.Is()` will work even if `store.ErrNoBuildScriptFile` is not the top level error.

pkg/platform/runtime/runtime.go
```diff
@@ -141,15 +141,32 @@ func (r *Runtime) validateCache() error {
...
+			if err != nil {
+				return errs.Wrap(err, "Unable to get remote build expression and time")
+			}
+			script, err := buildscript.NewFromCommit(atTime, expr)
+			if err != nil {
+				return errs.Wrap(err, "Unable to get local build script")
+			}
+			if err := r.store.StoreBuildScript(script); err != nil {
+				return errs.Wrap(err, "Unable to store build script")
+			}
```
Naatan, line 167: This shouldn't be in `validateCache()`, it's implicitly creating the buildscript to satisfy some external use-case of this function that is not evident from within this function.

I think we'll want to just always store this buildscript once we have a runtime sourced. Same way we store the build expression I'm guessing.

This really isn't critical for v0.43, but I also don't want this to add more technical debt. I'd be ok taking this out entirely and filing a followup to address the opt-in route.

pkg/platform/runtime/setup/setup.go
```diff
@@ -626,8 +626,15 @@ func (s *Setup) fetchAndInstallArtifactsFromBuildPlan(installFunc artifactInstal
...
-			return nil, nil, errs.Wrap(err, "Could not save build script.")
+		script, err := buildscript.NewFromCommit(buildResult.AtTime, buildResult.BuildExpression)
+		if err != nil {
+			return nil, nil, errs.Wrap(err, "Could not construct build script")
+		}
+		if err := s.store.StoreBuildScript(script); err != nil {
+			return nil, nil, errs.Wrap(err, "Could not save build script")
+		}
+		if err := buildscript.Update(s.target, script.AtTime, script.Expr, s.auth); err != nil {
+			return nil, nil, errs.Wrap(err, "Could not update build script.")
```
Naatan, line 637: This feels like it's gonna be a pain to maintain if we just have to remember to do this anytime we update a buildscript. I suggest we make this its own ticket and we leave the opt-in case unaddressed for this story.

Naatan (changes requested):

pkg/platform/runtime/runtime.go
mitchell-as: I filed DX-2650 for this.

pkg/platform/runtime/setup/setup.go
mitchell-as: I filed DX-2650 for this.

pkg/platform/runtime/runtime.go
mitchell-as: Okay, I'll remove it.

Naatan (approved):

PR #3179, opened 2024-03-18
Report artifact failures on `state eval`

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2645" title="DX-2645" target="_blank">TaskDX-2645</a>  `state eval` lists out log URI's for failed artifacts
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/buildplanner.go
+++ b/pkg/platform/model/buildplanner.go
@@ -637,7 +637,16 @@ func (bp *BuildPlanner) BuildTarget(owner, project, commitID, target string) err
 	return nil
 }
 
+type ErrFailedArtifacts struct {
+	Artifacts []*bpModel.Artifact
+}
+
+func (e ErrFailedArtifacts) Error() string {
+	return "ErrFailedArtifacts"
+}
+
 func (bp *BuildPlanner) PollBuildStatus(commitID, owner, project, target string) error {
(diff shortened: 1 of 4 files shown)


pkg/platform/model/buildplanner.go
```diff
@@ -637,7 +637,16 @@ func (bp *BuildPlanner) BuildTarget(owner, project, commitID, target string) err
 	return nil
 }
 
+type ErrFailedArtifacts struct {
+	Artifacts []*bpModel.Artifact
+}
+
+func (e ErrFailedArtifacts) Error() string {
+	return "ErrFailedArtifacts"
+}
```
MDrakos, line 646: nit: Is this just a placeholder for the error interface? Regardless, this function could provide some value by doing some of the processing you are doing in the rationalize function. Similar to how the planning error works.

MDrakos (approved):
Looks good!

pkg/platform/model/buildplanner.go
Naatan: It's just to satisfy the `error` interface. This would only ever show up in logs, as a stack trace.

I'd rather have `rationalizeError()` hold all the opinions on how these are rendered.

PR #3180, opened 2024-03-18
Add partial version detection back into `state init`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2526" title="DX-2526" target="_blank">TaskDX-2526</a>  state init automatically assumes wildcards
  </td></table>
  <br />
 

Incomplete version numbers will be given a '.x' suffix. For example, python@3.10 will end up requesting 3.10.x.

### DIFF
--- a/test/integration/init_int_test.go
+++ b/test/integration/init_int_test.go
@@ -180,7 +180,7 @@ func (suite *InitIntegrationTestSuite) TestInit_Resolved() {
 	// Run `state languages` to verify a full language version was resolved.
 	cp = ts.Spawn("languages")
 	cp.Expect("python")
-	cp.Expect("3.10 → 3.10.0")
+	cp.Expect("3.10 → 3.10.") // note: the patch version is variable, so just expect that it exists
 	cp.ExpectExitCode(0)
 }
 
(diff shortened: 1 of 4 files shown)


test/integration/init_int_test.go
```diff
@@ -180,7 +180,7 @@ func (suite *InitIntegrationTestSuite) TestInit_Resolved() {
 	// Run `state languages` to verify a full language version was resolved.
 	cp = ts.Spawn("languages")
 	cp.Expect("python")
-	cp.Expect("3.10 → 3.10.0")
+	cp.Expect("Auto → 3.10.") // note: the patch version is variable, so just expect that it exists
```
mitchell-as, line 183: This is pretty gross. Let me know if I should file a ticket to fix this down the road.

Naatan: Is this a typo? I can understand the "auto", that makes sense, but why is `3.10.` missing the `0` at the end?

Naatan (changes requested):

test/integration/init_int_test.go
mitchell-as: No, that is not a typo. If you give 3.10, then it will resolve to the latest patch release of Python. Today it is 3.10.13, but tomorrow it could be 3.10.14. This integration test should not need continuous maintenence, so I just test for that trailing dot, indicating it's a fully resolved version in the end.

PR #3187, opened 2024-03-19
Report CVE information for all dependencies

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2556" title="DX-2556" target="_blank">StoryDX-2556</a>  State install CVE information is recursive
  </td></table>
  <br />
 

This includes some substantial refactoring to runtime solving/updating logic, because in order to achieve the result we want we need to solve the runtime and interpret the resulting buildplan before we actually source that runtime.


internal/runbits/requirements/requirements.go
```diff
@@ -400,6 +355,137 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+
+	return rt, buildResult, &changedArtifacts, nil
+}
+
+func (r *RequirementOperation) cveReport(requirementName, requirementVersion string, artifactChangeset artifact.ArtifactChangeset, operation bpModel.Operation, ns *model.Namespace) error {
+	if !r.Auth.Authenticated() || operation == bpModel.OperationRemoved {
+		return nil
+	}
+
+	reqNameAndVersion := requirementName
```
MDrakos, line 425: Why initialize this as a new variable? It's only being used in the spinner output.

internal/runbits/requirements/requirements.go
```diff
@@ -400,6 +355,137 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+	if requirementVersion != "" {
+		requirementVersion = fmt.Sprintf("%s@%s", requirementName, requirementVersion)
```
MDrakos, line 427: This variable doesn't appear to be used anywhere in the function.

internal/runbits/requirements/requirements.go
```diff
@@ -431,8 +517,19 @@ func (r *RequirementOperation) updateCommitID(commitID strfmt.UUID) error {
 	return nil
 }
 
-func (r *RequirementOperation) shouldPromptForSecurity(vulnerabilities *model.Vulnerabilites) bool {
-	if !r.Config.GetBool(constants.SecurityPromptConfig) || vulnerabilities == nil {
+func (r *RequirementOperation) shouldPromptForSecurity(vulnerabilities model.VulnerableIngredientsByLevels) bool {
+	if !r.Config.IsSet(constants.SecurityPromptConfig) {
```
MDrakos, line 521: Both this, and the prompt level are now registered with defaults so I don't believe we need to be setting these values in the config database for the user.

internal/runbits/requirements/requirements.go
```diff
@@ -441,56 +538,88 @@ func (r *RequirementOperation) shouldPromptForSecurity(vulnerabilities *model.Vu
...
-	out.Notice(locale.Tr("warning_vulnerable_simple", strconv.Itoa(vulnerabilities.Length())))
+func (r *RequirementOperation) summarizeCVEs(out output.Outputer, vulnerabilities model.VulnerableIngredientsByLevels) {
+	out.Print("")
 
-	var pkgVersionVulns []string
-	if len(vulnerabilities.Critical) > 0 {
-		criticalOutput := fmt.Sprintf("[RED]%d Critical: [/RESET]", len(vulnerabilities.Critical))
-		criticalOutput += fmt.Sprintf("[CYAN]%s[/RESET]", strings.Join(vulnerabilities.Critical, ", "))
-		pkgVersionVulns = append(pkgVersionVulns, criticalOutput)
+	if vulnerabilities.CountPrimary == 0 {
```
MDrakos, line 562: nit: could be a switch.

internal/runbits/requirements/requirements.go
```diff
@@ -441,56 +538,88 @@ func (r *RequirementOperation) shouldPromptForSecurity(vulnerabilities *model.Vu
...
+		out.Print(locale.Tr("warning_vulnerable_directonly", strconv.Itoa(vulnerabilities.Count)))
+	} else {
+		out.Print(locale.Tr("warning_vulnerable", strconv.Itoa(vulnerabilities.CountPrimary), strconv.Itoa(vulnerabilities.Count-vulnerabilities.CountPrimary)))
 	}
 
-	if len(vulnerabilities.High) > 0 {
-		highOutput := fmt.Sprintf("[ORANGE]%d High: [/RESET]", len(vulnerabilities.High))
-		highOutput += fmt.Sprintf("[CYAN]%s[/RESET]", strings.Join(vulnerabilities.High, ", "))
-		pkgVersionVulns = append(pkgVersionVulns, highOutput)
+	if vulnerabilities.Critical.Count > 0 {
```
MDrakos, line 570: nit: a lot of repeated logic with each vulnerability level that could be helper functions.

internal/runbits/runtime/refresh.go
```diff
@@ -0,0 +1,113 @@
...
+			out.Notice(locale.T("update_runtime_info"))
+		}
+	}
+
+	return RefreshRuntimeByReference(rt, auth, out, proj, cfg)
+}
+
+// RefreshRuntimeByReference will update the given runtime if necessary. Unlike RefreshRuntime this won't print any UI
+// except for the progress of sourcing the runtime.
+func RefreshRuntimeByReference(
```
MDrakos, line 58: From the docstring and the function logic it's not clear why this has `ByReference` in the name. I could be missing something though.

internal/runbits/runtime/refresh.go
```diff
@@ -0,0 +1,113 @@
...
+	cfg Configurable,
+) (rerr error) {
+	if cfg.GetBool(constants.OptinBuildscriptsConfig) {
+		_, err := buildscript.Sync(proj, ptr.To(rt.Target().CommitUUID()), out, auth)
+		if err != nil {
+			return locale.WrapError(err, "err_update_build_script")
+		}
+	}
+
+	if rt.NeedsUpdate() {
```
MDrakos, line 72: nit: We could flip this, and the one in `UpdateByReference` and return early

pkg/platform/runtime/buildplan/changeset.go
```diff
@@ -32,3 +32,46 @@ func NewBaseArtifactChangesetByBuildPlan(build *model.Build, requestedOnly, buil
...
+// IsDependency iterates over all artifacts and their dependencies in the given changeset, and
+// returns whether or not the given artifact is a dependency of any of those artifacts or
+// dependencies.
+func IsDependency(a artifact.ArtifactID, changeset artifact.ArtifactChangeset, artifacts artifact.Map) bool {
+	for _, artifact := range changeset.Added {
+		if artifact.ArtifactID == a {
+			continue
+		}
+
+		for _, depId := range RecursiveDependenciesFor(artifact.ArtifactID, artifacts) {
```
MDrakos, line 56: I know this isn't related to your PR but I took a look at the `RecursiveDependenciesFor` function and it is only grabbing direct dependencies, not dependencies of the steps that generated the given artifact. Is this something we care to include?

pkg/platform/runtime/setup/setup.go
```diff
@@ -171,26 +171,36 @@ func New(target Targeter, eventHandler events.Handler, auth *authentication.Auth
...
-			panic(r) // We're just logging the panic while we have context, we're not meant to handle it here
-		}
+		s.solveUpdateRecover(recover())
+	}()
+
+	if s.target.InstallFromDir() != nil {
+		return nil, nil
+	}
+
+	// Request build
```
MDrakos, line 183:
```suggestion
```

pkg/platform/runtime/setup/setup.go
```diff
@@ -230,14 +240,35 @@ func (s *Setup) Update() (rerr error) {
...
 }
 
-func (s *Setup) updateArtifacts() ([]artifact.ArtifactID, error) {
+// Panics are serious, and reproducing them in the runtime package is HARD. To help with this we dump
+// the build plan when a panic occurs so we have something more to go on.
+func (s *Setup) solveUpdateRecover(r interface{}) {
+	if r == nil {
+		return
+	}
+	buildplan, err := s.store.BuildPlanRaw()
```
MDrakos, line 249: Right now, we are storing the build plan fairly late in the setup process (see [here](https://github.com/ActiveState/cli/blob/master/pkg/platform/runtime/setup/setup.go#L620)). If it's easy we should consider moving this up to happen earlier or file a follow up to do this in a different way. Regardless there is higher potential that we won't have the raw build plan from the `store` with the current implementation.

MDrakos (changes requested):

internal/runbits/dependencies/changesummary.go
```diff
@@ -25,12 +25,9 @@ const showUpdatedPackages = true
...
 	var addedId *artifact.ArtifactID
-	for _, candidateId := range changeset.Added {
-		if !isDependency(candidateId, changeset, artifacts) {
-			if addedId != nil {
-				return // more than two independent packages were added
-			}
-			foundId := candidateId
+	for _, candidate := range changeset.Added {
+		if !isDependency(candidate.ArtifactID, changeset, artifacts) {
+			foundId := candidate.ArtifactID
```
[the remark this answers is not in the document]
Naatan: I can't remember :\ I think I was probably debugging something. Will revert and see what happens with tests. Thanks for spotting.

internal/runbits/requirements/requirements.go
Naatan: This should've been `reqNameAndVersion = ...`, which also addresses your prior comment.

pkg/platform/runtime/buildplan/changeset.go
Naatan: Not sure I follow your meaning, it seems recursive? https://github.com/ActiveState/cli/blob/DX-2556/pkg/platform/runtime/buildplan/dependencies.go#L181-L181

pkg/platform/runtime/setup/setup.go
Naatan: Are you suggesting this as a general improvement, or you see a potential way for this PR to trip over this mechanic? If the former; that's a good idea - should file it on Jira! If the latter; can you elaborate on the issue you foresee?

pkg/platform/runtime/buildplan/changeset.go
MDrakos: `artf.Dependencies` is only the direct dependencies of the artifact. If we wanted the full closure we would recursively find the step that generated the artifact and include the dependences of that step.

MDrakos: That might be overkill for what we want though.

pkg/platform/runtime/setup/setup.go
MDrakos: Definitely a general improvement. Right now, depending on where the setup process failed we may not have the raw build plan because we are relying on the store. We could store it earlier in the process or make it part of the build response so we're almost guaranteed to have it. I'll file a story in Jira.

internal/runbits/dependencies/changesummary_test.go
```diff
@@ -11,49 +11,78 @@ import (
...
-	artifact8 := artifact.Artifact{
-		Name: "Package 2", Version: ptr.To("1.1"), ArtifactID: "8",
-		Dependencies: []artifact.ArtifactID{"9", "3"}}
-	artifact9 := artifact.Artifact{
-		Name: "Dependency 1", Version: ptr.To("2.1"), ArtifactID: "9",
-		Dependencies: []artifact.ArtifactID{"10", "5"}}
-	artifact10 := artifact.Artifact{
-		Name: "Common recursive dependency", Version: ptr.To("4.1"), ArtifactID: "10",
-		Dependencies: nil}
+	a1DependsOn2n3 := artifact.Artifact{
```
MDrakos, line 14: I went cross-eyed reading this test.

MDrakos (approved):
Looks good! The improvements to the runtime package are great!

internal/runbits/dependencies/changesummary_test.go
Naatan: lmao.. yeah I we cross-eyed trying to understand it. Before it just said `artifact1` and I had a hard time grokking from that how the test was functioning.

PR #3190, opened 2024-03-21
Added partial version detection for `state install`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2648" title="DX-2648" target="_blank">TaskDX-2648</a>  `state install` supports wildcard versions
  </td></table>
  <br />
 

Similar to `state init`.

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -2,6 +2,7 @@ package requirements
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


mitchell-as:
Test failure is one of the usual timeouts and not caused by this PR.

internal/runbits/requirements/requirements.go
```diff
@@ -218,6 +222,18 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 
 		requirementName = normalized
 
+		// If a bare version number was given, and if it is a partial version number (e.g. requests@2),
+		// we'll want to ultimately append a '.x' suffix.
+		if versionRe.MatchString(requirementVersion) {
```
MDrakos, line 227: nit: Any appetite to centralizing this logic that is shared between `init` and `install` now?

MDrakos (approved):
Nice!

internal/runbits/requirements/requirements.go
mitchell-as: No, they don't actually share much of the same code. One queries for a language, the other for a package/bundle.

PR #3192, opened 2024-03-22
Added ability to set default timestamp in buildexpression if it matches a commit time.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2611" title="DX-2611" target="_blank">BugDX-2611</a>  Buildscript conflicts should diff `at_time` properly
  </td></table>
  <br />
 

This is helpful for merging buildscripts involving older buildexpressions with non-variable at_time values.

### DIFF
--- a/pkg/platform/runtime/buildexpression/buildexpression.go
+++ b/pkg/platform/runtime/buildexpression/buildexpression.go
@@ -906,6 +906,31 @@ func (e *BuildExpression) SetDefaultTimestamp() error {
 	return nil
 }
 
+// MaybeSetDefaultTimestamp changes the solve node's "at_time" value to "$at_time" if and only if
+// the current value is the given timestamp.
+// Buildscripts prefer to use variables for at_time and define them outside the buildexpression as
+// the expression's commit time.
+// While modern buildexpressions use variables, older ones bake in the commit time. This function
+// exists primarily to update those older buildexpressions for use in buildscripts.
+func (e *BuildExpression) MaybeSetDefaultTimestamp(ts *strfmt.DateTime) error {
+	if ts == nil {
+		return nil // nothing to compare to
+	}
+	atTimeNode, err := e.getSolveAtTimeValue()
+	if err != nil {
+		return errs.Wrap(err, "Could not get %s node", AtTimeKey)
+	}
+	if strings.HasPrefix(*atTimeNode.Str, "$") {
+		return nil
+	}
+	if atTime, err := strfmt.ParseDateTime(*atTimeNode.Str); err == nil && atTime == *ts {
+		return e.SetDefaultTimestamp()
+	} else if err != nil {
+		return errs.Wrap(err, "Invalid timestamp: %s", *atTimeNode.Str)
+	}
+	return nil
+}
+
 // normalizeTimestamp normalizes the solve node's timestamp, if possible.
 // Platform timestamps may differ from the strfmt.DateTime format. For example, Platform
 // timestamps will have microsecond precision, while strfmt.DateTime will only have millisecond
--- a/pkg/platform/runtime/buildscript/buildscript.go
+++ b/pkg/platform/runtime/buildscript/buildscript.go
@@ -110,6 +110,12 @@ func NewFromCommit(atTime *strfmt.DateTime, expr *buildexpression.BuildExpressio
 		return nil, errs.Wrap(err, "Could not copy build expression")
 	}
 
+	// Update old expressions that bake in at_time as a timestamp instead of as a variable.
+	err = expr.MaybeSetDefaultTimestamp(atTime)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not set default timestamp")
+	}
+
 	return &Script{AtTime: atTime, Expr: expr}, nil
 }
 


pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -906,6 +906,31 @@ func (e *BuildExpression) SetDefaultTimestamp() error {
...
+	if ts == nil {
+		return nil // nothing to compare to
+	}
+	atTimeNode, err := e.getSolveAtTimeValue()
+	if err != nil {
+		return errs.Wrap(err, "Could not get %s node", AtTimeKey)
+	}
+	if strings.HasPrefix(*atTimeNode.Str, "$") {
+		return nil
+	}
```
MDrakos, line 925: Why are we returning in this case? Can we add a comment to explain why.

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -906,6 +906,31 @@ func (e *BuildExpression) SetDefaultTimestamp() error {
...
+	if atTime, err := strfmt.ParseDateTime(*atTimeNode.Str); err == nil && atTime == *ts {
```
MDrakos, line 926: I'm probably missing something here but what about the case where `err == nil` and `atTime != *ts`? Do we want to do nothing?

pkg/platform/runtime/buildexpression/buildexpression.go
```diff
@@ -906,6 +906,31 @@ func (e *BuildExpression) SetDefaultTimestamp() error {
 	return nil
 }
 
+// MaybeSetDefaultTimestamp changes the solve node's "at_time" value to "$at_time" if and only if
+// the current value is the given timestamp.
+// Buildscripts prefer to use variables for at_time and define them outside the buildexpression as
+// the expression's commit time.
+// While modern buildexpressions use variables, older ones bake in the commit time. This function
+// exists primarily to update those older buildexpressions for use in buildscripts.
+func (e *BuildExpression) MaybeSetDefaultTimestamp(ts *strfmt.DateTime) error {
```
MDrakos, line 915: nit: the prefix maybe got me worried we were introducing some freak functional stuff. We could rename the function or have a function call for a compatibility check followed by `e.setDefaultTimeStamp` in the calling code which I believe is more in line with our codebase. Up to you though.

MDrakos (changes requested):
Makes sense to me, just a few minor comments.

pkg/platform/runtime/buildexpression/buildexpression.go
mitchell-as: This is covered in the doc comment for the function:

```
MaybeSetDefaultTimestamp changes the solve node's "at_time" value to "$at_time" if and only if
the current value is the given timestamp.
```

If the node has a "$" prefix (i.e. it is "$at_time"), then we don't change it. (It's already a default timestamp.)

pkg/platform/runtime/buildexpression/buildexpression.go
mitchell-as: Yes, as stated in the doc comment, we only change the timestamp to default if it matches the given one. If it doesn't match, we leave it alone because the user has likely explicitly set that timestamp to be something other than the default at_time.

pkg/platform/runtime/buildexpression/buildexpression.go
mitchell-as: By name it might seem odd, but I believe the doc comment explains nicely what the function is for, and what it does. If there's something not clear, please let me know. Otherwise, please suggest a new name or an example of a compatibility check followed by `e.setDefaultTimeStamp`, because I don't understand what you mean :(

MDrakos: I just meant something like

```
be := NewBuildExpression(bytes)
if be.IsIncompatibleTimeStamp(ts) {
   err = be.UpdateTimestamp()
   ...
}
```

But I'm not married to the idea, just a suggestion as it's a pattern I've seen in our codebase before.

pkg/platform/runtime/buildexpression/buildexpression.go
MDrakos: > If the node has a "$" prefix (i.e. it is "$at_time"), then we don't change it. (It's already a default timestamp.)

While this part isn't specifically in the docstring, reading it and seeing this makes way more sense now. Thanks!

MDrakos (approved):

pkg/platform/runtime/buildexpression/buildexpression.go
mitchell-as: I see what you mean. I experimented with something like that before. My first inclination was to add a `func (e *BuildExpression) GetTimestamp() (*strfmt.DateTime, error)` method, but since "at_time" can be "$at_time", the function couldn't always return a timestamp, even if the "at_time" field had a valid value.

Then there was the idea of adding `func (e *BuildExpression) TimestampIs(other ts *strfmt.DateTime) (bool, err)`, but it would result in something quite complicated like:

```
if setDefault, err := expr.TimestampIs(other); err == nil && setDefault {
  err := expr.SetDefaultTimestamp()
  if err != nil {
    ...
  }
} else if err != nil {
  ...
}
```

As you can see, a simple

```
err := expr.MaybeSetDefaultTimestamp(other)
if err != nil {
  ...
}
```

seemed better to me :)

PR #3194, opened 2024-03-22
Quote `state exec` arguments on Windows.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2615" title="DX-2615" target="_blank">BugDX-2615</a>  state exec does not always pass arguments correctly
  </td></table>
  <br />


mitchell-as:
This appears to be solved by #3389 .

PR #3196, opened 2024-03-25
Rename builds to artifacts

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2663" title="DX-2663" target="_blank">StoryDX-2663</a>  Rename `state builds` to `state artifacts`, keep `state builds` as an alias
  </td></table>
  <br />


Naatan (changes requested):
Looks good! But can you please fix the rename so we don't loose history:

image

PR #3197, opened 2024-03-25
Ensure dependencies are reset if there is user input

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2655" title="DX-2655" target="_blank">BugDX-2655</a>  `state publish` failing on version constraint on master
  </td></table>
  <br />
 

For whatever reason this line existed on v0.43, but not on v0.44:

https://github.com/ActiveState/cli/blob/e85cd5d8b64191f27207396d4cc4c96c60488b5e/internal/runners/publish/publish.go#L326-L326

Likely a bad merge. This brings it back into v0.44, as well as considers the other ways dependencies can be specified.

### DIFF
--- a/internal/runners/publish/publish.go
+++ b/internal/runners/publish/publish.go
@@ -195,15 +195,6 @@ func (r *Runner) Run(params *Params) error {
 		}
 	}
 
-	// Validate user input
-	if params.Edit {
-		// Description is not currently supported for edit
-		// https://activestatef.atlassian.net/browse/DX-1886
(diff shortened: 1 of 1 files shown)


mitchell-as (changes requested):
I think you started on the wrong branch? Too many changes.

Naatan:
@mitchell-as my bad, I needed to target v0.44, because v0.43 is fine as is.

mitchell-as (approved):

PR #3199, opened 2024-03-26
Transitioning to buildscript is handled explicitly

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2619" title="DX-2619" target="_blank">TaskDX-2619</a>  Transitioning to buildscripts / build expressions is handled explicitly
  </td></table>
  <br />
 

This is addressing several things:

- Buildscripts are created explicitly; either because the project config needs migrating, or by running `state reset`.
- Project config now has a migration mechanic
- Runtime will error out if buildscripts are opted in but no buildscripts exist; you'll get an error informing you to run `state reset LOCAL`
- Runtime runbits were refactored to ensure this new buildscript error is properly propagated in all scenarios


mitchell-as:
If I opt into buildscripts and then attempt to checkout a project, it will immediately fail:

```
% state checkout ActiveState-CLI/Merge#447b8363-024c-4143-bf4e-c96989314fdf
Checking out project: ActiveState-CLI/Merge

█ Installing Runtime

 x Your project is missing its buildscript file, please run 'state reset LOCAL' to recreate it.
```

pkg/platform/api/buildplanner/model/buildplan.go
```diff
@@ -266,39 +282,51 @@ func (b *BuildPlanByProject) CommitID() (strfmt.UUID, error) {
 }
 
 type BuildPlanByCommit struct {
-	Commit *Commit `json:"commit"`
+	commit *Commit `json:"commit"`
```
MDrakos, line 285: I believe that by making this field private the `encoding/json` package will not be able to access it so marhsalling/unmarshalling may fail. Maybe that's changed but I remember getting caught by this before.

pkg/projectfile/projectfile.go
```diff
@@ -453,6 +466,29 @@ func Parse(configFilepath string) (_ *Project, rerr error) {
...
+	// Migrate project file if needed
+	if !migrationRunning && project.ConfigVersion != ConfigVersion && migrator != nil {
+		// Migrations may themselves utilize the projectfile package, so we have to ensure we don't start an infinite loop
+		migrationRunning = true
+		defer func() { migrationRunning = false }()
+
+		if project.ConfigVersion > ConfigVersion {
+			return nil, locale.NewInputError("err_projectfile_version_too_high")
+		}
+		updatedConfigVersion, errMigrate := migrator(project, ConfigVersion)
```
MDrakos, line 478: We will lose this error if anything on line 483 fails. We may want to log it in that case if this could be valuable.

pkg/projectfile/yamlfield.go
```diff
@@ -0,0 +1,50 @@
...
+type yamlField struct {
+	field string
+	value interface{}
+}
+
+func NewYamlField(field string, value interface{}) *yamlField {
+	return &yamlField{field: field, value: value}
+}
+
+func (y *yamlField) Save(path string) error {
```
MDrakos, line 22: nit: this feels low level enough to have a unit test.

MDrakos (changes requested):
This is a large PR so I tried to provide some higher level feedback. Overall the migration implementation looks good to me and makes sense.

internal/runbits/requirements/requirements.go
```diff
@@ -310,31 +311,40 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
 	pg.Stop(locale.T("progress_success"))
 	pg = nil
 
-	// Solve runtime
-	rt, buildResult, changedArtifacts, err := r.solve(commitID, ns)
-	if err != nil {
-		return errs.Wrap(err, "Could not solve runtime")
-	}
+	if strings.ToLower(os.Getenv(constants.DisableRuntime)) != "true" {
```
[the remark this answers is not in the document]
Naatan: Yeah agreed, but I don't see a way to address this without significant refactoring, which is what we're already planning on, just not for this story.

internal/runbits/runtime/rationalize.go
```diff
@@ -72,13 +67,34 @@ func rationalizeError(auth *authentication.Auth, proj *project.Project, rerr *er
...
+		errMsg := locale.Tr("err_build_artifact_failed_msg", artifactErr.Artifact.DisplayName)
+		*rerr = errs.WrapUserFacing(*rerr, locale.Tr("err_build_artifact_failed", errMsg,
+			strings.Join(artifactErr.Artifact.Errors, "\n"), artifactErr.Artifact.LogURL))
+
 	// If updating failed due to unidentified errors, and the user is not authenticated, add a tip suggesting that they authenticate as
 	// this may be a private project.
 	// Note since we cannot assert the actual error type we do not wrap this as user-facing, as we do not know what we're
 	// dealing with so the localized underlying errors are more appropriate.
-	case isUpdateErr && !auth.Authenticated():
+	case !auth.Authenticated(): // MUST BE LAST
```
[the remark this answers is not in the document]
Naatan: Because this is not asserting an error, it's decorating an unknown error. And all the other assertions are for asserting an error.

I've updated the code with a comment and some protection against accidental refactoring.

Note the original implementation was bugged:

https://github.com/ActiveState/cli/pull/3199/files/3c9e66d2268323d48391a6b19658bd20b4a7e327#diff-ad4a4ce35107eceb3ee56dfede324f0f7eb85af84005dbd6dc158f5681e58c92L28

It was ignoring errors if those non-error assertions were evaluated. We really should void using rationalizers to do anything that's not error assertion related.

internal/runbits/runtime/refresh.go
```diff
@@ -2,45 +2,67 @@ package runtime
...
+	"github.com/imacks/bitflags-go"
 )
 
-// RefreshRuntime should be called after runtime mutations.
-func RefreshRuntime(
+type Opts int
+
+const (
+	OptNone Opts = 1 << iota
+	OptMinimalUI
```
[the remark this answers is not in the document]
Naatan: I'll add comments. I have no intention of using this outside of this package, or even beyond the runtime refactor. This is effectively a temporary workaround to get around the increasingly lacking architecture.

internal/runners/reset/reset.go
```diff
@@ -96,27 +107,32 @@ func (r *Reset) Run(params *Params) error {
...
 
 	err = localcommit.Set(r.project.Dir(), commitID.String())
 	if err != nil {
 		return errs.Wrap(err, "Unable to set local commit")
 	}
 
-	err = runtime.RefreshRuntime(r.auth, r.out, r.analytics, r.project, commitID, true, target.TriggerReset, r.svcModel, r.cfg)
+	// Ensure the buildscript exists. Normally we should never do this, but reset is used for resetting from a corrupted
+	// state, so it is appropriate.
+	if r.cfg.GetBool(constants.OptinBuildscriptsConfig) {
```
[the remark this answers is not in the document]
Naatan: It's not really the same thing, the runtime instance is still getting called regardless of the env var, it just behaves differently. In this case we're not doing anything with the buildscript outside of this one call. Add to that the fact that we have some commands like commit and eval that do not care about this config, and it just makes more sense to have it in the runner.

pkg/platform/api/buildplanner/model/buildplan.go
Naatan: Urgh you're right, thanks for catching!

pkg/projectfile/projectfile.go
Naatan: Good catch! Perfect job for `errs.Pack()`.

internal/runbits/requirements/requirements.go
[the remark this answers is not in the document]
MDrakos: That's fair, I didn't want to tag this as anything other than a nit because I know we have the refactor coming.

internal/runbits/runtime/rationalize.go
```diff
@@ -72,13 +67,39 @@ func rationalizeError(auth *authentication.Auth, proj *project.Project, rerr *er
...
 	// If updating failed due to unidentified errors, and the user is not authenticated, add a tip suggesting that they authenticate as
 	// this may be a private project.
 	// Note since we cannot assert the actual error type we do not wrap this as user-facing, as we do not know what we're
 	// dealing with so the localized underlying errors are more appropriate.
-	case isUpdateErr && !auth.Authenticated():
-		*rerr = errs.AddTips(*rerr,
-			locale.T("tip_private_project_auth"),
-		)
+	default:
+		// Add authentication tip if we could not asser the error type
```
MDrakos, line 95:
```suggestion
		// Add authentication tip if we could not assert the error type
```

pkg/platform/runtime/setup/buildlog/buildlog.go
```diff
@@ -216,6 +217,11 @@ func NewWithCustomConnections(artifactMap artifact.Map,
 					break
 				}
 
+				if _, ok := artifactsDone[m.ArtifactID]; ok {
+					multilog.Critical("buildlogstreamer sent duplicate ArtifactSucceededMessage for artifact %s", m.ArtifactID.String())
+					break
+				}
```
MDrakos, line 223: Is this meant to be a part of another story?

MDrakos (dismissed):
Looks good to me! Just left a couple of minor comments, not meant to hold up the PR.

pkg/platform/runtime/setup/buildlog/buildlog.go
Naatan: Huh, yeah.. I had run `git reset`, not sure why this is still in :\

MDrakos (approved):

PR #3205, opened 2024-04-01
Use locale errors for flag errors.

mitchell-as:


internal/captain/values.go
```diff
@@ -194,7 +194,7 @@ func (p *PackageValueNSRequired) Set(s string) error {
 		return errs.Wrap(err, "PackageValueNSRequired.Set failed")
 	}
 	if p.Namespace == "" {
-		return fmt.Errorf("invalid package name format: %s (expected '<namespace>/<name>[@version]')", s)
+		return locale.NewInputError("err_package_value_ns_required", "invalid package name format: %s (expected '<namespace>/<name>[@version]')", s)
```
Naatan, line 197:
```suggestion
		return locale.NewInputError("err_package_value_ns_required", "invalid package name format: {{.V0}} (expected '<namespace>/<name>[@version]')", s)
```

Naatan (changes requested):

internal/captain/values.go
mitchell-as: 🤦 thanks

PR #3206, opened 2024-04-01
Correctly save to activestate.yaml the branch name for the given commit ID.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2658" title="DX-2658" target="_blank">BugDX-2658</a>  Activate by commitID from different then main branch saved incorrectly in as.yaml file
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/checkout/checkout.go
+++ b/internal/runbits/checkout/checkout.go
@@ -43,6 +43,11 @@ type Checkout struct {
 	auth       *authentication.Auth
 }
 
+type errCommitDoesNotBelong struct {
+	error
+	CommitID strfmt.UUID
+}
+
 func New(repo git.Repository, prime primeable) *Checkout {
 	return &Checkout{repo, prime.Output(), prime.Config(), prime.Analytics(), "", prime.Auth()}
 }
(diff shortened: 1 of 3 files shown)


internal/runbits/checkout/checkout.go
```diff
@@ -68,35 +73,38 @@ func (r *Checkout) Run(ns *project.Namespaced, branchName, cachePath, targetPath
...
-		branch, err = model.DefaultBranchForProject(pj)
-		if err != nil {
-			return "", errs.Wrap(err, "Could not grab branch for project")
+	commitID := ns.CommitID
+
+	switch {
+	// Fetch the branch the given commitID is on.
+	case commitID != nil:
+		for _, b := range pj.Branches {
+			if belongs, err := model.CommitBelongsToBranch(ns.Owner, ns.Project, b.Label, *commitID, r.auth); err == nil && belongs {
```
MDrakos, line 82: It's unfortunate the number of API calls we're going to have to make in order to determine this. I did some poking around the VCS API and didn't see anything. It _may_ be worth asking if something exists for this use case, but I doubt it so I'll leave it up to you.

MDrakos (approved):
Looks good! Just a thought in the comments.

PR #3208, opened 2024-04-01
Use `state revert REMOTE` instead of `state revert HEAD`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2674" title="DX-2674" target="_blank">TaskDX-2674</a>  As a user I can use intuitive terms to reference the local / remote project
  </td></table>
  <br />
 

Convert from HEAD to REMOTE, but show a deprecation message.

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1140,7 +1140,7 @@ err_user_libc_solution:
   other: If you specified a glibc version, please check it is configured → [ACTIONABLE]{{.V0}}[/RESET].
 warn_revert_head:
   other: |
-    [NOTICE]DEPRECATION WARNING[/RESET]: 'state revert HEAD' has been deprecated in favor of '[ACTIONABLE]state revert REMOTE[/RESET]'. Please use '[ACTIONABLE]state revert REMOTE[/RESET]' instead.
+    [NOTICE]DEPRECATION WARNING[/RESET]: '[ACTIONABLE]state revert HEAD[/RESET]' has been deprecated in favor of '[ACTIONABLE]state revert REMOTE[/RESET]'. Please use '[ACTIONABLE]state revert REMOTE[/RESET]' instead.
 err_revert_get_commit:
   other: "Could not fetch commit details for commit with ID: {{.V0}}"
 err_revert_commit:
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1140,7 +1140,7 @@ err_user_libc_solution:
   other: If you specified a glibc version, please check it is configured → [ACTIONABLE]{{.V0}}[/RESET].
 warn_revert_head:
   other: |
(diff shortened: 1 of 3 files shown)


internal/locale/locales/en-us.yaml
```diff
@@ -1138,6 +1138,9 @@ err_user_network_solution:
     If your issue persists consider reporting it on our forums at {{.V0}}.
 err_user_libc_solution:
   other: If you specified a glibc version, please check it is configured → [ACTIONABLE]{{.V0}}[/RESET].
+warn_revert_head:
+  other: |
+    [NOTICE]DEPRECATION WARNING[/RESET]: `state revert HEAD` has been deprecated in favor of `[ACTIONABLE]state revert REMOTE[/RESET]`. Please use `[ACTIONABLE]state revert REMOTE[/RESET]` instead.
```
mitchell-as, line 1143: The ticket did not give me a message I should be using, so let me know if this should be changed to something else.

Naatan, line 1143:
```suggestion
    [NOTICE]DEPRECATION WARNING[/RESET]: 'state revert HEAD' has been deprecated in favor of '[ACTIONABLE]state revert REMOTE[/RESET]'. Please use '[ACTIONABLE]state revert REMOTE[/RESET]' instead.
```

Naatan (changes requested):

internal/locale/locales/en-us.yaml
```diff
@@ -1138,6 +1138,9 @@ err_user_network_solution:
     If your issue persists consider reporting it on our forums at {{.V0}}.
 err_user_libc_solution:
   other: If you specified a glibc version, please check it is configured → [ACTIONABLE]{{.V0}}[/RESET].
+warn_revert_head:
+  other: |
+    [NOTICE]DEPRECATION WARNING[/RESET]: 'state revert HEAD' has been deprecated in favor of '[ACTIONABLE]state revert REMOTE[/RESET]'. Please use '[ACTIONABLE]state revert REMOTE[/RESET]' instead.
```
Naatan, line 1143:
```suggestion
    [NOTICE]DEPRECATION WARNING[/RESET]: '[ACTIONABLE]state revert HEAD[/RESET]' has been deprecated in favor of '[ACTIONABLE]state revert REMOTE[/RESET]'. Please use '[ACTIONABLE]state revert REMOTE[/RESET]' instead.
```

Naatan (changes requested):
Sorry didn't notice first pass.

internal/locale/locales/en-us.yaml
mitchell-as: FYI I intentionally did that so as to draw more attention to what we want the user to use, rather than what they just used. I will defer to you though.

Naatan: That's a good idea, I like it! My OCD was triggering at the inconsistency but there's a good reason for it so that settles it 🙏

Naatan (dismissed):

internal/locale/locales/en-us.yaml
mitchell-as: Okay, I've reverted the commit with your suggested changes. Please re-approve @Naatan .

Naatan (approved):

PR #3209, opened 2024-04-02
Enable wildcards for `state languages install`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2626" title="DX-2626" target="_blank">StoryDX-2626</a>  As a user I can use wildcard versions when running `state languages install`
  </td></table>
  <br />
 

This PR changes `state languages install` to piggy-back more off of the existing `requirements` package's capabilities for installing/updating packages by treating the language to be installed as just a package. It will inherit all the usual behaviors of `state install`, including `.x` wildcard support and silently appending wildcards for partial version numbers.

### DIFF
--- a/test/integration/languages_int_test.go
+++ b/test/integration/languages_int_test.go
@@ -57,13 +57,8 @@ func (suite *LanguagesIntegrationTestSuite) TestLanguages_install() {
 	cp.Expect("python")
 	cp.ExpectExitCode(0)
 
-	cp = ts.Spawn("languages", "install", "python")
-	cp.Expect("Language: python is already installed")
-	cp.ExpectExitCode(1)
-	ts.IgnoreLogErrors()
-
 	cp = ts.Spawn("languages", "install", "python@3.9.16")
-	cp.Expect("Language added: python@3.9.16")
+	cp.Expect("Language updated: python@3.9.16")
 	// This can take a little while
 	cp.ExpectExitCode(0, termtest.OptExpectTimeout(60*time.Second))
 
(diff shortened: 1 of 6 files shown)


test/integration/languages_int_test.go
```diff
@@ -57,13 +57,8 @@ func (suite *LanguagesIntegrationTestSuite) TestLanguages_install() {
 	cp.Expect("python")
 	cp.ExpectExitCode(0)
 
-	cp = ts.Spawn("languages", "install", "python")
-	cp.Expect("Language: python is already installed")
-	cp.ExpectExitCode(1)
-	ts.IgnoreLogErrors()
```
MDrakos, line 63: Isn't this still a valid test?

MDrakos (changes requested):

test/integration/languages_int_test.go
mitchell-as: I didn't think so because we are now fully utilizing the buildplanner, and it'll tell us if there's nothing new in the commit (i.e. the package/language is already installed), rather than us manually fetching the project's current language and doing a comparison in the runner.

MDrakos: Ahh okay, that makes sense. I still think we could have an expect here just to ensure that this command fails sensibly, but I'll leave that up to you.

MDrakos (approved):
Nice, glad we could get rid of some of those functions in the language runner.

test/integration/languages_int_test.go
mitchell-as: In this case, it will update Python because the checkout version is old. You actually have to run the install command twice before it will fail with "already" installed, and I'm concerned that it'll just balloon the integration test run time. Also, in my testing of this, macOS tended to have more build failures with Python 3.11, so I'd rather just defer to the existing `TestPackage_Duplicate()` to raise an error like this.

PR #3210, opened 2024-04-02
Handle duplicate artifact events at the source

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2697" title="DX-2697" target="_blank">TaskDX-2697</a>  Failed to add or update artifact bar: Artifact bar already exists
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/buildlog/buildlog.go
+++ b/pkg/platform/runtime/setup/buildlog/buildlog.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/pkg/platform/model"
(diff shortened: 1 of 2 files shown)


Naatan:
@mdrohmann flagged you for review since you know this API and can weigh in on whether this is appropriate.

If the API is expected to send the same event twice in certain scenario's we should not be logging this as critical (which means it goes to rollbar).

mdrohmann (commented):
It is currently possible for the buildlogstreamer to send the same event twice.  This happens, when nomad looses track of a build job and the HC re-schedules it.  In this case, the job _might_ actually be run twice.  It appears that the amount of times, we loose track of a build, has increased since we are using nomad, but it always was a possible state.

MDrakos (approved):

PR #3211, opened 2024-04-02
`state install` handles multiple requirements

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2525" title="DX-2525" target="_blank">StoryDX-2525</a>  state install and uninstall should support multiple packages
  </td></table>
  <br />
 

This PR updates the `executeRequirementOperation` to take a variadic `Requirements` object as an argument. To that end, a lot of the work in this PR was moving chunks of the original code to helper functions and having those functions work with the same variadic arguments.

There were definitely opportunities for a larger refactor here but I tried to refrain from changing the logic too much as we have follow up stories to the runtime refactor for cleaning this code up.


MDrakos:

internal/runbits/requirements/requirements.go
```diff
@@ -351,31 +268,198 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
+			continue
+		}
+		requirementsToValidate = append(requirementsToValidate, requirement)
+	}
+
+	if len(requirementsToValidate) == 0 {
+		return nil
+	}
+
+	pg := output.StartSpinner(r.Output, locale.Tr("progress_search", strings.Join(requirementNames(requirementsToValidate...), ", ")), constants.TerminalAnimationInterval)
```
mitchell-as, line 356: I wonder if it would be better to have individual progress bars for each requirement, so if there's a failure, it's easy for the user to see which one failed.

internal/runbits/requirements/requirements.go
```diff
@@ -679,14 +801,19 @@ func getSuggestions(ns model.Namespace, name string, auth *authentication.Auth)
...
+		case model.NamespacePlatform:
+			return platformCommitMessage(req.Operation, req.Name, req.Version, req.BitWidth)
+		case model.NamespacePackage, model.NamespaceBundle:
+			return packageCommitMessage(req.Operation, req.Name, req.Version)
+		}
+	} else {
+		return commitMessageMultiple(requirements...)
 	}
 
 	return ""
```
mitchell-as, line 819: nit: Maybe consider making this a

```
switch len(requirements) {
case 0:
  return ""
case 1:
  ...
  return packageCommitMessage(...)
default:
  return commitMessageMultiple(...)
```
?

internal/runbits/requirements/requirements.go
```diff
@@ -736,3 +863,16 @@ func packageCommitMessage(op bpModel.Operation, name, version string) string {
 	}
 	return locale.Tr(msgL10nKey, name, version)
 }
+
+func commitMessageMultiple(requirements ...*Requirement) string {
+	// TODO: Replace this placeholder with a proper message
```
mitchell-as, line 868: Do we want to file a follow-up story for this? Or is this something you still need to do for this PR?

internal/runners/languages/install.go
```diff
@@ -54,15 +54,12 @@ func (u *Update) Run(params *UpdateParams) error {
...
-		bpModel.OperationAdded,
-		nil,
-		&model.NamespaceLanguage,
-		nil)
+	return op.ExecuteRequirementOperation(nil, &requirements.Requirement{
+		Name:          lang.Name,
+		Version:       lang.Version,
+		NamespaceType: &model.NamespaceLanguage,
+		Operation:     bpModel.OperationAdded,
+	})
```
mitchell-as, line 62: Oooo, I like this very much!

internal/runners/packages/uninstall.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // UninstallRunParams tracks the info required for running Uninstall.
 type UninstallRunParams struct {
-	Package captain.PackageValueNoVersion
+	Packages captain.PackagesValue
```
mitchell-as, line 15: Was this an intentional change? Now users can use `state uninstall foo@x.y`, right?

internal/runners/packages/uninstall.go
```diff
@@ -33,23 +33,22 @@ func (u *Uninstall) Run(params UninstallRunParams, nsType model.NamespaceType) (
...
-
-	if params.Package.Namespace != "" {
-		ns = ptr.To(model.NewRawNamespace(params.Package.Namespace))
-	} else {
-		nsTypeV = &nsType
+	var reqs []*requirements.Requirement
+	for _, p := range params.Packages {
+		req := &requirements.Requirement{
+			Name:      p.Name,
+			Version:   p.Version,
```
mitchell-as, line 40: As a follow up to the previous comment, this used to be "".

mitchell-as (changes requested):
This looks pretty good overall, and is the next logical step for allowing multiple requirements. I think you got the major stuff down. There are some minor things left to address.

internal/runbits/requirements/requirements.go
```diff
@@ -134,109 +149,12 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
-				} else if strings.HasPrefix(knownVersion.Version, requirementVersion) {
-					appendVersionWildcard = true
-				}
-			}
-		}
-
-		pg.Stop(locale.T("progress_found"))
-		pg = nil
+	if err := r.validatePackages(requirements...); err != nil {
+		return locale.WrapError(err, "err_validate_packages", "Could not validate one or more packages")
```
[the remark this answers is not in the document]
MDrakos: This one already has an underlying user-facing error for when the ingredient search fails. I will update this to be an errs.Wrap at this line though.

internal/runbits/requirements/requirements.go
```diff
@@ -247,54 +165,53 @@ func (r *RequirementOperation) ExecuteRequirementOperation(
...
-	if operation == bpModel.OperationAdded && parentCommitID != "" {
-		req, err := model.GetRequirement(parentCommitID, *ns, requirementName, r.Auth)
-		if err != nil {
-			return errs.Wrap(err, "Could not get requirement")
-		}
-		if req != nil {
-			operation = bpModel.OperationUpdated
-		}
+	if err := r.checkForUpdate(parentCommitID, requirements...); err != nil {
+		return locale.WrapError(err, "err_check_for_update", "Could not check for requirements updates")
```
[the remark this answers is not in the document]
MDrakos: The underlying failure here will occur if we cannot fetch the checkpoint, which the user can't really do anything about.

internal/runbits/requirements/requirements.go
MDrakos: Initially I had that but it doesn't work with the mock

internal/runners/packages/uninstall.go
MDrakos: Ah right, didn't realize uninstall has a different version. I'll add a new value type for it.

internal/runbits/requirements/requirements.go
```diff
@@ -275,10 +280,18 @@ func (r *RequirementOperation) ExecuteRequirementOperation(ts *time.Time, requir
...
+	*locale.LocalizedError
+	Name string
+}
+
 func (r *RequirementOperation) resolveNamespaces(ts *time.Time, requirements ...*Requirement) error {
 	for _, requirement := range requirements {
 		if err := r.resolveNamespace(ts, requirement); err != nil {
-			return errs.Wrap(err, "Could not resolve namespace")
+			return &ResolveNamespaceError{
+				locale.WrapError(err, "err_resolve_namespace", "Could not resolve namespace for requirement: {{.Name}}"),
```
mitchell-as, line 292: I think you can simplify this to `&ResolveNamespaceError{Name: requirement.Name}` because you're not using this error message; you're constructing a new one. Regardless, I don't think we're supposed to use `locale.WrapError()` in new code anymore as long as rationalizers are present.

mitchell-as (changes requested):
Excellent. One last thing I think.

internal/runbits/requirements/requirements.go
MDrakos: Yeah, I wasn't quite sure about that. I'l still embedded the error interface as we need to satisfy that.

mitchell-as (approved):

PR #3212, opened 2024-04-02
Prompt the user if they wish to continue running the installer as administrator.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2647" title="DX-2647" target="_blank">StoryDX-2647</a>  We provide a warning to users when installing the state tool as administrator.
  </td></table>
  <br />
 

This is not recommended.

Note: this only works for Windows. Detecting whether or not the user is an admin or has admin privileges on macOS and Linux is not straightforward at all, if even possible: 

- https://stackoverflow.com/questions/27366298/check-if-application-is-running-as-administrator-in-golang
- https://serverfault.com/questions/568627/can-a-program-tell-it-is-being-run-under-sudo
- https://stackoverflow.com/questions/29733575/how-to-find-the-user-that-executed-a-program-as-root-using-golang

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -51,7 +51,11 @@ type Params struct {
 }
 
 func newParams() *Params {
-	return &Params{activate: &project.Namespaced{}, activateDefault: &project.Namespaced{}}
+	return &Params{
+		activate:        &project.Namespaced{},
+		activateDefault: &project.Namespaced{},
+		nonInteractive:  !term.IsTerminal(int(os.Stdin.Fd())),
+	}
 }
 
 func main() {
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -47,6 +47,7 @@ type Params struct {
 	activate        *project.Namespaced
 	activateDefault *project.Namespaced
 	showVersion     bool
+	nonInteractive  bool
 }
 
 func newParams() *Params {
@@ -109,7 +110,6 @@ func main() {
 		return
 	}
 
-	var garbageBool bool
 	var garbageString string
 
 	// We have old install one liners around that use `-activate` instead of `--activate`
@@ -189,8 +189,8 @@ func main() {
 				Name:  "version", // note: no shorthand because install.sh uses -v for selecting version
 				Value: &params.showVersion,
 			},
+			{Name: "non-interactive", Shorthand: "n", Hidden: true, Value: &params.nonInteractive}, // don't prompt
 			// The remaining flags are for backwards compatibility (ie. we don't want to error out when they're provided)
-			{Name: "nnn", Shorthand: "n", Hidden: true, Value: &garbageBool}, // don't prompt; useless cause we don't prompt anyway
 			{Name: "channel", Hidden: true, Value: &garbageString},
 			{Name: "bbb", Shorthand: "b", Hidden: true, Value: &garbageString},
 			{Name: "vvv", Shorthand: "v", Hidden: true, Value: &garbageString},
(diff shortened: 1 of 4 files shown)


cmd/state-installer/test/integration/installer_int_test.go
```diff
@@ -42,7 +42,7 @@ func (suite *InstallerIntegrationTestSuite) TestInstallFromLocalSource() {
 	// Run installer with source-path flag (ie. install from this local path)
 	cp := ts.SpawnCmdWithOpts(
 		suite.installerExe,
-		e2e.OptArgs(installationDir(ts)),
+		e2e.OptArgs(installationDir(ts), "-n"),
```
mitchell-as, line 45: Let me know if you think there's a better way of handling this than to just append "-n" to every invocation of "state-installer" in this test suite.

cmd/state-installer/installer.go
```diff
@@ -45,6 +48,22 @@ func NewInstaller(cfg *config.Instance, out output.Outputer, payloadPath string,
...
 
 func (i *Installer) Install() (rerr error) {
+	isAdmin, err := osutils.IsAdmin()
+	if err != nil {
+		return errs.Wrap(err, "Could not determine if running as Windows administrator")
+	}
+	if isAdmin {
+		prompter := prompt.New(!i.Params.nonInteractive, i.an)
+		defaultChoice := !prompter.IsInteractive() // default to no for interactive, yes for non-interactive
+		confirm, err := prompter.Confirm("", locale.T("installer_prompt_is_admin"), &defaultChoice)
```
Naatan, line 58:
```suggestion
	if isAdmin && !i.Params.nonInteractive {
		prompter := prompt.New(true, i.an)
		confirm, err := prompter.Confirm("", locale.T("installer_prompt_is_admin"), ptr.To(false))
```

Nit: I find it easier to understand what's happening this way.

cmd/state-installer/cmd.go
```diff
@@ -189,8 +189,8 @@ func main() {
 				Name:  "version", // note: no shorthand because install.sh uses -v for selecting version
 				Value: &params.showVersion,
 			},
+			{Name: "non-interactive", Shorthand: "n", Hidden: true, Value: &params.nonInteractive}, // don't prompt
```
Naatan, line 192: We should also auto-detect this if the user did not provide it. eg.

https://github.com/ActiveState/cli/blob/4a1a8585aa5c3b8805e26c4b274a043631b52ac2/cmd/state/output.go#L57-L57

Naatan (changes requested):

cmd/state-installer/cmd.go
```diff
@@ -47,10 +47,15 @@ type Params struct {
...
 	showVersion     bool
+	nonInteractive  bool
 }
 
 func newParams() *Params {
-	return &Params{activate: &project.Namespaced{}, activateDefault: &project.Namespaced{}}
+	return &Params{
+		activate:        &project.Namespaced{},
+		activateDefault: &project.Namespaced{},
+		nonInteractive:  !term.IsTerminal(int(os.Stdin.Fd())),
```
Naatan, line 57: Have you tested this? I can't recall if these values are preserved through captain. I imagine they should be if the user didn't supply the flag, but probably worth verifying.

cmd/state-installer/installer.go
```diff
@@ -45,6 +49,21 @@ func NewInstaller(cfg *config.Instance, out output.Outputer, payloadPath string,
...
+		return errs.Wrap(err, "Could not determine if running as Windows administrator")
+	}
+	if isAdmin && !i.Params.nonInteractive {
+		prompter := prompt.New(true, i.an)
+		confirm, err := prompter.Confirm("", locale.T("installer_prompt_is_admin"), ptr.To(false))
+		if err != nil {
+			return errs.Wrap(err, "Unable to confirm")
+		}
+		if !confirm {
+			return // halt installation
```
Naatan, line 63: I believe we normally return an input error in this scenario.

Naatan (changes requested):

cmd/state-installer/cmd.go
mitchell-as: Yes I have. They are preserved if unspecified.

Naatan (approved):

PR #3213, opened 2024-04-03
Show version constraints string in `state languages` instead of "Auto".

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2662" title="DX-2662" target="_blank">TaskDX-2662</a>  `state languages` should show larger/smaller than notation for version rather than auto
  </td></table>
  <br />


MDrakos (approved):
Simple and clean!

PR #3216, opened 2024-04-04
Unpack expect error messages

MDrakos:


internal/testhelpers/suite/suite.go
```diff
@@ -0,0 +1,50 @@
...
+type Suite struct {
+	testify.Suite
+}
+
+func (suite *Suite) NoError(err error, msgAndArgs ...interface{}) {
+	if err == nil {
+		return
+	}
+
+	errMsg := fmt.Sprintf("All error messages: %s", errs.JoinMessage(err))
```
MDrakos, line 25: Not sure about this string prefix.

MDrakos:
I wrote a test to verify that these overwritten functions are being called correctly. Unfortunately, I can keep the tests around as they are designed to fail.

internal/testhelpers/suite/suite.go
Naatan: I don't think it's necessary. Feel free to drop this and just use the result of JoinMessage.

internal/testhelpers/suite/suite.go
```diff
@@ -0,0 +1,50 @@
...
+	msgAndArgs = append(msgAndArgs, errMsg)
+	suite.Suite.NoError(err, msgAndArgs...)
```
Naatan, line 27: I think this will lead to the top level error being repeated. That's not a big deal, but it would be nice if we can avoid it. The actual NoError() function is really bare basic, I suggest just re-implementing it here and not calling out to testify's NoError.

internal/testhelpers/suite/suite.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/stretchr/testify/require"
+	testify "github.com/stretchr/testify/suite"
+)
+
+type tSuite interface {
```
Naatan, line 12: Might be worth adding some sort of method to identify this is OUR version of the suite. That would help avoid accidentally passing the wrong suite. Eg. `IsHelperSuite()`

internal/testhelpers/suite/suite.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (a *Assertions) NoError(err error, msgAndArgs ...interface{}) {
+	if err == nil {
+		return
+	}
+
+	errMsg := fmt.Sprintf("All error messages: %s", errs.JoinMessage(err))
+	msgAndArgs = append(msgAndArgs, errMsg)
+	a.Assertions.NoError(err, msgAndArgs...)
```
Naatan, line 45: Same as with the other one; should just implement the NoError logic here and avoid calling out to testify at all, so we don't have duplicated errors. And then you can probably share logic also.

Naatan (changes requested):

internal/testhelpers/suite/suite.go
MDrakos: I've reimplemented it but still modified it a bit as the testify implementation has interface checks and calls to `Helper()` methods. They are also returning a `bool` from their functions, which I believe we aren't using anywhere.

Naatan (approved):

PR #3223, opened 2024-04-05
Preserve `--ts now` from a previous commit when running `state uninstall`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2713" title="DX-2713" target="_blank">BugDX-2713</a>  Custom timestamp not preserved when running install
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/uninstall.go
+++ b/internal/runners/packages/uninstall.go
@@ -43,7 +43,7 @@ func (u *Uninstall) Run(params UninstallRunParams, nsType model.NamespaceType) (
 		nsTypeV = &nsType
 	}
 
-	ts, err := getTime(nil, u.prime.Auth(), u.prime.Project())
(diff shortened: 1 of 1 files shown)


mitchell-as:
Test failure is unrelated to this PR.

Naatan (changes requested):
This looks ok to me, but during stand Mark said that he could repro by running the second install without `--ts`. Were you unable to repro this yourself?

mitchell-as:
Yeah, I ran

1. `state publish`
2. `state install --ts now`
3. `state uninstall`
4. `state install`

Step 4 was failing because the atTime in step 3 was reverting to a platform-defined timestamp, rather than the latest inventory timestamp from step 2. With the fix, step 3 preserves the latest inventory timestamp, as does step 4, which finds the newly published package.

Naatan (approved):

PR #3227, opened 2024-04-08
Fix TestInUse integration test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2690" title="DX-2690" target="_blank">BugDX-2690</a>  Nightly failure: TestRuntimeIntegrationTestSuite/TestInUse
  </td></table>
  <br />
 

The old `ActiveState-CLI/Perl-5.32` project is not building and has a warning on the dashboard that the version of Perl being used is no longer supported. I made a new project for this test. Over time if we find integration tests with `Perl-5.32` are failing we should consider replacing the project with an up-to-date one.

### DIFF
--- a/test/integration/runtime_int_test.go
+++ b/test/integration/runtime_int_test.go
@@ -117,11 +117,8 @@ func (suite *RuntimeIntegrationTestSuite) TestInUse() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
-	cp := ts.SpawnWithOpts(
-		e2e.OptArgs("checkout", "ActiveState-CLI/Perl-5.36", "."),
-		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
-	)
-	cp.Expect("Checked out project", e2e.RuntimeSourcingTimeoutOpt)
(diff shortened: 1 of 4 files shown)


test/integration/runtime_int_test.go
```diff
@@ -117,8 +117,11 @@ func (suite *RuntimeIntegrationTestSuite) TestInUse() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
-	cp := ts.Spawn("checkout", "ActiveState-CLI/Perl-5.32", ".")
-	cp.Expect("Skipping runtime setup")
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("checkout", "ActiveState-CLI/Perl-5.36", "."),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
```
mitchell-as, line 122: This shouldn't be needed because we're sourcing the runtime in the next step (`state activate`). I think you can revert this entire block except for the Perl-5.36 change.

Presumably this is leftover from previous debugging?

mitchell-as (changes requested):

mitchell-as (approved):

PR #3228, opened 2024-04-08
Use powershell scripts instead of batch scripts for `state exec`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2673" title="DX-2673" target="_blank">TaskDX-2673</a>  Get rid of batch files for running user commands
  </td></table>
  <br />
 

Passing arguments through batch is not reliable, especially arguments with special characters like '<' in them.

On Windows, Go -> cmd -> executor/exe argument passthrough via `%*` is problematic because the batch script doesn't know how to do parameter escaping if a parameter contains special characters. Instead, use Go -> pwsh -> executor/exe so that powershell correctly forwards arguments.

This PR adds the following new integration tests:

- `state exec` uses powershell under the hood so that something like `state exec perl testargs.pl "<3"` works as expected (the original bug report) without the user needing to know anything about escaping.
- `state run foo "^<3"` where `foo` is a batch script explicitly written by the user works, but requires the '^' escape (presumably the user knows this, as I mention in a stand-alone comment). Note that State Tool invokes the user-defined batch script; it doesn't create its own script to facilitate running this.
- Within `state shell`, `foo "<3"` where `foo` is a powershell script explicitly written by the user works as expected. Note that State Tool invokes the user-defined script (whether it be powershell or batch); it doesn't create its own script to facilitate running this one.

### DIFF
--- a/internal/subshell/sscommon/sscommon.go
+++ b/internal/subshell/sscommon/sscommon.go
@@ -131,7 +131,13 @@ func runWithCmd(env []string, name string, args ...string) (rerr error) {
 	case ".bat":
 		// Use powershell to run the bat file because passing arguments with special characters to
 		// batch files has limitations.
-		ps1Script := fmt.Sprintf("& %q @args\nexit $LASTEXITCODE", name)
+		ps1Script := fmt.Sprintf(`
+		try {
+			& %q @args
+			exit $LASTEXITCODE
+		} catch {
+			exit 1 # file not found
+		}`, name)
 		ps1File, err := fileutils.WriteTempFile("*.ps1", []byte(ps1Script))
 		if err != nil {
 			return errs.Wrap(err, "Unable to write temporary powershell script")
--- a/internal/subshell/sscommon/sscommon.go
+++ b/internal/subshell/sscommon/sscommon.go
@@ -131,13 +131,7 @@ func runWithCmd(env []string, name string, args ...string) (rerr error) {
 	case ".bat":
 		// Use powershell to run the bat file because passing arguments with special characters to
 		// batch files has limitations.
-		ps1Script := fmt.Sprintf(`
-		try {
-			& %q @args
-			exit $LASTEXITCODE
-		} catch {
-			exit 1 # file not found
-		}`, name)
+		ps1Script := fmt.Sprintf("& %q @args\nexit $LASTEXITCODE", name)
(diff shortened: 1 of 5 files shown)


test/integration/run_int_test.go
```diff
@@ -313,6 +314,43 @@ func (suite *RunIntegrationTestSuite) TestRun_Perl_Variable() {
...
+      %s
+`, lang, cmd), "\n"))
+	suite.Require().NoError(err, "extra config is appended")
+
+	arg := "<3"
+	if runtime.GOOS == "windows" {
+		// The '<' needs to be escaped with '^', and I don't know why. There is no way around it.
+		// The other exec and shell integration tests that test arg passing do not need this escape.
+		// Only this batch test does.
+		arg = "^<3"
```
mitchell-as, line 348: I feel like this is okay because if you are explicitly writing a batch script, then you ought to know the ins and outs of how batch handles incoming arguments, so if you're doing `state run <name> <args>` with funky `<args>`, then you ought to know what needs escaping and what does not.

internal/subshell/sscommon/sscommon.go
```diff
@@ -127,7 +127,7 @@ func runWithCmd(env []string, name string, args ...string) error {
 	case ".bat":
 		// No action required
 	case ".ps1":
-		args = append([]string{"-file", name}, args...)
+		args = append([]string{"-executionpolicy", "bypass", "-file", name}, args...)
```
MDrakos, line 130: I wonder if we've worked through the potential implications of this... some customers might have locked down environments and passing this flag could cause execution to fail. I remember we had a discussion about this during stand but I can't recall all of the details.

MDrakos (commented):
This looks good but I haven't approved yet because of my concern in the comment below.

internal/subshell/sscommon/sscommon.go
mitchell-as: We spoke during stand that we'd like to have Mark test this in a brand-new, unprivileged environment. If that fails, we can undo all of this. In my testing, it seems to work, but I haven't tried all configurations.

MDrakos: I do wonder if that will catch potential issues with more complex locked-down environment, but I guess we will see if people complain.

MDrakos (approved):

PR #3234, opened 2024-04-09
Added --target behaviour to `state artifacts` and `state artifacts dl`

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2725" title="DX-2725" target="_blank">TaskDX-2725</a>  `state artifacts` takes an optional `--target` flag
  </td></table>
  <br />

### DIFF
--- a/cmd/state/internal/cmdtree/artifacts.go
+++ b/cmd/state/internal/cmdtree/artifacts.go
@@ -34,6 +34,11 @@ func newArtifactsCommand(prime *primer.Values) *captain.Command {
 				Value:       &params.CommitID,
 			},
 			{
+				Name:        "target",
+				Description: locale.Tl("artifacts_flags_target_description", "The target to report artifacts for"),
+				Value:       &params.Target,
+			},
+			{
 				Name:        "full-id",
 				Description: "List artifacts with their full identifier",
 				Value:       &params.Full,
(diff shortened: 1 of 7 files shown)


cmd/state/internal/cmdtree/artifacts.go
```diff
@@ -71,6 +76,11 @@ func newArtifactsDownloadCommand(prime *primer.Values) *captain.Command {
 				Description: locale.Tl("artifacts_download_flags_commit_description", "The commit ID to download artifacts from"),
 				Value:       &params.CommitID,
 			},
+			{
+				Name:        "target",
+				Description: locale.Tl("artifacts_flags_target_description", "The target to download artifacts from"),
```
mitchell-as, line 81: This is a duplicate locale key.

cmd/state/internal/cmdtree/artifacts.go
```diff
@@ -80,8 +90,8 @@ func newArtifactsDownloadCommand(prime *primer.Values) *captain.Command {
 				Required:    true,
 			},
 			{
-				Name:        "target",
-				Description: locale.Tl("artifacts_download_arg_target", "The target directory to download the artifact to"),
+				Name:        "path",
+				Description: locale.Tl("artifacts_download_arg_target", "The target path to download the artifact to"),
```
mitchell-as, line 94: nit: You may want to change this locale key since the name changed.

mitchell-as (changes requested):

cmd/state/internal/cmdtree/artifacts.go
Naatan: It's only duplicated between artifacts and artifacts dl, which are so closely related that I would consider this a "feature".

mitchell-as: I meant it's duplicated with another key added to this PR: https://github.com/ActiveState/cli/pull/3234/files#diff-ba71f664d43f1a5f805cd749ceab31582d01498132c3fdb97fc9bc941523058bR38

If this is intentional, then please add to en-us.yaml so that shared locale keys always use the same value.

mitchell-as (changes requested):

mitchell-as (approved):

PR #3237, opened 2024-04-10
Fixed build error.

mitchell-as:


Naatan (approved):
Can't trust those auto merges.. thanks @mitchell-as !

PR #3239, opened 2024-04-10
Check formatting on PRs

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2636" title="DX-2636" target="_blank">TaskDX-2636</a>  Enforce files ending with line break
  </td></table>
  <br />


Naatan (changes requested):
This looks good, but I worry this is going to cause a lot of maintenance overhead. There's a lot of unix/bash specific logic here that's prone to breakage as we've found with maintaining similar bash scripts.

Enforcing new lines is a pretty established best practice, there should be plenty of tools for us to use. 

Case in point: https://github.com/marketplace/actions/linelint

I'd suggest using something established and maintained rather than rolling our own.

MDrakos:
I looked at the action you linked as well as some other solutions. The main issue I can see is how we get these actions to only work with changed files. I don't believe `linelint` has a configuration option for this and it doesn't look like the repo has been touched in 2 years.

Other solutions have line linting as a feature but are also have a lot of other stuff we don't need which could lead to added headaches in getting them setup. Do we think we will want to add more linting in the future? If so I can see a case for us investing in a linting solution now.

This script is fairly simple, but does have the same issues that all other bash scripts have. For this use-case I think it makes sense but I'm happy to revisit other linting actions or give `linelint` a try.

Naatan:
Ok, let's go with this for now, if it turns into a maintenance overhead we should resort to a maintained solution sooner rather than later.

Naatan (approved):

PR #3241, opened 2024-04-10
`state artifacts` lists all available artifacts

Naatan:


internal/runners/artifacts/artifacts.go
```diff
@@ -76,6 +78,8 @@ type structuredArtifact struct {
 	URL  string `json:"url"`
 }
 
+const DefaultTarget = "__all__"
```
mitchell-as, line 81: nit: It might be worth commenting where this value comes from, as we're in a runner and not in the buildplanner package or whatever else may help shed light on this value.

Actually, now that I think about it, maybe this constant should live somewhere in the runtime package and we refer to it in this file?

mitchell-as (dismissed):
Approved, but with a nit that's a bit more than a nit, though it's not a showstopper.

mitchell-as (approved):

PR #3245, opened 2024-04-12
Do not require filepath argument to `state publish --edit`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2701" title="DX-2701" target="_blank">BugDX-2701</a>  `state publish --edit` fail work without path to the file 
  </td></table>
  <br />
 

The runner performs the necessary check and raises an input error if neither filepath nor --edit were provided.


Naatan (changes requested):
I think this is ok, though have you checked whether uploading a new ingredient errors out with an understandable error if no file was given? Effectively it is required when not using `--edit`.

mitchell-as:
Yes, as I mentioned in the PR description, the runner does this: https://github.com/ActiveState/cli/blob/3d071a6befefae384dc7b7017db4e259999cfffb/internal/runners/publish/publish.go#L103

Unless I'm missing the point?

Naatan (approved):

PR #3250, opened 2024-04-15
Add `state manifest` command

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2675" title="DX-2675" target="_blank">StoryDX-2675</a>  As a user I can run a single command and see my full order
  </td></table>
  <br />


MDrakos:
When reviewing the changed files I noticed that I started to move the tea view to it's own file in the`output` package. This must have been done before the weekend because I completely forgot about it and ended up implementing the rest of the story without using the view. I seem to recall we talked about it but there is nothing on the story about using the tea view. I've gone ahead and used it, but we can easily remove it and file a follow up if we want to go that route.

activestate.yaml
```diff
@@ -433,3 +433,4 @@ events:
   - name: file-changed
     scope: ["internal/locale/locales"]
     value: build
+config_version: 1
```
Naatan, line 436: Someones opted in :p We shouldn't commit this just yet. The intend is that you'd commit this along with the buildscript, which isn't quite ready yet.

Naatan (changes requested):
Haven't reviewed the code yet. Just based on my testing the command:

* This should not be taking over the terminal view and implementing scrolling behaviour. Just print to stdout, nothing fancy.
* This should not trigger an API solve, we are already storing the buildplan. So as long as the runtime cache isn't invalid we should already have the info we need.
* The arrows showing the request version vs actual should be white. Only the version numbers should be teal.


Figure this may already lead to substantial code changes so holding off on reviewing code.

internal/locale/locales/en-us.yaml
```diff
@@ -1591,4 +1591,9 @@ svcctl_cannot_create_debug_info:
...
   other: |
     Could not find compatible requirement for initial commit. Please ensure that you are trying to install a valid package.
-    To check for available packages run '[ACTIONABLE]state search <package_name>[/RESET]'.
\ No newline at end of file
+    To check for available packages run '[ACTIONABLE]state search <package_name>[/RESET]'.
+manifest_deprecation_warning:
+  other: |
+    [WARNING]Deprecation Warning:[/RESET] This command is deprecated, please use '[ACTIONABLE]state manfiest[/RESET]' instead.
+manifest_runtime_needs_update:
+  other: "Your project runtime needs to be updated.\nThe manifest will be generated based on the local runtime.\nTo update your runtime run [ACTIONABLE]state refresh[/RESET].\n"
```
Naatan, line 1599:
```suggestion
manifest_runtime_needs_update:
  other: "[WARNING]WARNING:[/RESET] Your runtime is out of date, in order to ensure an accurate manifest please run [ACTIONABLE]state refresh[/RESET].\n"
```

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -164,6 +164,8 @@ func New(prime *primer.Values, args ...string) *CmdTree {
 		newArtifactsDownloadCommand(prime),
 	)
 
+	manifestCmd := newManifestCommmand(prime)
+
```
Naatan, line 168:
```suggestion
```

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -215,6 +217,7 @@ func New(prime *primer.Values, args ...string) *CmdTree {
 		newCommitCommand(prime),
 		newPublish(prime),
 		newEvalCommand(prime),
+		manifestCmd,
```
Naatan, line 220:
```suggestion
		newManifestCommmand(prime),
```

No need for a variable if you're only passing it here.

cmd/state/internal/cmdtree/manifest.go
```diff
@@ -0,0 +1,30 @@
...
+		prime,
+		[]*captain.Flag{},
+		[]*captain.Argument{},
+		func(_ *captain.Command, _ []string) error {
+			return runner.Run()
+		},
+	)
+
+	cmd.SetGroup(PackagesGroup)
+	cmd.SetAliases("manifest")
```
Naatan, line 26:
```suggestion
```

Redundant

internal/locale/locales/en-us.yaml
```diff
@@ -1591,4 +1591,9 @@ svcctl_cannot_create_debug_info:
 err_initial_no_requirement:
   other: |
     Could not find compatible requirement for initial commit. Please ensure that you are trying to install a valid package.
-    To check for available packages run '[ACTIONABLE]state search <package_name>[/RESET]'.
\ No newline at end of file
+    To check for available packages run '[ACTIONABLE]state search <package_name>[/RESET]'.
+manifest_deprecation_warning:
+  other: |
+    [WARNING]Deprecation Warning:[/RESET] This command is deprecated, please use '[ACTIONABLE]state manfiest[/RESET]' instead.
```
Naatan, line 1597:
```suggestion
    [WARNING]Warning:[/RESET] This command is deprecated, please use '[ACTIONABLE]state manfiest[/RESET]' instead.
```

internal/runners/manifest/manifest.go
```diff
@@ -0,0 +1,157 @@
...
+
+	return nil
+}
+
+func (m *Manifest) fetchArtifacts() ([]*artifact.Artifact, error) {
+	if m.project == nil || strings.EqualFold(os.Getenv(constants.DisableRuntime), "true") {
+		return nil, nil
+	}
+
+	target := target.NewProjectTarget(m.project, nil, target.TriggerPackage)
```
Naatan, line 106: We should add a new trigger for manifest, `TriggerPackage` is inappropriate.

internal/runners/manifest/manifest.go
```diff
@@ -0,0 +1,157 @@
...
+	rt, err := runtime.New(target, m.analytics, m.svcModel, m.auth, m.cfg, m.out)
+	if err != nil {
+		return nil, locale.WrapError(err, "err_packages_update_runtime_init", "Could not initialize runtime.")
+	}
+
+	// This shouldn't happen, but it's good to warn the user if it does
```
Naatan, line 112:
```suggestion
```

It can happen easily. eg. state tool update, or user ran `state clean cache`.

internal/runners/manifest/manifest.go
```diff
@@ -0,0 +1,157 @@
...
+	if rt.NeedsUpdate() {
+		m.out.Notice(locale.T("manifest_runtime_needs_update"))
+	}
+
+	return rt.ResolvedArtifacts()
+}
+
+type vulns map[string]*model.VulnerabilityIngredient
+
+func (v vulns) getVulns(name, namespace string) (*model.VulnerabilityIngredient, bool) {
```
Naatan, line 122: Please use fully written words for symbols that aren't scoped to a small set of logic.

Also the name suggests multiple are returned, but it only deals with one.

This also goes for addVulns and fetchVulns.

internal/runners/manifest/requirements.go
```diff
@@ -0,0 +1,146 @@
...
+	VersionOutput   string `json:"version" locale:"manifest_version,Version"`
+	License         string `json:"license" locale:"manifest_license,License"`
+	Vulnerabilities string `json:"vulnerabilities" locale:"manifest_vulnerabilities,Vulnerabilities (CVEs)" opts:"omitEmpty"`
+	// Must be last of the output fields in order for our table renderer to include all the fields before it
+	NamespaceOutput string `json:"namespace" locale:"manifest_namespace,Namespace" opts:"omitEmpty,separateLine"`
+
+	// These fields are used for internal processing
+	name      string
+	namespace string
+}
```
Naatan, line 29: This is triggering my OCD a bit, the symbol naming is very inconsistent. Suggest dropping `Output` from the property suffixes, and instead adding it to the type name, to make it consistent with `requirementsOutput`.

internal/runners/manifest/requirements.go
```diff
@@ -0,0 +1,146 @@
...
+
+		normalized, err := platformModel.FetchNormalizedName(req.Namespace, req.Name, auth)
+		if err != nil {
+			multilog.Error("Failed to normalize '%s': %v", req.Name, err)
+		}
+
+		packages, err := platformModel.SearchIngredientsStrict(req.Namespace, normalized, false, false, nil, auth)
+		if err != nil {
+			multilog.Error("Failed to search for '%s': %v", req.Name, err)
+		}
```
Naatan, line 66: I did not realize the license info was not part of the buildplan, but that makes sense in retrospect. If we cannot retrieve this license info in bulk then we should drop this. It's nice to have but not to the point that we want to be spamming our API.

Note any time you are calling an API inside a loop alarm bells should be going off. We should always try to avoid this.

internal/runners/manifest/requirements.go
```diff
@@ -0,0 +1,146 @@
...
+			)
+		}
+
+		low, ok := counts[vulnModel.SeverityLow]
+		if ok && low > 0 {
+			vulnReport = append(
+				vulnReport,
+				locale.Tl("manifest_vulnerability_low", fmt.Sprintf("[GREEN]%d Low[/RESET]", low)),
+			)
+		}
```
Naatan, line 140: Should be able to turn this into a loop of some sort to avoid the duplication of logic.

pkg/platform/model/vcs.go
```diff
@@ -218,6 +218,24 @@ func LanguageFromNamespace(ns string) string {
 	return values[1]
 }
 
+func IsCustomNamespace(ns string) bool {
+	supportedNamespaces := []NamespaceType{
+		NamespacePackage,
+		NamespaceBundle,
+		NamespaceLanguage,
+		NamespacePlatform,
+		NamespaceOrg,
```
Naatan, line 227: Org is a custom namespace we would want to surface.

Also this logic feels veeery specific to `state manifest`, suggest moving the code there.

Naatan (changes requested):
Note the `For CVE info run 'state security'` line is missing.

This is moving in the right direction, but it feels a bit proof of concept grade at the moment. Please spend some more time finessing your architecture until you are personally satisfied. I don't mind if this takes an extra day of work or something, we're likely to be working on this code often so having this be in a good place architecturally will save us time in the long run.

MDrakos:
This feels like it's in a better place now. Since you're doing some refactoring and architectural work let me know if you're looking for something specific. I've gone over this code enough times that I feel like I'm starting to move things around just for the sake of it.

internal/locale/locales/en-us.yaml
```diff
@@ -1592,3 +1592,16 @@ err_initial_no_requirement:
...
+manifest_runtime_needs_update:
+  other: "[WARNING]WARNING:[/RESET] Your runtime is out of date, in order to ensure an accurate manifest please run [ACTIONABLE]state refresh[/RESET].\n"
+manifest_vulnerability_critical:
+  other: "[RED]{{.V0}} Critical[/RESET]"
+manifest_vulnerability_high:
+  other: "[ORANGE]{{.V0}} High[/RESET]"
+manifest_vulnerability_medium:
+  other: "[YELLOW]{{.V0}} Medium[/RESET]"
+manifest_vulnerability_low:
+  other: "[GREEN]{{.V0}} Low[/RESET]"
```
Naatan, line 1607: Would be good to standardize this and combine it with https://github.com/ActiveState/cli/blob/DX-2720/internal/runners/packages/searchView.go#L181-L181

internal/table/table.go
```diff
@@ -51,7 +52,11 @@ func (t *Table) Render() string {
 	var out string
 	if !t.HideHeaders {
 		out += "[NOTICE]" + renderRow(t.headers, colWidths) + "[/RESET]" + linebreak
-		out += "[DISABLED]" + strings.Repeat(dash, total) + "[/RESET]" + linebreak
+		if !t.HideDash {
+			out += "[DISABLED]" + strings.Repeat(dash, total) + "[/RESET]" + linebreak
+		} else {
+			out += "[DISABLED]" + strings.Repeat(" ", total) + "[/RESET]" + linebreak
+		}
```
Naatan, line 59: This doesn't appear to be working for me:

image

Also, we should probably just not render this line at all rather than make it "empty".

internal/runners/manifest/manifest.go
```diff
@@ -0,0 +1,155 @@
...
+	if rt.NeedsUpdate() {
+		m.out.Notice(locale.T("manifest_runtime_needs_update"))
+	}
+
+	return rt.ResolvedArtifacts()
+}
+
+func (m *Manifest) fetchVulnerabilities(reqs []bpModel.Requirement) (vulnerabilities, error) {
+	if !m.auth.Authenticated() {
+		return nil, nil
```
Naatan, line 125: This effectively means we will say "None Detected" under CVEs if you are not authenticated. Instead of "None Detected" we should print "Authenticate to View" in this scenario.

internal/runners/manifest/requirements.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
+	platformModel "github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+)
+
+type requirement struct {
+	Name            string                      `json:"name"`
+	Namespace       string                      `json:"namespace,omitempty"`
+	Version         resolvedVersion             `json:"version"`
```
Naatan, line 14: Can we rename this to ResolvedVersion? While reviewing I kept stumbling over this thinking we were printing the version only rather than requested versus actual. Obviously you're handling both, just the name is missleading which makes this harder to grok.

internal/runners/manifest/requirements.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return namespace
+}
+
+func isCustomNamespace(ns string) bool {
+	supportedNamespaces := []platformModel.NamespaceType{
+		platformModel.NamespacePackage,
+		platformModel.NamespaceBundle,
+		platformModel.NamespaceLanguage,
+		platformModel.NamespacePlatform,
```
Naatan, line 80:
```suggestion
```

Platforms aren't requirements.

internal/runners/manifest/version.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
+	platformModel "github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+)
+
+type resolvedVersion struct {
+	Requested string `json:"requested"`
+	Resolved  string `json:"resolved,omitempty"`
```
Naatan, line 12: Currently this is empty in JSON output, which is a bit awkward imo. I'd rather see Requested and Resolved as the same version.

You could probably use MarshalStructured and MarshalOutput on the `resolvedVersion` struct to achieve this.

internal/runners/manifest/version.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
+	platformModel "github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+)
+
+type resolvedVersion struct {
+	Requested string `json:"requested"`
```
Naatan, line 11: Currently with JSON output we render "Auto" as the value for Requested. For structured this should just be empty, as "Auto" is only meant for making it humanly readable.

Naatan (changes requested):
Looks much better, nice work! Just some minor iteration.

internal/table/table.go
MDrakos: Whoops, not sure how that happened. I've updated it not to print the line however the spacing isn't ideal if the line is printed or not.

Empty line:
```
  Name        Version          Vulnerabilities (CVEs)

  python      3.9.13           1 Critical, 7 High, 3 Medium
  libsass     auto → 0.21.0    None detected
  pexpect     4.4.0            None detected
  psutil      auto → 5.9.0     None detected
  requests    2.25.0           1 Medium

For CVE info run 'state security'
```

No line:
```
  Name        Version          Vulnerabilities (CVEs)
  python      3.9.13           1 Critical, 7 High, 3 Medium
  libsass     auto → 0.21.0    None detected
  pexpect     4.4.0            None detected
  psutil      auto → 5.9.0     None detected
  requests    2.25.0           1 Medium

For CVE info run 'state security'
```

Maybe we should revert this and keep the dash?

internal/runners/manifest/version.go
MDrakos: It doesn't appear that adding `MarshalStructured` to a struct field that itself has a `MarshallStructured` method will result in the first function being called. I've had to make that explicit, but it still works.

internal/runners/manifest/requirements.go
```diff
@@ -57,10 +57,22 @@ func (o requirements) MarshalOutput(_ output.Format) interface{} {
...
+		Requirements: requirementsOutput,
+	}
 }
 
-func (o requirements) MarshalStructured(_ output.Format) interface{} {
+func (o requirements) MarshalStructured(f output.Format) interface{} {
+	for _, req := range o.Requirements {
+		req.ResolvedVersion.MarshalStructured(f)
+
+		if !req.Vulnerabilities.authenticated {
```
Naatan, line 71:
```
state manifest --output json
An unexpected error occurred.
Error: runtime error: invalid memory address or nil pointer dereference
Stack trace: goroutine 37 [running]:
runtime/debug.Stack()
	/Users/runner/hostedtoolcache/go/1.20.14/x64/src/runtime/debug/stack.go:24 +0x65
github.com/ActiveState/cli/internal/sighandler.(*sigHandler).WaitForFunc.func1.1()
	/Users/runner/work/cli/cli/internal/sighandler/awaiting.go:48 +0x2d
panic({0x1daf8a0, 0x2b24cf0})
	/Users/runner/hostedtoolcache/go/1.20.14/x64/src/runtime/panic.go:884 +0x213
github.com/ActiveState/cli/internal/runners/manifest.requirements.MarshalStructured(...)
	/Users/runner/work/cli/cli/internal/runners/manifest/requirements.go:71
github.com/ActiveState/cli/internal/output.mediatorValue({0x1e4bfe0?, 0xc000858630?}, {0x1f3cddc, 0x4})
	/Users/runner/work/cli/cli/internal/output/mediator.go:57 +0x96
github.com/ActiveState/cli/internal/output.(*Mediator).Print(0xc0002f9500, {0x1e4bfe0?, 0xc000858630?})
	/Users/runner/work/cli/cli/internal/output/mediator.go:31 +0x3d
github.com/ActiveState/cli/internal/runners/manifest.(*Manifest).Run(0xc0000c1a80)
	/Users/runner/work/cli/cli/internal/runners/manifest/manifest.go:76 +0x25c
github.com/ActiveState/cli/cmd/state/internal/cmdtree.newManifestCommmand.func1(0xc000214f48?, {0xc000614360?, 0xc0008118c0?, 0x0?})
	/Users/runner/work/cli/cli/cmd/state/internal/cmdtree/manifest.go:21 +0x1d
github.com/ActiveState/cli/internal/captain.(*Command).cobraExecHandler.func2()
	/Users/runner/work/cli/cli/internal/captain/command.go:703 +0x1b3
github.com/ActiveState/cli/internal/sighandler.(*sigHandler).WaitForFunc.func1()
	/Users/runner/work/cli/cli/internal/sighandler/awaiting.go:50 +0x6f
created by github.com/ActiveState/cli/internal/sighandler.(*sigHandler).WaitForFunc
	/Users/runner/work/cli/cli/internal/sighandler/awaiting.go:47 +0xca

```

Naatan (changes requested):

Naatan (changes requested):
Also, at the moment when you have a clean cache it fails to access the buildplan and errors out with an "Execute failed" error. We should provide a user facing error for this scenario.

internal/runners/manifest/requirements.go
MDrakos: Bah, thanks for catching this!

internal/runners/manifest/version.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/model"
+	platformModel "github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/artifact"
+)
+
+type resolvedVersion struct {
+	Requested string `json:"requested,omitempty"`
```
Naatan, line 12:
```suggestion
	Requested string `json:"requested"`
```

I think it's friendlier to explicitly show an empty value.

internal/runners/manifest/requirements.go
```diff
@@ -66,9 +66,13 @@ func (o requirements) MarshalOutput(_ output.Format) interface{} {
 
 func (o requirements) MarshalStructured(f output.Format) interface{} {
 	for _, req := range o.Requirements {
+		if req == nil {
+			continue
+		}
```
Naatan, line 71: This feels like it should happen earlier, ie. when `o.Requirements` is calculated. Unless there's a reason for having empty entries?

test/integration/manifest_int_test.go
```diff
@@ -41,6 +42,66 @@ func (suite *ManifestIntegrationTestSuite) TestManifest() {
...
+
+	type requirement struct {
+		Name            string          `json:"name"`
+		Version         version         `json:"version"`
+		Vulnerabilities vulnerabilities `json:"vulnerabilities"`
+	}
+
+	type requirements struct {
+		Requirements []requirement `json:"requirements"`
+	}
```
Naatan, line 75: This could end up being painful to maintain as it makes us very opinionated on the exact format of the json output which isn't exactly what we intend to test here.

Suggest instead using https://github.com/ActiveState/cli/blob/DX-2720/test/integration/shared_int_test.go#L26-L26. I don't think you need to assert the exact version numbers or counts, maybe just check that we are producing output with the `requirements` key or something. 

I think this test should assert that we are in fact outputting json, and that the outputted json isn't an error message.

Another option is to simply expose the structs you used to generate the JSON, and unmarshal from that.

Naatan (approved):

PR #3254, opened 2024-04-17
Raise an input error for lack of commit ID in activestate.yaml unless using `state reset` to reset it.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2680" title="DX-2680" target="_blank">StoryDX-2680</a>  As a user I can expect State Tool to give me an actionable error when my project URL is malformed
  </td></table>
  <br />

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -183,12 +183,7 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	if pjPath != "" {
 		pjf, err := projectfile.FromPath(pjPath)
 		if err != nil {
-			if projectfile.IsFatalError(err) {
-				return err
-			} else if args[1] != "reset" {
-				return locale.NewInputError("err_bad_project_url_reset")
-			}
-			// otherwise, ignore things like invalid commit ID errors for `state reset`, which will fix them
+			return err
 		}
 		pj, err = project.New(pjf, out)
 		if err != nil {
@@ -254,6 +249,13 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 				)
 			}
 		}
+	}
+
+	// Centralize handling the lack of commit ID in the project to here.
+	// We cannot throw an error during projectfile parsing, because `state reset` needs a valid
+	// primer project struct to work with.
+	if childCmd != nil && childCmd.Name() != "reset" && pj != nil && !pj.Source().HasCommitID() {
+		return locale.NewInputError("err_bad_project_url_reset")
 	}
 
 	err = cmds.Execute(args[1:])
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -251,13 +251,6 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 		}
 	}
 
-	// Centralize handling the lack of commit ID in the project to here.
-	// We cannot throw an error during projectfile parsing, because `state reset` needs a valid
-	// primer project struct to work with.
-	if childCmd != nil && childCmd.Name() != "reset" && pj != nil && !pj.Source().HasCommitID() {
-		return locale.NewInputError("err_bad_project_url_reset")
-	}
-
 	err = cmds.Execute(args[1:])
 	if err != nil && !errs.IsSilent(err) {
 		cmdName := ""
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -183,7 +183,12 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	if pjPath != "" {
 		pjf, err := projectfile.FromPath(pjPath)
 		if err != nil {
-			return err
+			if projectfile.IsFatalError(err) {
+				return err
+			} else if args[1] != "reset" {
+				return locale.NewInputError("err_bad_project_url_reset")
+			}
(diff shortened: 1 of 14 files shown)


pkg/localcommit/localcommit.go
```diff
@@ -32,7 +34,7 @@ func Get(pjpath string) (strfmt.UUID, error) {
 	}
 
 	if !strfmt.IsUUID(proj.LegacyCommitID()) {
-		return "", locale.NewInputError("err_commit_id_invalid", "Commit ID in activestate.yaml is not UUID formatted: {{.V0}}.", proj.LegacyCommitID())
+		return "", &InvalidCommitID{locale.NewInputError("err_commit_id_invalid", "", proj.LegacyCommitID())}
```
mitchell-as, line 37: Note, the `err_commit_id_invalid` key is in `en-us.yaml`, so no need for this default text.

mitchell-as:
The lone windows failure is a sporadic network error, and unrelated to this PR.

mitchell-as:
The lone test failure is a sporadic one unrelated to this PR.

internal/locale/locales/en-us.yaml
```diff
@@ -534,6 +534,8 @@ err_invalid_project:
   other: Your activestate.yaml is missing the 'project' field.
 err_bad_project_url:
   other: "Invalid value for 'project' field in your activestate.yaml, please ensure it is in the format of: 'https://platform.activestate.com/org/project'."
+err_bad_project_url_reset:
+  other: "Invalid value for 'project' field in your activestate.yaml. Please run '[ACTIONABLE]state reset[/RESET]' to fix it."
```
mitchell-as, line 538: Let me know if you prefer a different message. The ticket didn't specify what exactly to use.

cmd/state/main.go
```diff
@@ -251,6 +251,13 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 		}
 	}
 
+	// Centralize handling the lack of commit ID in the project to here.
+	// We cannot throw an error during projectfile parsing, because `state reset` needs a valid
+	// primer project struct to work with.
+	if childCmd != nil && childCmd.Name() != "reset" && pj != nil && !pj.Source().HasCommitID() {
+		return locale.NewInputError("err_bad_project_url_reset")
```
Naatan, line 258: I think this error should come from projectfile.Parse(). We can then during prime initialization disregard this error if we detect we are running `reset`. 

Otherwise we are only surfacing errors for the projectfile detected in the primer, which isn't guaranteed to be the projectfile that ends up being used.

Naatan (changes requested):

cmd/state/main.go
mitchell-as: Sorry, I don't like it. `projectfile.Parse()` always returns `nil, err` on error. In my limited experience with Go, whenever you get an err, you can't use the return value because it's contents are undefined (not literally, but due to the error and where it occurred, the fields, etc. may not be populated as one would expect). In order to do what you're suggesting, inside `projectfile.Parse()` we would have to recognize "non-fatal" errors and continue (but save them/pack them for returning at the end of the function). I'm concerned that that would turn into a spaghetti mess.

mitchell-as: I did some work in the meantime to prove the point, and it is indeed a spaghetti mess. It's going to be a maintenence nightmare to keep track of when a project parse error is okay/non-fatal vs. not okay/fatal and how to continue. As you can see with status checks, I some seemingly unrelated unit test failures popped up.

Based on your feedback about the project being passed to the primer possibly not ultimately being the target project, I think we ought to abandon this. Is lack of commitID a common enough occurrence to warrant such intervention?

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3254#discussion_r1571064207

pkg/projectfile/projectfile.go
```diff
@@ -664,7 +664,7 @@ func (p *Project) parseURL() (projectURL, error) {
 
 func validateUUID(uuidStr string) error {
 	if ok := strfmt.Default.Validates("uuid", uuidStr); !ok {
-		return locale.NewError("invalid_uuid_val", "Invalid commit ID {{.V0}} in activestate.yaml. Please remove it and run `[ACTIONABLE]state pull[/RESET]` to reset it", uuidStr)
+		return locale.NewError("err_commit_id_invalid", "", uuidStr)
```
mitchell-as, line 667: This key is already defined in en-us.yaml, so there's no need for the optional text.

mitchell-as:
I think I may have found a better place to put this. I changed `localcommit.Get()` to throw an error that can be caught and handled as a user-facing error at the highest level (captain), but `state reset` can ignore this error and do its thing.

mitchell-as:
The test failures are the usual timeouts and unrelated to this PR.

pkg/localcommit/localcommit.go
```diff
@@ -14,6 +14,11 @@ import (
 // https://activestatef.atlassian.net/browse/DX-2524
 var proj *project.Project
 
+type InvalidCommitID struct {
+	*locale.LocalizedError
```
Naatan, line 18: You don't need to localize this, since it's being rationalized in captain. There shouldn't be any scenario in which we surface this error directly to the user.

Naatan (approved):

PR #3256, opened 2024-04-18
Do not print update/deprecation warnings/messages when running `state update`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2728" title="DX-2728" target="_blank">BugDX-2728</a>  We print the `Older version...` message in the summary of the update when user updated to the latest
  </td></table>
  <br />


mitchell-as:
I cannot actually test this with live messages. I have to use the debugger to verify the correct code path is hit when running `state update` vs. something else.

PR #3265, opened 2024-04-26
Async runtime sourcing

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2738" title="DX-2738" target="_blank">StoryDX-2738</a>  As a user I can use the state tool without sourcing the runtime
  </td></table>
  <br />


internal/runbits/runtime/request.go
```diff
@@ -0,0 +1,59 @@
...
+	an analytics.Dispatcher,
+	proj *project.Project,
+	customCommitID *strfmt.UUID,
+	trigger target.Trigger,
+	svcm *model.SvcModel,
+	cfg Configurable,
+	opts Opts,
+) *Request {
+
+	return &Request{
```
MDrakos, line 36: The idea with this type was to have a central mechanism for controlling and manipulating how the runtime sourcing behaves. This may be overkill for the use-case of this story and we could do this check in the actual refresh functions.

internal/locale/locales/en-us.yaml
```diff
@@ -1470,6 +1470,9 @@ checkout_project_statement:
   other: |
     Checked out project [ACTIONABLE]{{.V0}}[/RESET], located at [ACTIONABLE]{{.V1}}[/RESET].
     For editors and other tooling use the executables at: [ACTIONABLE]{{.V2}}[/RESET].
+checkout_project_statement_async:
+  other: |
+    Checked out project [ACTIONABLE]{{.V0}}[/RESET], located at [ACTIONABLE]{{.V1}}[/RESET].
```
Naatan, line 1475:
```suggestion
    Checked out project [ACTIONABLE]{{.V0}}[/RESET], located at [ACTIONABLE]{{.V1}}[/RESET].
    Run '[ACTIONABLE]state refresh[/RESET]' to source the runtime.
```

Naatan (changes requested):
I'm sorry, I don't think this meets the AC of minimizing boilerplate. Technically a minimal boilerplate addition here would be YET ANOTHER field to `SolveAndUpdate()` that holds the async boolean. That would mean that in terms of boilerplate there would simply be another arg added in all these instances. With the PR as is we're instead initializing a whole new struct and passing that, which is far more boilerplate (thus not "minimal").

Honestly we should've probably blocked this story by the runtime refactor, because once we do that the whole SolveAndUpdate() call having waaaaay too many arguments problem should be somewhat if not entirely solved. At the very least it'd have to be revisited as part of the refactor. And that then in itself would instruct the solution for what to do with async here.

I am really excited for this feature though, don't wanna hold it back too long.. How about we just address this by for now increasing the argument list with yet another argument? Accepting that a cleaner solution will be part of the runtime refactor.

internal/constants/constants.go
```diff
@@ -444,6 +444,9 @@ const DefaultBranchName = "main"
 // UnstableConfig is the config key used to determine whether the user has opted in to unstable commands
 const UnstableConfig = "optin.unstable"
 
+// AsyncRuntimeConfig is the config key used to determine whether the user has opted in to async runtimes
+const AsyncRuntimeConfig = "optin.unstable.async.runtime"
```
Naatan, line 448:
```suggestion
const AsyncRuntimeConfig = "optin.unstable.async_runtime"
```

internal/runbits/runtime/request.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/ActiveState/cli/internal/analytics"
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/runtime/target"
+	"github.com/ActiveState/cli/pkg/project"
+	"github.com/go-openapi/strfmt"
+)
+
+type Request struct {
```
Naatan, line 13: I think this can be dropped now?

internal/runbits/runtime/refresh.go
```diff
@@ -40,6 +51,7 @@ func SolveAndUpdate(
 	svcm *model.SvcModel,
 	cfg Configurable,
 	opts Opts,
+	async bool,
```
Naatan, line 54: I think you can safely drop this and just use `cfg` to get the async value. Then for the `state refresh` use-case you can check the `trigger`. That should bring the boilerplate further down.

internal/runners/languages/languages.go
```diff
@@ -76,13 +78,18 @@ func (l *Languages) Run() error {
...
-	artifacts, err := rt.ResolvedArtifacts()
-	if err != nil && !errs.Matches(err, store.ErrNoBuildPlanFile) {
-		return locale.WrapError(err, "err_language_resolved_artifacts", "Unable to resolve language version(s)")
+
+	var artifacts []*artifact.Artifact
+	if !async {
+		artifacts, err = rt.ResolvedArtifacts()
+		if err != nil && !errs.Matches(err, store.ErrNoBuildPlanFile) {
+			return locale.WrapError(err, "err_language_resolved_artifacts", "Unable to resolve language version(s)")
+		}
```
Naatan, line 92: This goes beyond the scope of async. Async means "don't source the runtime". It does not mean "don't evaluate the runtime". For this command we only need to do a solve, we don't need to source the entire runtime. I imagine there's other commands where this applies too.

Naatan (changes requested):

internal/constants/constants.go
MDrakos: We have rules to enforce config key structure and it flags the underscore in the key as invalid. Should we change that for this story?

internal/runbits/runtime/refresh.go
MDrakos: This `bool` also serves as an override for the commands that do not respect the config value. We could override the config in the runners but that feels awkward. I'm also not following how checking the trigger can help with that.

MDrakos: Never mind... I think I got there on my own 😆

internal/constants/constants.go
Naatan: hmmm, the rule is to avoid that we introduce something like `optin_unstable_async_runtime`, which would be inconsistent. But to me the dots are for grouping, so for example anything under `optin.unstable` is obviously about toggling the unstable setting for that given property. With that in mind `optin.unstable.async` doesn't really make sense as a "group" to me since we're unlikely to put anything else there, which is why I suggested `async_runtime`.

With that context out of the way; maybe we can update the rule to require at least one dot but to otherwise accept underscores? We'll have to guard against monstrosities like `optin_unstable.async_runtime` in our code reviews.. :p

internal/runbits/runtime/refresh.go
```diff
@@ -29,6 +40,16 @@ type Configurable interface {
 	GetBool(key string) bool
 }
 
+var overrideTriggers = map[target.Trigger]bool{
```
Naatan, line 43:
```suggestion
var overrideAsyncTriggers = map[target.Trigger]bool{
```

internal/runbits/runtime/refresh.go
```diff
@@ -51,6 +72,11 @@ func SolveAndUpdate(
 		return nil, rationalize.ErrHeadless
 	}
 
+	if cfg.GetBool(constants.AsyncRuntimeConfig) && !overrideTriggers[trigger] {
```
Naatan, line 75:
```suggestion
	if cfg.GetBool(constants.AsyncRuntimeConfig) && !overrideAsyncTriggers[trigger] {
```

internal/runbits/runtime/refresh.go
Naatan: Makes it easier to understand what this is for without having to find how this var is used.

Naatan (changes requested):
Almost there!

internal/runners/config/key.go
```diff
@@ -10,7 +10,7 @@ type Key string
 
 // Set implements the captain ArgMarshaler interface
 func (k *Key) Set(v string) error {
-	regex := regexp.MustCompile(`^[A-Za-z0-9\.]+$`)
+	regex := regexp.MustCompile(`^[A-Za-z0-9]+(\.[A-Za-z0-9_]+)*$`)
```
Naatan, line 13: Nit: you don't need a group here (you can drop the brackets) and the wildcard, ie. `^[A-Za-z0-9]+\.[A-Za-z0-9_]+$`.

I'm pretty sure that would match all the same things.

Naatan (approved):

internal/runners/config/key.go
mitchell-as: Sorry for butting in, but I disagree. The current solution matches an alphanumeric name followed by 0 or more alphanumeric+underscore names (separated by '.'). The proposed solution would match an alphanumeric name followed by a '.' followed by single alphanumeric+underscore name. The '.' and additional name would be mandatory.

PR #3267, opened 2024-04-29
Use `errs.JoinMessage()` when sending errors to Rollbar or the log.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2740" title="DX-2740" target="_blank">TaskDX-2740</a>  Ensure `multilog.Error()` calls use `errs.JoinMessage(err)` instead of just `err`.
  </td></table>
  <br />

### DIFF
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -244,10 +244,9 @@ func writeMessageDepth(depth int, level string, msg string, args ...interface{})
 		switch arg := arg.(type) {
 		case func() interface{}:
 			args[i] = arg
+		case error:
+			args[i] = errs.JoinMessage(arg)
 		default:
-			if err, ok := arg.(error); ok {
-				args[i] = errs.JoinMessage(err)
-			}
 		}
 	}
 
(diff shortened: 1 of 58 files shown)


internal/logging/logging.go
```diff
@@ -239,11 +239,15 @@ func writeMessageDepth(depth int, level string, msg string, args ...interface{})
...
 	// This allows lazy evaluation of arguments which are return values
+	// Also, unpack error objects.
 	for i, arg := range args {
 		switch arg := arg.(type) {
 		case func() interface{}:
 			args[i] = arg
 		default:
+			if err, ok := arg.(error); ok {
+				args[i] = errs.JoinMessage(err)
+			}
```
MDrakos, line 250: This switch is already a type-check, so I think it makes sense to move this to an `case` rather than use the default.

MDrakos (changes requested):

internal/logging/logging.go
mitchell-as: You are correct. I was thinking if the error was something like a `locale.WrappedError`, the `arg.(type)` would not go into a `case error`. However, the debugger proved otherwise and I learned something new. Thanks for pointing this out!

MDrakos (approved):

PR #3269, opened 2024-04-29
Increase state-svc wait timeout and output a notice after 5 seconds of waiting.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2742" title="DX-2742" target="_blank">TaskDX-2742</a>  Increase state-svc timeout
  </td></table>
  <br />

### DIFF
--- a/internal/svcctl/svcctl.go
+++ b/internal/svcctl/svcctl.go
@@ -21,11 +21,12 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/osutils"
+	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/profile"
 )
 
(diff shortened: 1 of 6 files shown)


internal/svcctl/svcctl.go
```diff
@@ -21,11 +21,12 @@ import (
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/osutils"
+	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/profile"
 )
 
 var (
-	pingRetryIterations = 32
+	pingRetryIterations = 45
```
mitchell-as, line 29: Using VERBOSE=true and trial and error I found this iteration gives a 31s timeout, which is close to the 30s target in the ACs.

Naatan (approved):

PR #3270, opened 2024-04-29
`state pull` checks for common ancestor

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2752" title="DX-2752" target="_blank">TaskDX-2752</a>  `state pull` should only allow pulling compatible history
  </td></table>
  <br />


test/integration/pull_int_test.go
```diff
@@ -128,6 +128,22 @@ func (suite *PullIntegrationTestSuite) assertMergeStrategyNotification(ts *e2e.S
 	suite.Assert().Equal(strategy, conflictEvents[0].Label)
 }
 
+func (suite *PullIntegrationTestSuite) TestPullNoCommonParent() {
```
mitchell-as, line 131: I like the test!

pkg/platform/model/vcs.go
```diff
@@ -344,6 +344,15 @@ func CommitHistoryPaged(commitID strfmt.UUID, offset, limit int64, auth *authent
...
 }
 
+func HasCommonParent(commit1, commit2 strfmt.UUID, auth *authentication.Auth) (bool, error) {
+	commonParent, err := CommonParent(&commit1, &commit2, auth)
+	if err != nil {
+		return false, errs.Wrap(err, "Could not get common parent")
+	}
+
+	return commonParent != nil, nil
+}
```
mitchell-as, line 354: I feel like this function is unnecessary, and that the user can simply call `model.CommonParent()` and assert that it is non-nil.

mitchell-as (changes requested):

PR #3271, opened 2024-04-30
On Linux, sort platform IDs by closest libc version to local libc version.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2756" title="DX-2756" target="_blank">TaskDX-2756</a>  Improve platform detection
  </td></table>
  <br />
 

Consider a Linux Platform project configured with the following libc's: `[2.17, 2.28]`:

System LibC | `state checkout` selects Platform project with LibC
-----------------|------------------
2.15 | 2.17
2.17 | 2.17
2.20 | 2.17
2.26 | 2.17
2.28 | 2.28
2.30 | 2.28

### DIFF
--- a/pkg/platform/model/inventory.go
+++ b/pkg/platform/model/inventory.go
@@ -2,7 +2,9 @@ package model
 
 import (
 	"fmt"
+	"regexp"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -22,6 +24,7 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/api/inventory/inventory_client/inventory_operations"
 	"github.com/ActiveState/cli/pkg/platform/api/inventory/inventory_models"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/sysinfo"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


pkg/platform/model/inventory.go
```diff
@@ -364,14 +382,61 @@ func filterPlatformIDs(hostPlatform, hostArch string, platformIDs []strfmt.UUID,
...
+		if err != nil {
+			return nil, errs.Wrap(err, "Unable to get system libc")
+		}
+		logging.Debug("Detected system libc: %v", libc)
+		localLibc, err := strconv.ParseFloat(libc.Version(), 32)
+		if err != nil {
+			return nil, errs.Wrap(err, "Libc version is not a number: %s", libc.Version())
+		}
+		logging.Debug("libcMap: %v", libcMap)
+		logging.Debug("Unsorted pids: %v", pids)
```
Naatan, line 409: Can we drop the debug calls? This seems useful for when you were implementing it, but shipping with this will likely just cause unnecessary noise in the log file.

If there's value in shipping it, can we combine it all into one debug call with the vital info you want?

Naatan (changes requested):
Looks great! Just some nitpicking on the debug calls..

pkg/platform/model/inventory.go
mitchell-as: Okay, I've dropped the debug logging.

Naatan (approved):

PR #3278, opened 2024-05-01
Defined the color Orange for Windows console output.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2744" title="DX-2744" target="_blank">BugDX-2744</a>  `state languages install python@3.9` on Windows missing CVE severity 
  </td></table>
  <br />
 

image

The windows console does not have orange like the ansi consoles. Since the non-bolded/highlighted version of the mix between red and green (yellow) is more orange than the bolded/highlighted version of that color, I've named them orange and yellow, respectively.


MDrakos (approved):
Nice!

PR #3280, opened 2024-05-02
Add ExternalErrors error type

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2735" title="DX-2735" target="_blank">TaskDX-2735</a>  We can mark errors as 'ExternalError' and avoid it being reported to rollbar
  </td></table>
  <br />
 

The major changes are in `internal/errs/errs.go` and `internal/locale/errors.go`. The rest of the changeset it updating user input errors to external errors. I tried to catch as many as I could. The _vast_ majority of `NewInputError` calls are correct so this mostly updates the wrapped errors.


internal/runners/initialize/init.go
```diff
@@ -189,10 +189,10 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 
 	version, err := deriveVersion(lang, languageVersion, r.auth)
 	if err != nil {
-		if inferred || !locale.IsInputError(err) {
+		if inferred || (!locale.IsInputError(err) && !errs.IsExternalError(err)) {
```
mitchell-as, line 192: This is at least the second time I've seen this. Would it make sense to add a function that would wrap this so if we were to add more types of errors, we'd only have to change that function instead of all of these instances?

internal/runners/update/lock.go
```diff
@@ -143,10 +143,10 @@ func confirmLock(prom prompt.Prompter) error {
 	return nil
 }
 
-func fetchExactVersion(an analytics.Dispatcher, svc *model.SvcModel, channel, version string) (string, error) {
+func fetchExactVersion(svc *model.SvcModel, channel, version string) (string, error) {
```
mitchell-as, line 146: Cool, thanks for catching this unused parameter and removing it!

mitchell-as (dismissed):
This looks fine to me. I've left a couple of comments with suggestions for improvements if we want to keep an eye towards the future. I'll leave it up to you if you think they are worth it or not.

internal/runners/initialize/init.go
MDrakos: This is a good idea. Thanks for the suggestion.

mitchell-as (approved):

PR #3284, opened 2024-05-03
State publish respects `EDITOR` environment variable

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2768" title="DX-2768" target="_blank">BugDX-2768</a>  state publish doesn't respect EDITOR env var
  </td></table>
  <br />

### DIFF
--- a/internal/runners/scripts/edit_test.go
+++ b/internal/runners/scripts/edit_test.go
@@ -3,7 +3,6 @@ package scripts
 import (
 	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


internal/runners/scripts/edit_test.go
```diff
@@ -100,92 +97,6 @@ func (suite *EditTestSuite) TestCreateScriptFile_Expand() {
 	suite.Equal(v, string(content))
 }
 
-func (suite *EditTestSuite) TestGetOpenCmd_EditorSet() {
```
mitchell-as, line 103: Did you consider moving this test to your new package? If not, is there a way we could unit- or integration-test your new package?

mitchell-as (changes requested):

mitchell-as (approved):
Looks great!

PR #3290, opened 2024-05-07
Enable `state publish` integration tests.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2739" title="DX-2739" target="_blank">TaskDX-2739</a>  Enable `state publish` integration tests.
  </td></table>
  <br />
 

Publish under 'private/ActiveState-CLI-Testing/<UUID>' namespace for each test run.


test/integration/publish_int_test.go
```diff
@@ -462,6 +466,8 @@ authors:
 			}
 		})
 	}
+
+	ts.IgnoreLogErrors()
```
Naatan, line 470: What is this ignoring? Please add a comment if we truly want this.

Naatan (changes requested):
Looks ok to me, just one minor thing.

test/integration/publish_int_test.go
mitchell-as: There are test cases that test failures, like omitting a filename or being unable to edit a description. These expected failures will be written to the log and cause nightly integration failures.

We have similar precedent for other integration tests that induce failures and verify that they actually do fail (and gracefully).

Naatan (approved):

PR #3292, opened 2024-05-07
Fix installs not overriding previous installs 

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2771" title="DX-2771" target="_blank">TaskDX-2771</a>  Ensure installs can override previous installs
  </td></table>
  <br />
 

I believe that the underlying issue has to do with our detection of state tool installations. With this fixed we shouldn't even be checking for empty directories when the state tool is installed.

To detect state tool installation we compile a list of potential state tool binary locations and call `TargetExists` on them. https://github.com/ActiveState/cli/blob/master/cmd/state-installer/installer.go#L233

I wrote a small script to test the `os` functions that `TargetExists` uses. For the state tool binary one of these fails.

For reference, here is the test script and the output

```
package main

import (
	"fmt"
	"os"
)

func main() {
	stateToolPath := `C:\Users\Mike\AppData\Local\ActiveState\StateTool\release\bin\state.exe`

	_, err := os.Stat(stateToolPath)
	if err != nil {
		fmt.Println("os.Stat error:", err)
	}

	_, err = os.Readlink(stateToolPath)
	if err != nil {
		fmt.Println("os.Readlink error:", err)
	}

	_, err = os.Lstat(stateToolPath)
	if err != nil {
		fmt.Println("os.Lstat error:", err)
	}
}
```

```
C:\Users\Mike\work\cli>go run osTest\main.go
os.Readlink error: readlink C:\Users\Mike\AppData\Local\ActiveState\StateTool\release\bin\state.exe: The file or directory is not a reparse point.
```

We _could_ do some Windows-specific handling here if we feel it is necessary.

With this fixed I think the empty directory detection should work as expected which is only for fresh installs. The ACs state that we check for the install root file but that shouldn't be there in this case.

I've also updated the error to a non-input error so that we can get some Rollbar reports if this continues to happen.

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -292,11 +292,11 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 
 	// If this is a fresh installation we ensure that the target directory is empty
 	if !stateToolInstalled && fileutils.DirExists(params.path) && !params.force {
-		contains, err := fileutils.DirContains(params.path, installation.InstallDirMarker)
+		empty, err := fileutils.IsEmptyDir(params.path)
 		if err != nil {
 			return errs.Wrap(err, "Could not check if install path is empty")
 		}
-		if !contains {
+		if !empty {
 			return locale.NewError("err_install_nonempty_dir", "Installation path must be an empty directory: {{.V0}}", params.path)
 		}
 	}
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -297,7 +297,7 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 			return errs.Wrap(err, "Could not check if install path is empty")
 		}
 		if !empty {
-			return locale.NewError("err_install_nonempty_dir", "Installation path must be an empty directory: {{.V0}}", params.path)
+			return locale.NewInputError("err_install_nonempty_dir", "Installation path must be an empty directory: {{.V0}}", params.path)
 		}
 	}
 
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -292,12 +292,12 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 
 	// If this is a fresh installation we ensure that the target directory is empty
 	if !stateToolInstalled && fileutils.DirExists(params.path) && !params.force {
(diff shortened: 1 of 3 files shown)


Naatan (changes requested):
Really nice find @MDrakos! We should fix this in `TargetExists()` though, this could be biting us in other areas.

`TargetExists()` is meant to solve the problem of `FileExists()` returning false when the target is a symlink. I think we could just evaluate `FileExists()` first and if that is false then we continue with the regular logic of `TargetExists()`.

cmd/state-installer/cmd.go
```diff
@@ -297,7 +297,7 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 			return errs.Wrap(err, "Could not check if install path is empty")
 		}
 		if !empty {
-			return locale.NewInputError("err_install_nonempty_dir", "Installation path must be an empty directory: {{.V0}}", params.path)
+			return locale.NewError("err_install_nonempty_dir", "Installation path must be an empty directory: {{.V0}}", params.path)
```
Naatan, line 300: This is still an input error. The point being that this isn't reporting to and flooding rollbar. We'll have to make sure to follow up via Athena rather than rollbar.

Naatan (changes requested):

MDrakos:
> Really nice find @MDrakos! We should fix this in `TargetExists()` though, this could be biting us in other areas.
> 
> `TargetExists()` is meant to solve the problem of `FileExists()` returning false when the target is a symlink. I think we could just evaluate `FileExists()` first and if that is false then we continue with the regular logic of `TargetExists()`.

Sure, I think it makes sense to update `TargetExists` however I do wonder if we should be calling it from the installer. We control how the state tool binaries are installed and I don't believe we use any type of linking.

Naatan:
> > Really nice find @MDrakos! We should fix this in `TargetExists()` though, this could be biting us in other areas.
> > `TargetExists()` is meant to solve the problem of `FileExists()` returning false when the target is a symlink. I think we could just evaluate `FileExists()` first and if that is false then we continue with the regular logic of `TargetExists()`.
> 
> Sure, I think it makes sense to update `TargetExists` however I do wonder if we should be calling it from the installer. We control how the state tool binaries are installed and I don't believe we use any type of linking.

We only control how the binaries are installed most of the time. We also facilitate users to unpack the payload directly, mainly intended for automations. 

Regardless it's not unheard of for developers to get creative and use symlinks. This should be an easy enough use-case for us to facilitate if not for the windows specific issue you found.

PR #3295, opened 2024-05-08
Show failed artifact error messages and logs in `state artifacts`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2751" title="DX-2751" target="_blank">StoryDX-2751</a>  As a user I can run `state artifacts` and see what platforms/artifacts are failing
  </td></table>
  <br />
 

Sample:

image

### DIFF
--- a/internal/runners/artifacts/artifacts.go
+++ b/internal/runners/artifacts/artifacts.go
@@ -78,9 +78,12 @@ type structuredPlatform struct {
 }
 
 type structuredArtifact struct {
-	ID   string `json:"id"`
-	Name string `json:"name"`
-	URL  string `json:"url"`
+	ID     string `json:"id"`
+	Name   string `json:"name"`
+	URL    string `json:"url"`
+	status string
+	Errors []string `json:"errors,omitempty"`
+	LogURL string   `json:"logUrl,omitempty"`
 }
 
 func New(p primeable) *Artifacts {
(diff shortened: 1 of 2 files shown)


internal/runners/artifacts/artifacts.go
```diff
@@ -201,14 +207,19 @@ func (b *Artifacts) Run(params *Params) (rerr error) {
...
-			if artifact.URL == "" {
+			switch {
+			case len(artifact.Errors) > 0:
+				b.out.Print(fmt.Sprintf("  • %s ([ERROR]%s[/RESET])", artifact.Name, locale.T("artifact_status_failed")))
+				b.out.Print(fmt.Sprintf("    ├─ %s: [ERROR]%s[/RESET]", locale.T("artifact_status_failed_message"), strings.Join(artifact.Errors, ": ")))
+				b.out.Print(fmt.Sprintf("    └─ %s: [ACTIONABLE]%s[/RESET]", locale.T("artifact_status_failed_log"), artifact.LogURL))
+				continue
+			case artifact.status == types.ArtifactSkipped:
+				b.out.Print(fmt.Sprintf("  • %s ([NOTICE]%s[/RESET])", artifact.Name, locale.T("artifact_status_skipped")))
+				continue
```
mitchell-as, line 221: I added this because for the project I tested with, one artifact had a status of "SKIPPED" that was being shown as "Building ...", which was not accurate. Even though this wasn't in the mock or ACs, I went ahead and made things change.

rawktron:
> This doesn't look much like the mock because the error messages and log URLs are much longer than expected, resulting in the wrapping you see in the screenshot in the description. I'm putting this up for review now to get feedback on how to proceed.

I'm not sure there's much we can do here about the really long URLs for the logs? The information we're outputting is all correct. I wouldn't remove anything. And wrapping will be influenced by terminal width etc. Not really sure what options we have really.

Naatan:
I've filed https://activestatef.atlassian.net/browse/DX-2828 to follow up on the UI issue.

Naatan (approved):

PR #3296, opened 2024-05-09
Buildscripts refactor

Naatan:


pkg/buildscript/internal/buildexpression/buildexpression.go
```diff
@@ -177,27 +144,29 @@ func New(data []byte) (*BuildExpression, error) {
...
+				}
+		},
+		"runtime": {
+				"state_tool_artifacts": {
+						"src": "$sources"
+				}
+		},
+		"in": "$runtime"
+	}
+}`))
```
Naatan, line 169: Checked in with Martin and he agreed it made sense to update this so we're not invoking anything legacy.

pkg/platform/model/buildplanner/buildplanner.go
```diff
@@ -20,32 +14,6 @@ type client struct {
...
-		}
-
-		beData, err := json.MarshalIndent(v, "", "  ")
-		if err != nil {
-			logging.Error("Failed to marshal build expression: %s", err)
-			return
-		}
-		logging.Debug("Build Expression: %s", string(beData))
-	}
-}
```
Naatan, line 47: I've dropped this because we're no longer sending *buildexpression.BuildExpression. Instead we simply send the byte representation. This is because we want to keep build expressions internal to buildscripts as much as possible.

I've not replaced this because you can already get this data by logging API requests.

internal/runbits/buildscript/buildscript.go
```diff
@@ -1,4 +1,4 @@
-package buildscript
+package buildscript_runbit
```
Naatan, line 1: Went ahead and named this according to what we want to start naming these based on our recent discussion on duplicate package naming.

pkg/buildscript/buildscript.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/types"
+	"github.com/ActiveState/cli/pkg/platform/api/mono/mono_models"
+	"github.com/go-openapi/strfmt"
+)
+
+// BuildScript is what we want consuming code to work with. This specifically makes the raw presentation private as no
+// consuming code should ever be looking at the raw representation, instead this package should facilitate the use-case
+// of the consuming code through convenience methods that are easy to understand and work with.
+type BuildScript struct {
+	// buildexpression is what we do all our modifications on. We will be planning work to move this to the raw type
```
mitchell-as, line 19: Do we have the JIRA ticket number for this? It may be helpful to reference it here.

pkg/buildscript/buildscript.go
```diff
@@ -0,0 +1,179 @@
...
+	expr, err := buildexpression.Unmarshal(be)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not unmarshal build expression")
+	}
+
+	return &BuildScript{expr, raw.AtTime}, nil
+}
+
+// UnmarshalBuildExpression will create buildscript using an existing build expression
+func UnmarshalBuildExpression(b []byte, atTime *time.Time) (*BuildScript, error) {
```
mitchell-as, line 59: Please add a comment about the atTime parameter and why it's needed or optional for this function.

pkg/buildscript/buildscript.go
```diff
@@ -0,0 +1,179 @@
...
+func (b *BuildScript) AtTime() *time.Time {
+	return b.atTime
+}
+
+func (b *BuildScript) SetAtTime(t time.Time) {
+	b.atTime = &t
+}
+
+func (b *BuildScript) SetDefaultAtTime() error {
+	return b.buildexpression.SetDefaultAtTime()
```
mitchell-as, line 123: Do we need to set `b.atTime = nil` here, too?

pkg/buildscript/internal/raw/marshal.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, errs.Wrap(err, "Could not marshal build expression")
+	}
+
+	expr, err := buildexpression.Unmarshal(be)
+	if err != nil {
+		return nil, errs.Wrap(err, "Could not unmarshal build expression")
+	}
+
+	return []byte(marshalFromBuildExpression(expr, r.AtTime)), nil
+}
```
mitchell-as, line 31: exploding-head-emojii

pkg/buildscript/internal/raw/marshal.go
```diff
@@ -0,0 +1,101 @@
...
+func UnmarshalBuildExpression(expr *buildexpression.BuildExpression, atTime *time.Time) (*Raw, error) {
+	return Unmarshal([]byte(marshalFromBuildExpression(expr, atTime)))
+}
+
+// marshalFromBuildExpression is a bit special in that it is sort of an unmarshaller and a marshaller at the same time.
+// It takes a build expression and directly translates it into the string representation of a build script.
+// We should update this so that this instead translates the build expression directly to the in-memory representation
+// of a buildscript (ie. the Raw structure). But that is a large refactor in and of itself that'll follow later.
+// For now we can use this to convert a build expression to a buildscript with an extra hop where we have to unmarshal
+// the resulting buildscript string.
```
mitchell-as, line 71: Thanks for the comment.

pkg/platform/api/buildplanner/request/createproject.go
```diff
@@ -1,13 +1,11 @@
 package request
 
-import "github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
-
-func CreateProject(owner, project string, private bool, expr *buildexpression.BuildExpression, description string) *createProject {
+func CreateProject(owner, project string, private bool, expr []byte, description string) *createProject {
```
mitchell-as, line 3: It'd be more helpful from a debugging and logging perspective to have the `expr` argument as a string. What do you think about changing it to a string?

pkg/platform/api/buildplanner/request/stagecommit.go
```diff
@@ -4,10 +4,9 @@ import (
 	"time"
 
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
-	"github.com/ActiveState/cli/pkg/platform/runtime/buildexpression"
 )
 
-func StageCommit(owner, project, parentCommit, description string, atTime *time.Time, expression *buildexpression.BuildExpression) *buildPlanByStageCommit {
+func StageCommit(owner, project, parentCommit, description string, atTime *time.Time, expression []byte) *buildPlanByStageCommit {
```
mitchell-as, line 9: Same question: what do you think about making `expression` a string type instead of `[]byte`?

pkg/platform/model/buildplanner/build.go
```diff
@@ -74,12 +74,12 @@ func (b *BuildPlanner) FetchCommit(commitID strfmt.UUID, owner, project string,
 		return nil, errs.Wrap(err, "failed to unmarshal build plan")
 	}
 
-	expression, err := buildexpression.New(commit.Expression)
+	script, err := buildscript.UnmarshalBuildExpression(commit.Expression, ptr.To(time.Time(commit.AtTime)))
```
mitchell-as, line 77: At this point I question the `UnmarshalBuildExpression()` name because if I'm looking to create a buildscript from an incoming buildexpression (and atTime), the last thing that comes to mind is this name, and it's not something I'd see in an autocompletion list and go "oh, that's what I need". I prefer `NewFromBuildExpression()` or something like that. That's just me though. Maybe this is a Go convention or something similar.

pkg/platform/model/buildplanner/build.go
```diff
@@ -74,12 +74,12 @@ func (b *BuildPlanner) FetchCommit(commitID strfmt.UUID, owner, project string,
...
 	if err != nil {
-		return nil, errs.Wrap(err, "failed to parse build expression")
+		return nil, errs.Wrap(err, "failed to parse build script")
```
mitchell-as, line 79: nit: technically it's a failure parsing the incoming `commit.Expression`, which is a build expression, right?

pkg/platform/model/buildplanner/commit.go
```diff
@@ -30,24 +28,23 @@ type StageCommitParams struct {
 	// ... or commits can have an expression (e.g. from pull). When pulling an expression, we do not
 	// compute its changes into a series of above operations. Instead, we just pass the new
 	// expression directly.
```
mitchell-as, line 30: Please update this comment.

pkg/platform/model/buildplanner/project.go
```diff
@@ -57,11 +57,15 @@ func (b *BuildPlanner) CreateProject(params *CreateProjectParams) (strfmt.UUID,
 		}
 	}
 
+	exprB, err := script.MarshalBuildExpression()
```
mitchell-as, line 60: nit: can we make this just `expr`? Seeing `exprB` makes me think there's an `exprA` somewhere and wonder why we need a B.

mitchell-as (changes requested):
Overall this looks good to me. There are some minor things to address, but it shouldn't be too difficult.

internal/runbits/buildscript/file.go
```diff
@@ -57,42 +55,48 @@ func Initialize(path string, auth *authentication.Auth) error {
...
+
 	buildplanner := buildplanner.NewBuildPlannerModel(auth)
-	expr, atTime, err := buildplanner.GetBuildExpressionAndTime(commitId.String())
+	script, err = buildplanner.GetBuildScript(commitId.String())
 	if err != nil {
 		return errs.Wrap(err, "Unable to get the remote build expression and time")
 	}
-	script, err = NewFromBuildExpression(atTime, expr)
+
+	sb, err := script.Marshal()
```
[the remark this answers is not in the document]
Naatan: I've renamed it to `scriptBytes`. Note I agree with your point of view though I'd caveat that it's ok as long as the var is immediately consumed right after, because then the full context is in view.

internal/runbits/buildscript/file.go
```diff
@@ -57,42 +55,48 @@ func Initialize(path string, auth *authentication.Auth) error {
...
 
-func Update(proj projecter, atTime *strfmt.DateTime, newExpr *buildexpression.BuildExpression) error {
+func Update(proj projecter, newScript *buildscript.BuildScript) error {
 	script, err := ScriptFromProject(proj)
 	if err != nil {
 		return errs.Wrap(err, "Could not read build script")
 	}
 
-	newScript, err := NewFromBuildExpression(atTime, newExpr)
+	equals, err := script.Equals(newScript)
```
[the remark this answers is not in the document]
Naatan: I kinda like having the logic of this be owned by the buildscript package, even if it is straight forward. Plus having a single Equals() function means having a single error handler. Otherwise we'd need to handle errors for both marshals.

pkg/buildscript/buildscript.go
Naatan: Will add 👍

pkg/buildscript/buildscript.go
Naatan: I've renamed this function. It wasn't really setting any atTime value, it was merely forcing it to be set to `$at_time`, which isn't appropriate on the buildscript level.

pkg/buildscript/internal/raw/marshal.go
Naatan: Yeah this took a while to untangle, and it's still pretty confusing. I think ultimately this will be fixed by having the ascript package and by moving most of the interpretation logic to buildscripts rather than buildexpressions.

pkg/platform/api/buildplanner/request/createproject.go
Naatan: This has since been updated. The argument is still bytes, but what we send on the wire is a string.

pkg/platform/model/buildplanner/build.go
Naatan: I'd like to stick with the Marshal/Unmarshal terms, as that's been a pattern throughout these refactors. If this becomes a maintenance overhead we can revisit.

Naatan:
Test failures appear unrelated to this PR.

mitchell-as (approved):
Well done!

PR #3297, opened 2024-05-09
Audit integration tests

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2755" title="DX-2755" target="_blank">TaskDX-2755</a>  Audit integration test timing
  </td></table>
  <br />
 

This PR is mostly just removing runtime sourcing where it is no longer necessary. Unfortunately, there isn't much more at this time that we can do. I've left some comments on the Jira ticket for ideas to improve the integration test timing but those are out of score for this story.


test/integration/cve_int_test.go
```diff
@@ -56,10 +56,6 @@ func (suite *CveIntegrationTestSuite) TestCveNoVulnerabilities() {
 	cp := ts.Spawn("cve")
 	cp.Expect("No CVEs detected")
 	cp.ExpectExitCode(0)
-
-	cp = ts.Spawn("cve", "report") // legacy alias
-	cp.Expect("No CVEs detected")
-	cp.ExpectExitCode(0)
```
mitchell-as, line 62: I'm pretty sure we want to keep asserting that this works. Is there a specific reason you removed it?

test/integration/deploy_int_test.go
```diff
@@ -158,7 +158,6 @@ func (suite *DeployIntegrationTestSuite) TestDeployPython() {
 	defer ts.Close()
 
 	ts.SetupRCFile()
-	suite.T().Setenv("ACTIVESTATE_HOME", ts.Dirs.HomeDir)
```
mitchell-as, line 161: I don't understand why these are being removed in this file. I feel like they are important.

internal/testhelpers/e2e/session.go
```diff
@@ -562,8 +562,16 @@ func (s *Session) Close() error {
...
+	host := os.Getenv(constants.APIHostEnvVarName)
+	if host == "" {
+		host = constants.DefaultAPIHost
+	}
+	serviceURL.Host = strings.Replace(serviceURL.Host, string(api.ServiceMono)+api.TestingPlatform, host, 1)
+
 	for _, proj := range s.createdProjects {
-		err := model.DeleteProject(proj.Owner, proj.Project, auth)
+		params := projects.NewDeleteProjectParams().WithOrganizationName(proj.Owner).WithProjectName(proj.Project)
+		_, err := mono.Init(serviceURL, nil).Projects.DeleteProject(params, auth.ClientAuth())
```
mitchell-as, line 574: Would you please explain why this change is necessary? I can't grok it :(

test/integration/analytics_int_test.go
```diff
@@ -571,7 +571,7 @@ func (suite *AnalyticsIntegrationTestSuite) TestHeapEvents() {
 	)
 
 	cp.Expect("Creating a Virtual Environment")
-	cp.Expect("Activated", e2e.RuntimeSourcingTimeoutOpt)
```
mitchell-as, line 574: Is this really okay? We're doing an activate, so it will take time to source the runtime before printing "Activated".

test/integration/auth_int_test.go
```diff
@@ -34,11 +34,10 @@ func (suite *AuthIntegrationTestSuite) TestAuth() {
 	suite.OnlyRunForTags(tagsuite.Auth, tagsuite.Critical)
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
-	user := ts.CreateNewUser()
 	ts.LogoutUser()
-	suite.interactiveLogin(ts, user.Username, user.Password)
+	suite.interactiveLogin(ts, e2e.PersistentUsername, e2e.PersistentPassword)
 	ts.LogoutUser()
-	suite.loginFlags(ts, user.Username)
```
mitchell-as, line 41: I feel like there is value in authenticating a new user, rather than an existing one. Is there a specific reason you wanted to change this?

test/integration/manifest_int_test.go
```diff
@@ -50,13 +50,11 @@ func (suite *ManifestIntegrationTestSuite) TestManifest_JSON() {
 
 	cp := ts.SpawnWithOpts(
 		e2e.OptArgs("checkout", "ActiveState/cli#9eee7512-b2ab-4600-b78b-ab0cf2e817d8", "."),
-		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
 	)
 	cp.Expect("Checked out project", e2e.RuntimeSourcingTimeoutOpt)
```
mitchell-as, line 54: I think you can remove the `e2e.RuntimeSourcingTimeoutOpt` param now.

test/integration/push_int_test.go
```diff
@@ -96,6 +96,7 @@ func (suite *PushIntegrationTestSuite) TestInitAndPush() {
 	ts.LoginAsPersistentUser()
 
 	cp = ts.SpawnWithOpts(e2e.OptArgs("push", namespace))
+	cp.SendLine("y")
```
mitchell-as, line 99: Do we have a record of this failing previously? What exactly was it that caused you decide to add this?

test/integration/run_int_test.go
```diff
@@ -192,7 +192,6 @@ func (suite *RunIntegrationTestSuite) TestTwoInterrupts() {
 	suite.createProjectFile(ts, 3)
 
 	ts.LoginAsPersistentUser()
-	defer ts.LogoutUser()
```
mitchell-as, line 195: This seems reasonable to me. I don't understand the rationale for removing this.

test/integration/shell_int_test.go
```diff
@@ -215,7 +215,6 @@ func (suite *ShellIntegrationTestSuite) TestUseShellUpdates() {
 	defer ts.Close()
 
 	suite.SetupRCFile(ts)
-	suite.T().Setenv("ACTIVESTATE_HOME", ts.Dirs.HomeDir)
```
mitchell-as, line 218: I feel like there's a good reason for having this.

test/integration/shell_int_test.go
```diff
@@ -349,7 +348,6 @@ func (suite *ShellIntegrationTestSuite) TestPs1() {
 
 	cp = ts.SpawnWithOpts(
 		e2e.OptArgs("shell", "small-python"),
-		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
```
mitchell-as, line 352: Doesn't `state shell` require a runtime having been setup? I'm surprised this does not fail.

test/integration/use_int_test.go
```diff
@@ -126,7 +126,6 @@ func (suite *UseIntegrationTestSuite) TestReset() {
 	defer ts.Close()
 
 	ts.SetupRCFile()
-	suite.T().Setenv("ACTIVESTATE_HOME", ts.Dirs.HomeDir)
```
mitchell-as, line 129: I feel like there's a good reason for this.

mitchell-as (changes requested):
While there is some value in what you've done, I feel like a lot of this PR goes against the spirit of the ticket, which was to identify bottlenecks in test runtime performance. A good chunk of the changes do not appear related to this goal, and I don't understand why they are present. (See comments below.) Once I understand, then perhaps I can approve, but right now they seem out of scope to me, and possibly detremental in some cases.

test/integration/cve_int_test.go
MDrakos: I'm not sure what our policy is regarding legacy alias testing. I don't know that a CVE test is the appropriate place to test that aliases work.

test/integration/deploy_int_test.go
MDrakos: They are which is why this is set as part of the environment sandboxing.

test/integration/analytics_int_test.go
MDrakos: Whoops, this might have been a mistake.

test/integration/auth_int_test.go
MDrakos: Hmm, that's interesting... what specifically is the value you see in creating a new user for each test invocation? To me this is testing that we can log in and out.

test/integration/push_int_test.go
MDrakos: The test would fail because it was expecting input here. I will try to remove it and see what happens.

test/integration/run_int_test.go
MDrakos: Each integration test runs with its own copy and configuration of the state tool meaning that each tests starts unauthenticated. I believe this is the only instance in our test suite where we defer a logout which is why it was removed.

test/integration/shell_int_test.go
MDrakos: No, `shell` will still happily create a subshell without a runtime.

internal/testhelpers/e2e/session.go
MDrakos: I believe that I was running into an issue where the test suite was trying to delete a project form a `testing.tld` domain... let me try to remove it and see what happens.

test/integration/analytics_int_test.go
MDrakos: It turns out the diff isn't showing the lines above. We aren't actually sourcing a runtime in this test so I think this is appropriate.

MDrakos:
I believe I've addressed or responded to everything you've pointed out. The test failures should be unrelated to this PR but let me know if something stands out to you or if my comments don't make sense.

test/integration/deploy_int_test.go
mitchell-as: Ohhh, I see. I wasn't aware it was done elsewhere.

test/integration/shell_int_test.go
mitchell-as: I see. I didn't know that.

test/integration/auth_int_test.go
mitchell-as: I'm not sure. It's one of those things: if it's not broken, don't fix it.

test/integration/cve_int_test.go
mitchell-as: I wouldn't know where else to test something like this. I won't press the point further though.

mitchell-as (approved):
Okay, this works for me. We'll see if anything breaks over the weekend :)

test/integration/cve_int_test.go
MDrakos: Whether or not the aliasing works should be tested in it's own test, ideally a unit test. 

I did a search and didn't find it there so I will file a follow-up story for us to add them. Part of that story should be a discussion amongst us to determine if we want to enforce alias testing in our integration tests as an added check to prevent removal of these aliases.

MDrakos:
> Okay, this works for me. We'll see if anything breaks over the weekend :)

If anything does fail over the weekend feel free to let me know or assign me to the story. I'll take responsibility this time 😄

PR #3298, opened 2024-05-10
Do not consider locale-wrapped errors to be unexpected errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2770" title="DX-2770" target="_blank">TaskDX-2770</a>  Improve "execute failed" error
  </td></table>
  <br />
 

I discovered an error in the previous implementation when there are `locale.NewInputError()`s. In that case, you'd get a chain that includes "unexpected error", but that's not true.

### DIFF
--- a/internal/captain/command.go
+++ b/internal/captain/command.go
@@ -699,7 +699,10 @@ func (c *Command) cobraExecHandler(cobraCmd *cobra.Command, args []string) (rerr
 		}
 
 		if err := c.execute(c, args); err != nil {
-			return locale.WrapError(err, "unexpected_error", "Command failed due to unexpected error. For your convenience, this is the error chain:\n{{.V0}}", errs.JoinMessage(err))
+			if !locale.IsError(err) {
+				return locale.WrapError(err, "unexpected_error", "Command failed due to unexpected error. For your convenience, this is the error chain:\n{{.V0}}", errs.JoinMessage(err))
+			}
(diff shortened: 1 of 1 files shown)


MDrakos (changes requested):
There is nothing in the linked story that indicates why this was brought back for a second round. Could you add some details here or in the ticket? As it stands it looks like if the error isn't localized the user will still get a fairly unhelpful message.

mitchell-as:
I did mention in the PR description what the issue was. Perhaps it wasn't as clear as I thought. Sorry about that.

If there's a locale error (i.e. user input error), you would get something like

```
% ../../cli/build/state search foobar

█ Searching for: foobar

  Cannot obtain language for package search.
 x Cannot obtain language for package search.
```

You should only see that last line, not the first one because this is an expected error (it is handled and localized).

With this fix, you would get:

```
% ../../cli/build/state search foobar                     

█ Searching for: foobar

 x Cannot obtain language for package search.
```

A non-localized error will still give a proper error message. For example, here is an error I forced:

```
% ../cli/build/state checkout org/project                         
Checking out project: org/project

█ Something Went Wrong

  Checkout failed:
      Could not get path to use: this is an unexpected error.
```

MDrakos (approved):
Thanks for pointing that out and catching this!

PR #3299, opened 2024-05-10
Initial support for language-less Platform projects.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2797" title="DX-2797" target="_blank">StoryDX-2797</a>  As a User I can deploy a runtime with no language core
  </td></table>
  <br />
 

We can `state checkout` language-less Platform projects, but we cannot yet create them via `state init`. See my comment in the JIRA ticket as to why.

### DIFF
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -952,6 +952,8 @@ envdef_unmarshal_error:
     This indicates a problem with the build. You can try updating your project
 err_arg_required:
   other: "The following argument is required:\n  Name: [NOTICE]{{.V0}}[/RESET]\n  Description: [NOTICE]{{.V1}}[/RESET]"
+err_init_no_language:
+  other: "A project has not been set using 'state use'. You must include the [NOTICE]language[/RESET] argument. For more help, run '[ACTIONABLE]state init --help[/RESET]'."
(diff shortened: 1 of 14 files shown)


MDrakos (commented):
The ticket only appears to talk about the `state checkout` use case. I can see value in adding handling for `state languages` and other commands but I'm unsure why we're doing that in this PR. `state push` in particular seems odd. Not requesting any changes here just asking for some clarification as to why we had to touch these other runners.

mitchell-as:
Every caller of `model.FetchLanguageForCommit()` and `model.LanguageByCommit()` needed to be changed to account for the fact that an error is no longer returned if a project does not have a language defined for it. That is why the changes are a bit more broad than just for checkout.

Because I had to touch on all these bits, I went the extra mile to try and make sure they work with language-less projects. Unfortunately, `init` didn't work out, but oh well.

MDrakos (approved):
> Every caller of `model.FetchLanguageForCommit()` and `model.LanguageByCommit()` needed to be changed to account for the fact that an error is no longer returned if a project does not have a language defined for it. That is why the changes are a bit more broad than just for checkout.

That's the bit of context I was missing! Thank you.

PR #3300, opened 2024-05-13
Move change summaries out of the runtime package.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2776" title="DX-2776" target="_blank">TaskDX-2776</a>  Move change summaries out of runtime package
  </td></table>
  <br />

### DIFF
--- a/internal/runners/checkout/checkout.go
+++ b/internal/runners/checkout/checkout.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runbits/checkout"
+	"github.com/ActiveState/cli/internal/runbits/dependencies"
 	"github.com/ActiveState/cli/internal/runbits/git"
 	"github.com/ActiveState/cli/internal/runbits/runtime"
 	"github.com/ActiveState/cli/internal/subshell"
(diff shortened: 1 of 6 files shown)


mitchell-as:
Integration test failures do not appear to be caused by this PR, but I'm not 100% sure, as this involves invoking pieces of the runtime refactor.

internal/runners/checkout/checkout.go
```diff
@@ -111,9 +112,14 @@ func (u *Checkout) Run(params *Params) (rerr error) {
 		u.out.Notice(output.Title(locale.T("installing_runtime_title")))
 	}
 
-	rti, err := runtime.SolveAndUpdate(u.auth, u.out, u.analytics, proj, nil, target.TriggerCheckout, u.svcModel, u.config, runtime.OptMinimalUI)
+	rti, commit, err := runtime.Solve(u.auth, u.out, u.analytics, proj, nil, target.TriggerCheckout, u.svcModel, u.config, runtime.OptMinimalUI)
```
Naatan, line 115: Solve does not print the same UI as SolveAndUpdate, eg.

https://github.com/ActiveState/cli/blob/version/0-44-0-RC4/internal/runbits/runtime/refresh.go#L60-L73

So either we need to print that UI here, or we need to update Solve to address for this use-case (and ensure that this in itself doesn't break the UI of other use-cases of Solve).

As it stands this change would affect the checkout UX in ways that are not intended for this story.

Naatan (changes requested):

internal/runners/checkout/checkout.go
mitchell-as: I spent a lot of time looking into this and have come to the conclusion that we don't need that same UI because we're using the `OptMinimalUI` option, which turns off the code path you linked to. Also, we are printing the "Installing runtime" title for checkout here: https://github.com/ActiveState/cli/pull/3300/files#diff-e583b0a3c4a4b63cc78777f9d66d26cf335db3d1f15ba5a4f7a00680f79ef303R115. I did a checkout with v43 vs this branch and the UI looks the same to me unless I'm missing something subtle.

The same thing applies to init. I don't see any changes between v43 and this branch.

mitchell-as: Okay, I don't think I can do this without large scale changes.

There are some very subtle differences in output between SolveAndUpdate and Solve. For example, Solve outputs

```
```

whereas SolveAndUpdate outputs

```
Resolving Dependencies ✔ Done
```

It looks like Solve was designed to be called by requirements, and SolveAndUpdate was designed to be called by checkout, init, etc.

I've spent a lot of time on this and am burned out on it.

PR #3303, opened 2024-05-14
Fix autoupdate checks

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2824" title="DX-2824" target="_blank">BugDX-2824</a>  `state --version` after install doesn't show version available
  </td></table>
  <br />
 

This fixes a couple of issues but we should have some follow up work planned for a more robust fix.

I found that the cobra search command will return the parent command if it can't find an appropriate child, at least in the case of `state --version` it will strip the flag and return the parent command if there are no more arguments.

Since we want to run the autoupdate check for `state --version` we have to update the handling in `main.go`. In going through this I found that we have 3 separate checks for "Should we run autoupdate". These should be centralized however I didn't address this for the bug fix as this is targeting an RC4 that we would like to release soon.

Lastly, we check the modification time of the installed binary and will _not_ run the autoupdate if the file is less than 24 hours old. I did not modify that as part of this PR but I can if we no longer desire that behaviour.

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -231,22 +231,28 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	cmds.OnExecStart(msger.OnExecStart)
 	cmds.OnExecStop(msger.OnExecStop)
 
-	if childCmd != nil && !childCmd.SkipChecks() && !out.Type().IsStructured() {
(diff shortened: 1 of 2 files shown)


MDrakos:
Closing this as we will file follow-up work to address and clean up the root cause of the issue.

PR #3304, opened 2024-05-15
Ensure executables passed to `state exec` exist.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2809" title="DX-2809" target="_blank">TaskDX-2809</a>  state exec should only run executables
  </td></table>
  <br />
 

Sample:

```
% ../../cli/build/state exec foobar python
Operating on project org/project, located at /path/to/project.

 x The executable 'foobar' was not found in your PATH or in your project runtime.

█ Need More Help?
 • Run → 'state exec --help' for general help.
 • Run 'state export env' to check project runtime paths.
```


MDrakos (commented):
I'm not sure that dumping the users PATH is valuable here. Perhaps we could just say that we couldn't find the executable and to ask them to ensure it's available on `PATH`?

I'm open to keeping it as is, just want to consider alternatives that don't fill up the terminal.

mitchell-as:
I thought about that, but the runtime paths do not show up on the user's PATH outside of `state exec`, so they would not know ahead of time which runtime paths to search.

@Naatan do you have an opinion on how we should handle this?

Naatan:
I would simplify the error message to:

> The executable 'foobar' was not found on your PATH or in your project runtime.

From there the user can debug by inspecting their PATH or by using `state export env`.

Maybe we can add a hint indicating that they can run `state export env` to make that easier.

I agree we shouldn't list out all the paths we iterated over, that feels more confusing than anything.

mitchell-as:
Okay, see the updated example in the PR description. I updated the message and added a tip about `state export env`.

PR #3307, opened 2024-05-15
Disconnect change summary from runtime package

Naatan:


Naatan:
@mitchell-as note this is still largely your code changes, I just updated the UI portions.

Note I think you were probably probably going down a wrong rabbit hole with the async stuff. Async only refers to sourcing the runtime. Solving the runtime is not affected by async.

mitchell-as (approved):

PR #3309, opened 2024-05-16
Test output expectations should fail if the process has prematurely exited.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2782" title="DX-2782" target="_blank">TaskDX-2782</a>  TermTest: Process exiting should immediately fail any pending expects
  </td></table>
  <br />


Naatan (changes requested):
Looks like this is still the older commit. Just marking as changes requested so you have a way to notify me when it's updated.

mitchell-as:
Yep, I was waiting for approval in the termtest repo before updating this PR.

Naatan (approved):

PR #3312, opened 2024-05-17
Do not fail a disallowed envdef merge if the two vars are ultimately equivalent.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2808" title="DX-2808" target="_blank">BugDX-2808</a>  Updating python2 (camel) runtime leads to `cannot merge environment` error
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/envdef/environment.go
+++ b/pkg/platform/runtime/envdef/environment.go
@@ -247,11 +247,14 @@ func (ev EnvironmentVariable) Merge(other EnvironmentVariable) (*EnvironmentVari
 	case Disallowed:
 		if len(ev.Values) != 1 || len(other.Values) != 1 || (ev.Values[0] != other.Values[0]) {
 			sep := string(ev.Separator)
-			return nil, fmt.Errorf(
-				"cannot merge environment definitions: no join strategy for variable %s with values %s and %s",
-				ev.Name,
-				strings.Join(ev.Values, sep), strings.Join(other.Values, sep),
-			)
+			// It's possible that the merged env vars will still be equal, so only error if they are not.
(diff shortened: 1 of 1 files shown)


pkg/platform/runtime/envdef/environment.go
```diff
@@ -247,11 +247,14 @@ func (ev EnvironmentVariable) Merge(other EnvironmentVariable) (*EnvironmentVari
 	case Disallowed:
 		if len(ev.Values) != 1 || len(other.Values) != 1 || (ev.Values[0] != other.Values[0]) {
 			sep := string(ev.Separator)
-			return nil, fmt.Errorf(
-				"cannot merge environment definitions: no join strategy for variable %s with values %s and %s",
-				ev.Name,
-				strings.Join(ev.Values, sep), strings.Join(other.Values, sep),
-			)
+			// It's possible that the merged env vars will still be equal, so only error if they are not.
+			if strings.Join(ev.Values, sep) != strings.Join(other.Values, sep) {
```
Naatan, line 251: Couldn't we just drop the condition on line 248 and use this one instead?

Naatan (changes requested):

pkg/platform/runtime/envdef/environment.go
mitchell-as: I think we could, but that would require potentially expensive `strings.Join()` calls for each of these env vars. The previous check is potentially much faster.

If you'd like me to make this the conditional we test with, I'll do it, but I wanted to raise this concern.

Naatan: I see, no that's a good point.. let's go with this. 🙏

Naatan (approved):

PR #3313, opened 2024-05-20
Added new `state export runtime` command.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2773" title="DX-2773" target="_blank">TaskDX-2773</a>  As a user I should be able to programmatically receive the runtime dir
  </td></table>
  <br />


mitchell-as:
The lone test failure is not due to this PR. It's one of the usual timeouts.

internal/runners/export/runtime.go
```diff
@@ -0,0 +1,106 @@
...
+			return &ErrProjectNotFound{params.Path}
+		}
+	}
+	if proj == nil {
+		return rationalize.ErrNoProject
+	}
+
+	e.out.Notice(locale.Tr("export_runtime_statement", proj.NamespaceString(), proj.Dir()))
+
+	rt, err := runtime.SolveAndUpdate(e.auth, e.out, e.analytics, proj, nil, target.TriggerActivate, e.svcModel, e.cfg, runtime.OptMinimalUI)
```
MDrakos, line 70: We should probably add a new trigger for this. I don't believe this requires an update to the analytics doc but we should check to make sure.

internal/runners/export/runtime.go
```diff
@@ -0,0 +1,106 @@
...
+	execDir := setup.ExecDir(runtimeDir)
+
+	env, err := rt.Env(false, true)
+	if err != nil {
+		return errs.Wrap(err, "Could not get runtime environment")
+	}
+
+	tmpl, err := template.New("env").Parse(`{{- range $k, $v := .}}
+  - {{$k}}: {{$v}}
+{{- end}}`)
```
MDrakos, line 86: Nit: even though this is small we should consider adding this to the assets package to help with readability and be in line with other templates in our codebase.

MDrakos (changes requested):
Looks really good, just two minor things.

internal/runners/export/runtime.go
mitchell-as: Good catch, thanks!

MDrakos (approved):

PR #3314, opened 2024-05-21
Use PowerShell for shortcuts

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2678" title="DX-2678" target="_blank">TaskDX-2678</a>  Replace OLE shortcut mechanic on Windows
  </td></table>
  <br />

### DIFF
--- a/internal/osutils/autostart/autostart_windows.go
+++ b/internal/osutils/autostart/autostart_windows.go
@@ -38,11 +38,7 @@ func enable(exec string, opts Options) error {
 	if err != nil {
 		return errs.Wrap(err, "Could not set icon for shortcut file")
 	}
-
-	err = s.SetWindowStyle(shortcut.Minimized)
-	if err != nil {
-		return errs.Wrap(err, "Could not set shortcut to minimized")
-	}
+	s.SetWindowStyle(shortcut.Minimized)
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


internal/osutils/autostart/autostart_windows.go
```diff
@@ -38,11 +38,7 @@ func enable(exec string, opts Options) error {
 	if err != nil {
 		return errs.Wrap(err, "Could not set icon for shortcut file")
 	}
-
-	err = s.SetWindowStyle(shortcut.Minimized)
-	if err != nil {
-		return errs.Wrap(err, "Could not set shortcut to minimized")
-	}
+	s.SetWindowStyle(shortcut.Minimized)
```
Naatan, line 41: It seems like on Windows we always want to set this window style, so why make it an option at all?

Naatan (changes requested):
Looks good! Just one minor question.

internal/osutils/autostart/autostart_windows.go
MDrakos: By default new shortcuts will always be the `Normal` window style. This function is just to set it if you want `Minimized` or `Maximized`. It's largely still there because of the original implementation.

Naatan: I understand, what I'm asking is if there's still any point on exposing this and not just making this the default behaviour on windows? Just cause the original implementation did it doesn't mean we need to keep doing it, and this seems like an appropriate time to make such changes.

Anyway I just wanted to bring it up, you have more context around this than I do so I'll leave it to you.

Naatan (approved):

PR #3318, opened 2024-05-22
Do not update to merge commit if there was a buildscript conflict.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2842" title="DX-2842" target="_blank">BugDX-2842</a>  Merge commit and manual conflict created at same time
  </td></table>
  <br />
 

This partially reverts #3249 because that solution was based on the buggy behavior fixed in #3282.

### DIFF
--- a/internal/runners/pull/pull.go
+++ b/internal/runners/pull/pull.go
@@ -157,16 +157,16 @@ func (p *Pull) Run(params *PullParams) (rerr error) {
 	}
 
 	if commitID != *resultingCommit {
-		err := localcommit.Set(p.project.Dir(), resultingCommit.String())
-		if err != nil {
-			return errs.Wrap(err, "Unable to set local commit")
-		}
-
 		if p.cfg.GetBool(constants.OptinBuildscriptsConfig) {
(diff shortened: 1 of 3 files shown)


Naatan (changes requested):
I would expect to see extra logic here. There's two scenario's this needs to cover:

1. There was a conflict that needs to be manually resolved, in this case the commit should be set to the commit ID of the remote project (ie. not a newly created merge commit).
2. There was a conflict which was automatically resolved, in this case the commit should be set to the merge commit. If a buildscript conflict occurs then we are in scenario 1, not 2.

Only one of these scenarios can be true at a time.

I think your current changes are only facilitating scenario 2.

mitchell-as:
Yeah, I think you're right. At this point, I've done so many iterations of this that I'm easily confused. Hopefully I've gotten a hold of this.

mitchell-as:
The lone test failure is a performance unit test failure unrelated to this PR.

Naatan (approved):
I think you got it, but yeah pretty heady so we'll see if this passes by Mark 😅 Nice work!

PR #3321, opened 2024-05-23
Update `state eval` to trigger a build

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2784" title="DX-2784" target="_blank">TaskDX-2784</a>  Update `state eval` to trigger a build
  </td></table>
  <br />

### DIFF
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -467,3 +467,4 @@ events:
   - name: file-changed
     scope: ["internal/locale/locales"]
     value: build
+config_version: 1
--- a/internal/runners/eval/rationalize.go
+++ b/internal/runners/eval/rationalize.go
@@ -22,7 +22,7 @@ func rationalizeError(rerr *error) {
 	switch {
 	case errors.Is(*rerr, rationalize.ErrNotAuthenticated):
 		*rerr = errs.WrapUserFacing(*rerr,
-			locale.T("err_init_authenticated"),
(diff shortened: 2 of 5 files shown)


MDrakos:
The use of `request.evaluate` is here: https://github.com/ActiveState/cli/pull/3321/files#diff-41a09860e62af08a595fc2e0b51a86f51dadbf949c74e073bc71e17a298fcfceL685 just outside of the diff

Since we were previously using the `build(target)` query on the buildplanner this was all contained in the `PollBuildStatus` function. Now we are explicitly requesting the target be built and then polling the build status like we were before. I followed up with PB and they confirmed this was the correct way to do this.

Naatan:
> The use of `request.evaluate` is here: [#3321 (files)](https://github.com/ActiveState/cli/pull/3321/files#diff-41a09860e62af08a595fc2e0b51a86f51dadbf949c74e073bc71e17a298fcfceL685) just outside of the diff
> 
> Since we were previously using the `build(target)` query on the buildplanner this was all contained in the `PollBuildStatus` function. Now we are explicitly requesting the target be built and then polling the build status like we were before. I followed up with PB and they confirmed this was the correct way to do this.

ahh I see, I was looking on the bleeding edge branch where this got significantly refactored, and this request isn't invoked.

I'm afraid this means you'll probably have to repeat this implementation for v45, as I doubt this will merge cleanly :\

Naatan (approved):

PR #3322, opened 2024-05-24
Initial integration tests for changelog summaries.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2789" title="DX-2789" target="_blank">TaskDX-2789</a>  Introduce integration tests for change summaries
  </td></table>
  <br />
 

Note: there were some obvious failures that I could easily fix. There is one failure that will be addressed by DX-2851.

### DIFF
--- a/internal/runbits/dependencies/changesummary.go
+++ b/internal/runbits/dependencies/changesummary.go
@@ -21,24 +21,18 @@ const showUpdatedPackages = true
 // OutputChangeSummary looks over the given build plans, and computes and lists the additional
 // dependencies being installed for the requested packages, if any.
 func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan, oldBuildPlan *buildplan.BuildPlan) {
-	requested := newBuildPlan.RequestedArtifacts()
-	if len(requested) == 0 {
-		return // nothing to do
-	}
-
 	addedString := []string{}
 	addedLocale := []string{}
-	for _, i := range requested.Ingredients() {
-		v := fmt.Sprintf("%s@%s", i.Name, i.Version)
-		addedString = append(addedLocale, v)
-		addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
-	}
-
+	added := buildplan.Ingredients{}
 	dependencies := buildplan.Ingredients{}
 	directDependencies := buildplan.Ingredients{}
 	changeset := newBuildPlan.DiffArtifacts(oldBuildPlan, false)
 	for _, a := range changeset.Added {
+		added = append(added, a.Ingredients...)
 		for _, i := range a.Ingredients {
+			v := fmt.Sprintf("%s@%s", i.Name, i.Version)
+			addedString = append(addedLocale, v)
+			addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
 			dependencies = append(dependencies, i.RuntimeDependencies(true)...)
 			directDependencies = append(dependencies, i.RuntimeDependencies(false)...)
 		}
--- a/internal/runbits/dependencies/changesummary.go
+++ b/internal/runbits/dependencies/changesummary.go
@@ -18,23 +18,27 @@ import (
 // dependency numbers.
 const showUpdatedPackages = true
 
-// OutputChangeSummary looks over the given artifact changeset and attempts to determine if a single
-// package install request was made. If so, it computes and lists the additional dependencies being
-// installed for that package.
-// `artifacts` is an ArtifactMap containing artifacts in the changeset, and `filter` contains any
-// runtime requirements/artifacts already installed.
-func OutputChangeSummary(out output.Outputer, changeset *buildplan.ArtifactChangeset, alreadyInstalled buildplan.Artifacts) {
+// OutputChangeSummary looks over the given build plans, and computes and lists the additional
+// dependencies being installed for the requested packages, if any.
+func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan, oldBuildPlan *buildplan.BuildPlan) {
+	requested := newBuildPlan.RequestedArtifacts()
+	if len(requested) == 0 {
+		return // nothing to do
+	}
+
 	addedString := []string{}
 	addedLocale := []string{}
-	added := buildplan.Ingredients{}
+	for _, i := range requested.Ingredients() {
+		v := fmt.Sprintf("%s@%s", i.Name, i.Version)
+		addedString = append(addedLocale, v)
+		addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
+	}
+
 	dependencies := buildplan.Ingredients{}
 	directDependencies := buildplan.Ingredients{}
+	changeset := newBuildPlan.DiffArtifacts(oldBuildPlan, false)
 	for _, a := range changeset.Added {
-		added = append(added, a.Ingredients...)
 		for _, i := range a.Ingredients {
-			v := fmt.Sprintf("%s@%s", i.Name, i.Version)
-			addedString = append(addedLocale, v)
-			addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
 			dependencies = append(dependencies, i.RuntimeDependencies(true)...)
 			directDependencies = append(dependencies, i.RuntimeDependencies(false)...)
 		}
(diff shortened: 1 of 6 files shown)


mitchell-as:
Test failures are expected and will be fixed by tickets to-be-filed when there are nightly failures. I'm not filing them yet because this PR hasn't landed and could go through some churn. Note that fixing these tests is outside the scope of the ticket, but I did make some easy fixes here.

internal/runbits/dependencies/changesummary.go
```diff
@@ -18,23 +18,27 @@ import (
...
+	}
+
 	addedString := []string{}
 	addedLocale := []string{}
-	added := buildplan.Ingredients{}
+	for _, i := range requested.Ingredients() {
+		v := fmt.Sprintf("%s@%s", i.Name, i.Version)
+		addedString = append(addedLocale, v)
+		addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
+	}
```
Naatan, line 35: This doesn't seem appropriate. This would give you all requested ingredients for a buildplan, not just the ones that were added.

test/integration/checkout_int_test.go
```diff
@@ -387,6 +387,26 @@ func (suite *CheckoutIntegrationTestSuite) TestNoLanguage() {
...
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("checkout", "ActiveState-CLI/small-python"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
+	)
+	cp.Expect("Resolving Dependencies")
+	cp.Expect("Done")
+	cp.Expect("Setting up the following dependencies:")
+	cp.Expect("└─ python@3.10.10")
+	cp.SendCtrlC() // don't care what happens after
+	cp.ExpectNotExitCode(0)
```
Naatan, line 407: Would be nice if we could check that only one dependency was listed, since we've had some bugs where we're giving too many.

We could potentially check if the terminal snapshot has only a single occurrence of `─`.

test/integration/init_int_test.go
```diff
@@ -219,6 +219,30 @@ func (suite *InitIntegrationTestSuite) TestInit_InferredOrg() {
...
+		e2e.OptArgs("init", "ActiveState-CLI/"+project, "--language", "python@3.10.10"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
+	)
+	cp.Expect("Resolving Dependencies")
+	cp.Expect("Done")
+	ts.NotifyProjectCreated("ActiveState-CLI", project)
+	cp.Expect("Setting up the following dependencies:")
+	cp.Expect("└─ python@3.10.10")
+	cp.SendCtrlC() // don't care what happens after
+	cp.ExpectNotExitCode(0)
```
Naatan, line 243: Same as with checkout; would be good if we can assert only a single dependency was printed.

test/integration/checkout_int_test.go
```diff
@@ -387,6 +387,26 @@ func (suite *CheckoutIntegrationTestSuite) TestNoLanguage() {
...
+	if runtime.GOOS == "windows" {
+		suite.T().Skip("cp.SendCtrlC() does not work on Windows")
+	}
+	suite.OnlyRunForTags(tagsuite.Checkout)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("checkout", "ActiveState-CLI/small-python"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
```
Naatan, line 400: Suggest instead using async runtimes, so we don't actually need to source this runtime. We only need to resolve dependencies which async runtimes should be able to do.

test/integration/init_int_test.go
```diff
@@ -219,6 +219,30 @@ func (suite *InitIntegrationTestSuite) TestInit_InferredOrg() {
...
+	suite.OnlyRunForTags(tagsuite.Init)
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	ts.LoginAsPersistentUser()
+
+	project := "test-init-change-summary-" + sysinfo.OS().String()
+	cp := ts.SpawnWithOpts(
+		e2e.OptArgs("init", "ActiveState-CLI/"+project, "--language", "python@3.10.10"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
```
Naatan, line 235: As with checkout; suggest using async runtimes.

test/integration/package_int_test.go
```diff
@@ -755,6 +755,37 @@ func (suite *PackageIntegrationTestSuite) TestCVE_Indirect() {
...
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
+	cp.Expect("Successfully set")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnWithOpts(
+		e2e.OptArgs("checkout", "ActiveState-CLI/small-python", "."),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
```
Naatan, line 769: Async runtimes

test/integration/package_int_test.go
```diff
@@ -755,6 +755,37 @@ func (suite *PackageIntegrationTestSuite) TestCVE_Indirect() {
...
+	)
+	cp.Expect("Checked out", e2e.RuntimeSourcingTimeoutOpt)
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnWithOpts(
+		e2e.OptArgs("install", "requests@2.31.0"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
```
Naatan, line 776: Asy.. you get it :p

test/integration/package_int_test.go
```diff
@@ -755,6 +755,37 @@ func (suite *PackageIntegrationTestSuite) TestCVE_Indirect() {
...
+
+	cp = ts.SpawnWithOpts(
+		e2e.OptArgs("checkout", "ActiveState-CLI/small-python", "."),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
+	)
+	cp.Expect("Checked out", e2e.RuntimeSourcingTimeoutOpt)
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnWithOpts(
+		e2e.OptArgs("install", "requests@2.31.0"),
```
Naatan, line 775: Perhaps worth pinning the timestamp here also so this doesn't break with inventory changes.

Naatan (changes requested):

test/integration/package_int_test.go
mitchell-as: For whatever reason, if I remove this, and there is no runtime checked out, the subsequent `state install` does not show dependency and CVE information. It just immediately prints "Added package: <name>". So this has to stay, unfortunately.

test/integration/package_int_test.go
mitchell-as: I tried this, but for whatever reason, the command just immediately prints "Added package: ".

mitchell-as:
I've addressed what I can. This is beginning to turn into a time sink for the scope of the ticket. Hopefully we can move on from it soon.

internal/runbits/dependencies/changesummary.go
Naatan: Ok but we shouldn't be pushing something that's not fixing the issue, especially when it introduces its own issues. This is creating another bug on top of the existing one.

Suggest dropping this and maybe adding some context to the description of DX-2851.

test/integration/package_int_test.go
Naatan: I don't follow, this one is specifically enabling runtimes?

test/integration/package_int_test.go
Naatan: Did you file a bug?

Naatan (changes requested):

test/integration/package_int_test.go
mitchell-as: This test needs to enable runtimes, or else the code in the linked conditional will not fire, and you will not get any dependency information or CVE information.

test/integration/package_int_test.go
mitchell-as: DX-2864

internal/runbits/dependencies/changesummary.go
mitchell-as: https://activestatef.atlassian.net/browse/DX-2851?focusedCommentId=52777

test/integration/package_int_test.go
Naatan: Oh I see what you mean. So this IS using Async runtimes, but because our integration tests always disable runtime sourcing and it's instead an opt-in, we have to opt-in here because otherwise the CVE info isn't printed. 

I think that's fine for now. We have a story for combining async runtimes with the disabled runtime mechanic, in so far as that's reasonable. Additionally we have a story about having minimal runtimes. I suspect soon the disabled runtime functionality will be gone altogether.

Naatan (approved):

PR #3330, opened 2024-05-28
Do not report silent installer errors to Rollbar.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2862" title="DX-2862" target="_blank">BugDX-2862</a>  Silenced errors should not be reported to rollbar
  </td></table>
  <br />
 

There is a question of how we want to report silent errors to analytics. Right now, it reports "error" to the CatInstaller label, and "fail" to the CatInstallerFunnel label. If this is okay, then no further action is needed. Otherwise, please let me know.

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -215,7 +215,9 @@ func main() {
 	an.Event(anaConst.CatInstallerFunnel, "pre-exec")
 	err = cmd.Execute(processedArgs[1:])
 	if err != nil {
-		errors.ReportError(err, cmd, an)
+		if !errs.IsSilent(err) {
+			errors.ReportError(err, cmd, an)
+		}
 		if locale.IsInputError(err) {
 			an.EventWithLabel(anaConst.CatInstaller, "input-error", errs.JoinMessage(err))
 			logging.Debug("Installer input error: " + errs.JoinMessage(err))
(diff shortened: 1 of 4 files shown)


Naatan (changes requested):
I think it makes more sense to place this conditional here:

https://github.com/ActiveState/cli/blob/master/internal/runbits/errors/errors.go#L227-L227

That way we still get the analytic events, just not the rollbar one.

mitchell-as:
That's interesting. So should we file a ticket to change our error reporting in state/main.go? https://github.com/ActiveState/cli/blob/927624eb09bd6ea9c34260b468a422fe1044fc66/cmd/state/main.go#L255-L263 `errors.ReportError()` is not being called at all if it's a silent error.

Naatan:
Hmm yeah this is a bit of a mess right now. I'd like to refactor this altogether, but not right now for this story.

I see some obvious wins that we can make within the context of this story, on top of the current changes I would suggest:

- Drop the IsSilent here: https://github.com/ActiveState/cli/blob/master/cmd/state/main.go#L255-L255
	- This ought to be a concern of ReportError, which your current changeset is addressing.	
- Drop the manual reporting here: https://github.com/ActiveState/cli/blob/master/cmd/state-remote-installer/main.go#L166-L172
	- This is entirely redundant with what ReportError does.

Naatan (changes requested):
Looks good, but given our findings in our ticket I'd also like to make the changes outlined here:

https://github.com/ActiveState/cli/pull/3330#issuecomment-2136125707

Naatan (approved):

PR #3331, opened 2024-05-29
Test in-progress builds.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2698" title="DX-2698" target="_blank">TaskDX-2698</a>  We have integration tests for in-progress builds
  </td></table>
  <br />


internal/runbits/runtime/refresh.go
```diff
@@ -1,6 +1,8 @@
 package runtime
```
mitchell-as, line 1: Note: the changes in this file are in #3332.

test/integration/package_int_test.go
```diff
@@ -783,6 +784,56 @@ func (suite *PackageIntegrationTestSuite) TestChangeSummary() {
...
+	cp = ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
+	cp.Expect("Successfully set")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("checkout", "ActiveState-CLI/Perl-5.36", ".")
+	cp.Expect("Checked out")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnWithOpts(
+		e2e.OptArgs("install", "private/"+e2e.PersistentUsername+"/hello-world", "--ts", "now"),
```
mitchell-as, line 820: The JIRA ticket says to not use `--ts now`, but I think that is based on outdated behavior. I need `--ts now` to install the newest version just published in the lines above.

test/integration/package_int_test.go
```diff
@@ -783,6 +784,56 @@ func (suite *PackageIntegrationTestSuite) TestChangeSummary() {
...
+	)
+	cp.Expect("Package added: hello-world")
+	cp.ExpectExitCode(0)
+
+	cp = ts.SpawnWithOpts(
+		e2e.OptArgs("exec", "main"),
+		e2e.OptAppendEnv(constants.DisableRuntime+"=false"),
+	)
+	cp.Expect("Build Log")
+	cp.Expect("Building")
```
mitchell-as, line 831: The JIRA ticket is scant on details as to what exactly to test, so I decided to look for this.

mitchell-as:
Test failures are known issues.

internal/runbits/runtime/refresh.go
mitchell-as: Update: I've closed that PR. This PR will have these temporary changes until the runtime refactor mentioned in https://github.com/ActiveState/cli/pull/3332#issuecomment-2140882402 will fix this properly.

test/integration/package_int_test.go
```diff
@@ -783,6 +784,56 @@ func (suite *PackageIntegrationTestSuite) TestChangeSummary() {
...
+		"--depend", "builder/autotools-builder@>=0", // for ./configure, make, make install
+		"--depend", "internal/mingw-build-selector@>=0", // for Windows to use mingw's GCC
+		filepath.Join(osutil.GetTestDataDir(), "hello-world-1.0.0.tar.gz"),
+		"--edit") // publish a new revision each time, forcing a build
+	cp.Expect("Successfully published")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
+	cp.Expect("Successfully set")
+	cp.ExpectExitCode(0)
```
Naatan, line 813:
```suggestion
```

I don't think this should be in this test at all, unless you unset it again right before you run the install command. With this set there shouldn't be any build happening at all. The "fix" to refresh.go is just working around the problem you introduced here.

The checkout already has runtime disabled, and for install you specifically enable it again. You can drop this code and the changes to refresh.go and everything should work.

Naatan (changes requested):

test/integration/package_int_test.go
```diff
@@ -783,6 +784,56 @@ func (suite *PackageIntegrationTestSuite) TestChangeSummary() {
 	cp.ExpectExitCode(0)
 }
 
+func (suite *PackageIntegrationTestSuite) TestBuildInProgress() {
+	suite.OnlyRunForTags(tagsuite.Package)
```
Naatan, line 788: Let's move this to `runtime_int_test.go`. I don't think package is appropriate, this isn't doing anything with `state packages`.

internal/runbits/runtime/refresh.go
mitchell-as: Based on subsequent comments, this no longer applies.

test/integration/package_int_test.go
mitchell-as: Yep, that shows just how little I understand about the async mechanic. Based on your feedback I got it to work.

Naatan (approved):

PR #3332, opened 2024-05-29
Fix `state exec` not working when `optin.unstable.async_runtime` is enabled.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2868" title="DX-2868" target="_blank">BugDX-2868</a>  `state exec` throws nil pointer error when `optin.unstable.async_runtime` is enabled.
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/runtime/refresh.go
+++ b/internal/runbits/runtime/refresh.go
@@ -1,6 +1,8 @@
 package runtime
 
 import (
+	"strings"
+
 	"github.com/ActiveState/cli/internal/analytics"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
@@ -40,9 +42,8 @@ type Configurable interface {
 	GetBool(key string) bool
 }
 
-var overrideAsyncTriggers = map[target.Trigger]bool{
+var overrideAsyncTriggersMap = map[target.Trigger]bool{
 	target.TriggerRefresh:  true,
-	target.TriggerExec:     true,
 	target.TriggerActivate: true,
 	target.TriggerShell:    true,
 	target.TriggerScript:   true,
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
The issue here is that async is not returning a runtime.

Async = don't source runtime updates. It does not mean "don't even initialize a runtime".

The bug here is taht async is returning nil. It should return the runtime as-is, without doing any update.

I don't think `overrideAsyncTriggers` should exist at all.

mitchell-as:
Well I need a runtime update for my test in #3331, and this is the smallest change that makes it work. Since you've already refactored this, is it worth me trying to refactor it now? I don't even understand the whole async thing, let alone why `overrideAsyncTriggers` should or should not exist. I didn't write the original code. I only went through the debugger to figure out why it wasn't working for `state exec`.

Naatan:
I believe what should happen is:

* Move the async conditional on line 76 to right after the runtime constructed (line 82)
* Drop `overrideAsyncTriggersMap`

Sorry if I'm not being clear, I thought I was but it can sometimes be hard to have the right perspective.

Basically when we request a runtime we do two things; we retrieve the runtime as it exists on disk, and then if we find that state to be outdated with the commit ID requested we update it, and then return the updated runtime.

What async tells us is to not go through that update logic. Just return the runtime as it exists on disk. It does not mean "don't even give me a runtime".

mitchell-as:
That didn't work. I'm getting a weird error when running `state exec`:

```
Operating on project ActiveState-CLI/Perl-5.36, located at /tmp/3916057409/work.                                                            
                                                                                                                                                      
          █ Something Went Wrong                                                                                                                      
                                                                                                                                                      
           x Could not retrieve environment information for your runtime.                                                                             
           x Your installation seems corrupted.                                                                                                       
            Please try to re-run this command, as it may fix the problem.  If the problem persists, please report it in our forum:                    
          [https://community.activestate.com/c/state-tool/.](https://community.activestate.com/c/state-tool/)                                                                                            
           x Your installation seems to be corrupted.                                                                                                 
                                                                                                                                                      
            Cannot find runtime definition file at /tmp/3916057409/cache/8d7d4588/_runtime_store/runtime.json.                                        
            You can try to manually delete the cache directory and re-activating the project.
```

I'm reluctant to spend time trying to debug and develop this further considering it will be throwaway code.

mitchell-as:
I'm going to close this. I only opened it so that we could track that `state exec` was broken for async runtimes. Since an upcoming runtime refactor will fix this, I think this PR is pointless.

Naatan:
Ah if it's only for async runtimes then yeah no harm waiting.

PR #3344, opened 2024-06-05
Report build success on heartbeat if we're not waiting for any more artifacts.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2874" title="DX-2874" target="_blank">BugDX-2874</a>  Build log is bogus
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/runtime/setup/buildlog/buildlog.go
+++ b/pkg/platform/runtime/setup/buildlog/buildlog.go
@@ -160,12 +160,19 @@ func NewWithCustomConnections(artifactMap buildplan.ArtifactIDMap,
 			return result
 		}
 
-		buildSuccess := func() {
-			if err := writeLogFile("", "Build Succeeded"); err != nil {
+		anyFailures := false
+		buildFinished := func() {
+			status := "Succeeded"
+			var event events.Eventer = events.BuildSuccess{}
+			if anyFailures {
+				status = "Failed"
+				event = events.BuildFailure{}
+			}
+			if err := writeLogFile("", "Build "+status); err != nil {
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
I think this is signalling a greater issue. Which is that we're not instrumenting `buildSuccess()` properly.

It seems `stillWaiting()` is based on `artifactsDone()`, which gets populated by both the `ArtifactSucceeded` as well as the `ArtifactFailed` events. So already that's a bug; we cannot say that `len(stillWaiting())==0` means the build was successful, only that we're "done".

I suggest updating `buildSuccess()` to `buildFinished()` as well as the terms it logs to the buildlog to say "finished" rather than "successful".

And we should have the conditional logic you introduced on heartbeat here instead be on `ArtifactFailed`, because waiting till the heartbeat doesn't feel necessary.

pkg/platform/runtime/setup/buildlog/buildlog.go
```diff
@@ -160,12 +160,19 @@ func NewWithCustomConnections(artifactMap buildplan.ArtifactIDMap,
...
-		buildSuccess := func() {
-			if err := writeLogFile("", "Build Succeeded"); err != nil {
+		anyFailures := false
+		buildFinished := func() {
+			status := "Succeeded"
+			var event events.Eventer = events.BuildSuccess{}
+			if anyFailures {
+				status = "Failed"
+				event = events.BuildFailure{}
+			}
```
Naatan, line 170: Nice!

Naatan (approved):

PR #3345, opened 2024-06-06
Consider a build ready if all of its installed artifacts are.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2874" title="DX-2874" target="_blank">BugDX-2874</a>  Build log is bogus
  </td></table>
  <br />
 

If a build is not ready, but all artifacts to install are, the buildlog streamer will be invoked and wait unnecessarily.


mitchell-as:
Mark reported on Slack that he installed this build and it fixed the failure he was able to reproduce.

Naatan (approved):
Nice!

PR #3350, opened 2024-06-10
Remove unused `state tutorial` and `state _ppm`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2866" title="DX-2866" target="_blank">TaskDX-2866</a>  Get rid of `state tutorial` and `state _ppm`
  </td></table>
  <br />


Naatan (changes requested):
We should be able to get rid of `runners/ppm` and `runbits/invoke.go` as well.

Please do an audit of anything else referenced in the ppm and tutorial runners for anything else we may be able to get rid of.

mitchell-as:
`runners/invoke.go` is used by `runners/protocol/protocol.go`, so it needs to remain. The `runners/ppm` package can go though. I had originally left it alone because it looked like it was in use, but that was not the case. I'm not sure why the linter didn't pick it up either.

Naatan (changes requested):
Let's refactor the protocol.go code. It seems to run InvokeSilent which is basically a really lightweight wrapper around `exec.Command`. We could just run `exec.Command` directly, or use `exeutils.ExecSimple()`.

Additionally, I think we can drop these tests/assertions:

* https://github.com/ActiveState/cli/blob/DX-2721/test/integration/analytics_int_test.go#L383-L383
* https://github.com/ActiveState/cli/blob/DX-2721/test/integration/activate_int_test.go#L410-L410

mitchell-as:
Test failures due to the usual timeouts, not anything related to this PR.

mitchell-as:
I did my best to identify anything related to PPM and remove it, but please let me know if I missed any more stuff.

internal/runners/protocol/protocol.go
```diff
@@ -60,5 +61,11 @@ func (p *Protocol) Run(params Params) error {
 		return locale.NewError("err_protocol_flag", "Invalid URL fragment, the only supported URL fragment is 'replace'")
 	}
 
-	return runbits.InvokeSilent("activate", fmt.Sprintf("--%s", parsed.Fragment), namespace.String())
+	// Execute state command
+	exe, err := os.Executable()
+	if err != nil {
+		return locale.WrapError(err, "err_protocol_os_executable", "Could not detect executable path of State Tool.")
+	}
+	_, _, err = osutils.ExecSimple(exe, []string{"activate", fmt.Sprintf("--%s", parsed.Fragment), namespace.String()}, nil)
```
Naatan, line 69:
```suggestion
	_, _, err = osutils.ExecuteAndPipeStd(exe, []string{"activate", fmt.Sprintf("--%s", parsed.Fragment), namespace.String()},[]string{})
```

Sorry ExecSimple was a bad call on my end. InvokeSilent pipes os.Stdout and os.Stderr to the cmd, whereas ExecSimple does not. We want to make sure here that we still pipe stdout and stderr, otherwise the end-user won't see anything.

Naatan (changes requested):

internal/runners/protocol/protocol.go
mitchell-as: Oh weird, maybe it's a Linux/macOS thing. I tested on Windows and it worked just fine. Regardless, I've made the change. Thanks for pointing it out.

Naatan: Huh, no idea what is happening on Windows. Strange..

Naatan (approved):

PR #3352, opened 2024-06-11
Mark client buildplanner API errors as input errors, not external errors.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2877" title="DX-2877" target="_blank">TaskDX-2877</a>  state init should give error about invalid formatting
  </td></table>
  <br />
 

Sample:

```
% ../../cli/build/state init mitchell-as/foo/bar --language python
Initializing Project

 x Failed to create project: Encountered unexpected error: Request failed: graphql: 
mono_api.create_project: Project name can only contain alphanumeric, '.' or '-' characters.
```

### DIFF
--- a/pkg/platform/model/buildplanner/build.go
+++ b/pkg/platform/model/buildplanner/build.go
@@ -28,8 +28,7 @@ const (
 	pollTimeout        = 30 * time.Second
 	buildStatusTimeout = 24 * time.Hour
 
-	codeExtensionKey   = "code"
-	statusExtensionKey = "status"
+	codeExtensionKey = "code"
 )
 
 type Commit struct {
@@ -118,11 +117,6 @@ func processBuildPlannerError(bpErr error, fallbackMessage string) error {
 		code, ok := graphqlErr.Extensions[codeExtensionKey].(string)
 		if ok && code == clientDeprecationErrorKey {
 			return &response.BuildPlannerError{Err: locale.NewExternalError("err_buildplanner_deprecated", "Encountered deprecation error: {{.V0}}", graphqlErr.Message)}
-		}
-		if status, ok := graphqlErr.Extensions[statusExtensionKey].(float64); ok {
-			if status >= 400 && status < 500 {
-				return locale.NewInputError("err_buildplanner_api_input_error", "{{.V0}}: {{.V1}}", fallbackMessage, bpErr.Error())
-			}
 		}
 	}
 	return &response.BuildPlannerError{Err: locale.NewExternalError("err_buildplanner", "{{.V0}}: Encountered unexpected error: {{.V1}}", fallbackMessage, bpErr.Error())}
--- a/pkg/platform/model/buildplanner/build.go
+++ b/pkg/platform/model/buildplanner/build.go
@@ -28,7 +28,8 @@ const (
 	pollTimeout        = 30 * time.Second
 	buildStatusTimeout = 24 * time.Hour
 
-	codeExtensionKey = "code"
+	codeExtensionKey   = "code"
+	statusExtensionKey = "status"
 )
 
 type Commit struct {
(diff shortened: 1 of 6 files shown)


mitchell-as:
Note: the buildplanner returns this error, which ends up being marked as a `locale.NewExternalError()`:

```
("*github.com/ActiveState/graphql.GraphErr")(0xc0008020d8)
*github.com/ActiveState/graphql.GraphErr {
	Message: "mono_api.create_project: Project name can only contain alphanume...+26 more",
	Extensions: map[string]interface {} [
		"status": *(*interface {})(0xc0007ca088), 
	],}
```

status is 400 (float64).

We don't seem to have anything in place to handle errors with status codes, and I'm not sure if or how we should start doing so.

Naatan (changes requested):
I don't understand how this changeset is meant to address the issue. You removed a localization error, but are still wrapping the same error. The only effects I could see this have is:

1. It has even less information reported (minus one localization error)
2. It now dumps the raw error rather than the localization error, and reports a critical error to rollbar telling us to fix the localization on this error.

What we have usually done when we want to forward API errors is something like:

```
locale.WrapInputError(err, "Could not create project: {{.V0}}", err.Error())
```

But only if we're certain the `err` we're dealing with holds the info we want to report.

mitchell-as:
It's an external error, so it's not reportable. I take your point though, and will try to figure out how to handle this type of external error better.

mitchell-as:
Okay, I've improved the handling around these types of API errors by marking client errors as input errors. See the updated example in the PR description for a real-world failure case.

pkg/platform/model/buildplanner/build.go
```diff
@@ -118,6 +119,11 @@ func processBuildPlannerError(bpErr error, fallbackMessage string) error {
 		if ok && code == clientDeprecationErrorKey {
 			return &response.BuildPlannerError{Err: locale.NewExternalError("err_buildplanner_deprecated", "Encountered deprecation error: {{.V0}}", graphqlErr.Message)}
 		}
+		if status, ok := graphqlErr.Extensions[statusExtensionKey].(float64); ok {
+			if status >= 400 && status < 500 {
+				return locale.NewInputError("err_buildplanner_api_input_error", "{{.V0}}: {{.V1}}", fallbackMessage, bpErr.Error())
```
Naatan, line 124: I don't think `err_buildplanner_api_input_error` exists. Least not finding it in my checkout.

pkg/platform/model/buildplanner/build.go
```diff
@@ -118,6 +119,11 @@ func processBuildPlannerError(bpErr error, fallbackMessage string) error {
 		if ok && code == clientDeprecationErrorKey {
 			return &response.BuildPlannerError{Err: locale.NewExternalError("err_buildplanner_deprecated", "Encountered deprecation error: {{.V0}}", graphqlErr.Message)}
 		}
+		if status, ok := graphqlErr.Extensions[statusExtensionKey].(float64); ok {
```
Naatan, line 122: I think the `ok` here is asserting `(float64)`. Couldn't it panic if the `statusExtensionKey` key doesn't exist?

pkg/platform/model/buildplanner/build.go
```diff
@@ -118,6 +119,11 @@ func processBuildPlannerError(bpErr error, fallbackMessage string) error {
...
+			if status >= 400 && status < 500 {
+				return locale.NewInputError("err_buildplanner_api_input_error", "{{.V0}}: {{.V1}}", fallbackMessage, bpErr.Error())
+			}
+		}
 	}
 	return &response.BuildPlannerError{Err: locale.NewExternalError("err_buildplanner", "{{.V0}}: Encountered unexpected error: {{.V1}}", fallbackMessage, bpErr.Error())}
```
Naatan, line 128: Wait so we were reporting this? It seems like external error is bugged then.

None of these changes are necessary I think. This sounds like a bug in reporting of external errors. External errors should still be shown to the user, external just means that we don't report it to rollbar. I think the fix here should be in our error printing of external errors.

Naatan (changes requested):

pkg/platform/model/buildplanner/build.go
mitchell-as: Right, it does not exist, so I'm using the `"{{.V0}}: {{.V1}}"` fallback string you see here.

pkg/platform/model/buildplanner/build.go
mitchell-as: Nope, I tried it in the playground. It doesn't panic.

pkg/platform/model/buildplanner/build.go
mitchell-as: Ugh yeah, it was bugged :(

Naatan (approved):

PR #3354, opened 2024-06-11
If the remote installer could not get an install payload, but no error happened, show a helpful error message.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2887" title="DX-2887" target="_blank">BugDX-2887</a>  Remote installer panic
  </td></table>
  <br />
 

See https://activestatef.atlassian.net/browse/DX-2887?focusedCommentId=53288 for the potential cause of `availableUpdate` being `nil`.

Note that we the logs will likely point to the source of failure, so we don't have to do any speculative logging.

### DIFF
--- a/cmd/state-remote-installer/main.go
+++ b/cmd/state-remote-installer/main.go
@@ -195,7 +195,7 @@ func execute(out output.Outputer, prompt prompt.Prompter, cfg *config.Instance, 
 	if err != nil {
 		return errs.Wrap(err, "Could not retrieve install package information")
 	} else if availableUpdate == nil {
-		return locale.NewError("remote_install_no_available_update", "Could not find package to install. Please try again later.")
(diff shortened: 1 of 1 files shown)


cmd/state-remote-installer/main.go
```diff
@@ -194,6 +194,8 @@ func execute(out output.Outputer, prompt prompt.Prompter, cfg *config.Instance,
 	availableUpdate, err := checker.CheckFor(channel, params.version)
 	if err != nil {
 		return errs.Wrap(err, "Could not retrieve install package information")
+	} else if availableUpdate == nil {
+		return locale.NewError("remote_install_no_available_update", "Could not find package to install. Please try again later.")
```
Naatan, line 198:
```suggestion
		return locale.NewError("remote_install_no_available_update", "Could not find installer to download. This could be due to networking issues or temporary maintenance. Please try again later.")
```

Naatan (changes requested):

Naatan (approved):

PR #3355, opened 2024-06-12
Fixed in-progress build detection.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2894" title="DX-2894" target="_blank">BugDX-2894</a>  Nightly failure: TestRuntimeIntegrationTestSuite/TestBuildInProgress
  </td></table>
  <br />
 

Artifacts can only be considered ready for install if they have a URL.

### DIFF
--- a/pkg/buildplan/filters.go
+++ b/pkg/buildplan/filters.go
@@ -65,9 +65,3 @@ func FilterSuccessfulArtifacts() FilterArtifact {
 			a.Status == types.ArtifactReady
 	}
 }
-
-func FilterInstallableArtifacts() FilterArtifact {
-	return func(a *Artifact) bool {
-		return a.URL != ""
-	}
-}
--- a/pkg/buildplan/filters.go
+++ b/pkg/buildplan/filters.go
@@ -65,3 +65,9 @@ func FilterSuccessfulArtifacts() FilterArtifact {
 			a.Status == types.ArtifactReady
 	}
 }
+
+func FilterInstallableArtifacts() FilterArtifact {
+	return func(a *Artifact) bool {
+		return a.URL != ""
+	}
+}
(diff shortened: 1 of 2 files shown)


pkg/buildplan/filters.go
```diff
@@ -65,3 +65,9 @@ func FilterSuccessfulArtifacts() FilterArtifact {
 			a.Status == types.ArtifactReady
 	}
 }
+
+func FilterInstallableArtifacts() FilterArtifact {
+	return func(a *Artifact) bool {
+		return a.URL != ""
+	}
+}
```
Naatan, line 73: I think we should just add this logic to:

https://github.com/ActiveState/cli/blob/f42d73c40e6d91def0163785de4eeb9b8cac9568/pkg/buildplan/filters.go#L45

Naatan (changes requested):

pkg/buildplan/filters.go
mitchell-as: I don't think so. For example, the change summary filters using `FilterInstallableArtifacts()` before sourcing/building/installing the runtime. With your proposal, anything not built yet would be excluded from the change summary.

The file your link points to looks very different from mine, signaling a pretty substantial refactor. I see you have a `FilterNeedsBuild()` filter. That could be useful, but I've run into cases where there's a "READY" status but the URL is not yet set.

I could just add a lambda func to setup.go that ensures there's a URL set. I'm afraid to increase the scope of the URL check because it might inadvertantly affect other things like change summaries.

Naatan: Hmm ok. What further complicates this is the refactor. This is a non-issue with the new implementation as it's doing a much more thorough job of calculating what is to be build/installed in the first place, and it uses this for the ready check. So the issue you're running into here doesn't exist anymore.

Let's make it a lambda in setup.go, since that will get deleted anyway.

Naatan (changes requested):

Naatan (approved):

PR #3356, opened 2024-06-13
State Tool does not call login API each invocation; uses JWT from svc

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2895" title="DX-2895" target="_blank">TaskDX-2895</a>  State Tool pulls JWT from state-svc
  </td></table>
  <br />
 

Please be extra thorough reviewing this. It's going into a patch release that we'll want to get out asap, and due to the long lifetime of tokens it's difficult to manually test all of this (though I have).


pkg/platform/model/svc.go
```diff
@@ -176,6 +177,29 @@ func (m *SvcModel) GetProcessesInUse(ctx context.Context, execDir string) ([]*gr
...
+	if err := m.request(ctx, r, &resp); err != nil {
+		return nil, errs.Wrap(err, "Error sending messages request")
+	}
+
+	jwt := mono_models.JWT{}
+	err := json.Unmarshal(resp.Payload, &jwt)
+	if err != nil {
+		return nil, errs.Wrap(err, "Error unmarshaling JWT")
+	}
+	logging.Debug("Response: %#v\n", jwt)
```
mitchell-as, line 198: Hmm, I've never seen this kind of format parameter before. They're getting more and more complex...

mitchell-as (dismissed):
Looks good to me. Maybe you want to get a second set of eyes to review it though.

pkg/platform/model/svc.go
Naatan: I should've dropped that, that was me debugging something.

Fwiw this variant of `%v` gives you more information about the structure, and not just its values.

https://gobyexample.com/string-formatting

Naatan:
@MDrakos flagged you for review as well since this is hard to manually test and this is going out in a patch release next week, so code review is extra important on this one.

MDrakos (approved):
This looks good to me too. There are some ACs from the story that I'm not sure were hit. Do we still want to keep them?

Namely the service pro actively grabbing the JWT. Right now it looks like that's only happening when the state tool makes a request to the service.

Naatan:
> Namely the service pro actively grabbing the JWT. Right now it looks like that's only happening when the state tool makes a request to the service.

Yeah I ended up not addressing that, when I wrote that I thought the JWT TTL was 15 minutes, but it's actually 1 hour. I could keep refreshing for a few hours, but I felt that this wasn't adding much value and could in fact by itself cause the API to still be hammered needlessly.

PR #3357, opened 2024-06-14
Update build expression to use state tool artifacts v1

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2902" title="DX-2902" target="_blank">TaskDX-2902</a>  Update default build expression to use `state_tool_artifacts_v1` instead of `state_tool_artifacts`
  </td></table>
  <br />

### DIFF
--- a/test/integration/init_int_test.go
+++ b/test/integration/init_int_test.go
@@ -24,30 +24,30 @@ type InitIntegrationTestSuite struct {
 
 func (suite *InitIntegrationTestSuite) TestInit() {
 	suite.OnlyRunForTags(tagsuite.Init, tagsuite.Critical)
-	suite.runInitTest(false, "python", "python3")
+	suite.runInitTest(false, true, "python", "python3")
 }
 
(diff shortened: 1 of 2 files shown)


test/integration/init_int_test.go
```diff
@@ -60,11 +60,20 @@ func (suite *InitIntegrationTestSuite) runInitTest(addPath bool, lang string, ex
 		computedArgs = append(computedArgs, ts.Dirs.Work)
 	}
 
+	env := []string{}
+	if sourceRuntime {
+		env = append(env, constants.DisableRuntime+"=false")
```
mitchell-as, line 65: I see this is targeted at v45, so when you merge this into v46, I think we'll want to flip the conditional (`if !sourceRuntime { env = append(env, constants.DisableRuntime+"=true") }`)

mitchell-as (approved):

test/integration/init_int_test.go
Naatan: Will update it in the runtime refactor PR.

PR #3358, opened 2024-06-17
Centralize condition for auto update checking.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2836" title="DX-2836" target="_blank">TaskDX-2836</a>  Improve handling of `Update Available` messaging
  </td></table>
  <br />
 

Also fix `command.Find()` to return `nil` if no child was found and rename it `FindChild()` for clarity.

Note: one of the checks listed for centralization in the ticket, https://github.com/ActiveState/cli/blob/87205e082c99853af80009361b73dbe1ef4cd351/cmd/state/autoupdate.go#L56, actually needs to be separate because its conditional (https://github.com/ActiveState/cli/blob/87205e082c99853af80009361b73dbe1ef4cd351/internal/updater/updater.go#L122-L126) depends on the update's contents. The conditional result cannot be determined prior to downloading the update.

### DIFF
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -242,7 +242,7 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	}
 
 	// Check to see if this state tool version is different from the lock version.
-	if (childCmd == nil || !childCmd.SkipChecks() || childCmd.Name() != "update") && pj != nil && pj.IsLocked() {
+	if (childCmd == nil || !childCmd.SkipChecks()) && pj != nil && pj.IsLocked() {
 		if (pj.Version() != "" && pj.Version() != constants.Version) ||
 			(pj.Channel() != "" && pj.Channel() != constants.ChannelName) {
 			return errs.AddTips(
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -225,7 +225,7 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	// Run the actual command
 	cmds := cmdtree.New(primer.New(pj, out, auth, prompter, sshell, conditional, cfg, ipcClient, svcmodel, an), args...)
 
-	childCmd, err := cmds.Command().Find(args[1:])
+	childCmd, err := cmds.Command().FindChild(args[1:])
 	if err != nil {
 		logging.Debug("Could not find child command, error: %v", err)
 	}
--- a/cmd/state/main.go
+++ b/cmd/state/main.go
@@ -234,23 +234,22 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
 	cmds.OnExecStart(msger.OnExecStart)
 	cmds.OnExecStop(msger.OnExecStop)
 
-	if childCmd != nil && !childCmd.SkipChecks() && !out.Type().IsStructured() {
-		// Auto update to latest state tool version
-		if updated, err := autoUpdate(svcmodel, args, cfg, an, out); err == nil && updated {
-			return nil // command will be run by updated exe
-		} else if err != nil {
-			multilog.Error("Failed to autoupdate: %v", err)
-		}
+	// Auto update to latest state tool version if possible.
+	if updated, err := autoUpdate(svcmodel, args, childCmd, cfg, an, out); err == nil && updated {
+		return nil // command will be run by updated exe
+	} else if err != nil {
+		multilog.Error("Failed to autoupdate: %v", err)
+	}
 
-		if childCmd.Name() != "update" && pj != nil && pj.IsLocked() {
-			if (pj.Version() != "" && pj.Version() != constants.Version) ||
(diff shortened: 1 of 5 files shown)


cmd/state/main.go
```diff
@@ -234,23 +234,22 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
...
-		if childCmd.Name() != "update" && pj != nil && pj.IsLocked() {
-			if (pj.Version() != "" && pj.Version() != constants.Version) ||
-				(pj.Channel() != "" && pj.Channel() != constants.ChannelName) {
-				return errs.AddTips(
-					locale.NewInputError("lock_version_mismatch", "", pj.Source().Lock, constants.ChannelName, constants.Version),
-					locale.Tr("lock_update_legacy_version", constants.DocumentationURLLocking),
-					locale.T("lock_update_lock"),
-				)
-			}
+	// Check to see if this state tool version is different from the lock version.
```
mitchell-as, line 244: This doesn't need to be locked behind a `shouldRunAutoUpdate()`, nor should it, because that function will return false in our integration tests due to `isFreshInstall()` being true.

internal/analytics/constants/constants.go
```diff
@@ -148,39 +148,47 @@ const ActCommandInputError = "command-input-error"
...
+// UpdateLabelFreshInstall is sent if we should not auto-update as we are on a fresh install
 const UpdateLabelFreshInstall = "fresh-install"
 
-// UpdateLabelLocked is the sent if we should not auto-update as the state tool is locked
+// UpdateLabelLocked is sent if we should not auto-update as the state tool is locked
 const UpdateLabelLocked = "locked"
 
+// UpdateLabelSkipChecks is sent if we should not auto-update because the command explicitly skips
+// auto update checks.
+const UpdateLabelSkipChecks = "skip-checks"
```
MDrakos, line 186: Do new labels require us to update the analytics reference? Can you please follow-up with Nathan after merging.

internal/analytics/constants/constants.go
```diff
@@ -148,39 +148,47 @@ const ActCommandInputError = "command-input-error"
 // ActExecutorExit is the event action used for executor exit codes
 const ActExecutorExit = "executor-exit"
 
-// UpdateLabelSuccess is the sent if an auto-update was successful
```
MDrakos, line 151: lol nice fix

MDrakos (approved):
Nice and clean!

internal/analytics/constants/constants.go
mitchell-as: I'm going to flag Nathan for review since this feels like some architecture changes in addition to the label thing.

cmd/state/main.go
```diff
@@ -234,23 +234,22 @@ func run(args []string, isInteractive bool, cfg *config.Instance, out output.Out
...
+	if (childCmd == nil || !childCmd.SkipChecks() || childCmd.Name() != "update") && pj != nil && pj.IsLocked() {
```
Naatan, line 245:
```suggestion
	if (childCmd == nil || (!childCmd.SkipChecks() || childCmd.Name() != "update")) pj != nil && pj.IsLocked() {
```

I think without this `childCmd.Name()` could panic?

Naatan, line 245: Can we add "SkipChecks" to the `update` command?

internal/analytics/constants/constants.go
Naatan: Yes we should. I've updated the doc.

Naatan (changes requested):

cmd/state/main.go
mitchell-as: No it will not, because `childCmd` is `nil` for `state --version`, and it does not panic. I tested this, and so do the integration tests.

mitchell-as: Looks like it is (https://github.com/ActiveState/cli/blob/87205e082c99853af80009361b73dbe1ef4cd351/cmd/state/internal/cmdtree/update.go#L33), so I've removed the extra check.

Naatan: Huh interesting, that kinda makes no sense. I'd feel a lot better if we incorporated my suggestion, or at least understood why this works despite the fact that it shouldn't. The way it's written at the moment I'd expect `childCmd.Name() != "update"` to be evaluated even when the first `childCmd == nil` evaluation returns `false`.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3358#discussion_r1647849913

cmd/state/main.go
mitchell-as: After discussion on slack, I think we can move forward as-is.

Naatan: Resolved on Slack; I was miss-reading this as `childCmd != nil`.

Naatan (approved):

PR #3361, opened 2024-06-20
Show dependency summary on `state commit`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2608" title="DX-2608" target="_blank">StoryDX-2608</a>  `state commit` execution provides list of dependencies as it done in case of `state install`
  </td></table>
  <br />
 

Sample:

```
% ../../cli/build/state commit
█ Commit Changes

Operating on project ActiveState-CLI/small-python, located at /path/to/small-python.

 • Creating commit ✔ Done
   Installing requests@2.28.2 includes 3 direct dependencies.
   ├─ certifi@2021.5.30
   ├─ charset-normalizer@3.0.1
   └─ urllib3@1.26.13

Commit '6b11f918-fa90-43de-9b34-a4da50a9ca9c' for project 'ActiveState-CLI/small-python' 
successfully created.

Run 'state refresh' to update your local runtime.
```

### DIFF
--- a/internal/runners/commit/commit.go
+++ b/internal/runners/commit/commit.go
@@ -11,12 +11,15 @@ import (
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runbits/buildscript"
+	"github.com/ActiveState/cli/internal/runbits/dependencies"
 	"github.com/ActiveState/cli/internal/runbits/rationalize"
+	"github.com/ActiveState/cli/internal/runbits/runtime"
 	"github.com/ActiveState/cli/pkg/localcommit"
 	bpResp "github.com/ActiveState/cli/pkg/platform/api/buildplanner/response"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/model/buildplanner"
+	"github.com/ActiveState/cli/pkg/platform/runtime/target"
 	"github.com/ActiveState/cli/pkg/project"
 )
 
@@ -82,6 +85,8 @@ func (c *Commit) Run() (rerr error) {
 		return rationalize.ErrNoProject
 	}
 
+	c.out.Notice(locale.Tr("operating_message", c.proj.NamespaceString(), c.proj.Dir()))
+
 	// Get buildscript.as representation
 	script, err := buildscript_runbit.ScriptFromProject(c.proj)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


test/integration/commit_int_test.go
```diff
@@ -72,7 +74,7 @@ func (suite *CommitIntegrationTestSuite) TestCommitAtTimeChange() {
 	suite.Require().NoError(err) // verify validity
 
 	// Update top-level at_time variable.
-	dateTime := "2023-03-01T12:34:56.789Z"
+	dateTime := "2023-06-21T12:34:56.789Z"
```
mitchell-as, line 77: I had to use a timestamp for a point where Python 3.10.12 exists on the Platform for solving. It didn't exist on the previous date, which resulted in a solve failure.

internal/runners/commit/commit.go
```diff
@@ -140,9 +145,26 @@ func (c *Commit) Run() (rerr error) {
...
 
+	// Solve runtime
+	_, rtCommit, err := runtime.Solve(c.auth, c.out, c.analytics, c.proj, &stagedCommitID, target.TriggerCommit, c.svcModel, c.cfg, runtime.OptNone)
+	if err != nil {
+		return errs.Wrap(err, "Could not solve runtime")
+	}
+
+	// Get old buildplan.
+	rtTarget := target.NewProjectTarget(c.proj, &stagedCommitID, target.TriggerCommit)
+	commit, err := bp.FetchCommit(localCommitID, rtTarget.Owner(), rtTarget.Name(), nil)
```
Naatan, line 159:
```suggestion
	commit, err := bp.FetchCommit(localCommitID, c.proj.Owner(), c.proj.Name(), nil)
```

internal/runners/commit/commit.go
```diff
@@ -140,9 +145,26 @@ func (c *Commit) Run() (rerr error) {
...
+	if err != nil {
+		return errs.Wrap(err, "Failed to fetch build result")
+	}
+	oldBuildPlan := commit.BuildPlan()
```
Naatan, line 163: We should wrap this in a progress spinner for solving.

Naatan (changes requested):

internal/runners/commit/commit.go
mitchell-as: `runtime.Solve()` has its own `Resolving Dependencies` spinner (as you can see in the output at the top of this PR). Is that sufficient?

Naatan: @mitchell-as it doesn't cover the FetchCommit call afterwards, which also is effectively a solve. I believe you can disable the UI from `runtime.Solve()` with a flag (`runtime.NoUI` or something like that).

This will be especially appropriate for the runtime refactor, which eliminates `runtime.Solve()` in favour of a method that does not produce any UI.

Naatan (approved):

PR #3362, opened 2024-06-20
Implement error interface instead of including error object field in struct.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2911" title="DX-2911" target="_blank">BugDX-2911</a>  INIT: Using wrong language name cause `invalid memory address or nil pointer dereference` error
  </td></table>
  <br />

### DIFF
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -78,8 +78,11 @@ var errNoOwner = errs.New("Could not find organization")
 var errNoLanguage = errs.New("No language specified")
 
 type errUnrecognizedLanguage struct {
-	error
 	Name string
+}
+
+func (e errUnrecognizedLanguage) Error() string {
+	return fmt.Sprintf("unrecognized language: %s", e.Name)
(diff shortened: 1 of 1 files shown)


internal/runners/initialize/init.go
```diff
@@ -78,10 +78,13 @@ var errNoOwner = errs.New("Could not find organization")
 var errNoLanguage = errs.New("No language specified")
 
 type errUnrecognizedLanguage struct {
-	error
 	Name string
 }
 
+func (e errUnrecognizedLanguage) Error() string {
+	return fmt.Sprintf("unrecognized language: %s", e.Name)
```
MDrakos, line 85: Does this error message get printed or does the user-facing error handle this?

mitchell-as: The user-facing error handles this. This message only shows up in the debug logs.

MDrakos (approved):

PR #3367, opened 2024-06-24
Refactor integration tests to use empty project by default.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2828" title="DX-2828" target="_blank">TaskDX-2828</a>  Update our integration tests to use a project with a minimal runtime
  </td></table>
  <br />
 

Also, do not disable runtimes by default in integration tests. Tests that want to disable runtimes should opt-in.

We are switching to empty projects for most integration tests, so disabling runtimes by default will no longer be necessary for the sake of speed.

Some things to note:

- We prefer `ts.PrepareProject()` instead of `state checkout`, as checkouts source the runtime (even if `async_runtime` is enabled).
- The following commands do not support `async_runtime`, and manually disable runtimes in the tests:
  - `state init`
  - `state import`
  - `state revert`
  - `state switch`
- There is no noticeable performance improvement or detriment with these changes. The lone data point when running "Test: all" is:
  - Linux: before: 43 mins; after: 44 mins
  - macOS: before: 47 mins; after: 49 mins
  - Windows: before: 1 hour 3 mins; after: 1 hour 4 mins
  - Note: the "after" times were done during the day, as opposed to the "before" times being done at night. CI might be a bit busier during the day, and quieter at night.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,6 @@ jobs:
     runs-on: ${{ matrix.sys.os }}
     env:
       ACTIVESTATE_CI: true
-      ACTIVESTATE_CLI_DISABLE_RUNTIME: true
       SHELL: bash
       GITHUB_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GITHUB_SHA_OVERRIDE: ${{ github.event.pull_request.head.sha || github.sha }}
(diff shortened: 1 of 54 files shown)


mitchell-as:
Test failures are known failures with bugs filed. This PR introduces no new failures.

Naatan (approved):
Awesome work! Shame we won't see a performance improvement from this, but that was never the goal. If anything fact that performance is the same suggests our runtime sourcing is performant as the only significant overhead is the API.

mitchell-as:
@Naatan please re-approve. I had to address a merge conflict that just changed commit_int_test.go, seen here: https://github.com/ActiveState/cli/commit/d044a5665f86204a54008f101db509baa4c570ac#diff-6053a79d5a732fb36672ca1d4e64385149ecfd4086206a8691d8cb1eb419648f

mitchell-as:
Never mind, it looks like I can put this one in myself since it was already approved once.

PR #3373, opened 2024-06-25
Fixed crash running `state export runtime` with async runtimes enabled.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2916" title="DX-2916" target="_blank">BugDX-2916</a>  Crash with async runtime and `state export runtime`.
  </td></table>
  <br />


Naatan (approved):
Given the runtime code is about to be replaced this is probably fine for now, but this should ideally actually fix the nil assertion rather than implicitly satisfy it.

PR #3380, opened 2024-06-28
Update termtest to fix process exit race condition

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2901" title="DX-2901" target="_blank">BugDX-2901</a>  Termtest: Not all output received when process exits
  </td></table>
  <br />


mitchell-as (approved):
I trust this is a direct import of what I already reviewed last week, and that no further changes were made.

PR #3382, opened 2024-06-28
State export log ignores some log files

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2892" title="DX-2892" target="_blank">StoryDX-2892</a>  `state export log -i` should not consider logs for `state export log`
  </td></table>
  <br />
 

One thing to note about this PR is that moving the logging of arguments to the `captain` package means that they won't be at the top of the log file anymore. I still think it makes sense to log the arguments in that package. The added requirement of only logging known commands and flags means that we have to log after command registration so that we can inspect the arguments and omit things we don't recognize.


internal/captain/command.go
```diff
@@ -880,3 +918,30 @@ func childCommands(cmd *Command) string {
 
 	return fmt.Sprintf("\n\nAvailable Commands:\n%s", table.Render())
 }
+
+func (c *Command) LogArgs() {
+	children, err := c.FindChildren(os.Args[1:])
```
Naatan, line 923: We should use `c.Args`. I think for those it already dropped the first entry of `os.Args`, but worth verifying.

Naatan, line 923: I'm pretty sure `FindChild()` already achieves the desired effect you're looking for. And it passes through cobra, guaranteeing that we get the right command. From there I think you could use `command.cobra.CommandPath()`, which should give you the full path.

internal/runners/export/log.go
```diff
@@ -57,12 +60,24 @@ func (l *Log) Run(params *LogParams) (rerr error) {
...
 		if regex.MatchString(file) {
 			timestamp, err := strconv.Atoi(regex.FindStringSubmatch(file)[1])
 			if err != nil {
 				continue
 			}
+
+			ignore, err := ignoreLogFile(file)
+			if err != nil {
+				logging.Error("failed to validate log file: %s", err.Error())
+				continue
```
Naatan, line 74: Bubble up errors please. This shouldn't ever cause an error, so if it does something is broken and we need to fix it. Logging just obfuscates the breakage.

internal/runners/export/log.go
```diff
@@ -85,3 +100,31 @@ func (l *Log) Run(params *LogParams) (rerr error) {
...
 }
+
+func ignoreLogFile(logFile string) (bool, error) {
+	file, err := os.Open(logFile)
+	if err != nil {
+		return false, errs.Wrap(err, "failed to open log file")
+	}
+	defer file.Close()
+
+	regex := regexp.MustCompile(`.*\] Args: \[(.*?)\], Flags: \[.*?\]`)
```
Naatan, line 111:
```suggestion
	regex := regexp.MustCompile(`Args: \[(.*?)\], Flags: \[.*?\]`)
```

Doesn't seem necessary for this function to have an opinion on what the prefix looks like.

internal/runners/export/log.go
```diff
@@ -85,3 +100,31 @@ func (l *Log) Run(params *LogParams) (rerr error) {
...
+		if regex.MatchString(logLine) {
+			match := regex.FindStringSubmatch(logLine)
+			if len(match) > 1 {
+				args = match[1]
+			}
+			break
+		}
+	}
+
+	if strings.Contains(args, "export") && strings.Contains(args, "log") {
```
Naatan, line 125: I'd prefer we be a but more explicit with the matches. However unlikely, it's easy enough to guard against something like `state exec -- myapp -i file.log --export` or something along those lines.

Naatan (changes requested):

internal/captain/command.go
```diff
@@ -920,18 +885,15 @@ func childCommands(cmd *Command) string {
 }
 
 func (c *Command) LogArgs() {
-	children, err := c.FindChildren(os.Args[1:])
+	child, err := c.FindChild(os.Args[1:])
```
Naatan, line 888: https://github.com/ActiveState/cli/pull/3382#discussion_r1673003268

internal/runners/export/log.go
```diff
@@ -85,3 +99,42 @@ func (l *Log) Run(params *LogParams) (rerr error) {
...
+		}
+
+		if i+1 >= len(splitArgs) {
+			break
+		}
+
+		if splitArgs[i+1] == "log" {
+			return true, nil
+		}
+	}
```
Naatan, line 137:
```suggestion
	if strings.HasPrefix(args, "export log") {
		return true, nil
	}
```

Naatan (changes requested):

internal/captain/command.go
MDrakos: For some reason all of my comments were `Pending` 🤦

Naatan:
@MDrakos can't respond to actual comment cause it's "old" and github UX is awful.

Suggest instead passing `os.Args` to the `LogArgs` function, that way captain isn't violating its own responsibilities by reaching out to `os.Args`.

That said, perhaps it would be better to just log args internally when the command is executed and the args are passed, ie. here:

https://github.com/ActiveState/cli/blob/26129e48e102938e68f2fcb10956e45488bc5d82/internal/captain/command.go#L216-L216

Naatan (changes requested):
Looks good! I think we just need a simple integration test covering this mechanic though, come to think of it. As this feels a little fragile considering how the logic is somewhat implicitly shared between different packages.

Naatan (approved):
Nice work!

PR #3383, opened 2024-07-01
Initial ability to `state install` anything.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2805" title="DX-2805" target="_blank">StoryDX-2805</a>  As a user I can use `state install` to install anything
  </td></table>
  <br />


internal/runbits/requirements/install.go
```diff
@@ -0,0 +1,366 @@
+package requirements
```
mitchell-as, line 1: I am starting the refactor of the monolithic `ExecuteRequirementsOperation()` here. The responsibility of this file is simply to handle installing requirements. I envision another file, _uninstall.go_, having the responsibility of uninstalling requirements.

test/integration/package_int_test.go
```diff
@@ -278,9 +278,9 @@ func (suite *PackageIntegrationTestSuite) TestPackage_detached_operation() {
 	cp.ExpectExitCode(0)
 
 	suite.Run("install non-existing", func() {
-		cp := ts.Spawn("install", "json")
+		cp := ts.Spawn("install", "argpars")
```
mitchell-as, line 281: See https://activestatef.atlassian.net/browse/DX-2805?focusedCommentId=53790 for an explanation of why this changed.

test/integration/package_int_test.go
```diff
@@ -512,14 +512,14 @@ func (suite *PackageIntegrationTestSuite) TestNormalize() {
 	cp.ExpectExitCode(0)
 
 	cp = ts.SpawnWithOpts(
-		e2e.OptArgs("install", "Charset_normalizer"),
+		e2e.OptArgs("install", "Charset-normalizer"),
```
mitchell-as, line 515: See https://activestatef.atlassian.net/browse/DX-2805?focusedCommentId=53790 for an explanation of why this changed.

internal/runbits/requirements/rationalize.go
```diff
@@ -71,6 +77,32 @@ func (r *RequirementOperation) rationalizeError(err *error) {
...
+				typeString = fmt.Sprintf("%s %s", cases.Title(language.English).String(lang), typeString)
+			}
+			messages = append(messages,
+				locale.Tl("err_package_multiple_matches_match",
+					"  {{.V0}}\t\t\t{{.V1}}\t\t[ACTIONABLE]{{.V2}}[/RESET]",
+					*match.Ingredient.Name,
+					typeString,
+					fmt.Sprintf("%s/%s", *match.Ingredient.PrimaryNamespace, *match.Ingredient.Name)))
+		}
+		*err = errs.WrapUserFacing(*err, strings.Join(messages, "\n"), errs.SetInput())
```
mitchell-as, line 104: Yes, the use of tab characters for formatting is awful, but our errors and user-facing errors mechanism expects strings, not objects, so I cannot pass the usual struct with tags to print nicely as a table. I'm open to suggestions or a redo of the mockup.

test/integration/install_int_test.go
```diff
@@ -64,7 +64,7 @@ func (suite *InstallIntegrationTestSuite) TestInstall_NoMatches_Alternatives() {
 	defer ts.Close()
 
 	ts.PrepareProject("ActiveState-CLI/small-python", "5a1e49e5-8ceb-4a09-b605-ed334474855b")
-	cp := ts.Spawn("install", "database")
+	cp := ts.Spawn("install", "dataa")
```
mitchell-as, line 67: See https://activestatef.atlassian.net/browse/DX-2805?focusedCommentId=53790 for a description of why this changed.

mitchell-as:
I'm putting this up for early review due to my comment here: https://activestatef.atlassian.net/browse/DX-2805?focusedCommentId=53790. I don't know if it constitutes a blocker or not. Please read through my annotations in this PR too, as they provide context. Also, I'm unsure how to emit a structured output error object.

Naatan (changes requested):
@mitchell-as let's move the story back to triage so we can discuss how we want to handle it.

There's a couple of implications here both for product and the user experience, as well as for the technical planning / maintainability of the feature. I think it's likely we'll want to pivot a little and probably split this off into multiple stories.

I like where you're headed with the refactor, I imagine we can still iterate on what you started here!

mitchell-as:
@Naatan since brought the ticket back into the sprint, please re-review with the discussion in mind. I'm unsure how it affects your comment.

mitchell-as:
Actually, never mind. I have to stop making suggestions first.

mitchell-as:
Test failures are known and not caused by this PR.

mitchell-as:
@Naatan since we brought the ticket back into the sprint, please re-review with the discussion in mind. I'm unsure how it affects your comment.

The only change since your last review is that we're only fetching suggestions if the project has one language (https://github.com/ActiveState/cli/pull/3383/commits/1870c3f6c655b3ef35905a8e007b81361b810d78).

Naatan (changes requested):
I think we should address a refactor of the requirement runbit first. You've started it here but I don't think the stage you're at is one I'm comfortable committing, as it essentially leaves us maintaining two very similar implementations; the new `Install()` and the old `ExecuteRequirementOperation()`. I'd rather we push the refactor through far enough that we don't have two diverging implementations. I'm not saying they should all use the same operation, but under the hood the logic should be shared and not duplicated.

I feel like this is too much scope creep for your story, as refactoring the requirements package ought to involve a bit more strategy and planning. I'd suggest blocking this story on the refactor unless you see another path.

mitchell-as:
Okay, I've blocked the ticket on the ExecuteRequirementOperation refactor. Once that's done, I'll open a new PR.

PR #3385, opened 2024-07-03
 Added change summary and CVE report to `state import`

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2550" title="DX-2550" target="_blank">StoryDX-2550</a>  Change and CVE information for `state import`
  </td></table>
  <br />
 

Also:
- Refactored to use a primer for passing to cve runbit.
- Enabled async runtime
- Fixed and re-enabled import integration tests that were disabled in the early buildplanner days

### DIFF
--- a/internal/runners/packages/import.go
+++ b/internal/runners/packages/import.go
@@ -3,14 +3,15 @@ package packages
 import (
 	"os"
 
-	"github.com/ActiveState/cli/internal/analytics"
+	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
-	"github.com/ActiveState/cli/internal/keypairs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
-	"github.com/ActiveState/cli/internal/prompt"
+	"github.com/ActiveState/cli/internal/runbits"
+	"github.com/ActiveState/cli/internal/runbits/cves"
+	"github.com/ActiveState/cli/internal/runbits/dependencies"
 	"github.com/ActiveState/cli/internal/runbits/rationalize"
 	"github.com/ActiveState/cli/internal/runbits/runtime"
 	"github.com/ActiveState/cli/pkg/buildscript"
@@ -18,20 +19,14 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/api"
 	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/types"
 	"github.com/ActiveState/cli/pkg/platform/api/reqsimport"
-	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/platform/model/buildplanner"
 	"github.com/ActiveState/cli/pkg/platform/runtime/target"
-	"github.com/ActiveState/cli/pkg/project"
 )
 
 const (
 	defaultImportFile = "requirements.txt"
 )
-
-type configurable interface {
-	keypairs.Configurable
-}
 
 // Confirmer describes the behavior required to prompt a user for confirmation.
 type Confirmer interface {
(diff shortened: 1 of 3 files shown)


mitchell-as:
Test failures are known and not due to this PR.

Naatan (changes requested):
I'm sorry this is doing a lot for what the story is asking for. I'm ok with a bit of scope creep to avoid or address technical debt, but this is starting to refactor cve reporting. I'm not against refactoring, but we need to be more intentional with those kinds of jobs. I imagine a lot of churn going into this PR as is if we are to include refactoring, and we'd be going through that churn without having done any requirement setting for the refactor in the first place.

I see two options:

1. Bring issue back to triage; we file a new refactor story as a blocker and we repurpose the code changes in this PR for the refactor.
2. Undo the refactoring changes and just focus on the requirements for the story, then keep these changes for a follow refactor story.

My preference would be number 2, but I'll defer to you as there's obviously a reason you resorted to the refactor, and so this may be a hard blocker for addressing the story.

mitchell-as:
Okay, this PR now utilizes the new cves runbit from #3386. From the PR description, this PR

- Adds change summary and CVE report to `state import`
- Refactors `state import` to use a primer for passing to cve runbit.
- Enabled async runtime for `state import`
- Fixed and re-enabled `state import` integration tests that were disabled in the early buildplanner days

internal/runners/packages/import.go
```diff
@@ -144,12 +135,51 @@ func (i *Import) Run(params *ImportRunParams) error {
...
 
-	if err := localcommit.Set(i.proj.Dir(), commitID.String()); err != nil {
+	pg.Stop(locale.T("progress_success"))
+	pg = nil
+
+	// Solve the runtime.
+	rt, rtCommit, err := runtime.Solve(auth, out, i.prime.Analytics(), proj, &commitID, target.TriggerImport, i.prime.SvcModel(), i.prime.Config(), runtime.OptNone)
+	if err != nil {
+		return errs.Wrap(err, "Could not solve runtime")
+	}
```
Naatan, line 145: This made me realize we're redundantly solving. Not asking for this to be addressed here, but for future context: https://activestatef.atlassian.net/browse/DX-2934

Filed it separately as this could easily get more complicated than just updating the model code and using the result from stageCommit directly (ie. UI implications).

internal/runners/packages/import.go
```diff
@@ -144,12 +135,51 @@ func (i *Import) Run(params *ImportRunParams) error {
...
+				// If the error is not a build error we want to retain the changes
+				if err2 := localcommit.Set(proj.Dir(), commitID.String()); err2 != nil {
+					return errs.Pack(err, locale.WrapError(err2, "err_package_update_commit_id"))
+				}
+			}
+			return errs.Wrap(err, "Failed to refresh runtime")
+		}
+	}
+
+	if err := localcommit.Set(proj.Dir(), commitID.String()); err != nil {
```
Naatan, line 178: This should happen before the runtime update, otherwise runtime update failing will cause the import to have no effect. This makes sense for `state install` because you're only touching one thing, but for import I think you'd want to retain the broken state so you can fix it.

Naatan (changes requested):

internal/runners/packages/import.go
mitchell-as: Huh, that's the opposite of what I'd expect. If I do `state import` and my requirements.txt is causing a build failure, I'd want to edit my requirements.txt and try again, not do a bunch of `state uninstall`, `state install` operations to try and fix things.

That said, I'll defer to you, but I wanted to double-check this is what you want.

mitchell-as: Going a step further, it's like `apt install <long list of deps>` and committing even if there's a dependency failure, and then expecting you to `apt uninstall` and `apt install` your way back to a working system. Now I understand we have `state reset` which may sidestep this issue entirely, but it's still an extra step. With a big red 'x' saying my runtime failed to build, I think it's reasonable as a user to assume nothing was committed.

Naatan: @mitchell-as that's a good point, I'm honestly not sure. I could see an argument for both sides, because maybe the nature of the issue is not one that can be addressed by editing the requirements.txt, and further tweaking is required. In this case you'd need the interim state.

That said, the story didn't ask for a change to this mechanic, so I don't think we should be introducing it here. Especially when the value prop isn't black and white.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3385#discussion_r1665887148

Naatan (approved):

PR #3386, opened 2024-07-03
Moved CVE reporting into its own runbit.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2932" title="DX-2932" target="_blank">TaskDX-2932</a>  Move CVE reporting out of the requirements runbit into its own runbit
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/requirements/requirements.go
+++ b/internal/runbits/requirements/requirements.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/logging"
-	configMediator "github.com/ActiveState/cli/internal/mediators/config"
 	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/output"
 	"github.com/ActiveState/cli/internal/primer"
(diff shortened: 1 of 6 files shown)


mitchell-as:
Failures are known and not introduced by this PR.

internal/runbits/requirements/requirements.go
```diff
@@ -62,6 +55,10 @@ func (pv *PackageVersion) Set(arg string) error {
 }
 
 type RequirementOperation struct {
+	prime primeable
+	// The remainder is redundant with the above. Refactoring this will follow in a later story so as not to blow
+	// up the one that necessitates adding the primer at this level.
+	// https://activestatef.atlassian.net/browse/DX-2869
```
mitchell-as, line 61: Copied from your PR to hopefully avoid a merge conflict.

internal/runbits/requirements/requirements.go
```diff
@@ -558,99 +555,6 @@ func (r *RequirementOperation) resolveRequirement(requirement *Requirement) erro
...
-	}
-
-	names := requirementNames(requirements...)
-	pg := output.StartSpinner(r.Output, locale.Tr("progress_cve_search", strings.Join(names, ", ")), constants.TerminalAnimationInterval)
-
-	var ingredients []*request.Ingredient
-	for _, requirement := range requirements {
-		if requirement.Operation == types.OperationRemoved {
-			continue
-		}
```
mitchell-as, line 574: This is the only functional change in the PR. I removed the `requirements` argument and construct the list of names from the `artifactChangeset`. I also removed the top-level iteration over `requirements`, which is superfluous.

Naatan (changes requested):
Looks good! But we're still passing prime members around as function parameters. I think it would be cleaner if these all shared a struct that has the prime recorded on it, just like our runners normally behave. 

A runbit really ought to just be a runner, but one that unlike regular runners can be imported by other runners. Point being the code itself should mostly follow the same pattern.

mitchell-as:
Is this kind of what you were looking for?

Naatan:
> Is this kind of what you were looking for? [36be5f3](https://github.com/ActiveState/cli/commit/36be5f3a038cb05f8a64963609e557826518a583)

Yep that's it. To be honest I'm not totally sold on the prime pattern, but as of right now this feels consistent with how we use it elsewhere.

Naatan (approved):

PR #3388, opened 2024-07-03
Fix dependency calculation

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2873" title="DX-2873" target="_blank">BugDX-2873</a>  `state install` reveal many inaccuracies in number of sub-dependencies, progress and summary
  </td></table>
  <br />
 

Eliminate common dependencies (eg. Python)
Fix direct dependencies inheriting non-direct dependencies

### DIFF
--- a/internal/runbits/dependencies/changesummary.go
+++ b/internal/runbits/dependencies/changesummary.go
@@ -33,16 +33,18 @@ func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan,
 			v := fmt.Sprintf("%s@%s", a.Name(), a.Version())
 			addedString = append(addedLocale, v)
 			addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
-		}
-		for _, i := range a.Ingredients {
-			dependencies = append(dependencies, i.RuntimeDependencies(true)...)
-			directDependencies = append(dependencies, i.RuntimeDependencies(false)...)
+
+			for _, i := range a.Ingredients {
+				dependencies = append(dependencies, i.RuntimeDependencies(true)...)
+				directDependencies = append(directDependencies, i.RuntimeDependencies(false)...)
+			}
 		}
 	}
 
 	dependencies = sliceutils.UniqueByProperty(dependencies, func(i *buildplan.Ingredient) any { return i.IngredientID })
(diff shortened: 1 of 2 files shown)


internal/runbits/dependencies/changesummary.go
```diff
@@ -33,16 +33,18 @@ func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan,
...
+				dependencies = append(dependencies, i.RuntimeDependencies(true)...)
+				directDependencies = append(directDependencies, i.RuntimeDependencies(false)...)
+			}
 		}
 	}
 
 	dependencies = sliceutils.UniqueByProperty(dependencies, func(i *buildplan.Ingredient) any { return i.IngredientID })
 	directDependencies = sliceutils.UniqueByProperty(directDependencies, func(i *buildplan.Ingredient) any { return i.IngredientID })
-	numIndirect := len(dependencies) - len(directDependencies)
+	commonDependencies := directDependencies.CommonRuntimeDependencies().ToIDMap()
```
Naatan, line 46: This identifies any dependencies that are used between all ingredients. Effectively this should give us Python, without us needing to know anything about languages.

This'll likely not work very well if someone does an install of multiple requirements across different namespaces, but that's incredibly niche/unlikely.

internal/runbits/dependencies/changesummary.go
```diff
@@ -33,16 +33,18 @@ func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan,
...
 			addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
-		}
-		for _, i := range a.Ingredients {
-			dependencies = append(dependencies, i.RuntimeDependencies(true)...)
-			directDependencies = append(dependencies, i.RuntimeDependencies(false)...)
+
+			for _, i := range a.Ingredients {
+				dependencies = append(dependencies, i.RuntimeDependencies(true)...)
+				directDependencies = append(directDependencies, i.RuntimeDependencies(false)...)
+			}
```
Naatan, line 40: Note 2 changes here:

1. We no longer look at dependencies if we're iterating over an addition that isn't what the user requested.
2. Fixed `directDependencies` appending to the `dependencies` slice.

internal/runbits/dependencies/changesummary.go
mitchell-as: That's a good solution I think.

mitchell-as (approved):

PR #3389, opened 2024-07-04
Work around Windows bug where arguments are not properly escaped

Naatan:


cmd/state-exec/cmd_windows.go
```diff
@@ -0,0 +1,90 @@
...
+func Command(name string, arg ...string) *exec.Cmd {
+	cmd := exec.Command(name, arg...)
+
+	// Go currently does not escape arguments properly on Windows, it account for spaces and tab characters, but not
+	// other characters that need escaping such as `<` and `>`.
+	// This can be dropped once we update to a Go version that fixes this bug: https://github.com/golang/go/issues/68313
+	cmd.SysProcAttr = &syscall.SysProcAttr{CmdLine: makeCmdLine(cmd.Args)}
+
+	return cmd
+}
```
Naatan, line 17: Note this is purposely duplicating logic in osutils, because a design principle behind executors is that we avoid reusing internal libraries so as to keep the executable as lightweight as possible.

test/integration/executor_int_test.go
```diff
@@ -90,3 +99,51 @@ func (suite *ExecutorIntegrationTestSuite) TestExecutorSizeOnDisk() {
...
+	)
+	suite.Require().NoError(err)
+	suite.Require().FileExists(reportExe)
+
+	// Force override ACTIVESTATE_CI to false, because communicating with the svc will fail, and if this is true
+	// the executor will interrupt.
+	// For this test we don't care about the svc communication.
+	env := e2e.OptAppendEnv("ACTIVESTATE_CI=false")
+
+	inputs := []string{"a<b", "b>a", "hello world", "&whoami", "imnot|apipe", "%NotAppData%", "^NotEscaped", "(NotAGroup)"}
```
mitchell-as, line 135: That's a nice set of test inputs!

cmd/state-exec/cmd_windows.go
mitchell-as: Yeah, I've accidentally included things that blew up the exe size. Fortunately we have a test to catch it 😅

PR #3390, opened 2024-07-05
Implement powershell support

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2852" title="DX-2852" target="_blank">StoryDX-2852</a>  As a user I can expect state tool to stay in powershell when I run `state shell`
  </td></table>
  <br />


internal/subshell/pwsh/pwsh.go
```diff
@@ -0,0 +1,169 @@
...
+func (v *SubShell) SetBinary(binary string) {
+	v.binary = binary
+}
+
+// WriteUserEnv - see subshell.SubShell
+func (v *SubShell) WriteUserEnv(cfg sscommon.Configurable, env map[string]string, envType sscommon.RcIdentification, userScope bool) error {
+	cmdShell := &cmd.SubShell{}
+	if err := cmdShell.WriteUserEnv(cfg, env, envType, userScope); err != nil {
+		return errs.Wrap(err, "Forwarded WriteUserEnv call failed")
+	}
```
Naatan, line 58: Super dirty, but all of this subshell stuff desperately needs a refactor, I didn't see much value in trying to get something satisfying to fit into something .. not (impossible).

internal/subshell/sscommon/rcfile.go
```diff
@@ -252,6 +253,8 @@ func SetupProjectRcFile(prj *project.Project, templateName, ext string, env map[
 		return nil, errs.Wrap(err, "Failed to read asset")
 	}
 
+	logging.Debug("Env: %v", env)
+
```
mitchell-as, line 257: Did you mean to leave this?

internal/subshell/pwsh/pwsh.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/ActiveState/cli/internal/output"
+	"github.com/ActiveState/cli/internal/subshell/cmd"
+	"github.com/ActiveState/cli/internal/subshell/sscommon"
+	"github.com/ActiveState/cli/pkg/project"
+)
+
+var escaper *osutils.ShellEscape
+
+func init() {
+	escaper = osutils.NewBatchEscaper()
```
mitchell-as, line 24: Are we sure this escaper works for powershell?

internal/subshell/pwsh/pwsh.go
mitchell-as: Do we have a ticket for refactoring all of this? If not, we may want to file one.

internal/subshell/pwsh/pwsh.go
```diff
@@ -0,0 +1,169 @@
...
+	return locale.NewError("err_writecompletions_notsupported", "{{.V0}} does not support completions.", v.Shell())
+}
+
+func (v *SubShell) RcFile() (string, error) {
+	home, err := user.HomeDir()
+	if err != nil {
+		return "", errs.Wrap(err, "Could not get home dir")
+	}
+
+	return filepath.Join(home, "Documents", "WindowsPowerShell", "Microsoft.PowerShell_profile.ps1"), nil
```
mitchell-as, line 86: Eww, littering the user's documents directory. I'm guessing PowerShell writes to this directory if needed? (It doesn't exist for me by default.) And are we sure this is the exact path? (e.g. no space between "Windows" and "PowerShell"?)

internal/subshell/sscommon/rcfile.go
```diff
@@ -392,8 +412,6 @@ func SetupProjectRcFile(prj *project.Project, templateName, ext string, env map[
 		return nil, errs.Wrap(err, "Failed to write to output buffer.")
 	}
 
-	logging.Debug("Using project RC: (%s) %s", tmpFile.Name(), o.String())
-
```
mitchell-as, line 396: Did you mean to leave this deleted?

internal/subshell/subshell.go
```diff
@@ -204,7 +211,7 @@ func DetectShell(cfg sscommon.Configurable) (string, string) {
 	}
 
 	isKnownShell := false
-	for _, ssName := range []string{bash.Name, cmd.Name, fish.Name, tcsh.Name, zsh.Name} {
+	for _, ssName := range []string{bash.Name, cmd.Name, fish.Name, tcsh.Name, zsh.Name, pwsh.Name} {
```
mitchell-as, line 214: You can remove the check for powershell on line 223 now since we officially recognize it.

tmp/test.go
```diff
@@ -0,0 +1 @@
+package tmp
```
mitchell-as, line 1: Is this intentional? Or was it accidentally committed?

mitchell-as (changes requested):
This largely looks okay (assuming the integration tests pass), but there are a few minor things to clarify.

internal/subshell/pwsh/pwsh.go
Naatan: No, but apparently we no longer invoke the logic that uses this. I've dropped it and added a comment.

internal/subshell/pwsh/pwsh.go
Naatan: This is powershell's equivalent of a bashrc file. 

https://activestatef.atlassian.net/browse/DX-2944

Powershell reads this file if it exists. It doesn't write to it afaik.

internal/subshell/sscommon/rcfile.go
Naatan: Yes, this shouldn't have been committed way back when, I'm trying to over time reduce the log file down to the helpful essentials.

internal/subshell/subshell.go
Naatan: Thanks for catching!

tmp/test.go
Naatan: Nope, should not have committed this. Thanks for catching.

internal/subshell/subshell.go
```diff
@@ -213,9 +220,6 @@ func DetectShell(cfg sscommon.Configurable) (string, string) {
 
 	if !isKnownShell {
 		logging.Debug("Unsupported shell: %s, defaulting to OS default.", name)
-		if !strings.EqualFold(name, "powershell") && name != "sh" {
-			rollbar.Error("Unsupported shell: %s", name) // we just want to know what this person is using
-		}
```
mitchell-as, line 218: We want to leave that second part of the conditional. It should be:

```
if name != "sh" {
  rollbar.Error("Unsupported shell: %s", name) // we just want to know what this person is using
}
```

mitchell-as (changes requested):

mitchell-as (approved):

PR #3396, opened 2024-07-09
Fix `state exec` unable to run complex arguments

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2946" title="DX-2946" target="_blank">BugDX-2946</a>  `state exec` mangels arguments on windows
  </td></table>
  <br />

### DIFF
--- a/internal/runners/exec/exec.go
+++ b/internal/runners/exec/exec.go
@@ -188,8 +188,8 @@ func (s *Exec) Run(params *Params, args ...string) (rerr error) {
 	lang := language.Bash
 	scriptArgs := fmt.Sprintf(`%q "$@"`, exeTarget)
 	if strings.Contains(s.subshell.Binary(), "cmd") {
-		lang = language.PowerShell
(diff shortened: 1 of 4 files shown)


internal/runners/exec/exec.go
```diff
@@ -188,8 +188,8 @@ func (s *Exec) Run(params *Params, args ...string) (rerr error) {
 	lang := language.Bash
 	scriptArgs := fmt.Sprintf(`%q "$@"`, exeTarget)
 	if strings.Contains(s.subshell.Binary(), "cmd") {
-		lang = language.PowerShell
-		scriptArgs = fmt.Sprintf("& %q @args\nexit $LASTEXITCODE", exeTarget)
```
Naatan, line 192: I reverted this because it's no longer needed, and for whatever reason this combined with the new solution I pushed caused a regression. I suspect because we were chaining cmd > ps1 > bat, whereas now it's just cmd > bat > bat.

mitchell-as (approved):

PR #3402, opened 2024-07-17
Refactor ExecuteRequirementOperation into separate install and uninstall methods.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1897" title="DX-1897" target="_blank">TaskDX-1897</a>  Refactor ExecuteRequirementOperation
  </td></table>
  <br />
 

This PR splits up the monolitic ExecuteRequirementsOperation() function into separate Install and Uninstall functions. It also has a separate InstallPlatform function because platforms are not ingredients.


mitchell-as:
Test failures are known and not introduced by this PR.

mitchell-as:
@Naatan while this PR is ready for review, I recognize we haven't defined the ACs for the refactor ticket. We should do that sooner rather than later since any work we do to `runbits/runtime/requirements.go` will need to be ported to this refactor PR. I've done this twice: the runtime refactor work you did, and then Mike's CVE fix.

I basically just extracted out the `state install` anything work I did from my initial refactor and cleaned things up a bit. Existing functionality remains unchanged.

internal/runners/languages/languages.go
```diff
@@ -16,6 +17,16 @@ import (
 	"github.com/ActiveState/cli/pkg/project"
 )
 
+type primeable interface {
```
mitchell-as, line 20: This was moved from the removed languages/install.go.

cmd/state/internal/cmdtree/bundles.go
```diff
@@ -45,7 +45,7 @@ func newBundlesCommand(prime *primer.Values) *captain.Command {
 func newBundleInstallCommand(prime *primer.Values) *captain.Command {
 	runner := packages.NewInstall(prime)
 
-	params := packages.InstallRunParams{}
+	params := &packages.InstallRunParams{NamespaceType: &model.NamespaceBundle}
```
mitchell-as, line 48: Instead of passing namespace type to the runner as an argument, pass it in the params. We'll do this for languages and platforms too.

internal/runners/packages/install.go
```diff
@@ -6,44 +6,82 @@ import (
...
+	Revision      captain.IntValue
+	NamespaceType *model.NamespaceType
 }
 
 // Install manages the installing execution context.
 type Install struct {
 	prime primeable
 }
 
+type errNamespaceMismatch struct {
```
mitchell-as, line 27: See https://github.com/ActiveState/cli/pull/3402/files#diff-9660d67939c62dc12e53e50721067074f16c958fd6f5f554b86d1b8de83be503R80 for why this exists.

internal/runbits/runtime/requirements/install.go
```diff
@@ -0,0 +1,448 @@
...
+			multilog.Error("Failed to normalize '%s': %v", req.Name, err)
+		}
+		ns = namespace.String()
+		//caseSensitive = true // ideally case-sensitive would be true (PB-4371)
+	}
+
+	return model.SearchIngredientsStrict(ns, name, caseSensitive, false, nil, r.Auth)
+}
+
+func (r *RequirementOperation) InstallPlatform(requirements []*Requirement) (rerr error) {
```
mitchell-as, line 320: For readability this needs to be a separate function because platforms are not requirements, so there are completely different code paths.

internal/runbits/runtime/requirements/uninstall.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/ActiveState/cli/internal/runbits/runtime"
+	"github.com/ActiveState/cli/internal/runbits/runtime/trigger"
+	"github.com/ActiveState/cli/pkg/buildscript"
+	"github.com/ActiveState/cli/pkg/localcommit"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/types"
+	"github.com/ActiveState/cli/pkg/platform/model"
+	bpModel "github.com/ActiveState/cli/pkg/platform/model/buildplanner"
+)
+
+func (r *RequirementOperation) Uninstall(requirements []*Requirement) (rerr error) {
```
mitchell-as, line 25: Like with installing platforms, this functionality should be in a separate area for readability and maintainability. Uninstalling requires more subtle code path differences from installing.

However, we can group requirements and platforms into the same function because there's only one code branch that's easy to follow.

mitchell-as:
I can hand this off to you @Naatan in its current state if you like. I accomplished what I really wanted, which was to split up the monolithic function into separate, more maintainable pieces. It works as expected and passes existing tests.

Naatan:
As discussed on Slack; I'm going for a larger refactor than this, so we'll close this PR and I'll open a new one.

PR #3405, opened 2024-07-18
`state import` uses a private namespace if added packages do not have one.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2941" title="DX-2941" target="_blank">StoryDX-2941</a>  Users can import and create projects from CycloneDX SBOMs using `state import`
  </td></table>
  <br />
 

This can happen with language-agnostic formats like CycloneDX and SPDX.

### DIFF
--- a/internal/runbits/org/org.go
+++ b/internal/runbits/org/org.go
@@ -1,0 +1,63 @@
+package org
+
+import (
+	"strings"
+
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/pkg/platform/authentication"
+	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/project"
+)
+
+var ErrNoOwner = errs.New("Could not find organization")
+
+type configurer interface {
+	GetString(string) string
+}
+
+// Get returns the name of an organization/owner in order of precedence:
+// - Returns the normalized form of the given org name.
+// - Returns the name of the most recently used org.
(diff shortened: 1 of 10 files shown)


internal/runbits/org/org.go
```diff
@@ -0,0 +1,63 @@
+package org
```
mitchell-as, line 1: This comes from the init runner, which selects a default org if one wasn't specified. This code is not new.

mitchell-as:
Test failures are known issues and not due to this PR.

internal/runbits/org/org.go
```diff
@@ -0,0 +1,63 @@
...
+	if desiredOrg != "" {
+		// Match the case of the organization.
+		// Otherwise the incorrect case will be written to the project file.
+		for _, org := range orgs {
+			if strings.EqualFold(org.URLname, desiredOrg) {
+				return org.URLname, nil
+			}
+		}
+		// Return desiredOrg for error reporting
+		return desiredOrg, ErrNoOwner
```
MDrakos, line 39: I believe that in go if an error is returned the other return value is not to be trusted. I think a better approach here may be to use an error type and have the desired org be a member of it.

MDrakos (changes requested):

internal/runbits/org/org.go
mitchell-as: Technically, fixing this is out of scope for the ticket (this is not new code), but I've gone ahead and done so.

MDrakos: Thank you! 🙏

MDrakos (approved):

PR #3406, opened 2024-07-19
StageCommit should return the commit so we do not have call FetchCommit.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2934" title="DX-2934" target="_blank">TaskDX-2934</a>  Update StageCommit to return buildplan / build expression
  </td></table>
  <br />


mitchell-as:
Fixed in #3420.

PR #3410, opened 2024-07-22
`state import` should solve the correct commitID.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2550" title="DX-2550" target="_blank">StoryDX-2550</a>  Change and CVE information for `state import`
  </td></table>
  <br />

### DIFF
--- a/internal/runners/packages/import.go
+++ b/internal/runners/packages/import.go
@@ -138,7 +138,7 @@ func (i *Import) Run(params *ImportRunParams) error {
 	pg = nil
 
 	// Solve the runtime.
-	rtCommit, err := bp.FetchCommit(latestCommit, proj.Owner(), proj.Name(), nil)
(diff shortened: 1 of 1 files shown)


internal/runners/packages/import.go
```diff
@@ -138,7 +138,7 @@ func (i *Import) Run(params *ImportRunParams) error {
 	pg = nil
 
 	// Solve the runtime.
-	rtCommit, err := bp.FetchCommit(latestCommit, proj.Owner(), proj.Name(), nil)
```
MDrakos, line 141: nit: Perhaps `latestCommit` should be renamed to `localCommit` as after we stage a commit this name is no longer valid.

MDrakos (approved):

MDrakos (approved):

PR #3413, opened 2024-07-24
Always update the runtime for the commands that use it.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2967" title="DX-2967" target="_blank">BugDX-2967</a>  `state shell` with `optin.unstable.async_runtime` does not work.
  </td></table>
  <br />
 

Ignoring the async runtime config option makes those commands unusable.

### DIFF
--- a/test/integration/shell_int_test.go
+++ b/test/integration/shell_int_test.go
@@ -504,6 +504,28 @@ func (suite *ShellIntegrationTestSuite) TestWindowsShells() {
 	cp.ExpectExitCode(0)
 }
 
+func (suite *ShellIntegrationTestSuite) TestAsync() {
+	suite.OnlyRunForTags(tagsuite.Shell)
+
+	ts := e2e.New(suite.T(), false)
+	defer ts.Close()
+
+	cp := ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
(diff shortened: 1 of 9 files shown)


test/integration/shell_int_test.go
```diff
@@ -504,6 +504,28 @@ func (suite *ShellIntegrationTestSuite) TestWindowsShells() {
 	cp.ExpectExitCode(0)
 }
 
+func (suite *ShellIntegrationTestSuite) TestAsync() {
```
mitchell-as, line 507: I'm wary of adding a test like this because it'll take a few minutes to run (particularly on Windows), but I don't see another way to test this without enabling `optin.unstable.async_runtime` for all tests. Doing so might be very disruptive as we currently have an "opt-in" mechanic for async runtimes.

mitchell-as:
The lone test failure is a timeout and unrelated to this PR.

Naatan (changes requested):
See my comment here https://activestatef.atlassian.net/browse/DX-2967?focusedCommentId=54360

I suggest abandoning this PR and the ticket.

mitchell-as:
Okay, abandoning.

PR #3414, opened 2024-07-24
Only remove the installation.InstallDirMarker after binaries were successfully removed.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2870" title="DX-2870" target="_blank">BugDX-2870</a>  Could not register config listener:  Could not determine install root from exec:
  </td></table>
  <br />
 

Otherwise, if removal of state-svc failed and it's scheduled for auto-start, it'll constantly fail to detect the install root and send the error to rollbar.

### DIFF
--- a/internal/assets/contents/scripts/removePaths.bat
+++ b/internal/assets/contents/scripts/removePaths.bat
@@ -32,17 +32,19 @@ echo "Waiting for process %exe% with PID %pid% to end..." >> %logfile%
 echo "Process %exe% has ended" >> %logfile%
 set success=true
 for %%i in (%paths%) do (
-    echo "Attempting to remove path %%i" >> %logfile%
-    if exist "%%i\" (
-        rmdir /s /q %%i 2>>&1 >> %logfile%
-    ) else if exist "%%i" (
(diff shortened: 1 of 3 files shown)


internal/assets/contents/scripts/removePaths.bat
```diff
@@ -32,17 +32,19 @@ echo "Waiting for process %exe% with PID %pid% to end..." >> %logfile%
...
-        rmdir /s /q %%i 2>>&1 >> %logfile%
-    ) else if exist "%%i" (
-        del /f /q %%i 2>>&1 >> %logfile%
-    )
-    if exist "%%i" (
-        echo "Could not remove path: %%i" >> %logfile%
-        set success=false
-    ) else (
-        echo "Successfully removed path %%i" >> %logfile%
+    if "%success%"=="true" (
```
mitchell-as, line 35: batch has no way to break out of a for loop, so we'll just skip over removal of the rest of the paths if one removal failed.

MDrakos (approved):

PR #3415, opened 2024-07-24
Enrich 'installation path must be an empty directory' errors to help troubleshoot.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2881" title="DX-2881" target="_blank">BugDX-2881</a>  Installation path must be an empty directory
  </td></table>
  <br />

### DIFF
--- a/cmd/state-installer/cmd.go
+++ b/cmd/state-installer/cmd.go
@@ -251,6 +251,8 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
 
 	an.Event(anaConst.CatInstallerFunnel, "exec")
 
+	usingDefaultInstallPath := params.path == ""
+
 	if params.path == "" {
 		var err error
 		params.path, err = installation.InstallPathForChannel(constants.ChannelName)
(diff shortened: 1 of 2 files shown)


cmd/state-installer/cmd.go
```diff
@@ -296,6 +298,17 @@ func execute(out output.Outputer, cfg *config.Instance, an analytics.Dispatcher,
...
 		if !empty {
+			if usingDefaultInstallPath {
+				// We're having trouble pinning down why these errors are occurring, so report the list of
+				// existing files to Rollbar to help diagnose.
+				if files, err := os.ReadDir(params.path); err == nil {
+					fileList := []string{}
+					for _, file := range files {
+						fileList = append(fileList, filepath.Join(params.path, file.Name()))
+					}
+					rollbar.Critical("Installation path must be an empty directory: %s\nExisting files:\n%s", params.path, strings.Join(fileList, "\n"))
```
Naatan, line 309:
```suggestion
				if files, err := fileutils.ListDirSimple(params.path, false); err == nil {
					rollbar.Critical("Installation path must be an empty directory: %s\nExisting files:\n%s", params.path, strings.Join(files, "\n"))
```

cmd/state-installer/installer.go
```diff
@@ -235,6 +235,10 @@ func installedOnPath(installRoot, channel string) (bool, string, error) {
 		return false, "", nil
 	}
 
+	if !fileutils.FileExists(filepath.Join(installRoot, installation.InstallDirMarker)) {
+		return false, "", nil
+	}
```
Naatan, line 240: What I meant with the AC is to return True if this was encountered. So add it to the candidates below.

I think we're avoiding asserting false if it didn't exist in order to support older state tool versions that didn't supply this marker.

Naatan (changes requested):

cmd/state-installer/cmd.go
mitchell-as: `fileutils.ListDirSimple()` does a recursive listing, which this ticket explicitly said not to do. Should I still use your suggestion?

Naatan: Ahh my bad, no; thanks for spotting that!

Naatan (approved):

PR #3418, opened 2024-07-29
`state import` with SBOMs should still update the commit.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2977" title="DX-2977" target="_blank">BugDX-2977</a>  Nightly failure: CycloneDX and SPDX SBOM imports
  </td></table>
  <br />
 

We expect the solve to fail, but we should still update the commit so the user can see what was added, even if the runtime is no longer viable.

Note: these tests were passing prior to the runtime refactor landing and the removal of `constants.DisableRuntime` in favor of the async runtime config option. The local commit was still being updated despite the runtime failing to solve. Now this is no longer the case, so this PR is trying to work around that.

### DIFF
--- a/internal/runners/packages/import.go
+++ b/internal/runners/packages/import.go
@@ -1,8 +1,10 @@
 package packages
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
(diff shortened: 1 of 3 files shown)


internal/runners/packages/import.go
```diff
@@ -143,7 +148,24 @@ func (i *Import) Run(params *ImportRunParams) (rerr error) {
...
+		if errors.As(err, &buildplannerErr) {
+			// When importing CycloneDX and SPDX SBOMs, we put all packages in the 'private/<org>'
+			// namespace, which will fail to solve. That is expected. However, we still want to update the
+			// local commit so the user can see what was imported, even if the runtime is not viable.
+			for _, verr := range buildplannerErr.ValidationErrors {
+				if strings.Contains(verr, "non-existent namespace: private/") {
+					if err := localcommit.Set(proj.Dir(), stagedCommitId.String()); err != nil {
+						return locale.WrapError(err, "err_package_update_commit_id")
+					}
+					return errImportSbomSolve
```
Naatan, line 161: This is too specific. We should not have awareness of sboms on the state tool side, and definitely should not be string matching errors from the API to localize them here. 

I think we should just make this the mechanic for import; we always update the commit regardless of whether the builder errored out. And we forward the API error, which should already be helpful enough on its own.

Naatan (changes requested):

Naatan (approved):

PR #3419, opened 2024-07-29
Use buildplanner ImpactReport endpoint to show change summary.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2955" title="DX-2955" target="_blank">StoryDX-2955</a>  Our change summary and cve summary use the buildplanner ImpactReport
  </td></table>
  <br />


pkg/platform/model/buildplanner/impactreport.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/request"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/response"
+)
+
+type ImpactReportParams struct {
+	Owner      string
+	Project    string
+	BeforeExpr []byte
+	AfterExpr  []byte
```
mitchell-as, line 13: We cannot use commit IDs because at this time they must be publicly attached to the project. Local commit IDs will not work. I've filed PB-5177 for this.

internal/runbits/dependencies/changesummary.go
```diff
@@ -1,43 +1,86 @@
...
+	params := &buildplanner.ImpactReportParams{
+		Owner:      prime.Project().Owner(),
+		Project:    prime.Project().Name(),
+		BeforeExpr: beforeExpr,
+		AfterExpr:  afterExpr,
+	}
+	report, err := bpm.ImpactReport(params)
+	if err != nil {
+		return errs.Wrap(err, "Failed to fetch impact report")
+	}
```
Naatan, line 53: Can we add a model function for this that abstracts all of this away? We'll be doing the same thing in cves.go.

internal/runbits/dependencies/changesummary.go
```diff
@@ -1,43 +1,86 @@
...
+	for _, i := range report.Ingredients {
+		if i.Before != nil {
+			alreadyInstalledVersions[strfmt.UUID(i.Before.IngredientID)] = i.Before.Version
+		}
+
+		if i.After == nil || !i.After.IsRequirement {
+			continue
+		}
+
+		if i.Before == nil {
```
Naatan, line 72: May as well pair this up with the identical conditional on line 64.

Naatan (changes requested):
In addition to flagged code, I think this should also be updating this code here:

https://github.com/ActiveState/cli/blob/dd0c713131c7a7f72af5c7f73b4967ef2bfa001c/internal/runbits/cves/cves.go#L43-L43

I'd also expect the `DiffArtifacts()` function to be dropped as part of this PR.

internal/runbits/dependencies/changesummary.go
mitchell-as: I would, but they're not identical :scream:

internal/runbits/dependencies/changesummary.go
mitchell-as: I've also changed the change and cve reports to accept an impact report so it's only computed once.

mitchell-as:
Test failures are a mix of known issues and timeouts. They are not caused by this PR.

pkg/platform/model/buildplanner/impactreport.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/json"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/pkg/buildscript"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/request"
+	"github.com/ActiveState/cli/pkg/platform/api/buildplanner/response"
+)
+
+type buildScripter interface {
+	BuildScript() *buildscript.BuildScript
```
mitchell-as, line 13: Using `*buildplanner.Commit` would result in an import cycle.

pkg/platform/model/buildplanner/impactreport.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type ImpactReportParams struct {
+	Owner   string
+	Project string
+	Before  buildScripter
```
mitchell-as, line 19: Using `*buildplanner.Commit` would result in an import cycle.

internal/runbits/cves/cves.go
```diff
@@ -39,36 +40,27 @@ func NewCveReport(prime primeable) *CveReport {
...
+		ingredients = append(ingredients, &request.Ingredient{
+			Namespace: i.Namespace,
+			Name:      i.Name,
+			Version:   i.After.Version,
+		})
+	}
+
+	if !c.prime.Auth().Authenticated() || len(ingredients) == 0 {
+		logging.Debug("Skipping CVE reporting")
+		return nil
```
Naatan, line 63: Can we put the auth check at the top and the `len()` check here? For the len check we don't need a debug entry.

This is pretty nitpicky I realize but it feels wrong to me to have both fire the same debug log, and especially for there not being any ingredients it doesn't seem worth logging at all.

internal/runbits/dependencies/changesummary.go
```diff
@@ -1,59 +1,28 @@
...
-		AfterExpr:  afterExpr,
-	}
-	report, err := bpm.ImpactReport(params)
-	if err != nil {
-		return errs.Wrap(err, "Failed to fetch impact report")
-	}
-
-	buildPlan := newCommit.BuildPlan()
-
+func OutputChangeSummary(out output.Outputer, report *response.ImpactReportResult, rtCommit *buildplanner.Commit) {
```
Naatan, line 25: It looks like we only use `rtCommit` to retrieve the buildplan, so we should just pass the buildplan here.

pkg/platform/model/buildplanner/impactreport.go
Naatan: Assuming there isn't an import cycle with buildscripts you could pass in the buildscript instead? It looks like that's all you're accessing on the commit.

I'd like to find a way to address this without interfaces if we can find a feasible solution.

Naatan (changes requested):
Looking good! I think we're almost there.

Naatan (approved):
Nice work!

PR #3431, opened 2024-08-01
Do not assume there is a local history when checking if a commit is within project history.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2992" title="DX-2992" target="_blank">BugDX-2992</a>  Nightly failure: TestArtifactsIntegrationTestSuite/TestArtifacts_Download_Remote
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/model/vcs.go
+++ b/pkg/platform/model/vcs.go
@@ -888,16 +888,18 @@ func GetCommitWithinCommitHistory(currentCommitID, targetCommitID strfmt.UUID, a
 // This function exists primarily as an existence check because the buildplanner API currently
 // accepts a query for a org/project#commitID even if commitID does not belong to org/project.
 // See DS-1705 (yes, DS, not DX).
-func GetCommitWithinProjectHistory(commitID strfmt.UUID, owner, name string, localCommitID strfmt.UUID, auth *authentication.Auth) (*mono_models.Commit, error) {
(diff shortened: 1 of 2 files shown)


pkg/platform/model/vcs.go
```diff
@@ -888,16 +888,18 @@ func GetCommitWithinCommitHistory(currentCommitID, targetCommitID strfmt.UUID, a
 // This function exists primarily as an existence check because the buildplanner API currently
 // accepts a query for a org/project#commitID even if commitID does not belong to org/project.
 // See DS-1705 (yes, DS, not DX).
-func GetCommitWithinProjectHistory(commitID strfmt.UUID, owner, name string, localCommitID strfmt.UUID, auth *authentication.Auth) (*mono_models.Commit, error) {
+func GetCommitWithinProjectHistory(commitID strfmt.UUID, owner, name string, localCommitID *strfmt.UUID, auth *authentication.Auth) (*mono_models.Commit, error) {
```
MDrakos, line 891: nit: We may want to consider renaming `commitID` to `searchCommitID` as it is hard to determine what the difference in `commitID` and `localCommitID` are. In fact, for the `state export buildplan` use-case I believe these two commit IDs will be the same, which further adds to the confusion.

MDrakos (approved):

PR #3432, opened 2024-08-01
Use volume-specific depot for Windows runtimes not on the main volume.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2943" title="DX-2943" target="_blank">TaskDX-2943</a>  We support setting up a runtime on the non-windows drive
  </td></table>
  <br />

### DIFF
--- a/pkg/runtime/depot.go
+++ b/pkg/runtime/depot.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -38,6 +39,15 @@ const (
 	deploymentTypeLink deploymentType = "link"
 	deploymentTypeCopy                = "copy"
 )
+
+type ErrVolumeMismatch struct {
+	DepotVolume string
+	PathVolume  string
+}
+
+func (e ErrVolumeMismatch) Error() string {
+	return fmt.Sprintf("volume mismatch: path volume is '%s', but depot volume is '%s'", e.PathVolume, e.DepotVolume)
+}
 
 type depot struct {
 	config    depotConfig
@@ -140,6 +150,10 @@ func (d *depot) DeployViaLink(id strfmt.UUID, relativeSrc, absoluteDest string) 
 
 	if !d.Exists(id) {
 		return errs.New("artifact not found in depot")
+	}
+
+	if err := d.validateVolume(absoluteDest); err != nil {
+		return errs.Wrap(err, "volume validation failed")
 	}
 
 	// Collect artifact meta info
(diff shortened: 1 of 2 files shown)


mitchell-as:
The lone windows failure is sporadic and unrelated to this PR.

pkg/runtime/runtime.go
```diff
@@ -46,7 +47,16 @@ func New(path string) (*Runtime, error) {
...
 
-	depot, err := newDepot()
+	// Windows does not support hard-linking across drives, so determine if the runtime path is on a
+	// separate drive than the default depot path. If so, use a drive-specific depot path when
+	// initializing the depot.
+	runtimeVolume := filepath.VolumeName(path)
+	storageVolume := filepath.VolumeName(storage.CachePath())
+	if runtimeVolume == storageVolume {
+		runtimeVolume = ""
+	}
```
Naatan, line 57: I think this logic can live within depot, no need for runtime to deal with these concerns.

Naatan: Also does this need to be windows specific logic? What does `filepath.VolumeName()` do on non-windows systems? Probably safer to avoid using this logic outside of windows, even if the APIs are satisfied.

Naatan (changes requested):

pkg/runtime/runtime.go
mitchell-as: The documentation says that `filepath.VolumeName()` returns the empty string for non-Windows OSes. However, I can wall it off behind a `runtime.GOOS` check.

pkg/runtime/runtime.go
```diff
@@ -46,7 +46,8 @@ func New(path string) (*Runtime, error) {
 		return nil, errs.Wrap(err, "Could not create runtime directory")
 	}
 
-	depot, err := newDepot()
+	runtimePath := path
+	depot, err := newDepot(runtimePath)
```
mitchell-as, line 50: Originally I wanted to pass `path` to `newDepot()`, but it didn't read right, so that's why I did the logic the way you originally saw it. I agree that the runtime shouldn't be concerned with what the depot can or cannot do, so I came up with something more readable here.

pkg/runtime/depot.go
```diff
@@ -45,9 +46,19 @@ type depot struct {
...
 
+	// Windows does not support hard-linking across drives, so determine if the runtime path is on a
+	// separate drive than the default depot path. If so, use a drive-specific depot path.
+	if runtime.GOOS == "windows" {
+		runtimeVolume := filepath.VolumeName(runtimePath)
+		storageVolume := filepath.VolumeName(storage.CachePath())
+		if runtimeVolume != storageVolume {
+			depotPath = filepath.Join(runtimeVolume+"\\", "activestate", depotName)
+		}
+	}
```
Naatan, line 60: I can see where you were coming from, it feels awkward dealing with the runtimePath here, since depot shouldn't have any concerns about that. But this might be an ok compromise for now, the only other solution I see is introduce some new construct that is responsible for this information, which feels like overkill for the current use-case.

That said, may be good to highlight this compromise as a comment so we hopefully avoid us further building on this by accident. ie. if we need to extend this logic then we should probably think more about the separate construct.

mitchell-as:
@Naatan I'm not sure what you're asking for. Are you asking for a `func (d *depot) Validate(absoluteDest string) error {}` below the two Deploy functions? Or are you asking for a `if filepath.VolumeName(absoluteDest) != filepath.VolumeName(d.depotPath) { return errs.New("volume mismatch") }` inside each of the Deploy functions?

If the former, what calls the validator and where? This is my first experience with the new runtime. I'm a n00b.

mitchell-as:
@Naatan ping

Naatan:
@mitchell-as sorry; I meant the latter; add some code to the two deploy functions to protect against miss-use. I suspect the low level error we would encounter would not be intuitive to interpret otherwise.

mitchell-as:
Thanks. Turns out that in order to avoid code duplication, I ended up doing both, heh.

Naatan (approved):

PR #3441, opened 2024-08-07
Poll the impact report if it returns a BuildPlanning status.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2996" title="DX-2996" target="_blank">BugDX-2996</a>  Sporadic failures while installing or uninstalling packages.
  </td></table>
  <br />
 

We poll up to 10 times starting at 500ms, increasing poll time for each retry. Ultimately, we try for up to ~30s before giving up.

### DIFF
--- a/pkg/platform/api/buildplanner/response/impactreport.go
+++ b/pkg/platform/api/buildplanner/response/impactreport.go
@@ -1,6 +1,8 @@
 package response
 
 import (
+	"errors"
+
 	"github.com/ActiveState/cli/internal/errs"
 )
 
@@ -17,10 +19,23 @@ type ImpactReportIngredient struct {
 	After     *ImpactReportIngredientState `json:"after"`
 }
 
+const buildPlannedType = "BuildPlanned"
+const buildPlanningType = "BuildPlanning"
+
+type ImpactReportBuildResult struct {
+	Type string `json:"__typename"`
+}
+
+type ImpactReportBuildResultError struct {
+	BuildBefore *ImpactReportBuildResult `json:"buildBefore"`
+	BuildAfter  *ImpactReportBuildResult `json:"buildAfter"`
+}
+
 type ImpactReportResult struct {
 	Type        string                   `json:"__typename"`
 	Ingredients []ImpactReportIngredient `json:"ingredients"`
 	*Error
+	*ImpactReportBuildResultError
 }
 
 type ImpactReportResponse struct {
@@ -34,10 +49,26 @@ type ImpactReportError struct {
 
 func (e ImpactReportError) Error() string { return e.Message }
 
+func IsImpactReportBuildPlanningError(err error) bool {
+	var impactReportErr *ImpactReportError
+	return errors.As(err, &impactReportErr) && impactReportErr.Type == buildPlanningType
+}
+
 func ProcessImpactReportError(err *ImpactReportResult, fallbackMessage string) error {
 	if err.Error == nil {
 		return errs.New(fallbackMessage)
 	}
 
-	return &ImpactReportError{err.Type, err.Message}
+	errType := err.Type
+	message := err.Message
+	buildResultError := err.ImpactReportBuildResultError
+	if beforeType := buildResultError.BuildBefore.Type; beforeType != buildPlannedType {
+		errType = beforeType
+		message += "\nbuildBefore status: " + beforeType
+	} else if afterType := buildResultError.BuildAfter.Type; afterType != buildPlannedType {
+		errType = afterType
+		message += "\nbuildAfter status: " + afterType
+	}
+
+	return &ImpactReportError{errType, message}
 }
--- a/pkg/platform/model/buildplanner/impactreport.go
+++ b/pkg/platform/model/buildplanner/impactreport.go
@@ -2,8 +2,10 @@ package buildplanner
 
 import (
 	"encoding/json"
+	"time"
 
 	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/logging"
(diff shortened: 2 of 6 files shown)


pkg/platform/model/buildplanner/impactreport.go
```diff
@@ -27,9 +32,26 @@ func (b *BuildPlanner) ImpactReport(params *ImpactReportParams) (*response.Impac
...
+	for i := 0; i < numRetries; i++ {
+		resp, err = run(b.client, params.Owner, params.Project, beforeExpr, afterExpr)
+		if err == nil || !response.IsImpactReportBuildPlanningError(err) {
+			break
+		}
+		logging.Debug("Impact report response was that the buildplanner was still planning; trying again")
+		time.Sleep(retryInterval * time.Duration(i+1))
+	}
+	if err != nil {
+		return nil, errs.Wrap(err, "failed to get impact report")
```
mitchell-as, line 45: The error will look like this:

```
█ Something Went Wrong

  Failed to fetch impact report:
      failed to get impact report: |-
          Could not resolve one or both input buildplans
          buildAfter status: BuildPlanning.
```

How should we present this to the user? In my experience, retrying the command worked.

Naatan: Let's make this user a user facing error that reads:

> Timed out waiting for build to resolve, please try again in a few minutes.

pkg/platform/model/buildplanner/impactreport.go
```diff
@@ -27,9 +32,26 @@ func (b *BuildPlanner) ImpactReport(params *ImpactReportParams) (*response.Impac
...
 
-	request := request.ImpactReport(params.Owner, params.Project, beforeExpr, afterExpr)
+	var resp *response.ImpactReportResponse
+	for i := 0; i < numRetries; i++ {
+		resp, err = run(b.client, params.Owner, params.Project, beforeExpr, afterExpr)
+		if err == nil || !response.IsImpactReportBuildPlanningError(err) {
+			break
+		}
+		logging.Debug("Impact report response was that the buildplanner was still planning; trying again")
+		time.Sleep(retryInterval * time.Duration(i+1))
```
Naatan, line 42:
```suggestion
		time.Sleep(retryInterval * (i+1))
```

I don't think you need `time.Duration()`, and it was a bit confusing to read.

pkg/platform/api/buildplanner/response/impactreport.go
```diff
@@ -34,10 +49,26 @@ type ImpactReportError struct {
...
 		return errs.New(fallbackMessage)
 	}
 
-	return &ImpactReportError{err.Type, err.Message}
+	errType := err.Type
+	message := err.Message
+	buildResultError := err.ImpactReportBuildResultError
+	if beforeType := buildResultError.BuildBefore.Type; beforeType != buildPlannedType {
+		errType = beforeType
+		message += "\nbuildBefore status: " + beforeType
```
Naatan, line 67: Can you comment why we care about the buildBefore? I'm not sure what this is telling us exactly, but based on what I can grok from here it sounds like this is telling us what the buildstatus was before we evaluating (ie. triggered) it. Why do we care about this?

Perhaps the impactReport API doesn't trigger builds, and so this is telling us that it's not planned, which effectively means it won't be planned unless some other external call happens. In which case retrying won't make any difference.

I think either we need to document the logic or we need to revise the logic.

Naatan (changes requested):

pkg/platform/model/buildplanner/impactreport.go
mitchell-as: Unfortunately I do. I tried `retryInterval * (i+1)` first, but got a compiler error. Google says I need to cast to `time.Duration()`. I don't like it, but what can you do?

pkg/platform/api/buildplanner/response/impactreport.go
mitchell-as: The error message we were getting was:

```
Could not resolve one or both input buildplans
```

Therefore I think we need to check the status of both input build expressions. If we only check afterExpr, we could open ourselves up to a subtle bug.

Recently I had an issue with the fact that the Platform did not recognize that a local commit belonged to a project (because it wasn't pushed yet). I can envision a scenario where a local commit's build expression submitted as `beforeExpr` could fail in a similar way.

mitchell-as:
It turns out all I needed to do was add a missing atTime field to the ImpactReport query. Now it hits the correct cache and immediately returns the plan we just solved for. No polling required :)

pkg/platform/model/buildplanner/impactreport.go
mitchell-as: As discussed in stand, we need a cast to satisfy the compiler. I did use a separate variable in an attempt to improve readability, though.

pkg/platform/api/buildplanner/response/impactreport.go
mitchell-as: Is it possible to make two commits in rapid succession such that the second commit is using the first commit as `beforeExpr`, and that `beforeExpr` may not have resolved yet in time for the second commit's call to ImpactReport? I agree that ImpactReport is not doing any triggering, but we are (StageCommit, etc.). It's just that I don't think we cannot guarantee that the trigger for `beforeExpr` is fully resolved every time.

Also, I've had experience with local commits not counted as belonging to the current project, so I can envision a scenario where a local commit's build expression can be the `beforeExpr` and fail to resolve for a while.

mitchell-as:
Abandoning since we're reverting the ImpactReport changes in #3445

PR #3444, opened 2024-08-12
Initial implementation of `state checkout --from-archive`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2981" title="DX-2981" target="_blank">StoryDX-2981</a>  `state checkout` can install from an offline source
  </td></table>
  <br />
 

We needed a `project.json` file, similar to offline installers, so here's a sample of the one I'm using in integration tests:

```
{
	"org_name": "ActiveState-CLI",
	"project_name": "AlmostEmpty",
	"commit_id": "6cd1cc1f-3886-439b-8373-c24ca06ab150",
	"branch": "main",
	"platform_id": "78977bc8-0f32-519d-80f3-9043f059398c"
}
```


mitchell-as:
The lone test failure is a timeout unrelated to this PR.

pkg/runtime/setup.go
```diff
@@ -40,6 +41,11 @@ type Opts struct {
 	EventHandlers        []events.HandlerFunc
 	BuildlogFilePath     string
 
+	// Options for setting up a runtime from an archive.
+	FromArchiveDir string
+	PlatformID     *strfmt.UUID
+	ArtifactExt    string
```
Naatan, line 47: Please make a struct to put these under. As they are all directly related and should never be used independently from one another.

pkg/runtime/setup.go
```diff
@@ -267,10 +277,24 @@ func (s *setup) onArtifactBuildReady(blog *buildlog.BuildLog, artifact *buildpla
...
+	} else {
+		// Read the artifact from the archive.
+		var err error
+		name := artifact.ArtifactID.String() + s.opts.ArtifactExt
+		artifactFile := filepath.Join(s.opts.FromArchiveDir, name)
+		logging.Debug("Reading file '%s' for '%s'", artifactFile, artifact.DisplayName)
+		b, err = fileutils.ReadFile(artifactFile)
+		if err != nil {
+			return errs.Wrap(err, "read from archive failed")
+		}
```
Naatan, line 297: I think since this is bypassing the event reporting of downloads we will have a progress indicator for downloads that never moves. 

We'll want to ensure that those events are still send, but we probably also want to update our runbit logic so that instead of "Downloading" we say "Unpacking" when using the archive approach.

test/integration/checkout_int_test.go
```diff
@@ -348,6 +349,29 @@ func (suite *CheckoutIntegrationTestSuite) TestChangeSummary() {
...
+	)
+	cp.Expect("Checking out project: ActiveState-CLI/AlmostEmpty")
+	cp.Expect("Setting up the following dependencies:")
+	cp.Expect("└─ zlib@1.3.1")
+	cp.Expect("Sourcing Runtime")
+	cp.Expect("Unpacking")
+	cp.Expect("Installing")
+	cp.Expect("All dependencies have been installed and verified")
+	cp.Expect("Checked out project ActiveState-CLI/AlmostEmpty")
+	cp.ExpectExitCode(0)
```
Naatan, line 372: Let's also assert that the resulting runtime has some of the expected files. This test currently is just testing the UI/UX of the command itself, we should also verify that the end result is what we expect it to be.

Naatan (changes requested):

cmd/state/internal/cmdtree/checkout.go
```diff
@@ -40,12 +40,16 @@ func newCheckoutCommand(prime *primer.Values) *captain.Command {
 				Description: locale.Tl("flag_state_checkout_force", "Leave a failed project checkout on disk; do not delete it"),
 				Value:       &params.Force,
 			},
+			{
+				Name:        "from-archive",
+				Description: locale.Tl("flag_state_checkout_from_archive", "Checkout from the given .tar.gz archive"),
+				Value:       &params.FromArchive,
+			},
```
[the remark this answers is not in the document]
mitchell-as: That sounds reasonable, as long as the archive path always has an extension. Otherwise the argument is ambiguous, as it can be either `org/project`, `project`, or `path/to/tar.gz`.

pkg/runtime/setup.go
mitchell-as: Okay, I'll add the download events. In testing, this step went by so fast I didn't even see a download bar. I only saw a brief "Unpacking" bar since setting up from an archive does go through this step.

internal/runbits/checkout/checkout.go
```diff
@@ -52,11 +52,13 @@ func (e errCommitDoesNotBelong) Error() string {
...
 }
 
+var errNoCommitID = errs.New("commitID is nil")
+
 func New(repo git.Repository, prime primeable) *Checkout {
 	return &Checkout{repo, prime.Output(), prime.Config(), prime.Analytics(), "", prime.Auth()}
 }
 
-func (r *Checkout) Run(ns *project.Namespaced, branchName, cachePath, targetPath string, noClone bool) (_ string, rerr error) {
+func (r *Checkout) Run(ns *project.Namespaced, branchName, cachePath, targetPath string, noClone, fromArchive bool) (_ string, rerr error) {
```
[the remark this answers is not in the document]
mitchell-as: I've gone with `onlyValidate`. With all of the requested changes, this makes sense to me, but let me know if you disagree and we can reconsider your suggestions.

cmd/state/internal/cmdtree/checkout.go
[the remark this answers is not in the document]
Naatan: Sounds good, I don't see any reason why it wouldn't have an extension.

internal/runbits/checkout/checkout.go
```diff
@@ -75,102 +74,115 @@ func (r *Checkout) Run(ns *project.Namespaced, branchName, cachePath, targetPath
...
+}
 
-		// Clone the related repo, if it is defined
-		if !noClone && pj.RepoURL != nil && *pj.RepoURL != "" {
-			err := r.repo.CloneProject(ns.Owner, ns.Project, path, r.Outputer, r.analytics)
-			if err != nil {
-				return "", locale.WrapError(err, "err_clone_project", "Could not clone associated git repository")
+// checkout performs the checkout and returns the checkout's owner, project, commitID, branchName,
+// and language.
+func (r *Checkout) checkout(
```
Naatan, line 101: It doesn't feel right to call this `checkout`. That's what the main `Run` function does (ie. creating the project file is effectively the checkout).

Suggest turning this into a function called `sanitize()` that does all the cleaning up of parameters, and then separately call the clone behaviour from `Run()`.

I'm not a big fan of the number of parameters this returns, but I also don't see an obvious solution that doesn't add boilerplate that I'm even less of a fan of. Just flagging it in case you have ideas.

internal/runbits/checkout/checkout.go
```diff
@@ -55,10 +54,10 @@ func (e errCommitDoesNotBelong) Error() string {
 var errNoCommitID = errs.New("commitID is nil")
 
 func New(repo git.Repository, prime primeable) *Checkout {
-	return &Checkout{repo, prime.Output(), prime.Config(), prime.Analytics(), "", prime.Auth()}
+	return &Checkout{repo, prime.Output(), prime.Config(), prime.Analytics(), prime.Auth()}
 }
 
-func (r *Checkout) Run(ns *project.Namespaced, branchName, cachePath, targetPath string, noClone, fromArchive bool) (_ string, rerr error) {
+func (r *Checkout) Run(ns *project.Namespaced, branchName, cachePath, targetPath string, noClone, onlyValidate bool) (_ string, rerr error) {
```
Naatan, line 60: `onlyValidate` doesn't seem an appropriate name to me. If supplied it in fact does no validations, and it only does the checkout (ie. create the project files).

Perhaps `bareCheckout` would be more appropriate?

internal/runners/checkout/checkout.go
```diff
@@ -77,30 +77,35 @@ func NewCheckout(prime primeable) *Checkout {
...
 		}
 		defer archive.Cleanup()
-		params.Namespace = archive.Namespace
+		ns = *archive.Namespace
 		params.Branch = archive.Branch
-	} else if params.Namespace.Project == "" {
-		return locale.NewInputError("err_checkout_namespace_required", "", locale.T("arg_state_checkout_namespace"), locale.T("arg_state_checkout_namespace_description"))
+	} else {
+		err := ns.Set(params.Namespace)
+		if err != nil {
```
Naatan, line 94:
```suggestion
		if err := ns.Set(params.Namespace); err != nil {
```

internal/runners/checkout/checkout.go
```diff
@@ -77,30 +77,35 @@ func NewCheckout(prime primeable) *Checkout {
...
+	if strings.HasSuffix(params.Namespace, checkout.ArchiveExt) {
 		var err error
-		archive, err = checkout.NewArchive(params.FromArchive)
+		archive, err = checkout.NewArchive(params.Namespace)
 		if err != nil {
 			return errs.Wrap(err, "Unable to read archive")
 		}
 		defer archive.Cleanup()
-		params.Namespace = archive.Namespace
+		ns = *archive.Namespace
```
Naatan, line 90:
```suggestion
		ns = archive.Namespace
```

Doesn't seem work dereferencing when all your use of it doesn't seem to care, and in fact you create a reference again when passing it to `checkout.Run()`. May as well avoid the potential nil pointer.

pkg/runtime/options.go
```diff
@@ -17,12 +17,8 @@ func WithPreferredLibcVersion(version string) SetOpt {
...
 }
 
-func WithFromArchive(dir string, platformID strfmt.UUID, ext string) SetOpt {
-	return func(opts *Opts) {
-		opts.FromArchiveDir = dir
-		opts.PlatformID = &platformID
-		opts.ArtifactExt = ext
-	}
+func WithArchive(dir string, platformID strfmt.UUID, ext string) SetOpt {
+	return func(opts *Opts) { opts.FromArchive = &fromArchive{dir, platformID, ext} }
```
Naatan, line 21:
```suggestion
	return func(opts *Opts) { 
		opts.FromArchive = &fromArchive{dir, platformID, ext} 
	}
```

pkg/runtime/setup.go
```diff
@@ -287,14 +292,24 @@ func (s *setup) obtain(artifact *buildplan.Artifact) (rerr error) {
...
 		if err != nil {
+			if err2 := s.fireEvent(events.ArtifactDownloadFailure{artifact.ArtifactID, err}); err2 != nil {
+				err = errs.Pack(err, errs.Wrap(err2, "Could not handle ArtifactDownloadFailure event"))
+			}
 			return errs.Wrap(err, "read from archive failed")
 		}
+
+		if err := s.fireEvent(events.ArtifactDownloadSuccess{artifact.ArtifactID}); err != nil {
+			return errs.Wrap(errs.Pack(err, err), "Could not handle ArtifactDownloadSuccess event")
+		}
```
Naatan, line 312: I just realized we are in fact handling unpack events separately. So I think it makes more sense that we don't send the download events and instead set the toDownload counter to 0 if we are installing from an archive, ie. https://github.com/ActiveState/cli/blob/6ff1b5ad197179213e3d93cf0ff22c516bff4e4b/pkg/runtime/setup.go#L114-L114

Apologies, I didn't have this in mind with my initial review.

test/integration/checkout_int_test.go
```diff
@@ -370,6 +370,13 @@ func (suite *CheckoutIntegrationTestSuite) TestCheckoutFromArchive() {
 	cp.Expect("All dependencies have been installed and verified")
 	cp.Expect("Checked out project ActiveState-CLI/AlmostEmpty")
 	cp.ExpectExitCode(0)
+
+	// Verify the zlib runtime files exist.
+	proj, err := project.FromPath(filepath.Join(ts.Dirs.Work, "AlmostEmpty"))
+	suite.Require().NoError(err)
+	cachePath := filepath.Join(ts.Dirs.Cache, runtime_helpers.DirNameFromProjectDir(proj.Dir()))
+	zlibH := filepath.Join(cachePath, "usr", "include", "zlib.h")
```
Naatan, line 378: I'm pretty sure this will be different on mac/windows.

Naatan (changes requested):

internal/runbits/checkout/checkout.go
mitchell-as: I've gone with `fetchProject` since I feel like that is the goal of this function. Fetch the project from the platform and return its complete set of parameters.

internal/runbits/checkout/checkout.go
mitchell-as: Okay.

test/integration/checkout_int_test.go
mitchell-as: Actually, it can be whatever I say it is since I produced the test archives :) But seriously, I untarred the ingredients and looked for a common one that had the same path and used that. The tests on all three platforms are passing, so it's all good.

internal/runners/checkout/checkout.go
mitchell-as: `ns` cannot be a pointer because of the `ns.Set()` function used earlier. The dereference here is needed, as is the passing of `&ns` to `checkout.Run()`. It's a product of our clever use the first argument being either a namespace or archive file.

pkg/runtime/setup.go
mitchell-as: Okay, well that complicated things. Let me know if my solution is sub-optimal.

test/integration/checkout_int_test.go
Naatan: Oh right, I forgot about that detail :p Ok we'll worry about this once we can test with proper payloads.

mitchell-as:
Closing in favor of #3451

Naatan:
@mitchell-as reopened as that other ticket was targeting this one so the review process was still workable. Though given I'm about to approve I guess this is just busywork 😅 Anyway now you have a signal to move the story along.

Naatan (approved):

PR #3447, opened 2024-08-14
Fix dependency names on `state commit`

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2966" title="DX-2966" target="_blank">BugDX-2966</a>  COMMIT: We get the message with a list of all dependencies and it is also not properly wrapped if too long.
  </td></table>
  <br />


internal/runners/commit/commit.go
```diff
@@ -159,7 +160,19 @@ func (c *Commit) Run() (rerr error) {
 	dependencies.OutputChangeSummary(out, rtCommit.BuildPlan(), oldCommit.BuildPlan())
 
 	// Report CVEs.
-	if err := cves.NewCveReport(c.prime).Report(rtCommit.BuildPlan(), oldCommit.BuildPlan()); err != nil {
+	newReqs, err := newRequirements(oldCommit.BuildScript(), rtCommit.BuildScript())
```
mitchell-as, line 163: This doesn't feel right, because we're going to need to duplicate this logic in other places that will eventually get CVE reporting. It looks like `state import` is not passing any names to `cves.NewCveReport()` either, so presumably it suffers from the same bug.

Can we put all of this logic in the `Report()` method? Kind of like how the change summary reporter diffs two buildplans? (https://github.com/ActiveState/cli/blob/0bb4d847b896c83a4d31cbba042b8ba65b06c933/internal/runbits/dependencies/changesummary.go#L30). If we need absolutely need buildscripts, then maybe we should be passing commits to the method instead of buildplans.

mitchell-as (changes requested):

internal/runbits/runtime/requirements/requirements.go
```diff
@@ -246,7 +246,7 @@ func (r *RequirementOperation) ExecuteRequirementOperation(ts *time.Time, requir
 
 	// Report CVEs
 	names := requirementNames(requirements...)
-	if err := cves.NewCveReport(r.prime).Report(commit.BuildPlan(), oldCommit.BuildPlan(), names...); err != nil {
+	if err := cves.NewCveReport(r.prime).Report(commit, oldCommit, names...); err != nil {
```
mitchell-as, line 249: Is there any way we can compute the `names` here inside `Report()` and remove these extra arguments? Maybe diffing the buildscripts will be enough?

It's unclear to me when we need to pass names, and when we do not. Currently only requirements passes names, and when other commands start calling this function, are they supposed to pass names or omit them? There's an ambiguity here and I'd either like to remove it, or add a documentation comment to `Report()` as to when names are required an when they can be omitted.

mitchell-as (changes requested):

internal/runbits/runtime/requirements/requirements.go
MDrakos: Now that we are using the build plan to get the requested requirements we should be okay to remove the `names` parameter.

mitchell-as (approved):
This looks fine to me, but before merging, please run the integration tests for install,import,package. They all use this code path and we want to make sure we didn't break anything.

PR #3448, opened 2024-08-15
Change summary shows added dependencies for updates.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2930" title="DX-2930" target="_blank">TaskDX-2930</a>  Change summary should be shown if package is updated
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/dependencies/changesummary.go
+++ b/internal/runbits/dependencies/changesummary.go
@@ -41,6 +41,27 @@ func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan,
 		}
 	}
 
+	if len(changeset.Added) > 0 && len(addedLocale) == 0 {
+		for _, a := range changeset.Added {
+			// Check if the added artifact is a direct dependency of a requested package.
+			// This can happen when an updated package brings in a new dependency.
+			for _, req := range requested {
+				if !isDirectDependency(a, req) {
+					continue
+				}
+				v := fmt.Sprintf("%s@%s", req.Name(), req.Version()) // requested package, not added package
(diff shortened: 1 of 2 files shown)


mitchell-as:
The lone test failure is sporadic and not due to this PR.

internal/runbits/dependencies/changesummary.go
```diff
@@ -41,6 +41,27 @@ func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan,
...
+				addedLocale = append(addedLocale, fmt.Sprintf("[ACTIONABLE]%s[/RESET]", v))
+
+				for _, i := range a.Ingredients { // added package, not requested package
+					dependencies = append(dependencies, i)
+					directDependencies = append(dependencies, i)
+				}
+				break
+			}
+		}
+	}
```
Naatan, line 63: While this might get us the result we want, it feels too implicit. Is there a reason we can't iterate over `changeset.Updated` the same way we do over `changeset.Added`? We'll still need to calculate dependencies that were actually changed, but at least that way we are explicitly interpreting just the requirements that were updated.

I'm worried the current solution will break in the future or cause edge case bugs.

Naatan (changes requested):

internal/runbits/dependencies/changesummary.go
mitchell-as: Okay, I've switched to iterating over `changeset.Upated` and then looking for a direct dependency in `changeset.Added`. Let me know if this isn't what you meant.

Naatan (approved):

PR #3451, opened 2024-08-16
Review iteration DX-2981

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2981" title="DX-2981" target="_blank">StoryDX-2981</a>  `state checkout` can install from an offline source
  </td></table>
  <br />
 

@mitchell-as turns out I was again wrong on my recollection regarding the progress indication :p Your most recent changes are perfect, and remove the ambiguity I kept stumbling over. So your changes actually clean this up very nicely.

That said I did want to experiment with the `ns` stuff as your comment didn't quite connect for me, hence the PR.

### DIFF
--- a/internal/runners/checkout/checkout.go
+++ b/internal/runners/checkout/checkout.go
@@ -77,32 +77,35 @@ func NewCheckout(prime primeable) *Checkout {
 }
 
 func (u *Checkout) Run(params *Params) (rerr error) {
-	var ns project.Namespaced
+	var err error
+	var ns *project.Namespaced
 	var archive *checkout.Archive
 
-	if strings.HasSuffix(params.Namespace, checkout.ArchiveExt) {
-		var err error
+	switch {
+	// Checkout from archive
+	case strings.HasSuffix(params.Namespace, checkout.ArchiveExt):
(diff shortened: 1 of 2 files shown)


internal/runners/checkout/checkout.go
```diff
@@ -77,32 +77,35 @@ func NewCheckout(prime primeable) *Checkout {
...
+	case strings.HasSuffix(params.Namespace, checkout.ArchiveExt):
 		archive, err = checkout.NewArchive(params.Namespace)
 		if err != nil {
 			return errs.Wrap(err, "Unable to read archive")
 		}
 		defer archive.Cleanup()
-		ns = *archive.Namespace
+		ns = archive.Namespace
 		params.Branch = archive.Branch
-	} else if err := ns.Set(params.Namespace); err != nil {
```
Naatan, line 92: `ns.Set()` exists to satisfy and work with the captain flag/argument mechanic. There's no point in using it in this context.

We should ideally have separated the types for "namespace" and "namespace flag". Don't wanna go that far for this PR though.

mitchell-as (approved):

PR #3452, opened 2024-08-16
Fix dependency output on `state init`

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3003" title="DX-3003" target="_blank">BugDX-3003</a>  `state init` missing required dependencies tree
  </td></table>
  <br />


MDrakos:
The lone test failure is also happening on the nightlies and not related to this PR

mitchell-as (approved):

PR #3455, opened 2024-08-20
If two artifacts provide the same file, re-link or re-copy that file from the artifact left after the other was undeployed/uninstalled.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2913" title="DX-2913" target="_blank">TaskDX-2913</a>  We track individual file ownership on the artifact level
  </td></table>
  <br />
 

The idea behind this PR is to:
- Retain the current of behavior of not modifying any currently deployed files. That is, if two artifacts both provide the same file, the one installed first "wins".
- Change the behavior of deleting any currently deployed files from an artifact. If two artifacts both provide the same file, when one of them is uninstalled, re-link or copy the file from the remaining artifact that is still installed.

For example, given the following two packages:

- `state install hello`, which provides a *bin/main* executable such that `state exec main` prints `Hello world!`.
- `state install hello-again`, which provides a *bin/main* executable such that `state exec main` prints `Hello world again!`.

Here's a scenario:
1. `state install hello`
2. `state install hello-again`
3. `state exec main` --> prints `Hello world!` from `hello`, which "wins" by providing *bin/main* first.
4. `state uninstall hello` --> identifies that `hello-again` also provides *bin/main*, and re-links/re-deploys it after uninstalling `hello`.
5. `state exec main` --> prints `Hello world again!`
6. `state uninstall hello-again` --> removes everything as expected.
7. `state exec main` --> errors saying that `main` was not found.

Please note this is an "extreme" scenario. In all likelihood, artifacts providing the same file will provide the same bits, so the contents will not matter. The file merely needs to exist. Therefore, this PR effectively does reference counting and will not completely remove a file until no more artifacts reference it.

### DIFF
--- a/internal/smartlink/smartlink.go
+++ b/internal/smartlink/smartlink.go
@@ -78,7 +78,7 @@ func Link(src, dest string) error {
 // UnlinkContents will unlink the contents of src to dest if the links exist
 // WARNING: on windows smartlinks are hard links, and relating hard links back to their source is non-trivial, so instead
 // we just delete the target path. If the user modified the target in any way their changes will be lost.
-func UnlinkContents(src, dest string) error {
+func UnlinkContents(src, dest string, ignorePaths ...string) error {
 	if !fileutils.DirExists(dest) {
 		return errs.New("dest dir does not exist: %s", dest)
 	}
@@ -87,6 +87,11 @@ func UnlinkContents(src, dest string) error {
 	src, dest, err = resolvePaths(src, dest)
 	if err != nil {
 		return errs.Wrap(err, "Could not resolve src and dest paths")
+	}
+
+	ignore := make(map[string]bool)
+	for _, path := range ignorePaths {
+		ignore[path] = true
 	}
 
 	entries, err := os.ReadDir(src)
@@ -101,8 +106,12 @@ func UnlinkContents(src, dest string) error {
 			continue
 		}
 
+		if _, yes := ignore[destPath]; yes {
+			continue
+		}
+
 		if fileutils.IsDir(destPath) {
-			if err := UnlinkContents(srcPath, destPath); err != nil {
+			if err := UnlinkContents(srcPath, destPath, ignorePaths...); err != nil {
 				return err // Not wrapping here cause it'd just repeat the same error due to the recursion
 			}
 		} else {
--- a/internal/smartlink/smartlink.go
+++ b/internal/smartlink/smartlink.go
@@ -47,7 +47,7 @@ func Link(src, dest string) error {
 	}
 
 	if fileutils.IsDir(src) {
-		if err := os.MkdirAll(dest, 0755); err != nil {
+		if err := fileutils.Mkdir(dest); err != nil {
 			return errs.Wrap(err, "could not create directory %s", dest)
 		}
 		entries, err := os.ReadDir(src)
@@ -62,10 +62,9 @@ func Link(src, dest string) error {
 		return nil
 	}
 
-	if destDir := filepath.Dir(dest); !fileutils.DirExists(destDir) {
-		if err := os.MkdirAll(destDir, 0755); err != nil {
-			return errs.Wrap(err, "could not create directory %s", destDir)
-		}
+	destDir := filepath.Dir(dest)
+	if err := fileutils.MkdirUnlessExists(destDir); err != nil {
+		return errs.Wrap(err, "could not create directory %s", destDir)
 	}
 
 	// Multiple artifacts can supply the same file. We do not have a better solution for this at the moment other than
--- a/internal/smartlink/smartlink.go
+++ b/internal/smartlink/smartlink.go
@@ -62,6 +62,12 @@ func Link(src, dest string) error {
 		return nil
 	}
 
+	if destDir := filepath.Dir(dest); !fileutils.DirExists(destDir) {
+		if err := os.MkdirAll(destDir, 0755); err != nil {
+			return errs.Wrap(err, "could not create directory %s", destDir)
+		}
+	}
+
 	// Multiple artifacts can supply the same file. We do not have a better solution for this at the moment other than
 	// favouring the first one encountered.
 	if fileutils.TargetExists(dest) {
@@ -78,7 +84,7 @@ func Link(src, dest string) error {
 // UnlinkContents will unlink the contents of src to dest if the links exist
 // WARNING: on windows smartlinks are hard links, and relating hard links back to their source is non-trivial, so instead
 // we just delete the target path. If the user modified the target in any way their changes will be lost.
-func UnlinkContents(src, dest string, ignorePaths ...string) error {
+func UnlinkContents(src, dest string) error {
 	if !fileutils.DirExists(dest) {
 		return errs.New("dest dir does not exist: %s", dest)
 	}
@@ -87,11 +93,6 @@ func UnlinkContents(src, dest string, ignorePaths ...string) error {
 	src, dest, err = resolvePaths(src, dest)
 	if err != nil {
 		return errs.Wrap(err, "Could not resolve src and dest paths")
-	}
-
-	ignore := make(map[string]bool)
-	for _, path := range ignorePaths {
-		ignore[path] = true
 	}
 
 	entries, err := os.ReadDir(src)
@@ -106,12 +107,8 @@ func UnlinkContents(src, dest string, ignorePaths ...string) error {
 			continue
 		}
 
-		if _, yes := ignore[destPath]; yes {
-			continue
-		}
-
 		if fileutils.IsDir(destPath) {
-			if err := UnlinkContents(srcPath, destPath, ignorePaths...); err != nil {
+			if err := UnlinkContents(srcPath, destPath); err != nil {
 				return err // Not wrapping here cause it'd just repeat the same error due to the recursion
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


mitchell-as:
Test failures are the usual timeouts and unrelated to this PR.

mitchell-as:
This is a first attempt. I realize this PR may go through a lot of churn.

pkg/runtime/depot.go
```diff
@@ -28,9 +28,10 @@ type depotConfig struct {
 }
 
 type deployment struct {
-	Type  deploymentType `json:"type"`
-	Path  string         `json:"path"`
-	Files []string       `json:"files"`
+	Type       deploymentType `json:"type"`
+	Path       string         `json:"path"`
+	Files      []string       `json:"files"`
+	InstallDir string         `json:"installDir"`
```
mitchell-as, line 34: This is needed to compute the on-disk path to a depot artifact deployment's relative sources. envdef usually provides a root directory like "installdir", and we need to retain that value to compute source paths outside of setup (i.e. when trying to re-link or re-copy files from another artifact).

Naatan, line 34: Can you rename this to `RelativeSrc`? I avoided calling it `InstallDir` as I found that term to be confusing. My first assumption as to its meaning would not be the actual meaning.

pkg/runtime/depot.go
```diff
@@ -270,16 +273,44 @@ func (d *depot) Undeploy(id strfmt.UUID, relativeSrc, path string) error {
...
+		return errs.Wrap(err, "failed to get shared files")
+	}
+	sharedFiles := make([]string, 0)
+	for file := range relinks {
+		sharedFiles = append(sharedFiles, file)
+	}
+
 	// Perform uninstall based on deployment type
-	if err := smartlink.UnlinkContents(filepath.Join(d.Path(id), relativeSrc), path); err != nil {
+	if err := smartlink.UnlinkContents(filepath.Join(d.Path(id), relativeSrc), path, sharedFiles...); err != nil {
```
Naatan, line 293: It doesn't seem worth "ignoring" the shared files here. You're still removing it on line 299. Can't we drop the ignore mechanic and line 299 through 301 and still have the same result?

pkg/runtime/depot.go
```diff
@@ -270,16 +273,44 @@ func (d *depot) Undeploy(id strfmt.UUID, relativeSrc, path string) error {
...
 		return errs.Wrap(err, "failed to unlink artifact")
 	}
 
+	// Re-link or re-copy any files provided by other artifacts.
+	for sharedFile, relinkSrc := range relinks {
+		if err := os.Remove(sharedFile); err != nil {
+			return errs.Wrap(err, "failed to remove file")
+		}
+		switch deployments[0].Type {
```
Naatan, line 302: Nit: assign to var so we're not operating on an index here.

pkg/runtime/depot.go
```diff
@@ -270,16 +273,44 @@ func (d *depot) Undeploy(id strfmt.UUID, relativeSrc, path string) error {
...
-	deploy := sliceutils.Filter(deployments, func(d deployment) bool { return d.Path == path })
-	if len(deploy) != 1 {
+	deployments = sliceutils.Filter(deployments, func(d deployment) bool { return d.Path == path })
+	if len(deployments) != 1 {
 		return errs.New("no deployment found for %s in depot", path)
 	}
 
+	// Determine if there are any files provided by another artifact that will need to be re-linked
+	// or re-copied after this artifact is uninstalled.
+	relinks, err := d.getSharedFilesToReLink(id, relativeSrc, path)
```
Naatan, line 283: We shouldn't use the term "link" here, because they may be installed via copy. Suggest using "redeploys" and "getSharedFilesToRedeploy".

pkg/runtime/depot.go
```diff
@@ -300,6 +331,46 @@ func (d *depot) validateVolume(absoluteDest string) error {
...
+// getSharedFilesToReLink returns a map of deployed files to re-link to (or re-copy from) another
+// artifact that provides those files. The key is the deployed file path and the value is the
+// source path from another artifact.
+func (d *depot) getSharedFilesToReLink(id strfmt.UUID, relativeSrc, path string) (map[string]string, error) {
+	// Map of deployed paths to other sources that provides those paths.
+	relink := make(map[string]string, 0)
+
+	// Get a listing of all files deployed by this artifact.
+	deployedDir := filepath.Join(d.Path(id), relativeSrc)
+	deployedFiles, err := fileutils.ListDirSimple(deployedDir, false)
```
Naatan, line 343: We should calculate this purely based on the deployment config. Since we store the installed files it should have all the information we need.

Main point here is that the config is our source of truth, and if the on-disk data doesn't agree with that then we have  an issue that we'll need to raise. If we inconsistently use a different source of truth these types of conflicts will be difficult to diagnose.

Additionally, the deployed dir is shared between multiple artifacts. So the listing you get here isn't representative for the given artifact.

pkg/runtime/depot.go
```diff
@@ -300,6 +331,46 @@ func (d *depot) validateVolume(absoluteDest string) error {
...
+
+			for _, deployment := range artifactDeployments {
+				for _, relativeDeploymentFile := range deployment.Files {
+					if relativeDeployedFile == relativeDeploymentFile {
+						// We'll want to relink this other artifact's copy after undeploying the currently deployed version.
+						deployedFile := filepath.Join(path, relativeDeployedFile)
+						newSrc := filepath.Join(d.Path(artifactId), deployment.InstallDir, relativeDeployedFile)
+						logging.Debug("More than one artifact provides '%s'", relativeDeployedFile)
+						logging.Debug("Will relink %s to %s", deployedFile, newSrc)
+						relink[deployedFile] = newSrc
```
Naatan, line 364: Should probably break here. What if we have three artifacts linking to the same file? We still want to address the first one encountered.

Naatan (changes requested):

pkg/runtime/depot.go
mitchell-as: `UnlinkContents()` will remove paths and also empty directories, so we need to notify it of paths to ignore. Then, we need to remove an existing link before re-linking because `Link()` won't re-link if the target already exists.

That is the reason for why I did what I did.

pkg/runtime/depot.go
mitchell-as: I don't see the point of this. Deployment iteration is not ordered, so we cannot make any sort of guarantee which of 2 remaining artifacts will get picked.

Furthermore, we'd have to break out of two for loops, not one, so that would either require a `goto` or refactor to use functions. I don't think it's worth it since we cannot make the guarantee mentioned earlier.

pkg/runtime/depot.go
mitchell-as: Good points!

pkg/runtime/depot.go
Naatan: Yeah I think having it create directories feels more appropriate. Not deleting the directories is how git operates, so I think we can do the same.

pkg/runtime/depot.go
Naatan: Not being able to make a guarantee is already the reality we live in now. That is a problem for another day that we cannot address in this story. All we can address is that as long as there are artifacts still targeting a file, that file should exist. And one of them will arbitrarily win the race.

The main reason I brought this up is because as currently written we would remove and then link the file again here: 

https://github.com/ActiveState/cli/pull/3455/files#diff-e57fe7a86f0cf670f9bb15dec637e18669380299ce1d661b3872f4e822c65bf2R299

Naatan (changes requested):
Changes so far are looking good to me! Please see my responses to your prior comments.

pkg/runtime/depot.go
mitchell-as: Okay, I added a `goto` to break out of 3 `for` loops.

Naatan (approved):
Looks good! I have two minor nits, up to you if you address those.

internal/smartlink/smartlink.go
```diff
@@ -62,6 +62,12 @@ func Link(src, dest string) error {
 		return nil
 	}
 
+	if destDir := filepath.Dir(dest); !fileutils.DirExists(destDir) {
+		if err := os.MkdirAll(destDir, 0755); err != nil {
+			return errs.Wrap(err, "could not create directory %s", destDir)
+		}
+	}
```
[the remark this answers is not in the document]
mitchell-as: Okay. I also found a case where we can use `fileutils.Mkdir()` instead of `os.Mkdir()` with a magic number.

pkg/runtime/depot.go
```diff
@@ -355,10 +345,12 @@ func (d *depot) getSharedFilesToRedeploy(id strfmt.UUID, deploy deployment, path
 						logging.Debug("More than one artifact provides '%s'", relativeDeployedFile)
 						logging.Debug("Will redeploy '%s' to '%s'", newSrc, deployedFile)
 						redeploy[deployedFile] = newSrc
+						goto nextDeployedFile
```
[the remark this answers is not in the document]
mitchell-as: Three consecutive

```
if fileFound {
  break
}
```

looked hideous, so I tried an anonymous function. Hopefully that passes your filter :)

Naatan (approved):

PR #3456, opened 2024-08-20
Add cycle support

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3002" title="DX-3002" target="_blank">StoryDX-3002</a>  Support cycles in the build plan
  </td></table>
  <br />


MDrakos:
Looking at the test runs it appears to be known or intermittent failures. I've tested this locally with regular checkouts and PR deploy checkouts of Go projects.

pkg/buildplan/raw/mock_test.go
```diff
@@ -192,3 +192,75 @@ var buildWithRuntimeDepsViaSrc = &Build{
 		},
 	},
 }
+
+var buildWithRuntimeDepsViaSrcCycle = &Build{
```
Naatan, line 196: Could you please add a comment highlighting the cycle.

Naatan (changes requested):
I think we should also have a test covering the `Dependencies()` methods on both the Artifact and Ingredient type in the buildplan. I suspect these either need to also be cycle aware, or we'll need the test to prevent them regressing as it feels like a likely point for cycle bugs to happen.

PR #3459, opened 2024-08-21
Drop the commit progress spinner, instead everything is handled under "solve"

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3016" title="DX-3016" target="_blank">StoryDX-3016</a>  We have a multi-spinner implementation
  </td></table>
  <br />
 

Note I had initially implemented the multi-spinner, you can still see it in the first commit. I ended up reverting this for a few reasons:

* Handling multiple spinners adds a lot of complications
	* A lot of work to implement ourselves. Can use existing lib (mpb) but have to deal with its limitations.	
	* Needs special non-interactive handling (example of something not handled by mpb).
	* Needs color-awareness; which works independently from our color mechanic.
	* Bridging the common use-cases of grouped spinners, single spinners and then non-interactive variants of those two adds a lot of architectural complexity.
* The UI/UX for interactive spinners is a bit awkward. We are committing and resolving dependencies at the same time. This distinction makes little sense to end-users and will just raise eyebrows.
* The UI/UX for non-interactive will be horribly handicapped, because we cannot show progress for two distinct indicators as it is non-interactive.

Add all this up and I figured we should just be avoiding this altogether at least for our current use-case. Instead I have gotten rid of the commit spinner. There is just the "Resolving Dependencies" spinner now, which also covers the commit stage.

Additionally, I noticed non-interactive was using "frames" which do not apply in non-interactive mode. So I fixed that.


MDrakos (changes requested):
This looks good to me however the conflicting spinners in the original ticket are located in the import running starting here: https://github.com/ActiveState/cli/blob/version/0-46-0-RC1/internal/runners/packages/import.go#L138

I didn't realize that we also had this in the requirement operation function so it's good you caught that but we should also address the spinners in the import runner. This will likely affect the integration tests as the are relying on a `Done` message.

MDrakos (dismissed):

PR #3460, opened 2024-08-22
First attempt using intel/cve-bin-tool to scan for CVEs during build.

mitchell-as:
<table>
<td>
  </td></table>
  <br />
 

The tool found CVEs in our sqlite implementation, so I updated it. However, it's a ~3.5M line diff :cry:


mitchell-as:
Closing, as we're going to use trivy. This scanner has false-positives and takes 6 minutes to run.

PR #3461, opened 2024-08-26
First attempt using aquasecurity/trivy-action to scan for CVEs during build.

mitchell-as:
<table>
<td>
  </td></table>
  <br />
 

Remediate go-retryablehttp and gqlparser CVEs.


.github/workflows/build.yml
```diff
@@ -214,6 +214,17 @@ jobs:
         shell: bash
         run: parallelize results Build-Executor
 
+      - # === Scan for CVEs (Linux only) ===
+        name: Scan for CVEs
+        if: runner.os == 'Linux'
```
Naatan, line 219: We will need to scan for all platforms. It's ok if this runs ON linux, so long as we verify all binaries for all platforms.

Additionally, we should only run this on release branches and the nightly I think, no point slowing down regular PRs.

We could add a new job that runs in between build.yml and release.yml. Because we are already transferring all platform bits from build to release.yml, so we could inject something in the middle there for CVE validation.

.github/workflows/build.yml
```diff
@@ -424,7 +435,7 @@ jobs:
         name: Install Go
         uses: actions/setup-go@v3
         with:
-          go-version: ${{ matrix.go-version }}
+          go-version: 1.22.x
```
Naatan, line 438: Should update the matrix version instead.

Naatan (changes requested):

.github/workflows/build.yml
mitchell-as: The matrix does not exist for this job. Our action linter caught this.

.github/workflows/build.yml
MDrakos: Well, that's what I was going to say...

For additional information you can run `go version -m <executable> ` on a Go binary to see that there are differences in the modules used. It doesn't appear that versions are different but the Windows build may include modules that Linux does not and vice versa.

It also appears that Trivy has releases for MacOS and Windows, if that helps: https://github.com/aquasecurity/trivy/releases

mitchell-as: Very well, I've added a new Scan action inbetween build and deploy that scans all platform-specific binaries in one go. I disagree that it should only run on release branches and nightlies. PRs should not introduce CVEs only for them to be caught later. We should be proactive.

Naatan: @MDrakos did I hijack your review? 😅 Not my intention. When I reviewed this I was under the impression I was flagged, maybe I got my wires crossed.

.github/workflows/build.yml
```diff
@@ -397,11 +397,32 @@ jobs:
...
+        with:
+          path: build/
+
+      - name: Scan for CVEs
+        if: runner.os == 'Linux'
+        uses: aquasecurity/trivy-action@0.20.0
+        with:
+          scan-type: rootfs
+          scan-ref: build
+          ignore-unfixed: true
```
Naatan, line 417:
```suggestion
          ignore-unfixed: false
```

If I understand this correctly `true` would only report it the first time.

.github/workflows/build.yml
```diff
@@ -397,11 +397,32 @@ jobs:
...
+    steps:
+      - name: Download All Build Session Artifacts
+        uses: actions/download-artifact@v2
+        with:
+          path: build/
+
+      - name: Scan for CVEs
+        if: runner.os == 'Linux'
+        uses: aquasecurity/trivy-action@0.20.0
+        with:
```
Naatan, line 414: Could you try adding `list-all-pkgs`? I'd like to see a report of the files that were scanned so we can assert that it did in fact scan the things we want it to.

.github/workflows/build.yml
```diff
@@ -397,11 +397,32 @@ jobs:
           name: session-build-${{ matrix.sys.os }}
           path: build/
 
+  scan:
```
Naatan, line 400: Please add the same `if` conditional as deploy. Which has it only run on nightly and release branches.

eg. 
```
if: contains(fromJSON('["refs/heads/master", "refs/heads/beta", "refs/heads/release", "refs/heads/LTS"]'), github.ref) || startsWith(github.event.pull_request.head.ref, 'version/')
```

Naatan (changes requested):

.github/workflows/build.yml
mitchell-as: No, this will ignore any CVEs that do not have fixes. We currently have an unfixed CVE. Without this option, the check will fail and we will have to force PRs to merge.

.github/workflows/build.yml
mitchell-as: You missed my comment about this:

> I disagree that it should only run on release branches and nightlies. PRs should not introduce CVEs only for them to be caught later. We should be proactive. It took a whopping 5s to scan all generated binaries.

Until that is addressed, I am not okay with doing what you requested.

.github/workflows/build.yml
Naatan: Ahh I see.. ok yeah that makes sense 👍

Naatan (approved):

PR #3462, opened 2024-08-26
Test CVE scanner on old branch.

mitchell-as:
<table>
<td>
  </td></table>
  <br />


mitchell-as:
Verified that trivy works for scanning binaries. It reports CVEs that have since been fixed. Closing.

PR #3465, opened 2024-08-27
Merge "PATH" with Window's "Path".

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3012" title="DX-3012" target="_blank">BugDX-3012</a>  Windows - `state exec` regression
  </td></table>
  <br />
 

This enables searching for exes in not just runtime PATH, but also Windows Path.

I believe this PR triggered the issue: https://github.com/ActiveState/cli/commit/d881640f32db1ffb3fb488a2b87f860cd6c82374, but I think it exposed the fact we aren't merging `PATH` and `Path` on Windows.

### DIFF
--- a/internal/runners/exec/exec.go
+++ b/internal/runners/exec/exec.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -31,7 +32,7 @@ import (
 	"github.com/ActiveState/cli/pkg/platform/model"
 	"github.com/ActiveState/cli/pkg/project"
 	"github.com/ActiveState/cli/pkg/projectfile"
-	"github.com/ActiveState/cli/pkg/runtime"
+	rt "github.com/ActiveState/cli/pkg/runtime"
 )
 
 type Configurable interface {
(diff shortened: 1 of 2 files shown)


mitchell-as:
I believe this PR triggered the issue: https://github.com/ActiveState/cli/commit/d881640f32db1ffb3fb488a2b87f860cd6c82374, but I think it exposed the fact we aren't merging `PATH` and `Path` on Windows.

mitchell-as:

Naatan (changes requested):
I may lack context here, but intuitively this feels wrong to me. If Windows is case insensitive then setting it to the sensitivity windows "assumes" doesn't solve anything. It sounds like the issue stems from us treating PATH as case sensitive on Windows, when it is not.

I think we need to make a larger change that has us always use `Path` on Windows for consistency, but also has us respecting case insensitive PATH matches.

It seems that this was always "broken" and we're just now realizing, right? So perhaps fixing the bug at hand at filing a story for a larger refactor is appropriate here. ie. your proposed solution for updating JUST `state exec`.

mitchell-as:
Okay, I've worked around this and filed DX-3030 in response.

mitchell-as:

Naatan (approved):

PR #3466, opened 2024-08-27
Update shell detection

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3006" title="DX-3006" target="_blank">TaskDX-3006</a>  Improve SHELL detection; base on parent process
  </td></table>
  <br />


MDrakos:
@Naatan I'm looking for feedback on how I addressed the integration test failures. I think the comment explain it but if there are any questions let me know. With this approach or a different one I will have to go through all of the integration tests and update them.

Of course feel free to review everything. I'm just asking for a review early before I address the remaining tests.

Naatan (changes requested):
This looks ok to me, but I don't think we need to make such sweeping changes. The only place this ought to matter is where we specifically test shell detection, which I imagine should be only one test.

I imagine what you're running into is that other tests have started to break because we shifted shell detection to be process based? If that's the case I think a more general purpose solution is appropriate, eg. add an env var that will trump shell detection that we can set on CI (`ACTIVESTATE_OVERRIDE_SHELL`?).

I'd prefer that over updating all these tests and the e2e code, as that feels less maintainable.

internal/subshell/subshell.go
```diff
@@ -182,20 +183,24 @@ func DetectShell(cfg sscommon.Configurable) (string, string) {
...
-			binary = "cmd.exe"
-		} else {
-			binary = "bash"
-		}
+		binary = os.Getenv(SHELL_ENV_VAR)
+	}
+
+	if binary == "" {
+		binary = OS_DEFULAT
 	}
```
Naatan, line 204: Can we change this into a switch case? We don't need to assert `binary == nil`, instead assert whether we have something to populate it with.

internal/subshell/subshell.go
```diff
@@ -182,20 +183,24 @@ func DetectShell(cfg sscommon.Configurable) (string, string) {
...
+		binary = os.Getenv(constants.OverrideShellEnvVarName)
+	}
+
+	if binary == "" {
+		binary = detectShellParent()
 	}
 
 	if binary == "" {
 		binary = configured
 	}
```
Naatan, line 196: This should trump detection.

internal/subshell/subshell.go
```diff
@@ -239,24 +244,21 @@ func DetectShell(cfg sscommon.Configurable) (string, string) {
...
 
-func detectShellWindows() string {
-	// Windows does not provide a way of identifying which shell we are running in, so we have to look at the parent
-	// process.
-
+func detectShellParent() string {
 	p, err := process.NewProcess(int32(os.Getppid()))
 	if err != nil && !errors.As(err, ptr.To(&os.PathError{})) {
-		panic(err)
+		logging.Error("Failed to get parent process: %v", err)
```
Naatan, line 250: Not sure why this was a panic, but we went from panicking to essentially ignoring it. Please return the error up the chain instead.

internal/subshell/subshell_lin_mac.go
```diff
@@ -18,3 +18,17 @@ var supportedShells = []SubShell{
 	&fish.SubShell{},
 	&cmd.SubShell{},
 }
+
+const (
+	SHELL_ENV_VAR = "SHELL"
+	OS_DEFULAT    = "bash"
```
Naatan, line 24:
```suggestion
	OS_DEFAULT    = "bash"
```

internal/subshell/subshell_win.go
```diff
@@ -3,8 +3,30 @@
...
 
 var supportedShells = []SubShell{
 	&cmd.SubShell{},
+	&pwsh.SubShell{},
+	&bash.SubShell{},
+}
+
+const (
+	SHELL_ENV_VAR = "COMSPEC"
+	OS_DEFULAT    = "cmd.exe"
```
Naatan, line 22:
```suggestion
	OS_DEFAULT    = "cmd.exe"
```

internal/subshell/subshell_lin_mac.go
```diff
@@ -18,3 +18,17 @@ var supportedShells = []SubShell{
...
+)
+
+func supportedShellName(name string) bool {
+	for _, subshell := range supportedShells {
+		if name == subshell.Shell() {
```
Naatan, line 29: Should probably make this case-insensitive.

internal/subshell/subshell.go
```diff
@@ -239,24 +244,21 @@ func DetectShell(cfg sscommon.Configurable) (string, string) {
...
 	}
 
-	for p != nil {
+	for p != nil && p.Pid != 0 {
 		name, err := p.Name()
 		if err == nil {
-			if strings.Contains(name, "cmd.exe") || strings.Contains(name, "powershell.exe") {
+			if supportedShellName(name) {
```
Naatan, line 256: I believe this can be an absolute path, so not guaranteed to just be the filename.

internal/subshell/subshell_win.go
```diff
@@ -3,8 +3,30 @@
...
+)
+
+func supportedShellName(name string) bool {
+	for _, subshell := range supportedShells {
+		if name == fmt.Sprintf("%s.exe", subshell.Shell()) {
```
Naatan, line 27: Suggest renaming `name` to `filename` to make it explicit that this is dealing in shell filenames and not shell names. Also please make the assertion case insensitive (I seem to recall windows sometimes refers to it as `CMD.EXE`, and windows is case-insensitive).

internal/testhelpers/e2e/session.go
```diff
@@ -189,6 +189,12 @@ func new(t *testing.T, retainDirs, updatePath bool, extraEnv ...string) *Session
 		require.NoError(session.T, err)
 	}
 
+	if runtime.GOOS == "windows" {
+		if err := cfg.Set(subshell.ConfigKeyShell, "cmd.exe"); err != nil {
+			require.NoError(session.T, err)
+		}
+	}
```
Naatan, line 196: Isn't this redundant with the env change? Also pretty sure the `cfg` instance you're operating on here isn't the one that the invoked state tool would be using, unless they happen to be the same version.

internal/testhelpers/e2e/spawn.go
```diff
@@ -74,7 +74,7 @@ func (s *SpawnedCmd) ExpectInput(opts ...termtest.SetExpectOpt) error {
 	expect := `expect'input from posix shell`
 	if cmdName != "bash" && shellName != "bash" && runtime.GOOS == "windows" {
 		if strings.Contains(cmdName, "powershell") || strings.Contains(shellName, "powershell") {
-			send = "echo \"`<expect input from powershell`>\""
```
Naatan, line 77: Are you sure this works? IIRC the backtick is an escape on powershell.

Naatan (changes requested):

internal/subshell/subshell.go
MDrakos: We can however this leads to us calling functions like `Getenv` or `detectShellParent()` twice. We could also move the function calls outside of the switch but then there will be cases where we are calling them needlessly. 

I'm not sure how much of a performance impact `detectShellParent()` has but we would be adding inefficiency for a marginal gain in readability. I'm fine with whatever direction we want to go but I wanted to flag this before I make the change.

internal/subshell/subshell.go
MDrakos: This was originally a windows only function which may be a reason for the panic?

No other functions in this package return errors either as `DetectShell` has fallbacks so it always returns a shell name. This function is also used but `subshell.New`. So even if we get a fallback shell from either the `SHELL` env var or the system default we would error out because we couldn't detect the parent.

Again, I just want to check before making this change.

internal/subshell/subshell.go
MDrakos: Oh... that was not how I read the ACs.

internal/testhelpers/e2e/spawn.go
MDrakos: You were right, I was running into a failure with this because it was defaulting back to `cmd.exe` and not powershell.

internal/subshell/subshell.go
```diff
@@ -188,19 +189,19 @@ func DetectShell(cfg sscommon.Configurable) (string, string) {
 	}
 
 	if binary == "" {
-		binary = detectShellParent()
+		binary = configured
```
Naatan, line 192: Sorry I realized upon reading the changes that configured had a different connotation than I remembered. In my mind this was "what the user configured", but actually this is just a fallback we ourselves set when we detect the shell.

This is awkward and should probably be revisited, but given the current behaviour I this should probably be a fallback after shell parent and environment.

Naatan (changes requested):

internal/testhelpers/e2e/spawn.go
MDrakos: Never mind, I was right. Here's the output on powershell

```
[ActiveState-CLI/small-python] C:\Users\runneradmin\AppData\Local\Temp\2179854665\work\small-python>echo "`<expect input from powershell`>" 
"`<expect input from powershell`>"    
```

And the expect

```
'Expected: "<expect input from powershell>": after 40s: timeout':
              'after 40s: timeout': timeout
```

Naatan: My guess is it depends on which shell you started from. So your new tests may want this variant but older tests will want the escaping? Worth perhaps looking for other places where we run ExpectInput and running those tests on Windows before merging your PR.

Naatan (approved):
Looks good, but please read the comment on ExpectInput.

internal/testhelpers/e2e/spawn.go
MDrakos: All of the integration tests that start powershell are also critical. Only one of them expects input which is the one that flagged this change.

PR #3471, opened 2024-09-03
Add timeout for complex import test.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3026" title="DX-3026" target="_blank">BugDX-3026</a>  Nightly failure: TestImportIntegrationTestSuite/TestImport/complex_requirements.txt
  </td></table>
  <br />
 

The default timeout is too low for this test.

### DIFF
--- a/test/integration/import_int_test.go
+++ b/test/integration/import_int_test.go
@@ -6,6 +6,9 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


mitchell-as:
I tried 10s, 20s, and then 30s before it passed on all platforms.

MDrakos (approved):

PR #3477, opened 2024-09-04
Do not re-link or re-copy a deployed file if it still exists.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3037" title="DX-3037" target="_blank">BugDX-3037</a>  INIT command fail with CRITICAL error
  </td></table>
  <br />

### DIFF
--- a/pkg/runtime/depot.go
+++ b/pkg/runtime/depot.go
@@ -332,9 +332,6 @@ func (d *depot) getSharedFilesToRedeploy(id strfmt.UUID, deploy deployment, path
 	// For each file deployed by this artifact, find another artifact (if any) that deploys its own copy.
 	for _, relativeDeployedFile := range deploy.Files {
 		deployedFile := filepath.Join(path, relativeDeployedFile)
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
This feels too vague, we uninstalled the artifact and so the files it owns should be gone. We're finding that this is not the case, which suggests there's an issue somewhere either with removing the files or with tracking the installed files. We need to address this issue at the root.

I can't recall; are we tracking which artifact owns the file? Or are we just tracking all files an artifact cares about regardless of whether they are the owner? We need to be doing the former.

mitchell-as:
I agree, it didn't feel right. After Mark gave me the ability to reproduce it, I found that we're considering artifact deployments outside the current one. Oops. We should only consider artifact deployments targetting the current one for re-linking and re-copying.

We only see this when there are multiple deployments on a given machine. We aren't seeing this on CI because we use a clean cache, etc. for each test.

Naatan (approved):

PR #3480, opened 2024-09-06
Added marshaling and pre-unmarshaling hooks for buildscript <-> buildexpression.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3031" title="DX-3031" target="_blank">TaskDX-3031</a>  Our buildscript package can mediate evaluation of client-side functions
  </td></table>
  <br />
 

This allows for outside packages to hook into:
- Unmarshaling from buildexpression to buildscript
- Marshaling from buildscript to buildexpression

Also: utilize this when marshaling `Req()` function to legacy object format. We cannot utilize the unmarshaling yet.

### DIFF
--- a/pkg/buildscript/marshal_buildexpression.go
+++ b/pkg/buildscript/marshal_buildexpression.go
@@ -21,6 +21,21 @@ const (
 	requirementComparatorKey          = "comparator"
 )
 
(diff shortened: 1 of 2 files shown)


Naatan (approved):
Nice, much simpler than I was expecting!

PR #3483, opened 2024-09-09

mitchell-as:
<table>
<td>
  </td></table>
  <br />
 

- Use triple-backticks to separate checkout info from buildscript. Using triple-quotes will require a new parser and/or lexer. Since we're experimenting, let's use what we have.
- Buildscripts always have an AtTime (it's no longer optional). I believe we had it optional for compatiblity with build expressions that didn't have a separate commit time, but all build expressions now have an associated commit time from what I can tell.
- Renamed `localcommit` package to `checkoutinfo`
  - `checkoutinfo` is a primer property and all runners go through it to get commit IDs and work with build script (initialize, update, etc.).
  - Removed `runbits/buildscript/file.go` in favor of this package.
- Changed `buildplanner.GetBuildScript()` to additionally accept an owner, project, branch, and commitID. It uses these fields to construct the "Project" field in the commit info section of buildscripts.
- `state reset LOCAL` will always re-initialize a build script.
- Internal change in the raw buildscript format to not store strings with quotes. (Unmarshaling strips quotes, and marshaling adds them back.)
- Users are shown a warning if they attempt to use an outdate build script (`state install`, etc.).


pkg/projectfile/projectfile.go
```diff
@@ -965,18 +965,13 @@ func createCustom(params *CreateParams, lang language.Language) (*Project, error
...
+		pjf.LoadProject("https://" + host)
+		pjf.SetNamespace(params.Owner, params.Project)
 		if params.BranchName != "" {
-			q.Set("branch", params.BranchName)
+			pjf.SetBranch(params.BranchName)
 		}
-
-		u.RawQuery = q.Encode()
-		params.ProjectURL = u.String()
+		params.ProjectURL = pjf.String()
```
mitchell-as, line 974: This is to avoid code duplication. I was using `NewProjectField()` in the buildplanner to construct the "Project" field in the checkout info block of buildscripts, and found that we can do the same thing here.

mitchell-as:
Test failure is unrelated to this PR.

internal/runbits/buildscript/file.go
```diff
@@ -40,8 +40,8 @@ func ScriptFromFile(path string) (*buildscript.BuildScript, error) {
 	return buildscript.Unmarshal(data)
 }
 
-func Initialize(path string, auth *authentication.Auth) error {
-	scriptPath := filepath.Join(path, constants.BuildScriptFileName)
+func Initialize(proj projecter, auth *authentication.Auth) error {
```
Naatan, line 43: It doesn't really make sense to me that this takes a project as an input. This story is effectively about introducing the project information to the buildscript, with the intend to eventually have this be the sole source of truth. For now we will also have this in the activestate.yaml, which is what the project package handles.

In order for this to not generate throwaway code or code that will need significant refactoring later, we would need to keep the buildscript and project package unaware of each other. And in order to expedite this we could use a mediator to handle mediating between the two (this is what I imagined `checkoutinfo` would be used for).

Passing in the project details as retrieved from the project package might be ok though. Just as long as buildscripts has zero awareness of the project package and vice versa. 

Basically think of how we want to use this package once its the source of truth, and write it that way. Then utilize the mediator to bridge the gap for now.

Naatan (changes requested):
Please see comment on the Initialize function. I didn't finish the review as I imagine this comment may bring about some significant changes.

internal/runbits/buildscript/file.go
mitchell-as: Fair enough. Originally I added new owner and project string fields, but later realized the projector interface was available and provided that information, so I cleaned it up to do that. I've gone back to passing owner and project arguments again.

internal/runners/reset/reset.go
```diff
@@ -126,16 +127,26 @@ func (r *Reset) Run(params *Params) error {
...
 	if err != nil {
 		return errs.Wrap(err, "Unable to set local commit")
 	}
 
 	// Ensure the buildscript exists. Normally we should never do this, but reset is used for resetting from a corrupted
 	// state, so it is appropriate.
 	if r.cfg.GetBool(constants.OptinBuildscriptsConfig) {
-		if err := buildscript_runbit.Initialize(r.project.Dir(), r.auth); err != nil {
-			return errs.Wrap(err, "Unable to initialize buildscript")
+		err := buildscript_runbit.Initialize(r.project.Dir(), r.project.Owner(), r.project.Name(), r.auth)
```
Naatan, line 138: You should also have the branch. The project URL we show in the buildscript should be identical to the one in the activestate.yaml.

internal/runners/reset/reset.go
```diff
@@ -126,16 +127,26 @@ func (r *Reset) Run(params *Params) error {
...
+			if errors.Is(err, buildscript.ErrOutdatedAtTime) {
+				// Remove the outdated build script and try again.
+				if err := buildscript_runbit.Remove(r.project.Dir()); err != nil {
+					return errs.Wrap(err, "Unable to remove outdated build script")
+				}
+				err = buildscript_runbit.Initialize(r.project.Dir(), r.project.Owner(), r.project.Name(), r.auth)
+			}
+			if err != nil {
+				return errs.Wrap(err, "Unable to initialize buildscript")
+			}
```
Naatan, line 149: This feels more complex than it needs to be. This is a reset, so we should just always remove the file if it exists and then initialize. Initialize should not have a care in the world about the buildscript already existing beyond obviously erroring out if it exists. Point being it's "Initialize" not "InitializeOrUpdate".

pkg/buildscript/unmarshal.go
```diff
@@ -31,23 +35,28 @@ func Unmarshal(data []byte) (*BuildScript, error) {
...
-			break
-		}
-		atTime, err := strfmt.ParseDateTime(strValue(value))
-		if err != nil {
-			return nil, errs.Wrap(err, "Invalid timestamp: %s", strValue(value))
+		return nil, ErrOutdatedAtTime
+	}
+
+	if raw.Info != nil {
+		for _, matches := range checkoutInfoPairRegex.FindAllStringSubmatch(*raw.Info, -1) {
```
Naatan, line 47: Suggest instead using a yaml parser for this. The checkout info is essentially yaml.

pkg/buildscript/unmarshal_buildexpression.go
```diff
@@ -71,20 +70,20 @@ func UnmarshalBuildExpression(data []byte, atTime *time.Time) (*BuildScript, err
 
 	// Extract the 'at_time' from the solve node, if it exists, and change its value to be a
 	// reference to "$at_time", which is how we want to show it in AScript format.
-	if atTimeNode, err := script.getSolveAtTimeValue(); err == nil && atTimeNode.Str != nil && !strings.HasPrefix(strValue(atTimeNode), `$`) {
-		atTime, err := strfmt.ParseDateTime(strValue(atTimeNode))
+	if atTimeNode, err := script.getSolveAtTimeValue(); err == nil && atTimeNode.Str != nil && !strings.HasPrefix(*atTimeNode.Str, `$`) {
```
Naatan, line 73: I think this could throw a panic if the user entered a non-string value for the at time node?

pkg/buildscript/unmarshal_buildexpression.go
```diff
@@ -390,7 +389,7 @@ func transformVersion(requirements *Assignment) *FuncCall {
 					{Assignment: &Assignment{"value", o.Value}},
 				}
 			case requirementComparatorKey:
-				f.Name = cases.Title(language.English).String(strValue(o.Value))
+				f.Name = cases.Title(language.English).String(*o.Value.Str)
```
Naatan, line 392: Potential nil panic if user entered different value type.

pkg/checkoutinfo/checkoutinfo.go
```diff
@@ -0,0 +1,114 @@
+package checkoutinfo
```
Naatan, line 1: I'd expect to see project info mediated here that is reflected in the URL. So org, name, namespace and branch.

Basically those fields should all be handled through this package, and the project/projectfile package can still work with them through the legacy pattern, same as we're doing with the commitID.

Naatan (changes requested):

pkg/buildscript/unmarshal_buildexpression.go
mitchell-as: No, because `atTimeNode.Str != nil` was checked earlier. If it's `nil`, we throw an error on line 82.

pkg/buildscript/unmarshal_buildexpression.go
mitchell-as: No, this is coming from the Platform. We are translating a platform buildexpression into a buildscript, so the user hasn't been involved yet.

internal/runners/reset/reset.go
mitchell-as: Good call, but that doubled the changeset size :/

pkg/checkoutinfo/checkoutinfo.go
mitchell-as: Right. I was hoping we could eliminate `checkoutinfo.SetCommitID()`, but buildscripts are still hidden behind a config wall. Until we remove that wall, we'll have the SetCommitID -> update activestate.yaml, update build script -> update activestate.yaml flow.

mitchell-as:
I expect more churn.

internal/runbits/buildscript/commitid.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/go-openapi/strfmt"
+
+	"github.com/ActiveState/cli/internal/constants"
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/pkg/buildscript"
+	"github.com/ActiveState/cli/pkg/checkoutinfo"
+)
+
+func CommitID(path string, cfg configurer) (strfmt.UUID, error) {
```
Naatan, line 14: Can we move this to checkoutinfo? I get where you're coming from, but runbits are about sharing controller logic. This isn't controller logic. And this type of muliplexing is exactly what the checkoutinfo package is intended for.

Naatan (changes requested):
Two things stand out to me at a high level:

- I think checkoutinfo should probably be a primer property
	- Because the reliance on cfg already implies primer anyway, and similar to project/projectfile this will almost always be relevant for every command.
- We're still sourcing project URL info from the project package
  - eg. owner, name and branch are still being sourced from `project`. I'd expect to see these sourced from `checkoutinfo`, and have legacy methods on the project methods like we did with the commitID. These legacy methods should probably only be called from checkoutinfo, but there might be edge cases.

mitchell-as:
I made checkout info a primer property and I think I addressed your other line item (sorry if I overlooked it; lots of changes).

I've updated the PR description with an overview of the entire thing.

pkg/checkoutinfo/checkoutinfo.go
```diff
@@ -0,0 +1,229 @@
...
+		if script, err := c.BuildScript(); err == nil {
+			commitID, err2 := script.CommitID()
+			if err2 != nil {
+				return "", errs.Wrap(err, "Could not get commit ID from build script")
+			}
+			return commitID, nil
+		} else if !errors.Is(err, ErrBuildscriptNotExist) {
+			return "", errs.Wrap(err, "Could not get build script")
+		}
+		// Fall back on activestate.yaml.
```
Naatan, line 76: If the buildscript exists then I think we should just fail here. Falling back to the as.yaml at that point is working around some bug.

pkg/checkoutinfo/checkoutinfo.go
```diff
@@ -0,0 +1,229 @@
...
+func (c *CheckoutInfo) Owner() string {
+	return c.project.Owner()
+}
+
+func (c *CheckoutInfo) Name() string {
+	return c.project.Name()
+}
+
+func (c *CheckoutInfo) Branch() string {
+	return c.project.BranchName()
```
Naatan, line 62: These should behave the same as the `CommitID()` function; ie. collect from buildscript if opted in.

pkg/checkoutinfo/checkoutinfo.go
```diff
@@ -0,0 +1,229 @@
...
+		return nil // buildscripts are not enabled, so nothing more to do
+	}
+
+	// Update commitID in Project field of build script.
+	logging.Debug("Updating commitID in buildscript")
+	buildscriptPath := filepath.Join(c.project.Dir(), constants.BuildScriptFileName)
+
+	if !fileutils.FileExists(buildscriptPath) {
+		return c.InitializeBuildScript(commitID)
+	}
```
Naatan, line 133: Please error out with the "run state reset LOCAL" error. This file should exist if opted in, it not existing is an error condition. Working around this magically like this is what we did originally and it caused a lot of issues.

pkg/checkoutinfo/checkoutinfo.go
```diff
@@ -0,0 +1,229 @@
...
+	// Read from activestate.yaml.
+	commitID := c.project.LegacyCommitID()
+	if !strfmt.IsUUID(commitID) {
+		return "", &ErrInvalidCommitID{commitID}
+	}
+
+	return strfmt.UUID(commitID), nil
+}
+
+func (c *CheckoutInfo) BuildScript() (*buildscript.BuildScript, error) {
```
Naatan, line 88: I'm not sure about having this, `UpdateBuildScript()`, and `InitBuildScript()` inside `checkoutinfo`. Checkoutinfo is from a consumer point of view about project information, not about buildscripts, that's just an under the hood mechanic. 

Also, we don't handle project retrieval or initialization in this package, so handling buildscripts here feels inconsistent.

Let's put it this way: if we KNOW we want a buildscript, or a project/projectile (the pkg) then we should not be using `checkoutinfo`.

pkg/checkoutinfo/checkoutinfo.go
Naatan: Also their `Set()` methods should be in `checkoutinfo` also.. otherwise we're not syncing the buildscript when those are changed.

Naatan (changes requested):

pkg/checkoutinfo/checkoutinfo.go
mitchell-as: I'm going to remove the `Owner()`, `Name()`, and `Branch()` accessors since they're not used externally. The ACs make no mention of having checkoutinfo being the gatekeeper of accessors, just for setters like you mentioned here.

If we make checkoutinfo the sole source of truth for Owner, Name, and Branch, then this PR will double in size. I'd rather do that in another ticket.

mitchell-as:
This is a mess. I'm going to restart.

mitchell-as:
Restarting is more of a mess, so this'll have to do.

mitchell-as:
No, I can't do it in this PR. I'll start from scratch.

PR #3484, opened 2024-09-09
Add hashGlobs query to state service

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3032" title="DX-3032" target="_blank">TaskDX-3032</a>  We have a query on the state-svc for calculating a hash for a series of globs
  </td></table>
  <br />


MDrakos:
When I ran `state run generate-graph` the generated code was using types that weren't present in our dependencies. I've update the `gqlgen` dependency to match that of the `gqlgen` tool that is installed by `generate-graph`. That is why there are so many vendor changes.

internal/hash/file_hasher.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"encoding/base64"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"time"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/cespare/xxhash"
```
MDrakos, line 12: xxhash is a fast hash algorithm. For more info see here: https://xxhash.com/
<sub>this link is safe for work</sub>

Naatan (changes requested):

.github/workflows/build.yml
```diff
@@ -174,7 +174,7 @@ jobs:
         name: Check Format
         id: check_format
         shell: bash
-        if: "!contains(fromJSON('[\"refs/heads/beta\", \"refs/heads/release\", \"refs/heads/LTS\", \"refs/heads/master\"]'), github.ref) && !startsWith(github.event.pull_request.head.ref, 'version/')"
+        if: '!contains(fromJSON(''["refs/heads/beta", "refs/heads/release", "refs/heads/LTS", "refs/heads/master"]''), github.ref) && !startsWith(github.event.pull_request.head.ref, ''version/'')'
```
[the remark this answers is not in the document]
MDrakos: Whoops, that's my YAML auto-formatter again...

internal/hash/file_hasher.go
```diff
@@ -0,0 +1,75 @@
...
+	Set(key string, value interface{}, expiration time.Duration)
+}
+
+type FileHasher struct {
+	cache fileCache
+}
+
+func NewFileHasher() *FileHasher {
+	return &FileHasher{
+		cache: cache.New(cache.NoExpiration, cache.NoExpiration),
```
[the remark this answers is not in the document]
MDrakos: It doesn't appear that limiting the cache size is supported by this library. Rather than implement it ourselves I just opted for the 24hr clean up.

internal/hash/file_hasher.go
```diff
@@ -0,0 +1,75 @@
...
+
+			hash = fmt.Sprintf("%x", fileHasher.Sum(nil))
+		}
+
+		if err := file.Close(); err != nil {
+			return "", errs.Wrap(err, "Could not close file: %s", f)
+		}
+
+		fh.cache.Set(cacheKey(file.Name(), fileInfo.ModTime()), hash, cache.NoExpiration)
+		fmt.Fprintf(hasher, "%x", hash)
```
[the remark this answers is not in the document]
MDrakos: Added a comment the hasher is an `io.Writer` so this is writing the hash of the current file to the overall hash.

internal/hash/file_hasher.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return base64.StdEncoding.EncodeToString(hasher.Sum(nil)), nil
```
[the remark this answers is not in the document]
MDrakos: `hasher.Sum()` returns a slice of bytes and I don't believe these values are guaranteed to be printable characters. They could be control characters or invalid UFT-8 sequences. The encoding ensures that the output is safe to print and transmit.

cmd/state-svc/internal/hash/file_hasher.go
```diff
@@ -24,11 +25,11 @@ type FileHasher struct {
 
 func NewFileHasher() *FileHasher {
 	return &FileHasher{
-		cache: cache.New(cache.NoExpiration, cache.NoExpiration),
+		cache: cache.New(cache.NoExpiration, 24*time.Hour),
```
Naatan, line 28:
```suggestion
		cache: cache.New(24*time.Hour, 24*time.Hour),
```

Otherwise the schedule would run every 24hr, but it would then find there is no expiration so it wouldn't actually expire anything.

Why these are separate inputs is beyond me.

cmd/state-svc/internal/hash/file_hasher.go
```diff
@@ -24,11 +25,11 @@ type FileHasher struct {
...
 	}
 }
 
-func (fh *FileHasher) HashFiles(files []string) (string, error) {
+func (fh *FileHasher) HashFiles(files []string) (hash string, rerr error) {
```
Naatan, line 32:
```suggestion
func (fh *FileHasher) HashFiles(files []string) (string, error) {
```

Doesn't look like naming these is used for anything. And it can actually cause bug so better to avoid.

Naatan (changes requested):

cmd/state-svc/internal/hash/file_hasher.go
MDrakos: Forgot that we need the named error return for the deferred error handling https://github.com/ActiveState/cli/pull/3484/files#diff-f95330cc381d368f14952fbad1b24b3a714997bab5851fba341014a2271774d0R41

Naatan: ah I see. In that case I suggest renaming `hash` to `_`, so it doesn't clash with your code. You are using a `hash` var in the function body. I don't think it clashes as currently written, but it feels error-prone.

Naatan (changes requested):

Naatan (approved):

PR #3485, opened 2024-09-10
Sort artifact changesets by name.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3028" title="DX-3028" target="_blank">StoryDX-3028</a>  Requirement ordering is consistent
  </td></table>
  <br />
 

Unfortunately, it's not as simple as sorting buildplan artifacts once and you're done. For example, the `DiffArtifacts()` function creates a map of artifacts, which is inherently unsorted, so we have to sort it after the fact. (It doesn't matter whether or not the incoming artifacts are sorted.) Other functions throughout the codebase have the capability to do something similar (if they do not already). We could either:

1. Identify and rewrite all functions that utilize artifacts, ingredients, and requirements to be mindful of sort order, and keep this in mind for all future functions we write.
2. Sort as needed, attempting to capture the largest number of use cases at as high a level as possible.

This PR chooses option 2, which sorts the `DiffArtifacts()` output. This results in CVE lists, change summary lists, and `state upgrade` lists outputting sorted lists of artifacts.

### DIFF
--- a/pkg/buildplan/buildplan.go
+++ b/pkg/buildplan/buildplan.go
@@ -2,6 +2,7 @@ package buildplan
 
 import (
 	"encoding/json"
+	"sort"
 
 	"github.com/go-openapi/strfmt"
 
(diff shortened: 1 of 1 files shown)


Naatan (approved):
I like your solution. Handling this at a low level and accepting there might still be edge cases was basically the approach of the story anyway.

PR #3486, opened 2024-09-10
Update deprecated GitHub Actions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3043" title="DX-3043" target="_blank">TaskDX-3043</a>  GitHub Actions has deprecated 'actions/upload-artifact: v2'
  </td></table>
  <br />

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,7 +68,7 @@ jobs:
 
       - # === Install Go ===
         name: Install Go
-        uses: actions/setup-go@v3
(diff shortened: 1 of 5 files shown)


mitchell-as:
We cannot avoid the scan failure, because the fix version is Go 1.22.7, and that version is not yet available with the setup-go action (1.22.6 is the latest).

This should clear up over time once the action can install 1.22.7, and future builds of the State Tool will get it automatically.

MDrakos (approved):

PR #3487, opened 2024-09-10
Refactored requirement operation and added multi-target/multi-language compatibility

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-1897" title="DX-1897" target="_blank">TaskDX-1897</a>  Refactor ExecuteRequirementOperation
  </td></table>
  <br />
 

A lot went into this one. More than I had anticipated, but less than I would've liked. 

Overall this focusses on trying to simplify and streamline the logic for installing and uninstalling packages. Similar to Mitchell's original PR this separates the install and uninstall actions. It also separates platform installs/uninstalls from requirement operations altogether, since platforms are not requirements. Finally, I turned this into a top level runner rather than a runbit, because there is no need for command centric runners as all the behaviour is already covered.

Summary of side-effects:

*   Multiple install targets:
    *   We no longer print the actual package names we're searching for, instead we just say a generic "searching for packages"
        *   Reasoning: Depending on how input is passed not all packages are actually searched for. 
    *   The `--revision` flag has been dropped.
        *   Reasoning: this flag is targeting a single package operation, but `state install` now supports multiple operations. The only way to specify a revision would be to use buildscripts.
*   Multiple languages:
    *   We can no longer normalize names, as this requires a namespace and we cannot determine the namespace without exact matching.
        *   Means users have to give the exact ingredient name, casing and all.
*   Misc:
    *   Added an `--expand` flag to `state manifest` which will show requirements by their full namespace.
        *   This is to give a user a tool to review namespaces in case they need to disambiguate their input for `state uninstall` (eg. multiple packages with the same name, but different namespace).
    * Commit messages and install/uninstall success messages have been updated. Eg. instead of "Package added: pkgName" it will now say: "Added: namespace/to/pkgName".
    	* Adding the namespace is intentional as the user will need to reason about them when working with multi-language projects or projects with private ingredients.


pkg/buildscript/mutations.go
```diff
@@ -33,7 +35,11 @@ func (b *BuildScript) UpdateRequirement(operation types.Operation, requirement t
 	return nil
 }
 
-func (b *BuildScript) addRequirement(requirement types.Requirement) error {
+func (b *BuildScript) AddRequirement(requirement types.Requirement) error {
```
mitchell-as, line 38: By changing the visibility of this method, it's possible that `UpdateRequirement()` is no longer needed. Please consider removing it.

pkg/buildscript/mutations.go
```diff
@@ -72,34 +78,44 @@ type RequirementNotFoundError struct {
 	*locale.LocalizedError // for legacy non-user-facing error usages
 }
 
-func (b *BuildScript) removeRequirement(requirement types.Requirement) error {
+// RemoveRequirement will remove any matching requirement. Note that it only operates on the Name and Namespace fields.
+// It will not verify if revision or version match.
+func (b *BuildScript) RemoveRequirement(requirement types.Requirement) error {
```
mitchell-as, line 83: By changing the visibility of this method, it's possible that `UpdateRequirement()` is no longer needed. Please consider removing it.

pkg/buildscript/mutations.go
```diff
@@ -125,7 +141,7 @@ func (b *BuildScript) UpdatePlatform(operation types.Operation, platformID strfm
 	return nil
 }
 
-func (b *BuildScript) addPlatform(platformID strfmt.UUID) error {
+func (b *BuildScript) AddPlatform(platformID strfmt.UUID) error {
```
mitchell-as, line 144: By changing the visibility of this method, it's possible that `UpdatePlatform()` is no longer needed. Please consider removing it.

pkg/buildscript/mutations.go
```diff
@@ -143,7 +159,7 @@ type PlatformNotFoundError struct {
 	*locale.LocalizedError // for legacy non-user-facing error usages
 }
 
-func (b *BuildScript) removePlatform(platformID strfmt.UUID) error {
+func (b *BuildScript) RemovePlatform(platformID strfmt.UUID) error {
```
mitchell-as, line 162: By changing the visibility of this method, it's possible that `UpdatePlatform()` is no longer needed. Please consider removing it.

pkg/platform/model/vcs.go
```diff
@@ -83,7 +84,7 @@ const (
 	NamespaceCamelFlagsMatch = `^camel-flags$`
 
 	// NamespaceOrgMatch is the namespace used for org specific requirements
-	NamespaceOrgMatch = `^org\/`
+	NamespaceOrgMatch = `^private\/`
```
mitchell-as, line 87: Good catch!

pkg/platform/model/vcs.go
```diff
@@ -169,6 +174,18 @@ func (n Namespace) String() string {
 	return n.value
 }
 
+func ParseNamespace(ns string) Namespace {
```
mitchell-as, line 177: Nice, I've been hoping for a function like this for a while now.

pkg/platform/model/vcs.go
```diff
@@ -204,21 +221,24 @@ func NewNamespacePlatform() Namespace {
 	return Namespace{NamespacePlatform, "platform"}
 }
 
-const OrgNamespacePrefix = "private/"
-
 func NewOrgNamespace(orgName string) Namespace {
```
mitchell-as, line 224: Nit: for consistency we should rename this `NewNamespaceOrg()` to match the other functions.

test/integration/install_int_test.go
```diff
@@ -25,10 +25,27 @@ func (suite *InstallIntegrationTestSuite) TestInstall() {
...
+
+	ts.PrepareProject("ActiveState-CLI/small-python", "5a1e49e5-8ceb-4a09-b605-ed334474855b")
+
+	cp := ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("install", "djang")
+	cp.Expect("No results found")
+	cp.Expect("Did you mean")
+	cp.Expect("language/python/djang")
```
mitchell-as, line 45: Please double-check you mean to expect "djang" instead of "django".

test/integration/package_int_test.go
```diff
@@ -277,122 +270,26 @@ func (suite *PackageIntegrationTestSuite) TestPackage_detached_operation() {
...
-
-	cp = ts.Spawn("history", "--output=json")
-	cp.ExpectExitCode(0)
-
 	suite.Run("install", func() {
 		cp := ts.Spawn("install", "requests", "urllib3@1.25.6")
-		cp.Expect(fmt.Sprintf("Operating on project %s/python3-pkgtest", user.Username))
-		cp.ExpectRe("(?:Package added|being built)", termtest.OptExpectTimeout(30*time.Second))
+		cp.Expect("Operating on project ActiveState-CLI/small-python")
+		cp.Expect("Added: language/python/requests", e2e.RuntimeSourcingTimeoutOpt)
```
mitchell-as, line 276: We have async runtimes enabled, so normally we don't need timeouts like this. Could you add a comment (perhaps top-level) explaining why we're using timeouts here? (Including the functions below.)

Based on the diff, it looks like we're expecting a long "resolving dependencies" window due to the Platform needing to build packages, but you wouldn't know that from looking at this code by itself.

test/integration/package_int_test.go
```diff
@@ -431,16 +328,46 @@ scripts:
...
+	cp = ts.Spawn("uninstall", "requests")
+	cp.Expect("project has been updated")
+	cp.ExpectExitCode(0)
+}
+
 func (suite *PackageIntegrationTestSuite) TestPackage_UninstallDoesNotExist() {
 	suite.OnlyRunForTags(tagsuite.Package)
 
-	ts := e2e.New(suite.T(), false)
+	ts := e2e.New(suite.T(), true)
```
mitchell-as, line 364: Did you mean to leave this change in place?

mitchell-as (changes requested):
At a high level this looks fine to me. Good job! Some small things to address before merging, though.

activestate.yaml
```diff
@@ -48,6 +48,7 @@ constants:
 scripts:
   - name: install-deps-dev
     language: bash
+    standalone: true
```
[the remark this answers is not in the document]
Naatan: Because I find often while testing that it re-sources my runtime because I invoked it with another state tool version (ie. the one I'm working on vs my installed state tool). And end of the day none of our scripts actually require the runtime, they were always standalone, just never marked as such.

internal/locale/locales/en-us.yaml
```diff
@@ -1145,23 +1083,25 @@ err_exec_recursion:
     To allow recursion, set [ACTIONABLE]{{.V1}}=true[/RESET]
 package_ingredient_alternatives:
   other: |
-    No results found for search term "[NOTICE]{{.V0}}[/RESET]". Did you mean:
+    No results found for search term {{.V0}}. Did you mean:
```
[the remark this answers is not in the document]
Naatan: It's still highlighted, but because it now takes multiple values it happens outside of the locale file:

https://github.com/ActiveState/cli/blob/02ba52766043b37a4fa9ebca874a7d8b9bf054ff/internal/runners/install/rationalize.go#L29-L29

internal/runbits/reqop_runbit/update.go
```diff
@@ -0,0 +1,143 @@
...
+	pg = nil
+
+	// Report changes and CVEs to user
+	dependencies.OutputChangeSummary(out, newCommit.BuildPlan(), oldCommit.BuildPlan())
+	if err := cves.NewCveReport(prime).Report(newCommit.BuildPlan(), oldCommit.BuildPlan()); err != nil {
+		return errs.Wrap(err, "Could not report CVEs")
+	}
+
+	// Start runtime sourcing UI
+	if !cfg.GetBool(constants.AsyncRuntimeConfig) {
```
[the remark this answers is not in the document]
Naatan: This was not intentional on my part, I think I copied this from the old code. But reading through it now I think it makes sense to retain it here. The comment in Update says that we still want the actual solving logic to run so we can raise errors. But that solving logic has already ran in this case when we created the commit. So solving it again in the async use-case wouldn't really give us any benefit.

I'll add a comment explaining as much.

internal/runners/install/install.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	// Prepare updated buildscript
+	script := oldCommit.BuildScript()
+	if err := prepareBuildScript(script, reqs, ts); err != nil {
+		return errs.Wrap(err, "Could not prepare build script")
+	}
+
+	// Update local checkout and source runtime changes
+	if err := reqop_runbit.UpdateAndReload(i.prime, script, oldCommit, locale.Tr("commit_message_added", reqs.String()), trigger.TriggerInstall); err != nil {
```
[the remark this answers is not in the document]
Naatan: We've had some runbits that became dumping grounds, hence why I wanted it a bit more specific. I agree `reqop` is a bit vague but I don't have any better ideas that are both specific and not overly verbose (eg. `requirementoperation_runbit`).

internal/runners/platforms/remove.go
```diff
@@ -1,59 +1,145 @@
...
+	bpModel "github.com/ActiveState/cli/pkg/platform/model/buildplanner"
 )
 
 // RemoveRunParams tracks the info required for running Remove.
 type RemoveRunParams struct {
 	Params
 }
 
-// Remove manages the removeing execution context.
+// Remove manages the adding execution context.
```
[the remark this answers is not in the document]
Naatan: Lol thanks for catching. This was definitely because of me copying the runner and doing a `s/remove/add`..

pkg/buildscript/mutations.go
Naatan: Yeah that was my expectation, but `state import` and buildscript merging still makes good use of it.

pkg/buildscript/mutations.go
Naatan: Oh this one I actually can remove! Thanks for catching.

pkg/platform/model/vcs.go
Naatan: 😅

test/integration/package_int_test.go
Naatan: This was addressing solve times, not sourcing times. You're totally right for flagging this though. I've introduced a new `RuntimeSolvingTimeoutOpt`.

test/integration/package_int_test.go
Naatan: Nope, great catch!

mitchell-as (changes requested):
Looks good, but there are some test failures to address. Then it's good to go!

mitchell-as (approved):

Naatan:
The only test failure appears to be a sporadic one unrelated to this PR. Gonna move on ahead.

PR #3488, opened 2024-09-11
Wrong libc should be an input error.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3041" title="DX-3041" target="_blank">BugDX-3041</a>  `CRT` error in the log for user input error while platform incompatible
  </td></table>
  <br />
 

Note: I had to cherry pick the GitHub Actions fixes made in #3486 because they are only present in v47.

### DIFF
--- a/internal/runbits/runtime/rationalize.go
+++ b/internal/runbits/runtime/rationalize.go
@@ -109,7 +109,8 @@ func RationalizeSolveError(proj *project.Project, auth *auth.Auth, rerr *error) 
 				*rerr = errs.NewUserFacing(locale.Tr(
 					"err_alternate_branches",
 					noMatchingPlatformErr.HostPlatform, noMatchingPlatformErr.HostArch,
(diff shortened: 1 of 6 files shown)


internal/runbits/runtime/rationalize.go
```diff
@@ -109,7 +109,8 @@ func RationalizeSolveError(proj *project.Project, auth *auth.Auth, rerr *error)
 				*rerr = errs.NewUserFacing(locale.Tr(
 					"err_alternate_branches",
 					noMatchingPlatformErr.HostPlatform, noMatchingPlatformErr.HostArch,
-					proj.BranchName(), strings.Join(branches, "\n - ")))
+					proj.BranchName(), strings.Join(branches, "\n - ")),
+					errs.SetInput())
```
mitchell-as, line 113: This is the change made to fix the bug.

Naatan (approved):

PR #3490, opened 2024-09-16
Do not validate commit ID from within projectfile.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3050" title="DX-3050" target="_blank">BugDX-3050</a>  The attempt to fix invalid URL by `state reset` fail
  </td></table>
  <br />
 

Doing so prevents `state reset` from fixing a bad ID.

Any runners that need a commit ID get it from localcommit, which raises an invalid commit ID error, so this validation is superfluous.

### DIFF
--- a/pkg/projectfile/projectfile.go
+++ b/pkg/projectfile/projectfile.go
@@ -31,7 +31,6 @@ import (
 	"github.com/ActiveState/cli/internal/sliceutils"
 	"github.com/ActiveState/cli/internal/strutils"
 	"github.com/ActiveState/cli/pkg/sysinfo"
-	"github.com/go-openapi/strfmt"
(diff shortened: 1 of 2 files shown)


Naatan (approved):
I don't like removing functionality to fix a bug. But I also see little value in the functionality being removed here. It wasn't a very friendly error for the user, and from what I can tell through the blame log this was introduced for headless commits, which we no longer have.

So accepting this one but beware of fixing bugs this way, more often than not this'd be an inappropriate fix.

mitchell-as:
I agree. I tried to fix this properly, but ran into a mountain of issues because project parsing was occurring so early in main.go. After going back to the drawing board, I came up with this solution.

PR #3491, opened 2024-09-17
Remove shell PS1/prompt modifications.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3020" title="DX-3020" target="_blank">StoryDX-3020</a>  Drop PS1 modifications from shells
  </td></table>
  <br />


mitchell-as:
The lone test failure is unrelated to this PR.

MDrakos (approved):
Looks good. There may be some other tests that expect the shell modification. Might be worth doing a quick audit, but we'll also catch them on the nightlies.

mitchell-as:
Yeah, I did an audit and found two total tests. Perhaps there's another one hidden, but yeah, the nightlies will catch it if it exists.

PR #3492, opened 2024-09-17
Added test that verifies a bare `state` is not making any API calls.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2897" title="DX-2897" target="_blank">TaskDX-2897</a>  We have an integration test that checks for API requests on plain state tool invocations
  </td></table>
  <br />

### DIFF
--- a/test/integration/api_int_test.go
+++ b/test/integration/api_int_test.go
@@ -1,9 +1,12 @@
 package integration
 
 import (
+	"path/filepath"
+	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


mitchell-as:

I also did a local test with `state pkg` and it failed due to making an API call, so we're not deluding ourselves with a test that would never fail.

MDrakos (approved):

PR #3493, opened 2024-09-17
Added `ACTIVESTATE_CLI_IGNORE_ENV` for ignoring env vars during runtime setup.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3011" title="DX-3011" target="_blank">TaskDX-3011</a>  Support overriding PYTHONPATH
  </td></table>
  <br />
 

For example, PYTHONPATH.

### DIFF
--- a/pkg/runtime/internal/envdef/environment.go
+++ b/pkg/runtime/internal/envdef/environment.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/osutils"
+	"github.com/ActiveState/cli/internal/sliceutils"
 	"github.com/thoas/go-funk"
 
 	"github.com/ActiveState/cli/internal/fileutils"
@@ -134,14 +135,10 @@ func NewEnvironmentDefinition(fp string) (*EnvironmentDefinition, error) {
 		}
 
 		// Remove any environment variables to ignore.
-		for i := 0; i < len(ed.Env); {
-			if _, exists := ignore[ed.Env[i].Name]; !exists {
-				i++
-				continue
-			}
-			ed.Env[i] = ed.Env[len(ed.Env)-1]
-			ed.Env = ed.Env[:len(ed.Env)-1]
-		}
+		ed.Env = sliceutils.Filter(ed.Env, func(e EnvironmentVariable) bool {
+			_, exists := ignore[e.Name]
+			return !exists
+		})
 	}
 
 	return ed, nil


pkg/runtime/runtime.go
```diff
@@ -157,6 +159,14 @@ func (r *Runtime) getEnv(inherit bool) (map[string]string, map[string]string, er
 		return empty, empty, errs.Wrap(err, "Failed to get environment variables")
 	}
 
+	if ignores := os.Getenv(constants.IgnoreEnvEnvVarName); ignores != "" {
+		for _, name := range strings.Split(ignores, ",") {
+			if _, exists := vars[name]; exists {
+				delete(vars, name)
+			}
+		}
+	}
```
Naatan, line 168: This is deleting environment variables regardless of whether we created them. If inherit is true then you might be deleting the very PYTHONPATH the user was trying to set, along with ours.

I think we need to address this inside the envdef package. And it should prevent defining the env var, rather than delete it after the fact (unless we can do it without inheritance playing a role).

Naatan (changes requested):

pkg/runtime/runtime.go
mitchell-as: No, there's an integration test that verifies we're leaving the user's set PYTHONPATH alone. This deletion only affects the runtime environment, which has not yet been merged with the current one. Since this runtime environment does not contain PYTHONPATH, when merging with the current one, the user's set PYTHONPATH will remain.

I tried to make this happen in envdef, but there are at least three different places where environment mutations happen, and all would need the check:

- https://github.com/ActiveState/cli/blob/d893ec6d386118897fdf4ce87dde6931687c87d5/pkg/runtime/internal/envdef/environment.go#L178-L222
- https://github.com/ActiveState/cli/blob/d893ec6d386118897fdf4ce87dde6931687c87d5/pkg/runtime/internal/envdef/environment.go#L370-L379
- https://github.com/ActiveState/cli/blob/d893ec6d386118897fdf4ce87dde6931687c87d5/pkg/runtime/internal/envdef/environment.go#L337-L362

I thought it would be better to capture it at the top-level than all those low-level places where things can slip through the cracks if we're not careful.

Naatan: I agree having a central place for it would be ideal, but this still feels too error prone to me. The fact that it passes in your tests worries me more than that it settles my concern, because reading through this code suggests that it should not work.

If we are to place the logic here then there needs to be an understanding of why that test is passing, and the code should clearly communicate either via logic or comments that this is working with an awareness of inheritance.

Naatan (changes requested):

mitchell-as:
Okay, I had the realization I could try to cut of envdef loading at the source, and that appears to work. Perhaps this will be okay?

pkg/runtime/internal/envdef/environment.go
```diff
@@ -125,6 +126,24 @@ func NewEnvironmentDefinition(fp string) (*EnvironmentDefinition, error) {
...
+		}
+
+		// Remove any environment variables to ignore.
+		for i := 0; i < len(ed.Env); {
+			if _, exists := ignore[ed.Env[i].Name]; !exists {
+				i++
+				continue
+			}
+			ed.Env[i] = ed.Env[len(ed.Env)-1]
+			ed.Env = ed.Env[:len(ed.Env)-1]
```
mitchell-as, line 143: Order doesn't matter in envdefs, right?

Naatan: I don't believe so, but this doesn't seem to change the order?

Note `sliceutils.Filter()` would be good here (easier to digest).

pkg/runtime/internal/envdef/environment.go
```diff
@@ -125,6 +126,24 @@ func NewEnvironmentDefinition(fp string) (*EnvironmentDefinition, error) {
...
+	if ignores := os.Getenv(constants.IgnoreEnvEnvVarName); ignores != "" {
+		ignore := make(map[string]bool)
+		for _, name := range strings.Split(ignores, ",") {
+			ignore[name] = true
+		}
+
+		// Remove any environment variables to ignore.
+		for i := 0; i < len(ed.Env); {
+			if _, exists := ignore[ed.Env[i].Name]; !exists {
+				i++
```
Naatan, line 139: If the first entry matches it seems like you would have an infinite loop here, because `i` will forever remain `0`.

Naatan (changes requested):

PR #3494, opened 2024-09-17
Allow config options to be an enum.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2927" title="DX-2927" target="_blank">TaskDX-2927</a>  ConfigMediator supports an enum type
  </td></table>
  <br />
 

Use it for 'security.prompt.level'.

### DIFF
--- a/internal/runners/config/set.go
+++ b/internal/runners/config/set.go
@@ -56,7 +56,10 @@ func (s *Set) Run(params SetParams) error {
 			return locale.WrapInputError(err, "Invalid integer value")
 		}
 	case configMediator.Enum:
-		enums := option.Default.(*configMediator.Enums)
+		enums, ok := option.Default.(*configMediator.Enums)
+		if !ok {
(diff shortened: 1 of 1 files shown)


internal/runners/config/set.go
```diff
@@ -52,6 +55,14 @@ func (s *Set) Run(params SetParams) error {
 		if err != nil {
 			return locale.WrapInputError(err, "Invalid integer value")
 		}
+	case configMediator.Enum:
+		enums := option.Default.(*configMediator.Enums)
```
MDrakos, line 59: I believe this is a potential panic if the type check fails. We should add an `ok` check here.

MDrakos (changes requested):

internal/runners/config/set.go
mitchell-as: I had more faith in us, but very well, I've added a check :)

MDrakos (approved):

PR #3495, opened 2024-09-17
Add ability for runners to ignore async runtimes.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3054" title="DX-3054" target="_blank">BugDX-3054</a>  Once `optin.unstable.async_runtime` is true no commands sourcing the runtime
  </td></table>
  <br />
 

This is needed for `state refresh` and `state shell`.

### DIFF
--- a/internal/runbits/runtime/runtime.go
+++ b/internal/runbits/runtime/runtime.go
@@ -49,6 +49,7 @@ type Opts struct {
 	Archive  *checkout.Archive
 
 	ValidateBuildscript bool
+	IgnoreAsync         bool
 }
 
 type SetOpt func(*Opts)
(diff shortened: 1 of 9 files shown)


internal/runbits/runtime/runtime.go
```diff
@@ -49,6 +49,7 @@ type Opts struct {
 	Archive  *checkout.Archive
 
 	ValidateBuildscript bool
+	IgnoreAsync         bool
```
mitchell-as, line 52: I'm open to this name changing.

Naatan (approved):

PR #3496, opened 2024-09-18
Fix checkout/init change summaries

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3039" title="DX-3039" target="_blank">BugDX-3039</a>  Sub-dependencies number is the same for all dependencies when checkout the project
  </td></table>
  <br />
 

Ran into several issues in fixing this:

- Too many artifacts were being associated with an ingredient
- Checkout/init summaries were not filtering out common dependencies
- Common runtime dep calculation was plain broken
- Once above was fixed; Languages were not getting dependency counts as they were considered "common deps"
- Testing with buildplans was painful because slice ordering was inconsistent


pkg/buildplan/buildplan.go
```diff
@@ -45,7 +68,7 @@ func Unmarshal(data []byte) (*BuildPlan, error) {
 }
 
 func (b *BuildPlan) Marshal() ([]byte, error) {
-	return json.Marshal(b.raw)
+	return json.MarshalIndent(b.raw, "", "  ")
```
MDrakos, line 71: I don't think this has an effect on parsing, but MarshalIndent is for a more readable output. Was this for debugging and do we want to keep it?

pkg/buildplan/hydrate.go
```diff
@@ -212,7 +214,7 @@ func (b *BuildPlan) sanityCheck() error {
 	// Ensure all artifacts have an associated ingredient
 	// If this fails either the API is bugged or the hydrate logic is bugged
 	for _, a := range b.Artifacts() {
-		if len(a.Ingredients) == 0 {
+		if raw.IsStateToolMimeType(a.MimeType) && len(a.Ingredients) == 0 {
```
MDrakos, line 217: We're only concerned with state tool artifacts having ingredients now?

pkg/buildplan/buildplan.go
```diff
@@ -32,6 +33,28 @@ func Unmarshal(data []byte) (*BuildPlan, error) {
...
+	for _, a := range b.raw.Artifacts {
+		sort.Slice(a.RuntimeDependencies, func(i, j int) bool { return a.RuntimeDependencies[i] < a.RuntimeDependencies[j] })
+	}
+	for _, step := range b.raw.Steps {
+		sort.Slice(step.Inputs, func(i, j int) bool { return step.Inputs[i].Tag < step.Inputs[j].Tag })
+		sort.Slice(step.Outputs, func(i, j int) bool { return step.Outputs[i] < step.Outputs[j] })
+		for _, input := range step.Inputs {
+			sort.Slice(input.NodeIDs, func(i, j int) bool { return input.NodeIDs[i] < input.NodeIDs[j] })
+		}
+	}
```
MDrakos, line 56: nit: Might be nice to have this in a method similar to how you have `b.cleanup()` above.

pkg/buildplan/mock_test.go
```diff
@@ -9,14 +12,15 @@ import "github.com/ActiveState/cli/pkg/buildplan/raw"
...
+//
 //	-> 00000000-0000-0000-0000-000000000002
 //	  -> 00000000-0000-0000-0000-000000000003
 //	    -> 00000000-0000-0000-0000-000000000001 (Cycle back to the first artifact)
 func createMockArtifactWithCycles() *Artifact {
 	// Create the artifacts with placeholders
-	artifact0001 := &Artifact{ArtifactID: "00000000-0000-0000-0000-000000000001"}
-	artifact0002 := &Artifact{ArtifactID: "00000000-0000-0000-0000-000000000002"}
-	artifact0003 := &Artifact{ArtifactID: "00000000-0000-0000-0000-000000000003"}
+	artifact0001 := &Artifact{ArtifactID: "00000000-0000-0000-0000-000000000001", MimeType: types.XArtifactMimeType}
```
MDrakos, line 21: I added these mime types when I was trying to filter on them. Are they still needed?

MDrakos (approved):
My comments and questions are all nits

pkg/buildplan/buildplan.go
Naatan: It was indeed for debugging, but I figured it would be worth keeping. It'll be useful if we're doing proxy debugging. And the extra whitespace really doesn't have a noticeable effect on performance.

pkg/buildplan/hydrate.go
Naatan: They're the only type of artifact that is guaranteed to have an ingredient, so the only thing we can reasonably validate.

pkg/buildplan/buildplan.go
Naatan: Good call!

MDrakos (approved):

PR #3497, opened 2024-09-18
Add `state export deptree` commands

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3056" title="DX-3056" target="_blank">TaskDX-3056</a>  We can debug dependency trees 
  </td></table>
  <br />


internal/runners/export/deptree/artifacts.go
```diff
@@ -0,0 +1,142 @@
+package deptree
```
Naatan, line 1: Please note the code in runners/export/deptree was writen for debugging purposes. I made architectural shortcuts for sure, this code is decidedly unstable and does not follow best practice. I've committed it as-is because it still holds a lot of value to us, but please take it for what it is in your review.

internal/runners/export/deptree/artifacts.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (d *DeptreeByArtifacts) printArtifacts(
+	parents []*buildplan.Artifact,
+	as buildplan.Artifacts,
+	platformID strfmt.UUID,
+	level int,
+	levelLimit int) {
+	indent := strings.Repeat("  ", level)
+	if level == levelLimit {
+		d.prime.Output().Print(indent + indentValue + "[ORANGE]Recursion limit reached[/RESET]")
```
MDrakos, line 110: Nit: this isn't an error but a tip/note about passing the `--limit` flag to increase the recursion might helpful.

pkg/buildplan/buildplan.go
```diff
@@ -33,6 +36,33 @@ func Unmarshal(data []byte) (*BuildPlan, error) {
...
+		sort.Slice(step.Outputs, func(i, j int) bool { return step.Outputs[i] < step.Outputs[j] })
+		for _, input := range step.Inputs {
+			sort.Slice(input.NodeIDs, func(i, j int) bool { return input.NodeIDs[i] < input.NodeIDs[j] })
+		}
+	}
+
+	v, _ := b.Marshal()
+	vs := string(v)
+	_ = vs
+	fileutils.WriteFile(fmt.Sprintf("/tmp/buildplan-%d.json", time.Now().Unix()), v)
```
MDrakos, line 64: Didn't we move all of this sorting to it's own function in your last PR that fixed the dependency count?

internal/runners/export/deptree/artifacts.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+type primeable interface {
+	primer.Auther
+	primer.Outputer
+	primer.Configurer
+	primer.Projecter
+	primer.Analyticer
+	primer.SvcModeler
+}
```
MDrakos, line 27: Only auth, project, and output appear to be used here.

internal/runners/export/deptree/artifacts.go
```diff
@@ -0,0 +1,142 @@
...
+	if platformID == "" {
+		platformID, err = model.FilterCurrentPlatform(sysinfo.OS().String(), bp.Platforms(), "")
+		if err != nil {
+			return errs.Wrap(err, "Could not get platform ID")
+		}
+	}
+
+	levelLimit := params.LevelLimit
+	if levelLimit == 0 {
+		levelLimit = 10
```
MDrakos, line 79: Nit: maybe this magic number could be a constant that is shared between runners in this package? May also be worth mentioning this default in the flags description.

MDrakos (approved):
Looks good, my comments don't flag anything functional. I'm just noting some potential improvements to consider.

pkg/buildplan/buildplan.go
Naatan: Yes, it seems something went wrong with the merge. Thanks for catching!

PR #3502, opened 2024-09-23
Test for hardlink support on Windows before doing it.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2943" title="DX-2943" target="_blank">TaskDX-2943</a>  We support setting up a runtime on the non-windows drive
  </td></table>
  <br />


mitchell-as:
Test failures are either known or sporadic. They are unrelated to this PR.

pkg/runtime/links_windows.go
```diff
@@ -0,0 +1,50 @@
...
+			return false
+		}
+	}
+
+	logging.Debug("Attempting to link '%s' to '%s'", lnk, target)
+	err := smartlink.Link(target, lnk)
+	if err != nil {
+		logging.Debug("Test link creation failed: %v", err)
+	}
+	return err == nil
```
Naatan, line 49: Please cleanup the source and target.

pkg/runtime/depot.go
```diff
@@ -146,6 +148,10 @@ func (d *depot) Put(id strfmt.UUID) error {
 
 // DeployViaLink will take an artifact from the depot and link it to the target path.
 func (d *depot) DeployViaLink(id strfmt.UUID, relativeSrc, absoluteDest string) error {
+	if !d.supportsHardLinks {
+		return d.DeployViaCopy(id, relativeSrc, absoluteDest)
+	}
```
Naatan, line 153: This needs to happen at a higher level. Recall we record how files were deployed in a config file. We'll need to ensure that this config has the correct value also.

Naatan (changes requested):

pkg/runtime/depot.go
mitchell-as: I don't see where this can happen. The next level up is in setup.go (https://github.com/ActiveState/cli/blob/84273faa27f55bc371a897ef902b9b2c8f9c9edd/pkg/runtime/setup.go#L466-L477), and I'm not sure we want to expose and make setup change its behavior based on whether or not the depot supports hard links -- it feels like an internal implementation issue.

Also, the storage of deployment type in a config file happens in the DeployVia*() methods, so it will have the correct value.

If there is more to do, please advise.

Naatan: ok, we're gonna have to hoist this up to setup.go then. The depots only concern is making it happen, setup.go owns the logic of interpreting envdef and invoking the appropriate deployer. We may want to revisit that given all the recent changes, but probably not for this story, unless addressing it in setup.go becomes difficult or creates other challenges.

Naatan (changes requested):
See https://github.com/ActiveState/cli/pull/3502#discussion_r1772288515

mitchell-as:
Test failures are other known nightly failures. All failures due to the "Package added:" -> "Added:" locale change are fixed.

Naatan (changes requested):
Please move the hardlink detection logic and concerns to setup. Now we have an entanglement in responsibilities between setup and depot.

pkg/runtime/setup.go
```diff
@@ -470,10 +472,14 @@ func (s *setup) install(id strfmt.UUID) (rerr error) {
 		if err := envDef.ApplyFileTransforms(s.path); err != nil {
 			return errs.Wrap(err, "Could not apply env transforms")
 		}
-	} else {
+	} else if s.supportsHardLinks {
 		if err := s.depot.DeployViaLink(id, envDef.InstallDir, s.path); err != nil {
 			return errs.Wrap(err, "Could not deploy artifact via link")
 		}
+	} else {
```
Naatan, line 479: I think you can drop this and make line 468 read 

```
if envDef.NeedsTransforms() || !s.supportsHardlinks {
```

The fact that it will run `ApplyFileTransforms` doesn't matter, since there wouldn't be any to apply. Though for robustness it wouldn't hurt to wrap a conditional around that.

Naatan (changes requested):

pkg/runtime/setup.go
```diff
@@ -465,21 +465,19 @@ func (s *setup) install(id strfmt.UUID) (rerr error) {
...
 	}
 
-	if envDef.NeedsTransforms() {
+	if envDef.NeedsTransforms() || !s.supportsHardLinks {
 		if err := s.depot.DeployViaCopy(id, envDef.InstallDir, s.path); err != nil {
 			return errs.Wrap(err, "Could not deploy artifact via copy")
 		}
-		if err := envDef.ApplyFileTransforms(s.path); err != nil {
-			return errs.Wrap(err, "Could not apply env transforms")
+		if s.supportsHardLinks {
```
Naatan, line 472:
```suggestion
		if envDef.NeedsTransforms() {
```

Naatan (changes requested):

pkg/runtime/setup.go
mitchell-as: 🤦

mitchell-as:
Maybe this is okay now? Honestly, this PR has gone through so much churn I wouldn't be surprised if I'm still missing something.

Naatan (approved):
Got there 😅

PR #3504, opened 2024-09-24
Fix nil pointer panic

Naatan:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3072" title="DX-3072" target="_blank">BugDX-3072</a>   Panic: 'runtime error: index out of range [0] with length 0' when install package
  </td></table>
  <br />


mitchell-as (approved):
That'll do it.

PR #3505, opened 2024-09-24
The only commands that should ignore async runtimes are `refresh`, `deploy`, and `activate`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3071" title="DX-3071" target="_blank">TaskDX-3071</a>  Limit async runtime to JUST `state refresh`
  </td></table>
  <br />
 

Show a warning to the user when async runtime cases update skips.


mitchell-as:
Test failures look to be known failures. The one failure I know for sure was due to async runtime was fixed.

Naatan (approved):

PR #3509, opened 2024-09-24
Fix install test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3074" title="DX-3074" target="_blank">BugDX-3074</a>  Nightly failure: TestInstallIntegrationTestSuite/TestInstall_NoMatches_Alternatives
  </td></table>
  <br />
 

The underlying issue is that when we don't find an ingredient for the requested requirement we don't have a namespace. In this example when we call the search endpoint with `database` and no namespace we first get a bunch of results for Perl.

I've updated the `getSuggestions` function to compose namespaces for each language and call the search endpoint for each one.

### DIFF
--- a/internal/runners/install/rationalize.go
+++ b/internal/runners/install/rationalize.go
@@ -61,9 +61,32 @@ func (i *Install) rationalizeError(rerr *error) {
 }
 
 func (i *Install) getSuggestions(req *requirement, languages []model.Language) ([]string, error) {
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
This isn't a viable solution as it would scale up our search calls for each namespace being targeted in our project. The search endpoint is known to be slow. Even the newly optimized one. Already with just the one lookup it's kinda awkward when having to wait for the error to appear.

Perhaps the new search endpoint will allow us to search across multiple namespaces. Suggest marking this issue blocked by DX-2994.

PR #3517, opened 2024-10-01
Added ceremonial break.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3088" title="DX-3088" target="_blank">BugDX-3088</a>  Buildexpression unmarshalling may be dropping args
  </td></table>
  <br />
 

Function calls in JSON have a single argument object with a list of values. We already assert that the object is of length 1 (the list of args; https://github.com/ActiveState/cli/blob/2deb60816da0be22d8f7aeef6a5fcead2274cc86/pkg/buildscript/unmarshal_buildexpression.go#L259-L264), but a cursory glance at a later loop in the code may lead one to believe there is more than one object. I've added a ceremonial break with a comment for clarity.

We're not dropping any args; the interface contains a list of all arguments.

### DIFF
--- a/pkg/buildscript/unmarshal_buildexpression.go
+++ b/pkg/buildscript/unmarshal_buildexpression.go
@@ -283,6 +283,7 @@ func unmarshalFuncCall(path []string, m map[string]interface{}) (*FuncCall, erro
 		}
 
 		argsInterface = value
+		break // technically this is not needed since there's only one element in m
 	}
 
 	args := []*Value{}


Naatan (changes requested):
This doesn't seem like it's solving anything. It's now defaulting to the first entry rather than the last. The issue here seems to be that it's not considering all arguments, which so far is a non-issue as no function calls exist at this level (yet!) that have more than one argument.

mitchell-as:
No, this fix is making a point that the length of m is expected to be 1 (thanks to the assertion above). It is impossible for there to be more than one "argument". What you take to be an argument is actually a list of arguments (which you can see being processed below).

This is not a bug. If there will be more than one "thing" in the argument object, then we will have to file a task to expand what we're doing here.

Naatan (approved):
My bad; I missed the comment about `m>1`.

PR #3520, opened 2024-10-02
If `ACTIVESTATE_API_KEY` is set, try to authenticate with it, or explicitly error out.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3080" title="DX-3080" target="_blank">TaskDX-3080</a>  We error out if the stored API key is invalid
  </td></table>
  <br />

### DIFF
--- a/internal/runners/auth/auth.go
+++ b/internal/runners/auth/auth.go
@@ -1,6 +1,9 @@
 package auth
 
 import (
+	"os"
+
+	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/keypairs"
 	"github.com/ActiveState/cli/internal/locale"
 	"github.com/ActiveState/cli/internal/output"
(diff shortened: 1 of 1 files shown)


internal/runners/auth/auth.go
```diff
@@ -91,6 +94,14 @@ func (a *Auth) authenticate(params *AuthParams) error {
 		return auth.AuthenticateWithToken(params.Token, a.Auth)
 	}
 
+	if apiKey := os.Getenv(constants.APIKeyEnvVarName); apiKey != "" {
+		err := auth.AuthenticateWithToken(apiKey, a.Auth)
+		if err != nil {
+			return locale.WrapError(err, "err_auth_api_key", "Failed to authenticate with [ACTIONABLE]{{.V0}}[/RESET] environment variable", constants.APIKeyEnvVarName)
+		}
+		return nil
+	}
```
MDrakos, line 103: I wonder if this error reporting should be in the lower level `auth` package. The user would only get this error message if they explicitly ran `state auth` but we are attempting to authenticate with the api key environment variable through multiple code paths in the `auth` package. See [here](https://github.com/ActiveState/cli/blob/master/pkg/platform/authentication/auth.go#L218) for one.

mitchell-as: That's what I did originally, but I couldn't trigger it without `state auth`. We have lots of places that check `auth.Authenticated()` and bail out with "you must be logged in" if not. I was only able to show my message using `state auth`. The only callers of your linked method are "Sync" (not applicable to this PR), and "Client". All instances of the latter appear to be walled behind `auth.Authenticated()` checks, so in practice, I think we're okay.

Given that the confusion arises around the whole `state auth` flow, I figured this was an okay place to put it.

Naatan (changes requested):
This is addressing ONLY the `state auth` case. The issue here is when we pass the API key through an env var, which is the more common use-case. This scenario would not give any sort of error.

Also re the slack thread with Damien raising the 60 second polling of auth which seemed to come from this PR; we should ensure that we do not retry when the API key is known to be invalid. 

I'll update the story to clarify these points.

mitchell-as:
This jumped up in complexity, so closing.

PR #3522, opened 2024-10-02
Fixed test failure.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3097" title="DX-3097" target="_blank">BugDX-3097</a>  Nightly failure:  TestProgressIntegrationTestSuite/TestProgress
  </td></table>
  <br />
 

The solver is getting faster!

### DIFF
--- a/test/integration/progress_int_test.go
+++ b/test/integration/progress_int_test.go
@@ -19,16 +19,17 @@ func (suite *ProgressIntegrationTestSuite) TestProgress() {
 	defer ts.Close()
 
 	cp := ts.Spawn("checkout", "ActiveState-CLI/Empty")
+	cp.Expect("Resolving Dependencies")
+	cp.ExpectRe(`[^.]+?✔ Done`, e2e.RuntimeSolvingTimeoutOpt)
 	cp.Expect(locale.T("install_runtime"))
 	cp.Expect("Checked out", e2e.RuntimeSourcingTimeoutOpt)
-	suite.Assert().NotContains(cp.Output(), "...")
 	cp.ExpectExitCode(0)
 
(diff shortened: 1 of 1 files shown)


test/integration/progress_int_test.go
```diff
@@ -19,16 +19,17 @@ func (suite *ProgressIntegrationTestSuite) TestProgress() {
 	defer ts.Close()
 
 	cp := ts.Spawn("checkout", "ActiveState-CLI/Empty")
+	cp.Expect("Resolving Dependencies")
+	cp.ExpectRe(`[^.]+?✔ Done`, e2e.RuntimeSolvingTimeoutOpt)
```
Naatan, line 23: This first checkout seems to specifically be about asserting that it does NOT use dot progress, but these changes seem to make it so that it is asserting that it DOES use dot progress.

Naatan (changes requested):

test/integration/progress_int_test.go
mitchell-as: This test is expecting that there are no dots between "Resolving Dependencies" and  "✔ Done", so it's asserting that it does NOT use dot progress.

I had to do a few iterations of this test, and during these iterations I found that `cp.Output()` gave me all the escape sequences of the \\, |, and / interactive progress between "Resolving Dependencies" and "✔ Done". I am confident this test is working as intended.

Naatan: Ahh.. the way the regex is written I'd assume it was checking for `... ✔ Done`. I had interpreted the dot as a literal dot rather than a regex rule. Might be worth adding a comment for clarity.

Naatan (approved):

PR #3527, opened 2024-10-03

mitchell-as:
<table>
<td>
  </td></table>
  <br />

### DIFF
--- a/internal/runbits/buildscript/file.go
+++ b/internal/runbits/buildscript/file.go
@@ -19,15 +19,16 @@ import (
 
 // projecter is a union between project.Project and setup.Targeter
 type projecter interface {
-	ProjectDir() string
+	Dir() string
 	Owner() string
 	Name() string
+	BranchName() string
 }
 
 var ErrBuildscriptNotExist = errors.New("Build script does not exist")
 
 func ScriptFromProject(proj projecter) (*buildscript.BuildScript, error) {
-	path := filepath.Join(proj.ProjectDir(), constants.BuildScriptFileName)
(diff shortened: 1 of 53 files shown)


internal/runbits/buildscript/file.go
```diff
@@ -47,24 +48,24 @@ type primeable interface {
 	primer.SvcModeler
 }
 
-func Initialize(path string, auth *authentication.Auth, svcm *model.SvcModel) error {
-	scriptPath := filepath.Join(path, constants.BuildScriptFileName)
+func Initialize(proj projecter, auth *authentication.Auth, svcm *model.SvcModel) error {
```
mitchell-as, line 51: I know that we would prefer to not make buildscripts aware of projects, but since we're not having a primed checkoutinfo anymore, this is the next best thing. I could pass dir, owner, name, branch if you want though.

Naatan (changes requested):
I think based on our conversation yesterday that we can avoid infecting all our buildscript logic with check info, because we determined the only time it really matters is at checkout / init.

So I would suggest instead adding something for decorating a buildscript with the checkoutinfo, which checkout and init can manually call.

The buildscript itself should then be smart enough to retain the project information data during conversion to/from build expression. So rather than essentially bake it in every time, it's more of a "retain what you already had" operation.

I'd suggest adding a marshaller/unmarshaller specifically for the checkoutinfo, as it would help with the solution described above, as well as address your concern about buildscript parsing issues interrupting checkoutinfo based logic.

The churn continues, sorry. Let me know if you're growing sick of this story.

mitchell-as:
Yeah, let's put this one on the backburner then.

PR #3528, opened 2024-10-03
Add config list

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3055" title="DX-3055" target="_blank">StoryDX-3055</a>  Users can list available and configured config keys/values with `state config` 
  </td></table>
  <br />


internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+
+	if len(v) > 100 {
+		v = v[:100] + "..."
+	}
+
+	if value == opt.Default {
+		return fmt.Sprintf("[GREEN]%s[/RESET]", v)
+	}
+
+	return fmt.Sprintf("[BOLD][RED]%s*[/RESET]", v)
```
Naatan, line 97: It looks like I screwed this up a bit in the mock (fixed now), but the color of green/blue is for booleans being true/false. It does not relate to default or not. 

Additionally, we want to highlight values that are defined by the user as bold and with an asterisk. This indicates whether the user has specified a value, it does not specify whether the value also happens to be the same as default.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+}
+
+func formatDefault[T any](defaultValue T) string {
+	v := fmt.Sprintf("%v", defaultValue)
+	if v == "" {
+		v = "\"\""
+	}
```
Naatan, line 104: Currently string values have no quotes around them, but for formatValue they do. I suggest having formatValue and formatDefault invoke the same logic for quoting and string lengths (default values could be long too).

cmd/state/internal/cmdtree/config.go
```diff
@@ -10,13 +10,13 @@ import (
 func newConfigCommand(prime *primer.Values) *captain.Command {
```
Naatan, line 10: We should mark it unstable, since technically this is a new command.

That means we're marking a command that used to be "stable" as "unstable", but given the original command was just printing help output I don't see a problem with that.

I'm pretty sure unstable doesn't propagate to child commands (ie. config set), but please make sure.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+}
+
+type configOutput struct {
+	out     output.Outputer
+	cfg     *config.Instance
+	options []mediator.Option
+	data    []configData
+}
+
+func (c *List) Run(usageFunc func() error) error {
```
Naatan, line 46: We no longer need `usageFunc`.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
 		cfg: prime.Config(),
 	}, nil
 }
 
-func (c *Config) Run(usageFunc func() error) error {
-	return usageFunc()
+type configData struct {
+	Key     string `locale:"key,Key"`
+	Value   string `locale:"value,Value"`
+	Default string `locale:"default,Default"`
```
Naatan, line 36: We've started moving away from magical output handling, please see how I handled table layout here:

https://github.com/ActiveState/cli/blob/d43bc4921ea8fccf7f1848f4ba2e22004121e145/internal/runners/upgrade/upgrade.go#L248-L248

Basically it's just manually constructing the table and rendering the resulting string itself.

Naatan: Also please add json keys, right now json output is capitalized.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
 	primer.Analyticer
 }
 
-func NewConfig(prime primeable) (*Config, error) {
-	return &Config{
+func NewList(prime primeable) (*List, error) {
```
Naatan, line 26: Please store prime, eg. https://github.com/ActiveState/cli/blob/d43bc4921ea8fccf7f1848f4ba2e22004121e145/internal/runners/upgrade/upgrade.go#L46-L46

After the runtime refactor this is the new pattern I'd like to move us to. We still need the primeable interface but at least we are removing a bit of symbol soup this way.

internal/mediators/config/registry.go
```diff
@@ -74,3 +84,15 @@ func GetDefault(opt Option) interface{} {
 	}
 	return opt.Default
 }
+
+// AllRegistered returns all registered options, excluding hidden ones
+func AllRegistered() []Option {
```
Naatan, line 89:
```suggestion
// Registered returns all registered options, excluding hidden ones
func Registered() []Option {
```

"All" seems redundant.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+	cfg     *config.Instance
+	options []mediator.Option
+	data    []configData
+}
+
+func (c *List) Run(usageFunc func() error) error {
+	registered := mediator.AllRegistered()
+	sort.SliceStable(registered, func(i, j int) bool {
+		return registered[i].Name < registered[j].Name
+	})
```
Naatan, line 50: May as well do this in the mediator function.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+
+	if v == "" {
+		return "\"\""
+	}
+
+	if len(v) > 100 {
+		v = v[:100] + "..."
+	}
+
+	if value == opt.Default {
```
Naatan, line 93: Should use `config.GetDefault()`, as this isn't handling enum defaults.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+		return registered[i].Name < registered[j].Name
+	})
+
+	var data []configData
+	for _, opt := range registered {
+		configuredValue := c.cfg.Get(opt.Name)
+		data = append(data, configData{
+			Key:     formatKey(opt.Name),
+			Value:   formatValue(opt, configuredValue),
+			Default: formatDefault(mediator.GetDefault(opt)),
```
Naatan, line 58: For JSON output the formatting here isn't desirable. eg. string values now have double quotes.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+		options: registered,
+		data:    data,
+	})
+
+	return nil
+}
+
+func formatKey(key string) string {
+	return fmt.Sprintf("[CYAN]%s[/RESET]", key)
+}
```
Naatan, line 74: Nit: this function feels unnecessary. It's only called once and its logic is of equivalent length to the function definition itself.

internal/runners/config/config.go
```diff
@@ -18,13 +23,103 @@ type primeable interface {
...
+
+func formatDefault[T any](defaultValue T) string {
+	v := fmt.Sprintf("%v", defaultValue)
+	if v == "" {
+		v = "\"\""
+	}
+	return fmt.Sprintf("[DISABLED]%s[/RESET]", v)
+}
+
+func (c *configOutput) MarshalOutput(format output.Format) interface{} {
```
Naatan, line 108: Please handle structured output manually, eg. https://github.com/ActiveState/cli/blob/d43bc4921ea8fccf7f1848f4ba2e22004121e145/internal/runners/upgrade/upgrade.go#L152-L152

I'm not completely sure yet what the go-forward pattern is going to be yet, but I am sure it's going to be a lot more explicit. I'd like to move away from the magical handling.

Naatan (changes requested):

cmd/state/internal/cmdtree/config.go
MDrakos: It appears that we _do_ set the child commands to unstable as well. https://github.com/ActiveState/cli/blob/master/internal/captain/command.go#L410

Naatan (changes requested):

Naatan (approved):

PR #3529, opened 2024-10-04
Added installable githook for creating a platform commit with the buildscript changes.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3092" title="DX-3092" target="_blank">StoryDX-3092</a>  We have a git hook that runs `state commit` automatically on `git commit`
  </td></table>
  <br />
 

The hook also auto-updates the buildscript with any resulting changes prior to commit.

The end result is that the committed buildscript will be the latest one, but it may NOT be what the user intended to commit. For example, if the ordering of requirements in the platform commit's buildscript is different, the buildscript committed locally will have the new order, not the order the user originally intended to commit.

### DIFF
--- a/scripts/git-hooks/buildscript-commit
+++ b/scripts/git-hooks/buildscript-commit
@@ -1,4 +1,6 @@
 #!/bin/bash
+# Git pre-commit hook to run `state commit` prior to committing any build script changes.
+# Usage: call this script from your git pre-commit hook or use it as your pre-commit hook.
 
 if git diff --name-only --cached | grep -q buildscript.as; then
 	echo "Running 'state commit', as buildscript.as is being committed."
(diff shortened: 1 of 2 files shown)


Naatan (changes requested):
This looks good to me, but I don't think we should be using this as a githook on our project. We are very often playing with the buildscript locally in a way that is not meant to be committed.

Suggest putting this in `scripts/git-hooks/buildscript-commit`.

mitchell-as:
It may look like it, but it wasn't enabled as our pre-commit hook. I agree that we don't want that. If it was in a *.githooks* directory (with the dot), then it would be.

Regardless, I have applied your suggested name change and added some usage information to compensate.

Naatan (approved):

PR #3530, opened 2024-10-04
Remove no-op pops.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3087" title="DX-3087" target="_blank">BugDX-3087</a>  Buildexpression unmarshalling has potential pathing issues
  </td></table>
  <br />
 

We still want to keep context, but popping is not necessary since mutations are done locally (`path = append(path, ...)`), and do not affect the values passed by callers.


MDrakos (approved):
Do we even need the `path` value at all? The `pop` calls were doing nothing here but that means that the `path` value is likely incorrect in many places.

mitchell-as:
We do need path, and the values were correct in the locations they were used.

PR #3531, opened 2024-10-07
Build scripts can distinguish between solve nodes for different targets.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2960" title="DX-2960" target="_blank">TaskDX-2960</a>  We can interpret a buildscript where the solve node is not top level
  </td></table>
  <br />
 

Clients need to pass the target when querying requirements, platforms, etc. Use "" for the default target (i.e. the name assigned to 'main').

### DIFF
--- a/pkg/buildscript/queries.go
+++ b/pkg/buildscript/queries.go
@@ -47,8 +47,8 @@ func (r UnknownRequirement) IsRequirement() {}
 
 // Returns the requirements for the given target.
 // If no target is given, uses the default target (i.e. the name assigned to 'main').
-func (b *BuildScript) Requirements(target ...string) ([]Requirement, error) {
-	requirementsNode, err := b.getRequirementsNode(target...)
+func (b *BuildScript) Requirements(targets ...string) ([]Requirement, error) {
+	requirementsNode, err := b.getRequirementsNode(targets...)
 	if err != nil {
 		return nil, errs.Wrap(err, "Could not get requirements node")
 	}
@@ -99,8 +99,8 @@ func (b *BuildScript) Requirements(target ...string) ([]Requirement, error) {
 // DependencyRequirements is identical to Requirements except that it only considers dependency type requirements,
 // which are the most common.
 // ONLY use this when you know you only need to care about dependencies.
-func (b *BuildScript) DependencyRequirements(target ...string) ([]types.Requirement, error) {
-	reqs, err := b.Requirements(target...)
+func (b *BuildScript) DependencyRequirements(targets ...string) ([]types.Requirement, error) {
+	reqs, err := b.Requirements(targets...)
 	if err != nil {
 		return nil, errs.Wrap(err, "Could not get requirements")
 	}
@@ -113,8 +113,8 @@ func (b *BuildScript) DependencyRequirements(target ...string) ([]types.Requirem
 	return deps, nil
 }
 
-func (b *BuildScript) getRequirementsNode(target ...string) (*Value, error) {
-	node, err := b.getSolveNode(target...)
+func (b *BuildScript) getRequirementsNode(targets ...string) (*Value, error) {
+	node, err := b.getSolveNode(targets...)
 	if err != nil {
 		return nil, errs.Wrap(err, "Could not get solve node")
 	}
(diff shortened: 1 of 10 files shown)


pkg/buildscript/queries.go
```diff
@@ -169,15 +194,45 @@ func (b *BuildScript) getSolveNode() (*Value, error) {
...
+		}
+		a := arg.Assignment
+		if a.Key == srcKey && a.Value.Ident != nil {
+			node, err := b.getSolveNode(*a.Value.Ident)
+			if err != nil {
+				return nil, errs.Wrap(err, "Could not get solve node from target")
+			}
+			return node, nil
+		}
+	}
```
MDrakos, line 229: Why can't this solve node be identified by the `getTargetNode` above? Is it because the `target` value wasn't found? Can you please add a more detailed comment.

pkg/buildscript/unmarshal.go
```diff
@@ -50,5 +50,15 @@ func Unmarshal(data []byte) (*BuildScript, error) {
...
 
+	// Verify there are no duplicate key assignments.
+	// This is primarily to catch duplicate solve nodes for a given target.
+	seen := make(map[string]bool)
+	for _, assignment := range raw.Assignments {
+		if _, exists := seen[assignment.Key]; exists {
+			return nil, locale.NewInputError(locale.Tl("err_buildscript_duplicate_keys", "Build script has duplicate '{{.V0}}' assignments", assignment.Key))
+		}
+		seen[assignment.Key] = true
+	}
```
MDrakos, line 61: Does this still allow multiple solve nodes for different targets? The ACs in the story don't stipulate that it has to be multiple solve nodes for a _single_ target that raise an error. Maybe the story needs to be updated?

pkg/buildscript/queries.go
```diff
@@ -147,7 +150,23 @@ func getVersionRequirements(v *Value) []types.VersionRequirement {
 	return reqs
 }
 
-func (b *BuildScript) getSolveNode() (*Value, error) {
+func isSolveFuncName(name string) bool {
+	return name == solveFuncName || name == solveLegacyFuncName
+}
+
+func (b *BuildScript) getTargetNode(target string) (*Value, error) {
```
MDrakos, line 157: nit: Any taste for a variadic argument here? Would allow us to forgo the passing of `""` and be ready for multiple targets when the time comes.

MDrakos (changes requested):

pkg/buildscript/unmarshal.go
mitchell-as: Yes, you can have more than one solve node in a buildscript. There's only an issue if you have a buildscript whose target has two different solve nodes. For example:

```
runtime = state_tool_artifacts_v1(src = sources)
sources = solve(...)
sources = solve(... something different...)
another_solve = solve(...)
```

This error would identify `sources` has having duplicate assignments. `another_solve` is totally fine, as it's a different target or whatever.

pkg/buildscript/queries.go
mitchell-as: Yes! I was looking for an excuse to make this an optional argument, and multiple target support justifies it!

pkg/buildscript/queries.go
mitchell-as: Consider this build script snippet:

```
runtime = state_tool_artifacts_v1(src = sources)
sources = solve(...)
main = runtime
```

`getTargetNode()` would return `sources` (`main -> runtime -> src`). Since `sources` is not a solve node (it's a reference), we have to look again for the `sources` node. The second time, `getTargetNode("sources")` will give us the solve node we need.

pkg/buildscript/queries.go
```diff
@@ -169,15 +195,50 @@ func (b *BuildScript) getSolveNode() (*Value, error) {
...
 		return nil
 	}
+
 	if node := search(b.raw.Assignments); node != nil {
 		return node, nil
 	}
+	return nil, errNodeNotFound
+}
+
+func (b *BuildScript) getSolveNode(target ...string) (*Value, error) {
```
MDrakos, line 205: nit: This should probably be `targets`

MDrakos (approved):
Thanks for the clarifications!

MDrakos (approved):
Thank you for dealing with my nits 😆

PR #3539, opened 2024-10-11
Handle Invalid API Key

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3080" title="DX-3080" target="_blank">TaskDX-3080</a>  We error out if the stored API key is invalid
  </td></table>
  <br />

### DIFF
--- a/internal/runners/auth/auth.go
+++ b/internal/runners/auth/auth.go
@@ -1,9 +1,6 @@
 package auth
 
 import (
-	"os"
-
-	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/keypairs"
 	"github.com/ActiveState/cli/internal/locale"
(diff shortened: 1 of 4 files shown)


internal/runners/auth/auth.go
```diff
@@ -91,6 +92,14 @@ func (a *Auth) authenticate(params *AuthParams) error {
 		return auth.AuthenticateWithToken(params.Token, a.Auth)
 	}
 
+	if token := a.Auth.AvailableAPIToken(); token != "" {
+		err := auth.AuthenticateWithToken(token, a.Auth)
+		if err != nil {
+			return errs.Wrap(err, "Failed to authenticate with API key")
+		}
+		return nil
+	}
```
Naatan, line 101: What was this added for? This scenario is already handled, afaik.

Naatan (changes requested):
Have you tested this? Please add an integration test for this.

I'm pretty sure the code as written won't relay anything to the user because it's failing at the state-svc level, but I could be wrong.

MDrakos:
Closing this as there is a new blocker. See story for details.

PR #3543, opened 2024-10-16
Check process in use

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3040" title="DX-3040" target="_blank">TaskDX-3040</a>  Windows, Unable to delete cache folder
  </td></table>
  <br />


MDrakos:
Spent some time trying to fix the integration test failure on Windows as I was in the code but it was becoming a rabbit hole. I will double check that there is a ticket for it and create one if not. The failure is unrelated to this PR.

internal/runners/clean/cache.go
```diff
@@ -76,9 +80,16 @@ func (c *Cache) removeCache(path string, force bool) error {
...
 	}
 
-	logging.Debug("Removing cache path: %s", path)
-	err := removeCache(c.path)
+	inUse, err := c.checkPathInUse(path)
 	if err != nil {
+		return errs.Wrap(err, "Failed to check if path is in use")
+	}
+	if inUse {
+		return locale.NewError("err_clean_in_use")
```
Naatan, line 88: Should make this an input error (same for the other one below).

internal/runners/clean/cache.go
```diff
@@ -109,3 +128,19 @@ func (c *Cache) removeProjectCache(projectDir, namespace string, force bool) err
 
 	return nil
 }
+
+func (c *Cache) checkPathInUse(path string) (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), model.SvcTimeoutMinimal)
```
Naatan, line 133: I don't think we should have such an aggressive timeout here. Especially for process in use checks, which on Windows tend to be slow. Do we bubble up timeout errors or that would just result in no processes in use being asserted?

Naatan (changes requested):

internal/runners/clean/cache.go
MDrakos: Good point. I've increased it a fair amount but let me know if we should do further adjustments. 

Looking at the client code the error from the context timeout _should_ be bubbled up as a `context deadline exceeded`. Won't be super useful to the user... we could maybe catch that specific error and continue with the removal if we encounter it.

internal/runners/clean/cache.go
Naatan: You updated the other one but not this one 😅

Naatan (changes requested):

internal/runners/clean/cache.go
MDrakos: 🤦‍♂️

Naatan (approved):

PR #3547, opened 2024-10-17
Handle more buildplanner errors

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2821" title="DX-2821" target="_blank">TaskDX-2821</a>  Handle Buildplanner errors with user facing messages
  </td></table>
  <br />

### DIFF
--- a/pkg/platform/api/buildplanner/response/commit.go
+++ b/pkg/platform/api/buildplanner/response/commit.go
@@ -54,7 +54,7 @@ func (c *ProjectCommitResponse) PostProcess() error {
 func ProcessBuildError(build *BuildResponse, fallbackMessage string) error {
 	logging.Debug("ProcessBuildError: build.Type=%s", build.Type)
 	if build.Type == types.PlanningErrorType {
-		return processPlanningError(build.Message, build.SubErrors)
+		return processPlanningError(build.Message, build.Error.SubErrors)
 	} else if build.Error == nil {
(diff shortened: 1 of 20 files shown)


pkg/platform/api/buildplanner/response/commit.go
```diff
@@ -133,9 +133,9 @@ type TargetNotFound struct {
 	PossibleTargets []string `json:"possibleTargets"`
 }
 
-type ValidationError struct {
-	SubErrors []*BuildExprError `json:"subErrors"`
-}
+// ValidationError represents a validation error that occurred during planning.
+// Contains message and subErrors which is handled separately
+type ValidationError struct{}
```
MDrakos, line 138: We now have multiple errors with `SubErrors` so I moved that to the common `Error` struct

MDrakos:
I found that a lot of the errors we are currently handling wouldn't be very enriched by the extra fields. I also took the time in this PR to handle errors that we weren't before.

Naatan (changes requested):
We should be using the rationalizer for this. These are already returning typed errors, which are at least to some level already rationalized. Adding localized errors underneath them feels inappropriate. We should only be marking input errors outside of rationalizers if there is a clear reason why rationalizing doesn't work there.

MDrakos:
Closing this PR and filed another story that I think will better address our buildplanner error handling.

PR #3548, opened 2024-10-17
Fix path ordering

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3116" title="DX-3116" target="_blank">BugDX-3116</a>  On Windows - `state exec` output sporadically fail to provide correct output
  </td></table>
  <br />

### DIFF
--- a/internal/runners/exec/exec.go
+++ b/internal/runners/exec/exec.go
@@ -6,7 +6,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


MDrakos:
> We should address this inside the envdef package, because `exec` is just giving us the symptom. Fixing it only in exec still leaves places like `state shell` exposed.

I'm not sure addressing this in a lower level package will be a quick fix. We are using the `virtualenvironment` package in this runner which works exclusively with maps. Most, if not all, of our environment handling is done with maps. We would have to change the return type or add another function that gives an ordered data structure. At this point a quick fix may not be ideal. We may want to address the Windows path merging story.

Naatan (changes requested):
If the issue is to do with maps and sorting; we should just eliminate the entry we don't want. So maybe some logic that checks if both PATH and Path is set, and then eliminate the one we don't want. Then sorting is irrelevant.

That's still a quick fix, because I think the proper fix here is that we teach our environment management (of any kind) to be case insensitive on Windows. At which point there shouldn't be any duplicate entries to begin with.

Naatan (changes requested):
Looks good, but please move the logic here:

https://github.com/ActiveState/cli/blob/5bbb1c9cf8705a01d6aaf62bd9760f94805e3c10/pkg/runtime/internal/envdef/collection.go#L56

Naatan (approved):

PR #3551, opened 2024-10-21
Ensure remote installer can run within restricted powershells.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2989" title="DX-2989" target="_blank">TaskDX-2989</a>  We integration test our remote-installer with restricted powershell permissions
  </td></table>
  <br />

### DIFF
--- a/test/integration/remote_installer_int_test.go
+++ b/test/integration/remote_installer_int_test.go
@@ -43,7 +43,10 @@ func (suite *RemoteInstallIntegrationTestSuite) TestInstall() {
 		}
 
 		policy := getPolicy()
-		defer setPolicy(policy)
+		defer func() {
+			setPolicy(policy)
+			suite.Assert().Equal(policy, getPolicy(), "execution policy was not reset to '"+policy+"'; subsequent test results may be invalid")
+		}()
 
 		setPolicy("Restricted")
 		suite.Assert().Equal("Restricted", getPolicy(), "should have set powershell policy to 'Restricted'")
--- a/test/integration/remote_installer_int_test.go
+++ b/test/integration/remote_installer_int_test.go
@@ -2,14 +2,19 @@ package integration
 
 import (
 	"fmt"
+	"os/exec"
 	"path/filepath"
+	"runtime"
+	"strings"
 	"testing"
 
+	svcAutostart "github.com/ActiveState/cli/cmd/state-svc/autostart"
 	anaConst "github.com/ActiveState/cli/internal/analytics/constants"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/environment"
 	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/osutils"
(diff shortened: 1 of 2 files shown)


test/integration/remote_installer_int_test.go
```diff
@@ -25,6 +30,25 @@ func (suite *RemoteInstallIntegrationTestSuite) TestInstall() {
...
+			suite.Require().NoError(err, "error getting policy: "+string(policy))
+			return strings.TrimSpace(string(policy))
+		}
+		setPolicy := func(policy string) {
+			output, err := exec.Command("powershell.exe", "Set-ExecutionPolicy", "-ExecutionPolicy", policy).CombinedOutput()
+			suite.Require().NoError(err, "error setting policy: "+string(output))
+		}
+
+		policy := getPolicy()
+		defer setPolicy(policy)
```
MDrakos, line 46: Could we make this a defer func and ensure the policy is set back to the default once we're done. Just to make sure it doesn't affect other tests.

test/integration/remote_installer_int_test.go
```diff
@@ -25,6 +30,25 @@ func (suite *RemoteInstallIntegrationTestSuite) TestInstall() {
 	ts := e2e.New(suite.T(), false)
 	defer ts.Close()
 
+	// Setup running the remote installer in restricted powershell mode.
+	if runtime.GOOS == "windows" {
+		getPolicy := func() string {
+			policy, err := exec.Command("powershell.exe", "Get-ExecutionPolicy").CombinedOutput()
```
MDrakos, line 36: Would using `ts.SpawnCmd` make this only affect the current shell? I imagine it wouldn't because this is likely a system-wide setting...

MDrakos (changes requested):
Sorry for the delay on the review.

test/integration/remote_installer_int_test.go
mitchell-as: I don't understand. I am deferring resetting the policy to its default. Line 45 fetches the current default policy, and line 46 defers setting it to that default when we're done.

test/integration/remote_installer_int_test.go
mitchell-as: It's a system-wide setting. I could play around with flags passed to `Set-ExecutionPolicy` to maybe affect the current session, but that doesn't reflect the nature of the test, which is to run in a restricted environment (presumably system-wide).

test/integration/remote_installer_int_test.go
MDrakos: What I mean is something like this:
```
defer func() {
	setPolicy(policy)
	suite.Assert().Equal(policy, getPolicy(), "should have set powershell policy to '"+policy+"'")
}()
```

Just to _ensure_ the policy has been reverted. If you feel this is unnecessary or overkill that's fine.

mitchell-as: Okay, thanks for clarifying. I've added your suggested assert.

MDrakos (approved):

PR #3552, opened 2024-10-21
Fix merge integration test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3115" title="DX-3115" target="_blank">BugDX-3115</a>  Nightly failure: TestPullIntegrationTestSuite/TestMergeBuildScript
  </td></table>
  <br />

### DIFF
--- a/internal/runners/install/install.go
+++ b/internal/runners/install/install.go
@@ -114,6 +114,7 @@ func (i *Install) Run(params Params) (rerr error) {
 	var oldCommit *bpModel.Commit
 	var reqs requirements
 	var ts time.Time
+	var override bool
 	{
 		pg = output.StartSpinner(out, locale.T("progress_search"), constants.TerminalAnimationInterval)
 
@@ -133,6 +134,7 @@ func (i *Install) Run(params Params) (rerr error) {
 		if err != nil {
 			return errs.Wrap(err, "Unable to get timestamp from params")
 		}
+		override = params.Timestamp.String() != ""
 
 		// Get languages used in current project
 		languages, err := model.FetchLanguagesForBuildScript(oldCommit.BuildScript())
@@ -153,7 +155,7 @@ func (i *Install) Run(params Params) (rerr error) {
 
 	// Prepare updated buildscript
 	script := oldCommit.BuildScript()
-	if err := prepareBuildScript(script, reqs, ts); err != nil {
+	if err := prepareBuildScript(script, reqs, ts, override); err != nil {
 		return errs.Wrap(err, "Could not prepare build script")
 	}
 
@@ -340,8 +342,8 @@ func (i *Install) renderUserFacing(reqs requirements) {
 	i.prime.Output().Notice("")
 }
 
-func prepareBuildScript(script *buildscript.BuildScript, requirements requirements, ts time.Time) error {
-	script.SetAtTime(ts)
+func prepareBuildScript(script *buildscript.BuildScript, requirements requirements, ts time.Time, override bool) error {
+	script.SetAtTime(ts, override)
 	for _, req := range requirements {
 		requirement := types.Requirement{
 			Namespace:          req.Resolved.Namespace,
--- a/internal/runners/install/install.go
+++ b/internal/runners/install/install.go
@@ -114,7 +114,6 @@ func (i *Install) Run(params Params) (rerr error) {
 	var oldCommit *bpModel.Commit
 	var reqs requirements
 	var ts time.Time
-	var override bool
 	{
 		pg = output.StartSpinner(out, locale.T("progress_search"), constants.TerminalAnimationInterval)
 
@@ -134,7 +133,6 @@ func (i *Install) Run(params Params) (rerr error) {
 		if err != nil {
 			return errs.Wrap(err, "Unable to get timestamp from params")
 		}
-		override = params.Timestamp.String() != ""
 
 		// Get languages used in current project
 		languages, err := model.FetchLanguagesForBuildScript(oldCommit.BuildScript())
@@ -155,7 +153,7 @@ func (i *Install) Run(params Params) (rerr error) {
 
 	// Prepare updated buildscript
 	script := oldCommit.BuildScript()
-	if err := prepareBuildScript(script, reqs, ts, override); err != nil {
+	if err := prepareBuildScript(script, reqs, ts); err != nil {
 		return errs.Wrap(err, "Could not prepare build script")
 	}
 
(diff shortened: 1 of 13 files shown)


pkg/platform/model/buildplanner/buildscript.go
```diff
@@ -52,10 +52,10 @@ func (b *BuildPlanner) GetBuildScript(commitID string) (*buildscript.BuildScript
 	}
 
 	script := buildscript.New()
-	script.SetAtTime(time.Time(resp.Commit.AtTime))
 	if err := script.UnmarshalBuildExpression(resp.Commit.Expression); err != nil {
 		return nil, errs.Wrap(err, "failed to parse build expression")
 	}
+	script.SetAtTime(time.Time(resp.Commit.AtTime))
```
MDrakos, line 58: This update is to follow the same structure as here: https://github.com/ActiveState/cli/blob/master/pkg/platform/model/buildplanner/build.go#L113

It appears we want to the buildscript `at_time` to be the commit `at_time` even if the buildscript has it's own time set.

Naatan (changes requested):
I think we should fix this issue closer to the root. The issue here is that UnmarshalBuildExpression is apparently overriding the AtTime. We should fix that.

We might still be setting at time in the raw representation, we should move that out of raw and just set it on the buildscript struct itself. Raw effectively represents the build expression, which does not contain the at time at all.

mitchell-as:
There was a time before we had `"$at_time"` in build expressions. In those days, it was a timestamp. When we find a timestamp, we should be using it, not overriding it with a later atTime. I don't agree with the latest change, but maybe I'm missing something.

Naatan (changes requested):
Reading the code now, I realize that this should just be an update to our testing code. The reason the test was failing is we were essentially testing how we want things to work with new buildscripts, but it was failing cause it was dealing with an old builscript. The solution with old buildscripts was appropriate: Take the at_time value, and record it on the buildscript level. Next time the user commits either it is bumped or at the very least it will be recorded on the commit level.

So sorry for going back on what we discussed during stand; but I think the fix here ought to be no code changes to anything but the integration test. And the integration test fix in theory would just be to bump the time stamp on the commit that it starts from for the comparison (ie. so it uses a build expression without the at time value hardcoded on it).

MDrakos:
From our discussion the following points are how we want to handle the `at_time` value along with how this solution _should_ address them.
- We respect the hardcoded `at_time` value
	- This is done via storing the `at_time` found in the solve node separately from the preferred time value.
	- We return the commit `at_time` if that was set externally, otherwise we default to the solve at time.
	- When we unmarshall the build expression the solve `at_time` could either be a time stamp or a reference.
- BUT; any future commits we make transfer it to the commit level and use a variable instead
	- This is already handled by our current implementation
	- The `buildscript` library updates the `at_time` in the solve node to be a reference if it finds a timestamp there.
	- When we make a commit we supply the `at_time` as part of the input parameters which sets the commit's `at_time`
- If `at_time` is already used as a variable then we don't do anything special other than respect the commit at time and not touch the build expression value
	- This is done by updating the `SetAtTime()` and `AtTime()` methods to work with the preferred (commit) `atTime` and not the one found in the solve node if set.
	- `SetAtTime()` is currently being called when we unmarshall the buildexpression from a commit response which sets the `atTime` value to the commit at time.
- I as a developer shouldn't need to think about this when using the buildscript library externally
	- Since `SetAtTime()` is working with a separate value than the solve node external callers of the library can be assured that calling `SetAtTime()` with a time value will be preserved and returned when calling `AtTime()`

Naatan (changes requested):
I don't think this is meeting the criteria:

* We respect the hardcoded at_time value

Because we are always returning `atTime` when it is set. Which will always be the case. I think you are not seeing this in your tests because you are likely still setting atTime before unmarshalling.

Essentially the following test should pass (pseudo code):

```
func TestLegacyAtTime(t *testing.T) {
  bs := newBuildScript()
  err := bs.UnmarshalBuildExpression(bexp)
  require.NoError(t, err)
  err := bs.SetAtTime(bogusValue)
  require.ErrorAs(err, ErrAtTimeAlreadySet)
  require.Equal(t, bs.AtTime(), buildExpressionAtTime)
}
```

So that would be asserting that the build expression atTime is preserved and overrides the commit atTime.

The error check on SetAtTime is just an in-the-moment solution to the problem of "I as a developer do not need special knowledge". Because when working with legacy atTime values it straight up isn't appropriate to override the at time with that of the commit. I think erroring out in this case is ok, because then at least the failure will be obvious to the developer and they can then handle it better (eg. ignore that type of error and move on).

MDrakos:
@Naatan You're right I misinterpreted that point. I didn't realize we want to retain the solve node's `atTime`. Going through the code there appears to be some cases where we want to override the `atTime` like when we pass the `--ts` flag or when merging. I've added the ability to specify overriding the time value but we may want to go with a different pattern.

internal/runners/install/install.go
```diff
@@ -114,6 +114,7 @@ func (i *Install) Run(params Params) (rerr error) {
 	var oldCommit *bpModel.Commit
 	var reqs requirements
 	var ts time.Time
+	var override bool
```
Naatan, line 117: I think for the install action we always want to override.

internal/runners/platforms/add.go
```diff
@@ -98,7 +98,7 @@ func (a *Add) Run(params AddRunParams) (rerr error) {
 
 	// Prepare updated buildscript
 	script := oldCommit.BuildScript()
-	script.SetAtTime(ts)
+	script.SetAtTime(ts, false)
```
Naatan, line 101:
```suggestion
	script.SetAtTime(ts, true)
```

We're making a change, which by definition means we're bumping the atTime (unless the current atTime is higher than the latest, but that's handled by the Expand function).

internal/runners/upgrade/upgrade.go
```diff
@@ -111,7 +111,7 @@ func (u *Upgrade) Run(params *Params) (rerr error) {
 	if err != nil {
 		return errs.Wrap(err, "Failed to fetch latest timestamp")
 	}
-	bumpedBS.SetAtTime(ts)
+	bumpedBS.SetAtTime(ts, params.Timestamp.String() != "")
```
Naatan, line 114:
```suggestion
	bumpedBS.SetAtTime(ts, true)
```

pkg/buildscript/buildscript.go
```diff
@@ -56,7 +56,10 @@ func (b *BuildScript) AtTime() *time.Time {
 	return b.atTime
 }
 
-func (b *BuildScript) SetAtTime(t time.Time) {
+func (b *BuildScript) SetAtTime(t time.Time, override bool) {
```
Naatan, line 59:
```suggestion
// SetAtTime sets the AtTime value, if the buildscript already has an AtTime value  and `override=false` then the value passed here will be discarded.
// Override should in most cases only be true if we are making changes to the buildscript.
func (b *BuildScript) SetAtTime(t time.Time, override bool) {
```

Naatan (changes requested):

internal/runners/install/install.go
MDrakos: Ah okay, so anytime we make a change we want to update the timestamp? I was under the impression that we want to preserve the timestamp in the solve node and I was only bumping it when the user explicitly passed the `--ts` flag.

Naatan: We only want to preserve it when we retrieve the commit. Once we're acting on the buildscript (ie. making changes) the existing timestamp should always be overwritten.

Naatan (approved):

PR #3553, opened 2024-10-22
Runtime envdef handles case-insensitive environment variable names on Windows.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3030" title="DX-3030" target="_blank">TaskDX-3030</a>  Handle/merge PATH case insensitively on Windows
  </td></table>
  <br />

### DIFF
--- a/pkg/runtime/internal/envdef/environment.go
+++ b/pkg/runtime/internal/envdef/environment.go
@@ -395,8 +395,10 @@ func (ed *EnvironmentDefinition) getEnvBasedOn(envLookup map[string]string) (map
 		if len(ev.Values) > 0 {
 			res[pev.Name] = pev.ValueString()
 			if pev.Name != osName {
-				// On Windows, delete the case-insensitive version. Our case-sensitive version has already
-				// processed the value of the case-insensitive version.
+				// On Windows, delete the redundant (case-insensitive) version that our case-sensitive
+				// version could conflict with. (Our version has already processed the value of the
+				// redundant version.)
+				// For example, delete "Path" while preserving our "PATH".
 				delete(res, osName)
 			}
 		}
--- a/pkg/runtime/internal/envdef/environment.go
+++ b/pkg/runtime/internal/envdef/environment.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/ActiveState/cli/internal/constants"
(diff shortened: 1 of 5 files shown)


mitchell-as:
Failures are unrelated to this PR.

pkg/runtime/internal/envdef/environment.go
```diff
@@ -344,35 +345,60 @@ func (ev *EnvironmentVariable) ValueString() string {
...
+
 	for _, ev := range ed.Env {
 		pev := &ev
+		osName := pev.Name
+		if runtime.GOOS == "windows" {
+			if name, ok := osEnvNames[strings.ToLower(osName)]; ok {
+				osName = name
+			}
+		}
+		osValue, hasOsValue := envLookup[osName]
```
Naatan, line 380: Can you rename this to `osKey` rather than `Value`? Since that is what this is tracking, and since this function does in fact also deal with values. It makes it a bit error prone to grok. Also can we link this up in naming with `osName` so it's clear one is an override for the other? Maybe `osKey` and `osOverrideKey` or something like that.

I stumbled trying to understand this code, both here as well as further below on line 397 which implies the code here, but due to the naming it took me a while to fully grok.

I realize some of this code was already there, but still, worth addressing.

Naatan (changes requested):
In addition to the comments, can we make sure there is a test case covering this mechanic (specifically: PATH and Path on Windows).

pkg/runtime/internal/envdef/environment.go
```diff
@@ -344,35 +345,60 @@ func (ev *EnvironmentVariable) ValueString() string {
...
 			res[pev.Name] = "" // unset
 		}
 		// only add environment variable if at least one value is set (This allows us to remove variables from the environment.)
 		if len(ev.Values) > 0 {
 			res[pev.Name] = pev.ValueString()
+			if pev.Name != osName {
+				// On Windows, delete the case-insensitive version. Our case-sensitive version has already
+				// processed the value of the case-insensitive version.
+				delete(res, osName)
+			}
```
Naatan, line 401:
```suggestion
			if pev.Name != osName {
				// Delete the redundant value that ours could conflict with
				// Eg. this would delete "Path" while preserving our "PATH"
				delete(res, osName)
			}
```

Naatan (changes requested):

Naatan (approved):

PR #3554, opened 2024-10-23
Re-enable CheckoutFromArchive test.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3057" title="DX-3057" target="_blank">BugDX-3057</a>  TestCheckoutFromArchive uses bad mock
  </td></table>
  <br />
 

See comment here about how the test archives were originally constructed: https://activestatef.atlassian.net/browse/DX-3057?focusedCommentId=56035. They were not being mocked. It's real-world data.


pkg/buildplan/filters.go
```diff
@@ -44,9 +46,12 @@ const NamespaceInternal = "internal"
...
 	return func(a *Artifact) bool {
-		internalIngredients := sliceutils.Filter(a.Ingredients, func(i *Ingredient) bool {
-			return i.Namespace == NamespaceInternal
-		})
+		internalIngredients := []*Ingredient{}
+		if os.Getenv(constants.InstallInternalDependenciesEnvVarName) != "true" {
+			internalIngredients = sliceutils.Filter(a.Ingredients, func(i *Ingredient) bool {
+				return i.Namespace == NamespaceInternal
+			})
+		}
```
mitchell-as, line 54: The checkout archive installs a `shared/zlib` package, which would be excluded by default. In order to keep things small and simple, add an env var override to avoid excluding internal packages.

Naatan (changes requested):
This doesn't feel appropriate to me. It seems the issue here is that this test was asserting something as installed which as it turns out shouldn't have been installed, that in itself was a bug that it seems we indirectly addressed at one point.

The solution here is to update the test code, we shouldn't need to introduce special logic in the buildplan for this. Can't we update the project we generate data from to use a real dependency that we can then assert?

PR #3556, opened 2024-10-25
Use fixed timestamp when asserting package versions.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3128" title="DX-3128" target="_blank">BugDX-3128</a>  Nightly failure: TestPackageIntegrationTestSuite/TestChangeSummaryShowsAddedForUpdate
  </td></table>
  <br />


MDrakos (approved):
This is a better solution than constantly updating the dependency version 👍

PR #3559, opened 2024-10-25
Use runtime sourcing timeout opt to fix failing test.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3131" title="DX-3131" target="_blank">BugDX-3131</a>  Nightly failure: TestActivateIntegrationTestSuite/TestActivateBranch
  </td></table>
  <br />

### DIFF
--- a/test/integration/activate_int_test.go
+++ b/test/integration/activate_int_test.go
@@ -582,11 +582,8 @@ func (suite *ActivateIntegrationTestSuite) TestActivateBranch() {
 
 	namespace := "ActiveState-CLI/Branches"
 
-	cp := ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
-	cp.ExpectExitCode(0)
-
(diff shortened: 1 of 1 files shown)


MDrakos (changes requested):
We're not doing anything with the runtime in this test. It appears that the issue is that it's downloading a runtime even when the async runtime config value is set?

mitchell-as:
This test asserts that the branch is available and activated. (A subsequent test asserts we throw an error if a branch does not exist.) Yes, we're not doing anything after activating, but activating in and of itself is the test.

With regard to async runtimes, we had a recent ticket where we defined that only a select few commands can source a runtime, and activate is one of them. We shouldn't be setting async runtimes in activation tests, as it's meaningless now.

This test is failing only on Windows where it times out, so we need the runtime sourcing timeout.

MDrakos (approved):
Thank you for clearing that up! Just wanted to be sure I understood the change.

PR #3566, opened 2024-10-28
Localize an access denied error when attempting to migrate an activestate.yaml.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3137" title="DX-3137" target="_blank">BugDX-3137</a>  Rollbar: MUST ADDRESS: Error does not have localization: Could not parse projectfile
  </td></table>
  <br />


mitchell-as:
Good call.

Naatan (approved):

PR #3568, opened 2024-10-29
Do not log JSON output pipe errors to rollbar.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3135" title="DX-3135" target="_blank">BugDX-3135</a>  Rollbar: projects [-o]: Could not write json output, error: 'write /dev/stdout: The pipe is being closed.': The pipe is being closed.
  </td></table>
  <br />
 

The system is hanging up on us and there is nothing we can do about it.

### DIFF
--- a/internal/output/json.go
+++ b/internal/output/json.go
@@ -109,7 +109,7 @@ func (f *JSON) Error(value interface{}) {
 
 func isPipeClosedError(err error) bool {
 	pipeErr := errors.Is(err, syscall.EPIPE)
-	if runtime.GOOS == "windows" && errors.Is(err, syscall.Errno(242)) {
+	if runtime.GOOS == "windows" && errors.Is(err, syscall.Errno(232)) {
 		// Note: 232 is Windows error code ERROR_NO_DATA, "The pipe is being closed".
 		// See https://go.dev/src/os/pipe_test.go
 		pipeErr = true
--- a/internal/output/json.go
+++ b/internal/output/json.go
@@ -2,8 +2,11 @@ package output
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"runtime"
+	"syscall"
 
 	"github.com/ActiveState/cli/internal/colorize"
 	"github.com/ActiveState/cli/internal/locale"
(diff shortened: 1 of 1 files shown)


internal/output/json.go
```diff
@@ -92,8 +99,22 @@ func (f *JSON) Error(value interface{}) {
...
+			logging.Error("Could not write json output, error: %v", err) // do not log to rollbar
+		} else {
+			multilog.Error("Could not write json output, error: %v", err)
+		}
+	}
+}
+
+func isPipeClosedError(err error) bool {
+	pipeErr := errors.Is(err, syscall.EPIPE)
+	if runtime.GOOS == "windows" && errors.Is(err, syscall.Errno(242)) {
```
MDrakos, line 112: As long as the comment is correct which I think it is.
```suggestion
	if runtime.GOOS == "windows" && errors.Is(err, syscall.Errno(232)) {
```

internal/output/json.go
```diff
@@ -92,8 +99,22 @@ func (f *JSON) Error(value interface{}) {
...
+		if isPipeClosedError(err) {
+			logging.Error("Could not write json output, error: %v", err) // do not log to rollbar
+		} else {
+			multilog.Error("Could not write json output, error: %v", err)
+		}
+	}
+}
+
+func isPipeClosedError(err error) bool {
+	pipeErr := errors.Is(err, syscall.EPIPE)
```
MDrakos, line 111: Should we also check for `io.ErrClosedPipe`?

internal/output/json.go
```diff
@@ -92,8 +99,22 @@ func (f *JSON) Error(value interface{}) {
...
+	if runtime.GOOS == "windows" && errors.Is(err, syscall.Errno(242)) {
+		// Note: 232 is Windows error code ERROR_NO_DATA, "The pipe is being closed".
```
MDrakos, line 113: Do we know if the Go library uses error code `109` as well? It is `ERROR_BROKEN_PIPE`

https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-

MDrakos (changes requested):

internal/output/json.go
mitchell-as: 🤦 Thanks

internal/output/json.go
mitchell-as: I don't think so, because the error message is "io: read/write on closed pipe" (https://pkg.go.dev/io#pkg-variables). That's not consistent with the rollbar error message we are seeing. I feel like we want to be as narrow as possible right now for this ticket since we're turning off rollbar logging in this case. If we start seeing this, then we can revisit.

internal/output/json.go
mitchell-as: I don't know. Regardless, that error message is "The pipe has been ended." As mentioned previously, I think we want to be as narrow as possible right now when turning off rollbar logging. I haven't seen that error message in rollbar yet.

MDrakos (approved):

PR #3575, opened 2024-10-31
Fixes and tests for ingredientcall end-to-end

Naatan:


cmd/state-svc/internal/hash/file_hasher.go
```diff
@@ -32,15 +34,16 @@ type hashedFile struct {
...
 
 func (fh *FileHasher) HashFiles(wd string, globs []string) (_ string, _ []hashedFile, rerr error) {
-	sort.Strings(globs) // ensure consistent ordering
+	fs := os.DirFS(wd)
 	hashedFiles := []hashedFile{}
-	hasher := xxhash.New()
+	hashes := []string{}
 	for _, glob := range globs {
-		files, err := filepath.Glob(glob)
+		files, err := doublestar.Glob(fs, glob)
```
Naatan, line 46: `filepath.Glob()` does not support double star globs (ie. `*/**`) nor specifying the working directory. This library addresses both limitations.

cmd/state-svc/internal/hash/file_hasher.go
```diff
@@ -53,47 +56,61 @@ func (fh *FileHasher) HashFiles(wd string, globs []string) (_ string, _ []hashed
...
+	if hashedFiles == nil {
+		return "", nil, nil
+	}
+
+	// Ensure the overall hash is consistently calculated
+	sort.Slice(hashedFiles, func(i, j int) bool { return hashedFiles[i].Path < hashedFiles[j].Path })
+	h := xxhash.New()
+	for _, f := range hashedFiles {
+		fmt.Fprintf(h, "%016x", f.Hash)
+	}
```
Naatan, line 111: Changed it to calculate the overall hash based on a sorted slice of all the individual hashes for all files. This way we get the same hash even if the files are sorted differently, or more importantly (what I ran into); if the patterns are different but produce the same result.

internal/archiver/archiver.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/mholt/archiver/v3"
+)
+
+type FileMap struct {
+	Source string
+	Target string
+}
+
+func CreateTgz(filepath string, fileMaps []FileMap) error {
```
Naatan, line 17: Archiver supports creating archives, but its API is pointlessly convoluted. This package is just a wrapper that doesn't require us to write so much boilerplate just to create a simple archive.

test/integration/buildscript_int_test.go
```diff
@@ -39,6 +48,124 @@ func (suite *BuildScriptIntegrationTestSuite) TestBuildScript_NeedsReset() {
 	suite.Require().FileExists(filepath.Join(ts.Dirs.Work, constants.BuildScriptFileName), ts.DebugMessage(""))
 }
 
+func (suite *BuildScriptIntegrationTestSuite) TestBuildScript_IngredientFunc() {
+	suite.T().Skip("Please enable once ingredient publishing via buildscript on the API is live")
```
MDrakos, line 52: Could we link this to a story? Even if it's a while before that story gets resolved.

pkg/buildscript/unmarshal_buildexpression.go
```diff
@@ -172,7 +175,7 @@ func unmarshalValue(path []string, valueInterface interface{}) (*value, error) {
 		result.List = &values
 
 	case string:
-		if sliceutils.Contains(path, ctxIn) || strings.HasPrefix(v, "$") {
+		if len(path) >= 2 && path[len(path)-2] == ctxIn || strings.HasPrefix(v, "$") {
```
MDrakos, line 178: 😬 We should probably prioritize the story to revisit the `path` value.

MDrakos (changes requested):
Looks really good, just a few minor things to consider.

internal/fileutils/fileutils_test.go
```diff
@@ -634,3 +634,68 @@ func TestIsWritableDir(t *testing.T) {
 		t.Fatalf("Path should not be writable: %s", pathWithNoPermission)
 	}
 }
+
+func TestCommonParentPath(t *testing.T) {
```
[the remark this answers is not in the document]
Naatan: Good call! 👍

pkg/buildscript/unmarshal_buildexpression.go
Naatan: I'll clean it up a little. At the moment no story is planned, since Mitchell said it's otherwise all working as intended, this is a one off bug.

That said we will surely be refactoring this library as a whole once buildscripts are more stable.

pkg/buildscript/unmarshal_buildexpression.go
```diff
@@ -172,7 +175,8 @@ func unmarshalValue(path []string, valueInterface interface{}) (*value, error) {
 		result.List = &values
 
 	case string:
-		if sliceutils.Contains(path, ctxIn) || strings.HasPrefix(v, "$") {
+		parentNode, hasParentNode := sliceutils.GetString(path, -2)
```
MDrakos, line 178: Nice 👍

MDrakos (dismissed):

internal/archiver/archiver.go
```diff
@@ -0,0 +1,73 @@
...
+	tgz := archiver.NewTarGz()
+	if err := tgz.Create(f); err != nil {
+		return errs.Wrap(err, "Could not create tar.gz")
+	}
+	defer tgz.Close()
+
+	for _, fileMap := range fileMaps {
+		source := fileMap.Source
+		if !filepath.IsAbs(source) {
+			source = filepath.Join(workDir, source)
```
MDrakos, line 33: From reading the code in this file it's not clear to me why we're appending the source path to the `workDir` rather than just using the absolute path here.

MDrakos (approved):
Looks good to me, just one comment to consider.

internal/archiver/archiver.go
Naatan: I don't follow, this is what is making the path absolute. Are you seeing another absolute path value that I can use?

MDrakos: Ahh okay, maybe a comment on the `FileMap` struct would help here? I was expecting to see a call to `filepath.Abs` if the path isn't absolute but if that's how the file map is constructed it makes sense as well.

Naatan: `filepath.Abs()` I believe works with the working directory, which is a global so something we want to avoid. I'll add a comment to clarify. Fwiw the reason the paths are relative is to facilitate archiving, otherwise the archives also have absolute paths in them (ie. rather than having setup.py at the root of the archive it'd be at `c:\Users\Runner~1\` etc. (oddly specific example that totally does not indicate I struggled with this..).

MDrakos (approved):

PR #3576, opened 2024-10-31
Initial support for rendering wrapped, bulleted lists.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2829" title="DX-2829" target="_blank">TaskDX-2829</a>  Failure messages remain indented with line breaks
  </td></table>
  <br />
 

Also do not put wrapped links on their own line. Links are not wrapped and indented because they would not be copy-pastable into browsers without manual removal of spaces and/or newlines.

Screenshot 2024-11-01 at 6 07 26 PM

### DIFF
--- a/internal/colorize/wrap.go
+++ b/internal/colorize/wrap.go
@@ -4,8 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
-
-	"github.com/ActiveState/cli/internal/logging"
+	"unicode/utf8"
 )
 
 type WrappedLines []WrappedLine
@@ -17,8 +16,8 @@ type WrappedLine struct {
 
 func (c WrappedLines) String() string {
 	var result string
-	for _, crop := range c {
-		result = result + crop.Line
+	for _, wrapped := range c {
+		result = result + wrapped.Line
 	}
 
 	return result
@@ -28,43 +27,56 @@ var indentRegexp = regexp.MustCompile(`^([ ]+)`)
 var isLinkRegexp = regexp.MustCompile(`\s*(\[[^\]]+\])?https?://`)
 
 func Wrap(text string, maxLen int, includeLineEnds bool, continuation string) WrappedLines {
+	// Determine indentation of wrapped lines based on any leading indentation.
 	indent := ""
 	if indentMatch := indentRegexp.FindStringSubmatch(text); indentMatch != nil {
 		indent = indentMatch[0]
 		if len(text) > len(indent) && strings.HasPrefix(text[len(indent):], "• ") {
+			// If text to wrap is a bullet item, indent extra to be flush after bullet.
 			indent += "  "
 		}
 	}
 
-	maxLen -= len(continuation)
+	// If wrapping includes continuation text, reduce maximum wrapping length accordingly.
+	maxLen -= utf8.RuneCountInString(StripColorCodes(continuation))
 
 	entries := make([]WrappedLine, 0)
 	colorCodes := colorRx.FindAllStringSubmatchIndex(text, -1)
 	colorNames := colorRx.FindAllStringSubmatch(text, -1)
 
-	isLineEnd := false
 	entry := WrappedLine{}
+	// Iterate over the text, one character at a time, and construct wrapped lines while doing so.
 	for pos, amend := range text {
-		inColorTag := inRange(pos, colorCodes)
+		isLineEnd := amend == '\n'
+		isTextEnd := pos == len(text)-1
 
-		isLineEnd = amend == '\n'
-
+		// Add the current character to the wrapped line.
+		// Update the wrapped line's length as long as the added character is not part of a tag like
+		// [ERROR] or [/RESET].
 		if !isLineEnd {
 			entry.Line += string(amend)
-			if !inColorTag {
+			if !inRange(pos, colorCodes) {
 				entry.Length++
 			}
 		}
+		atWrapPosition := entry.Length == maxLen
 
-		// Ensure the next position is not within a color tag and check conditions that would end this entry
-		if isLineEnd || (!inRange(pos+1, colorCodes) && (entry.Length == maxLen || pos == len(text)-1)) {
-			wrapped := ""
-			wrappedLength := len(indent)
-			nextCharIsSpace := pos+1 < len(text) && isSpace(text[pos+1])
-			if !isLineEnd && entry.Length == maxLen && !nextCharIsSpace && pos < len(text)-1 {
-				// Put the current word on the next line, if possible.
-				// Find the start of the current word and its printed length, taking color ranges and
-				// multi-byte characters into account.
+		// When we've reached the end of the line, either naturally (line or text end), or when we've
+		// reached the wrap position (maximum length), we need to wrap the current word (if any) and
+		// set up the next (wrapped) line.
+		// Note that if we've reached the wrap position but there's a tag immediately after it, we want
+		// to include the tag, so do not wrap in that case.
+		if isLineEnd || isTextEnd || (atWrapPosition && !inRange(pos+1, colorCodes)) {
+			wrapped := ""                // the start of the next (wrapped) line
+			wrappedLength := len(indent) // the current length of the next (wrapped) line
+
+			// We need to prepare the next (wrapped) line unless we're at line end, text end, not at wrap
+			// position, or the next character is a space (i.e. no wrapping needed).
+			nextCharIsSpace := !isTextEnd && isSpace(text[pos+1])
+			if !isLineEnd && !isTextEnd && atWrapPosition && !nextCharIsSpace {
+				// Determine the start of the current word along with its printed length (taking color
+				// ranges and multi-byte characters into account).
+				// We need to know these things in order to put it on the next line (if possible).
 				i := len(entry.Line) - 1
 				for ; i > 0; i-- {
 					if isSpace(entry.Line[i]) {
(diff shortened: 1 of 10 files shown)


Naatan (changes requested):
First review based on screenshot alone, I'll review the code once we get it looking right so we're not churning on code that's still gonna be changed.

- The indented message seems to be missing one space 
- The indentation marks should not be given the color of the message (ie. they should remain black/default regardless of the message color).

I recognize the second point might require some significant business logic. Please move it back to triage if this creates a lot of additional work. This isn't worth a ton of investment, but we also shouldn't land the change as is since it's just trading one visual bug for another.

internal/colorize/wrap_test.go
```diff
@@ -7,82 +7,82 @@ import (
 	"testing"
 )
 
-func Test_GetCroppedText(t *testing.T) {
```
Naatan, line 10: Please add tests covering the new mechanic. Particularly color codes not covering the "continuation".

internal/colorize/wrap.go
```diff
@@ -100,6 +115,19 @@ func inRange(pos int, ranges [][]int) bool {
 	return false
 }
 
+// colorTag returns the currently active color tag (if any) at the given position.
+func colorTag(pos int, ranges [][]int, names [][]string) string {
```
Naatan, line 119: This assumes there's only ever one color tag, but we could have: `[BOLD][RED]`, for example.

internal/colorize/wrap.go
```diff
@@ -69,8 +76,16 @@ func GetCroppedText(text string, maxLen int, includeLineEnds bool) CroppedLines
...
+				if i > 0 && i < len(entry.Line)-1 && !isLinkRegexp.MatchString(entry.Line[i:]) {
+					tag := colorTag(pos, colorCodes, colorNames)
+					if continuation != "" && tag != "" {
+						// Do not colorize the continuation.
+						wrapped = fmt.Sprintf("%s[/RESET]%s%s%s", indent, continuation, tag, entry.Line[i:])
+					} else {
+						wrapped = indent + continuation + entry.Line[i:]
+					}
+					logging.Debug("continuation: '%s'", continuation)
+					logging.Debug("wrapped: '%s'", wrapped)
```
Naatan, line 88:
```suggestion
```

internal/output/renderers/bulletlist.go
```diff
@@ -0,0 +1,89 @@
...
+// printer.
+func NewBulletList(prefix string, bullets, items []string) *bulletList {
+	if len(bullets) != 4 {
+		multilog.Error("Invalid bullet list: 4 bullets required")
+		bullets = BulletTree
+	}
+	return &bulletList{prefix, items, bullets}
+}
+
+func (b *bulletList) MarshalOutput(format output.Format) interface{} {
```
Naatan, line 48: Unless this creates some deep rabbit hole I'd prefer we don't address the magical output handling that we're slowly phasing out. So this could just be a `String()` function, or whatever you feel is appropriate so long as it's not trying to satisfy the output interface.

That also means that use-cases of this functionality will need to handle structured vs non-structured themselves. That part might be a rabbit hole, depending on the use-cases we have..

Naatan: Also worth having a unit test for this.

internal/output/renderers/bulletlist.go
```diff
@@ -0,0 +1,89 @@
...
+// The set of bullets should contain four items: the bullet for the first item (e.g. ""); the
+// bullet for each subsequent item (e.g. "├─"); the bullet for an item's wrapped lines, if any
+// (e.g. "│"); and the bullet for the last item (e.g. "└─").
+// The returned list can be passed to a plain printer. It should not be passed to a structured
+// printer.
+func NewBulletList(prefix string, bullets, items []string) *bulletList {
+	if len(bullets) != 4 {
+		multilog.Error("Invalid bullet list: 4 bullets required")
+		bullets = BulletTree
+	}
```
Naatan, line 44: Should return the error.

internal/output/renderers/bulletlist.go
```diff
@@ -0,0 +1,89 @@
...
+	prefix  string
+	items   []string
+	bullets []string
+}
+
+// BulletTree outputs a list like:
+//
+// ├─ one
+// ├─ two
+// │ wrapped
```
Naatan, line 22:
```suggestion
// ├─ two
// │  wrapped
```

internal/output/renderers/bulletlist.go
```diff
@@ -0,0 +1,89 @@
...
+// └─ three
+var BulletTree = []string{output.TreeMid, output.TreeMid, output.TreeLink + " ", output.TreeEnd}
+
+// HeadedBulletTree outputs a list like:
+//
+// one
+// ├─ two
+// │ wrapped
```
Naatan, line 30:
```suggestion
// ├─ two
// │  wrapped
```

internal/output/renderers/bulletlist.go
```diff
@@ -0,0 +1,89 @@
...
+	bullets []string
+}
+
+// BulletTree outputs a list like:
+//
+// ├─ one
+// ├─ two
+// │ wrapped
+// └─ three
+var BulletTree = []string{output.TreeMid, output.TreeMid, output.TreeLink + " ", output.TreeEnd}
```
Naatan, line 24: Since we're using presets anyway, can we make this typed so it's easier to understand the code while reading?

eg.

```
type Bullets {
  Start string
  Mid string
  Link string
  End string
}
```

internal/runbits/dependencies/changesummary.go
```diff
@@ -130,8 +127,9 @@ func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan,
 			item = fmt.Sprintf("[ACTIONABLE]%s@%s[/RESET] → %s (%s)", oldVersion.Name, oldVersion.Version, item, locale.Tl("updated", "updated"))
 		}
 
-		out.Notice(fmt.Sprintf("  [DISABLED]%s[/RESET] %s", prefix, item))
```
Naatan, line 133: This seems to be dropping the `[DISABLED]`.

internal/runbits/dependencies/changesummary.go
```diff
@@ -130,8 +127,9 @@ func OutputChangeSummary(out output.Outputer, newBuildPlan *buildplan.BuildPlan,
...
+		items[i] = item
 	}
 
+	out.Notice(renderers.NewBulletList("  ", renderers.BulletTree, items))
```
Naatan, line 133: Fwiw, with regards to dropping the magical output handing, this could simply become:

```
	out.Notice(renderers.NewBulletList("  ", renderers.BulletTree, items).String())
```

internal/runbits/dependencies/summary.go
```diff
@@ -44,10 +41,9 @@ func OutputSummary(out output.Outputer, directDependencies buildplan.Artifacts)
 			subdepLocale = locale.Tl("summary_subdeps", "([ACTIONABLE]{{.V0}}[/RESET] sub-dependencies)", strconv.Itoa(numSubs))
 		}
 
-		item := fmt.Sprintf("[ACTIONABLE]%s@%s[/RESET] %s", ingredient.Name, ingredient.Version, subdepLocale)
-
-		out.Notice(fmt.Sprintf("[DISABLED]%s[/RESET] %s", prefix, item))
```
Naatan, line 49: Why drop the DISABLED item here?

Naatan (changes requested):

internal/output/renderers/bulletlist.go
mitchell-as: By making it typed as suggested later in the review, there cannot be an error to return anymore.

internal/colorize/wrap.go
```diff
@@ -0,0 +1,161 @@
...
+		result = result + wrapped.Line
+	}
+
+	return result
+}
+
+var indentRegexp = regexp.MustCompile(`^([ ]+)`)
+var isLinkRegexp = regexp.MustCompile(`\s*(\[[^\]]+\])?https?://`)
+
+func Wrap(text string, maxLen int, includeLineEnds bool, continuation string) WrappedLines {
```
mitchell-as, line 29: Despite me doing a rename, GitHub thinks this entire file is new. I refactored this method and added `colorTags()` below. That's it. Everything else is the same.

internal/colorize/wrap.go
```diff
@@ -0,0 +1,161 @@
...
+		start, stop := intRange[0], intRange[1]
+		if pos >= start && pos <= stop-1 {
+			return true
+		}
+	}
+	return false
+}
+
+// colorTags returns the currently active color tags (if any) at the given position.
+func colorTags(pos int, ranges [][]int, names [][]string) []string {
```
mitchell-as, line 140: This is new after the rename.

internal/colorize/wrap_test.go
```diff
@@ -0,0 +1,159 @@
...
+				escape := func(v string) string {
+					return strings.Replace(v, "\n", "\\n", -1)
+				}
+				t.Errorf("Wrap() = %v, want %v (wrap data: %s)", escape(got.String()), escape(tt.text), escape(fmt.Sprintf("%#v", got)))
+			}
+		})
+	}
+}
+
+func TestWrapBullet(t *testing.T) {
```
mitchell-as, line 134: Despite renaming this file, GitHub thinks all of its contents are new. I added this test and the one that follows. The tests above I left alone.

internal/runbits/dependencies/changesummary.go
mitchell-as: There's a new `BulletTreeDisabled` preset.

internal/runbits/dependencies/summary.go
mitchell-as: There's a new `BulletTreeDisabled` preset I'm using now.

Naatan (approved):
Nice work! Almost felt like I was reviewing a Komodo PR..

PR #3577, opened 2024-11-01
Handle more panics

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3136" title="DX-3136" target="_blank">BugDX-3136</a>  Rollbar: Could not report messages as CheckMessages return an error: Error sending messages request:
  </td></table>
  <br />

### DIFF
--- a/cmd/state-svc/internal/messages/messages.go
+++ b/cmd/state-svc/internal/messages/messages.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"sync"
 	"time"
 
@@ -15,6 +16,7 @@ import (
 	"github.com/ActiveState/cli/internal/httputil"
 	"github.com/ActiveState/cli/internal/logging"
 	"github.com/ActiveState/cli/internal/poller"
+	"github.com/ActiveState/cli/internal/runbits/panics"
 	"github.com/ActiveState/cli/internal/strutils"
 	auth "github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/ActiveState/cli/pkg/sysinfo"
@@ -43,6 +45,9 @@ func New(cfg *config.Instance, auth *auth.Auth) (*Messages, error) {
 	}
 
 	poll := poller.New(1*time.Hour, func() (interface{}, error) {
+		defer func() {
+			panics.LogAndPanic(recover(), debug.Stack())
+		}()
 		resp, err := fetch()
 		return resp, err
 	})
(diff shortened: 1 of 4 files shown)


internal/poller/poller.go
```diff
@@ -19,8 +21,15 @@ type Poller struct {
 }
 
 func New(interval time.Duration, pollFunc func() (interface{}, error)) *Poller {
+	wrappedFn := func() (interface{}, error) {
+		defer func() {
+			panics.LogAndPanic(recover(), debug.Stack())
+		}()
+		return pollFunc()
+	}
```
MDrakos, line 29: Panics can occur in the goroutine where the poll func is started. If we don't want this automatic wrapping we can update each caller of this library to manually inject it into their pollfunc.

cmd/state-svc/internal/messages/messages.go
```diff
@@ -74,7 +74,11 @@ func (m *Messages) Check(command string, flags []string) ([]*graph.MessageInfo,
 	if cacheValue == nil {
 		return []*graph.MessageInfo{}, nil
 	}
-	allMessages := cacheValue.([]*graph.MessageInfo)
+
+	allMessages, ok := cacheValue.([]*graph.MessageInfo)
+	if !ok {
+		return nil, errs.New("Could not get messages from cache")
```
Naatan, line 80:
```suggestion
		return nil, errs.New("cacheValue has unexpected type: %T", cacheValue)
```

cmd/state-svc/internal/messages/messages.go
```diff
@@ -110,7 +114,11 @@ func check(params *ConditionParams, messages []*graph.MessageInfo, lastReportMap
 		logging.Debug("Checking message %s", message.ID)
 		// Ensure we don't show the same message too often
 		if lastReport, ok := lastReportMap[message.ID]; ok {
-			lastReportTime, err := time.Parse(time.RFC3339, lastReport.(string))
+			lr, ok := lastReport.(string)
+			if !ok {
+				return nil, errs.New("Could not get last reported time for message %s as it's not a string", message.ID)
```
Naatan, line 119:
```suggestion
				return nil, errs.New("Could not get last reported time for message %s as it's not a string: %T", message.ID, lastReport)
```

internal/poller/poller.go
Naatan, line 29: I think we need to implement this as part of the pollFunc that we pass. Reason being that if we catch it here the stacktrace won't point to what caused the panic.

Naatan (changes requested):

Naatan (approved):

PR #3579, opened 2024-11-04
Implement user-facing errors for environment setup commands.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-2855" title="DX-2855" target="_blank">StoryDX-2855</a>  User facing errors for Environment Setup commands
  </td></table>
  <br />
 

For all `locale.WrapError()` calls in these runners, I navigated up the call chain to see what kinds of errors were being returned.
- For the cases where a locale or input error (e.g. projectfile package) was ultimately being produced, I replaced the runner error with `errs.Wrap()` since the extra localization was superfluous, and no user-facing error was needed.
- For the cases where there was no locale or input error, I replaced the error with an error struct, and ultimately a user facing error.

### DIFF
--- a/internal/runners/initialize/init.go
+++ b/internal/runners/initialize/init.go
@@ -20,7 +20,6 @@ import (
 	"github.com/ActiveState/cli/internal/primer"
 	"github.com/ActiveState/cli/internal/runbits/buildscript"
 	"github.com/ActiveState/cli/internal/runbits/dependencies"
-	"github.com/ActiveState/cli/internal/runbits/errors"
 	"github.com/ActiveState/cli/internal/runbits/org"
 	"github.com/ActiveState/cli/internal/runbits/rationalize"
 	"github.com/ActiveState/cli/internal/runbits/runtime"
@@ -89,6 +88,8 @@ type errUnrecognizedLanguage struct {
 func (e errUnrecognizedLanguage) Error() string {
 	return fmt.Sprintf("unrecognized language: %s", e.Name)
 }
+
+var errDeleteProjectAfterError = errs.New("could not delete initialized project")
 
 // New returns a prepared ptr to Initialize instance.
 func New(prime primeable) *Initialize {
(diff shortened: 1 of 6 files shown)


internal/runners/initialize/init.go
```diff
@@ -200,11 +200,7 @@ func (r *Initialize) Run(params *RunParams) (rerr error) {
 
 	version, err := deriveVersion(lang, languageVersion, r.auth)
 	if err != nil {
-		if inferred || errors.IsReportableError(err) {
-			return locale.WrapError(err, "err_init_lang", "", languageName, languageVersion)
-		} else {
-			return locale.WrapExternalError(err, "err_init_lang", "", languageName, languageVersion)
-		}
+		return errs.Wrap(err, "Unable to get language version")
```
mitchell-as, line 203: `deriveVersion()` has undergone lots of changes relatively recently and, from what I gather, will return a localized error when appropriate (e.g. invalid language).

MDrakos (commented):
The code and the errors look fine to me.

> For the cases where a locale or input error (e.g. projectfile package) was ultimately being produced, I replaced the runner error with errs.Wrap() since the extra localization was superfluous, and no user-facing error was needed.

I believe we want to move towards error localization _only_ happening in the rationalizers. Is this something that is feasible to do with some of the errors you found? I may be wrong so feel free to correct me however I'm fairly certain this is where we want to go.

mitchell-as:
The low-level locale errors I found in packages like projectfile touch a TON of other bits of code. I felt that was outside the scope of this ticket, since this is targetted at just the environment commands.

MDrakos:
> The low-level locale errors I found in packages like projectfile touch a TON of other bits of code. I felt that was outside the scope of this ticket, since this is targetted at just the environment commands. The road to pure user-facing errors is a long one.

That's what I expected. Thanks for the clarification!

MDrakos (approved):

PR #3580, opened 2024-11-05
Update GraphQL Response Handling

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3101" title="DX-3101" target="_blank">TaskDX-3101</a>  Update graphql response handling logic to automatically handle grabbing query result
  </td></table>
  <br />
 

I fairly confident that the remaining test failures are not related to this PR. 

Essentially we are extracting a data path from the request string. We then use this as the point at which we start to unmarshal the response data. If we feel that the path inference is too magical we can remove it and require that the call that constructs the request also declares where response unmarshalling should begin, or some other approach.


Naatan (changes requested):
I could be wrong here, I haven't played with the code like you have, but this feels like it should be solve-able with less logic.

What about something like this?

https://stackoverflow.com/a/11067084

ie. we always pass the response to a map with a raw json message, the key will be known to us, so we just grab the raw json message via the key and then continue unmarshalling with the response structure that was passed.

PR #3581, opened 2024-11-05
Update check-format script

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3077" title="DX-3077" target="_blank">BugDX-3077</a>  check_format raising files that were not touched
  </td></table>
  <br />
 

The reason the script raised an issue in this PR: https://github.com/ActiveState/cli/pull/3504 is because it was run when the target branch was `master` and not the RC branch that it should have been. We rely on `GITHUB_BASE_REF` to identify the target branch. I looked into some ways to work around this but there doesn't seem to be a clear solution so rather than try I've just opted to skip the check if the target branch is master.

### DIFF
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -389,28 +389,19 @@ scripts:
     standalone: true
     description: "Checks if the code is formatted correctly"
     value: |
-      set -e
-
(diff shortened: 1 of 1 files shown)


mitchell-as (approved):
That explanation works for me!

PR #3588, opened 2024-11-08
Target & Verify action should append to Jira labels, not overwrite.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3152" title="DX-3152" target="_blank">BugDX-3152</a>  "Target & Verify" automation removes an existing "Found_By_QA_Automation" label
  </td></table>
  <br />

### DIFF
--- a/scripts/internal/workflow-helpers/jira.go
+++ b/scripts/internal/workflow-helpers/jira.go
@@ -181,6 +181,7 @@ func UpdateJiraFixVersion(client *jira.Client, issue *jira.Issue, versionID stri
 		Key: issue.Key,
 		Fields: &jira.IssueFields{
 			FixVersions: []*jira.FixVersion{{ID: versionID}},
+			Labels:      issue.Fields.Labels,
 		},
 	}
 


mitchell-as:
Tested by adding a label to the JIRA issue, then running Target & Verify, and checking that "WasNextFeasible" was appended instead of overwriting the existing label.

mitchell-as:
image

mitchell-as:
@Naatan ping

mitchell-as:
The concern you raised may look problematic based on the diff GitHub shows, but the troublesome line is here: https://github.com/ActiveState/cli/pull/3588/files#diff-2224e300e5d1da6fa67f6cf4cfc1d6b288f60bda27d6d0181bce500076850273R192. We need to initialize the struct with the current set of labels if we want to append to it. Otherwise, we're appending to an empty list, which will overwrite the current label.

I think we're fine.

Naatan (approved):

PR #3591, opened 2024-11-13
Fix Publish Integration Test

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3160" title="DX-3160" target="_blank">BugDX-3160</a>  Nightly failure: TestPublishIntegrationTestSuite/TestPublish/*
  </td></table>
  <br />
 

We now have two graphQL client implementations. I've ported the changes from https://github.com/ActiveState/cli/pull/3580 into this PR but the real solution will be to unify the client code. I've filed a story for that here: https://activestatef.atlassian.net/browse/DX-3162

### DIFF
--- a/internal/gqlclient/gqlclient.go
+++ b/internal/gqlclient/gqlclient.go
@@ -312,8 +312,9 @@ func (c *Client) runWithFiles(ctx context.Context, gqlReq RequestWithFiles, resp
 		logging.Debug("gqlclient: response: %s", responseData)
 	}
 
+	intermediateResp := make(map[string]interface{})
 	gr := &graphResponse{
-		Data: response,
+		Data: &intermediateResp,
 	}
 	req = req.WithContext(ctx)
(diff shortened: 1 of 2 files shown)


internal/gqlclient/gqlclient.go
```diff
@@ -371,5 +372,20 @@ func (c *Client) runWithFiles(ctx context.Context, gqlReq RequestWithFiles, resp
 	if err := json.Unmarshal(resp, &gr); err != nil {
 		return errors.Wrap(err, "decoding response")
 	}
-	return nil
+
+	if len(intermediateResp) == 1 {
```
mitchell-as, line 376: Nit: a comment describing why this conditional is here or what conditions it may occur under would probably be useful for anyone debugging around this area in the future.

mitchell-as (dismissed):

mitchell-as (approved):

PR #3592, opened 2024-11-13
Add a bit of lag to the JWT renewal time.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3151" title="DX-3151" target="_blank">BugDX-3151</a>  `state checkout` with expired JWT gives unhelpful error message
  </td></table>
  <br />
 

It was possible to land within the interval between actual expired time and expected expired time.

I did my best to try and reproduce this, but I wasn't able to. I genuinely think I landed in the brief window of time where the JWT expired on the Platform, but state-svc or state tool thought it was still valid. Reducing the time state-svc or state tool think it's valid should avoid this. Either way, it's not something our users will run into on a regular basis (and waiting a minute or two for the state-svc poller to refresh would have silently fixed this error).

### DIFF
--- a/pkg/platform/authentication/auth.go
+++ b/pkg/platform/authentication/auth.go
@@ -317,7 +317,7 @@ func (s *Auth) UpdateSession(accessToken *mono_models.JWT) {
 	s.bearerToken = accessToken.Token
 	clientAuth := httptransport.BearerToken(s.bearerToken)
 	s.clientAuth = &clientAuth
-	s.lastRenewal = ptr.To(time.Now())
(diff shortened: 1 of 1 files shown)


Naatan (changes requested):
This is effectively duplicating the logic here:

https://github.com/ActiveState/cli/blob/8e47470c8c8b075aa0f23eca986631eec4f1a1f1/pkg/platform/authentication/auth.go#L151-L151

Suggest instead lowering the lifetime to say 50 minutes:

https://github.com/ActiveState/cli/blob/8e47470c8c8b075aa0f23eca986631eec4f1a1f1/pkg/platform/authentication/auth.go#L41-L41

Naatan (approved):

PR #3593, opened 2024-11-14
Recognize circular links

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3146" title="DX-3146" target="_blank">BugDX-3146</a>  Checkout fail for `ActiveStateCommunityTesting/Py-3.10.13-Data-Science`
  </td></table>
  <br />

### DIFF
--- a/internal/smartlink/smartlink.go
+++ b/internal/smartlink/smartlink.go
@@ -7,7 +7,6 @@ import (
 	"github.com/ActiveState/cli/internal/errs"
 	"github.com/ActiveState/cli/internal/fileutils"
 	"github.com/ActiveState/cli/internal/logging"
-	"github.com/ActiveState/cli/internal/multilog"
 )
 
 // LinkContents will link the contents of src to desc
@@ -41,7 +40,15 @@ func LinkContents(src, dest string) error {
 // Link creates a link from src to target. MS decided to support Symlinks but only if you opt into developer mode (go figure),
 // which we cannot reasonably force on our users. So on Windows we will instead create dirs and hardlinks.
 func Link(src, dest string) error {
-	var err error
+	isCircular, err := isCircularLink(src)
+	if err != nil {
+		return errs.Wrap(err, "could not check for circular link")
+	}
+	if isCircular {
+		logging.Warning("Skipping linking '%s' to '%s' as it is a circular link", src, dest)
+		return nil
+	}
+
 	src, dest, err = resolvePaths(src, dest)
 	if err != nil {
 		return errs.Wrap(err, "Could not resolve src and dest paths")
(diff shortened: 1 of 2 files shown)


internal/smartlink/smartlink.go
```diff
@@ -55,6 +56,15 @@ func Link(src, dest string) error {
 			return errs.Wrap(err, "could not read directory %s", src)
 		}
 		for _, entry := range entries {
+			circular, err := isCircularLink(filepath.Join(src, entry.Name()))
+			if err != nil {
+				return errs.Wrap(err, "could not check for circular link")
+			}
+			if circular {
+				multilog.Error("Build contains a circular symlink: %s -> %s", filepath.Join(src, entry.Name()), dest)
+				continue
```
Naatan, line 65: I think we should error out on this, unless BE has a solution? Continuing would just give you a corrupted runtime.

internal/smartlink/smartlink.go
```diff
@@ -80,6 +90,32 @@ func Link(src, dest string) error {
...
+		target = resolved
+	}
+
+	// If the target points at the parent directory, we have a circular link
+	if filepath.Dir(src) == target {
+		return true, nil
+	}
+
+	return false, nil
+}
```
Naatan, line 117: This doesn't feel right. It's totally valid to have a link like this:

```
- dir1/file1
- dir1/dir2 > ../dir1
```

The problem is that our Linking mechanism doesn't understand this and will continue to recurse forever. We need to fix that, whilst still supporting a link that goes back up the path.

Naatan (changes requested):

internal/smartlink/smartlink.go
MDrakos: BE considers this an error case and has patched it, but advised that it could happen again so we should guard against it.

I'm not sure how this comment works with your next one. We should error our if we encounter a circular link but also support it?

internal/smartlink/smartlink.go
MDrakos: Since this library doesn't link directories I'm not sure how we can support this beyond what this PR is doing. If we encounter a circular directory we skip it, all of the contents of the directory will still be linked.

 I've moved the check to make it more clear. I think that in the end, all we can do is ensure all of the files in the directories and subdirectories have been linked with the same layout as the source.

Naatan (changes requested):
I still feel like this is solving the wrong problem.

The problem isn't that the link points to itself. It works fine in the tar archive, so it should work fine on disk. We should not be verifying the validity of symlinks created by the builders.

What we should guard against is our linking mechanism hitting a recursion. Like in this case we should be able to tell that we already traversed the directory the symlink goes to. Heck, we probably shouldn't traverse at all if we hit a symlink, we should just be copying the symlink. In fact I'm pretty sure that's what we do, but in this case it's a symlink of a file To a directory, so we might not be identifying that properly.

The current changeset is working around this problem by identifying a very specific way that this might arise. I think we should instead make our linking mechanic robust enough that this is a non-issue in the first place.

MDrakos:
It seems to me that in order to solve this problem we either need to make the `smarlink` package a lot smarter so that it can keep track of things it's already linked, or support linking of files to directories. I don't think this decision should be made here so I'm going to push this back to triage.

PR #3594, opened 2024-11-15
Add Date Recognition To Messages

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3161" title="DX-3161" target="_blank">TaskDX-3161</a>  Improvements to messaging mechanic to support maintenance messages
  </td></table>
  <br />
 

I had some trouble interpreting the ACs here. This should cover all of the requirements but we may want to take a different approach.

### DIFF
--- a/cmd/state-svc/internal/messages/condition.go
+++ b/cmd/state-svc/internal/messages/condition.go
@@ -4,6 +4,7 @@ import (
 	"regexp"
 	"strings"
 	"text/template"
+	"time"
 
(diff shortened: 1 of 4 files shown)


MDrakos:
After some more thorough testing this approach will not work with older version of the state tool. For now, I will keep this up for reference and put my findings in the story details as we may need to try something different.

MDrakos:
Closing this in favour of a new implementation.

PR #3595, opened 2024-11-15
Update prompt to allow `--force` overrides.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3166" title="DX-3166" target="_blank">StoryDX-3166</a>  Bring back the `--force` flag to allow running an action regardless of prompts OR defaults
  </td></table>
  <br />


mitchell-as:

mitchell-as:
Note: this will also be submitted for architecture review upon approval.

internal/prompt/prompt.go
```diff
@@ -158,7 +177,15 @@ func (p *Prompt) Select(title, message string, choices []string, defaultChoice *
...
 // Confirm prompts user for yes or no response.
-func (p *Prompt) Confirm(title, message string, defaultChoice *bool) (bool, error) {
+func (p *Prompt) Confirm(title, message string, defaultChoice *bool, forcedChoice *bool) (bool, error) {
+	if p.isForced {
+		if forcedChoice == nil {
+			return false, errs.New("No force option given for force-enabled prompt")
+		}
+		logging.Debug("Prompt %s confirmed with choice %v in force mode", title, forcedChoice)
+		return *forcedChoice, nil
+	}
```
MDrakos, line 187: Do we even need to check `isForced` here? I'm just not seeing why we have to set the prompt to a forced mode in order to use the forced choice. It's a bit different than the interactive use case.

internal/runners/clean/config.go
```diff
@@ -52,14 +53,19 @@ func (c *Config) Run(params *ConfigParams) error {
...
 
-	if !params.Force {
-		ok, err := c.confirm.Confirm(locale.T("confirm"), locale.T("clean_config_confirm"), ptr.To(true))
-		if err != nil {
-			return locale.WrapError(err, "err_clean_config_confirm", "Could not confirm clean config choice")
-		}
-		if !ok {
-			return locale.NewInputError("err_clean_config_aborted", "Cleaning of config aborted by user")
-		}
+	defaultChoice := !c.confirm.IsInteractive()
```
MDrakos, line 56: nit: Could probably get rid of this with `ptr.To(!c.confirm.IsInteractive()` below

MDrakos (changes requested):
Looks good to me. Our prompt code is a bit hard to grok. We might want to consider a runbit or revisiting some of the code in the future.

internal/prompt/prompt.go
mitchell-as: We do, because we are always supplying a forced choice for the prompts that support it. Just because we're supplying the choice doesn't mean force is enabled, so we do need to check.

When we initialize a prompt, we're supplying the `isInteractive` field inheritently, which depends on the terminal or if the `--non-interactive` flag is passed. That's the only thing different with the interactive/non-interactive case; we're still checking for `isInteractive` in this method.

Note: We cannot initialize prompt with a `isForced` because that information is not yet known, hence the need to set forced mode later.

I hope this makes sense.

internal/runners/clean/config.go
mitchell-as: We could, but the in-line argument wouldn't be as descriptive. This reads better and falls in line with other `Confirm()` calls.

internal/prompt/prompt.go
MDrakos: It does, but I still think we could get away without having an `isForced` property here. We pass nil to the `check` function if we're not forcing a choice otherwise it's explicit. This feels like we're forcing (no pun intended) this forced state to follow the same pattern as the interactive state when they are different things.

Anyways that's just my 2 cents. I'm fine with whatever way you want to do this. 👍

MDrakos (dismissed):

cmd/state/internal/cmdtree/packages.go
```diff
@@ -67,6 +68,11 @@ func newInstallCommand(prime *primer.Values) *captain.Command {
 				Description: locale.T("package_flag_ts_description"),
 				Value:       &params.Timestamp,
 			},
+			{
+				Name:        "force",
+				Description: locale.Tl("package_flag_force_description", "Ignore security policy preventing packages with CVEs from being installed (not recommended)"),
+				Value:       &force,
+			},
```
Naatan, line 75: We should not be adding force as a local flag. It should be a global flag just like non-interactive. Whether it is relevant to use on a command and communicating that through the UI is not a requirement, just like with non-interactive.

internal/prompt/prompt.go
```diff
@@ -18,14 +19,17 @@ type EventDispatcher interface {
...
 }
 
-// Prompter is the interface used to run our prompt from, useful for mocking in tests
+// Prompter is the interface used to run our prompt from
 type Prompter interface {
 	Input(title, message string, defaultResponse *string, flags ...ValidatorFlag) (string, error)
 	InputAndValidate(title, message string, defaultResponse *string, validator ValidatorFunc, flags ...ValidatorFlag) (string, error)
 	Select(title, message string, choices []string, defaultResponse *string) (string, error)
-	Confirm(title, message string, defaultChoice *bool) (bool, error)
+	Confirm(title, message string, defaultChoice *bool, forcedChoice *bool) (bool, error)
```
Naatan, line 27: This is only setting the forced value for confirm. The story is about all prompts. Anywhere where we handle defaultResponse and thus non-interactive we should also handle force.

internal/prompt/prompt.go
```diff
@@ -39,18 +43,33 @@ type Prompt struct {
...
+	p.isInteractive = interactive
+}
+
+// EnableForce forces confirm prompts to return the force value (which is often different from the
+// non-interactive value).
+func (p *Prompt) EnableForce() {
+	p.isForced = true
+}
+
+func (p *Prompt) IsForceEnabled() bool {
```
Naatan, line 69: Can we name this to  be consistent with the Interactive mode?

ie. `SetForced()` and `IsForced()`.

Naatan (changes requested):

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -318,10 +327,6 @@ func newStateCommand(globals *globalOptions, prime *primer.Values) *captain.Comm
 		},
 		[]*captain.Argument{},
 		func(ccmd *captain.Command, args []string) error {
-			if globals.Verbose {
-				logging.CurrentHandler().SetVerbose(true)
```
mitchell-as, line 322: This is set in the `OnUse` handler function, so this is not needed.

internal/locale/locales/en-us.yaml
```diff
@@ -23,6 +23,8 @@ flag_state_output_description:
   other: "Set the output method. Possible values: plain, simple, json, editor"
 flag_state_non_interactive_description:
   other: Run the State Tool without any prompts
+flag_state_force_description:
+  other: Run the State Tool without any prompts, overriding any safe defaults
```
Naatan, line 27:
```suggestion
flag_state_non_interactive_description:
  other: Assume default values for any prompts
flag_state_force_description:
  other: Disable prompts and perform the requested action regardless of defaults values (this can result in destructive or unsafe actions, such as installing packages with critical CVEs)
```

internal/prompt/prompt.go
```diff
@@ -81,10 +101,24 @@ func interactiveInputError(message string) error {
...
+	if p.isForced {
+		response := forcedResponse
+		if response == nil {
+			response = defaultResponse
+		}
+		if response != nil {
+			p.out.Notice(locale.Tr("prompt_using_force", *response))
+			return *response, nil
+		}
+		return "", errs.New("No force option given for forced prompt")
```
Naatan, line 116: We should localize this through rationalized errors. We can do the rationalising here: https://github.com/ActiveState/cli/blob/8e47470c8c8b075aa0f23eca986631eec4f1a1f1/internal/captain/rationalize.go#L13-L13

Naatan (changes requested):
Great work! Just some minor touchup.

Naatan (approved):
Nice work!

PR #3598, opened 2024-11-18
Handle recursive links in runtime sources.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3167" title="DX-3167" target="_blank">StoryDX-3167</a>  Our depot linking mechanic is aware of directories it has already traversed
  </td></table>
  <br />


mitchell-as:
Wrong branch

PR #3599, opened 2024-11-18
Reorganize BuildPlanner Error Handling

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3127" title="DX-3127" target="_blank">TaskDX-3127</a>  Update BuildPlanner error handling
  </td></table>
  <br />
 

This PR updates the errors we request in the buildplanner graphQL queries to be a bit more standardized. I've also removed unused error types. The actual handling of errors hasn't changed, just how the underlying information is passed to the callers.


pkg/platform/api/buildplanner/response/commit.go
```diff
@@ -38,7 +38,7 @@ func processPlanningError(message string, subErrors []*BuildExprError) error {
 			}
 		}
 
-		if se.Type != types.RemediableSolveErrorType && se.Type != types.GenericSolveErrorType {
+		if se.Type != types.RemediableSolveErrorType && se.Type != types.GenericSolveErrorType && se.Type != types.RemediableError {
```
Naatan, line 41: Are you sure RemediableSolveErrorType and RemediableError are distinct? I think we only did the solver remediations, so RemediableError might just be the base error, but they're effectively the same.

Naatan (changes requested):
Looks good, just one question on the remediable error type.

pkg/platform/api/buildplanner/response/commit.go
MDrakos: They are different but after some inspection it looks like the `RemediableError` is experimental so I've removed any references to it. https://github.com/ActiveState/TheHomeRepot/blob/master/service/buildplanner/buildplan.graphql#L2388

Naatan (approved):

PR #3600, opened 2024-11-19
Handle recursive links in runtime sources.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3167" title="DX-3167" target="_blank">StoryDX-3167</a>  Our depot linking mechanic is aware of directories it has already traversed
  </td></table>
  <br />

### DIFF
--- a/internal/smartlink/smartlink.go
+++ b/internal/smartlink/smartlink.go
@@ -10,7 +10,7 @@ import (
 )
 
 // LinkContents will link the contents of src to desc
-func LinkContents(src, dest string) error {
+func LinkContents(src, dest string, visited map[string]bool) error {
 	if !fileutils.DirExists(src) {
 		return errs.New("src dir does not exist: %s", src)
 	}
@@ -24,12 +24,23 @@ func LinkContents(src, dest string) error {
 		return errs.Wrap(err, "Could not resolve src and dest paths")
 	}
 
+	if visited == nil {
+		visited = make(map[string]bool)
+	}
+	if _, exists := visited[src]; exists {
+		// We've encountered a recursive link. This is most often the case when the resolved src has
+		// already been visited. In that case, just link the dest to the src (which may be a directory;
+		// this is fine).
+		return linkFile(src, dest)
+	}
+	visited[src] = true
+
 	entries, err := os.ReadDir(src)
 	if err != nil {
 		return errs.Wrap(err, "Reading dir %s failed", src)
 	}
 	for _, entry := range entries {
-		if err := Link(filepath.Join(src, entry.Name()), filepath.Join(dest, entry.Name()), nil); err != nil {
+		if err := Link(filepath.Join(src, entry.Name()), filepath.Join(dest, entry.Name()), visited); err != nil {
 			return errs.Wrap(err, "Link failed")
 		}
 	}
@@ -39,7 +50,7 @@ func LinkContents(src, dest string) error {
 
 // Link creates a link from src to target. MS decided to support Symlinks but only if you opt into developer mode (go figure),
 // which we cannot reasonably force on our users. So on Windows we will instead create dirs and hardlinks.
-func Link(src, dest string, visited map[string]int) error {
+func Link(src, dest string, visited map[string]bool) error {
 	var err error
 	src, dest, err = resolvePaths(src, dest)
 	if err != nil {
@@ -47,19 +58,15 @@ func Link(src, dest string, visited map[string]int) error {
 	}
 
 	if visited == nil {
-		visited = make(map[string]int)
+		visited = make(map[string]bool)
 	}
-	if count, exists := visited[src]; exists {
+	if _, exists := visited[src]; exists {
 		// We've encountered a recursive link. This is most often the case when the resolved src has
-		// already been visited. We will recurse into the directory no more than once, so that any
-		// runtime paths that reference the link will not silently fail.
-		if count > 1 {
-			return nil
-		}
-		visited[src]++
-	} else {
-		visited[src] = 1
+		// already been visited. In that case, just link the dest to the src (which may be a directory;
+		// this is fine).
+		return linkFile(src, dest)
 	}
+	visited[src] = true
 
 	if fileutils.IsDir(src) {
 		if err := fileutils.Mkdir(dest); err != nil {
--- a/internal/smartlink/smartlink.go
+++ b/internal/smartlink/smartlink.go
@@ -24,17 +24,6 @@ func LinkContents(src, dest string, visited map[string]bool) error {
 		return errs.Wrap(err, "Could not resolve src and dest paths")
 	}
 
-	if visited == nil {
-		visited = make(map[string]bool)
-	}
-	if _, exists := visited[src]; exists {
-		// We've encountered a recursive link. This is most often the case when the resolved src has
-		// already been visited. In that case, just link the dest to the src (which may be a directory;
-		// this is fine).
-		return linkFile(src, dest)
-	}
-	visited[src] = true
-
 	entries, err := os.ReadDir(src)
 	if err != nil {
 		return errs.Wrap(err, "Reading dir %s failed", src)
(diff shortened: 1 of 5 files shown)


internal/smartlink/smartlink.go
```diff
@@ -24,12 +24,23 @@ func LinkContents(src, dest string) error {
...
+	if visited == nil {
+		visited = make(map[string]bool)
+	}
+	if _, exists := visited[src]; exists {
+		// We've encountered a recursive link. This is most often the case when the resolved src has
+		// already been visited. In that case, just link the dest to the src (which may be a directory;
+		// this is fine).
+		return linkFile(src, dest)
+	}
+	visited[src] = true
```
Naatan, line 36: I don't think we need this check here. I think just the one on the Link function already covers our use-case.

internal/smartlink/smartlink.go
```diff
@@ -39,13 +50,24 @@ func LinkContents(src, dest string) error {
...
```
Naatan, line 69: This should not be creating folder links. Or really any links. If we've already visited it then the link already exists and we should cancel out. But importantly folder links are to be avoided; they do not work on windows (due to it using hard links).

internal/smartlink/smartlink_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLinkContentsWithCircularLink(t *testing.T) {
```
Naatan, line 13: Can we also add a test for the issue that brought up this bug? I imagine we have a commit for the project that was surfacing the bug, we should be able to use that.

Naatan (changes requested):

internal/smartlink/smartlink.go
mitchell-as: The ticket specifically calls out this scenario:

```
- dir1/file1
- dir1/dir2 > ../dir1
```

The archive we get has a link from dir2 back to dir1. I suppose technically dir2 is not a folder since there are no folder links. Regardless, it's a symlink to a directory, which is perfectly valid. We want to emulate this in case any runtime objects use paths containing dir2. If we didn't create dir2 at all, then the runtime would be silently broken.

I also went back and looked at Mike's original PR, and he didn't make any links if he detected recursion. You rejected it. My understanding is that we want the link.

I am aware this is broken on Windows. The unit test specifically expects an error on Windows, and our runtime will fail to install with an error rather than stack overflow due to infinite recursion.

If you are expecting something else based on the ticket description, prior ticket description and comments, and prior PR, then please share it. I was not and am not able to discern that information.

internal/smartlink/smartlink_test.go
mitchell-as: We could, but it's a massive runtime that would really bog down CI. I think the unit test is sufficient.

internal/smartlink/smartlink.go
mitchell-as: Long story short: we need a reference to the circular src so runtime paths that use it won't silently fail. Since you don't want the "dir1/dir2 > ../dir1" link specified in the ticket, I've made it so we make the dir2 directory, and link its circular contents from dir1 no more than once (i.e. dir1/dir2/dir2 will not exist, but all non-recursive items in dir1 will be present in dir2).

Naatan: Ahh I see what you mean. Hmm, I guess in this case we just defer to the artifact. Can you update the logic so that we do place a folder link but only if the folder was a link itself in the archive?

So basically src>target may be a directory symlink if src it itself a symlink. Otherwise we are in an error condition, which I don't think would ever actually trigger because how can you hit recursion if not through this exact symlink pattern.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3600#discussion_r1852699599

internal/smartlink/smartlink.go
mitchell-as: @Naatan What you are requesting is basically just a revert of the most recent commit. I don't have any issues with that, however this will result in an error on Windows, which does not support links to directories. (The user will get an error stating this during runtime setup, but that's better than the stack overflow that was previously happening.)

It's worth pointing out that the original issue was Linux-onlly, and that I doubt a tarball produced for Windows would have such a recursive symlink structure, so it's probably safe to do the revert. At the same time, this current solution will handle that case without an error.

Please advise how you'd like me to proceed.

Naatan: I think the original logic was to create symlinks irrespective of whether the src was a symlink. Again I acknowledge that the src being a symlink is likely the only way that this permutation could happen, but there's value in making the logic explicit. Not only will that give us a clear indication of the problem if we do run into it, it also makes it less likely that we break this logic when we revisit the code.

That does indeed mean on Windows there is an error condition, but one that would only happen if we have a Windows artifact with a symlink, which itself would be an error regardless of whatever else we're doing with the depot.

So I think your original solution was spot on, I'd just like to make the logic explicit for maintainability.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3600#discussion_r1852939493

internal/smartlink/smartlink.go
mitchell-as: Okay, I hope I understood. If not, please be patient with me. It's been a long day.

internal/smartlink/smartlink.go
```diff
@@ -39,13 +39,26 @@ func LinkContents(src, dest string) error {
...
 	var err error
 	src, dest, err = resolvePaths(src, dest)
 	if err != nil {
 		return errs.Wrap(err, "Could not resolve src and dest paths")
 	}
 
+	if visited == nil {
+		visited = make(map[string]bool)
+	}
+	if _, exists := visited[src]; exists && srcWasSymlink {
```
Naatan, line 54: Hmm I think this is now trying to solve the same issue two ways. If the src is a symlink then we should just make the deployment a link also, and stop iterating further down because it's a symlink. In this case what we already visited is irrelevant.

So I think we can get rid of the visited logic altogether, and simply revise it to make dir links if the src is a dir link, and otherwise create the dir and continue recursion.

Naatan (changes requested):

mitchell-as:
I am a bit nervous about these changes, as I don't know if they'll have unintended consequences. The critical tests seem to be passing, so that brings a little comfort.

mitchell-as:
@Naatan ping

internal/smartlink/smartlink.go
```diff
@@ -47,6 +47,13 @@ func Link(src, dest string) error {
 	}
 
 	if fileutils.IsDir(src) {
+		if isSymlink(src) {
```
Naatan, line 50:
```suggestion
		if fileutils.IsSymlink(src) {
```

internal/smartlink/smartlink.go
```diff
@@ -131,15 +138,20 @@ func UnlinkContents(src, dest string) error {
 	return nil
 }
 
+func isSymlink(src string) bool {
+	target, err := fileutils.SymlinkTarget(src)
+	return err == nil && src != target
+}
+
```
Naatan, line 145:
```suggestion
```

`src != target` does not assert whether something is a symlink. Is this a concern we have? We could add the error check and return an error, but this doesn't feel relevant to the issue at hand. ie. the bug we're addressing isn't that a file is linking to itself.

internal/smartlink/smartlink.go
```diff
@@ -131,15 +138,20 @@ func UnlinkContents(src, dest string) error {
...
 // This is to ensure that we're always comparing apples to apples when doing string comparisons on paths.
 func resolvePaths(src, dest string) (string, string, error) {
 	var err error
-	src, err = fileutils.ResolveUniquePath(src)
+	src, err = filepath.Abs(filepath.Clean(src))
 	if err != nil {
 		return "", "", errs.Wrap(err, "Could not resolve src path")
 	}
-	dest, err = fileutils.ResolveUniquePath(dest)
+	dest, err = filepath.Abs(filepath.Clean(dest))
```
Naatan, line 154: Please revert. Absolute path does not give a canonical result, whereas ResolveUniquePath does (or at least -tries to-).

Naatan (changes requested):

PR #3601, opened 2024-11-19
Add Date Recognition To Messages

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3161" title="DX-3161" target="_blank">TaskDX-3161</a>  Improvements to messaging mechanic to support maintenance messages
  </td></table>
  <br />
 

Tested the release version of the state tool with a timestamped message and it printed as expected without any errors in the log.


cmd/state-svc/internal/messages/messages.go
```diff
@@ -140,11 +165,23 @@ func check(params *ConditionParams, messages []*graph.MessageInfo, lastReportMap
...
+			logging.Debug("Skipping message %s as it is outside of its date range", message.ID)
+			continue
+		}
+
 		// Validate the conditional
 		if message.Condition != "" {
 			result, err := strutils.ParseTemplate(fmt.Sprintf(`{{%s}}`, message.Condition), params, funcMap)
 			if err != nil {
-				return nil, errs.Wrap(err, "Could not parse condition template for message %s", message.ID)
+				logging.Warning("Could not parse condition template for message %s: %v", message.ID, err)
```
mitchell-as, line 183: Is there a risk this will cause us to miss potential issues that Rollbar would otherwise notify us of?

mitchell-as (approved):
Nicely done! Just one question for clarification and this is good to go.

cmd/state-svc/internal/messages/messages.go
mitchell-as: Actually, never mind, because this is covered in the last AC. I'm good.

MDrakos: Yeah, I waffled a bit on this too until I reread the AC.

PR #3602, opened 2024-11-19
Added "--portable" flag to `state checkout` for copying runtime files instead of linking them.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3159" title="DX-3159" target="_blank">StoryDX-3159</a>  As a User I want all of my required files to go to the runtime path if selected
  </td></table>
  <br />
 

Also make `state deploy` runtimes portable by default.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -255,6 +255,8 @@ func (p *Project) Lock() string { return p.projectfile.Lock }
 // Cache returns the cache information for this project
 func (p *Project) Cache() string { return p.projectfile.Cache }
 
+func (p *Project) IsPortable() bool { return p.projectfile.Portable && p.projectfile.Cache != "" }
+
 // Namespace returns project namespace
 func (p *Project) Namespace() *Namespaced {
 	return &Namespaced{Owner: p.projectfile.Owner(), Project: p.projectfile.Name()}
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -255,7 +255,7 @@ func (p *Project) Lock() string { return p.projectfile.Lock }
 // Cache returns the cache information for this project
 func (p *Project) Cache() string { return p.projectfile.Cache }
 
-func (p *Project) IsPortable() bool { return p.projectfile.Portable && p.projectfile.Cache != "" }
+func (p *Project) IsPortable() bool { return p.projectfile.Portable }
 
 // Namespace returns project namespace
 func (p *Project) Namespace() *Namespaced {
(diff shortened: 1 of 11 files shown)


mitchell-as:
I think the solution is just to always deploy by copy. Can it be that simple? Or is there a catch?

pkg/project/project.go
```diff
@@ -255,6 +255,8 @@ func (p *Project) Lock() string { return p.projectfile.Lock }
 // Cache returns the cache information for this project
 func (p *Project) Cache() string { return p.projectfile.Cache }
 
+func (p *Project) IsPortable() bool { return p.projectfile.Portable && p.projectfile.Cache != "" }
```
Naatan, line 258: Why does Cache impact `IsPortable()`? I think the two are compatible.

Naatan:
> I think the solution is just to always deploy by copy. Can it be that simple? Or is there a catch?

No I think it is that simple. The higher pointing was IIRC because we weren't sure what impact this would have on the architecture. It appears not much..

Naatan (changes requested):

pkg/project/project.go
mitchell-as: `state checkout org/project --portable` would end up putting a portable runtime in the local cache directory. To me that doesn't seem particularly useful. If a user did that, they might wonder what exactly that means. I feel like `--portable` is an add-on to `--runtime-path`. As a user, that's how I would see it at least. I'm not wedded to the idea though. I'll leave it up to you.

Naatan: > `state checkout org/project --portable` would end up putting a portable runtime in the local cache directory. To me that doesn't seem particularly useful. If a user did that, they might wonder what exactly that means. I feel like `--portable` is an add-on to `--runtime-path`. As a user, that's how I would see it at least. I'm not wedded to the idea though. I'll leave it up to you.

You're probably right in that most use-cases of `--portable` would probably want to be combined with `--runtime-path`, but that's a layer of flexibility that we should leave up to the user. There's no reason I can see why we would want to force the two flags together.

There could be other use-cases, eg. a runtime that doesn't like being symlinked. Not all tools are symlink aware. 

So unless there is some overhead you're worried about I'd say let's unlink them.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3602#discussion_r1853055605

mitchell-as:
Okay, I hope I understood. If not, please be patient with me. It's been a long day.

PR #3603, opened 2024-11-20
Unify graphQL client code

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3162" title="DX-3162" target="_blank">TaskDX-3162</a>  Unify our GraphQL client code
  </td></table>
  <br />
 

For the PR I've folded our existing `gqlclient` into the `graphql` package. Happy to revisit naming if we feel it's easier the other way around.

The `runWithFiles` method is pulled directly from the old `gqlclient`. There rest of the new functionality is in the `Run` and `RunWithContext`. The rest should just be test fixes and renaming.


Naatan (changes requested):
Looks good! But long as we're unifying and cleaning up, can we please use a name other than `graphql`? It's a bit generic, we already have both a graphql api package and a client package. Suggest using the `gqlclient` name.

Naatan (approved):

PR #3606, opened 2024-11-22
Show prompts, even if running in non-interactive or force mode.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3166" title="DX-3166" target="_blank">StoryDX-3166</a>  Bring back the `--force` flag to allow running an action regardless of prompts OR defaults
  </td></table>
  <br />
 

Prompts were never shown in non-interactive mode, and the previous PR for this ticket didn't change that. This PR changes the prompt library to also show prompts in non-interactive and force mode before the "continuing because non-interactive/--force" notices.

```
% state reset --force      
█ Reset to a Commit

Operating on project org/project, located at 
/path/to/project.

Your project will be reset to 6d79f2ae-f8b5-46bd-917a-d4b2558ec7b8

Resetting is destructive. You will lose any changes that were not pushed. Are 
you sure you want to do this? (y/N)
Continuing because the '--force' flag is set.
█ Sourcing Runtime

```

```
% state install datetime -n     
█ Installing Package

Operating on project org/project, located at /path/to/project.

• Searching for packages in the ActiveState Catalog ............
Multiple Matches
Your query for datetime has multiple matches. Which one would you like to use?

> DateTime (language/perl)
  DateTime (language/python)
Using 'DateTime (language/perl)' because State Tool is running in non-interactive mode.
 ✔ Found
```

### DIFF
--- a/internal/prompt/prompt.go
+++ b/internal/prompt/prompt.go
@@ -106,26 +106,6 @@ func interactiveInputError(message string) error {
 // If the prompt is non-interactive, it returns defaultResponse.
 // If the prompt is forced, it returns forcedResponse if not nil, or defaultResponse.
 func (p *Prompt) InputAndValidate(title, message string, defaultResponse *string, forcedResponse *string, validator ValidatorFunc, flags ...ValidatorFlag) (string, error) {
-	if p.isForced {
-		response := forcedResponse
-		if response == nil {
-			response = defaultResponse
-		}
-		if response != nil {
-			p.out.Notice(locale.Tr("prompt_using_force", *response))
-			return *response, nil
-		}
-		return "", ErrNoForceOption
-	}
-
-	if !p.isInteractive {
-		if defaultResponse != nil {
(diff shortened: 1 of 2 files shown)


internal/prompt/prompt.go
```diff
@@ -139,6 +119,29 @@ func (p *Prompt) InputAndValidate(title, message string, defaultResponse *string
...
+	if !p.isInteractive {
+		nonInteractiveResponse = defaultResponse
+		if nonInteractiveResponse == nil {
+			return "", interactiveInputError(message)
+		}
+	}
+
+	if p.out.Type().IsStructured() {
+		return *nonInteractiveResponse, nil
+	}
```
Naatan, line 143: I think this is addressing what we talked about during stand?

Suggest leaving it out of this PR, so we address it all at once in the story I filed. Otherwise this could create some friction in testing as that's not a change Mark is expecting as part of this story.

This goes for all instances of this logic.

Naatan (changes requested):

Naatan (approved):

PR #3607, opened 2024-11-26
Do not run remote installer test in non-interactive mode.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3173" title="DX-3173" target="_blank">BugDX-3173</a>  Nightly failure: TestRemoteInstallIntegrationTestSuite/TestInstall/install-prbranch-and-channel_(@master) 
  </td></table>
  <br />
 

It turns out we need to be running in non-interactive mode because the Windows installer detects installation as admin and prompts you to continue. I've updated the test to stop trying to send "Y" to the TOS prompt and instead assert it's continuing due to non-interactive mode. There's an extra assertion for Windows so we do not forget this use-case.

### DIFF
--- a/test/integration/remote_installer_int_test.go
+++ b/test/integration/remote_installer_int_test.go
@@ -74,7 +74,7 @@ func (suite *RemoteInstallIntegrationTestSuite) TestInstall() {
 			installPath := filepath.Join(ts.Dirs.Work, "install")
 			stateExePath := filepath.Join(installPath, "bin", constants.StateCmd+osutils.ExeExtension)
 
-			args := []string{"-n"}
+			args := []string{"--non-interactive"}
 			if tt.Version != "" {
 				args = append(args, "--version", tt.Version)
 			}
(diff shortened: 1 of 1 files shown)


test/integration/remote_installer_int_test.go
```diff
@@ -74,7 +74,7 @@ func (suite *RemoteInstallIntegrationTestSuite) TestInstall() {
 			installPath := filepath.Join(ts.Dirs.Work, "install")
 			stateExePath := filepath.Join(installPath, "bin", constants.StateCmd+osutils.ExeExtension)
 
-			args := []string{"-n"}
+			args := []string{"--non-interactive"}
```
MDrakos, line 77: Did `-n` not work?

MDrakos (approved):

test/integration/remote_installer_int_test.go
mitchell-as: It did, but I like that this is more explicit.

PR #3609, opened 2024-11-26
Propagate auth errors from service

MDrakos:


mitchell-as:
It's not clear to me what other authentication issues could happen that ought to be captured by the poller. The ticket does not make it clear which cases we should expect to have happen and handle. The `ACTIVESTATE_API_KEY` env var Mike handled in this PR and which is used in this description seems to be the only one.

Putting this up for early review. Maybe it's enough for now.

Naatan:
@MDrakos I know it's been a while; do you recall if there was still work missing from this PR?

Naatan:
@MDrakos ping

MDrakos:
@Naatan Bah sorry not sure how I lost track of this. I don't think there was any functional work left to do just maybe some improvements do how the code is structured.

PR #3613, opened 2024-12-13
Stop creating new users.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3189" title="DX-3189" target="_blank">TaskDX-3189</a>  Get rid of `CreateNewUser()` in our testing suite
  </td></table>
  <br />
 

The Dashboard requires e-mail confirmation which we cannot do.


mitchell-as:
DX-3191 has been filed for the Scan failure.

PR #3616, opened 2024-12-16
Force prompt option should override non-interactive option.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3193" title="DX-3193" target="_blank">BugDX-3193</a>  Windows ST install fail if install as Administrator
  </td></table>
  <br />
 

Also assert installing as admin prompt happens on Windows.


internal/prompt/prompt.go
```diff
@@ -129,9 +129,7 @@ func (p *Prompt) InputAndValidate(title, message string, defaultResponse *string
 		if nonInteractiveResponse == nil {
 			return "", ErrNoForceOption
 		}
-	}
-
-	if !p.isInteractive {
+	} else if !p.isInteractive {
```
mitchell-as, line 132: The bug here is that we were checking for force and setting `nonInteractiveResponse` but if `!p.isInteractive` is also true (it's usually not), then `nonInteractiveResponse` would be overwritten with a different value.

tjsanterre (approved):

PR #3625, opened 2024-12-24
Turn on non-interactive mode for structured output.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3176" title="DX-3176" target="_blank">StoryDX-3176</a>  `--output json` should always set non-interactive
  </td></table>
  <br />

### DIFF
--- a/cmd/state/internal/cmdtree/cmdtree.go
+++ b/cmd/state/internal/cmdtree/cmdtree.go
@@ -295,7 +295,12 @@ func newStateCommand(globals *globalOptions, prime *primer.Values) *captain.Comm
 				Shorthand:   "o",
 				Description: locale.T("flag_state_output_description"),
 				Persist:     true,
-				Value:       &globals.Output,
+				OnUse: func() {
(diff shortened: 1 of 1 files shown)


mitchell-as:
The lone failure is a sporadic one.

cmd/state/internal/cmdtree/cmdtree.go
```diff
@@ -295,7 +295,12 @@ func newStateCommand(globals *globalOptions, prime *primer.Values) *captain.Comm
 				Shorthand:   "o",
 				Description: locale.T("flag_state_output_description"),
 				Persist:     true,
-				Value:       &globals.Output,
+				OnUse: func() {
+					if prime.Output().Type().IsStructured() {
+						globals.NonInteractive = true
+					}
```
mitchell-as, line 301: This function is called after our outputer is updated to use the type given on the command line, so we know at this time if it's structured (e.g. json) or not (e.g. plain).

PR #3630, opened 2025-01-16
Return a commit even if there's a polling error.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3205" title="DX-3205" target="_blank">BugDX-3205</a>  Nightly failure: CycloneDX and SPDX SBOM imports do not create commits
  </td></table>
  <br />
 

Recent Platform changes have resulted in every stageCommit call responding with "build planning".

`state import` does not care if the plan ultimately fails due to requirements not being found. It should still make the commit.

### DIFF
--- a/pkg/platform/model/buildplanner/commit.go
+++ b/pkg/platform/model/buildplanner/commit.go
@@ -72,7 +72,7 @@ func (b *BuildPlanner) StageCommit(params StageCommitParams) (*Commit, error) {
 	if resp.Build.Status == raw.Planning {
 		resp.Build, err = b.pollBuildPlanned(resp.CommitID.String(), params.Owner, params.Project, nil)
 		if err != nil {
-			return nil, errs.Wrap(err, "failed to poll build plan")
+			return &Commit{resp, nil, nil}, errs.Wrap(err, "failed to poll build plan")
 		}
 	}
 


pkg/platform/model/buildplanner/commit.go
```diff
@@ -72,7 +72,7 @@ func (b *BuildPlanner) StageCommit(params StageCommitParams) (*Commit, error) {
 	if resp.Build.Status == raw.Planning {
 		resp.Build, err = b.pollBuildPlanned(resp.CommitID.String(), params.Owner, params.Project, nil)
 		if err != nil {
-			return nil, errs.Wrap(err, "failed to poll build plan")
+			return &Commit{resp, nil, nil}, errs.Wrap(err, "failed to poll build plan")
```
mitchell-as, line 75: We do the same thing earlier in this function:

https://github.com/ActiveState/cli/blob/2d271e1d91fa4a506569dece28812e3725cd8ae8/pkg/platform/model/buildplanner/commit.go#L65-L67

It's up to the caller to handle error conditions, and `state import` does something with the returned commit, even if there's an error:

https://github.com/ActiveState/cli/blob/2d271e1d91fa4a506569dece28812e3725cd8ae8/internal/runners/packages/import.go#L129-L141

Other callers just error if `err` is not nil. The point is, it's okay to return the commit returned by StageCommit, even if there was ultimately an error (the commit was created after all; it just doesn't build).

tjsanterre (approved):

PR #3632, opened 2025-01-22
Start building on and for ARM.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3211" title="DX-3211" target="_blank">StoryDX-3211</a>  [Vectra] We produce and automate ARM64 Linux builds for State Tool 
  </td></table>
  <br />
 

I'm not sure if we want to merge this into the v48 branch for a "soft" release from CI, or if we want to leave this in PR form for BE to play with.


mitchell-as:
Flagging for review to figure out what we want to do with this (see description). I am aware we might want to let this bake for a while.

Naatan (approved):
Nice work!

PR #3640, opened 2025-02-10
Add unique(ish) ID to error messages for easier Rollbar cross-referencing.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3210" title="DX-3210" target="_blank">StoryDX-3210</a>  For non user-facing input errors we report an error ID
  </td></table>
  <br />
 

Sample:

```
% state checkout org/project
█ Something Went Wrong

  oops!.

█ Need More Help?
 • Run → 'state checkout --help' for general help.
   When doing so, please reference the following error ID: 90cb6932.
```

Note the last line of output.

### DIFF
--- a/internal/runbits/errors/errors.go
+++ b/internal/runbits/errors/errors.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/thoas/go-funk"
 
 	"github.com/ActiveState/cli/internal/analytics"
@@ -30,6 +31,19 @@ type ErrorTips interface {
 
 type OutputError struct {
 	error
+}
+
+var errorId string
+
+// Returns a relatively unique ID for error reporting.
+// We report this ID in non-user-facing errors and on Rollbar so we can cross-reference them.
+// Repeated calls return the same ID.
+func getErrorId() string {
+	if errorId == "" {
+		errorId = uuid.New().String()[:8]
+	}
+
+	return errorId
 }
 
 func (o *OutputError) MarshalOutput(f output.Format) interface{} {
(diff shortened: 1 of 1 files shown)


mitchell-as:
The Scan job failure is likely due to the fact the Go install action hasn't updated to the latest, newly released version of Go. It should go away soon.

internal/runbits/errors/errors.go
```diff
@@ -197,7 +215,7 @@ func ReportError(err error, cmd *captain.Command, an analytics.Dispatcher) {
 	var action string
 	errorMsg := err.Error()
 	if IsReportableError(err) {
-		multilog.Critical("Returning error:\n%s\nCreated at:\n%s", errs.JoinMessage(err), stack)
+		multilog.Critical("Returning error (ID: %s):\n%s\nCreated at:\n%s", getErrorId(), errs.JoinMessage(err), stack)
```
mitchell-as, line 218: This will call `rollbar.Critical`, which will forward our message and error ID to rollbar.

internal/runbits/errors/errors.go
```diff
@@ -32,6 +33,19 @@ type OutputError struct {
 	error
 }
 
+var errorId string
+
+// Returns a relatively unique ID for error reporting.
+// We report this ID in non-user-facing errors and on Rollbar so we can cross-reference them.
+// Repeated calls return the same ID.
```
mitchell-as, line 40: This is fine because there can only ever be one error we report to the user that we'd have to lookup on Rollbar.

internal/runbits/errors/errors.go
```diff
@@ -32,6 +33,19 @@ type OutputError struct {
 	error
 }
 
+var errorId string
+
+// Returns a relatively unique ID for error reporting.
```
mitchell-as, line 38: We don't need a totally unique ID, just something to search for in Rollbar to find the user's log.

SkyLeite (approved):

PR #3647, opened 2025-02-13
Added integration test for `state checkout --portable`.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3227" title="DX-3227" target="_blank">TaskDX-3227</a>  We have an integration test that covers the `--portable` flag / use-case
  </td></table>
  <br />

### DIFF
--- a/test/integration/checkout_int_test.go
+++ b/test/integration/checkout_int_test.go
@@ -395,6 +395,39 @@ func (suite *CheckoutIntegrationTestSuite) TestCheckoutFromArchive() {
 	cp.ExpectExitCode(0)
 }
 
+func (suite *CheckoutIntegrationTestSuite) TestCheckoutPortable() {
+	suite.OnlyRunForTags(tagsuite.Checkout)
+	ts := e2e.New(suite.T(), false)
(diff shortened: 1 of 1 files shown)


test/integration/checkout_int_test.go
```diff
@@ -395,6 +395,39 @@ func (suite *CheckoutIntegrationTestSuite) TestCheckoutFromArchive() {
...
+
+	// Verify there are no symlinks in the runtime.
+	files, err := fileutils.ListDir(targetDir, true)
+	suite.Require().NoError(err)
+	for _, file := range files {
+		path := filepath.Join(targetDir, file.Name())
+		resolvedPath, err := fileutils.ResolvePath(path)
+		suite.Require().NoError(err)
+		suite.Assert().NotContains(resolvedPath, "depot", "runtime file '%s' should not be linked to the depot", file.Name())
+	}
```
Naatan, line 428: If this works, great! But we could theoretically still have symlinks originating from the artifact themselves.

Naatan (approved):

PR #3651, opened 2025-02-18
Recognize build expressions that use `Any()` for a version requirement.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3224" title="DX-3224" target="_blank">TaskDX-3224</a>  Ensure compatibility with new Req format
  </td></table>
  <br />


Naatan (changes requested):
Well that was "easy" 👍 Could you just add a test that verifies our use-case is solved?

Naatan (approved):

PR #3653, opened 2025-02-19
Do not use deprecated powershell command.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3228" title="DX-3228" target="_blank">BugDX-3228</a>  Cannot install State Tool on Windows CI
  </td></table>
  <br />
 

Attempting to use the deprecated command completely breaks the ability to install State Tool v0.47 on systems where that command is removed (e.g. GitHub Actions windows-2019 runner).

PR #3654, opened 2025-02-19
Do not use deprecated powershell command.

mitchell-as:


mitchell-as:
Closing in favor of #3653 since changing the windows runner will likely have more unintended consequences.

PR #3659, opened 2025-02-20
Allow buildscript functions without arguments.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3230" title="DX-3230" target="_blank">TaskDX-3230</a>  Build script parser needs to recognize `Any()` function in `Req()` version requirements.
  </td></table>
  <br />

### DIFF
--- a/pkg/buildscript/raw.go
+++ b/pkg/buildscript/raw.go
@@ -71,5 +71,5 @@ type null struct {
 
 type funcCall struct {
 	Name      string   `parser:"@Ident"`
-	Arguments []*value `parser:"'(' @@ (',' @@)* ','? ')'"`
+	Arguments []*value `parser:"'(' (@@ (',' @@)* ','?)? ')'"`
 }
--- a/test/integration/buildscript_int_test.go
+++ b/test/integration/buildscript_int_test.go
@@ -168,9 +168,8 @@ func (suite *BuildScriptIntegrationTestSuite) TestBuildScriptRequirementVersionA
 	cp := ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
 	cp.ExpectExitCode(0)
 
-	// Enable as part of DX-3230.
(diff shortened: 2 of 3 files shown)


pkg/buildscript/raw.go
```diff
@@ -71,5 +71,5 @@ type null struct {
 
 type funcCall struct {
 	Name      string   `parser:"@Ident"`
-	Arguments []*value `parser:"'(' @@ (',' @@)* ','? ')'"`
+	Arguments []*value `parser:"'(' (@@ (',' @@)* ','?)? ')'"`
```
mitchell-as, line 74: This is basically saying that a function call can have zero or one argument list (consisting of 1 or more arguments).

mitchell-as:
The ARM failure is expected since the Platform doesn't support ARM projects yet.

Naatan (approved):

test/integration/buildscript_int_test.go
```diff
@@ -168,9 +168,8 @@ func (suite *BuildScriptIntegrationTestSuite) TestBuildScriptRequirementVersionA
 	cp := ts.Spawn("config", "set", constants.AsyncRuntimeConfig, "true")
 	cp.ExpectExitCode(0)
 
-	// Enable as part of DX-3230.
-	//cp = ts.Spawn("config", "set", constants.OptinBuildscriptsConfig, "true")
-	//cp.ExpectExitCode(0)
+	cp = ts.Spawn("config", "set", constants.OptinBuildscriptsConfig, "true")
+	cp.ExpectExitCode(0)
```
Naatan, line 172: This intentional for this PR?

mitchell-as: Yes, there was a removed comment indicated this ought to be enabled now. Previously there was a parse error, but this PR fixes that, so enabling this verifies the fix.

PR #3660, opened 2025-02-24
Implement an artifact cache in the depot.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3213" title="DX-3213" target="_blank">StoryDX-3213</a>  Keep a cache of unused artifacts in the depot to speed up checkouts / branch switches
  </td></table>
  <br />
 

We cache up to 500MB of unused artifacts by default, but this size is configurable.

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -493,3 +493,6 @@ const IgnoreEnvEnvVarName = "ACTIVESTATE_CLI_IGNORE_ENV"
 
 // ProgressUrlPathName is the trailing path for a project's build progress.
 const BuildProgressUrlPathName = "distributions"
+
+// RuntimeCacheSizeConfigKey is the config key for the runtime cache size.
+const RuntimeCacheSizeConfigKey = "runtime.cache_size"
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -495,4 +495,4 @@ const IgnoreEnvEnvVarName = "ACTIVESTATE_CLI_IGNORE_ENV"
 const BuildProgressUrlPathName = "distributions"
 
 // RuntimeCacheSizeConfigKey is the config key for the runtime cache size.
-const RuntimeCacheSizeConfigKey = "runtime.cache_size"
+const RuntimeCacheSizeConfigKey = "runtime.cache.size"
--- a/pkg/runtime/depot.go
+++ b/pkg/runtime/depot.go
@@ -19,7 +19,6 @@ import (
 	"github.com/ActiveState/cli/internal/installation/storage"
 	"github.com/ActiveState/cli/internal/logging"
 	configMediator "github.com/ActiveState/cli/internal/mediators/config"
-	"github.com/ActiveState/cli/internal/multilog"
 	"github.com/ActiveState/cli/internal/sliceutils"
 	"github.com/ActiveState/cli/internal/smartlink"
 )
@@ -51,6 +50,8 @@ type artifactInfo struct {
 	InUse          bool  `json:"inUse"`
 	Size           int64 `json:"size"`
 	LastAccessTime int64 `json:"lastAccessTime"`
+
+	id strfmt.UUID // for convenience when removing stale artifacts; should NOT have json tag
 }
 
 type ErrVolumeMismatch struct {
@@ -220,7 +221,10 @@ func (d *depot) DeployViaLink(id strfmt.UUID, relativeSrc, absoluteDest string) 
 		Files:       files.RelativePaths(),
 		RelativeSrc: relativeSrc,
 	})
-	d.recordUse(id)
+	err = d.recordUse(id)
+	if err != nil {
+		return errs.Wrap(err, "Could not record artifact use")
+	}
 
 	return nil
 }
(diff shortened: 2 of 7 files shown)


mitchell-as:
The ARM failure is expected since ARM projects are not yet supported.

The lone Windows failure is a sporadic failure and unrelated to this PR.

internal/constants/constants.go
```diff
@@ -493,3 +493,6 @@ const IgnoreEnvEnvVarName = "ACTIVESTATE_CLI_IGNORE_ENV"
 
 // ProgressUrlPathName is the trailing path for a project's build progress.
 const BuildProgressUrlPathName = "distributions"
+
+// RuntimeCacheSizeConfigKey is the config key for the runtime cache size.
+const RuntimeCacheSizeConfigKey = "runtime.cache_size"
```
Naatan, line 498:
```suggestion
const RuntimeCacheSizeConfigKey = "runtime.cache.size"
```

pkg/runtime/depot.go
```diff
@@ -253,10 +278,28 @@ func (d *depot) DeployViaCopy(id strfmt.UUID, relativeSrc, absoluteDest string)
...
 
 	return nil
 }
 
+func (d *depot) recordUse(id strfmt.UUID) {
+	// Ensure a cache entry for this artifact exists and then update its last access time.
+	if _, exists := d.config.Cache[id]; !exists {
+		size, err := fileutils.GetDirSize(d.Path(id))
+		if err != nil {
+			multilog.Error("Could not get artifact size on disk: %v", err)
```
Naatan, line 291: I'd rather we return it. If this fails something is definitely wrong cause we own the depot. Realistically this should pretty much never fail.

pkg/runtime/depot.go
```diff
@@ -253,10 +278,28 @@ func (d *depot) DeployViaCopy(id strfmt.UUID, relativeSrc, absoluteDest string)
...
+			size = 0
+		}
+		logging.Debug("Recording artifact '%s' with size %.1f MB", id.String(), float64(size)/float64(MB))
+		d.config.Cache[id] = &artifactInfo{Size: size}
+	} else {
+		logging.Debug("Recording use of artifact '%s'", id.String())
+	}
```
Naatan, line 298:
```suggestion
		d.config.Cache[id] = &artifactInfo{Size: size}
	} 
```

Guessing this was for debugging during implementation. Doesn't seem useful to have in the log going forward and would be pretty spammy.

pkg/runtime/depot.go
```diff
@@ -422,3 +465,39 @@ func someFilesExist(filePaths []string, basePath string) bool {
...
 	return false
 }
+
+// removeStaleArtifacts iterates over all unused artifacts in the depot, sorts them by last access
+// time, and removes them until the size of cached artifacts is under the limit.
+func (d *depot) removeStaleArtifacts() {
+	type artifact struct {
+		id   strfmt.UUID
+		info *artifactInfo
+	}
```
Naatan, line 475: Suggest encoding id in artifactInfo and avoiding this type altogether, though I might be missing something here..

pkg/runtime/depot.go
```diff
@@ -422,3 +465,39 @@ func someFilesExist(filePaths []string, basePath string) bool {
...
+
+	sort.Slice(unusedArtifacts, func(i, j int) bool {
+		return unusedArtifacts[i].info.LastAccessTime < unusedArtifacts[j].info.LastAccessTime
+	})
+
+	for _, artifact := range unusedArtifacts {
+		if totalSize <= d.cacheSize {
+			break // done
+		}
+		logging.Debug("Removing cached artifact '%s', last accessed on %s", artifact.id.String(), time.Unix(artifact.info.LastAccessTime, 0).Format(time.UnixDate))
```
Naatan, line 495: The debug line above seems useful to have in the log going forward, but this one would probably be a bit spammy.

pkg/runtime/depot.go
```diff
@@ -422,3 +465,39 @@ func someFilesExist(filePaths []string, basePath string) bool {
...
+		if err := os.RemoveAll(d.Path(artifact.id)); err == nil {
+			totalSize -= artifact.info.Size
+		} else {
+			multilog.Error("Could not delete old artifact: %v", err)
```
Naatan, line 499: Should return this error.

test/integration/runtime_int_test.go
```diff
@@ -212,6 +212,36 @@ func (suite *RuntimeIntegrationTestSuite) TestIgnoreEnvironmentVars() {
...
+	cp.Expect("Downloading")
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("switch", "mingw") // should not remove cached shared/zlib artifact
+	cp.ExpectExitCode(0)
+
+	cp = ts.Spawn("install", "shared/zlib")
+	cp.Expect("Installing")
+	cp.ExpectExitCode(0)
+	suite.Assert().NotContains(cp.Snapshot(), "Downloading", "shared/zlib should have been cached")
```
Naatan, line 232: In addition to this it would be nice if we could also assert that the artifact is still in the depot. Because the "Downloading" contract is very implicit. eg. we could rename "Downloading" to "Obtaining" and not think to touch this test, and it would still be passing.

Naatan (approved):
Nice work! I just have some relatively nitpicky stuff, overall this looks solid to me.

Naatan (changes requested):
Sorry pressed wrong button

internal/constants/constants.go
mitchell-as: I will change this, but note that the AC specified the underscore. We also have `optin.unstable.async_runtime` vs. `optin.unstable.async.runtime`.

pkg/runtime/depot.go
mitchell-as: The depot cache is a map of artifact IDs to artifactInfo. I didn't want to duplicate information, but I can add it as a non-json-tagged field so it's not duplicated in the cache on disk at least.

mitchell-as:
The test failures are known issues and unrelated to this PR.

internal/constants/constants.go
Naatan: Bah you're right, thanks for pointing out. I was trying to recall the consistency of the naming but this could really go either way. Generally I would use dot if there might be more settings that follow under a similar category, eg. maybe we would want `runtime.cache.location`, in which cache `cache_size` would be awkward and inconcistent.

I think let's go with the dots since it does seem plausible we might have more here. And starting to wonder if we should just outlaw underscore for config keys, but that's for another day.

Naatan (approved):
Nice work!

PR #3663, opened 2025-03-03
Go back to setup-state-tool GitHub Action on Windows.

mitchell-as:


.github/workflows/build.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       - # === Install State Tool ===
         name: Install State Tool
         uses: ActiveState/setup-state-tool@v1
-        if: "!contains(matrix.sys.os, 'arm') && runner.os != 'Windows'"
+        if: "!contains(matrix.sys.os, 'arm')"
```
mitchell-as, line 84: Go back to using this action on Windows; we were using it prior to v0.47.0's release. v0.47.0 broke it. Now that v0.47.1 has been released, it will work again.

tjsanterre (approved):

PR #3664, opened 2025-03-03
Remove `state init` language inference.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/DX-3229" title="DX-3229" target="_blank">TaskDX-3229</a>  Remove default language mechanic from `state init`
  </td></table>
  <br />
 

It was too confusing. Just go back to requiring the `--language` argument.


mitchell-as:
The ARM test failure is expected since we do not yet support ARM projects on the Platform.

PR #3683, opened 2025-07-11
Ecosystem implementation

Naatan:


pkg/runtime/setup.go
```diff
@@ -269,11 +281,18 @@ func (s *setup) update() error {
 		}
 	}
 
+	// Tell applicable ecosystems to perform any final uninstall actions.
```
mitchell-as, line 284: I'm not convinced we can have `Apply()` do both uninstall and install at the same time, so I'm calling `Apply()` here after all of the `Remove()` calls, and then again later in this file after all of the `Add()` calls.

Naatan:
@mitchell-as could you open a new PR? I'll close this one. Since I opened it I can't properly review it.

PR #3688, opened 2025-07-17
feat: CP-955 migrate ProjectErrorsTool

samueld-activator:


cmd/state-mcp/internal/toolregistry/tools.go
```diff
@@ -37,4 +41,86 @@ func HelloWorldTool() Tool {
...
+			}
+
+			// at this point we can handle request to create the final output
+			// OR just return the full response as text
+			jsonBytes, _ := json.Marshal(response)
+			p.Output().Print(string(jsonBytes))
+
+			return mcp.NewToolResultText(
+				strings.Join(p.Output().History().Print, "\n"),
+			), nil
```
Naatan, line 123: This is all runner logic. The registry should only be about registering tools and routing input to the runner.

cmd/state-mcp/internal/toolregistry/tools.go
```diff
@@ -37,4 +41,86 @@ func HelloWorldTool() Tool {
...
+												}
+											}
+										}
+									}
+								}
+							}`,
+				QueryVars: map[string]interface{}{
+					"organization": ns.Owner,
+					"project":      ns.Project,
+				},
```
Naatan, line 108: Please use the existing request/response architecture, see:

https://github.com/ActiveState/cli/tree/169471b44a507f115252ba0a1eceb100a163e94d/pkg/platform/api/buildplanner

cmd/state-mcp/internal/toolregistry/tools.go
```diff
@@ -37,4 +41,86 @@ func HelloWorldTool() Tool {
...
+			}
+			response := BuildNodesResponse{}
+			err = runner.Run(request, &response)
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			// at this point we can handle request to create the final output
+			// OR just return the full response as text
+			jsonBytes, _ := json.Marshal(response)
```
Naatan, line 118: We should never drop errors, unless the code in question is for some quick one off use-case. Error handling in Go can be annoying, but for better or worse it is a core part of the language and dropping them often leads to needle in the haystack type problems.

cmd/state-mcp/internal/toolregistry/tools.go
```diff
@@ -37,4 +41,86 @@ func HelloWorldTool() Tool {
...
+			"projecterrors",
+			mcp.WithDescription("Project errors tool - shows errors for a project"),
+			mcp.WithString("namespace", mcp.Description("Project namespace in format 'owner/project'")),
+		),
+		Handler: func(ctx context.Context, p *primer.Values, mcpRequest mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			namespace, _ := mcpRequest.RequireString("namespace")
+
+			ns, err := project.ParseNamespace(namespace)
+			if err != nil {
+				return mcp.NewToolResultError("Invalid namespace format. Use 'owner/project' format."), nil
```
Naatan, line 78: You're dropping the actual error here, might be useful to include that.

cmd/state-mcp/internal/toolregistry/tools.go
```diff
@@ -37,4 +41,86 @@ func HelloWorldTool() Tool {
...
+							}`,
+				QueryVars: map[string]interface{}{
+					"organization": ns.Owner,
+					"project":      ns.Project,
+				},
+			}
+			response := BuildNodesResponse{}
+			err = runner.Run(request, &response)
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
```
Naatan, line 113:
```suggestion
				return mcp.NewToolResultError(errs.JoinMessage(err)), nil
```

When you call `err.Error()` you're only getting the top level error. Each error in Go can be wrapping or packing other errors underneath it. `errs.JoinMessage()` is a helper function that joins the messages of all errors together.

cmd/state-mcp/internal/toolregistry/tools.go
```diff
@@ -37,4 +41,86 @@ func HelloWorldTool() Tool {
...
+		),
+		Handler: func(ctx context.Context, p *primer.Values, mcpRequest mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			namespace, _ := mcpRequest.RequireString("namespace")
+
+			ns, err := project.ParseNamespace(namespace)
+			if err != nil {
+				return mcp.NewToolResultError("Invalid namespace format. Use 'owner/project' format."), nil
+			}
+
+			runner := graphql.New(p, api.ServiceBuildPlanner)
```
Naatan, line 81: You can use: `gqlclient.NewWithOpts()`, eg.

https://github.com/ActiveState/cli/blob/169471b44a507f115252ba0a1eceb100a163e94d/pkg/platform/model/buildplanner/buildplanner.go#L43

pkg/platform/api/graphql/graphql.go
```diff
@@ -6,8 +6,13 @@ import (
...
 
-func New(auth *authentication.Auth) *gqlclient.Client {
-	url := api.GetServiceURL(api.ServiceGraphQL)
+func New(auth *authentication.Auth, services ...api.Service) *gqlclient.Client {
+	var service api.Service = api.ServiceGraphQL
+	if len(services) > 0 {
+		service = services[0]
+	}
+
+	url := api.GetServiceURL(service)
```
Naatan, line 15: The api/graphql package is terribly named; it was our first graphql api, and it was simply called; graphql. It had no distinct name. So this is not a generic graphql api client, this is a specific one. Allowing it to use different endpoints like you have here might work, but goes against the API pattern we're using.

As noted further up, suggest using the same approach as here:

https://github.com/ActiveState/cli/blob/169471b44a507f115252ba0a1eceb100a163e94d/pkg/platform/model/buildplanner/buildplanner.go#L43

We might want to wrap this into its own client inside the api/buildplanner dir. Not sure why we didn't do that. Either way don't let that block you.

Naatan (changes requested):

pkg/platform/api/graphql/graphql.go
samueld-activator: Done

internal/runners/mcp/projecterrors/projecterrors.go
```diff
@@ -0,0 +1,197 @@
...
+
+	"github.com/ActiveState/cli/internal/primer"
+	"github.com/ActiveState/cli/pkg/buildplan"
+	"github.com/ActiveState/cli/pkg/platform/model"
+	"github.com/ActiveState/cli/pkg/platform/model/buildplanner"
+	"github.com/ActiveState/cli/pkg/project"
+)
+
+type ProjectErrorsRunner struct {
+	primer    *primer.Values
```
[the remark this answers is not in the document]
samueld-activator: Added output, auth and svcModel, which are coming from the primer

internal/runners/mcp/projecterrors/projecterrors.go
```diff
@@ -0,0 +1,197 @@
...
+			defer wg.Done()
+			depError, err := CheckWasDependencyError(client, ingredient.LogURL)
+			if err != nil {
+				mu.Lock()
+				errors = append(errors, fmt.Errorf("error checking dependency error for %s: %w", ingredient.Name, err))
+				mu.Unlock()
+				return
+			}
+			ingredient.IsDependencyError = depError
+		}(&(*failedIngredients)[i])
```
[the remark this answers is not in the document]
samueld-activator: I get "invalid operation: cannot index failedArtifacts (variable of type *[]FailedArtifact)"

failedArtifacts is a pointer, and this is why I can't index it. I need to dereference using *, but I can do that before "range artifacts" and use &artifacts[i] here

PR #3690, opened 2025-07-22
Initial ecosystem implementation for JavaScript.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/CP-935" title="CP-935" target="_blank">StoryCP-935</a>  Implement Javascript ecosystem installer in State Tool
  </td></table>
  <br />
 

image

### DIFF
--- a/pkg/runtime/ecosystem/javascript.go
+++ b/pkg/runtime/ecosystem/javascript.go
@@ -54,7 +54,7 @@ func (e *JavaScript) Apply() error {
 		return nil // nothing to do
 	}
 
-	args := []string{"install", "--install-links", "-g"}
+	args := []string{"install", "-g"}
 	for _, arg := range e.packages {
 		args = append(args, arg)
 	}
--- a/pkg/runtime/ecosystem/javascript.go
+++ b/pkg/runtime/ecosystem/javascript.go
@@ -1,6 +1,8 @@
 package ecosystem
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/ActiveState/cli/internal/errs"
@@ -54,11 +56,15 @@ func (e *JavaScript) Apply() error {
 		return nil // nothing to do
 	}
 
+	binDir := filepath.Join(e.runtimePath, "usr", "bin")
 	args := []string{"install", "-g", "--offline"}
 	for _, arg := range e.packages {
 		args = append(args, arg)
 	}
-	_, stderr, err := osutils.ExecSimple(filepath.Join(e.runtimePath, "exec", "npm"), args, nil)
+	env := []string{
+		fmt.Sprintf("PATH=%s%s%s", binDir, string(os.PathListSeparator), os.Getenv("PATH")),
+	}
+	_, stderr, err := osutils.ExecSimple(filepath.Join(binDir, "npm"), args, env)
 	if err != nil {
 		return errs.Wrap(err, "Error running npm: %s", stderr)
 	}
--- a/pkg/runtime/ecosystem/javascript.go
+++ b/pkg/runtime/ecosystem/javascript.go
@@ -54,7 +54,7 @@ func (e *JavaScript) Apply() error {
 		return nil // nothing to do
 	}
 
-	args := []string{"install", "-g"}
+	args := []string{"install", "-g", "--offline"}
 	for _, arg := range e.packages {
 		args = append(args, arg)
 	}
--- a/pkg/runtime/ecosystem/javascript.go
+++ b/pkg/runtime/ecosystem/javascript.go
@@ -36,6 +36,13 @@ func (e *JavaScript) Add(artifact *buildplan.Artifact, artifactSrcPath string) (
 	}
 	var packageName string
 	for _, file := range files {
+		if file.Name() == "runtime.json" {
+			err = injectEnvVar(file.AbsolutePath(), "NPM_CONFIG_PREFIX", "${INSTALLDIR}/usr")
+			if err != nil {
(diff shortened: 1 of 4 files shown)


mitchell-as:
1. I am installing our tar-gzipped artifacts directly. I am NOT pulling down from npm. I am essentially running `npm install /path/to/artifact1.tgz /path/to/artifact2.tgz ...`. It's all local and it is supported by npm.
```
   Synopsis
         npm install [<package-spec> ...]

   [...]

       A package is:

       • a) a folder containing a program described by a package.json
         /configuring-npm/package-json file

       • b) a gzipped tarball containing (a)

   [...]
```
2. Install links does not make symlinks.
```
   install-links

       • Default: false

       • Type: Boolean

       When set file: protocol dependencies that exist outside of the project
       root will be packed and installed as regular dependencies instead of
       creating a symlink. This option has no effect on workspaces.
```
That said, upon closer inspection I don't think I actually need it since I'm installing artifacts directly, and those are not considered to be "file: protocol" dependencies.

3. Thank you for linking the document; I do recall reading it a long time ago, but it was not linked in the ticket, so I kind of forgot about it. That said, I'm not sure how going the package.json route is better than installing tar-gzipped artifacts directly with npm. The latter seems like a shortcut for the former.

mitchell-as:
I removed the unnecessary `--install-links` option and added an `--offline` option to force npm into offline mode to make sure it's not pulling anything in from outside in-case our solve is wrong.

pkg/runtime/ecosystem/javascript.go
```diff
@@ -0,0 +1,66 @@
...
+func (e *JavaScript) Remove(artifact *buildplan.Artifact) error {
+	return nil // TODO: CP-956
+}
+
+func (e *JavaScript) Apply() error {
+	if len(e.packages) == 0 {
+		return nil // nothing to do
+	}
+
+	args := []string{"install", "-g", "--offline"}
```
Naatan, line 57:
```suggestion
	args := []string{"install", "--offline"}
```

This installs it into the global package directory, which the local project will make use of. I imagine you should be able to just drop this.

Naatan (changes requested):

pkg/runtime/ecosystem/javascript.go
mitchell-as: Okay, I've experimented with the `NPM_CONFIG_PREFIX` and it functions like Java's `CLASSPATH` such that we can have runtime-specific npm operations.

We still need `-g` because it will install to the runtime's _node_modules/_ directory instead of making a node_modules in the current working directory and installing to that.

pkg/runtime/internal/envdef/environment.go
```diff
@@ -82,6 +82,19 @@ func (j *VariableJoin) MarshalText() ([]byte, error) {
 	return []byte(res), nil
 }
 
+func (j VariableJoin) MarshalJSON() ([]byte, error) {
```
mitchell-as, line 85: I need this because `MarshalText()` was not being called by the JSON marshaler, resulting in integers (not strings) being written to our injected runtime.json files.

pkg/runtime/ecosystem/javascript.go
Naatan: I'm not grokking from that why the `-g` flag is necessary? To me dropping `-g` would address that ticket?

For docker there are 100% going to be manual stitching steps for now. We shouldn't be forcing things in the State Tool that only address that use-case.

mitchell-as: If I drop `-g`, `state install` will create a _node_modules/_ folder in the current directory that I ran `state install` in. Is that desirable? I don't think so. I think we want the packages to be installed in the runtime cache's _node_modules/_ folder. To make that happen, we need the `-g`.

To put it another way, when you run `state install` from your project directory, do you want a node_modules/ to be there in that directory you've checked out your Platform project in? That's not how it is for our other runtimes. All of their dependencies are located in the runtime cache, which is what `-g` is doing.

Naatan: I see. Isn't that a matter of running it from a different working directory then? I mean the npm execution we do, not the `state install command` itself.

It might also necessitate we create a bogus `package.json`.

Either way I'd consider this in scope for this PR. ie. it's a requirement for this PR to install inside our virtual environment.

Naatan (changes requested):
https://github.com/ActiveState/cli/pull/3690#discussion_r2226822377

pkg/runtime/ecosystem/javascript.go
```diff
@@ -34,6 +36,13 @@ func (e *JavaScript) Add(artifact *buildplan.Artifact, artifactSrcPath string) (
 	}
 	var packageName string
 	for _, file := range files {
+		if file.Name() == "runtime.json" {
+			err = injectEnvVar(file.AbsolutePath(), "NPM_CONFIG_PREFIX", "${INSTALLDIR}/usr")
+			if err != nil {
+				return nil, errs.Wrap(err, "Unable to add NPM_CONFIG_PREFIX to runtime.json")
+			}
+			continue
+		}
```
Naatan, line 45: Why is this needed? The intend is that us running `npm` frees us up from this responsibility.

mitchell-as:
Based on the discussion during stand, it sounds like what I'm doing is okay: packages are installed into the virtual runtime environment instead of the current working directory. Please correct me if I'm wrong.

Naatan (approved):
Nice work! You had me playing catchup, sorry for the churn.

PR #3691, opened 2025-07-23
feat: Add MCP tools and migrate aws/aws-sdk-go to v2

samueld-activator:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/CP-990" title="CP-990" target="_blank">TaskCP-990</a>  Migrate other MCP tools to state tool
  </td></table>
  <br />

### DIFF
--- a/scripts/ci/s3-deployer/main_test.go
+++ b/scripts/ci/s3-deployer/main_test.go
@@ -15,7 +15,7 @@ func init() {
 }
 
 func TestCreateSession(t *testing.T) {
-	createSession()
+	createClient()
 	// succeeds if no panic/exit
 }
 


internal/runners/mcp/downloadlogs/downloadlogs.go
```diff
@@ -0,0 +1,41 @@
...
+	output output.Outputer
+	logUrl string
+}
+
+func New(p *primer.Values, logUrl string) *DownloadLogsRunner {
+	return &DownloadLogsRunner{
+		output: p.Output(),
+		logUrl: logUrl,
+	}
+}
```
Naatan, line 22: This is going to sound pretty pedantic, but I'm a sucker for consistency; helps with maintainbility.

The pattern is that we share context like the primer with the constructor. And for the inputs we use a Param struct. So logUrl would be params.

This enables casting the value through our CLI flag parser. Something we might want to do with our MCP implementation down the road to reduce boilerplate and improve maintainability (fewer moving pieces).

For now it's just for consistency. Which will help us refactor this later on if either we want to teach our MCP implementation how to pass these values, or if we want to hook this up to a CLI.

I won't comment on each runner, but this applies to all runners this PR is introducing.

Example:

https://github.com/ActiveState/cli/blob/1b6ba3929e3d78610fc0f712c9e39f2de6f0a3a0/internal/runners/hello/hello_example.go#L32-L32

You can see the flag parser magic here:

https://github.com/ActiveState/cli/blob/1b6ba3929e3d78610fc0f712c9e39f2de6f0a3a0/cmd/state/internal/cmdtree/hello_example.go#L31-L31

internal/runners/mcp/downloadlogs/downloadlogs.go
```diff
@@ -0,0 +1,41 @@
...
+
+	if response.StatusCode == 200 {
+		body, err := io.ReadAll(response.Body)
+		if err != nil {
+			return fmt.Errorf("error reading response body: %v", err)
+		}
+		runner.output.Print(string(body))
+		return nil
+	} else {
+		return fmt.Errorf("wrong status code fetching logs: %d", response.StatusCode)
```
Naatan, line 39: Fwiw the body often contains an error message in this case. Might be worth including.

internal/runners/mcp/downloadsource/downloadsource.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	if !strings.HasSuffix(fileName, ".tar.gz") && !strings.HasSuffix(fileName, ".tgz") {
+		return fmt.Errorf("file '%s' is not a .tar.gz file", fileName)
+	}
+
+	err = ProcessTarGz(tempFile, runner.targetFile, runner.output)
+	if err != nil {
+		return err
+	}
```
Naatan, line 92: It doesn't seem like you're using tempFile for IO operations, you're doing everything in memory. Couldn't you just pass the reader to ProcessTarGz?

internal/runners/mcp/downloadsource/downloadsource.go
```diff
@@ -0,0 +1,207 @@
...
+	configOptions = append(configOptions, config.WithSharedConfigProfile("sso"))
+	configOptions = append(configOptions, config.WithRegion("us-east-1"))
+
+	awsCfg, err := config.LoadDefaultConfig(context.Background(), configOptions...)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to load AWS config: %w", err)
+	}
+
+	// Create S3 client
+	client := s3.NewFromConfig(awsCfg)
```
Naatan, line 113: Is this going to rely on being authenticated? Might be worth explicitly handling that error.

internal/runners/mcp/downloadsource/downloadsource.go
```diff
@@ -0,0 +1,207 @@
...
+	fileName := filepath.Base(parsedURL.Path)
+	if fileName == "" || fileName == "." {
+		fileName = "downloaded-file"
+	}
+
+	return resp.Body, fileName, nil
+}
+
+func ProcessTarGz(file *os.File, targetFile string, output output.Outputer) error {
+	gzReader, err := gzip.NewReader(file)
```
Naatan, line 155: You can probably reduce this function down significantly by using our unarchiver package:

https://github.com/ActiveState/cli/blob/1b6ba3929e3d78610fc0f712c9e39f2de6f0a3a0/internal/unarchiver/unarchiver.go#L85-L85

Not asking for changes, up to you.

Naatan (changes requested):
Nice work so far!

internal/runners/mcp/downloadsource/downloadsource.go
samueld-activator: Correct, good catch

internal/runners/mcp/downloadsource/downloadsource.go
samueld-activator: if it's not authed, it will fail below when downloading the file

internal/runners/mcp/downloadsource/downloadsource.go
samueld-activator: I took a look at that and it saves to a local file; I am just using the reader here and also I do custom checks, like skipping dirs.

internal/runners/mcp/downloadlogs/downloadlogs.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+		runner.output.Print(string(body))
+		return nil
+	} else {
+		body, err := io.ReadAll(response.Body)
+		if err != nil {
+			return fmt.Errorf("error reading response body: %v", err)
+		}
+		return fmt.Errorf("error fetching logs: status %d, %s", response.StatusCode, body)
+	}
```
Naatan, line 52:
```suggestion
	body, err := io.ReadAll(response.Body)
	if err != nil {
		return fmt.Errorf("error reading response body: %v", err)
	}
	if response.StatusCode != 200 {
		return fmt.Errorf("error fetching logs: status %d, %s", response.StatusCode, body)
	}
	runner.output.Print(string(body))
	return nil
```

Naatan (approved):
Nice work! Minor nit on code structure, up to you what you do with it.

Naatan (dismissed):

Naatan (approved):

PR #3694, opened 2025-07-25
Add `security.reporting` config to disable CVE reporting

Naatan:
I couldn't test on a PR deploy because the CVE service is unavailable.

This adds `state config set security.reporting true|false`.

It defaults to `true`, which means it will report on security related issue (ie. CVEs). Turning this to false will negate `security.prompt.enabled` and `security.prompt.level`.

The primary use-case for this is avoiding disruptions during test cycles where the CVE service may be unavailable.


mitchell-as:
Why do we need this? There's no ticket attached. Is the service itself broken completely?

Naatan:
@mitchell-as It's come up a few times, and I ran into this myself. It was straight up blocking me from testing.

Since we're targeting internal use-cases I'm less concerned with following a rigid story based process for this.

PR #3695, opened 2025-07-29
Initial ecosystem implementation for Rust.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/CP-984" title="CP-984" target="_blank">StoryCP-984</a>  Implement Rust ecosystem installer in State Tool
  </td></table>
  <br />
 

Cargo stores package information and caches in directories with location-specific checksums (e.g. _index.crates.io-6f16d22bbaa15001f_). At the time of writing, our language core has a cargo registry located at _usr/registry/_, but it cannot be used because its checksums were constructed at build-time (e.g. _usr/registry/index/index.crates.io-1949cf8c6b5b557f/_), so Cargo cannot make use of them. Rather than using that registry location (and potentially overwriting things), I've picked _usr/cargo_ (e.g. _usr/cargo/registry/index/index.crates.io-6f16d22bbaa15001f/_ and _usr/cargo/registry/cache/index.crates.io-6f16d22bbaa15001f/_).

State Tool takes an incoming artifact with a Rust crate and copies that crate to the Cargo cache. Whenever Cargo needs that artifact, it will get it from the cache rather than downloading it from crates.io.

While Cargo has an `--offline` switch, it cannot be used because we need to actually construct a local mirror for the crates.io repository with all of the packages, metadata, etc. in order for it to "solve" and fetch the right crates. We'll also have to jump though some more hoops with a runtime-specific _cargo.toml_ config. That's beyond the scope of this PR.

Screenshot 2025-07-29 at 3 10 51 pm

As you can see from the screenshot, I was able to compile and run a Rust project that made use of two packages installed into our runtime environment. If they were downloaded from crates.io, there would be a "DOWNLOADED" notice before the "COMPILING" ones. Since there are none, Cargo used our artifact crates.

Note: this feature only works with the crates.io registry. It will not work with any other registry, private or otherwise.

### DIFF
--- a/pkg/runtime/ecosystem/rust.go
+++ b/pkg/runtime/ecosystem/rust.go
@@ -1,0 +1,129 @@
+package ecosystem
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/ActiveState/cli/internal/locale"
+	"github.com/ActiveState/cli/internal/osutils"
+
+	"github.com/ActiveState/cli/pkg/buildplan"
+)
+
+type Rust struct {
+	runtimeDir     string
+	crateCacheName string // e.g. index.crates.io-6f16d22bbaa15001f
+}
+
+func (e *Rust) Init(runtimePath string, buildplan *buildplan.BuildPlan) error {
+	e.runtimeDir = runtimePath
+
+	var err error
+	e.crateCacheName, err = getCacheDirName(runtimePath)
+	switch {
+	case err != nil:
+		return errs.Wrap(err, "Unable to get cargo registry cache dir")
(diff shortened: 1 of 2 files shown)


pkg/runtime/ecosystem/rust.go
```diff
@@ -0,0 +1,129 @@
...
+func (e *Rust) Init(runtimePath string, buildplan *buildplan.BuildPlan) error {
+	e.runtimeDir = runtimePath
+
+	var err error
+	e.crateCacheName, err = getCacheDirName(runtimePath)
+	switch {
+	case err != nil:
+		return errs.Wrap(err, "Unable to get cargo registry cache dir")
+	case e.crateCacheName == "":
+		return locale.NewError("rust_cache_not_found", "The Rust runtime does not have a cache directory for the crates.io index")
```
mitchell-as, line 30: This only happens if we fail to create the directory. It's not an external or input error.

PR #3696, opened 2025-07-30
Try vendoring dependencies in Rust ecosystem.

mitchell-as:
It looks like we can unpack our crates and trick Cargo into thinking they are vendored dependencies. Cargo does not reach out to crates.io at all. It only looks at what is vendored.

Screenshot 2025-07-30 at 4 31 59 pm


Naatan (approved):
Nice work! Glad you found something.

PR #3698, opened 2025-08-06
Initial ecosystem implementation for .NET/C#.

mitchell-as:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/CP-980" title="CP-980" target="_blank">StoryCP-980</a>  Implement C# ecosystem installer in State Tool
  </td></table>
  <br />
 

The `dotnet` command, provided by the .NET runtime/SDK, functions like `npm` for JavaScript. It replaces nuget's functionality. (This means it is possible to use `dotnet` to pull down packages from the nuget online package registry.) We cannot use nuget in Linux because of it's dependency on Mono, which is a large framework we should absolutely **not** depend on.

However, `dotnet` requires a project to operate in, so we cannot install _*.nupkg_ artifacts directly like we can with npm. Instead, we unpack _*.nupkg_ artifacts into a directory structure `dotnet` recognizes. Then users can build their own projects using our packages.

image

The order of operations is:
- Checkout a .NET Platform project.
- Install a .NET package.
- Start a shell for the project, which sets the `$NUGET_PACKAGES` environment variable that points to our virtual environment's package listing.
- Go to your .NET project and run `dotnet restore`, which will read your .NET project's project file and verify you have all the required dependencies (which should be in our virtual environment).
- Build and run the .NET project.

After the `state install` command, I turned off my network to confirm that `dotnet` can detect and use packages in our virtual environment.

It's a bit unfortunate that `dotnet restore` needs the `--source $NUGET_PACKAGES` argument, but oh well. It's certainly possible that people building against a local package listing have to do something similar to prevent a reachout to the nuget registry, so it may not be so bad.

### DIFF
--- a/pkg/runtime/ecosystem/dotnet.go
+++ b/pkg/runtime/ecosystem/dotnet.go
@@ -1,0 +1,114 @@
+package ecosystem
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ActiveState/cli/internal/errs"
+	"github.com/ActiveState/cli/internal/fileutils"
+	"github.com/ActiveState/cli/internal/unarchiver"
+
+	"github.com/ActiveState/cli/pkg/buildplan"
+)
+
+type DotNet struct {
+	runtimePath string
+	nupkgDir    string
+}
+
+func (e *DotNet) Init(runtimePath string, buildplan *buildplan.BuildPlan) error {
+	e.runtimePath = runtimePath
+	e.nupkgDir = filepath.Join("usr", "nupkg")
+	err := fileutils.MkdirUnlessExists(filepath.Join(e.runtimePath, e.nupkgDir))
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/runtime/ecosystem/dotnet.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type DotNet struct {
+	runtimePath string
+	nupkgDir    string
+}
+
+func (e *DotNet) Init(runtimePath string, buildplan *buildplan.BuildPlan) error {
+	e.runtimePath = runtimePath
+	e.nupkgDir = filepath.Join("usr", "nupkg")
```
mitchell-as, line 22: The default .NET package registry lives in the user's _~/.nuget/packages/_ directory. We can override this with the `$NUGET_PACKAGES` environment variable, and we'll point it to this directory: _<runtime>/usr/nupkg/_.

PR #3712, opened 2025-08-18
DX-3158-Allow publish .whl files

icanhasmath:


emanuelc-activestate (approved):
This change has been built and verified .whl upload on publish

PR #3716, opened 2025-08-25
Initial support for removing dynamically imported artifacts.

mitchell-as:
I could not test Java and Rust because the language cores are not building. However, Rust is pretty much identical to .NET with how we are vendoring artifacts and Java is also similar, but with individual jars instead of directories. I am confident they work as expected.


internal/captain/values.go
```diff
@@ -155,9 +155,9 @@ func (p *PackageValue) Set(s string) error {
 	}
 	switch {
 	case strings.Contains(s, ":"):
-		v := strings.Split(s, ":")
-		p.Namespace = strings.TrimSpace(strings.Join(v[0:len(v)-1], ":"))
-		p.Name = strings.TrimSpace(v[len(v)-1])
+		namespace, name, _ := strings.Cut(s, ":")
```
mitchell-as, line 158: With Java, we're doing things like `state install language/java:junit:junit` and `state uninstall language/java:junit:junit`, so we want to split the string only once on the first ':' and keep 'junit:junit' intact.

pkg/runtime/ecosystem.go
```diff
@@ -36,12 +36,10 @@ func artifactMatchesEcosystem(a *buildplan.Artifact, e ecosystem) bool {
 	return false
 }
 
-func namespacesMatchesEcosystem(namespaces []string, e ecosystem) bool {
-	for _, namespace := range e.Namespaces() {
-		for _, n := range namespaces {
-			if n == namespace {
-				return true
-			}
+func namespaceMatchesEcosystem(namespace string, e ecosystem) bool {
```
mitchell-as, line 39: I simplified this function because it wasn't being used as written, and this form is better for my needs.

pkg/runtime/depot.go
```diff
@@ -286,7 +292,9 @@ func (d *depot) DeployViaCopy(id strfmt.UUID, relativeSrc, absoluteDest string)
 // Track will record an artifact deployment.
 // This is automatically called by `DeployVia*()` functions.
 // This should be called for ecosystems that handle installation of artifacts.
-func (d *depot) Track(id strfmt.UUID, deploy *deployment) error {
+// The artifact parameter is only necessary for tracking dynamically imported artifacts after being
+// added by an ecosystem.
+func (d *depot) Track(id strfmt.UUID, deploy *deployment, artifact *buildplan.Artifact) error {
```
Naatan, line 297:
```suggestion
func (d *depot) Track(artifact *buildplan.Artifact, deploy *deployment) error {
```

Could you try to make this the behaviour? The reason this was using id was to keep it as dumb as possible. Obviously for artifacts that ship has now sailed, I'd prefer we're consistent then in so far as that is possible.

This also means updating the Deploy functions. The real question mark here is going to be the Save() function, which far as I can tell is "abusing" the Track() function in order to clean up the cache. I don't recall specifically what the exact behaviour is there nor can I quickly grok it, but I would suggest spending some time there to try and see if we can remove its reliance on Track(). If this turns into a deep rabbit hole please push back.

pkg/runtime/setup.go
```diff
@@ -553,6 +544,33 @@ func (s *setup) uninstall(id strfmt.UUID) (rerr error) {
 		return errs.Wrap(err, "Could not unload artifact envdef")
 	}
 
+	// If this is a dynamically imported artifact, tell the ecosystem to remove/undeploy it.
+	if artifact, exists := s.depot.config.Cache[id]; exists && artifact.Namespace != "" {
```
Naatan, line 548: Could you update the depot.Exists() function to return the artifact? I'd prefer we don't access private properties on the depot to keep the responsibilities clear.

pkg/runtime/setup.go
```diff
@@ -553,6 +544,33 @@ func (s *setup) uninstall(id strfmt.UUID) (rerr error) {
...
+		if ecosys := filterEcosystemMatchingNamespace(s.ecosystems, artifact.Namespace); ecosys != nil {
+			installedFiles := []string{}
+			// Find record of our deployment
+			if deployments, ok := s.depot.config.Deployments[id]; ok {
```
Naatan, line 552: Please make a public method on the depot to access this. Possibly you could even combine this with the `artifact,extists` logic above seeing as the deployment has the namespace in it. Something like `depot.GetDeployments(id) []Deployment`. Then check the namespace off os those.

pkg/runtime/setup.go
```diff
@@ -553,6 +544,33 @@ func (s *setup) uninstall(id strfmt.UUID) (rerr error) {
...
+				deployments = sliceutils.Filter(deployments, func(d deployment) bool { return d.Path == s.path })
```
Naatan, line 553: Might need `fileutils.ResolveUniquePath()` here. Note I thought we had something like `fileutils.IsEqualPath()` but apparently not, could be worth adding.

internal/captain/values.go
Naatan: Oh joy! I really wish we could think of a better way to represent this. Not for this PR, but starting to prefer just passing a flag at this point, eg. `state install -l java junit:junit`. Or maybe even make it a required argument; `state install java junit:junit`.

Naatan (changes requested):
Nice work! Just some minor churn.

Naatan (approved):
Nice work! Definitely feel like the depot is in a poor state right now due to the mix of responsibilities with DI, and the Deploy methods / Track methods stick out like a sore thumb. But that's a bigger issues, I like that this PR is going as far as reasonable to keep those painpoints minimal.

PR #3717, opened 2025-08-26
Allow configuration of API host

MDrakos:
https://activestatef.atlassian.net/browse/CP-1050

### DIFF
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/ActiveState/cli/internal/rollbar"
 	"github.com/ActiveState/cli/internal/runbits/panics"
 	"github.com/ActiveState/cli/internal/svcctl"
+	"github.com/ActiveState/cli/pkg/platform/api"
 	"github.com/ActiveState/cli/pkg/platform/authentication"
 	"github.com/inconshreveable/mousetrap"
 )
@@ -69,6 +70,7 @@ func main() {
 	}
 	rollbar.SetupRollbar(constants.StateServiceRollbarToken)
 	rollbar.SetConfig(cfg)
+	api.SetConfig(cfg)
 
 	if os.Getenv("VERBOSE") == "true" {
 		logging.CurrentHandler().SetVerbose(true)
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -72,6 +72,12 @@ func main() {
 	rollbar.SetConfig(cfg)
 	api.SetConfig(cfg)
 
+	if err := api.RegisterConfigListener(cfg); err != nil {
+		multilog.Critical("Could not register config listener: %v", errs.JoinMessage(err))
+		exitCode = 1
+		return
+	}
+
 	if os.Getenv("VERBOSE") == "true" {
 		logging.CurrentHandler().SetVerbose(true)
 	}
--- a/cmd/state-svc/main.go
+++ b/cmd/state-svc/main.go
@@ -70,13 +70,8 @@ func main() {
 	}
 	rollbar.SetupRollbar(constants.StateServiceRollbarToken)
 	rollbar.SetConfig(cfg)
-	api.SetConfig(cfg)
 
-	if err := api.RegisterConfigListener(cfg); err != nil {
-		multilog.Critical("Could not register config listener: %v", errs.JoinMessage(err))
-		exitCode = 1
-		return
-	}
+	api.SetConfig(cfg)
 
 	if os.Getenv("VERBOSE") == "true" {
 		logging.CurrentHandler().SetVerbose(true)
--- a/internal/config/instance.go
+++ b/internal/config/instance.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ActiveState/cli/internal/logging"
 	mediator "github.com/ActiveState/cli/internal/mediators/config"
 	"github.com/ActiveState/cli/internal/multilog"
+	"github.com/ActiveState/cli/internal/osutils/lockfile"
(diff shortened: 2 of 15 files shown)


pkg/platform/api/settings.go
```diff
@@ -164,11 +179,35 @@ func getProjectHost(service Service) *string {
 	return &url.Host
 }
 
+func getProjectHostFromConfig() string {
+	cfg, err := config.New()
```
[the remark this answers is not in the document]
MDrakos: Hmm this one is a bit awkward. This package is so low-level that passing the config in here will lead to A LOT of cascading changes. Many function signatures will need to be updated to pass a config instance or an interface. I've added a proposal for DB protection in [this](https://github.com/ActiveState/cli/pull/3717/commits/867a099d276701fccb845de87ea48a5c2807cb8e) commit. 

We could consider something like above with a lockfile, some sort of global registry of config instances (though this doesn't protect us for inter-process access), bite the bullet and do the updates required to pass the config instance around, or some other solution that I'm not thinking of.

internal/config/instance.go
```diff
@@ -123,6 +128,17 @@ func (i *Instance) setWithCallback(key string, valueF func(currentValue interfac
 		return nil
 	}
 
+	pidLock, err := lockfile.NewPidLock(i.lockfile)
+	if err != nil {
+		return errs.Wrap(err, "Could not create lock file for %s", i.lockfile)
+	}
+	defer pidLock.Close()
```
Naatan, line 135: This would close the lock after the connection has been created, which doesn't solve the potential conflict. In order for the lock to be valid it would have to be closed when the config connection is closed, but if that were the case then your use-case would not work either as it would encounter the lock..

I think this might be a scenario where it makes more sense to break our rule of "no globals" given the application of the package (api) is so global, and not contextual. Perhaps in time we can properly solve this once and for all by threading a context everywhere.

So I'd suggest calling something like `api.Init(cfg)` from `main.go`. Just make sure that tests are not using this global, because that would break concurrent tests.

Naatan (changes requested):

Naatan:
@MDrakos oh also, I don't think this is addressing the following AC:

> The state-svc also picks up the new config change

I don't particularly care how, that could be restarting the svc I imagine.

internal/config/instance.go
MDrakos: Maybe I'm misunderstanding how the lockfile library works but I believe it's accessing the lockfile only for database writes and then releasing it. Which should prevent us from running into conflicts between the state tool and the service. I'm realizing now that I'm attempting to solve a problem that already exists so I'll look into the `app.Init(cfg)` approach.

MDrakos:
> @MDrakos oh also, I don't think this is addressing the following AC:
> 
> > The state-svc also picks up the new config change
> 
> I don't particularly care how, that could be restarting the svc I imagine.

Yeah, I tried to call this out in one of the integration tests [here](https://github.com/ActiveState/cli/pull/3717/files#diff-f0ded22a4966c54f117268aa10786113c2235b8dc676b19b9fd010a9106d381bR111). The service will pick up the config change after it restarts, or if the config value was set before the service started. The test above my comment specifically sets the config value before the service starts to try to confirm that.

Naatan:
> > @MDrakos oh also, I don't think this is addressing the following AC:
> > > The state-svc also picks up the new config change
> > 
> > 
> > I don't particularly care how, that could be restarting the svc I imagine.
> 
> Yeah, I tried to call this out in one of the integration tests [here](https://github.com/ActiveState/cli/pull/3717/files#diff-f0ded22a4966c54f117268aa10786113c2235b8dc676b19b9fd010a9106d381bR111). The service will pick up the config change after it restarts, or if the config value was set before the service started. The test above my comment specifically sets the config value before the service starts to try to confirm that.

I could have been more explicit in writing that AC, but the intend is that it picks up the change as it's made, without the user manually restarting the state-svc. It should not require the user to manually restart the svc. If this turns out to be too deep a rabbit hole we can probably descope.

Naatan (changes requested):
Looking good! Just some churn on the svc restarting stuff -- https://github.com/ActiveState/cli/pull/3717#issuecomment-3228921718

cmd/state-svc/main.go
```diff
@@ -69,6 +70,13 @@ func main() {
 	}
 	rollbar.SetupRollbar(constants.StateServiceRollbarToken)
 	rollbar.SetConfig(cfg)
+	api.SetConfig(cfg)
+
+	if err := api.RegisterConfigListener(cfg); err != nil {
+		multilog.Critical("Could not register config listener: %v", errs.JoinMessage(err))
+		exitCode = 1
+		return
+	}
```
Naatan, line 79: You're already calling `api.SetConfig()`, that ought to take care of registering the config listener imo.

pkg/platform/api/settings.go
```diff
@@ -109,6 +118,24 @@ var urlsByService = map[Service]*url.URL{
...
+}
+
+var configuredAPIHost string
+
+func RegisterConfigListener(cfg *config.Instance) error {
+	configMediator.AddListener(constants.APIHostConfig, func() {
+		configuredAPIHost = cfg.GetString(constants.APIHostConfig)
+	})
+	return nil
+}
```
Naatan, line 135:
```suggestion
func RegisterConfigListener(cfg *config.Instance) {
	configMediator.AddListener(constants.APIHostConfig, func() {
		configuredAPIHost = cfg.GetString(constants.APIHostConfig)
	})
}
```

There's no error to handle.

Naatan (changes requested):
Looking good, nearly there!

PR #3718, opened 2025-08-26
Allow configuration of analytics pixel URL

MDrakos:


test/integration/analytics_int_test.go
```diff
@@ -644,6 +644,40 @@ func (suite *AnalyticsIntegrationTestSuite) TestCIAndInteractiveDimensions() {
...
+	for _, e := range events {
+		if e.URL == "https://example.com" {
+			foundCount++
+		}
+	}
+
+	// Because the service has already reported some events with the default configuration values,
+	// we expect to find at least one event with the new configuration values after the service is restarted.
+	suite.Greater(foundCount, 1)
+}
```
Naatan, line 679: Note as written this would pass even if the state-svc didn't pick up the change because it's also counting events from the main state tool invocation.

internal/analytics/analytics.go
```diff
@@ -15,6 +16,16 @@ type Dispatcher interface {
...
 
+var AnalyticsURL string
+
 func init() {
 	configMediator.RegisterOption(constants.ReportAnalyticsConfig, configMediator.Bool, true)
+	configMediator.RegisterOption(constants.AnalyticsPixelOverrideConfig, configMediator.String, "")
+}
+
+func RegisterConfigListener(cfg *config.Instance) error {
+	configMediator.AddListener(constants.AnalyticsPixelOverrideConfig, func() {
```
Naatan, line 27: Far as I can tell this does not speak with the state-svc. So a listener placed on the svc will not fire if the config was changed on the state tool side.

I like this pattern though, and it seems to me we ought to support this across state tool and the svc. It also means the code as is is completely self sufficient aside from the state-svc AC, but if we address that separately this code will automatically get the benefit.

I've filed this as a followup: https://activestatef.atlassian.net/browse/CP-1083, and we'll drop the AC from the current set of stories regarding the svc picking up the change. I don't think it's critical for the milestone we're targeting.

Good thinking on taking this approach!

cmd/state-svc/main.go
```diff
@@ -70,6 +71,12 @@ func main() {
 	rollbar.SetupRollbar(constants.StateServiceRollbarToken)
 	rollbar.SetConfig(cfg)
 
+	if err := analytics.RegisterConfigListener(cfg); err != nil {
+		multilog.Critical("Could not register config listener: %v", errs.JoinMessage(err))
+		exitCode = 1
+		return
+	}
```
Naatan, line 78: Analytics is already using a single instance that is propagated everywhere, so there should be no need for globals. You can do the config listener registration in `analytics.New()`.

internal/analytics/client/sync/reporters/pixel.go
```diff
@@ -4,35 +4,26 @@ import (
...
-}
+type PixelReporter struct{}
 
 func NewPixelReporter() *PixelReporter {
-	var pixelUrl string
-
-	// Attempt to get the value for the pixel URL from the environment.  Fall back to default if that fails
-	if pixelUrl = os.Getenv(constants.AnalyticsPixelOverrideEnv); pixelUrl == "" {
-		pixelUrl = constants.DefaultAnalyticsPixel
-	}
```
Naatan, line 24: We should still respect the env override if it is set.

internal/analytics/analytics.go
```diff
@@ -15,6 +16,16 @@ type Dispatcher interface {
 	Close()
 }
 
+var AnalyticsURL string
```
Naatan, line 19: This should still be recorded on the PixelReporter. Unlike with the API host the use of analytics is not pervasive, it's all a single instance.

Naatan (changes requested):

internal/analytics/analytics.go
MDrakos: I'm _fairly_ certain that the service will pick up this change. When we set a config value the runner [makes a request to the service](https://github.com/ActiveState/cli/blob/master/internal/runners/config/set.go#L90) which in turn [notifies the listeners in the service](https://github.com/ActiveState/cli/blob/master/cmd/state-svc/internal/resolver/resolver.go#L259).

cmd/state-svc/main.go
MDrakos: I can't find a confirmation of this. There is no `analytics.New()`. We have multiple analytics clients and while the async client is created once and passed around throughout the state tool the sync client has multiple instantiations in the state service.

internal/analytics/analytics.go
MDrakos: I've moved this to the pixel reporter but as stated above I can't confirm that there will be a single instance

internal/analytics/analytics.go
Naatan: Ohh nice! I wasn't grokking that yesterday. Thanks for digging! 🙏

cmd/state-svc/main.go
MDrakos: In the state service we have two invocations of `sync.New()` one for the service itself and another for the resolver because I believe the latter is running in a goroutine. See [here](https://github.com/ActiveState/cli/blob/master/cmd/state-svc/main.go#L105) and [here](https://github.com/ActiveState/cli/blob/master/cmd/state-svc/internal/resolver/resolver.go#L88) also it looks like we're only using the pixel reporter in the sync client...

internal/analytics/client/sync/reporters/pixel.go
```diff
@@ -7,32 +7,49 @@ import (
...
+	"github.com/ActiveState/cli/internal/config"
 	"github.com/ActiveState/cli/internal/constants"
 	"github.com/ActiveState/cli/internal/errs"
+
+	configMediator "github.com/ActiveState/cli/internal/mediators/config"
 )
 
 type PixelReporter struct {
-	url string
+	cfg *config.Instance
```
Naatan, line 18: I think we can still keep the url here, no? It's static, it only changes when the config listener fires, at which point we can update the value here. The config listener is set at construction time, so we can just loose the config after the listener is placed.

internal/analytics/client/sync/reporters/pixel.go
```diff
@@ -7,32 +7,49 @@ import (
...
+	switch {
+	case envUrl != "":
+		pixelUrl = envUrl
+	case cfgUrl != "":
+		pixelUrl = cfgUrl
+	default:
+		pixelUrl = constants.DefaultAnalyticsPixel
+	}
+
+	pixelURL, err := url.Parse(pixelUrl)
```
Naatan, line 50: Suggest making this its own function that gets invoked both in the constructor and when the config value changes. You could make that a private method on PixelReporter which then also updates the url.

Naatan (changes requested):

PR #3722, opened 2025-08-29
fix: CP-1086 Specify latest timestamp (vs nil) when retrieving latest ingredient revision

samueld-activator:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/CP-1086" title="CP-1086" target="_blank">BugCP-1086</a>  Error when retrieving some ingredients from project failures
  </td></table>
  <br />

### DIFF
--- a/internal/runners/mcp/projecterrors/projecterrors.go
+++ b/internal/runners/mcp/projecterrors/projecterrors.go
@@ -109,11 +109,11 @@ func (runner *ProjectErrorsRunner) Run(params *Params) error {
 // If found, assume the issue is resolved so that a new build could be retried.
 func CheckDependencyFixes(auth *authentication.Auth, failedArtifacts []*buildplan.Artifact) (map[strfmt.UUID]bool, error) {
 	fixed := make(map[strfmt.UUID]bool)
+	latest, err := model.FetchLatestRevisionTimeStamp(auth)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/runners/mcp/projecterrors/projecterrors.go
```diff
@@ -110,9 +110,13 @@ func (runner *ProjectErrorsRunner) Run(params *Params) error {
 func CheckDependencyFixes(auth *authentication.Auth, failedArtifacts []*buildplan.Artifact) (map[strfmt.UUID]bool, error) {
 	fixed := make(map[strfmt.UUID]bool)
 	for _, artifact := range failedArtifacts {
+		latest, err := model.FetchLatestRevisionTimeStamp(auth)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to fetch latest timestamp: %w", err)
+		}
```
Naatan, line 116:
```suggestion
fixed := make(map[strfmt.UUID]bool)
latest, err := model.FetchLatestRevisionTimeStamp(auth)
if err != nil {
	return nil, fmt.Errorf("Failed to fetch latest timestamp: %w", err)
}
for _, artifact := range failedArtifacts {
```

Avoid doing API calls within a for loop if not needed.

Naatan (changes requested):

PR #3724, opened 2025-08-29
Add config flag support to installers

MDrakos:


internal/keypairs/rsa.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // MinimumRSABitLength is the minimum allowed bit-length when generating RSA keys.
-const MinimumRSABitLength int = 12
```
MDrakos, line 16: This and other RSA changes were required to get unit tests passing... I believe this is due to the update to go 1.24

mitchell-as (approved):

PR #3725, opened 2025-09-02
Allow configuration of notifications URL

MDrakos:
<table>
<td>
  <a href="https://activestatef.atlassian.net/browse/CP-1087" title="CP-1087" target="_blank">StoryCP-1087</a>  The notifications URL can be configured with `state config`
  </td></table>
  <br />


MDrakos:
Integration tests are passing. Unit tests are failing but the fix is in https://github.com/ActiveState/cli/pull/3724

PR #3736, opened 2025-09-09
Re-enable integration test.

mitchell-as:


mitchell-as:
Tests are passing now, so no need to keep skipping it.

PR #3808, opened 2026-06-01
ENG-1372: State Tool forwards platform JWT (+ versioned User-Agent) on build-log-streamer WS

antoine-activestate:
## Summary

Makes newer State Tool versions identify themselves on the build-log-streamer WebSocket Upgrade, so the server (ENG-1246 / ENG-1367) can authorize the stream and monitor client versions. Two things now ride on the Upgrade:

1. **Platform JWT via `Sec-WebSocket-Protocol`.** When authenticated, the client offers `bearer.<jwt>` alongside the real subprotocol `build-log-streamer.activestate.com.v1` (which the server echoes back; its allow-list contains only the real one, so the token never appears in the upgrade *response*). This mirrors the dashboard (ENG-1374), which can't set custom WS request headers — both clients stay symmetric.
2. **Versioned `User-Agent`.** The standard `state/<version>-SHA...; <channel> (<os>; <ver>; <arch>)` agent is now sent on the Upgrade so the server can size/monitor State Tool versions (e.g. the unauthenticated tail) before tightening the gate.

### Scope
The goal here is just that **newer** State Tool versions start sending the token + UA. Forcing *old* shipped versions to stop connecting anonymously (min-version / deprecation strategy) is a separate follow-up — see the ticket's "Long-term" section.

## Design note
The token is threaded as a plain **string** (`prime.Auth().BearerToken()` → `runtime.WithAuthToken` → `Opts.AuthToken` → `buildlog` → `Connect`), so `pkg/runtime` gains no dependency on the `authentication` package. Empty token = anonymous (unchanged behavior). `api.UserAgent()` was extracted as a package-level function (the `*RoundTripper` method delegates) so the WS dialer reuses the exact agent string the HTTP clients use.

## Files
- `pkg/platform/api/buildlogstream/streamer.go` — `Connect(ctx, jwt)` sets subprotocols + User-Agent.
- `pkg/platform/api/api.go` — extract package-level `UserAgent()`.
- `pkg/runtime/{options.go,setup.go,internal/buildlog/buildlog.go}` — thread `AuthToken` through.
- `internal/runbits/runtime/runtime.go` — `WithAuthToken(prime.Auth().BearerToken())`.
- `pkg/platform/api/buildlogstream/streamer_test.go` — **new**; real WS handshake against an httptest server.

## Test plan
- [x] `go build` + gofmt clean across touched packages.
- [x] New `TestConnect_*` tests do a real WS handshake against a mock server and assert the server received: `Sec-WebSocket-Protocol: bearer.<jwt>, build-log-streamer.activestate.com.v1` when authenticated, only the real subprotocol when anonymous, and a `state/` User-Agent in both cases.
- [ ] End-to-end JWT *validation* requires the server-side `Sec-WebSocket-Protocol` reader (TheHomeRepot, currently in the deferred ENG-1367 PR); testable against a PR deploy once that lands.

🔒 ENG-1372

### DIFF
--- a/internal/runbits/runtime/runtime.go
+++ b/internal/runbits/runtime/runtime.go
@@ -257,6 +257,7 @@ func Update(
 		runtime.WithAnnotations(proj.Owner(), proj.Name(), commitID),
 		runtime.WithEventHandlers(pg.Handle, ah.handle),
 		runtime.WithPreferredLibcVersion(prime.Config().GetString(constants.PreferredGlibcVersionConfig)),
+		runtime.WithAuthToken(prime.Auth().BearerToken()),
 	}
 	if opts.Archive != nil {
 		rtOpts = append(rtOpts, runtime.WithArchive(opts.Archive.Dir, opts.Archive.PlatformID, checkout.ArtifactExt))
--- a/pkg/platform/api/api.go
+++ b/pkg/platform/api/api.go
@@ -62,6 +62,12 @@ func (r *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 
 // UserAgent returns the user agent used by the State Tool
 func (r *RoundTripper) UserAgent() string {
+	return UserAgent()
+}
+
+// UserAgent returns the user agent used by the State Tool, including the
+// version (so the server can monitor State Tool versions) plus OS/arch.
+func UserAgent() string {
 	var osVersionStr string
 	osVersion, err := sysinfo.OSVersion()
 	if err != nil {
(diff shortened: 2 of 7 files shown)


antoine-activestate:
Superseded by #3809 — branch renamed `antoines/ENG-1372-ws-subprotocol-jwt` → `antoines/ENG-1372` to shorten the State Tool IPC socket path that was overflowing the macOS/Windows `sun_path` limit and failing the integration-test step. Same commits + the Copilot doc-comment fixes.

PR #3812, opened 2026-06-15
Remediate CVEs.

mitchell-as:
All I did was run `go get` for package versions that fixed CVEs, ran `git mod tidy`, and then `git mod vendor`. No other changes.


mitchell-as:
ARM test failures are expected, and the others are the usual sporadic ones.

PR #3816, opened 2026-06-23
ENG-1635: Decrypt and install encrypted private artifacts in the runtime setup path

mitchell-as:
[ENG-1635: Decrypt and install encrypted private artifacts in the runtime setup path](https://activestatef.atlassian.net/browse/ENG-1635)

The consume side of the private ingredient work ([ENG-1563](https://activestatef.atlassian.net/browse/ENG-1563)). When a project uses an encrypted private ingredient, the State Tool now decrypts it with the organization's key while setting up the runtime, so it installs like any other package. If the key is missing or the key service is unreachable, everything else still installs and the user is told which ingredients were skipped — they're picked up automatically on the next run once the key is available.

The outer artifact downloads and unpacks as usual. Before it's committed to the cache, an encrypted payload (detected by its envelope, not its filename) is decrypted with the org key fetched in `runbits` (ENG-1632) and extracted through the untrusted-source sanitizer (ENG-1640). Decrypted artifacts are kept out of cache eviction. A wrong key fails the install with a clear error naming the artifact; a missing key skips just that ingredient and never blocks the rest of the runtime.

**Deviations:** the inner payload is a tar.gz rather than a wheel/zip, so it can carry symlinks and permissions and matches the existing artifact format; the cache flag is named `Private` rather than the ticket's `Pinned`. Architecture/import tests and the end-to-end `state install` test are deferred (ENG-1642 and a follow-up).

**Base branch:** targets `mitchell/eng-1640` (in review) so the diff is only this change; GitHub will retarget it to `version/0-48-1-RC2` once the upstream PRs land.

Tested for detection, decrypt, wrong-key and missing-key handling, cache-eviction survival, and progress/skip reporting.

### DIFF
--- a/internal/runbits/runtime/runtime.go
+++ b/internal/runbits/runtime/runtime.go
@@ -1,6 +1,7 @@
 package runtime_runbit
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"os"
@@ -22,6 +23,7 @@ import (
 	"github.com/ActiveState/cli/internal/rtutils/ptr"
 	buildscript_runbit "github.com/ActiveState/cli/internal/runbits/buildscript"
 	"github.com/ActiveState/cli/internal/runbits/checkout"
+	"github.com/ActiveState/cli/internal/runbits/orgkey"
 	"github.com/ActiveState/cli/internal/runbits/rationalize"
 	"github.com/ActiveState/cli/internal/runbits/runtime/progress"
 	"github.com/ActiveState/cli/internal/runbits/runtime/trigger"
@@ -36,7 +38,6 @@ import (
 	"github.com/ActiveState/cli/pkg/runtime_helpers"
 	"github.com/ActiveState/cli/pkg/sysinfo"
 	"github.com/go-openapi/strfmt"
-	"golang.org/x/net/context"
 )
 
 func init() {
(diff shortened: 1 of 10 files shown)


internal/runbits/runtime/runtime.go
```diff
@@ -285,6 +289,20 @@ func Update(
...
 
+	// Fetch the organization key for private ingredients, if a key service is configured.
+	orgKeyProvider := orgkey.New(prime.Config(), proj.Owner())
+	if orgKeyProvider.Configured() {
+		defer orgKeyProvider.Close()
+		key, keyID, err := orgKeyProvider.Key(context.Background())
+		if err != nil {
+			prime.Output().Notice(locale.Tl("warn_orgkey_unavailable",
+				"[WARNING]Warning:[/RESET] Could not fetch the organization key: {{.V0}}. Encrypted private ingredients will be skipped and retried once the key service is reachable.",
+				errs.JoinMessage(err)))
```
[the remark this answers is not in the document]
mitchell-as: It's fine as-is.

mitchell-as:
Test failures are known and/or sporadic, and unrelated to this PR.

internal/runbits/runtime/runtime.go
```diff
@@ -285,6 +289,20 @@ func Update(
...
 	rtOpts = append(rtOpts, runtime.WithCacheSize(prime.Config().GetInt(constants.RuntimeCacheSizeConfigKey)))
 
+	// Fetch the organization key for private ingredients, if a key service is configured.
+	orgKeyProvider := orgkey.New(prime.Config(), proj.Owner())
+	if orgKeyProvider.Configured() {
+		defer orgKeyProvider.Close()
+		key, keyID, err := orgKeyProvider.Key(context.Background())
+		if err != nil {
+			prime.Output().Notice(locale.Tl("warn_orgkey_unavailable",
+				"[WARNING]Warning:[/RESET] Could not fetch the organization key: {{.V0}}. Encrypted private ingredients will be skipped and retried once the key service is reachable.",
```
MDrakos: I think the user will have to try a command that updates the runtime in order for these to be retired, so it's not _just_ restarting their service.

MDrakos (changes requested):
One minor change

internal/runbits/runtime/runtime.go
mitchell-as: You're right. I've updated both this message and the one below to have the user run `state refresh` once the key is available in order to install the private artifacts.

MDrakos (approved):

PR #3819, opened 2026-06-24
Verify downloaded artifact checksums.

mitchell-as:
[ENG-1638: Verify artifact checksums and enforce anti-downgrade on private artifacts](https://activestatef.atlassian.net/browse/ENG-1638)

This feature existed previously, but was lost in a runtime refactor.

We don't need the anti-rollback mechanism prescribed by the LLM in the ticket.

### DIFF
--- a/pkg/runtime/setup.go
+++ b/pkg/runtime/setup.go
@@ -2,6 +2,8 @@ package runtime
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


mitchell-as:
Test failures are known, expected, or sporadic. They are unrelated to this PR.

PR #3820, opened 2026-06-24
ENG-1636: Detect re-published private-ingredient content in the depot

mitchell-as:
[ENG-1636: Detect re-published private-ingredient content in the depot](https://activestatef.atlassian.net/browse/ENG-1636)

Part of the private ingredient work ([ENG-1563](https://activestatef.atlassian.net/browse/ENG-1563)). A private artifact can be re-published with new content under the same artifact ID (the timeless, mutable-pointer model). A cache keyed by ID alone would serve the old plaintext forever; this makes the State Tool notice the content changed and re-fetch it.

Each private depot entry now records the build-plan checksum its content was stored under. The cache-hit decision compares that against the build plan and, on a mismatch, re-fetches and re-deploys the artifact (undeploying the stale plaintext first). The comparison is gated on the depot's `Private` flag, so public artifacts and non-private projects are untouched.

Because a re-publish keeps the commit hash unchanged, the hash short-circuits would otherwise skip the check entirely. So `state refresh`/`state pull` now re-solve and re-check content **only when the runtime actually has private artifacts** — routine commands keep the fast path.

Two commits: the depot checksum + content-aware cache-hit, then the refresh/pull re-check.

**Base branch:** targets `mitchell/eng-1638` (where this was branched from, for the checksum primitive); GitHub will retarget to `version/0-48-1-RC2` as the upstream PRs land.

Tested for the checksum comparison (fresh/stale/public/absent/empty), the stored-checksum write, and private-artifact detection by deployment path.

### DIFF
--- a/internal/runbits/runtime/runtime.go
+++ b/internal/runbits/runtime/runtime.go
@@ -57,6 +57,8 @@ type Opts struct {
 
 	ValidateBuildscript bool
 	IgnoreAsync         bool
+
+	CheckForPrivateArtifactUpdates bool
 }
 
 type SetOpt func(*Opts)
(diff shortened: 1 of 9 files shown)


mitchell-as:
We don't need this. Artifact IDs are unique, even for new revisions. The LLM got confused when generating this ticket.

PR #3821, opened 2026-06-25
ENG-1643: Add end-to-end integration test for encrypted publish and install

mitchell-as:
[ENG-1643: Add end-to-end integration test for encrypted publish and install](https://activestatef.atlassian.net/browse/ENG-1643)

The acceptance test for the private ingredient work ([ENG-1563](https://activestatef.atlassian.net/browse/ENG-1563)). It exercises the full encrypted round trip: build and publish a protected package from local source, then install it and confirm the decrypted content lands in the runtime.

`TestPublishBuildEncrypted` publishes a tiny Python package (carrying a unique sentinel) with `state publish --build`, installs it, then reads the decrypted wheel back out of the depot and checks for the sentinel — something only the plaintext could contain. The org key is handed to both sides through an environment variable, so the test needs no HTTPS key service.

Getting the round trip to pass surfaced a few production fixes along the way: `state publish --build` now depends on the `private-builder` (the default builder produced an empty artifact) and wraps only the encrypted payload, the consume side finds and decrypts that payload even when the builder nests it under the install directory, and the payload filename is now a single shared constant across the produce and consume sides.

**Base branch:** targets `mitchell/eng-1634`; GitHub will retarget it to `version/0-48-1-RC2` once the upstream PRs land.

### DIFF
--- a/test/integration/publish_int_test.go
+++ b/test/integration/publish_int_test.go
@@ -556,10 +556,82 @@ func (suite *PublishIntegrationTestSuite) TestPublishBuildEncrypted() {
 	cp.Expect("All dependencies have been installed and verified", e2e.RuntimeBuildSourcingTimeoutOpt)
 	cp.ExpectExitCode(0)
 
-	// Decryption proof: the decrypted wheel must be present in the depot and
-	// contain our sentinel. A failed decrypt would skip the artifact, leaving no
-	// wheel behind.
-	suite.assertDecryptedWheelContains(ts, sentinel)
+	// Decryption proof: the decrypted content must be present in the depot and
+	// contain our sentinel. A failed decrypt would skip the artifact, leaving the
+	// sentinel absent. This is inlined (rather than in a helper) so a failure
+	// surfaces the spawned-command output and state logs the e2e harness dumps.
+	// The search is deliberately broad — any wheel or file anywhere under the
+	// depot — because the exact on-disk path depends on how the artifact is
+	// packaged on install.
+	depot := filepath.Join(ts.Dirs.Cache, "depot")
+
+	// Log the depot's top two levels so a failure shows how the artifact landed
+	// (e.g. an "install" dir with a wheel vs. a leftover "payload.enc").
+	if entries, err := os.ReadDir(depot); err == nil {
+		for _, e := range entries {
+			suite.T().Logf("depot/%s", e.Name())
+			if e.IsDir() {
+				if sub, err := os.ReadDir(filepath.Join(depot, e.Name())); err == nil {
+					for _, s := range sub {
+						suite.T().Logf("depot/%s/%s", e.Name(), s.Name())
+					}
+				}
+			}
+		}
+	} else {
+		suite.T().Logf("could not read depot %s: %v", depot, err)
+	}
+
+	var wheels []string
+	fileCount := 0
+	found := false
+	walkErr := filepath.WalkDir(depot, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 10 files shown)


mitchell-as:
The test failures are sporadic ones around the publish endpoint. The only test that matters is the new one, which passes:

```
✅ TestPublishIntegrationTestSuite/TestPublishBuildEncrypted (32.48s)
```

test/integration/publish_int_test.go
```diff
@@ -472,6 +479,175 @@ authors:
...
+		return false
+	}
+	defer zr.Close()
+	for _, f := range zr.File {
+		rc, err := f.Open()
+		if err != nil {
+			continue
+		}
+		content, _ := io.ReadAll(rc)
+		rc.Close()
```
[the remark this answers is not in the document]
mitchell-as: Done in 4b8fdb7f4 — `wheelContains` now captures the `io.ReadAll` error and skips that zip entry rather than testing partial bytes.

test/integration/publish_int_test.go
```diff
@@ -472,6 +479,175 @@ authors:
...
+		fileCount++
+
+		// A decrypted wheel is a zip; scan its entries for the sentinel.
+		if strings.HasSuffix(d.Name(), ".whl") {
+			wheels = append(wheels, path)
+			if suite.wheelContains(path, sentinel) {
+				found = true
+			}
+			return nil
+		}
```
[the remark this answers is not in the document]
mitchell-as: Done in 4b8fdb7f4 — the walk returns `filepath.SkipAll` once the sentinel is found, so it stops instead of scanning the rest of the depot.

mitchell-as:
Note: "install" is misleading. State Tool installs the decrypted wheel by linking to it from the runtime for now. We only recently learned that we want to do a proper wheel install into the runtime, and that will be in its own ticket.

test/integration/publish_int_test.go
```diff
@@ -472,6 +479,181 @@ authors:
...
+	))
+	cp.Expect("Successfully published")
+	cp.ExpectExitCode(0)
+
+	// Install the freshly published ingredient, forcing resolution at the current
+	// timestamp so the new revision is picked up rather than a cached solve.
+	cp = ts.SpawnWithOpts(e2e.OptArgs(
+		"install", ingredientNamespace+":"+ingredientName, "--ts=now",
+	))
+	cp.Expect("All dependencies have been installed and verified", e2e.RuntimeBuildSourcingTimeoutOpt)
```
MDrakos, line 556: In other PRs I recall that if the private ingredient failed at this stage there would be a warning. Is that still the case? Can we be sure that decrypt and install happened correctly here for the private ingredient? Especially now that it isn't a STAR.

MDrakos: I see I just needed to scroll down further to get part of my answer. I assume this test still doesn't cover the actual installation of the wheel yet? If not, we will want to ensure this test is expanded once that functionality lands.

MDrakos (approved):
Looks good to me, see my minor comments.

Another idea would be to expand this test to a slightly more complex python package for another layer of correctness.

test/integration/publish_int_test.go
mitchell-as: You are correct that this PR does not install the wheel -- this code was written before we knew about that requirement. As mentioned in my previous comment above your review, there's a follow-up ticket for installing the wheel. In that PR this test will be updated to verify the wheel was installed.

I am under the impression any ingredient installation failure, private or not, is a runtime failure, not a warning. The runtime setup mechanic didn't have the ability to distinguish private ingredients before.

PR #3832, opened 2026-07-06
Speed up file installation on Windows

icanhasmath:
## Why

Installing runtimes with the CLI is noticeably slow on Windows. Tracing the install pipeline (download → unpack into depot → deploy → transform → track) surfaced two Windows-specific costs in the deploy step, both per-file:

1. **Redundant path-resolution syscalls.** `smartlink.Link` re-resolved *both* src and dest via `ResolveUniquePath` for every entry it recursed into. On Windows that calls `GetLongPathName` — a syscall — so a runtime with many files paid ~2 extra syscalls per file.
2. **Serialized deploys.** `DeployViaLink` held the depot-wide `fsMutex` for the entire link operation, so despite the install worker pool (`maxConcurrency = 5`), file linking ran one artifact at a time. On Windows each link is an individual, relatively slow syscall, so serializing them hurts.

## What changed

**Resolve link paths once per tree instead of once per file** ([smartlink.go](../blob/windows-install-perf/internal/smartlink/smartlink.go))
- Split the recursion into an unexported `link()` worker that assumes already-resolved inputs and does not re-resolve. The root is resolved once; descendant paths are resolved-parent + real (long) entry name, so they need no further resolution.
- Symlink-to-file semantics are preserved exactly (still links to the dereferenced target); the circular-symlink test still passes.
- An "already exists" error from `linkFile` is now treated the same as the existing `TargetExists` skip, so concurrent deploys of a shared file degrade to the documented "first one wins" behavior rather than failing.

**Deploy linked artifacts concurrently** ([depot.go](../blob/windows-install-perf/pkg/runtime/depot.go))
- Dropped `fsMutex` from `DeployViaLink` so linking parallelizes across the worker pool. Safe because linking touches no shared in-memory depot state (`Exists` guards its own map via `mapMutex`, `depotPath` is immutable), dir creation is idempotent, and smartlink now skips already-existing targets.
- `DeployViaCopy` **keeps** the lock: two artifacts copying the same file would race on `os.Create` and could corrupt the destination. Copy is the less common path (file transforms / no hard links / portable mode).

## Testing

- `go build` + `go test` pass for `internal/smartlink` and `pkg/runtime` (including the circular-symlink test).
- ⚠️ Not yet exercised on a real Windows install. The syscall reduction and added parallelism are mechanically certain wins; a timed install on Windows would confirm the magnitude.

## Deliberately out of scope (possible follow-ups)
- Windows Defender exclusion for the depot/install dir — likely the single biggest real-world factor, but it means modifying the user's AV config (admin, invasive) and is a product decision.
- Avoiding the double file-write on transforms, and collapsing the redundant per-artifact tree walks (`ListDir` + `GetDirSize`).

### DIFF
--- a/pkg/runtime/depot.go
+++ b/pkg/runtime/depot.go
@@ -214,10 +214,14 @@ func (d *depot) MarkPrivate(id strfmt.UUID) error {
 
 // DeployViaLink will take an artifact from the depot and link it to the target path.
 // It should return deployment info to be used for tracking the artifact.
+//
+// Unlike DeployViaCopy this does NOT hold d.fsMutex for the duration of the operation, so multiple
+// artifacts can be linked into the runtime concurrently (the install worker pool runs several at once).
+// This is safe because linking touches no shared in-memory depot state (d.Exists guards its own access
+// via mapMutex, and d.depotPath is immutable after construction), directory creation is idempotent, and
+// smartlink treats an already-existing target as a skip. This is a major win on Windows where each file
+// link is an individual, relatively expensive syscall.
 func (d *depot) DeployViaLink(id strfmt.UUID, relativeSrc, absoluteDest string) (*deployment, error) {
-	d.fsMutex.Lock()
-	defer d.fsMutex.Unlock()
-
 	if exists, _ := d.Exists(id); !exists {
 		return nil, errs.New("artifact not found in depot")
 	}
@@ -262,6 +266,11 @@ func (d *depot) DeployViaLink(id strfmt.UUID, relativeSrc, absoluteDest string) 
 
 // DeployViaCopy will take an artifact from the depot and copy it to the target path.
 // It should return deployment info to be used for tracking the artifact.
+//
+// This intentionally holds d.fsMutex for the duration of the copy. Unlike linking, copying a file that
+// two artifacts both provide would race on os.Create and could corrupt the destination, so copies are
+// serialized. The copy path is only used when an artifact needs file transforms, hard links are
+// unsupported, or portable mode is requested, so it is the less common case.
 func (d *depot) DeployViaCopy(id strfmt.UUID, relativeSrc, absoluteDest string) (*deployment, error) {
 	d.fsMutex.Lock()
 	defer d.fsMutex.Unlock()
--- a/internal/smartlink/smartlink.go
+++ b/internal/smartlink/smartlink.go
@@ -28,8 +28,12 @@ func LinkContents(src, dest string) error {
 	if err != nil {
 		return errs.Wrap(err, "Reading dir %s failed", src)
 	}
+	// src and dest are already resolved above, so recurse via link() which does not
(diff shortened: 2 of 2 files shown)


mitchell-as:
FYI I ran the "Test: critical" integration tests to make sure this isn't breaking anything unexpected. The failures that did occur are sporadic and unrelated to this PR.

pkg/runtime/depot.go
```diff
@@ -214,10 +214,14 @@ func (d *depot) MarkPrivate(id strfmt.UUID) error {
 
 // DeployViaLink will take an artifact from the depot and link it to the target path.
 // It should return deployment info to be used for tracking the artifact.
+//
+// Unlike DeployViaCopy this does NOT hold d.fsMutex for the duration of the operation, so multiple
+// artifacts can be linked into the runtime concurrently (the install worker pool runs several at once).
+// This is safe because linking touches no shared in-memory depot state (d.Exists guards its own access
+// via mapMutex, and d.depotPath is immutable after construction), directory creation is idempotent, and
+// smartlink treats an already-existing target as a skip. This is a major win on Windows where each file
```
mitchell-as, line 222: I am not terribly comfortable with this assertion. It's probably referring to this code in smartlink.go:

```go
	// Multiple artifacts can supply the same file. We do not have a better solution for this at the moment other than
	// favouring the first one encountered.
	if fileutils.TargetExists(dest) {
		logging.Warning("Skipping linking '%s' to '%s' as it already exists", src, dest)
		return nil
	}
```

If we remove the fs mutex and then come up with a better solution, we lose the mutex guarantee. If we want to remove the mutex, we should update the referenced code's comment to mention the lack of an existing mutex (the one we're removing).

I like the additional comment above `DeployViaCopy` that we need the mutex to avoid copying to the same destination, but I'm not convinced we can forgo the need for mutex when creating links.

MDrakos: We could try making `fsMutex` an `RWMutex` and have `DeployViaLink` take an `RLock` and `DeployViaCopy/Put/Undeploy` take `Lock`. This way the safer operations which are linking can be done in parallel but copying which isn't as safe still has to be sequential.

MDrakos (commented):
This PR needs some tests, especially to test the concurrency element which this PR addresses.

Some ideas from my robot all run with `-race`:
- A deterministic unit test for the new os.IsExist branch: pre-create the target,
  call Link, assert it returns nil and leaves the existing target intact.
- A concurrent same-file link test in smartlink (two goroutines, same dest) to hit
  the TargetExists -> linkFile -> os.IsExist path.
- If we take the RWMutex: a pkg/runtime test deploying a copy artifact and a link
  artifact that share a file concurrently, asserting the depot source stays
  byte-for-byte identical — the regression guard for the truncation case.
- A quick equivalence test for the Link refactor (regular dir / symlink->file /
  symlink->dir), since only the circular case is covered today.

PR #3833, opened 2026-07-08
Added run script for testing encrypted private ingredients.

mitchell-as:
This is for local testing purposes. I know it's not a deliverable for the initiative.

```
% state run serve-org-keys ActiveState-CLI-Testing

Operating on project ActiveState/cli, located at /path/to/cli.

█ Running Script: serve-org-keys

█ Script Output

Generating HTTPS certificate.
Starting org key server.
Using random encryption key (see below)
--key Ndt9NfWAEOq6FMyq3EKaPtnF7CypxXr0SxkKGqiiTCg=
serving /v1/org-key for org 'ActiveState-CLI-Testing' on https://127.0.0.1:8443

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ cmd/state/versioninfo.json
 cmd/state/resource.syso
 cmd/state-svc/versioninfo.json
 cmd/state-svc/resource.syso
+test/ssl/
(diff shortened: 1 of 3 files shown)


MDrakos (approved):
We actually might need something like this when we enable CS to do some of their own testing.

PR #3839, opened 2026-07-14
serve-org-keys: run under cmd (Windows) without bash or openssl

icanhasmath:
## What

Makes the `serve-org-keys` run script work in a Windows **cmd** shell, not just bash.

## Why

`serve-org-keys` was a single `language: bash` script. On a cmd shell it either wouldn't run or relied on tools (`openssl`, `python3`) that on Windows only come from git-bash. This lets the script run in a pure cmd environment.

## Changes

- **`activestate.yaml`**
  - Constrained the existing bash `serve-org-keys` to `if: ne .Shell "cmd"`.
  - Added a `language: batch` twin constrained to `if: eq .Shell "cmd"`. The State Tool filters by shell constraint (`FilterUnconstrained`), so the right one resolves automatically per shell — same mechanism `install-deps-dev` uses.
  - The batch variant detects the interpreter (`python` → `py -3` → `python3`) and drops the `openssl` dependency.
- **`scripts/orgkeyserver.py`**
  - Added `--gen-cert`, which generates the self-signed cert/key itself via the `cryptography` package (SAN `DNS:localhost, IP:127.0.0.1`), so no `openssl` binary is needed. Falls back to a clear error if `cryptography` isn't installed.

## Verification

- Cert generation: valid X.509, correct SAN (deduped), cert/key modulus match.
- Server starts and serves `/v1/org-key` with the generated cert.
- `ast.parse` clean; `--help` and required-arg enforcement work.

## Notes

- The bash variant is unchanged and still uses `openssl` (always present where bash is).
- The cmd path trades the `openssl` dependency for `cryptography`; since Python is already required, `pip install cryptography` is the trivial follow-up if absent.

### DIFF
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -499,19 +499,21 @@ scripts:
       )
       set "key=%~2"
 
-      echo Generating HTTPS certificate.
-      if not exist test\ssl mkdir test\ssl
-      openssl req -x509 -newkey rsa:2048 -nodes -days 365 ^
-        -keyout test\ssl\key.pem -out test\ssl\cert.pem ^
-        -subj "/CN=localhost" ^
-        -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" 2> nul
+      set "PY="
+      where python >nul 2>nul && set "PY=python"
+      if not defined PY (where py >nul 2>nul && set "PY=py -3")
+      if not defined PY (where python3 >nul 2>nul && set "PY=python3")
+      if not defined PY (
+        echo Error: no Python interpreter found on PATH ^(tried python, py, python3^).
+        exit /b 1
+      )
 
-      echo Starting org key server.
+      echo Generating HTTPS certificate and starting org key server.
       if "%key%"=="" (
         echo Using random encryption key ^(see below^)
-        python3 -- scripts\orgkeyserver.py --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org %org%
+        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org %org%
       ) else (
-        python3 -- scripts\orgkeyserver.py --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%" --key "%key%"
+        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%" --key "%key%"
       )
 
 events:
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -472,19 +472,12 @@ scripts:
       fi
       key="$2"
 
-      echo "Generating HTTPS certificate."
-      if [ ! -d test/ssl ]; then mkdir test/ssl; fi
-      openssl req -x509 -newkey rsa:2048 -nodes -days 365 \
-        -keyout test/ssl/key.pem -out test/ssl/cert.pem \
-        -subj "/CN=localhost" \
-        -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" 2> /dev/null
-
       echo "Starting org key server."
       if [ -z "$key" ]; then
         echo "Using random encryption key (see below)"
-        python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org "$org"
+        python3 -- scripts/orgkeyserver.py --org "$org"
       else
-        python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org "$org" --key "$key"
+        python3 -- scripts/orgkeyserver.py --org "$org" --key "$key"
       fi
   - name: serve-org-keys
     language: batch
@@ -508,12 +501,12 @@ scripts:
         exit /b 1
       )
 
-      echo Generating HTTPS certificate and starting org key server.
+      echo Starting org key server.
       if "%key%"=="" (
         echo Using random encryption key ^(see below^)
-        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%"
+        %PY% scripts\orgkeyserver.py --org "%org%"
       ) else (
-        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%" --key "%key%"
+        %PY% scripts\orgkeyserver.py --org "%org%" --key "%key%"
       )
 
 events:
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -482,7 +482,7 @@ scripts:
       echo "Starting org key server."
       if [ -z "$key" ]; then
         echo "Using random encryption key (see below)"
-        python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org $org
+        python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org "$org"
       else
         python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org "$org" --key "$key"
       fi
@@ -511,7 +511,7 @@ scripts:
       echo Generating HTTPS certificate and starting org key server.
       if "%key%"=="" (
         echo Using random encryption key ^(see below^)
-        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org %org%
+        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%"
       ) else (
         %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%" --key "%key%"
       )
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -461,6 +461,7 @@ scripts:
     value: go run $project.path()/scripts/to-buildexpression/main.go $@
   - name: serve-org-keys
     language: bash
+    if: ne .Shell "cmd"
     description: Runs a server that serves org keys for testing encrypted private ingredients
     value: |
       org="$1"
@@ -470,14 +471,14 @@ scripts:
         exit 1
       fi
       key="$2"
-      
+
       echo "Generating HTTPS certificate."
       if [ ! -d test/ssl ]; then mkdir test/ssl; fi
       openssl req -x509 -newkey rsa:2048 -nodes -days 365 \
         -keyout test/ssl/key.pem -out test/ssl/cert.pem \
         -subj "/CN=localhost" \
         -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" 2> /dev/null
-        
+
       echo "Starting org key server."
       if [ -z "$key" ]; then
         echo "Using random encryption key (see below)"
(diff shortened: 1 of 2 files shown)


activestate.yaml
```diff
@@ -470,21 +471,50 @@ scripts:
...
         -keyout test/ssl/key.pem -out test/ssl/cert.pem \
         -subj "/CN=localhost" \
         -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" 2> /dev/null
-        
+
       echo "Starting org key server."
       if [ -z "$key" ]; then
         echo "Using random encryption key (see below)"
         python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org $org
       else
```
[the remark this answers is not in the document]
icanhasmath: Fixed in 1103c8c1b — quoted `--org "$org"` in the no-key branch to match the else branch and avoid word-splitting/expansion.

activestate.yaml
```diff
@@ -470,21 +471,50 @@ scripts:
...
+      if not defined PY (
+        echo Error: no Python interpreter found on PATH ^(tried python, py, python3^).
+        exit /b 1
+      )
+
+      echo Generating HTTPS certificate and starting org key server.
+      if "%key%"=="" (
+        echo Using random encryption key ^(see below^)
+        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org %org%
+      ) else (
```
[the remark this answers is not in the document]
icanhasmath: Fixed in 1103c8c1b — quoted `--org "%org%"` in the cmd no-key branch so spaces and cmd metacharacters (e.g. `&`) are passed as data.

scripts/orgkeyserver.py
```diff
@@ -40,6 +46,68 @@
...
+    if key_dir:
+        os.makedirs(key_dir, exist_ok=True)
+    with open(key_path, "wb") as f:
+        f.write(key.private_bytes(
+            serialization.Encoding.PEM,
+            serialization.PrivateFormat.TraditionalOpenSSL,
+            serialization.NoEncryption(),
+        ))
+    with open(cert_path, "wb") as f:
+        f.write(cert.public_bytes(serialization.Encoding.PEM))
```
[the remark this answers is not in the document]
icanhasmath: Fixed in 1103c8c1b — the generated key is now `chmod`ed to 0600 (verified 600 on POSIX; harmless no-op on Windows). This also matches the epic's key-at-rest `0600` decision.

scripts/orgkeyserver.py
```diff
@@ -11,6 +11,11 @@
         -subj "/CN=localhost" \
         -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
 
+Or let this script generate one for you (needs the 'cryptography' package, no
+openssl binary required -- useful on native Windows / cmd):
+
+    python3 scripts/orgkeyserver.py --gen-cert --tls-cert cert.pem --tls-key key.pem
+
```
mitchell-as: I'm going to propose we strike these and always use Python to generate the cert.

scripts/orgkeyserver.py
```diff
@@ -40,6 +46,68 @@
...
+        import ipaddress
+
+        from cryptography import x509
+        from cryptography.hazmat.primitives import hashes, serialization
+        from cryptography.hazmat.primitives.asymmetric import rsa
+        from cryptography.x509.oid import NameOID
+    except ImportError:
+        raise SystemExit(
+            "--gen-cert requires the 'cryptography' package (pip install "
+            "cryptography), or supply --tls-cert/--tls-key generated with openssl."
```
mitchell-as: Since I propose we use Python to always generate the cert, and the runtime should include it, I think we can remove the try except clause here and just assume cryptography is available.

scripts/orgkeyserver.py
```diff
@@ -83,6 +156,9 @@ def main():
     p = argparse.ArgumentParser(description=__doc__.splitlines()[0])
     p.add_argument("--tls-cert", required=True, help="server TLS certificate (PEM)")
     p.add_argument("--tls-key", required=True, help="server TLS private key (PEM)")
+    p.add_argument("--gen-cert", action="store_true",
+                   help="generate a self-signed cert/key at --tls-cert/--tls-key "
+                        "before serving (no openssl binary required)")
```
mitchell-as: Let's remove these and add a cert dir constant at the top that points to "test/ssl" by default. This change will probably affect the signature of `generate_self_signed()` above.

scripts/orgkeyserver.py
```diff
@@ -98,6 +174,9 @@ def main():
         raw_key = secrets.token_bytes(KEY_SIZE)
         print("--key", base64.standard_b64encode(raw_key).decode("ascii"), file=sys.stderr)
 
+    if args.gen_cert:
+        generate_self_signed(args.tls_cert, args.tls_key, args.host)
```
mitchell-as: Let's make this the default.

activestate.yaml
```diff
@@ -470,21 +471,50 @@ scripts:
...
         -keyout test/ssl/key.pem -out test/ssl/cert.pem \
         -subj "/CN=localhost" \
         -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" 2> /dev/null
-        
+
       echo "Starting org key server."
       if [ -z "$key" ]; then
         echo "Using random encryption key (see below)"
-        python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org $org
+        python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org "$org"
```
mitchell-as: We won't need `--tls-cert` and `--tls-key` anymore.

activestate.yaml
```diff
@@ -470,21 +471,50 @@ scripts:
...
       else
         python3 -- scripts/orgkeyserver.py --tls-cert test/ssl/cert.pem --tls-key test/ssl/key.pem --org "$org" --key "$key"
```
mitchell-as: We won't need `--tls-cert` and `--tls-key` anymore.

activestate.yaml
```diff
@@ -470,21 +471,50 @@ scripts:
...
+      if not defined PY (where python3 >nul 2>nul && set "PY=python3")
+      if not defined PY (
+        echo Error: no Python interpreter found on PATH ^(tried python, py, python3^).
+        exit /b 1
+      )
+
+      echo Generating HTTPS certificate and starting org key server.
+      if "%key%"=="" (
+        echo Using random encryption key ^(see below^)
+        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%"
```
mitchell-as: We won't need `--tls-cert`, `--tls-key`, or `--gen-cert` anymore.

activestate.yaml
```diff
@@ -470,21 +471,50 @@ scripts:
...
+      ) else (
+        %PY% scripts\orgkeyserver.py --gen-cert --tls-cert test\ssl\cert.pem --tls-key test\ssl\key.pem --org "%org%" --key "%key%"
```
mitchell-as: We won't need `--tls-cert`, `--tls-key`, or `--gen-cert` anymore.

mitchell-as (changes requested):

activestate.yaml
```diff
@@ -470,21 +471,50 @@ scripts:
...
+
       echo "Generating HTTPS certificate."
       if [ ! -d test/ssl ]; then mkdir test/ssl; fi
       openssl req -x509 -newkey rsa:2048 -nodes -days 365 \
         -keyout test/ssl/key.pem -out test/ssl/cert.pem \
         -subj "/CN=localhost" \
         -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" 2> /dev/null
-        
+
       echo "Starting org key server."
```
mitchell-as: Since we're using Python to generate certs, we can remove this.

scripts/orgkeyserver.py
icanhasmath: Done in dcc45c26f — struck the openssl instructions; the script now always generates the cert in Python.

scripts/orgkeyserver.py
icanhasmath: Done in dcc45c26f — removed the --tls-cert/--tls-key args, added a CERT_DIR="test/ssl" constant at the top; generate_self_signed(host) now writes there and returns the paths.

scripts/orgkeyserver.py
icanhasmath: Done in dcc45c26f — removed the try/except; cryptography is imported directly and assumed present (bundled in the runtime).

activestate.yaml
icanhasmath: Done in dcc45c26f — dropped --tls-cert/--tls-key/--gen-cert from the cmd invocation.

activestate.yaml
icanhasmath: Done in dcc45c26f — removed the openssl cert-prep/mkdir lines; the script handles cert generation into test/ssl itself.

scripts/orgkeyserver.py
icanhasmath: Done in dcc45c26f — cert generation is now the default on every run (removed --gen-cert).

activestate.yaml
icanhasmath: Done in dcc45c26f — dropped --tls-cert/--tls-key from the bash invocation.

activestate.yaml
icanhasmath: Done in dcc45c26f — dropped --tls-cert/--tls-key from the bash invocation.

activestate.yaml
icanhasmath: Done in dcc45c26f — dropped --tls-cert/--tls-key/--gen-cert from the cmd invocation.

icanhasmath:
Addressed the review in dcc45c26f — the script now **always generates its cert in Python**, so openssl is gone entirely:

- Added a `CERT_DIR = "test/ssl"` constant; `generate_self_signed(host)` writes `cert.pem`/`key.pem` there and returns the paths.
- Cert generation is now the default on every run — removed `--gen-cert`, `--tls-cert`, and `--tls-key`.
- Removed the `ImportError` guard; `cryptography` is assumed present (bundled in the runtime).
- Both `activestate.yaml` variants now just run the script with `--org` (and optional `--key`); the openssl/mkdir prep is gone.

Verified: default `python3 scripts/orgkeyserver.py --org <org>` auto-creates `test/ssl/{cert,key}.pem` (key `0600`, SAN `localhost`+`127.0.0.1`) and serves `/v1/org-key`.

Config note: the CA path is now fixed, so `state config set privateingredient.key_service_ca test/ssl/cert.pem`.

scripts/orgkeyserver.py
```diff
@@ -4,26 +4,18 @@
...
 Run:
 
-    python3 scripts/orgkeyserver.py --tls-cert cert.pem --tls-key key.pem
+    python3 scripts/orgkeyserver.py
 
 Point the State Tool at it (the base URL only; the tool appends /v1/org-key):
 
     state config set privateingredient.key_service_url https://127.0.0.1:8443
-    state config set privateingredient.key_service_ca   /path/to/cert.pem
+    state config set privateingredient.key_service_ca   test/ssl/cert.pem
```
mitchell-as: Maybe try /path/to/test/ssl/cert.pem to be more explicit. We should use absolute paths.

mitchell-as (changes requested):

mitchell-as:
@icanhasmath If you want you can update the top line of activestate.yaml to use bac333be-be69-4334-bfeb-f326f49af681 (top commit at https://platform.activestate.com/ActiveState/cli/releases). My brief testing is showing that `state run serve-org-keys <org>` works with the new runtime.

mitchell-as (changes requested):
Please see previous requested documentation change. Also, see previous comment about maybe using the new ActiveState/cli commit that uses a new runtime.

scripts/orgkeyserver.py
icanhasmath: Done — `generate_self_signed()` now resolves the cert dir via `os.path.abspath()`, so it always returns absolute paths, and the server prints `key_service_ca <absolute path>` on startup for copy-paste into the config. Updated the docstring example to `/path/to/test/ssl/cert.pem`.

mitchell-as (approved):

PR #3842, opened 2026-07-15
Add machine-wide (all users) config layer without shared auth

icanhasmath:
## Summary

Adds a read-only, **machine-wide config layer** that applies to every user on a machine, while keeping authentication strictly **per-user**.

Motivation: administrators want to install config settings that apply to all users on a machine, without sharing auth credentials across those users.

## How it works

An optional YAML file at a system location (not rooted in any user's home dir):

| OS | Path |
|---|---|
| Linux | `/etc/activestate/cli-<channel>/config.yaml` |
| macOS | `/Library/Application Support/activestate/cli-<channel>/config.yaml` |
| Windows | `%PROGRAMDATA%\activestate\cli-<channel>\config.yaml` |

The location can be overridden with `ACTIVESTATE_CLI_SYSTEM_CONFIGDIR` (mainly for testing).

Example `config.yaml`:
```yaml
report.errors: false
autoupdate: false
api.host: "https://api.internal.example.com"
```

**Precedence** (`Instance.Get`): user value > machine-wide value > built-in default. Machine-wide values act as defaults a user can still override locally.

## Auth is never shared — by design

The machine-wide layer only serves **registered** config options. Because the auth token (`apiToken`) is *not* a registered option, credentials can never be read from the shared file. This is structural, not a filter — even an `apiToken:` entry in the shared file is ignored.

## Notes / decisions

- `IsSet(key)` still reflects only the user's own store; the machine-wide layer behaves like a default, not an explicit set.
- The file is read once at startup; a missing or unparseable file logs and is treated as empty (never blocks CLI startup).

## Testing

- Unit tests cover: machine-wide default applied, user override winning, missing-file fallback, and the auth-never-leaks guarantee (`internal/config/instance_test.go`).
- Verified end-to-end with locally-built Linux and cross-compiled Windows binaries.

### DIFF
--- a/internal/config/instance.go
+++ b/internal/config/instance.go
@@ -29,6 +29,10 @@ type Instance struct {
 	closeThread bool
 	db          *sql.DB
 	closed      bool
+	// systemConfig holds machine-wide (all users) config values, loaded read-only at startup.
+	// It only ever provides values for registered config options, which structurally excludes
+	// credentials such as the auth token (apiToken is not a registered option).
+	systemConfig map[string]interface{}
 }
 
 func New() (*Instance, error) {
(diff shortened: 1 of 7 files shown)


icanhasmath:
Superseded by #3843, which takes a different approach: instead of a machine-wide config file, all config options can be set via environment variables (canonical ACTIVESTATE_CONFIG_* names, with the existing bespoke vars kept as aliases). Auth stays out of scope since credentials are not registered config options.

PR #3843, opened 2026-07-15
ENG-2005-Environment-variable overrides for all config options

icanhasmath:
## Summary

Every registered config option can now be overridden by an environment variable, so configuration can be applied **machine-wide (all users)** without a shared config file and without sharing auth. Supersedes #3842 (the config-file approach).

## Canonical + legacy env vars

Each option gets a **canonical** override auto-derived from its key (`ACTIVESTATE_CONFIG_<KEY>`, dots/dashes → underscores, uppercased):

| Config key | Canonical env var |
|---|---|
| `autoupdate` | `ACTIVESTATE_CONFIG_AUTOUPDATE` |
| `api.host` | `ACTIVESTATE_CONFIG_API_HOST` |
| `update.info.endpoint` | `ACTIVESTATE_CONFIG_UPDATE_INFO_ENDPOINT` |
| `privateingredient.mtls_cert` | `ACTIVESTATE_CONFIG_PRIVATEINGREDIENT_MTLS_CERT` |
| …all 22 | `ACTIVESTATE_CONFIG_*` |

The five options that already had bespoke env vars keep working as **aliases** (`ACTIVESTATE_API_HOST`, `ACTIVESTATE_CLI_ANALYTICS_PIXEL`, `ACTIVESTATE_NOTIFICATIONS_OVERRIDE`, `ACTIVESTATE_TEST_UPDATE_URL`, `ACTIVESTATE_TEST_UPDATE_INFO_URL`). The canonical variable wins if both are set.

## Behavior

- Resolved in `config.Instance.Get` with precedence **env > stored > default**, so overrides take effect everywhere, not just in display.
- Values are coerced to the option's type (bool/int/string/enum).
- Only **registered** options are eligible — credentials such as `apiToken` are not registered options, so they stay out of scope (no shared auth).

## `state config` visibility

- Reports the effective value and a **Source** column showing the overriding env var name, `local` (user-set), or `default`.
- JSON output (`state config --output json`) gains `source`, `env` (active var), and `envVar` (canonical var) fields.

## Operational note

`state-svc` must inherit these env vars to honor keys it consumes (e.g. `notifications.endpoint`) — set them system-wide (`/etc/environment`, service environment) so the daemon sees them too.

## Testing

- New unit tests: canonical derivation, alias precedence, type coercion, empty-var handling ([registry_test.go](internal/mediators/config/registry_test.go)); effective-value/source and get behavior ([env_test.go](internal/runners/config/env_test.go)).
- Verified end-to-end with a real binary across bool/int/string keys, canonical and legacy vars.

### DIFF
--- a/internal/config/instance.go
+++ b/internal/config/instance.go
@@ -180,9 +180,7 @@ func (i *Instance) rawGet(key string) interface{} {
 func (i *Instance) Get(key string) interface{} {
 	opt := mediator.GetOption(key)
 
-	// An environment variable override takes precedence over any stored or default value, so that
-	// config can be applied machine-wide (all users) via the environment. Only registered options
-	// can be overridden this way, which keeps unregistered keys (e.g. credentials) out of scope.
+	// An environment variable override takes precedence over any stored or default value.
 	if mediator.KnownOption(opt) {
 		if value, _, ok := mediator.EnvOverride(opt); ok {
 			return value
--- a/internal/config/instance.go
+++ b/internal/config/instance.go
@@ -178,11 +178,22 @@ func (i *Instance) rawGet(key string) interface{} {
 }
 
 func (i *Instance) Get(key string) interface{} {
+	opt := mediator.GetOption(key)
+
+	// An environment variable override takes precedence over any stored or default value, so that
+	// config can be applied machine-wide (all users) via the environment. Only registered options
+	// can be overridden this way, which keeps unregistered keys (e.g. credentials) out of scope.
+	if mediator.KnownOption(opt) {
+		if value, _, ok := mediator.EnvOverride(opt); ok {
+			return value
+		}
+	}
+
 	result := i.rawGet(key)
 	if result != nil {
 		return result
 	}
-	if opt := mediator.GetOption(key); mediator.KnownOption(opt) {
+	if mediator.KnownOption(opt) {
 		return mediator.GetDefault(opt)
 	}
 	return nil
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1547,6 +1547,10 @@ default:
   other: Default
 source:
   other: Source
+config_source_local:
+  other: local
+config_source_default:
+  other: default
 vulnerabilities:
   other: Vulnerabilities (CVEs)
 dependency_row:
--- a/internal/locale/locales/en-us.yaml
+++ b/internal/locale/locales/en-us.yaml
@@ -1545,6 +1545,8 @@ value:
   other: Value
 default:
   other: Default
+source:
+  other: Source
 vulnerabilities:
   other: Vulnerabilities (CVEs)
 dependency_row:
--- a/internal/mediators/config/registry.go
+++ b/internal/mediators/config/registry.go
@@ -1,6 +1,24 @@
 package config
 
-import "sort"
+import (
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cast"
+)
+
+// EnvVarPrefix is prepended to a config key to derive its canonical environment-variable override.
+const EnvVarPrefix = "ACTIVESTATE_CONFIG_"
+
+var envVarReplacer = strings.NewReplacer(".", "_", "-", "_")
+
+// CanonicalEnvVarName returns the environment variable that overrides the given config key. Every
+// registered config option can be overridden this way, e.g. "update.info.endpoint" maps to
+// "ACTIVESTATE_CONFIG_UPDATE_INFO_ENDPOINT".
+func CanonicalEnvVarName(key string) string {
+	return EnvVarPrefix + strings.ToUpper(envVarReplacer.Replace(key))
+}
 
 type Type int
 
@@ -25,9 +43,13 @@ var EmptyEvent = func(value interface{}) (interface{}, error) {
 
 // Option defines what a config value's name and type should be, along with any get/set events
 type Option struct {
-	Name         string
-	Type         Type
-	Default      interface{}
+	Name    string
+	Type    Type
+	Default interface{}
+	// EnvAliases are additional (legacy/bespoke) environment variables that override this option, in
+	// addition to its canonical CanonicalEnvVarName. They are kept for backwards compatibility with
+	// env vars that predate the canonical ACTIVESTATE_CONFIG_* scheme (e.g. ACTIVESTATE_API_HOST).
+	EnvAliases   []string
 	GetEvent     Event
 	SetEvent     Event
 	isRegistered bool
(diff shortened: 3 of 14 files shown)


internal/mediators/config/registry.go
```diff
@@ -52,28 +74,68 @@ func NewEnum(options []string, default_ string) *Enums {
...
+func coerceToType(t Type, raw string) interface{} {
+	switch t {
+	case Bool:
+		return cast.ToBool(raw)
+	case Int:
+		return cast.ToInt(raw)
+	default: // String, Enum
+		return raw
+	}
 }
```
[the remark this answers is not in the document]
icanhasmath: Good catch — fixed in 19678dc. `EnvOverride` now strictly coerces the value to the option's type via `cast.ToBoolE`/`ToIntE` and validates enum values against the allowed set. Anything that doesn't parse (e.g. `ACTIVESTATE_CONFIG_AUTOUPDATE=notabool` or an out-of-set enum) is ignored, so it falls back to the stored/default value instead of silently becoming false/0. This mirrors the validation `state config set` already does. Added tests in `registry_test.go` (TestEnvOverrideInvalidIgnored) and verified end-to-end: an invalid bool leaves `autoupdate` at its default, an invalid enum leaves `security.prompt.level` at `critical`.

Note: I kept the fallback silent rather than logging, because importing multilog/logging into the mediators/config package would create an import cycle (rollbar/logging register config options). The Source column showing `local`/`default` instead of the env var name signals that the override didn't take effect.

internal/locale/locales/en-us.yaml
```diff
@@ -1545,6 +1545,8 @@ value:
   other: Value
 default:
   other: Default
+source:
+  other: Source
```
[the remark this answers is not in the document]
icanhasmath: Fixed in 19678dc — added `config_source_local` and `config_source_default` to en-us.yaml so the Source column no longer relies on the Tl fallback.

icanhasmath:
Thanks for the review. Addressed both Copilot findings in 19678dc:

1. **Invalid env values** — `EnvOverride` now strictly validates/coerces each variable to the option's type (`cast.ToBoolE`/`ToIntE`, and enum membership check). Unparseable bools/ints and out-of-set enums are ignored and fall back to the stored/default value instead of silently becoming `false`/`0`. Kept the fallback silent to avoid an import cycle (logging packages register config options); the Source column showing `local`/`default` rather than the var name makes a non-applied override visible. Added `TestEnvOverrideInvalidIgnored`.
2. **Missing locale IDs** — added `config_source_local` and `config_source_default` to en-us.yaml.

All config tests pass and I re-verified end-to-end (invalid bool leaves `autoupdate` at default; invalid enum leaves `security.prompt.level` at `critical`; valid values still apply).

mitchell-as:
There's a regression in `state config`: (https://github.com/ActiveState/cli/actions/runs/29457176664/job/87492845470?pr=3843#step:27:920)

```
TestConfigIntegrationTestSuite/TestList (300ms)
      session.go:339: 
          'Expected: "true*": consumer stopped, reason: pty closed':
              'consumer stopped, reason: pty closed': pty closed
          Stack:
          ./internal/testhelpers/e2e/session.go:513:e2e.(*Session).DebugMessage
          ./internal/testhelpers/e2e/session.go:339:e2e.(*Session).newSpawnOpts.func1
          ./vendor/github.com/ActiveState/termtest/expect.go:77:termtest.(*TermTest).ExpectErrorHandler
          ./vendor/github.com/ActiveState/termtest/expect.go:95:termtest.(*TermTest).ExpectCustom
          ./vendor/github.com/ActiveState/termtest/expect.go:104:termtest.(*TermTest).Expect
          ./test/integration/config_int_test.go:113:integration.(*ConfigIntegrationTestSuite).TestList

Snapshot for Cmd '/tmp/530567214/bin/state config set optin.buildscripts true':
          █ Set config value                                                                                                                          
                                                                                                                                                      
        
Snapshot for Cmd '/tmp/530567214/bin/state config':
          █ Listing Configuration Keys and Values                                                                                                     
                                                                                                                                                      
            Key                                    Value         Source     Default                                                                   
            api.host                               ""            default    ""                                                                        
            autostart.svc                          true          default    true                                                                      
            autoupdate                             true          default    true                                                                      
            notifications.endpoint                 ""            default    ""                                                                        
            optin.buildscripts                     true          local      false                                                                     
```

The test is expecting that last line to have `true*` due to the previous `state config set optin.buildscripts true` command.

icanhasmath:
> The test is expecting that last line to have `true*` due to the previous `state config set optin.buildscripts true` command.

Diagnosis
Your PR intentionally replaced the old * marker (which appended * to any value differing from default) with a proper Source column — renderConfigValue no longer appends *, and renderConfigSource renders environment (bold env-var name) / local / default. 

TestList is stale: after state config set optin.buildscripts true, the row is now optin.buildscripts  true  local  false — value true, Source local — instead of true*. The test still asserts cp.Expect("true*"), which can never match now. 

@mitchell-as do you agree with this - and therefor the fix is to update the test?

mitchell-as:
Yes, I would update the text to expect "local" instead of "true*"

mitchell-as:
@icanhasmath  Furthermore, I would add a new integration test that includes a test with environment as the source. This will test your new improvement. You can model it after the test you just fixed.

internal/mediators/config/registry.go
```diff
@@ -52,28 +74,95 @@ func NewEnum(options []string, default_ string) *Enums {
...
 
 // Registers a hidden config option without get/set events.
 func RegisterHiddenOption(key string, t Type, defaultValue interface{}) {
-	registerOption(key, t, defaultValue, EmptyEvent, EmptyEvent, true)
+	registerOption(key, t, defaultValue, nil, EmptyEvent, EmptyEvent, true)
+}
+
+// RegisterOptionWithEnv registers a config option along with one or more legacy/bespoke environment
+// variables that override it, in addition to its canonical ACTIVESTATE_CONFIG_* variable.
+func RegisterOptionWithEnv(key string, t Type, defaultValue interface{}, envAliases ...string) {
```
mitchell-as: Since envAliases is optional, I think we can fold this into the existing `RegisterOption()`. No need for a separate function.

Make sure we keep the comment that envAliases is only for legacy env vars that do not have the "ACTIVESTATE_CONFIG_*" prefix.

internal/mediators/config/registry.go
```diff
@@ -1,6 +1,24 @@
...
+)
+
+// EnvVarPrefix is prepended to a config key to derive its canonical environment-variable override.
+const EnvVarPrefix = "ACTIVESTATE_CONFIG_"
+
+var envVarReplacer = strings.NewReplacer(".", "_", "-", "_")
+
+// CanonicalEnvVarName returns the environment variable that overrides the given config key. Every
+// registered config option can be overridden this way, e.g. "update.info.endpoint" maps to
+// "ACTIVESTATE_CONFIG_UPDATE_INFO_ENDPOINT".
```
mitchell-as: Can we replace this comment with a config option we'd actually override with an env var? This isn't one of them. Maybe use one of the config option that triggered this PR in the first place.

internal/config/instance.go
```diff
@@ -178,11 +178,22 @@ func (i *Instance) rawGet(key string) interface{} {
 }
 
 func (i *Instance) Get(key string) interface{} {
+	opt := mediator.GetOption(key)
+
+	// An environment variable override takes precedence over any stored or default value, so that
+	// config can be applied machine-wide (all users) via the environment. Only registered options
+	// can be overridden this way, which keeps unregistered keys (e.g. credentials) out of scope.
```
mitchell-as: Let's just keep this at "An environment variable override takes precedence over any stored or default value."

State Tool is designed for individual use, not a single, machine-wide instance. We don't want to hint the latter in the comments.

internal/mediators/config/registry_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCanonicalEnvVarName(t *testing.T) {
+	assert.Equal(t, "ACTIVESTATE_CONFIG_API_HOST", CanonicalEnvVarName("api.host"))
+	assert.Equal(t, "ACTIVESTATE_CONFIG_AUTOUPDATE", CanonicalEnvVarName("autoupdate"))
+	assert.Equal(t, "ACTIVESTATE_CONFIG_UPDATE_INFO_ENDPOINT", CanonicalEnvVarName("update.info.endpoint"))
```
mitchell-as: nit: we wouldn't actually override this one. Can we pick a different one that has double dots? How about "security.prompt.level"?

internal/runners/config/get.go
```diff
@@ -23,6 +23,9 @@ func NewGet(prime primeable) *Get {
 
 func (g *Get) Run(params GetParams) error {
 	key := params.Key.String()
+
+	// cfg.Get already applies any environment-variable override for the key, so the value reported
+	// here is the one the State Tool will actually use.
```
mitchell-as: We don't need this comment.

mitchell-as (changes requested):
This is looking pretty good. There are some changes I'd like to see though. See below.

internal/mediators/config/registry.go
icanhasmath: Done in 5fcf95d0f — folded into `RegisterOption` via a variadic `envAliases ...string` and removed `RegisterOptionWithEnv`. Existing no-alias callers are unchanged. Kept the note that envAliases are only for legacy env vars that don't use the ACTIVESTATE_CONFIG_* prefix (on both `RegisterOption` and the `EnvAliases` field).

internal/mediators/config/registry.go
icanhasmath: Good point — changed the example to `api.host` -> `ACTIVESTATE_CONFIG_API_HOST` in 5fcf95d0f.

internal/config/instance.go
icanhasmath: Done in 5fcf95d0f — trimmed to just "An environment variable override takes precedence over any stored or default value." and dropped the machine-wide framing. (The registered-only scoping that keeps credentials out is still enforced by the `mediator.KnownOption(opt)` guard right below.)

internal/mediators/config/registry_test.go
icanhasmath: Done in 5fcf95d0f — replaced with `security.prompt.level` -> `ACTIVESTATE_CONFIG_SECURITY_PROMPT_LEVEL` (real user-facing option, double dots).

internal/runners/config/get.go
icanhasmath: Removed in 5fcf95d0f.

icanhasmath:
@mitchell-as thanks for the review — all five addressed in 5fcf95d0f:

1. Folded `RegisterOptionWithEnv` into a variadic `RegisterOption(key, type, default, envAliases ...string)` and removed the separate function; kept the legacy-alias note.
2. Doc example → `api.host` -> `ACTIVESTATE_CONFIG_API_HOST`.
3. Trimmed the `Instance.Get` comment to just the env-precedence sentence (no machine-wide framing).
4. Test example → `security.prompt.level` instead of the test-only update endpoint.
5. Removed the unnecessary comment in `config get`.

Build and all config unit tests pass. Ready for another look.

mitchell-as (approved):

PR #3845, opened 2026-07-17
CS-2357-Declare asInvoker execution level for remote installer

icanhasmath:
## Problem

`state-remote-installer.exe` fires a Windows UAC elevation prompt on launch. When an admin authenticates at that prompt, the State Tool ends up installed under the **admin** account instead of the target (invoking) user. Reported by David Cartin / KSA — see epic [CS-2338](https://activestatef.atlassian.net/browse/CS-2338).

## Root cause

The remote installer ships with **no embedded Windows application manifest**, so it declares no `requestedExecutionLevel`. Windows' legacy UAC *installer-detection* heuristic auto-flags executables whose filename contains "install"/"setup"/"update" for elevation **unless** the binary explicitly declares `asInvoker`. Because `state-remote-installer.exe` matches that heuristic and had no manifest, Windows prompts for elevation.

Before this change:
- `cmd/state-remote-installer/versioninfo.json` had `"ManifestPath": ""` — `goversioninfo` embeds no manifest when this is empty.
- No `.manifest` file existed anywhere in the project's own source.

## Fix

- Add `cmd/state-remote-installer/state-remote-installer.exe.manifest` declaring `<requestedExecutionLevel level="asInvoker" uiAccess="false" />`.
- Point `ManifestPath` in `versioninfo.json` at that manifest so `goversioninfo` (`go generate`, run by the `build-remote-installer` step) embeds it into `resource.syso` at build time.

This opts the binary out of installer-detection auto-elevation, so it runs with the invoking user's token and restores per-user install behavior. The version-stamping generator (`scripts/versioninfo-generator`) preserves `ManifestPath` through its struct, so version stamping is unaffected.

## Verification

Built on Linux/WSL here, so the Windows resource path could not be exercised locally. On a Windows build host:

```
cd cmd/state-remote-installer && go generate   # regenerates resource.syso
go build .
sigcheck -m state-remote-installer.exe          # confirm asInvoker block present
```

Then confirm launching as a standard user no longer triggers UAC.

## Follow-up (not in this PR)

`state-installer.exe` (`cmd/state-installer`) similarly has no manifest and is subject to the same heuristic — worth confirming whether it needs the same treatment.

Jira: [CS-2357](https://activestatef.atlassian.net/browse/CS-2357) (under epic [CS-2338](https://activestatef.atlassian.net/browse/CS-2338))

[CS-2338]: https://activestatef.atlassian.net/browse/CS-2338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CS-2357]: https://activestatef.atlassian.net/browse/CS-2357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,10 @@ cmd/state/versioninfo.json
 cmd/state/resource.syso
 cmd/state-svc/versioninfo.json
 cmd/state-svc/resource.syso
+cmd/state-installer/versioninfo.json
+cmd/state-installer/resource.syso
+cmd/state-exec/versioninfo.json
+cmd/state-exec/resource.syso
+cmd/state-mcp/versioninfo.json
+cmd/state-mcp/resource.syso
 test/ssl/
--- a/activestate.yaml
+++ b/activestate.yaml
@@ -94,6 +94,9 @@ scripts:
       if [[ "$GOOS" == "windows" || "$OS" == "Windows_NT" ]]; then
         go run scripts/versioninfo-generator/main.go version.txt cmd/state/versioninfo.json "State Tool"
         go run scripts/versioninfo-generator/main.go version.txt cmd/state-svc/versioninfo.json "State Service"
+        go run scripts/versioninfo-generator/main.go version.txt cmd/state-installer/versioninfo.json "State Installer"
+        go run scripts/versioninfo-generator/main.go version.txt cmd/state-exec/versioninfo.json "State Executor"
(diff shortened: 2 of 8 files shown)


mitchell-as (approved):
This looks fine to me *crosses fingers*

PR #3851, opened 2026-07-21
ENG-1924: Promote staging files without adding a release/ prefix

mitchell-as:
[ENG-1924: Promote staging files without adding a release/ prefix](https://activestatef.atlassian.net/browse/ENG-1924)

The promote-staging-to-production step was moving beta release files to release/update/state/beta/ instead of update/state/beta/, which forced a manual copy in the S3 console after every promotion. The promoter stripped the staging/ prefix and then prepended release/, so files landed one directory too deep. Because the updater only ever fetches from update/state/<channel>/, this actually misplaced files for every channel, not just beta.

The fix drops the release/ prefix so promotion maps staging/<key> straight to <key>; the channel (beta, release, ...) is already part of the path. Beta files now land in update/state/beta/ and release-channel files in update/state/release/, matching what the updater serves.

Base branch: targets version/0-48-1-RC3.

### DIFF
--- a/scripts/ci/s3-promoter/main.go
+++ b/scripts/ci/s3-promoter/main.go
@@ -75,8 +75,7 @@ func run() {
 	// Copy each staging object to production location and delete the staging version
 	for _, obj := range stagingObjects {
 		stagingKey := *obj.Key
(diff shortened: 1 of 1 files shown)


mitchell-as:
Tested this when promoting RC3 to production and it worked.

PR #3853, opened 2026-09-11
Try to avoid macOS integration test failures due to gotestfmt crash.

mitchell-as:


mitchell-as:
Integration test failures are known issues and unrelated to this PR.

The important thing is that the integration tests on macos-latest ran instead of errored, so the problem is resolved.

